PR #851, opened 2015-07-28
Generalize LP.Join to take an arbitrary condition.

### LINKED ISSUE #847: conjunction of two equi-joins is itself an equi-join
Query works correctly:

```sql
SELECT c1.name, c2.city
 FROM table1 c1
 JOIN table2 c2 ON c1.userId = c2.userId;
```

Query never returns:

```sql
SELECT c1.name, c2.city
 FROM table1 c1
 JOIN table2 c2 ON c1.userId = c2.userId AND c1.userId = 101;
```

sellout:
Fixes #847.

The compiler is simplified to merely compile the JoinRelation clause,
while there is a new (optional) LP phase to “align” the
condition (previously done by the compiler). And then the mongo planner
separates the list of expressions for the left and right sides of the join.

Also,
* add `cataM`,
* “fix” lpParaZygoHistoM to match other zygoHisto constructions, and
* change the default it config to use mongodb_3_0.

### DIFF
--- a/core/src/main/scala/slamdata/engine/logicalplan.scala
+++ b/core/src/main/scala/slamdata/engine/logicalplan.scala
@@ -39,8 +39,8 @@ object LogicalPlan {
       fa match {
         case x @ ReadF(_) => G.point(x)
         case x @ ConstantF(_) => G.point(x)
-        case JoinF(left, right, tpe, rel, lproj, rproj) =>
-          G.apply4(f(left), f(right), f(lproj), f(rproj))(JoinF(_, _, tpe, rel, _, _))
+        case JoinF(left, right, tpe, rel) =>
+          G.apply3(f(left), f(right), f(rel))(JoinF(_, _, tpe, _))
         case InvokeF(func, values) => G.map(Traverse[List].sequence(values.map(f)))(InvokeF(func, _))
         case x @ FreeF(_) => G.point(x)
         case LetF(ident, form0, in0) =>
@@ -52,8 +52,8 @@ object LogicalPlan {
       v match {
         case x @ ReadF(_) => x
         case x @ ConstantF(_) => x
-        case JoinF(left, right, tpe, rel, lproj, rproj) =>
-          JoinF(f(left), f(right), tpe, rel, f(lproj), f(rproj))
+        case JoinF(left, right, tpe, rel) =>
+          JoinF(f(left), f(right), tpe, f(rel))
         case InvokeF(func, values) => InvokeF(func, values.map(f))
         case x @ FreeF(_) => x
         case LetF(ident, form, in) => LetF(ident, f(form), f(in))
@@ -64,8 +64,8 @@ object LogicalPlan {
       fa match {
         case x @ ReadF(_) => F.zero
         case x @ ConstantF(_) => F.zero
-        case JoinF(left, right, tpe, rel, lproj, rproj) =>
-          F.append(F.append(f(left), f(right)), F.append(f(lproj), f(rproj)))
+        case JoinF(left, right, tpe, rel) =>
+          F.append(F.append(f(left), f(right)), f(rel))
         case InvokeF(func, values) => Foldable[List].foldMap(values)(f)
         case x @ FreeF(_) => F.zero
         case LetF(_, form, in) => {
@@ -78,8 +78,7 @@ object LogicalPlan {
       fa match {
         case x @ ReadF(_) => z
         case x @ ConstantF(_) => z
-        case JoinF(left, right, tpe, rel, lproj, rproj) =>
-          f(left, f(right, f(lproj, f(rproj, z))))
+        case JoinF(left, right, tpe, rel) => f(left, f(right, f(rel, z)))
         case InvokeF(func, values) => Foldable[List].foldRight(values, z)(f)
         case x @ FreeF(_) => z
         case LetF(ident, form, in) => f(form, f(in, z))
@@ -93,7 +92,7 @@ object LogicalPlan {
     override def render(v: LogicalPlan[_]) = v match {
       case ReadF(name)                 => Terminal("Read" :: nodeType, Some(name.pathname))
       case ConstantF(data)             => Terminal("Constant" :: nodeType, Some(data.toString))
-      case JoinF(_, _, tpe, rel, _, _) => Terminal("Join" :: nodeType, Some(tpe.toString + ", " + rel))
+      case JoinF(_, _, tpe, _) => Terminal("Join" :: nodeType, Some(tpe.toString))
       case InvokeF(func, _     )       => Terminal(func.mappingType.toString :: "Invoke" :: nodeType, Some(func.name))
       case FreeF(name)                 => Terminal("Free" :: nodeType, Some(name.toString))
       case LetF(ident, _, _)           => Terminal("Let" :: nodeType, Some(ident.toString))
@@ -103,9 +102,9 @@ object LogicalPlan {
     def equal[A](v1: LogicalPlan[A], v2: LogicalPlan[A])(implicit A: Equal[A]): Boolean = (v1, v2) match {
       case (ReadF(n1), ReadF(n2)) => n1 == n2
       case (ConstantF(d1), ConstantF(d2)) => d1 == d2
-      case (JoinF(l1, r1, tpe1, rel1, lproj1, rproj1),
-            JoinF(l2, r2, tpe2, rel2, lproj2, rproj2)) =>
-        A.equal(l1, l2) && A.equal(r1, r2) && A.equal(lproj1, lproj2) && A.equal(rproj1, rproj2) && tpe1 == tpe2
+      case (JoinF(l1, r1, tpe1, rel1),
+            JoinF(l2, r2, tpe2, rel2)) =>
+        A.equal(l1, l2) && A.equal(r1, r2) && A.equal(rel1, rel2) && tpe1 == tpe2
       case (InvokeF(f1, v1), InvokeF(f2, v2)) => Equal[List[A]].equal(v1, v2) && f1 == f2
       case (FreeF(n1), FreeF(n2)) => n1 == n2
       case (LetF(ident1, form1, in1), LetF(ident2, form2, in2)) =>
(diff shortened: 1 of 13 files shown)


core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -2400,9 +2396,9 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
       planLog("select city from zips").map(_.map(_.name)).toEither must
         beRight(Vector(
           "SQL AST", "Variables Substituted", "Annotated Tree",
-          "Logical Plan", "Simplified", "Logical Plan (projections preferred)",
-          "Workflow Builder", "Workflow (raw)", "Workflow (finished)",
-          "Physical Plan", "Mongo"))
+          "Logical Plan", "Simplified", "Logical Plan (aligned joins)",
+          "Logical Plan (projections preferred)", "Workflow Builder",
+          "Workflow (raw)", "Workflow (finished)", "Physical Plan", "Mongo"))
     }
```
jdegoes: Would be good to get a new planner test here to verify the multiple conditions case generates the correct plan.

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -814,12 +865,9 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
             Vector.empty -> sa.run(ng)))
       }
 
-    val annotateƒ = zipPara(
-      selectorƒ[OutputM[WorkflowBuilder]],
-      liftPara(jsExprƒ[OutputM[WorkflowBuilder]]))
-
     (for {
-      prep <- log("Logical Plan (projections preferred)")(Optimizer.preferProjections(logical).point[M])
+      align <- log("Logical Plan (aligned joins)")(swizzle(StateT[Error \/ ?, NameGen, Term[LogicalPlan]](s => logical.cataM[PlannerError \/ ?, Term[LogicalPlan]](alignJoinsƒ).map((s, _)))))
```
mossprescott: Can we get a `swizzle2` or something? This line should read as `cataM(alignJoinsƒ)` and at the moment it's all types.

core/src/main/scala/slamdata/engine/logicalplan.scala
```diff
@@ -232,20 +229,18 @@ object LogicalPlan {
   def lpParaZygoHistoM[M[_]: Monad, A, B](
     t: Term[LogicalPlan])(
     f: LogicalPlan[(Term[LogicalPlan], B)] => B,
-    g: LogicalPlan[(B, Cofree[LogicalPlan, A])] => M[A]):
+    g: LogicalPlan[Cofree[LogicalPlan, (B, A)]] => M[A]):
```
mossprescott: Did you need this type for `workflowƒ`, or just wanted it to have the correct name? Because you've made the morphism slightly more powerful, right?

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -756,6 +759,15 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
       }
     }
 
+    def splitConditions: Ann => Option[List[(Ann, Ann)]] =
+      _.tail match {
+      case InvokeF(relations.And, terms) =>
+        terms.map(splitConditions).sequence.map(_.foldLeft[List[(Ann, Ann)]](Nil)(_ ++ _))
```
mossprescott: Ug. I cringe whenever I have to write this `foldLeft(Nil)` with the type annotated. Is there no incantation that would just reduce it with List's Monoid?

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -768,25 +780,64 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
...
+          c => {
+            val (leftKeys, rightKeys) = c.unzip
+            lift((HasWorkflow(left) |@|
+              HasWorkflow(right) |@|
+              leftKeys.map(HasWorkflow).sequenceU |@|
+              leftKeys.map(HasJs).sequenceU |@|
+              rightKeys.map(HasWorkflow).sequenceU |@|
+              rightKeys.map(HasJs).sequenceU)((l, r, lk, lj, rk, rj) =>
+              lift((lj.map(_._2.map(_(comp.map(_._2))).sequenceU).sequenceU |@| rj.map(_._2.map(_(comp.map(_._2))).sequenceU).sequenceU)((largs, rargs) =>
+                join(l, r, tpe, lk, (lj zip largs).map(l => l._1._1(l._2.map(κ(JsFn.identity)))), rk, (rj zip rargs).map(r => r._1._1(r._2.map(κ(JsFn.identity))))))).join)).join
```
mossprescott: Sigh.

core/src/main/scala/slamdata/engine/logicalplan.scala
sellout: Yeah, it was necessary because we need to get the JS for a child of the condition.

If we have a condition like `l.a = r.a AND l.b = r.b`, we need both JS and Workflow versions of `l.a`, `l.b`, `r.a`, and `r.b`.

I actually didn’t realize it was the more idiomatic `zygoHisto` until I tried to figure out what to call it and look at other libraries.

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
sellout: Yeah, this is a mess … I think operads could clean some of this up. It got messier because rather than just having a `PartialJs` (which is `(Function, List)`), we now have `List[PartialJs]`. From my end it doesn’t look too bad, because it’s _much_ nicer than the interim version that built up a single JsFn and Workflow here rather than deferring that to WorkflowBuilder 😉

There probably are a couple functions I could pull out of that, though. Because we do the same ops to the left and right keys.

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
mossprescott: I would start with a couple of tests for `alignJoinsƒ`, which would be simpler and would help to make it clear what it does anyway. but a single planner test wouldn't hurt, too.

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -2415,7 +2411,8 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
       planLog("select date_part('foo', bar) from zips").map(_.map(_.name)).toEither must
         beRight(Vector(
           "SQL AST", "Variables Substituted", "Annotated Tree",
-          "Logical Plan", "Simplified", "Logical Plan (projections preferred)"))
+          "Logical Plan", "Simplified", "Logical Plan (aligned joins)",
+          "Logical Plan (projections preferred)"))
```
mossprescott: Since join alignment can fail, how about another test here, too. Basically just showing that your new swizzling does the right thing.

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
sellout: *facepalm* Yeah, that should just be `fold`.

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -768,25 +780,64 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
...
+        Term[LogicalPlan] => PlannerError \/ Term[LogicalPlan] =
+      _.unFix match {
+        case InvokeF(And, terms) =>
+          terms.map(alignCondition(lt, rt)).sequenceU.map(Invoke(And, _))
+        case InvokeF(Or, terms) =>
+          terms.map(alignCondition(lt, rt)).sequenceU.map(Invoke(Or, _))
+        case InvokeF(Not, terms) =>
+          terms.map(alignCondition(lt, rt)).sequenceU.map(Invoke(Not, _))
+        case x @ InvokeF(func: Mapping, List(left, right)) =>
+          if (Tag.unwrap(left.foldMap(x => Tags.Disjunction(x == lt))) && Tag.unwrap(right.foldMap(x => Tags.Disjunction(x == rt))))
```
mossprescott: I'm really not a fan of this Tags.Disjunction business. Is there not a simple method/function on/over List that would do the trick? Something like `any(left.map(_ == lt))`? I'm not even sure I understand what this is doing well enough to know what you need.

sellout: Aww, no fair. That’s old code 😉

You’re right, though. It’s not a list, but `Foldable` has `any`, so `left.any(_ == lt)` should do it. Much better.

core/src/test/scala/slamdata/engine/logicalplan.scala
```diff
@@ -39,8 +39,8 @@ class LogicalPlanSpecs extends Spec {
 
   def joinGen[A: Arbitrary]: Gen[LogicalPlan[A]] = for {
     tpe <- Gen.oneOf(List(Inner, LeftOuter, RightOuter, FullOuter))
-    (l, r, lproj, rproj) <- Arbitrary.arbitrary[(A, A, A, A)]
-  } yield JoinF(l, r, tpe, std.RelationsLib.Eq, lproj, rproj)
+    (l, r, comp) <- Arbitrary.arbitrary[(A, A, A)]
```
mossprescott: Not sure what this generator gets used for, but surely `comp` should at least be a boolean expression. I guess we can deal with it when we decide to do something more interesting with it.

sellout: No, not in an Arbitrary … we can add filters on Arbitraries, but this should generate something with any `A` … it’s not even fixpoint, so it _can’t_ be a boolean LP.

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
mossprescott: If it's green, it's fair game ;-) Heck, if it's white, but close enough to show up in the diff, then it's fair game.

I'm still confused, but now I'll have half a chance. This is saying, "if the (part of) the left side of the condition refers to the left side of the join and similar on the right." So what you really want to say is `left contains lt && right contains rt`. Better names wouldn't hurt, either.

core/src/test/scala/slamdata/engine/logicalplan.scala
mossprescott: Right. Well that makes sense but makes this less useful (eventually). I'm just now thinking through the implications of using `cond: A`, which suggests the generator is useful fro something already. Let's worry about it later.

mossprescott:
Is this the right time to move LogicalPlan to the new fixplate conventions (new package, etc.)?

core/src/test/scala/slamdata/engine/logicalplan.scala
sellout: It’s already useful for testing type class laws. Also, with the `Term` Arbitrary, we have a `Term[LogicalPlan]` Arbitrary, and with that we should be able to make a filtered version that type checks.

Also, Join is not really different from a function at this point (still waffling on whether to make that change), and our `InvokeF` Arbitrary does nothing to ensure that other functions type check, so why should Join be different?

sellout:
The problem I have with that is git doesn’t like renaming files (especially if you split them into pieces). So, it mucks up the diffs a ton.

core/src/main/scala/slamdata/engine/physical/mongodb/planner.scala
sellout: Simplifying this also encouraged me to make the check more exhaustive, so 👍🏾.

mossprescott:
Right, so a separate PR then. I think it's a good practice to do refactorings in separate commits anyway, although git pushes you away from that to some extent.

core/src/test/scala/slamdata/engine/logicalplan.scala
mossprescott: Well, Join has the type, so there's that.

I think we'll find that generating truly arbitrary LPs and then filtering them through type checking doesn't accomplish anything other than making our fans spin up. So this instance is *only* good for testing the type classes. One might be tempted to ask if that means that the LP type is under-constrained, but I guess we knew that and made that choice deliberately. Anyway I'm OK with this as is.

sellout: I’m not sure how efficient it would be, but I think the untypers would help with this.

mossprescott:
Push, please?

PR #854, opened 2015-07-31
Create a custom SlamData Predef.

sellout:
* Disable all automatic Scala imports,
* add a new, more restrictive `Predef`,
* rename `slamdata.java` to `slamdata.stacktrace`,
* move `Disjunction` & `Validation` matchers into `slamdata.specs2`,
* rename matchers to align with those in scalaz-specs2 (and migrate some
  tests over), and
* upgrade sbt-buildinfo.


core/src/main/scala/slamdata/Predef.scala
```diff
@@ -0,0 +1,101 @@
...
+  type Map[A, +B] = I.Map[A, B]
+  val  Map        = I.Map
+  type Set[A] = I.Set[A]
+  val  Set    = I.Set
+  type Seq[+A] = I.Seq[A]
+  type Stream[+A] = I.Stream[A]
+  val  Stream     = I.Stream
+  val  #::        = Stream.#::
+
+  def ??? = P.???
```
mossprescott: Maybe substitute @djspiewak's [fancy version](https://gist.github.com/djspiewak/a31d774ae78e398839b2) that tells you the expected type?

mossprescott: Does this even work given our draconian compiler options?

core/src/main/scala/slamdata/engine/fp/package.scala
```diff
@@ -252,7 +243,7 @@ trait SKI {
   //     be more appropriate, but the code points are larger than 2 bytes, so
   //     Scala doesn't handle them.
 
-  /** Probably not useful; implemented here mostly because it's amusing. */
+  /** Probably not useful; implemented here m*.ostly because it's amusing. */
```
mossprescott: Emacs junk.

core/src/main/scala/slamdata/engine/fp/package.scala
```diff
@@ -264,8 +255,8 @@ trait SKI {
   def κ[A, B, C, D](x: D): (A, B, C) => D                   = (_, _, _) => x
   def κ[A, B, C, D, E, F, G](x: G): (A, B, C, D, E, F) => G = (_, _, _, _, _, _) => x
 
-  /** A shorter name for the identity function. */
-  def ɩ[A]: A => A = Predef.identity
+  /** A shorter name for the identity* function. */
```
mossprescott: More emacs decoration.

core/src/main/scala/slamdata/engine/javascript/jscore.scala
```diff
@@ -353,7 +353,7 @@ final case class JsFn(base: JsCore.Ident, expr: Term[JsCore]) {
   override def toString = JsCore.toUnsafeJs(apply(JsCore.Ident("_").fix).simplify).render(0)
 
   val commonBase = JsCore.Ident("$")
-  override def equals(obj: Any) = obj match {
+  override def equals(obj: scala.Any) = obj match {
```
mossprescott: :-)

mossprescott:
I like the way this simplifies our imports and makes it more obvious when we're stepping outside our usual style, but it's definitely in the category of "making things less obvious for newcomers". I suggest two things to mitigate this:

1. Make sure `import slamdata.engine.Predef._` is the first line of *every* source file.
2. Start a "Developer's/Contributor's Guide" doc in the project mentioning this and our compiler options.

I'd say most, if not all, of what you get with `import fp._` also belongs in our `Predef`. I'm thinking of `κ` and `ι`, the ops for `ListMap`, the instances for RenderTree/Show, etc. As long as you're touching almost every file's imports, maybe now's the time?

core/src/main/scala/slamdata/Predef.scala
sellout: Using @djspiewak’s version is a good idea.

Also, I’m not sure how problematic it is with `-Ywarn-dead-code`. I know there were `???` that triggered that warning when we added the option, but `???` does still exist in our codebase, so
¯\\\_(ツ)\_/¯

core/src/main/scala/slamdata/engine/fp/package.scala
sellout: Ugh, stupid race condition.

sellout:
I _think_ it is the first line in every file now. Considering you can’t do much without it, it has to be there.

I do have [a Dev Guide in the wiki, with a section on Predef](https://github.com/slamdata/slamengine/wiki/Scala%5BSlamData%5D). It’s still pretty skeletal, but yeah.

sellout:
Oh, and the reason I didn’t put `fp` in there yet is because it has to be moved out of `slamdata.engine` first. As do RenderTree, etc. So I wanted to do that incrementally (as well as replacing Option and List in the Predef with Maybe and IList).

core/src/main/scala/slamdata/Predef.scala
sellout: I didn’t have any luck getting Daniel’s version to work (ended up with a couple “No TypeTag available for A” errors). Happy to switch to it if we can, though.

PR #856, opened 2015-08-04
Convert Joins from an LP node to a Func.

sellout:
Now that we handle arbitrary conditions, Join doesn’t need any special
handling, so it’s simpler to handle it as functions.

Also, `Cross` is just an `InnerJoin` where the condition is constantly
`true`, so that’s how it’s implemented now.

Finally, add some fp package coverage:

* Add a simple `gpara` test,
* delete some dead code, and
* fix a bug in EnvT uncovered by the new coverage.

### DIFF
--- a/core/src/main/scala/slamdata/engine/std/set.scala
+++ b/core/src/main/scala/slamdata/engine/std/set.scala
@@ -79,7 +79,7 @@ trait SetLib extends Library {
 
   val InnerJoin = Transformation(
     "INNER JOIN",
-    "Joins two sets …",
+    "Returns a new set containing the pairs values from the two sets that satisfy the condition.",
     Type.Top :: Type.Top :: Type.Bool :: Nil,
     noSimplification,
     partialTyper {
@@ -92,7 +92,7 @@ trait SetLib extends Library {
 
   val LeftOuterJoin = Transformation(
     "LEFT OUTER JOIN",
-    "Joins two sets …",
+    "Returns a new set containing the pairs values from the two sets that satisfy the condition, plus all other values from the left set.",
     Type.Top :: Type.Top :: Type.Bool :: Nil,
     noSimplification,
     partialTyper {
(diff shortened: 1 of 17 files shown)


core/src/main/scala/slamdata/engine/std/set.scala
```diff
@@ -77,17 +77,58 @@ trait SetLib extends Library {
       case t           => success(Type.Set(t) :: Type.Bool :: Nil)
     })
 
-  val Cross = Transformation(
-    "CROSS",
-    "Computes the Cartesian product of two sets",
-    Type.Top :: Type.Top :: Nil,
+  val InnerJoin = Transformation(
+    "INNER JOIN",
+    "Joins two sets …",
```
mossprescott: How about an actual description? Surely it can be cribbed from somebody's SQL doc.

core/src/main/scala/slamdata/engine/std/set.scala
```diff
@@ -77,17 +77,58 @@ trait SetLib extends Library {
...
+    "LEFT OUTER JOIN",
+    "Joins two sets …",
+    Type.Top :: Type.Top :: Type.Bool :: Nil,
+    noSimplification,
+    partialTyper {
+      case List(s1, _, Type.Const(Data.Bool(false))) =>
+        Type.Obj(Map("left" -> s1, "right" -> Type.Null), None)
+      case List(Type.Const(Data.Set(Nil)), _, _) => Type.Const(Data.Set(Nil))
+      case List(s1, s2, _) =>
+        Type.Obj(Map("left" -> s1, "right" -> (s2 | Type.Null)), None)
```
mossprescott: Slick, this is almost all the description I need. Users might not feel the same way, at the point that we use the descriptions to generate docs.

mossprescott:
That's about all I have to say about this, which seems straightforward. 

I *would* like to come up with some way to not get dinged on the coverage as we adopt the new style, but I suppose that's not really part of this. It's a shame to be put in the position of scrounging for dead code just to get the build to succeed.

PR #861, opened 2015-08-11
Fixes and improvements arising from arrayConcat cases

### LINKED ISSUE #676: mongodb: duplicated $match ops in query with array splicing and filter
After #634, this query now works, but similar $match ops appear before and after the splice:

```sql
select loc || [ pop ] from zips where city = 'BOULDER'
```

See the pending planner test marked with this issue number.

mossprescott:
Handle several previously broken query patterns involving arrayConcat.

Some of these patterns led to working but inefficient plans, due to falling back on merging. They now are explicitly handled so as to generate the expected plan (fixes #676).

Pulling this thread led to a whole series of loosely related issues:

- add (simple) arrayConcat expression to query generator
- additional cases in objectConcat and arrayConcat
- don't inline $unwinds that are preceded by pipeline ops
- `normalize` WorkflowBuilders at each step, to eliminate unnecessary nesting
- handle more shapes when re-ordering $match ops around $simpleMap
- actually useful RenderTree[JsCore/JsFun]
- `ExprOp.toJs` is now implemented as a fold with `toJsƒ` implementing the simple, reusable part.
- implement `paraM`
- implement `Selector.mapUpFieldsM`


core/src/main/scala/slamdata/engine/analysis/fixplate.scala
```diff
@@ -161,6 +161,9 @@ sealed trait term {
     def para[A](f: F[(Term[F], A)] => A)(implicit F: Functor[F]): A =
       f(unFix.map(t => t -> t.para(f)(F)))
 
+    def paraM[M[_]: Monad, A](f: F[(Term[F], A)] => M[A])(implicit F: Traverse[F]): M[A] =
+      unFix.map(v => v.paraM(f).map(v -> _)).sequence.flatMap(f)
+
```
sellout: 👍🏾, but wondering if we should start defining the non `M` cases in terms of `Id`.

core/src/main/scala/slamdata/engine/debug.scala
```diff
@@ -166,6 +166,9 @@ object RenderTree {
 
   def show[A](a: A)(implicit RA: RenderTree[A]): Cord = Show[RenderedTree].show(RA.render(a))
 
+  /** For use with `<|`, mostly. */
+  def print[A: RenderTree](label: String, a: A): Unit = println(label + ":\n" + show(a))
+
```
sellout: 👍🏾, but this is going to conflict with my new PR 😭

core/src/main/scala/slamdata/engine/physical/mongodb/findquery.scala
```diff
@@ -68,15 +69,16 @@ sealed trait Selector {
...
-      case Doc(pairs)          => Doc(pairs.map { case (field, expr) => f(field) -> expr })
-      case And(left, right)    => And(left.mapUp0(f), right.mapUp0(f))
-      case Or(left, right)     => Or(left.mapUp0(f), right.mapUp0(f))
-      case Nor(left, right)    => Nor(left.mapUp0(f), right.mapUp0(f))
-      case Where(_)            => this // FIXME: need to rename fields referenced in the JS (#383)
+      case Doc(pairs)       => pairs.toList.map { case (field, expr) => f(field).map(_ -> expr) }.sequenceU.map(ps => Doc(ps.toListMap))
+      case And(left, right) => (left.mapUpFieldsM(f) |@| right.mapUpFieldsM(f))(And(_, _))
+      case Or(left, right)  => (left.mapUpFieldsM(f) |@| right.mapUpFieldsM(f))(Or(_, _))
+      case Nor(left, right) => (left.mapUpFieldsM(f) |@| right.mapUpFieldsM(f))(Nor(_, _))
+      case Where(_)         => this.point[M] // FIXME: need to rename fields referenced in the JS (#383)
```
sellout: BSON, you _will_ become a functor.

core/src/main/scala/slamdata/engine/analysis/fixplate.scala
mossprescott: That would reduce duplication and possibly improve coverage. On the other hand, it may be worth keeping around the simpler, more efficient, non-monadic forms, especially if any of these folds are used a lot.

Aside: this is something I've often wanted a language/platform to provide. Say I have a simple, correct implementation, and also a less simple but more efficient implementation. I want to provide both, assert that their results are the same, and have that automatically checked.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -273,6 +273,65 @@ object WorkflowBuilder {
     case ArraySpliceBuilderF(src, _) => 1 + src
   }
 
+  /**
+    Simplify/coalesce certain shapes, eliminating extra layers that make it harder
+    to pattern match. Should be applied before `objectConcat`, `arrayConcat`, or `merge`.
+   */
+  def normalize(w: WorkflowBuilder): WorkflowBuilder = {
```
sellout: Good – we should be able to simplify some other WorkflowBuilder functions by moving more stuff to this.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -982,6 +1047,45 @@ object WorkflowBuilder {
...
+                Expr(-\/($var(rbase))))(List(_, _))),
+              inputs1, op1)
+          }
+        case (ShapePreservingBuilderF(_, in1, op1), ArrayBuilderF(Term(ShapePreservingBuilderF(_, in2, op2)), _)) => delegate
+
+        case (
+          ShapePreservingBuilderF(src1, inputs1, op1),
+          ArrayBuilderF(Term(
+            GroupBuilderF(Term(
+              ShapePreservingBuilderF(src2, inputs2, op2)),
```
sellout: We need fixpoint wrappers for these Builders.

core/src/main/scala/slamdata/engine/physical/mongodb/expression/package.scala
```diff
@@ -276,4 +242,42 @@ package object expression {
...
+      // same as above, but for arrays
+      case $andF(
+        ($lte($literal(Bson.Arr(List())), f1), _),
+        ($lt(f2, $literal(b1)), _))
+          if f1 == f2 && b1 == Bson.Binary(scala.Array[Byte]())=>
+        toJs(f1).map(f =>
+          JsFn(JsFn.base,
+            JsCore.BinOp(JsCore.Instance, f(JsFn.base.fix), JsCore.Ident("Array").fix).fix))
+
+      case op => toJsƒ(op.map(_._2))
```
sellout: This fold isn’t too bad, but I’m inclined to at least expose the “real” fold (`paraM`) as `toJsƒ`, and maybe hide the cataM in that function.

Also, using `histoM` instead of `paraM` would eliminate the recursive calls, but it may complicate other things, so I’m not sure it’s a win.

sellout:
Looks pretty good – only two things are revisiting the `toJs` fold and perhaps making fixpoint wrappers for WorkflowBuilder.

core/src/main/scala/slamdata/engine/physical/mongodb/expression/package.scala
mossprescott: I need to expose the simpler fold so it can be used in WorkflowBuilder.normalize, but now saying that think I might need something better there.

Also, I don't really like the way the these cases match on the tuples but use only the first element, so I think I'll try breaking that out as a partial function and then just compose the two in `toJs`.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
mossprescott: I guess I'd prefer to do that in a separate PR. It's going to touch a lot of lines of code in a trivial way.

sellout: Yeah, probably good to do along with the repackaging.

mossprescott:
Updated.

PR #862, opened 2015-08-11
Various FP improvements.

sellout:
* Use Simulacrum for all type classes,
* add new `FunctorT` type class for operations over `Term` and `Cofree`, and
* rename `Js.render` to `.pprint`,
* extract `fixplate`, `fp`, and `RenderTree` from `slamdata.engine`, and
* standardize imports (see the project Wiki).


wemrysi:
Would `FoldableT` be a better name than `FunctorT` for the new typeclass, given the operations it defines?

sellout:
It was originally called `FoldableF`, but I think the fact that it only has `Foldable` operations currently is just a matter of it being incomplete. I don’t know if `FunctorT` is the right name either, but “functor” is just shorter than saying “unary type operator”.

core/src/main/scala/slamdata/fp/FunctorT.scala
```diff
@@ -0,0 +1,22 @@
+package slamdata.fp
+
+import slamdata.Predef._
+
+import scalaz._; import Scalaz._
+import simulacrum.typeclass
+
+@typeclass trait FunctorT[T[_[_]]] {
```
mossprescott: I think a typeclass is worthy of a comment explaining its intended use, especially in this case since earlier discussion indicates the current methods don't entirely reflect it.

Do realistically intend to implement this for types other that `Term`? If not, why not just define these methods there?

core/src/main/scala/slamdata/fixplate.scala
```diff
@@ -533,4 +526,4 @@ trait binding extends attr {
   }
 }
 
-object fixplate extends binding
+package object fixplate extends binding
```
mossprescott: :+1: I could *never* remember what package fixplate was hiding in!

mossprescott: Having said that, when we pull this out to a separate repo as `scala-fixplate`, is `slamdata.fixplate` still the right package?

core/src/main/scala/slamdata/fp/FunctorT.scala
mossprescott: Oh, Cofree. Nevermind, I'd still like to see an attempt to articulate what this is for in a comment, though.

sellout: Yep – already defined for Cofree. That’s why I pulled them out of Term.

core/src/main/scala/slamdata/fixplate.scala
sellout: I think when we pull things out of this repo, we should publish them as standalone packages (EG, just `fixplate`). Although, I do think this also needs to be renamed, because it doesn’t really follow the Fixplate model anymore.

core/src/main/scala/slamdata/fp/FunctorT.scala
wemrysi: :+1: for an attempt at articulation.

core/src/main/scala/slamdata/fp/FunctorT.scala
```diff
@@ -0,0 +1,22 @@
...
+  def all[F[_]: Foldable](t: T[F])(p: T[F] ⇒ Boolean): Boolean
+  def any[F[_]: Foldable](t: T[F])(p: T[F] ⇒ Boolean): Boolean
+
+  def contains[F[_]: EqualF: Foldable](t: T[F], c: T[F])(implicit T: Equal[T[F]]): Boolean =
+    any(t)(_ ≟ c)
+
+  def foldMap[F[_]: Foldable, Z: Monoid](t: T[F])(f: T[F] => Z): Z =
+    foldMapM[F, Free.Trampoline, Z](t)(f(_).pure[Free.Trampoline]).run
+
+  def foldMapM[F[_]: Foldable, M[_]: Monad, Z: Monoid](t: T[F])(f: T[F] => M[Z]): M[Z]
```
wemrysi: Of the operations defined, `foldMapM` appears to be the only necessary one for a complete definition as `all` and `any` can have default implementations defined in terms of `foldMap`.

wemrysi:
Regarding naming, I'd still vote for `Foldable{F,T}` as any non-foldable operations would likely warrant a separate abstraction.

core/src/main/scala/slamdata/fp/FunctorT.scala
sellout: Yeah, I was just avoiding the line noise caused by scalaz’s type tag approach to conflicting implicits (personally, I prefer just putting multiple implicits in scope, then explicitly passing one). But probably better to not require people to implement them.

sellout:
I don’t think they do warrant a separate abstraction – the only thing that separates them (I think) is the constraint on `F`. I don’t think there’s a case where something that provides `FunctorT` wouldn’t also provide `FoldableT` and `TraverseT`.

sellout:
Ok, wavering on this … `FoldableT` might be the right name for now. If we find use for operations for other type classes, we can generalize at that point.

core/src/main/scala/slamdata/fp/FunctorT.scala
sellout: Ok, that definitely was not as bad as I feared. 👍🏾

wemrysi:
I think adding operations that place additional constraints on `F[_]` and only depend on `foldMapM` or derivatives is within the scope of this class. If we ran into a situation where we found an interesting operation that wasn't implementable in terms of `foldMapM` (or derivatives), requiring us to define another abstract function, then I think we'd want to step back and see if there is an additional abstraction lurking.

jdegoes:
> If we ran into a situation where we found an interesting operation that wasn't implementable in terms of foldMapM 

:+1:

sellout:
Updated.

sellout:
One other thing with this PR – using simulacrum means that all operations are “injected”. IE, any type with a `RenderTree`, instance has `.render`, and since there’s a RenderTree ⇒ Show implicit, it also has `.show`/`.shows`.

Among other things, this cleans up our RenderTree instances that had tons of implicits just to give short names to the various instances they depended on.

wemrysi:
:+1: looks good.

PR #867, opened 2015-08-12
Fix sequencing of effects when server internally restarts for config change

### LINKED ISSUE #865: Manual restart of slamengine required after adding or removing database
**NOTE**: This issue only occurs in the bundled version of SlamData that is downloaded from our website. I cannot reproduce with my own build and the latest `slamengine` jar.

When I mount a database, it does not appear in the list unless I manually quit `SlamData.app` (which I assume is what is running `slamengine`) and restart it. The same thing happens when removing a database.

mossprescott:
Fixes #865.

Also some refactoring to make some of this logic avaliable to the tests, and
add an integration test for server restart with new config.


mossprescott:
Testing locally with the installer bundle, replacing the jar with my local build, I'm now seeing the server restarting normally, and the requests look good, but I still don't see new mounts until I refresh the page. On the other hand, deleted mounts disappear right away. Maybe a client problem now?

jonsterling:
@mossprescott Sounds like the remaining issue is a client thing; I think even the Getting Started guide tells you to hit the folder icon in the top left (effectively refreshing) to see the new mount...

mossprescott:
Should be good to go after a small fix for Travis.

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -41,16 +41,21 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
+    }
+    val rez: slamdata.engine.Evaluator.EnvironmentError \/ Int = Server.run(port, config, "", 1.seconds, mounter, configWriter).run.run
 
     try {
       body
     }
     finally {
-      ignore(srv.shutdown.run)
+      // NB: in case the test provoked a deferred restart
+      java.lang.Thread.sleep(500)
```
jdegoes: This will lead to sporadic test failures.

PR #868, opened 2015-08-13
Generalize fixplate.

sellout:
* Rename to `recursionschemes`,
* move folds and unfolds into type classes,
* merge `FoldableT` into the `Recursive` type class, and
* rename `Term` to `Fix`.


sellout:
It may help to look at sellout/slamengine@e4e13cb0be0571529265ff990e3785803ff85257, because the other commit just renames `Term` and adds a ton of noise to the diff.

jonsterling:
Strong agreement on renaming `Term` to `Fix` :smiley:

mossprescott:
Mostly looks great, and if there's nothing to be done about the new type annotations, then I can live with them.

:+1:

core/src/main/scala/slamdata/recursionschemes/package.scala
```diff
@@ -0,0 +1,324 @@
...
+  }
+  implicit val FixCorecursive = new Corecursive[Fix] {
+    def embed[F[_]](t: F[Fix[F]]) = Fix(t)
+  }
+
+  implicit def CofreeRecursive[A] = new Recursive[Cofree[?[_], A]] {
+    def project[F[_]](t: Cofree[F, A]) = t.tail
+  }
+
+  // NB: Not currently possible because this requires `Functor[F]` and that
```
wemrysi: Should be possible once we're able to upgrade to scalaz-7.2 due to `Free` refactoring.

core/src/main/scala/slamdata/recursionschemes/package.scala
```diff
@@ -0,0 +1,324 @@
...
+      val t = RF.render(v.unFix)
+        NonTerminal(t.nodeType, t.label, v.children.map(render(_)))
+    }
+  }
+
+  implicit def FixEqual[F[_]](implicit F: Equal ~> λ[α => Equal[F[α]]]):
+    Equal[Fix[F]] =
+  Equal.equal { (a, b) => F(FixEqual[F]).equal(a.unFix, b.unFix) }
+
+  sealed trait Hole
```
wemrysi: What is the purpose of this type? It doesn't seem to be used anywhere.

sellout: I’m not sure. The functions below it come from http://hackage.haskell.org/package/fixplate-0.1.5/docs/Data-Generics-Fixplate-Open.html#g:2 but I don’t see any reference to anything like `Hole` in that code. @jdegoes – any recollection?

wemrysi: Looks like it is referenced in the `Shape` and the aforementioned link says it is useful for drawing trees as well.

wemrysi:
:+1: Nice work!

sellout:
@mossprescott – I think we can do something about the new type annotations eventually.

With something like Scalaz’s `…IdV` traits to inject the unfolds into `A`,
```scala
Corecursive[Term].futu(81)(extract2and3)
```
hopefully becomes
```scala
81.futu[Term](extract2and3)
```

And @tpolecat’s [partially-applied type constructors](http://tpolecat.github.io/2015/07/30/infer.html) can help reduce the not-new annotations that need to be provided on things like `cataM`.

Issue #822, opened 2015-07-03
API: clean up and improve error messages


mossprescott:
Currently all errors are returned in a JSON-formatted response of the form `{ "error": "..." }`, with a free-form message which looks different and contains different amounts of information for each type of error. In many cases these messages are just taken from the `message` of some exception or other.

We should adopt a common style for these messages, and capture useful information as separate fields in the response.
- [ ] Make the value under `error` an object with at least one field, `status`.
- [ ] `status` should contain a fixed text summarizing the problem. These messages should be short, grammatical, capitalized, and end with a period, e.g. `"'Destination' header missing."`
- [ ] The same status should be used in the HTTP status line, e.g. `400 'Destination' header missing.`
- [ ] Any additional detail which may help the user/client to understand the problem should be included as separate fields in the `error` object, e.g. the text of an unanticipated runtime error message.
- [ ] Where possible, units of data should be captured as *separate* fields, so that they can be easily captured and used without parsing strings, e.g. `"status": "Invalid path.", "detail": "The specified path is not contained within something or other.", "path": "../foo", "targetPath": "/"`

Needing clarity:
- What exactly constitutes "grammatical"?
- Should error responses ever include information that's visible in the request (e.g. `path` in the example above, probably). 
- Is there any need to document the fields, other than in the API tests?

To implement this cleanly, we probably need a simple, clean abstraction for representing errors to be rendered into this form, and then we need a way to map various kinds of errors that arise in the API services to that form. It will also mean capturing some errors in a way that preserves more detail than we currently have. Also see #774.

Split from #757. See discussion on PR #819.

jdegoes:
> What exactly constitutes "grammatical"?

Must contain (possibly implicit) subject, explicit verb, and explicit object.

> Should error responses ever include information that's visible in the request (e.g. path in the example above, probably).

This would be helpful. The client can reconstruct it, of course, but it may not keep 1-to-1 track of request to response. A lot of times these will end up in some log file so the context won't be known.

> Is there any need to document the fields, other than in the API tests?

No, let's just assume the client can, if necessary, either log them or dynamically pull them out into a way that a user can understand.

sellout:
I have a lot of work done on #774, so we should talk before any effort is put into representing errors.

jdegoes:
@sellout How much of this ticket is already done with your work on #774?

mossprescott:
[...]

PR #870, opened 2015-08-14
In the API, test connection to new mount before saving

### LINKED ISSUE #863: mongodb: test connection when adding/editing a mount
Currently, when a mount is added or edited through the API, we "validate" the configuration (which amounts to parsing the URI), but do not perform a connection test. Running `Backend.test` on the config would uncover problems with connectivity and authentication earlier.

Also, the connection test does not currently do any write operations. Adding that would make the test even more robust. See https://github.com/slamdata/slamengine/blob/6f275ac8f3c33bcfb9131535884818a5937b11cc/core/src/main/scala/slamdata/engine/backend.scala#L353-L358.

mossprescott:
Fixes #863.

Call `Backend.test` from the API before saving a new mount (previously it was
used only by the now defunct admin UI).

Make that test somewhat more robust by also writing a temp file so that it
fails in a read-only configuraiton.

Capture additional errors from MongoDB in our error types. Unfortunately, much
of the interesting information is available only in the message strings of the
MongoDB exceptions, so for now that's just captured and wrapped.

Also, went back to the simpler server setup for the bulk of the API tests. The
server restarting code path is now used only by the one test the depends on it
(and tests it), while the rest use the simpler, single-threaded setup.
Hopefully this will resolve the flakiness we've seen in Jenkins and elsewhere.

### DIFF
--- a/core/src/main/scala/slamdata/engine/evaluator.scala
+++ b/core/src/main/scala/slamdata/engine/evaluator.scala
@@ -72,6 +72,9 @@ object Evaluator {
       def message = "no database found"
     }
     final case class InvalidConfig(message: String) extends EnvironmentError
+    final case class ConnectionFailed(message: String) extends EnvironmentError
+    final case class InvalidCredentials(message: String) extends EnvironmentError
+    final case class InsufficientPermissions(message: String) extends EnvironmentError
     final case class EnvPathError(error: PathError) extends EnvironmentError {
       def message = error.message
     }
@@ -83,6 +86,16 @@ object Evaluator {
     }
     final case class UnsupportedVersion(backend: Evaluator[_], version: List[Int]) extends EnvironmentError {
       def message = "Unsupported " + backend + " version: " + version.mkString(".")
+    }
+
+    import argonaut._, Argonaut._
+    implicit val EnvironmentErrorEncodeJson = EncodeJson[EnvironmentError] {
+      case MissingDatabase              => Json("error" := "Authentication database not specified in connection URI.")
+      case ConnectionFailed(msg)        => Json("error" := "Invalid server and / or port specified.", "errorDetail" := msg)
+      case InvalidCredentials(msg)      => Json("error" := "Invalid username and/or password specified.", "errorDetail" := msg)
+      case InsufficientPermissions(msg) => Json("error" := "Database user does not have permissions on database.", "errorDetail" := msg)
+      case EnvWriteError(pe)            => Json("error" := "Database user does not have necessary write permissions.", "errorDetail" := pe.message)
+      case e                            => Json("error" := e.message)
     }
   }
 
@@ -114,6 +127,27 @@ object Evaluator {
     def apply(message: String): EnvironmentError = EnvironmentError.InvalidConfig(message)
     def unapply(obj: EnvironmentError): Option[String] = obj match {
       case EnvironmentError.InvalidConfig(message) => Some(message)
+      case _                       => None
+    }
+  }
+  object ConnectionFailed {
+    def apply(message: String): EnvironmentError = EnvironmentError.ConnectionFailed(message)
+    def unapply(obj: EnvironmentError): Option[String] = obj match {
+      case EnvironmentError.ConnectionFailed(message) => Some(message)
+      case _                       => None
+    }
+  }
+  object InvalidCredentials {
+    def apply(message: String): EnvironmentError = EnvironmentError.InvalidCredentials(message)
+    def unapply(obj: EnvironmentError): Option[String] = obj match {
+      case EnvironmentError.InvalidCredentials(message) => Some(message)
+      case _                       => None
+    }
+  }
+  object InsufficientPermissions {
+    def apply(message: String): EnvironmentError = EnvironmentError.InsufficientPermissions(message)
+    def unapply(obj: EnvironmentError): Option[String] = obj match {
+      case EnvironmentError.InsufficientPermissions(message) => Some(message)
       case _                       => None
     }
   }
--- a/core/src/main/scala/slamdata/engine/evaluator.scala
+++ b/core/src/main/scala/slamdata/engine/evaluator.scala
@@ -89,13 +89,18 @@ object Evaluator {
     }
 
     import argonaut._, Argonaut._
-    implicit val EnvironmentErrorEncodeJson = EncodeJson[EnvironmentError] {
-      case MissingDatabase              => Json("error" := "Authentication database not specified in connection URI.")
-      case ConnectionFailed(msg)        => Json("error" := "Invalid server and / or port specified.", "errorDetail" := msg)
-      case InvalidCredentials(msg)      => Json("error" := "Invalid username and/or password specified.", "errorDetail" := msg)
-      case InsufficientPermissions(msg) => Json("error" := "Database user does not have permissions on database.", "errorDetail" := msg)
-      case EnvWriteError(pe)            => Json("error" := "Database user does not have necessary write permissions.", "errorDetail" := pe.message)
-      case e                            => Json("error" := e.message)
+    implicit val EnvironmentErrorEncodeJson = {
+      def format(message: String, detail: Option[String]) =
+        Json(("error" := message) :: detail.toList.map("errorDetail" := _): _*)
+
+      EncodeJson[EnvironmentError] {
+        case MissingDatabase              => format("Authentication database not specified in connection URI.", None)
+        case ConnectionFailed(msg)        => format("Invalid server and / or port specified.", Some(msg))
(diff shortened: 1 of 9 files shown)


mossprescott:
This PR captures all the errors requested in #863, with the actual error text from MongoDB. Unfortunately, the relevant detail seems to be available *only* in the message string (see https://github.com/mongodb/mongo-java-driver/blob/8866de115d45ea73a39da9360159d3fbb6390dd9/driver-core/src/main/com/mongodb/connection/BaseCluster.java#L160-L163), so it will take some doing to extract it in a more readable form. We also have some work to do on our side to capture more fine-grained info in the errors and render it more nicely (see #822).

Currently the errors appear as follows:

Invalid connection URI:
```javascript
{"error":"invalid connection URI: mongodb:/admin:admin@localhost/"}
```

No database in connection URI:
```javascript
{"error":"no database found"}
```

Bad host/port:
```javascript
{"error":"connection timeout: Timed out after 30000 ms while waiting for a server that matches PrimaryServerSelector. Client view of cluster state is {type=UNKNOWN, servers=[{address=localhost1:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketException: localhost1}, caused by {java.net.UnknownHostException: localhost1}}]"}
```

No credentials:
```javascript
{"error":"listing files: Query failed with error code 13 and error message 'not authorized for query on admin.system.namespaces' on server localhost:27017"}
```

Bad user/password:
```javascript
{"error":"connection timeout: Timed out after 30000 ms while waiting for a server that matches PrimaryServerSelector. Client view of cluster state is {type=UNKNOWN, servers=[{address=localhost:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSecurityException: Exception authenticating}, caused by {com.mongodb.MongoCommandException: Command failed with error 18: 'auth failed' on server localhost:27017. The full response is { \"ok\" : 0.0, \"errmsg\" : \"auth failed\", \"code\" : 18 }}}]"}
```

No rights on db:
```javascript
{"error":"connection timeout: Timed out after 30000 ms while waiting for a server that matches PrimaryServerSelector. Client view of cluster state is {type=UNKNOWN, servers=[{address=localhost:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSecurityException: Exception authenticating}, caused by {com.mongodb.MongoCommandException: Command failed with error 18: 'auth failed' on server localhost:27017. The full response is { \"ok\" : 0.0, \"errmsg\" : \"auth failed\", \"code\" : 18 }}}]"}
```

No write permission:
```javascript
{"error":"write failed: Command failed with error 13: 'not authorized on test to execute command { insert: \"tmp.gen_8977b06_0\", ordered: true, documents: [ { _id: ObjectId('55ce2ec22473282031bd9805'), a: 1 } ] }' on server localhost:27017. The full response is { \"ok\" : 0.0, \"errmsg\" : \"not authorized on test to execute command { insert: \\\"tmp.gen_8977b06_0\\\", ordered: true, documents: [ { _id: ObjectId('55ce2ec22473282031bd9805'), a: 1 } ] }\", \"code\" : 13 }; value: Obj(Map(a -> Int(1)))"}
```

jdegoes:
These error messages need to be improved; they don't provide enough information for an end-user. Since the error message is only available via String, this implies you're going to have to match against these strings to detect which of the cases it is. The raw message from MongoDB can go into the details field, and can go into the primary field if no match can be found. This is important enough we should test for all these scenarios and ensure we identify them with the MongoDB client library we are using (although that can be a separate PR if you can't get it in today).

Here are better error messages for these cases:

> No database in connection URI:

"Authentication database not specified in connection URI. 

> Bad host/port:

"Invalid server and / or port specified."

> No credentials:

"No username and / or password specified."

> Bad user/password:

"Invalid username and / or password specified."

> No rights on db:

"Database user does not have permissions on database."

> No write permission:

"Database user does not have necessary write permissions."

jdegoes:
Btw, I'd match against their "poor man's error type", which appears to be the part of the message preceding the first `:`, e.g. 'connection timeout', etc. Also from your examples, bad username / password cannot be distinguished from no rights, so I'd just use the bad username / password message on that one. Arguably, no rights is a bad username / password, anyway.

mossprescott:
That prefix is coming from me, actually, based on the exception type, which carries a little information, but as you said, it's a timeout for many of these cases.

mossprescott:
Updated with error messages as requested above. 

Note: this PR does not include any automated tests, because that will require some additional setup for each test environment. But this is certainly going to be prone to breakage in future MongoDB versions and won't get regularly exercised except inadvertently, so it's probably worth investing the time. I'll write up a ticket.

core/src/main/scala/slamdata/engine/evaluator.scala
```diff
@@ -72,15 +72,31 @@ object Evaluator {
...
+    final case class EnvWriteError(error: Backend.ProcessingError) extends EnvironmentError {
+      def message = "write failed: " + error.message
+    }
     final case class UnsupportedVersion(backend: Evaluator[_], version: List[Int]) extends EnvironmentError {
       def message = "Unsupported " + backend + " version: " + version.mkString(".")
     }
+
+    import argonaut._, Argonaut._
+    implicit val EnvironmentErrorEncodeJson = EncodeJson[EnvironmentError] {
+      case MissingDatabase              => Json("error" := "Authentication database not specified in connection URI.")
```
jdegoes: A local helper function could generate this Json structure given a string, and it would eliminate the duplication in all these cases.

wemrysi: Should we include an "error code" or other form of identifier to make it easier identify a particular type of error?

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1618,8 +1618,29 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
+      // Exercises the server restarting code path in server.scala.
+       def withRestartableServer[A](backend: Backend, config: Config)(body: => A): A = {
+         val rez: slamdata.engine.Evaluator.EnvironmentError \/ Int = Server.run(port, config, "", 1.seconds, tester, mounter(backend), configWriter).run.run
+
+         try {
+           body
+         }
+         finally {
+           // NB: the test will have provoked a deferred restart
+           java.lang.Thread.sleep(500)
```
jdegoes: If the test passing is dependent on these (and I assume it is or they wouldn't be here), we can count on it to fail some percentage of the time. Of course, I realize one sleep was in there before, but now we have two, and a bad situation is growing worse. 

What condition are you guarding against? Making sure the server has restarted?

core/src/main/scala/slamdata/engine/evaluator.scala
jdegoes: That makes sense if the front-end can use that information in some way, and if the error code cannot be communicated via HTTP Status Code (which is preferred for a REST API). Currently, all these are user-land errors, so the front-end can't do anything than print a message, so I think it's OK to skip for now.

web/src/test/scala/slamdata/engine/api/fs.scala
mossprescott: First, this is better than what was here before in that only this one test now has to deal with this craziness, whereas before all the rest of the tests were subjected to it, even though they didn't need to be.

However, when we do need to this, it's worse than you think.

When one of these config-modifying requests come in, we need to do three things in sequence:

1. validate and save the new configuration
1. respond to the request
1. shutdown the server and start a new one with the new configuration

The problem is that we can't do (3) until after (2) has completed, but (2) happens somewhat outside of our control, because it involves http4s running the `Task` we provide it with . So the way that works is when http4s runs our Task, we do (1), spawn a Java thread to run the restart Task after a delay, and then return the value which http4s will turn into a response. By the time our thread gets around to restarting the server, the response will have gone out (we hope).

The condition we'd like to observe is that the response has actually been delivered, but there is no way to do that that we could figure out so we came up with this hack. This has been in place for months now; we're just talking about it because I wrote a test to try to prove that it actually works.

Given that, the test needs to be sure that the restart sequence has happened, and then also wants to make sure that the second server instance has shut down before the next test is allowed to run, hence the multiple silly delays.

jdegoes:
> When one of these config-modifying requests come in, we need to do three things in sequence:

Indeed, this is the source of the trouble. It's not a good architecture. The only reason the server *needs* restarting is because of a port change. Adding, changing, or deleting mounts is no cause of restart. That, at least, would eliminate most of the current trouble.

Actually, it would eliminate all of them (I believe), since on a port change, you don't care how long it takes for the old server to actually wind down.

jdegoes:
Please create a ticket for cleaning up restarts, and I'll merge this in.

mossprescott:
See #874.

PR #872, opened 2015-08-14
Refactor server config to use typesafe paths

### LINKED ISSUE #858: incorrect formation of paths on windows
```scala
  def defaultPath: Task[String] = Task.delay {
    import scala.util.Properties._

    val commonPath = "SlamData/slamengine-config.json"

    if (isWin)
      envOrElse("LOCALAPPDATA", propOrElse("user.home", ".")) + commonPath
    else
      propOrElse("user.home", ".") +
        (if (isMac) "/Library/Application Support/" else "/.config/") +
        commonPath
  }
```

On Windows, there is no path separator between the `commonPath` and whatever crap comes out of the environment variable.

Note that such bugs cannot happen with a library abstraction like [purescript-pathy](http://github.com/slamdata/purescript-pathy). :wink:

wemrysi:
Integrates `pathy.Path` into the server config, replacing `String` paths with typesafe versions.

Introduces `FsPath` to encapsulate paths that may be in a volume (e.g. `c:\foo\bar`, `\\c\foo\bar`). I'm not sure if this is something we'd want to extend pathy with or if it lies outside the scope of that library. The library is advertised and dealing with filesystem paths, so in that sense it might belong, though there are other kinds of paths, like URIs, which pathy could be used for and also have a scheme prefix, similar to windows volumes.

Also, the system currently makes the assumption that any paths provided via the `LOCALAPPDATA` environment variable or the `user.home` JVM property are absolute, let me know if this is a poor assumption.

Fixes #858.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,6 +61,7 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
     "io.argonaut"       %% "argonaut"                  % "6.1-M6"       % "compile, test",
     "org.jboss.aesh"    %  "aesh"                      % "0.55"         % "compile, test",
     "org.typelevel"     %% "shapeless-scalaz"          % slcVersion     % "compile, test",
+    "com.slamdata"      %% "pathy"                     % "0.0.1-SNAPSHOT" % "compile, test",
     "com.github.mpilquist" %% "simulacrum"             % "0.3.0"        % "compile, test",
     "org.scalaz"        %% "scalaz-scalacheck-binding" % scalazVersion  % "test",
     "org.specs2"        %% "specs2-core"               % "2.4"          % "test",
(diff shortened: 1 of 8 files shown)


jdegoes:
Travis failed.

wemrysi:
Yes, I realize I had build and tested locally against a newer version of pathy than what is on sonatype. I'm publishing a new snapshot.

wemrysi:
Merged in master and resolved conflicts. Also, forgot to mention here earlier, but updated pathy so building normally now.

wemrysi:
Hm, looks like the coverage alarm has been tripped with the result of the merge, I'll look and see what I can do to get us back in the green.

wemrysi:
I refactored a few things to separate the pure from impure, allowing for better testablility and improved coverage.

jdegoes:
:+1:

PR #875, opened 2015-08-16
Don’t throw away errors on server start-up.

sellout:


web/src/main/scala/slamdata/engine/api/server.scala
```diff
@@ -29,8 +29,7 @@ import scalaz._
 import scalaz.concurrent._
 
 object Server {
-  var serv: EnvironmentError \/ org.http4s.server.Server =
-    -\/(InvalidConfig("No server running."))
+  var serv: Option[org.http4s.server.Server] = None
```
jdegoes: Can we use `IORef` or something here? I wish there was a `TaskRef` but it looks like such a thing does not exist.

web/src/main/scala/slamdata/engine/api/server.scala
```diff
@@ -167,13 +166,15 @@ object Server {
         } yield ())
     }
 
-    ignore(start.run.run)
+    start.run.attemptRun.fold(
+      e => System.err.println(e.getMessage),
```
jdegoes: You should be able to make the whole thing pure except for the final `run`. This is currently less functional than it can be.

sellout: We tried and failed to do this a couple times, but today I finally figured out why it was so baffling – our constant function (as well as `Function.const`) aren’t lazy on their argument, so a value you’re expecting to have chained on a `Task` is evaluated eagerly.

So yeah, got this one 👍🏾

web/src/main/scala/slamdata/engine/api/server.scala
sellout: wedens pointed me at oncue/remotely, which is Apache licensed and has a `Task`-based `IORef`. I couldn’t figure out how to add it as a dependency, so I just copied their IORef.scala into our tree. It works locally, but I’m not sure if that’s the right way to “depend on” their code.

wemrysi: In this case it seems like copying the code is reasonable given the tradeoff of depending on remotely + dependencies for 50 LOC that likely isn't considered part of the remotely public API anyhow. We could submit a PR to scalaz and then switch over to it eventually.

web/src/main/scala/slamdata/engine/api/server.scala
jdegoes: I'm not sure I understand, can you explain the problem here? Impure code should always be wrapped in `Task.delay`, so that it becomes pure. Once you do that, you can reason about it like any other pure value.

sellout: Yeah, I guess it _is_ an issue with impure code, since it was the result of reading from a `var`, but of course, that doesn’t _look like_ impure code. So, switching to IORef fixes the issue as well – in that case, reading from the mutable reference returns a `Task`, so we’re good. Obviating the lazy constant function.

In fact, in the interim, I had fixed it and it looked something like `someTask.flatMap(κ(Task.delay(serv.something)))`, but I thought “that delay is unnecessary, because that function shouldn’t be evaluated until after `serv` is set properly by `someTask`.” So I went back to the broken `someTask.map(κ(serv.something))` and tried to figure out why that wasn’t behaving how I expected.

jdegoes: > Yeah, I guess it is an issue with impure code, since it was the result of reading from a var, but of course, that doesn’t look like impure code. So, switching to IORef fixes the issue as well

Yes, it's absolutely important we "not lie" with the types, or try to mask problems caused by lying. Moving away from vars and to `IORef` should definitely help here. Another thing we could to is try to create `IO` / `Task` wrappers around APIs, so we're not having to constantly `Task.delay` things. Easier to stay in purely functional land (we wrap stuff like `System.out.println` enough it might be worth it at some point).

sellout: Yeah – I have a note to look at aloiscochard/matterhorn for giving us `IO`-based wrappers for printing, etc.

web/src/main/scala/slamdata/engine/api/server.scala
sellout: What I’ve done is copy the entire file (including Verizon copyright notice, only modified to deal with `-Yno-imports`) into `web/src/main/scala/remotely/` so that the original package is maintained.

Should I add anything else, say to link to the GitHub repo or anything?

jdegoes: That's probably fine. I'm surprised there doesn't exist `IO ~> Task` already in Scalaz, which would just let us use / lift `IORef`.

wemrysi: Good point, I guess we could just define something like
```scala
val fromIO: IO ~> Task = new (IO ~> Task) {
  def apply[A](io: IO[A]): Task[A] = Task.delay(io.unsafePerformIO())
}
```
ourselves and then reuse the existing `IO` machinery.

jdegoes: :+1: This is preferable to adding another copyright notice we have to incorporate into the app.

web/src/main/scala/slamdata/engine/api/server.scala
```diff
@@ -174,13 +178,11 @@ object Server {
...
+    start.map(
+      κ(fromIO(serv.read).flatMap(_.fold(
+        Task.delay(System.err.println("Server failed to start. Exiting.")))(
+        κ(waitForInput.flatMap(κ(destroyServer))))))).fold(
       e => Task.delay(System.err.println(e.message)),
-      κ(for {
-        _ <- waitForInput
-        _ <- Task.delay { serv.fold(κ(()), x => ignore(x.shutdownNow)) }
-      } yield ())).run
+      ɩ).join.attemptRun.fold(e => System.err.println(e.getMessage), ɩ)
```
mossprescott: How about rewriting this as `val exec = start.map(...).join.attempt.map(_.fold(...))` followed by `exec.run`? That would make it clearer when `run` is happening, and that it really is at the end of the proverbial universe. At the moment it's a bit lost for me in the middle of join and fold.

mossprescott:
I don't see any problems with this, although it's still a bit hard for me to decipher so I'm not sure I would spot them if they were there.

My only concern is the introduction of `IO`, when we've only used `Task` for similar purposes so far. I guess I'd rather stick to one way of modeling effects unless there's some good reason to bring in a new thing. Wouldn't some version of `TaskRef` server equally well?

web/src/main/scala/slamdata/engine/api/server.scala
jdegoes: Indeed, and technically the `System.err.println` is not pure. Obviously, you like your folds :laughing: , but this whole thing would be clearer in a `for` comprehension, IMO.

web/src/main/scala/slamdata/engine/api/server.scala
```diff
@@ -25,13 +25,12 @@ import java.io.File
...
-import scalaz._
-import scalaz.syntax.std.option._
+import scalaz._, Scalaz._
 import scalaz.concurrent._
+import scalaz.effect._
 
 object Server {
-  var serv: EnvironmentError \/ org.http4s.server.Server =
-    -\/(InvalidConfig("No server running."))
+  val serv = IO.newIORef(None: Option[org.http4s.server.Server]).unsafePerformIO
```
jdegoes: This should be made private.

sellout:
Remotely’s `IORef` was a TaskRef, but @jdegoes and @wemrysi voted for IO over that. I’m fine with either.

web/src/main/scala/slamdata/engine/api/server.scala
jdegoes: > Wouldn't some version of TaskRef server equally well?

Such a thing needs to exist, but probably in Scalaz. I was not too excited about copy / pasting Remotely code into the project. Also Remotely's Task reference is impure (albeit, so is the way we're using `IORef`).

web/src/main/scala/slamdata/engine/api/server.scala
sellout: I _don’t_ like `fold`! But I like `for` even less 😉

jdegoes:
:+1:

PR #876, opened 2015-08-17
Convert sql.Expr to fixpoint style.

### LINKED ISSUE #3: parser: convert SQL AST and semantic analysis to fixplate
Currently the SQL AST uses a monomorphic representation, which necessitates a lot of machinery to annotate the tree -- brittle machinery, at that.

Now that the fixplate port is sufficiently mature, we need to swap over from the existing AST to a fixed point functor representation similar to LogicalPlan.

This is mostly a rote process, the trickiest part will be deciding which parts of fixplate ot use for each semantic phase.

sellout:
Fixes #3.

We don’t yet take much advantage of this, however:

* Annotations now only exist on `Expr`s, and there is no longer a `Node` type;
* the `CrossRelation` data type is removed as it’s just a restricted `JoinRelation`; and
* `Expr.mapUpM` is gone in favor of `cataM`.

This is further prep for annotations becoming (un)folds.


jdegoes:
Let's say this fixes #3. At least as good as we need until we get the new expression-oriented SQL parser + ADT.

core/src/main/scala/slamdata/engine/compiler.scala
```diff
@@ -285,7 +278,7 @@ trait Compiler[F[_]] {
         relations =
           if (namedRel.size <= 1) namedRel
           else {
-            val filtered = namedRel.filter(x => Path(x._1).filename == node.sql)
+            val filtered = namedRel.filter(x => Path(x._1).filename == node.para(sqlƒ))
```
mossprescott: I count 12 occurrences of this. Maybe worth a `sql` op?

mossprescott:
I don't think we have any tests of the RenderTree instance for Node/Expr. Did you at least eyeball the output to make sure it's sensible?

No other comments. Looks good.

core/src/main/scala/slamdata/engine/compiler.scala
sellout: Whoops – meant to do that. But what do you think of calling it `pprint` (like we do for `Js`)? Since 1. I think we should rename `Expr` to `Sql` and 2. `sql` causes confusion since the package name is also `sql`.

sellout:
The `RenderTree` instance hasn’t changed, although we should do that eventually, so it operates on `ExprF[_]` rather than `Expr`. Also, yeah, we should have tests 😄

sellout:
@jdegoes – Yeah, I think that’s reasonable.

core/src/main/scala/slamdata/engine/compiler.scala
mossprescott: Sure, assuming we're not replacing it with a partial isomorphism :)

I would prefer Sql to Expr also, just because we have lots of things called Expr.

sellout: Hopefully it _will_ be a partial isomorphism. But that’s up to @jedesah 😉

jdegoes:
@mossprescott You good with this?

mossprescott:
:+1:

PR #878, opened 2015-08-17
reduce the MongoDB driver's serverSelectionTimeout to 5 seconds

### LINKED ISSUE #877: It takes 30s for a bad mount configuration to be rejected
As of #863 / #870, the backend now verifies that it can connect to a mount before saving it; this is useful to the frontend, since it can now present feedback to the user and avoid entering an inconsistent state.

However, if an incorrect mount is provided (e.g. with the wrong port), it takes 30 seconds for it to be rejected. On the frontend, this translates to the user waiting for 30 seconds before they receive the following feedback: 

```
{"error":"Invalid server and / or port specified.","errorDetail":"Timed out after 30000 ms while waiting for a server that matches PrimaryServerSelector. Client view of cluster state is {type=UNKNOWN, servers=[{address=ds045089.mongolab.com:45087, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketOpenException: Exception opening socket}, caused by {java.net.ConnectException: Connection refused}}]"}
```

It this something that can be improved? Ideally it would fail a lot sooner, but if that's not possible, we should probably open a ticket on the frontend to show some indication to the user that we are validating the mount (since it just looks like the app got wedged).

mossprescott:
Fixes #877.

### DIFF
--- a/core/src/main/scala/slamdata/engine/physical/mongodb/util.scala
+++ b/core/src/main/scala/slamdata/engine/physical/mongodb/util.scala
@@ -26,9 +26,13 @@ import scalaz.Memo
 import com.mongodb._
 
 object util {
+  private val DefaultOptions =
+    (new com.mongodb.MongoClientOptions.Builder)
+      .serverSelectionTimeout(5000)
+
   private val mongoClient: String => Task[MongoClient] = {
     val memo = Memo.mutableHashMapMemo[String, MongoClient] { (connectionUri: String) =>
-      new MongoClient(new MongoClientURI(connectionUri))
+      new MongoClient(new MongoClientURI(connectionUri, DefaultOptions))
     }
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/slamdata/engine/physical/mongodb/util.scala
```diff
@@ -26,9 +26,13 @@ import scalaz.Memo
 import com.mongodb._
 
 object util {
+  private val DefaultOptions =
```
jdegoes: This should be inlined (placed into the code block) because it may be mutated by MongoDB.

mossprescott: Ah, the _builder_ is mutable, yes. OK will do.

PR #880, opened 2015-08-18
handle home directory without trailing slash on posix/mac

### LINKED ISSUE #879: default config location is incorrect on Mac
Since adopting scala-pathy, we are trying to load the configuration from `currentDir` (e.g., `./Library/Application Support/SlamData/slamengine-config.json`), when the path should be relative to the user's home directory (Java's `user.dir`). pathy doesn't have a constructor for that, but perhaps it should.

mossprescott:
Fixes #879.

Requires updated pathy.


jdegoes:
Pathy changes have been merged.

mossprescott:
After doing a release on pathy, should be good to go.

Issue #780, opened 2015-06-16
match not being pulled to before project resulting in terrible performance


jdegoes:
```sql
select distinct demographic.gender as gender,  
  sum("twitter.retweets.count") as sum
  from "/csesocial/Sprite_Simple"  
  where demographic.gender is not NULL  
  group by demographic.gender
```

jdegoes:
This may be the same as or different from slamdata/slamengine#418.

mossprescott:
This seems to be another case, as in #418, where the filter doesn't get compiled to a selector, but could.

```
    {
      "$project": {
        "__tmp2": { "$not": { "$eq": ["$demographic.gender", { "$literal": null }] } },
        "__tmp3": "$$ROOT"
      }
    },
    { "$match": { "__tmp2": true } },
```

Definitely in the low-hanging fruit category.

PR #883, opened 2015-08-19
implement Not in the selector phase

### LINKED ISSUE #418: mongodb: inefficient plan when filtering expression includes `not`
Whenever `not` appears in filtering, we currently generate a JS selector. In many cases, we could fold the negation into a non-JS selector, and stay in the aggregation pipeline.

For example: `where city not like 'BOULDER%'`, `where not (pop > 1000)`.

mossprescott:
Fixes #418.

Add `negate` method on Selector with unit tests.

Also rewrite `not(eq(...))` as `$ne`, because `$eq` isn't a thing and `$not`
requires an operator, and switch to the 3.0 driver's new BsonRegularExpression
wrapper, since `$not: { $regex: "..."}` isn't allowed.

### DIFF
--- a/core/src/main/scala/slamdata/engine/physical/mongodb/findquery.scala
+++ b/core/src/main/scala/slamdata/engine/physical/mongodb/findquery.scala
@@ -78,6 +78,27 @@ sealed trait Selector {
       case Nor(left, right) => (left.mapUpFieldsM(f) |@| right.mapUpFieldsM(f))(Nor(_, _))
       case Where(_)         => this.point[M] // FIXME: need to rename fields referenced in the JS (#383)
     }
+
+  def negate: Selector = {
+    def expr(x: Selector.SelectorExpr): Selector.SelectorExpr = x match {
+      case Selector.Expr(cond) => Selector.NotExpr(cond)
+      case Selector.NotExpr(cond) => Selector.Expr(cond)
+    }
+
+    def loop(s: Selector): Selector = s match {
+      case Selector.Doc(pairs) =>
+        pairs.toList
+          .map { case (f, x) => Selector.Doc(ListMap(f -> expr(x))) }
+          .reduceOption[Selector](Selector.Or(_, _))
+          .getOrElse(Selector.Where(Js.Bool(false)))
+      case Selector.And(l, r) => Selector.Or(loop(l), loop(r))
+      case Selector.Or(l, r)  => Selector.Nor(l, r)
+      case Selector.Nor(l, r) => Selector.Or(l, r)
+      case Selector.Where(x)  => Selector.Where(Js.UnOp("!", x))
+    }
+
+    loop(this)
+  }
 }
 
 object Selector {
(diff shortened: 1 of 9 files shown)


core/src/main/scala/slamdata/engine/physical/mongodb/findquery.scala
```diff
@@ -78,6 +78,27 @@ sealed trait Selector {
...
+      case Selector.Expr(cond) => Selector.NotExpr(cond)
+      case Selector.NotExpr(cond) => Selector.Expr(cond)
+    }
+
+    def loop(s: Selector): Selector = s match {
+      case Selector.Doc(pairs) =>
+        pairs.toList
+          .map { case (f, x) => Selector.Doc(ListMap(f -> expr(x))) }
+          .reduceOption[Selector](Selector.Or(_, _))
+          .getOrElse(Selector.Where(Js.Bool(false)))
```
jdegoes: What's this `where false` about?

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -488,6 +488,27 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
...
+         $read(Collection("db", "zips")),
+         $match(
+           Selector.Or(
+             Selector.Doc(ListMap[BsonField, Selector.SelectorExpr](
+               BsonField.Name("pop") -> Selector.NotExpr(Selector.Gt(Bson.Int64(0))))),
+             Selector.Doc(ListMap[BsonField, Selector.SelectorExpr](
+               BsonField.Name("pop") -> Selector.NotExpr(Selector.Lt(Bson.Int64(1000)))))))))
+    }
+
+    "plan filter with not and equality" in {
```
jdegoes: It would be good to get a `not` planner test for reg exes.

it/src/test/resources/tests/filterWithNot.test
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "filter with negated regex selector",
+
+  "query": "select city, state from zips where city not like 'BOULD%' and pop = 18174",
```
jdegoes: This is good but you should include at least one more test which does a not on a complex composite, with `and`s, `or`s, and a `not` inside it.

core/src/main/scala/slamdata/engine/physical/mongodb/findquery.scala
mossprescott: An empty `Selector.Doc` matches anything, so it's negation needs to match nothing. I tried to think of a way to do that with a non-where selector but drew a blank, and since we don't actually ever generate empty Docs, we aren't negatively impacted by the Where (which has to be done in map-reduce).

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
mossprescott: There's nothing interesting about regexes when it comes to handling $not in the planner. I included an integration test because I had to change the way we represent regexes at the bson level.

it/src/test/resources/tests/filterWithNot.test
mossprescott: Sure. I'll add a test with a more complex condition. The only hard part is thinking of something that makes enough sense that I can verify the result.

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
jdegoes: K

mossprescott:
Let's not merge quite yet. I want to add one more test to explicitly close #780.

mossprescott:
Ready.

jdegoes:
:+1:

PR #891, opened 2015-08-24
Put Expression on the right of various disjunctions.

sellout:
Expression tends to be the “default” state, and so we’ve been doing
things like `.map(_.swap)` and `.fold(Some(_), None)`. This eliminates
that extra hoop to jump through.

The three types changed are `WorkflowBuilder.Expr`,
`WorkflowBuilder.GroupValue`, and `Reshape.Shape`.

Also, update Scala to 2.11.7 (some dependency was forcing it anyway) and
add a missing `SpliceBuilder` case to `WorkflowBuilder.objectConcat`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
   headers := Map(
     "scala" -> Apache2_0("2014 - 2015", "SlamData Inc."),
     "java"  -> Apache2_0("2014 - 2015", "SlamData Inc.")),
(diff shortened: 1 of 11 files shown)


mossprescott:
I only spotted one place where anything cleaned up, but I'm sure it's just because of the size of the diff. But, hey, if you're happy, I'm happy.

sellout:
Yeah, this is mostly because of changes in my typechecking branch, but it was fairly easily separable. There are four simplifications in this diff though 😉

PR #892, opened 2015-08-24
Project a single value for any expression appearing in the group keys

### LINKED ISSUE #459: compiler: GROUP BY should be compiled to something that distincts by something or other
This is a well-known issue, but I haven’t found a bug associated with it.

In a query like `select city from zips group by city` we do
```js
db.zips.aggregate([
    { "$group" : { "city" : { "$push" : "$city"} , "_id" : "$city"}},
    { "$unwind" : "$city"},
    { "$project" : { "city" : "$city" , "_id" : false}},
    { "$out" : "tmp.gen_0"}
  ],
  { allowDiskUse: true })
db.tmp.gen_0.find()
```
when really we want
```js
db.zips.aggregate([
    { "$group" : { "_id" : "$city"}},
    { "$project" : { "city" : "$_id" , "_id" : false}},
    { "$out" : "tmp.gen_0"}
  ],
  { allowDiskUse: true })
db.tmp.gen_0.find()
```
IE, we shouldn’t unwind unaggregated fields if those fields are part of the group key.

mossprescott:
Fixes #459.

New Func `Arbitrary`, to select a single value from a set. Translated to $first for MongoDB.

During compilation, keep track of the compiled form of each grouping key, and then when any of them occurs in a projection (even embedded in a larger expression), wrap with `Arbitrary(...)` to simulate SQL semantics.

Handle an additional case in objectConcat that arises from the new shape.

Update query generator to avoid generating anything that would (correctly) lead to a plan with two consecutive group ops, because that pattern is flagged as a common characteristic of busted plans.

Update integration test queries to remove DISTINCT where it's no longer needed.

### DIFF
--- a/it/src/test/resources/tests/groupByFlatten.test
+++ b/it/src/test/resources/tests/groupByFlatten.test
@@ -1,7 +1,7 @@
 {
     "name": "group by flattened field",
     "data": "slamengine_commits.data",
-    "query": "select distinct substring(parents[*].sha, 0, 1), count(*) from slamengine_commits group by substring(parents[*].sha, 0, 1)",
+    "query": "select substring(parents[*].sha, 0, 1), count(*) from slamengine_commits group by substring(parents[*].sha, 0, 1)",
     "predicate": "containsExactly",
     "expected": [{ "0": "0", "1": 1 },
                  { "0": "1", "1": 2 },
--- a/core/src/main/scala/slamdata/engine/compiler.scala
+++ b/core/src/main/scala/slamdata/engine/compiler.scala
@@ -63,11 +63,14 @@ trait Compiler[F[_]] {
         this.subtables ++ that.subtables)
   }
 
+  private final case class Grouped(src: Fix[LogicalPlan], keys: Set[Fix[LogicalPlan]])
+
   private final case class CompilerState(
     tree:         AnnotatedTree[Expr, Annotations],
     fields:       List[String],
     tableContext: List[TableContext],
-    nameGen:      Int)
+    nameGen:      Int,
+    grouped:      Option[Grouped])
 
   private object CompilerState {
     /**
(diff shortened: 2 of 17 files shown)


jdegoes:
:+1: :boom: :fireworks: :+1: 

Will do miracles for customer support. :)

I'd like to see a test (could be planner or integration level, your choice) that the logic can handle trivial expression equality in queries like `select age*2 from foo group by age*2`)

jdegoes:
@sellout I'm going to do a release on this today, when you get to a good breaking point, can you review this? Thanks!

it/src/test/resources/tests/groupByFlatten.test
```diff
@@ -1,7 +1,7 @@
 {
     "name": "group by flattened field",
     "data": "slamengine_commits.data",
-    "query": "select distinct substring(parents[*].sha, 0, 1), count(*) from slamengine_commits group by substring(parents[*].sha, 0, 1)",
```
mossprescott: @jdegoes: Here's one example of an expression being used the key and in a projection.

jdegoes: But with the `containsAtLeast` predicate, the result is not meaningful since there could very well be duplicates.

sellout:
This looks really good. But, I’m curious about the “dubious” failing case you mentioned earlier. What is that?

mossprescott:
An example is `select count(city), city from zips group by city`, which produces the wrong result, compared to `select count(*), city from zips group by city` (and there's a pending test for that). Even more perverse is something like `select min(city) ... group by city`.

sellout:
Ah, yeah. I saw that one.

it/src/test/resources/tests/groupByFlatten.test
mossprescott: Isn't it `containsExactly`? The next test is also `containsExactly`, with a less-complex expression. Also there's one more test I missed that I'll add to the branch

jdegoes:
> An example is select count(city), city from zips group by city, 

This is a really good example of why we need to improve the rule for adding `arbitrary`. 

What does this translate to now, or does it just flat-out fail?

it/src/test/resources/tests/groupByFlatten.test
jdegoes: No, it's `containtsAtLeast` (arguably wrong), although the similar one following is `containsExactly`.

Ready to merge when you add the test.

mossprescott:
That particular query doesn't fail, but you end up with the total count after grouping, not the count for each group, so `16584` in every row. This PR *always* adds Arbitrary, so you get `select(count(arbitrary(city)) ...`, which I decided isn't really wrong, because that doesn't make sense in standard SQL anyway.

jdegoes:
>  This PR always adds Arbitrary, so you get select(count(arbitrary(city)) ..., which I decided isn't really wrong, because that doesn't make sense in standard SQL anyway.

OK, let's devote some brain space to this in the future. A more sensible (less surprising?) thing may be adding it if it's not already reduced.

mossprescott:
Makes sense.

PR #895, opened 2015-08-25
Mongo Connection String parser & etc.

MasseGuillaume:
The goal of this PR is to
- [x] ~~replace [this regex](https://github.com/slamdata/slamengine/blob/368e07ebf41d3487be648728834d0d726a9aa475/core/src/main/scala/slamdata/engine/config/config.scala#L63) with a [rfc3986][rfc3986] compliant parser~~
- [x] Convert `host`/`addHosts` to `NonEmptyList[(String, Option[Int])]` (see [todo][todo])
- [x] Convert `opts` to a ~~`Map[String, String]`~~ `Map[String, NonEmptyList[String]]`  (see [todo][todo])
[rfc3986]: https://www.ietf.org/rfc/rfc3986.txt
[todo]: https://github.com/slamdata/slamengine/blob/368e07ebf41d3487be648728834d0d726a9aa475/core/src/main/scala/slamdata/engine/config/config.scala#L78-L79

### DIFF
--- a/core/src/main/scala/quasar/config/config.scala
+++ b/core/src/main/scala/quasar/config/config.scala
@@ -28,6 +28,8 @@ import scalaz._, Scalaz._
 import scalaz.concurrent.Task
 import pathy._, Path._
 
+import com.mongodb._
+
 final case class SDServerConfig(port0: Option[Int]) {
   val port = port0.getOrElse(SDServerConfig.DefaultPort)
 }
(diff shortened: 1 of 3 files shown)


jdegoes:
[This](https://github.com/slamdata/purescript-uri/tree/master/src/Data) might be a useful reference. Or might not, depending on how well you read Haskell / PureScript. :wink:

jdegoes:
>  Convert opts to a Map[String, String] (see todo)

This is actually lossy, because there may exist more than one query string parameter with the same name. The correct type for such a thing is `Map[String, NonEmptyList[String]]`.

sellout:
Is there not already an RFC3986-compliant URI library for Scala? Seems like something, say, http4s would also want to use. So, even if one needs to be written (and hey, might as well write one that uses slamdata/scala-pathy), perhaps it could be a separate library.

MasseGuillaume:
@sellout good call !

I will use this one instead: https://github.com/http4s/http4s/blob/v0.8.3/core/src/main/scala/org/http4s/Uri.scala#L125

MasseGuillaume:
There is a small inconsistency in the [mongodb connection string](https://docs.mongodb.org/v3.0/reference/connection-string/#connection-string-uri-format)

`mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]`

`host1:port1,host2:port2` is not valid per [rfc3986#section-3.2.2](https://tools.ietf.org/html/rfc3986#section-3.2.2). Meanwhile `host1,host2:port` is valid.

mossprescott:
I'm only a little bit surprised to find out that MongoDB's "connection
strings" aren't actually valid URIs :-P

The reference you want to be looking at is
https://docs.mongodb.org/v3.0/reference/connection-string/. This might
indeed make it hard to use a standards-compliant parser, though.

By the same token, mongodb connection string parameters can't have multiple
values in any meaningful way, although the format probably doesn't forbid
them appearing. It might be with looking into how the drivers handle that.

jdegoes:
We have to support the more general (if non-conforming) MongoDB format.

As an aside, requiring that multiple hosts all use the same port (as per RFC) seems a bit onerous...

MasseGuillaume:
@jdegoes @mossprescott @sellout this is ready for review.

jdegoes:
Review by @mossprescott

MasseGuillaume:
> Maybe it makes sense to go one step further and make it a proper case class, and make it the ?
> parameter in MongoDbConfig, but perhaps not as that could tie the serialized configuration to the 
> particular MongoDB driver version.

@mossprescott I propose we leave this for a separate PR.

wemrysi:
@MasseGuillaume would you please resolve the merge conflicts with master?

MasseGuillaume:
https://github.com/sbt/sbt/issues/2217

core/src/main/scala/quasar/config/config.scala
```diff
@@ -54,36 +56,73 @@ final case class MongoDbConfig(connectionUri: String) extends BackendConfig {
...
-        case _ => None
-      }
+    def unapply(uri: String): Option[(
+      Option[String], // Username
+      Option[String], // Password
+      NonEmptyList[(String, Option[Int])],
+      Option[String], 
+      MongoConnectionOption)] = {
+      import scala.collection.JavaConverters._
+      \/.fromTryCatchNonFatal{
```
wemrysi: If `new ConnectionString(uri)` succeeds, then we can assume all the components of the URI are valid (it's a shame properties like `getHosts` don't use more precise types to reflect this, or extract the port number even). So in this case, and because we're crushing any errors into `None` anyways, you could probably refactor this to something like

```scala
for {
  cs    <- \/.fromTryCatchNonFatal(new ConnectionString(uri)).toOption
  hosts <- cs.getHosts.asScala.toList.toNel
} yield {
 ...
}
```
and avoid having to come up with error messages only to throw them away.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -54,36 +56,73 @@ final case class MongoDbConfig(connectionUri: String) extends BackendConfig {
...
+        val cs = new ConnectionString(uri)
+        val hosts = cs.getHosts.asScala.toList.map{hp =>
+          val hpSep = ':'
+          val hpSepCount = hp.count(_ == hpSep)
+          if(hpSepCount == 1) {
```
wemrysi: This logic doesn't handle IPv6 hostnames (which contain multiple `:`). You'll probably need to add some checks similar to what the [`ConnectionString` implementation](https://github.com/mongodb/mongo-java-driver/blob/master/driver-core/src/main/com/mongodb/ConnectionString.java#L621) uses (again, would have been nice if the driver would have extracted the ports for us while it was validating them).

MasseGuillaume: done in e36ba27261a232a9b5df6f8e724b343fed2ed526

core/src/main/scala/quasar/config/config.scala
MasseGuillaume: hum not sure about the second part. the big trycatch block will catch incorrect cast in parameters (intValue & etc).

does this make sense: aa1f8724a82dd457d9e8a90e3e849ec68cfdfe0f ?

wemrysi: You've handled`nulls` with `Option` and, AFAIK, `intValue` or `booleanValue` shouldn't throw anything as they're just returning the wrapped primitives.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -49,41 +51,81 @@ sealed trait BackendConfig {
...
-    // TODO: Convert host/addHosts to NonEmptyList[(String, Option[Int])] and
-    //       opts to a Map[String, String]
-    def unapply(uri: String):
-        Option[(Option[String], Option[String], String, Option[Int], Option[String], Option[String], Option[String])] =
-      uri match {
-        case UriPattern(user, pass, host, port, addHosts, authDb, opts) =>
-          Some((Option(user), Option(pass), host, Option(port).flatMap(_.parseInt.toOption), orNone(addHosts), Option(authDb), Option(opts)))
-        case _ => None
+    def validate(uri: String): \/[String, (
+      Option[String],                       // Username
```
jedesah: I think it might be a good idea to create a case class for this instead of putting comments like this.

MasseGuillaume: yes 100% agree with you. I just wanted to change the implementation (Regex) to something more robust. Right now, in the code base only unapply is used to extract some information.

jedesah:
After reviewing this PR, @wemrysi and I came to the conclusion that it would be much better if `MongoDbConfig` was safe by construction instead of relying on the validate function which was intended for another purpose. Therefore I submitted a PR that adopts a different approach to improving the situation vis-à-vis the mongo connection string. It's not perfect as we fall back to the java driver definition of a `ConnectionString`, but it's better in the sense that a `MongoDbConfig` is guaranteed to have a valid connection String. One day it might make sense to use one of the many MongoDB Scala drivers and use it's notion of a connection string. I for one, would like to see a compile time checked string interpolator for connection Strings and I think this would be a great contribution to such a driver library.

Thank you @MasseGuillaume for the effort you put into this PR and for bringing to our attention issues with the MongoDB connection strings. Your PR was merged in spirit and we hope to avoid leaving your PRs hanging for one month in the future.

jedesah:
For reference, the other PR was #944

PR #900, opened 2015-08-29
update to http4s 0.8.6

### LINKED ISSUE #886: stack overflow when making multiple requests
This might not be immediately reproducible as it occurs while I'm testing slamdata/slamdata#409, so until that is merged I can't quite figure out how else to make it happen.

Basically though, if I try and run a markdown cell with a query like this embedded in it 3 or more times (2 is fine):
```sql
SELECT city FROM "/testDb/zips" LIMIT 500
```

It results in this:
```
Exception in thread "pool-1-thread-2" java.lang.StackOverflowError
        at scalaz.stream.Process$.fail(Process.scala:786)
        at org.http4s.blaze.util.ProcessWriter$class.org$http4s$blaze$util$ProcessWriter$$Try(ProcessWriter.scala:104)
        at org.http4s.blaze.util.ProcessWriter$class.org$http4s$blaze$util$ProcessWriter$$go(ProcessWriter.scala:79)
        at org.http4s.blaze.util.ProcessWriter$$anonfun$org$http4s$blaze$util$ProcessWriter$$go$5.apply(ProcessWriter.scala:66)
        at org.http4s.blaze.util.ProcessWriter$$anonfun$org$http4s$blaze$util$ProcessWriter$$go$5.apply(ProcessWriter.scala:65)
        at scalaz.concurrent.Future$$anonfun$runAsync$1.apply(Future.scala:142)
        at scalaz.concurrent.Future$$anonfun$runAsync$1.apply(Future.scala:142)
        at scalaz.concurrent.Future.listen(Future.scala:76)
        at scalaz.concurrent.Future.runAsync(Future.scala:142)
        at scalaz.concurrent.Task.runAsync(Task.scala:152)
        at org.http4s.blaze.util.ProcessWriter$class.org$http4s$blaze$util$ProcessWriter$$go(ProcessWriter.scala:65)
        at org.http4s.blaze.util.ProcessWriter$$anonfun$org$http4s$blaze$util$ProcessWriter$$go$5.apply(ProcessWriter.scala:66)
        at org.http4s.blaze.util.ProcessWriter$$anonfun$org$http4s$blaze$util$ProcessWriter$$go$5.apply(ProcessWriter.scala:65)
        at scalaz.concurrent.Future$$anonfun$runAsync$1.apply(Future.scala:142)
        at scalaz.concurrent.Future$$anonfun$runAsync$1.apply(Future.scala:142)
        at scalaz.concurrent.Future.listen(Future.scala:76)
        at scalaz.concurrent.Future.runAsync(Future.scala:142)
        at scalaz.concurrent.Task.runAsync(Task.scala:152)
        ... (repeat ProcessWriter to Task stack ad nauseam)
```

This might happen a few times when attempting to run the cell, but then at a certain point it seems to start working, and from then on running it is fine until I kill the server and then it starts happening again. It seems it may be related to the size of the data set, as running it with smaller limits is always fine.

The odd thing is, the queries are run in series when evaluating the cell, so it shouldn't make a difference to slamengine even if the cell I'm trying to run has that query in it once or a thousand times.

mossprescott:
Fixes #886 (now known as SD-888), a stack overflow when reading several
thousand rows via the API with gzip encoding.

The bug was fixed in http4s, which is updated here. This commit adds
integration tests of the backend and of the API implementation, showing that
each is capable of emitting a 100k rows result set.

### DIFF
--- a/it/src/test/scala/slamdata/engine/fs/filesystem.scala
+++ b/it/src/test/scala/slamdata/engine/fs/filesystem.scala
@@ -7,6 +7,7 @@ import slamdata.engine.fs._, Path._
 
 import org.specs2.execute.{Result}
 import org.specs2.scalaz.DisjunctionMatchers
+import scalaz._, Scalaz._
 import scalaz.concurrent._
 import scalaz.stream._
 
@@ -15,10 +16,12 @@ class FileSystemSpecs extends BackendTest with DisjunctionMatchers {
   import Errors._
   import slamdata.engine.fs._
 
-  def oneDoc: Process[Task, Data] =
+  val oneDoc: Process[Task, Data] =
     Process.emit(Data.Obj(ListMap("a" -> Data.Int(1))))
-  def anotherDoc: Process[Task, Data] =
+  val anotherDoc: Process[Task, Data] =
     Process.emit(Data.Obj(ListMap("b" -> Data.Int(2))))
+  def manyDocs(n: Int): Process[Task, Data] =
+    Process.range(0, n).map(n => Data.Obj(ListMap("a" -> Data.Int(n))))
 
   tests {  case (backendName, fs) =>
     val TestDir = testRootDir(fs) ++ genTempDir.run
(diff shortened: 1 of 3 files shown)


mossprescott:
@sellout: I had a simple, dumb bug in the test we were looking at (see https://github.com/http4s/http4s/issues/400#issuecomment-135931371). Embarrassing, but I'm glad to have it fixed.

jdegoes:
@mossprescott You initially had some concerns this wasn't perhaps fixed (even with the patch in http4s), are those all cleared up now?

mossprescott:
Yes, it turned out I had a bug in my test, which bryce-anderson found for
me.

jdegoes:
:+1:

PR #903, opened 2015-09-01
Upgrade Argonaut from 6.1-M6 to 6.1

jedesah:
6.1-M6 was being evicted by sbt anyway and producing an annoying warning message


wemrysi:
:+1:

PR #904, opened 2015-09-01
Document and test behavior of Backend.scan

jedesah:
Implementation of scan contained some bizarre wizardry that I believe is worthwhile to document and test. Update comment that was out of date.

### DIFF
--- a/core/src/main/scala/slamdata/engine/physical/mongodb/filesystem.scala
+++ b/core/src/main/scala/slamdata/engine/physical/mongodb/filesystem.scala
@@ -37,6 +37,10 @@ trait MongoDbFileSystem extends PlannerBackend[Workflow.Crystallized] {
       e => Process.eval[ETask[ResultError, ?], Data](EitherT.left(Task.now(ResultPathError(e)))),
       col => {
         val skipper = (it: com.mongodb.client.FindIterable[org.bson.Document]) => it.skip(offset.toInt)
+        // NB As per the MondoDB documentation, we inverse the value of limit in order to retrieve a single batch.
+        // The documentation around this is very confusing, but it would appear that proceeding this way allows us
+        // to retrieve the data in a single batch. Behavior of the limit parameter on large datasets is tested in
+        // FileSystemSpecs and appears to work as expected.
         val limiter = (it: com.mongodb.client.FindIterable[org.bson.Document]) => limit.map(v => it.limit(-v.toInt)).getOrElse(it)
 
         val skipperAndLimiter = skipper andThen limiter
--- a/core/src/main/scala/slamdata/engine/backend.scala
+++ b/core/src/main/scala/slamdata/engine/backend.scala
@@ -121,16 +121,33 @@ sealed trait Backend { self =>
 
   // Filesystem stuff
 
+  /**
+   * A stream of data found at the specified path
+   * @param path The path for which to retrieve the specified data
+   * @param offset The offset from which to start streaming the Data. If the collection contains 10 elements, specifying
+   *               an offset of 5 will stream the last 5 data elements in the collection. offset must be a positive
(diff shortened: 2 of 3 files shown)


core/src/main/scala/slamdata/engine/physical/mongodb/filesystem.scala
```diff
@@ -37,6 +37,10 @@ trait MongoDbFileSystem extends PlannerBackend[Workflow.Crystallized] {
       e => Process.eval[ETask[ResultError, ?], Data](EitherT.left(Task.now(ResultPathError(e)))),
       col => {
         val skipper = (it: com.mongodb.client.FindIterable[org.bson.Document]) => it.skip(offset.toInt)
+        // NB As per the MondoDB documentation, we inverse the value of limit in order to retrieve a single batch.
+        // The documentation around this is very confusing, but it would appear that proceeding this way allows us
+        // to retrieve the data in a single batch. Behavior of the limit parameter on large datasets is tested in
+        // FileSystemSpecs and appears to work as expected.
         val limiter = (it: com.mongodb.client.FindIterable[org.bson.Document]) => limit.map(v => it.limit(-v.toInt)).getOrElse(it)
```
wemrysi: After staring at the docs a bit I _think_ the effect this has is, for some limit L and a batch size B, to return a single batch of documents of size min(L, floor(B / ||document||)), where ||document|| is the size in bytes of a document. Thus, when L is very large and/or documents themselves are very large, it appears that fewer than L documents may be returned even when the cardinality of the collection is >= L.

In other words, it is going to return exactly one batch of documents, if L documents do not fit in a single batch, tough luck.

jedesah: At one point (depending on which page of the documentation I was looking at), that was my understanding as well, however I would have expected the below test to fail if that were the case though. I also tried it the below test with larger values and it worked as well, so I reverted to make the test faster.

wemrysi: The [documentation](http://docs.mongodb.org/manual/reference/method/cursor.batchSize/) for the mongo shell `cursor.batchSize` seems to corroborate our understanding. To get the test to fail, I _think_ you'd have to have more documents than would fit into a single BSON document, which appears to be 16MB, and then set a limit that would exceed this size and notice fewer documents returned.

That said, could we just pass the positive value to `limit` and sidestep this alltogether? We're already handling closing of the cursor via the `resource` combinator, so I'm not sure if using the negative limit is buying us anything.

jdegoes: > That said, could we just pass the positive value to limit and sidestep this alltogether? We're already handling closing of the cursor via the resource combinator, so I'm not sure if using the negative limit is buying us anything.

:+1:

wemrysi:
@jedesah The comments look good, and much appreciated.

I think if we make the change to `limiter` discussed in the other comments then we don't need the additional test, correct?

jedesah:
@wemrysi I am a little wary of changing something I failed to wrap my head around. But if you are quite confident that making the change to a positive number is a strict improvement over the current implementation, then we can go ahead and do it. In theory, it would break one of our tests if it were to change the semantics in a bad way.

wemrysi:
@jedesah Fair enough, would you mind opening an issue to investigate this particular behavior so we don't forget about it?

Also, I think we should either omit the new test, or devise one that requests more data than will fit into a single `BSON` document. As-is, I don't think it requests enough data to give us confidence that we aren't going to prematurely truncate a response from `scan0` involving a limit.

jedesah:
@wemrysi Should be good to go now! :-)

jedesah:
These two commits disappeared from master. It probably has to do with the renaming to quasar.

PR #907, opened 2015-09-03
Fix #922 Fail-fast integration tests when no backend is configured

jedesah:


it/src/test/scala/slamdata/engine/backendtest.scala
```diff
@@ -15,28 +15,28 @@ import slamdata.engine.config._
...
 
-  private def envName(backend: String): String = "SLAMDATA_" + backend.toUpperCase
+  def envName(backend: String): String = "SLAMDATA_" + backend.toUpperCase
 
+  /**
+   * Load backend config from environement variable.
+   * Fails if it cannot parse the config
+   * Returns None if there is no config.
+   * @param name
+   * @return
```
sellout: Use or remove these JavaDoc tags. (We tend not to use them.)

it/src/test/scala/slamdata/engine/backendtest.scala
```diff
@@ -53,7 +53,12 @@ trait BackendTest extends Specification {
         }
       }
     } yield backend.map(name -> _)
-  }).sequenceU.map(_.flatten)
+  }).sequenceU.flatMap(_.flatten match {
+    case Nil => Task.fail(new java.lang.Exception(
+      s"No backend to test. Consider setting one of these environment variables ${TestConfig.backendNames.map(TestConfig.envName)}"
```
sellout: This ends up dumping a stack trace, right? Any way we can present it a bit more nicely to the user? EG, handling `Throwable` in `tests` and outputting the message.

jdegoes: A simple equation: `Task.fail = #fail`. :laughing:

sellout: Yeah, perhaps … but this is in tests, as long as it blows up when something goes wrong 😉

Not that I wouldn’t prefer a disjunction or something …

sellout: So, we decided to leave the current approach in, because it seems to be the only way to make the tests actually fail (at least until we move to less-magical specs2 3.x … which requires scalaz 7.2.0 …).

PR #909, opened 2015-09-04
Feature/blaze client

MasseGuillaume:
Not ready for review.

Goals
- [x] remove dependency on dispatch since we use http4s
- [x] improve code readability

Findings
* look like there is a [CORS middleware built-in http4s](https://github.com/http4s/http4s/blob/v0.8.6/server/src/main/scala/org/http4s/server/middleware/CORS.scala)
* not sure if it uses [gzip middleware](http://http4s.org/api/0.8/index.html#org.http4s.server.middleware.GZip$) or not
* inconsistencies in HTTP 404 response: sometime returns json, sometime returns string
* when PUT creates a new ressource return 201

TODO
- [x] create another pr for repl/sbt
- [x] rebase/sync


mossprescott:
@MasseGuillaume To get this merged, I think it would help to separate it into separate PRs, at least one for fixing the console and another for switching to http4s's client.

MasseGuillaume:
@mossprescott no problemo!

MasseGuillaume:
This is ready for review

wemrysi:
@MasseGuillaume would you mind splitting this into two commits? One with all the semantic changes and additions and another with the renames and other organizational changes (like extracting `ResponseFormatSpecs` into its own file)? This will make it much easier to review.

MasseGuillaume:
@wemrysi sure. doing this now.

wemrysi:
@MasseGuillaume Thanks, much appreciated!

build.sbt
```diff
@@ -48,26 +44,33 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
...
-    "org.typelevel"     %% "shapeless-scalaz"          % slcVersion     % "compile, test",
-    "com.slamdata"      %% "pathy"                     % "0.0.1-SNAPSHOT" % "compile, test",
-    "com.github.mpilquist" %% "simulacrum"             % "0.3.0"        % "compile, test",
-    "org.scalaz"        %% "scalaz-scalacheck-binding" % scalazVersion  % "test",
-    "org.specs2"        %% "specs2-core"               % "2.4"          % "test",
-    "org.typelevel"     %% "scalaz-specs2"             % "0.3.0"        % "test",
-    "org.typelevel"     %% "shapeless-scalacheck"      % slcVersion     % "test",
-    "net.databinder.dispatch" %% "dispatch-core"       % "0.11.1"       % "test"),
+  libraryDependencies ++= {
+    val scalazVersion  = "7.1.3"
```
jedesah: I like this formatting! :+1:

MasseGuillaume:
rebasing on master

MasseGuillaume:
stange ivy resolution failed..

MasseGuillaume:
https://github.com/http4s/http4s/issues/415

build.sbt
```diff
@@ -23,6 +19,7 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
   },
   autoCompilerPlugins := true,
   exportJars := true,
+  fullResolvers -= "jcenter" at "https://jcenter.bintray.com/", // until https://github.com/http4s/http4s/issues/415 is fixed
```
jedesah: This particular issue has been fixed since it's not http4s' fault. The comment should track the actual issue as a future maintainer might believe we can remove this now. Also, as a side comment, it might be worth mentioning this issue to the sbt team just so that they are aware of the impact of their change.

MasseGuillaume: make sense, let's see what's the result of this issue: https://github.com/sbt/sbt/issues/2217

jedesah:
GitHub is sucking and won't let me add inline comments because apparently 1000 lines is too much...
- **line 60**: Why the change from () => List to ListBuffer? We made it that way to convey the fact that it may change as test code gets executed. If there is a good reason for this change, let us know and update the documentation to explain the behavior, if not, revert the change and maybe update the documentation so that the next person does not make the same mistake :-)
- **line 47** on the old file: Why remove the comment about extending specs2?
- **line 67**: Any particular reason to fully qualify the mutable `ListBuffer`?
- **line 252**: Why remove the comment here?
- **line 347**: How are we checking that we get NotFound?
- **line 352**: Idem
- **line 357**: Idem
I think I will stop here for now

wemrysi:
In `web/src/test/scala/quasar/api/fs.scala`:

* There appears to be a fair amount of trailing whitespace introduced, including lines consisting of only trailing whitespace
* Two examples testing server restarts, on **lines 1838-1870** in the previous version, seem to have gone missing. Were they moved somewhere that I just missed?
* Can we rename `coorsHeaders` to `corsHeaders`, similarly with `coors`? Does the extra "o" signify something?
* How about changing the various `withXXXX` methods to accept a function returning a `Task[A]`, this would get rid of all the calls to `run` everywhere
* What does the `1` suffix in `Mount1` signify? Can we rename to `Mount`? Also, how about renaming the `type` property to `kind` just to avoid having to escape it?
* Could we rename the handful of methods which have a variant ending in "2" to end in "Uri" instead? I'd also be ok with using overloaded methods in this case.

MasseGuillaume:
- [ ] line 60: Why the change from () => List to ListBuffer? We made it that way to convey the fact that it may change as test code gets executed. If there is a good reason for this change, let us know and update the documentation to explain the behavior, if not, revert the change and maybe update the documentation so that the next person does not make the same mistake :-)
- [x] line 47 on the old file: Why remove the comment about extending specs2? **none undone**
- [x] line 67: Any particular reason to fully qualify the mutable ListBuffer? **none added import**
- [x] line 252: Why remove the comment here? **none undone**
- [ ] line 347: How are we checking that we get NotFound? **I dont understand your question**
- [ ] line 352: Idem
- [ ] line 357: Idem I think I will stop here for now
- [ ] There appears to be a fair amount of trailing whitespace introduced, including lines consisting of - only trailing whitespace
- [ ] Two examples testing server restarts, on lines 1838-1870 in the previous version, seem to have gone missing. Were they moved somewhere that I just missed? **ops I will put them back, I might need your help here**
- [x] Can we rename coorsHeaders to corsHeaders, similarly with coors? Does the extra "o" signify something? **renamed**
- [ ] How about changing the various withXXXX methods to accept a function returning a Task[A], this would get rid of all the calls to run everywhere **I think spec3 has support for Task, I prefer to leave it as is so when we upgrade it does not change that much**
- [ ] What does the 1 suffix in Mount1 signify? Can we rename to Mount? Also, how about renaming the type property to kind just to avoid having to escape it? **I just took a shortcut to disambiguate things, I dont understand the second part**
- [x] Could we rename the handful of methods which have a variant ending in "2" to end in "Uri" instead? I'd also be ok with using overloaded methods in this case. **done**

jedesah:
@MasseGuillaume Nevermind my comment about line 347 through 357. I was looking quickly and did not realized that you factored out expectations. It's a good change!

As a side note, what does the `expectFs` name mean? We might want to consider another name since it was not obvious to me what it is doing. It seems to be expecting a json response.

jedesah:
@MasseGuillaume specs2 3.x has support for `Task`? That is news to me. Do you have a link? Wouldn't the code change less when upgrading to specs2 3.x if you were to follow @wemrysi's suggestion about the withXXXX methods? I have to agree that `run`'s everywhere kinda suck. With `dispatch` it would have been a pain in the ass to have that signature and have to do `Task.delay{...}` everywhere, but now it would best I think. It would also allow us to put a timeout on all the requests in one go to avoid the tests hanging if the server were to misbehave or something (does http4s timeout on it's own?).

jedesah:
@MasseGuillaume What do you mean exactly, when you say the test fails with an empty list? It seemed to be working previously.

MasseGuillaume:
@jedesah I guess the evaluation order is different with Task from http4s

jedesah:
@MasseGuillaume Could you elaborate on why it's not possible to use the function with Task, you can make that list into a `Task` if it makes things easier, that would be another good way of attracting attention to the effectfullness/mutability of that List.

jedesah:
@MasseGuillaume I would really like to get this PR merged ASP for many reasons. Maybe tomorrow (Friday) or on the weekend we can pair up and work at getting to a state where it can be merged.

PR #910, opened 2015-09-07
[SD-876] Only restart api server when port number changed

wemrysi:
Changes the behavior of `FileSystemApi` to only restart the api server when the port number is changed. For mounting changes, a local cache of the configuration is updated (in a `TaskRef`) after writing the updated config to disk.

The server main is also refactored in terms of `scalaz.stream.Process` which makes running + restarting uniform and allows us to test restarting the server deterministically.

### DIFF
--- a/web/src/main/scala/slamdata/engine/api/fs.scala
+++ b/web/src/main/scala/slamdata/engine/api/fs.scala
@@ -124,12 +124,20 @@ object ResponseFormat {
   }
 }
 
-final case class FileSystemApi(backend: Backend, contentPath: String, config: Config, tester: BackendConfig => EnvTask[Unit], reloader: Config => Task[Unit]) {
+final case class FileSystemApi(
+  backend: Backend, contentPath: String, initialConfig: Config,
+  tester: BackendConfig => EnvTask[Unit],
+  reloader: Config => Task[Unit],
+  configWriter: Config => Task[Unit]) {
+
   import Method.{MOVE, OPTIONS}
 
   val CsvColumnsFromInitialRowsCount = 1000
 
   val LineSep = "\r\n"
+
+  private def writeConfig(cache: TaskRef[Config], cfg: Config): Task[Unit] =
+    configWriter(cfg) *> cache.write(cfg)
 
   private def rawJsonLines[F[_]](codec: DataCodec, v: Process[F, Data])(implicit EE: EncodeJson[DataEncodingError]): Process[F, String] =
     v.map(DataCodec.render(_)(codec).fold(EE.encode(_).toString, ɩ))
@@ -317,23 +325,27 @@ final case class FileSystemApi(backend: Backend, contentPath: String, config: Co
     }
   }
 
-  def serverService(config: Config, reloader: Config => Task[Unit]) = {
+  def serverService(ref: TaskRef[Config]) = {
     HttpService {
-      case req @ PUT -> Root / "port" => for {
-        body <- EntityDecoder.decodeString(req)
-        r    <- body.parseInt.fold(
-          e => NotFound(e.getMessage),
-          i => for {
-            _    <- reloader(config.copy(server = SDServerConfig(Some(i))))
-            // TODO: If the requested port is unavailable the server will restart
-            //       on a random one, thus this response text may not be accurate.
-            resp <- Ok("changed port to " + i)
-          } yield resp)
-      } yield r
+      case req @ PUT -> Root / "port" =>
+        EntityDecoder.decodeString(req).flatMap(body =>
+          body.parseInt.fold(
+            e => NotFound(e.getMessage),
+            i => for {
+              cfg  <- ref.read
+              _    <- reloader(cfg.copy(server = SDServerConfig(Some(i))))
+              // TODO: If the requested port is unavailable the server will restart
+              //       on a random one, thus this response text may not be accurate.
+              resp <- Ok("changed port to " + i)
+            } yield resp))
+
       case DELETE -> Root / "port" => for {
-        _    <- reloader(config.copy(server = SDServerConfig(None)))
-        resp <- Ok("reverted to default port")
+        cfg  <- ref.read
+        dprt =  SDServerConfig.DefaultPort
+        _    <- reloader(cfg.copy(server = SDServerConfig(Some(dprt))))
+        resp <- Ok(s"reverted to default port ($dprt)")
       } yield resp
+
       case req @ GET -> Root / "info" =>
         Ok(versionAndNameInfo)
     }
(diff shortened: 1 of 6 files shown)


web/src/main/scala/slamdata/engine/api/fs.scala
```diff
@@ -343,73 +357,78 @@ final case class FileSystemApi(backend: Backend, contentPath: String, config: Co
...
-            else
-              -\/(Conflict(errorBody("Can't add a mount point above the existing mount point at " + k, None)))))
-        }.sequenceU.fold(
-          ɩ,
-          κ(respond(for {
-            upd <- addPath(path, req)
-          } yield (if (upd) "updated" else "added") + " " + path)))
+        ref.read.flatMap(cfg =>
+          (if (cfg.mountings contains path) None else ensureNoOverlaps(cfg, path)) getOrElse
+          respond(addPath(cfg, path, req) map (upd => s"${upd ? "updated" | "added"} $path")))
```
mossprescott: Although we have no explicit policy, we've been avoiding string interpolation almost entirely, and certainly not putting logic in it. I know @jdegoes is not a fan; good time to make a decision and maybe add something to the wiki?

web/src/main/scala/slamdata/engine/api/server.scala
```diff
@@ -27,11 +27,17 @@ import scala.concurrent.duration._
...
-import scalaz.effect._
+import scalaz.stream._
+
+import org.http4s.server.{Server => Http4sServer}
+import org.http4s.server.blaze.BlazeBuilder
 
 object Server {
-  private val serv =
-    IO.newIORef(None: Option[org.http4s.server.Server]).unsafePerformIO
+  private def info(msg: => String): Task[Unit] =
```
mossprescott: A good idea but can we find a better home for these?

web/src/main/scala/slamdata/engine/api/server.scala
```diff
@@ -48,51 +54,85 @@ object Server {
...
+      case -\/(err)                         => Task.fail(err)
+      case \/-(s)                           => Task.delay(s.close()).as(None)
+    })
 
-    config => for {
-      _       <- configWriter(config)
-      _       <- runAsync(restart(config).run.map(ignore))
-    } yield ()
+  /** An available port number. */
+  def anyAvailablePort: Task[Int] = Task.delay {
```
mossprescott: Code moving around makes it hard to follow this diff. Which I'm ok with if there's a good reason to re-order things. Was there?

web/src/main/scala/slamdata/engine/api/fs.scala
```diff
@@ -317,21 +325,27 @@ final case class FileSystemApi(backend: Backend, contentPath: String, config: Co
...
+              //       on a random one, thus this response text may not be accurate.
+              resp <- Ok("changed port to " + i)
+            } yield resp))
+
       case DELETE -> Root / "port" => for {
-        _    <- reloader(config.copy(server = SDServerConfig(None)))
-        resp <- Ok("reverted to default port")
+        cfg  <- ref.read
+        dprt =  SDServerConfig.DefaultPort
+        _    <- reloader(cfg.copy(server = SDServerConfig(Some(dprt))))
```
mossprescott: I think we want `None` here, just to avoid writing the (current) default port to the config file. If we ever change our minds on it, we don't want to have to ask users to go in and manually delete it.

web/src/main/scala/slamdata/engine/api/fs.scala
```diff
@@ -343,73 +357,78 @@ final case class FileSystemApi(backend: Backend, contentPath: String, config: Co
...
-        if (config.mountings.contains(path))
-          for {
-            _    <- reloader(config.copy(mountings = config.mountings - path))
-            resp <- Ok("deleted " + path)
-          } yield resp
-        else
-          Ok()
+        ref.read.flatMap(cfg =>
+          if (cfg.mountings contains path)
+            writeConfig(ref, cfg.copy(mountings = cfg.mountings - path)) *>
```
mossprescott: Not sure I like the trailing `*>` as an alternative to `for { _ <- ...`, but this is another thing I feel deserves to be approached consistently.

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1697,62 +1714,76 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
     }
+  }
 
-    "restart sequence" should {
-      // Exercises the server restarting code path in server.scala.
-       def withRestartableServer[A](backend: Backend, config: Config)(body: => A): A = {
-         val rez: slamdata.engine.Evaluator.EnvironmentError \/ Int = Server.run(port, config, "", 1.seconds, tester, mounter(backend), configWriter).run.run
+  "/server" should {
+    def withServerExpectingRestart[A, B](backend: Backend, config: Config, timeoutMillis: Long = 10000)
+                                        (causeRestart: => A)(afterRestart: => B): B = {
```
mossprescott: :grinning: I like it.

web/src/main/scala/slamdata/engine/api/fs.scala
jdegoes: The reason for `*>` is that it's strictly less powerful than `>>=` (AKA `flatMap`). It's considered good practice (one I agree with) to use the least specific abstraction that gets the job done. 

However, I think we do need some standardization on `flatMap` versus `for { ... }` since we have mixed styles; also on various syntactical alternatives to `ap` (`|@|`, etc.).

web/src/main/scala/slamdata/engine/api/server.scala
jdegoes: Also probably worth pulling in [Journal](https://github.com/oncue/journal) or similar at some point.

web/src/main/scala/slamdata/engine/api/fs.scala
wemrysi: Ah, you're absolutely right, that's much better behavior (and indeed what was there originally!).

web/src/main/scala/slamdata/engine/api/fs.scala
wemrysi: Agree on the need to standardize, I tend (and prefer) to use combinators where possible out of habit for the reason John mentions. I also think they are clearer in many cases, i.e.

```
for {
  a <- fa
} yield g(a)
```
vs `fa map g`, or

```
for {
  a <- fa
  b <- g(a)
} yield b
```
vs `fa flatMap g` or even `fa >>= g`.

That said, I'll try not to change styles just for change's sake, only when it is a strict improvement.

web/src/main/scala/slamdata/engine/api/fs.scala
wemrysi: Ok, thanks for pointing that out, I'll omit the interpolation. This probably isn't the best place to discuss, but what are the arguments against it?

web/src/main/scala/slamdata/engine/api/server.scala
wemrysi: The only reason was to keep the file organized in a more-or-less top-down fashion, but I sympathize with moving code making things hard to understand. I should have moved things in one commit and modified them in another to make things more clear.

web/src/main/scala/slamdata/engine/api/fs.scala
mossprescott: I can't speak for the rest. My own preference would probably be to use simple interpolation but avoid anything more complicated. Good topic for a tech hangout?

web/src/main/scala/slamdata/engine/api/fs.scala
```diff
@@ -124,13 +124,21 @@ object ResponseFormat {
   }
 }
 
-final case class FileSystemApi(backend: Backend, contentPath: String, config: Config, tester: BackendConfig => EnvTask[Unit], reloader: Config => Task[Unit]) {
+final case class FileSystemApi(
+  backend: Backend, contentPath: String, initialConfig: Config,
+  tester: BackendConfig => EnvTask[Unit],
+  reloader: Config => Task[Unit],
```
jedesah: Sorry, a little late to the party, what is the difference between `reloader` and `configWriter`? I think it would not be unreasonable to add a little bit of documentation on this class to explain the various fields.

wemrysi: Yeah, the types here aren't very helpful =(. In short, `reloader` starts a new server with the provided `Config` whilst `configWriter` persists the given `Config`. Consider this a +1 for tracking effects in the types!

jedesah: Not sure I understand, aren't `reloader` and `Config` kind of like callbacks essentially? I don't understand what you mean by
>starts a new server with the provided Config

I really think we should add some ScalaDoc here. I will be glad to do it once I have a better understanding of what the different fields are.

PR #911, opened 2015-09-09
Avoid adding Arbitrary when group keys are already reduced

mossprescott:
Resolves SD-898.

Add a transformation of the LogicalPlan after is is generated from the SQL
AST and before returning it from the compiler. Implement it as a series of
steps: identifying keys, annotating expressions that match the keys,
inserting Arbitrary where no reduction is already present. Two of those
steps use new, fairly generic, folds: `topDownTransform` (for `Cofree`) and
`boundParaAttribute`.

Remove the previous approach to inserting Arbitrary, which added the keys
to the compiler's state.

Shape-preserving functions are now identified with a separate mappingType
from other transformations (grouping and joins), so that keys can be
tracked when filtering happens in between grouping and projections (e.g.
SQL's `HAVING`).

For debugging, added RenderTree instances for several primitive types.


mossprescott:
@sellout: do you want to have a look at this as well?

core/src/main/scala/slamdata/recursionschemes/package.scala
```diff
@@ -321,4 +329,23 @@ package object recursionschemes {
 
     loop(t.unFix, B.initial)
   }
+
+  /**
+   Annotate (the original nodes of) a tree, by applying a function to the
+   "bound" nodes. The function is also applied to the bindings themselves
+   to determine their annotation.
+   */
+  def boundParaAttribute[F[_]: Functor, A](t: Fix[F])(f: Fix[F] => A)(implicit B: Binder[F], R1: RenderTree[Fix[F]], R2: RenderTree[Cofree[F, A]]): Cofree[F, A] = {
```
wemrysi: Can we get rid of the `RenderTree` constraints?

core/src/main/scala/slamdata/engine/compiler.scala
```diff
@@ -615,35 +604,66 @@ trait Compiler[F[_]] {
...
+  import LogicalPlan._
+
   def apply[F[_]]: Compiler[F] = new Compiler[F] {}
 
-  def trampoline = apply[Free.Trampoline]
+  def trampoline = apply[scalaz.Free.Trampoline]
 
   def compile(tree: AnnotatedTree[Expr, Annotations]): SemanticError \/ Fix[LogicalPlan] = {
-    trampoline.compile(tree).run
+    trampoline.compile(tree).run.map(reduceGroupKeys)
```
wemrysi: Does it make sense to move the application of `reduceGroupKeys` into the `Compiler#compile` method on the trait, so that any implementation will make use of it?

core/src/main/scala/slamdata/recursionschemes/package.scala
mossprescott: Ah, yes, that's leftover debugging code.

core/src/main/scala/slamdata/engine/compiler.scala
mossprescott: Yes, I think so.

wemrysi:
:+1: unless @sellout has anything to add

wemrysi:
Whoops, sorry misclick.

core/src/main/scala/slamdata/engine/compiler.scala
```diff
@@ -615,35 +604,66 @@ trait Compiler[F[_]] {
...
+          case InvokeF(func, src :: _) if func.mappingType == MappingType.ManyToMany =>
+            groupedKeys(src.unFix, newSrc)
+          case _ => None
+        }
+      }
+
+      Fix(t.map(_._1)) ->
+        groupedKeys(t.map(_._1), Fix(t.map(_._1))).getOrElse(t.foldMap(_._2))
+    }
+    val keys: List[Fix[LogicalPlan]] = boundCata(tree)(keysƒ)._2
```
sellout: I think this should use `boundPara`, then `keysƒ` doesn’t return a tuple, and you eliminate the `_2` projection here.

core/src/main/scala/slamdata/recursionschemes/package.scala
```diff
@@ -321,4 +329,23 @@ package object recursionschemes {
 
     loop(t.unFix, B.initial)
   }
+
+  /**
+   Annotate (the original nodes of) a tree, by applying a function to the
+   "bound" nodes. The function is also applied to the bindings themselves
+   to determine their annotation.
+   */
+  def boundParaAttribute[F[_]: Functor, A](t: Fix[F])(f: Fix[F] => A)(implicit B: Binder[F]): Cofree[F, A] = {
```
sellout: This should be `f: F[(Fix[F], A)] => A` to be `para`.

core/src/main/scala/slamdata/engine/compiler.scala
```diff
@@ -615,35 +604,66 @@ trait Compiler[F[_]] {
...
+
+    // Step 1: annotate nodes containing the keys.
+    val ann: Cofree[LogicalPlan, Boolean] = boundParaAttribute(tree)(keys contains _)
```
sellout: So, with the change to `boundParaAttribute` I suggested below, this becomes a little more complicated. I think that `A` in `keysƒ` should be unFixed already, so you have a `List[LogicalPlan[Fix[LogicalPlan]]]`, then here, `keys contains _.map(_._1)`

core/src/main/scala/slamdata/engine/compiler.scala
```diff
@@ -615,35 +604,66 @@ trait Compiler[F[_]] {
...
+
+        case _ =>
+          if (t.head) Cofree(false, InvokeF(agg.Arbitrary, List(strip(t))))
+          else t
+      }
+    }
+    val rewritten: Cofree[LogicalPlan, Boolean] = topDownTransform(ann)(rewriteƒ)
+
+    // Step 3: strip the annotation.
+    Recursive[Cofree[?[_], Boolean]].forget(rewritten)
```
sellout: I think you can use `ana` instead of this new fold.
```scala
def rewriteƒ(t: Cofree[LogicalPlan, Boolean]): LogicalPlan[Cofree[LogicalPlan, Boolean]] = {
  def strip(v: Cofree[LogicalPlan, Boolean]) = Cofree(false, v.tail)

  t.tail match {
    case InvokeF(func, arg :: Nil) if func.mappingType == MappingType.ManyToOne =>
      InvokeF(func, List(strip(arg)))

    case _ =>
      if (t.head) InvokeF(agg.Arbitrary, List(strip(t)))
      else t.tail
  }
}
```
then there’s no need to `forget` at the end, either.

sellout:
So, just being picky about folds. Otherwise, looks good.

sellout:
Oh yeah, and I feel like the pair of folds in `reduceGroupKeys` is some kind of (generalized) mutumorphism, which would eliminate the `Cofree`, but we don’t have anything like that yet. Might 10% it at some point.

core/src/main/scala/slamdata/recursionschemes/package.scala
mossprescott: So, what would you suggest as a name, given this signature, or do you think it's actually a bad idea? Since the `A` is captured in Cofree, I didn't think it made sense to include it in `f`'s argument, and `F[Fix[F]]` seemed perverse.

Given that, it's analogous to `attribute`, but deals with bindings, and provides the recursive value, hence the frankensteinish name.

core/src/main/scala/slamdata/engine/compiler.scala
mossprescott: Pretty sure I tried that and it didn't give me what I needed, which was the fully substituted plan at each node. I could try it again and make sure I didn't go in a big circle, which wouldn't be a huge surprise.

core/src/main/scala/slamdata/recursionschemes/package.scala
sellout: I don’t think it’s a bad idea, since you don’t need the `A` anyway. Also, this isn’t really a fold, since it doesn’t matter what order the nodes get processed in.

So, this could be `boundAttribute`, and what we’ve called `attribute` in the past could be renamed to `cataAttribute` (since that’s the algebra it uses).

core/src/main/scala/slamdata/recursionschemes/package.scala
```diff
@@ -321,4 +329,23 @@ package object recursionschemes {
...
+    def loop(t: F[Fix[F]], b: B.G[(Fix[F], Cofree[F, A])]): (Fix[F], Cofree[F, A]) = {
+      val newB = B.bindings(t, b)(loop(_, b))
+      B.subst(t, newB).fold {
+        val m: F[(Fix[F], Cofree[F, A])] = t.map(x => loop(x.unFix, newB))
+        val t1 = Fix(m.map(_._1))
+        t1 -> Cofree(f(t1), m.map(_._2))
+      } { case (x, _) =>
+        x -> attrK(Fix(t), f(Fix(t)))
```
sellout: This bit seems odd to me: `f(Fix(t))` – you’re calling the function on the `Free` that you found a substitution for, when the expectation of `bound…` is that the function will never be called on `Free`.

Shouldn’t it be `f(x)`?

sellout: Also, can we restrict the weird `->` syntax to only places where a tuple represents a key-value mapping?

core/src/main/scala/slamdata/engine/compiler.scala
sellout: Ah, right … `boundPara` gives you the `Free` in the `Fix[F]`, only giving you the `A` from the bound value. I remember hemming and hawing over it for a while.

The reason it landed where it did is because in something like `preferProjectionsƒ`, where `A` is `Fix[F]`, you probably want to be able to preserve at least some of the variable references in the result.

core/src/main/scala/slamdata/recursionschemes/package.scala
sellout: Can you add a test of `bound[Para]Attribute` that ensures `Free`s are maintained in the result while not being passed to `f`. I _think_ they’re maintained, but as I mentioned in my penultimate comment, I think the second condition might be failing.

mossprescott: Working on it, and it looks like you're right. I don't care about annotating those nodes in this case, but it's worth getting right (and having a unit test, regardless).

core/src/main/scala/slamdata/engine/compiler.scala
mossprescott: `ana` doesn't apply because `Cofree` isn't Corecursive. That's the same reason I had to write topDownTransform specifically for Cofree.

sellout: You don’t need `Cofree` to be Corecursive. `Cofree` is the `A`, `Fix` is the `T`.

`Corecursive[Fix].ana(ann)(rewriteƒ)` (the `val rewritten` is unnecessary, just return the result of the `ana`).

sellout:
LGTM. It would be great to unify topDownTransform, but may be impossible 😉

Now just conflicts.

mossprescott:
Should be ready to go now.

PR #912, opened 2015-09-09
Enable deleting dbs and clusters in MongoDB.

sellout:
Also add a `.skipped` test decorator.

SD-761 #done

Review by @mossprescott, @wemrysi.

### DIFF
--- a/core/src/main/scala/slamdata/engine/physical/mongodb/collection.scala
+++ b/core/src/main/scala/slamdata/engine/physical/mongodb/collection.scala
@@ -38,23 +38,28 @@ final case class Collection(databaseName: String, collectionName: String) {
 object Collection {
   import PathError._
 
-  def fromPath(path: Path): PathError \/ Collection = {
-    val rel = path.asRelative
-    val segs = rel.dir.map(_.value) ++ rel.file.map(_.value).toList
-    for {
-      first    <- segs.drop(1).headOption \/> PathTypeError(path, Some("has no segments"))
-      rest     =  segs.drop(2)
-      db       <- DatabaseNameParser(first)
-      collSegs <- rest.map(CollectionSegmentParser(_)).sequenceU
-      _        <- if (collSegs.isEmpty)
-                    -\/(InvalidPathError("path names a database, but no collection: " + path))
-                  else \/-(())
-      coll     =  collSegs.mkString(".")
-      _        <- if (utf8length(db) + 1 + utf8length(coll) > 120)
-                    -\/(InvalidPathError("database/collection name too long (> 120 bytes): " + db + "." + coll))
-                  else \/-(())
-    } yield Collection(db, coll)
+  def foldPath[A](path: Path)(clusterF: => A, dbF: String => A, collF: Collection => A):
+      PathError \/ A = {
+    val abs = path.asAbsolute
+    val segs = abs.dir.map(_.value) ++ abs.file.map(_.value).toList
+    segs match {
+      case Nil => \/-(clusterF)
+      case first :: rest => for {
+        db       <- DatabaseNameParser(first)
+        collSegs <- rest.map(CollectionSegmentParser(_)).sequenceU
+        coll     =  collSegs.mkString(".")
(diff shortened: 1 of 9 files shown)


core/src/main/scala/slamdata/engine/physical/mongodb/collection.scala
```diff
@@ -38,24 +38,29 @@ final case class Collection(databaseName: String, collectionName: String) {
...
-      collSegs <- rest.map(CollectionSegmentParser(_)).sequenceU
-      _        <- if (collSegs.isEmpty)
-                    -\/(InvalidPathError("path names a database, but no collection: " + path))
-                  else \/-(())
-      coll     =  collSegs.mkString(".")
-      _        <- if (utf8length(db) + 1 + utf8length(coll) > 120)
-                    -\/(InvalidPathError("database/collection name too long (> 120 bytes): " + db + "." + coll))
-                  else \/-(())
-    } yield Collection(db, coll)
+  def foldPath[A](path: Path)(clusterF: => A, dbF: String => A, collF: Collection => A):
```
mossprescott: Had to think for a minute about what `clusterF` was supposed to be abbreviating :-)

core/src/test/scala/slamdata/specs2/decorators.scala
```diff
@@ -26,3 +26,16 @@ trait PendingWithAccurateCoverage extends PendingUntilFixed {
 }
 
 object PendingWithAccurateCoverage extends PendingWithAccurateCoverage
+
+trait SkippedDecorator extends PendingUntilFixed {
```
mossprescott: Not sure I see the point, if this is just a new syntax for skipped.

Can we do something like `PendingWithAccurateCoverage` and test an env. variable so that the test runs in Travis?

core/src/main/scala/slamdata/engine/physical/mongodb/collection.scala
```diff
@@ -38,24 +38,29 @@ final case class Collection(databaseName: String, collectionName: String) {
...
+      PathError \/ A = {
+    val abs = path.asAbsolute
+    val segs = abs.dir.map(_.value) ++ abs.file.map(_.value).toList
+    segs match {
+      case Nil => \/-(clusterF)
+      case first :: rest => for {
+        db       <- DatabaseNameParser(first)
+        collSegs <- rest.map(CollectionSegmentParser(_)).sequenceU
```
wemrysi: Any particular reason to prefer this to `rest.traverseU(CollectionSegmentParser(_))`?

core/src/main/scala/slamdata/engine/physical/mongodb/filesystem.scala
```diff
@@ -199,17 +201,23 @@ sealed trait MongoWrapper {
     _ = c.drop
   } yield ()
 
+  def dropDatabase(name: String): Task[Unit] = Task.delay(db(name).drop)
+
+  val dropAllDatabases: Task[Unit] =
+    listDatabases.flatMap(_.map(dropDatabase).sequence.map(_.concatenate))
```
wemrysi: I think `listDatabases.flatMap(_.traverse_(dropDatabase))` is a bit more straightforward.

core/src/test/scala/slamdata/specs2/decorators.scala
sellout: Part of the point is that I don’t know what the _old_ syntax for skipped is. And I don’t know why these things shouldn’t be consistent.

Having a decorator that does something only in isolated environments is even better though. I’ll do that.

core/src/main/scala/slamdata/engine/physical/mongodb/collection.scala
sellout: clusterFunc

core/src/main/scala/slamdata/engine/physical/mongodb/filesystem.scala
sellout: Yes, in both cases.

PR #914, opened 2015-09-10
Refactor tests in web project

jedesah:
In order to make pieces of test infrastructure more easily usable from within other contexts. In particular, I am interested in making the test infrastructure available from within the it project in order to write tests cases for bugs.

### DIFF
--- a/web/src/test/scala/slamdata/engine/api/fs.scala
+++ b/web/src/test/scala/slamdata/engine/api/fs.scala
@@ -10,6 +10,7 @@ import slamdata.engine.fs._, Path._
 import slamdata.specs2._
 
 import scala.concurrent.duration._
+import scala.collection.mutable
 
 import argonaut._, Argonaut._
 import com.ning.http.client.Response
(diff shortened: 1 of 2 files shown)


web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -19,129 +20,151 @@ import scalaz._, Scalaz._
...
 
-  var historyBuff = scala.collection.mutable.ListBuffer[Action]()
-  def history = historyBuff.toList
+  /**
+   * Start a server, with the given backend, execute something, and then tear
+   * down the server.
+   * The body receives a mutable ListBuffer. This buffer contains every Config that
+   * was asked to be reloaded since the server started.
+   */
+  def withServerRecordReloads[A](backend: Backend, config: Config, port: Int = 8888)(body: (Req,mutable.ListBuffer[Config]) => A): A = {
```
sellout: What if you remove the `port` parameter, and then either just make it a `val` in the function for now or use `Server.anyAvailablePort` (which requires weaving `Task`, but allows you to remove the `sequential` flag)?

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1462,30 +1484,30 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
           result() must_== "added /local/"
-          history must_== List(Action.Reload(Config(SDServerConfig(Some(port)), Map(
+          reloads.toList must_== List(Config(SDServerConfig(None), Map(
             Path("/foo/") -> MongoDbConfig("mongodb://localhost/foo"),
             Path("/non/root/mounting/") -> MongoDbConfig("mongodb://localhost/mounting"),
-            Path("/local/") -> MongoDbConfig("mongodb://localhost/test")))))
+            Path("/local/") -> MongoDbConfig("mongodb://localhost/test"))))
         }
-      }
+      }.pendingUntilFixed("Will be changed by outstanding PR")
```
sellout: This outstanding PR (#910) has been merged, so you should merge and remove these `pending`s.

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1735,30 +1759,27 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
       try {
         (forCfg(Some((port, config))) *> exec).runFor(timeoutMillis)
       } finally {
-        historyBuff.clear
+        // Do we need to close down servers here?
```
sellout: This should be flagged with `TODO:` … but better to get an answer. It does seem like they should be shut down.

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: Already merged, but forgot to remove the pending (nice catch!). But that means they fail :-(

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: @wemrysi This function was re-implemented in your PR. Do you have any insight on where or not we are missing some server closing down logic here?

web/src/test/scala/slamdata/engine/api/fs.scala
sellout: But you can change them from `withServerRecordReloads` to `withServer` now (unless they change the port) … they still fail? If so, more work to do 😉

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1521,26 +1543,26 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
           errorFromBody(resp) must_== \/-("Can't add a mount point below the existing mount point at /foo/")
-          history must_== Nil
+          mock.actions must_== Nil
         }
       }
 
       "be 400 with missing file-name" in {
-        withServer(noBackends, config1) {
-          val req = root.POST
+        withServerRecordReloads(noBackends, config1) { (client, reloads) =>
```
sellout: This can be `withServer` now.

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1521,26 +1543,26 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
           errorFromBody(resp) must_== \/-("The 'X-File-Name' header must be specified")
-          history must_== Nil
+          reloads.toList must_== Nil
         }
       }
 
       "be 400 with invalid MongoDB path (no trailing slash)" in {
-        withServer(noBackends, config1) {
-          val req = root.POST
+        withServerRecordReloads(noBackends, config1) { (client, reloads) =>
```
sellout: Also this.

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1548,13 +1570,13 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
           errorFromBody(resp) must_== \/-("Not a directory path: /local")
-          history must_== Nil
+          reloads.toList must_== Nil
         }
       }
 
       "be 400 with invalid JSON" in {
-        withServer(noBackends, config1) {
-          val req = root.POST
+        withServerRecordReloads(noBackends, config1) { (client, reloads) =>
```
sellout: And this … etc. Just search for all the RecordReloads, and if they don’t try to change the port, you can use `withServer`.

web/src/test/scala/slamdata/engine/api/fs.scala
```diff
@@ -1735,30 +1759,27 @@ class ApiSpecs extends Specification with DisjunctionMatchers with PendingWithAc
...
-        history must_== Nil
       }
     }
 
     "restart on new port when PUT /port succeeds" in {
       val newPort = 8889
 
-      withServerExpectingRestart(noBackends, config1)({
-        val result1 = Http((root / "info") > code)
+      withServerExpectingRestart(noBackends, config1)({ client =>
```
sellout: Ok, hrmm, what is the distinction between `withServerExpectingRestart` and `withServerRecordReloads`? Seems like there should be only one.

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: @sellout Your comment made me realize, the `withServer` function should probably make sure no reloads happened. Right now, it just does not care whether the server was reloaded or not. Thoughts?

Yes, they are still failing, because the config has the port and I changed the test to expect None in order for it to compile. I think the reason with are still seeing "reloads" is because @wemrysi added a kinda of new "callback" kind of thing that records a config change even if it did not restart the server and right now the `withServerRecordReloads` is listening to that callback as well. Trying to understand exactly what the different fiels on `FileSystemApi` mean.

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: I guess the distinction is that a "reload" would be the config being changed, whether or not that triggers a restart. Are you suggesting we get ride of the check that makes sure that the config was changed?

jedesah: As to why the implementation of those two methods are so drastically different, I am not sure why that is, @wemrysi would surely have some insight on this.

web/src/test/scala/slamdata/engine/api/fs.scala
wemrysi: If there is any closing logic missing, it is in the impl. of servers, I don't think this call site needs to worry about it (so the try ... finally can go away). I'll check that `servers` is doing the right thing and submit a separate PR if it isn't.

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: @sellout We still need to record the Config changes in order for the tests to make sense so we can't use `withServer` for these tests. Unless I am missing something...

sellout: Yeah, I think you’re right … I overgeneralized from it not restarting the server anymore.

Also, re: `withServer`, yeah, it seems like checking that there were no reloads in there is a good idea.

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: I think we want to be explicit about the fact that the config did not change because the server recognized a malformed request. Of course, we don't want to test the world, but because these requests normally change the config, I think it is reasonable to do so and that would maintain the semantic that was there before the refactor.

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: Actually, @wemrysi changed my mind about `withServer`, the tests that care about that can record the config changes and make sure the list is empty, the tests using `withServer` don't care even though most of them should not see any reloads.

jedesah: And when I say reloads, I mean config changes, sorry :-)

web/src/test/scala/slamdata/engine/api/fs.scala
jedesah: By weaving `Task`, you mean `Task.run`? :P. I think it's that or I change `withServer` to return a `Task` and add a specs2 function that accepts a `Task` "test" and runs it. That's actually what I did in one of the codebases I maintained at my previous company. The advantage is that you can add a timeout at one place and never worry about a runaway `Task.run` hanging your test run. But, honestly, if that sounds appealing, I think it warrants it's own PR.

sellout: Well, I meant `withServer` should return Task, but then, yes, we’d have to call `.run` everywhere we use it.

Definitely having specs2 handling Task would be nice, and should be its own PR.

web/src/test/scala/slamdata/engine/api/fs.scala
wemrysi: `withServerRecordReloads` just needs a better name as it records config changes, not server restarts (as changing the config no longer equates to a server restart), `withServerExpectingRestart` provides a deterministic way of making assertions about the environment before and after a real http4s server restart.

mossprescott:
I have no comments other than about the javadoc, and would like to get this merged before I embark on renaming the world. @jr, if I don't hear from you in the next hour or so I'll probably just merge as is.

jedesah:
@mossprescott As per my previous comment, I feel the same way. Let me just make the obvious changes and let's merge as is.

PR #916, opened 2015-09-11
Add infrastructure required in order concisely write integration tests

jedesah:
- Add test classes of web project to it (integration) classpath
- Add convenience methods used by tests as conceivably sripts one day
- Create constants for MongoDB names

### DIFF
--- a/it/src/test/scala/quasar/Interactive.scala
+++ b/it/src/test/scala/quasar/Interactive.scala
@@ -1,0 +1,57 @@
+package quasar
+
+import Predef._
+import Backend.ProcessingTask
+import fs.Path
+import sql.{Query, SQLParser}
+
+import scalaz.Scalaz
+import scalaz.concurrent.Task
+import scalaz.stream.Process
+import Scalaz._
+
+/**
+ * Convenience methods that trade strong type safety for simplicity.
+ * Intended to be used in test code where a failure is unexpected and may be useful in
+ * a scripting or console context for playing around with the Quasar API.
+ */
+package object interactive {
+  /**
+   * Execute an SQL query on the backend and store the result in the provided `destinationPath`
+   * @param backend The Backend on which to run this query.
+   * @param query The SQL query to run
+   * @param destinationPath The path at which to store the result of the query.
+   * @return The result path where the result of this query was stored. For now, this should be equal to the
+   *         `destinationPath` specified as an argument, but that might eventually change. All failures are encoded as
+   *         a `Task` failure for convenience at the expense of safety and strong typing. The failure will be a
+   *         `scala.Exception` that contains a string message describing the nature of the failure.
+   */
+  def run(backend: Backend, query: String, destinationPath: String): Task[ResultPath] = {
+    val parser = new SQLParser()
+    for {
+      sql <- Task.fromDisjunction(parser.parse(Query(query)).leftMap(parseError => new scala.Exception("Parse error" + parseError.message)))
+      query = QueryRequest(sql, Some(Path(destinationPath)), Variables(Map()))
+      resultPath <- backend.run(query).fold(
+        err => Task.fail(new scala.Exception(err.message)),
+        a => a.run.flatMap(either => Task.fromDisjunction(either.leftMap(e => new scala.Exception("Evaluation error" + e.message))))
+      )._2
+    } yield resultPath
+  }
+
+  /**
+   * Execute an SQL query on the backend and return the result as a stream.
+   * @param backend The `Backend` on which to run this query.
+   * @param query The SQL query to evaluate
+   * @return A Stream of `Data` representing the result of the query.
+   */
+  def eval(backend: Backend, query: String): Process[ProcessingTask, Data] = {
+    val parser = new SQLParser()
+    parser.parse(Query(query)).fold(
+      err => Process.fail(new scala.Exception("Parser Error" + err.message)),
+      sql => backend.eval(QueryRequest(sql, None, Variables(Map()))).run._2.fold(
+        err => Process.fail(new scala.Exception("Compilation Error" + err.message)),
+        process => process
+      )
+    )
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -124,4 +124,4 @@ lazy val core = (project in file("core")) settings (oneJarSettings: _*) enablePl
 
 lazy val web = (project in file("web")) dependsOn (core % "test->test;compile->compile") settings (oneJarSettings: _*) enablePlugins(AutomateHeaderPlugin)
 
-lazy val it = (project in file("it")) dependsOn (core % "test->test;compile->compile", web) settings (standardSettings: _*) enablePlugins(AutomateHeaderPlugin)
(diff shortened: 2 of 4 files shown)


core/src/main/scala/slamdata/engine/backend.scala
```diff
@@ -83,6 +84,26 @@ sealed trait Backend { self =>
       })
   }
 
+  /**
+   * Convenience method for executing an SQL query on this backend. This method is usefull for certain tests
+   * and in a scripting context where a robust handling of all error scenarios is not required.
+   * @param query The SQL query to run
+   * @return The result path where the result of this query was stored. The return type is weakly typed, partly for
+   *         convenience, partly because I did not find a Tuple2T and was unwilling to create one. If anything bad
```
sellout: “Tuple2T” sounds like `WriterT`.

core/src/main/scala/slamdata/engine/backend.scala
```diff
@@ -83,6 +84,26 @@ sealed trait Backend { self =>
       })
   }
 
+  /**
+   * Convenience method for executing an SQL query on this backend. This method is usefull for certain tests
```
sellout: Minor misspelling of “useful”.

core/src/main/scala/slamdata/engine/backend.scala
```diff
@@ -83,6 +84,26 @@ sealed trait Backend { self =>
...
+   * and in a scripting context where a robust handling of all error scenarios is not required.
+   * @param query The SQL query to run
+   * @return The result path where the result of this query was stored. The return type is weakly typed, partly for
+   *         convenience, partly because I did not find a Tuple2T and was unwilling to create one. If anything bad
+   *         occurs, the `Task` will fail with an `Exception` containing a message describing the error.
+   */
+  def run(query: String, destinationPath: String): Task[ResultPath] = {
```
sellout: These functions should be in “test”, not “main”. We don’t `fail` outside of test code.

jdegoes: > These functions should be in “test”, not “main”. We don’t fail outside of test code.

:+1: Shouldn't clutter up production code with testing code. And yeah, no `fail`ing in prod code! :wink:

core/src/main/scala/slamdata/engine/backend.scala
```diff
@@ -83,6 +84,26 @@ sealed trait Backend { self =>
...
+    val parser = new SQLParser()
+    for {
+      sql <- Task.fromDisjunction(parser.parse(Query(query)).leftMap(parseError => new java.lang.Exception("Parse error" + parseError.message)))
```
jdegoes: Ditto.

core/src/main/scala/slamdata/engine/backend.scala
```diff
@@ -106,6 +127,23 @@ sealed trait Backend { self =>
...
   /**
+   * Convenience method for evaluating an SQL query on this backend. This method is useful for certain tests
+   * and in a scripting context where robust handling of all error scenarios is not required.
+   * @param query The SQL query to evaluate
+   * @return A Stream of `Data` representing the result of the query.
+   */
+  def eval(query: String): Process[ProcessingTask, Data] = {
+    val parser = new SQLParser()
+    parser.parse(Query(query)).fold(
+      err => Process.fail(new scala.Exception("Parser Error" + err.message)),
```
jdegoes: Ditto.

core/src/main/scala/slamdata/engine/backend.scala
jedesah: I don't remember why I said that so I just removed the comment, I think a vanilla Task is more convenient anyway. If it's going to be weakly typed, might as well be really weakly typed!

core/src/main/scala/slamdata/engine/backend.scala
jedesah: thx, fixed.

core/src/main/scala/slamdata/engine/backend.scala
jedesah: So the reason I consciously put these functions outside of the tests, which I don't consider to be de facto production code (but maybe I have the wrong definition of production code) is because I thought they might be rather useful for a developer playing around with the API on the console or something. Of course I added a lengthy comment explaining why one would want to avoid using this in production code. Now I don't have a strong objection to putting this in test code if that's what everyone wants (though It's not my preference), but I do think you guys bring up a good point that comments cannot be trusted. I was thinking a solution might be to take an implicit `unsafe` or `partial` argument which would require one to import `quasar.unsafe` in order to use these functions and of course importing `quasar.unsafe` is not allowed outside of test files in the Quasar project. What do you think?

sellout: The problem with `unsafe` is that it’s just convention. `fail` will still work inside a function that doesn’t require that implicit. That means we have to audit, and auditing falls apart with any complexity. We should be able to grep for (or, better, have a WartRemover check for) `fail` and see nothing come up.

If not in tests (and I agree, there are various helpful things that are arguably more general than testing), they should be in a separate subproject, so we can be super-strict with `core`.

I do think that such a subproject could be very useful, but probably better to put in tests for now, and pull those sorts of things into an `interactive` or whatever subproject later.

jedesah:
@mossprescott Sorry for the delay, it took me a while to realize this PR was broken, fixed it yesterday, feel like it's in a good state for merge now.

mossprescott:
This seem helpful. My only real concern is that without anybody running it it could rot. Can we get a couple of trivial tests to prove that it works in every build?

jedesah:
@mossprescott PR #917 depends on this PR and uses the functions. Possibly should not have separated the two.

it/src/test/scala/quasar/Interactive.scala
```diff
@@ -0,0 +1,57 @@
...
+   */
+  def run(backend: Backend, query: String, destinationPath: String): Task[ResultPath] = {
+    val parser = new SQLParser()
+    for {
+      sql <- Task.fromDisjunction(parser.parse(Query(query)).leftMap(parseError => new scala.Exception("Parse error" + parseError.message)))
+      query = QueryRequest(sql, Some(Path(destinationPath)), Variables(Map()))
+      resultPath <- backend.run(query).fold(
+        err => Task.fail(new scala.Exception(err.message)),
+        a => a.run.flatMap(either => Task.fromDisjunction(either.leftMap(e => new scala.Exception("Evaluation error" + e.message))))
+      )._2
```
mossprescott: In a typical debugging situation, the actual result isn't nearly as interesting as the log/plan that was generated. In my own hackish test harness, I have code to print the final plan (mongodb/javascript), or to print the last two (adding the Workflow version), or to print the WorkflowBuilder, or to print just everything (including the annotated SQL and LogicalPlan). 

I'm not sure what would be the most convenient way to capture it, but at least returning it is probably a good start.

mossprescott:
OK, with #917 coming I'm happy to merge this as is, unless you want to make any adjustments.

jedesah:
@mossprescott I think finding a way to provide the log is a good idea as well as moving to it -> main. But since it's not been used right now, maybe we can merge as is for now. Having moved it already, I don't want to move these methods until we come to a clear consensus. Your call.

PR #917, opened 2015-09-11
Ready/test for SD 892

jedesah:
Depends on PR #916


jedesah:
I forgot a `println` that I need to remove. But the tests are still failing on Travis but not on my machine! I assume it has something to do with the whole renaming the MongoDB instance, but I can't put my finger on it.

mossprescott:
@jdesah: Looks like you need to update the paths in `scripts/build`: https://travis-ci.org/slamdata/quasar/builds/81478243#L986.

jedesah:
@mossprescott Yes, I am looking at this currently. But I am investigating if I can just get ride of this loading step altogether now.

jedesah:
Hmm... it looks like Travis does not like loading the data collections from within the JVM process. Not enough memory?

jedesah:
Hurray! 11 days later... (don't tell John)

core/src/main/scala/quasar/fs/path.scala
```diff
@@ -48,6 +48,9 @@ final case class Path(dir: List[DirNode], file: Option[FileNode]) {
     case _=> this
   }
 
+  // TODO: Consider adding documentation to this method/renaming/reimplementing because
```
mossprescott: Should fix this or remove the TODO.

it/src/main/scala/quasar/utils.scala
```diff
@@ -0,0 +1,15 @@
...
+import Predef._
+
+package object utils {
+  /**
+   * Removes the extension from a file name
+   * @param fileName File name with an extension
+   * @return The name of the file without it's extension. Passing in "zips.data" would yield "zips".
+   *         Passing in "zips.latest.data" would yield "zips.latest".
+   */
+  def removeExtension(fileName: String):String = {
```
mossprescott: This duplicates code in `scala-pathy`. I'd rather see us porting code over to that than adding one-off functions like this. Short of that, maybe it could go in the existing (quasi-deprecated) `Path` object?

it/src/test/scala/quasar/Bug892.scala
```diff
@@ -0,0 +1,155 @@
...
+import scalaz._, Scalaz._
+
+import Predef._
+import config.{SDServerConfig, Config}
+import fs.Path
+import sql.{Query, SQLParser}
+import Errors._
+import specs2.DisjunctionMatchers
+
+class Bug892 extends BackendTest with NoTimeConversions with DisjunctionMatchers {
```
mossprescott: Can we give this test a meaningful name instead of referring to the bug by number? The bug will be lost to history, but this test will be around forever testing *something*.

it/src/test/scala/quasar/backendtest.scala
```diff
@@ -98,17 +99,63 @@ trait BackendTest extends Specification {
...
     *
+    * @param data A list of data sources to use in this test. The data will be loaded if
+    *             not already loaded and deleted at the end of the test. Currently, the data is actually
+    *             NOT deleted at the end of the test, but that is an undesired behavior that should hopefully
+    *             be fixed in the future.
+    *             Currently as well, the sources must have a non-empty [[String]] as a description
+    *             in order for the fixture to derive a [[Path]] at which to store the data. Eventually, this path
+    *             location could be randomized since we provide the paths as an argument for the body, so it should
+    *             not hardcode any [[String]]s in theory. Under such circumstances, a [[Source]] with a empty string
+    *             description would be accepted without issues.
```
mossprescott: Not really a big fan of documenting what *should* or *could* happen, so would prefer to keep this simple.

Also, `Source` isn't really our style. Why not something like `(data: (String, Process[Task, String]))` which would 1) be pure and 2) make it clear that the name is required.

it/src/test/scala/quasar/fs/filesystem.scala
```diff
@@ -45,7 +45,7 @@ class FileSystemSpecs extends BackendTest with DisjunctionMatchers with SkippedO
 
       "have zips" in {
         // This is the collection we use for all of our examples, so might as well make sure it's there.
-        fs.ls(prefix).map(_ must contain(FilesystemNode(Path("./zips"), Plain))).run.run must beRightDisjunction
+        fs.ls(prefix).map(_ must contain(FilesystemNode(ZipsDir.rebase(prefix).getOrElse(scala.sys.error("")), Plain))).run.run must beRightDisjunction
```
mossprescott: I'm still not clear on what this is testing, since `zips` is no longer being imported, right?

scripts/build
```diff
@@ -32,6 +32,7 @@ QUASAR_TEST_PATH_PREFIX="/${QUASAR_MONGODB_TESTDB}/"
 QUASAR_MONGODB_2_6="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_2_6}\"}}"
 QUASAR_MONGODB_3_0="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_0}\"}}"
 
+# TODO: Explain why we need to do this instead of just doing sbt clean oneJar at the end
```
mossprescott: Good idea but how about providing that explanation instead of the TODO?

mossprescott:
I see no major problems, but a couple of overall issues:

The new test is of a dubious nature, I think. If I understand correctly, it's attempting to reproduce a problem that was reported, but we were never able to reproduce. In other words, this test is not accompanying a fix. If it's proving that some important use case works, then that's worth something, but I can't really see what that case is by reading the test. All of which just suggests that the test may not be justifying its cost.

Second, we're mostly avoiding adding TODOs to the code base. If there's an important issue but we can't afford to hold up a merge, we'll usually write a new JIRA issue. In that case, any TODO (or "pending" test) should reference the new issue. Otherwise, even though it might be a good idea, we've said it's not worth cluttering the code with things that we aren't working on. Make sense?

scripts/build
jedesah: I would gladly do that if I knew the explanation :-) Do you know why? If not, then I will remove the TODO and put sbt clean oneJar at the end.

mossprescott: From reading the script, it's where we copy the web jar for testing; we used to make a copy of it before enabling coverage, so we could capture the un-instrumented build, but I don't see that happening here. In any case, `sbt clean` won't touch it.

jedesah: Check out my most recent PR where I addressed this issue so that I could remove the TODO on this PR without feeling like a bad boy scout.

jedesah:
Pinging @sellout to chime in on the value of the tests around SD-892. It is true that they never uncovered anything and I am totally okay with dropping them. At the same time they triggered many failures on Travis that exposed all kinds of problems with the tests in general.

it/src/test/scala/quasar/backendtest.scala
jedesah: Why not?

jedesah:
@mossprescott Thx for the review comments, I think they really helped improve the PR.

I haven't updated the documentation on `backendShould` because I wasn't clear on what was wrong with it.

PR #918, opened 2015-09-11
Change the project name to Quasar.

mossprescott:
SD-937 #done

Move everything from both `slamengine` and `slamdata.engine` packages to one package called `quasar`.

Update the README and CLA to refer to the project as Quasar.

Update references to external tools, including JIRA and Gitter.


mossprescott:
Need a volunteer to review: @sellout, @jr, @wemrysi? @jdegeos?

jdegoes:
Is this purely cosmetic?

mossprescott:
Only names are changed, if that's what you mean. I did not (intentionally) change any behavior.

jdegoes:
This looks fine. Obviously, though, the diff doesn't show what *might* have been missed, only what has changed. I need to do a release, would prefer to do it before this change — waiting for fix from @wemrysi. As soon as I get a new release out, I'm good to merge this.

mossprescott:
OK. As you can see, I did update any publicly-visible reference to slamengine, even things that won't be in place until after the repo is moved (like Gitter and Travis). What you can't see in the diff is that some less visible things are not changed:

Some of the build scripts still use filenames/paths involving `slamengine`. I figured updating that stuff could be a separate PR since it affects only us.

The default config location is unchanged: `.../slamengine-config.json`. I'm not sure if we want to worry about migrating existing installs, so I just left it as is, figuring we could decide later.

jdegoes:
I think it's worth migrating the config file (try to read quasar-config, if that fails, read slamengine-config; but always save quasar-config). We can update the slamengine script references separately.

mossprescott:
Alright I'll work on dealing with the config file.

mossprescott:
The config file is now read from the new location. Will need conflicts resolved after the other fix is merged, of course.

mossprescott:
@jdegoes: this is ready to go. Waiting on you for a release first.

PR #919, opened 2015-09-12
[SD-946] Fix regression where backend not updated when config changed

wemrysi:
This fixes the regression introduced in #910 where the backend isn't updated when the config changes. It also includes tests that should prevent this regression in the future.

### DIFF
--- a/core/src/main/scala/slamdata/engine/backend.scala
+++ b/core/src/main/scala/slamdata/engine/backend.scala
@@ -283,6 +283,7 @@ object Backend {
     }
   }
 
+  type ProcErrT[F[_], A] = EitherT[F, ProcessingError, A]
   type ProcessingTask[A] = ETask[ProcessingError, A]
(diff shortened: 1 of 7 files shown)


jdegoes:
@wemrysi Don't forget to nominate a reviewer or this will get stuck in the queue. :smile:

wemrysi:
@jdegoes Thanks for the reminder, assigned to @mossprescott mostly for continuity with #910. @jedesah you may want to take a look as well given your recent work on these tests.

mossprescott:
So I like this approach, which is to make the backend, along with the config, part of the state in FileSystemApi, while keeping them both pure values. Eventually, they will be unified and this approach will still work.

Aside from the formatting issues, I'm ready to merge this.

wemrysi:
Ok, cleaned up the formatting and removed the unused test helper.

jonsterling:
thank the ʟoʀᴅ this is merged! ;-)

Issue #283, opened 2014-09-08
planner: DISTINCT * does not work


mossprescott:
The problem is that every doc has a necessarily unique `_id` field, so DISTINCT will never be able to remove anything. Without knowing the shape, we cannot specify the particular fields to match on.

Furthermore: the aggregation pipeline cannot exclude the _id field, so projecting to a temp name without _id will not work. There may be a workaround involving MapReduce.

See #211.

sellout:
Fixed by #630.

PR #920, opened 2015-09-14
`distinct` without `Workflow`.

sellout:
SD-949 #done

This reimplements `distinctBy` (and adds `distinct`) without converting
to `Workflow`. It also fixes some test cases that I _think_ should
already have been working, but were overlooked when #283 was fixed.

Most significantly, the required replacing `DocVar` with `Base` for rewriting.

`Base` is a richer abstraction that allows us to distinguish between an
entire document (IE, DocVar.ROOT) and a subset of a document (which was
not distinguished from ROOT previously).

This required various other fixes to work properly:

* add a NEL `apply` to `BsonField`,
* render BSON as JS rather than `repr`,
* move all of the `ExprBuilder` coalescing from `expr` to `normalize`,
* extract `findKeys` from `distinct`,
* move a couple `merge` cases that we preventing splices from unifying
  properly,
* use `apoM` instead of `topDownTransformM` in `normalize` (this prevents
  double-processing (and thus failure) in the case when a substitution
  contains a projection itself),
* add `anaM` and `apoM` recursion schemes,
* convert `Schema` elements from `String` to `BsonField.Name`,
* eliminate an unused `fold1Builders` call, and
* propagate the correct `Schema` from `GroupBuilder`s.


sellout:
@jedesah I can’t seem to assign this PR to you – I think you’re not a member of the SlamData org.

Anyway, it’s yours, but I’m fairly certain you’ll want to have @mossprescott look at it with you.

jdegoes:
Under MRA, we'll have:

```haskell
a `distinct by` b = arbitrary a `group by` b

distinct a = arbitrary a `group by` a
```

Not that this helps you now. :laughing:

sellout:
This actually brings the mongo implementation to what we need to support that MRA version – you can see this line in the new `distinctBy` is exactly what you say: https://github.com/slamdata/slamengine/pull/920/files#diff-3c2fe70b67eb3d507c03b856f657becaR1578

The one difference is that our current `distinctBy` adopts SQL’s order-preserving semantics, so we propagate sort keys as necessary.

sellout:
Oh, and this PR means we now properly handle excluding unwanted fields (IE, `_id` and synthetic projections) from all `groupBy`s, which is necessary for the MRA version.

core/src/main/scala/slamdata/engine/physical/mongodb/bson.scala
```diff
@@ -292,6 +292,12 @@ object BsonField {
   sealed trait Root
   final case object Root extends Root
 
+  def apply(v: NonEmptyList[BsonField.Leaf]): BsonField = v match {
+    case NonEmptyList(head) => head
+    case NonEmptyList(head, tail @ _*) =>
+      Path(NonEmptyList.nel(head, tail.toList))
```
mossprescott: I think this can be simplified to just `case _ => Path(v)`.

core/src/main/scala/slamdata/engine/physical/mongodb/bson.scala
```diff
@@ -292,6 +292,12 @@ object BsonField {
...
+  }
+
   def apply(v: List[BsonField.Leaf]): Option[BsonField] = v match {
     case Nil => None
     case head :: Nil => Some(head)
```
mossprescott: Meanwhile, these two cases become `case head :: tail => apply(NonEmptyList.nel(head, tail))`.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -214,7 +221,7 @@ object WorkflowBuilder {
   implicit def WorkflowBuilderEqualF = new EqualF[WorkflowBuilderF] {
     def equal[A](v1: WorkflowBuilderF[A], v2: WorkflowBuilderF[A])(implicit A: Equal[A]) = (v1, v2) match {
       case (CollectionBuilderF(g1, b1, s1), CollectionBuilderF(g2, b2, s2)) =>
-        g1 == g2 && b1 == b2 && s1 === s2
+        g1 == g2 && b1 == b2 && s1 == s2
```
mossprescott: This is pretty confusing, given that `s1 === s2` appears on the next line. Of course, `s1` is referring to totally different types, but still. How is one supposed to know when it's necessary to do `===`?

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -299,40 +306,88 @@ object WorkflowBuilder {
...
+            }).map(f))
+
+        src.unFix match {
+          case ExprBuilderF(wb0, -\/(js1)) =>
+            exprToJs(outerExpr).map(js => ExprBuilder(wb0, -\/(js1 >>> js))).toOption
+          case ExprBuilderF(src0, contents) =>
+            inln(Expr(contents))(expr => ExprBuilder(src0, \/-(expr)))
+
+          case DocBuilderF(src, innerShape) =>
+            // inln(Doc(innerShape))(expr => ExprBuilder(src0, \/-(expr)))
```
mossprescott: dead code

core/src/main/scala/slamdata/engine/physical/mongodb/bson.scala
wemrysi: Or even `v.toNel.map(apply)`.

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -1523,15 +1523,39 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
...
+                Call(ident("remove"),
+                  List(ident("x"), jscore.Literal(Js.Str("_id")))))))),
+            ListMap()),
+          $group(
+            grouped("__tmp2" -> $first($field("__tmp1"))),
+            \/-($field("__tmp1"))),
+          $project(
+            reshape("value" -> $field("__tmp2")),
+            ExcludeId)))
+    }
```
mossprescott: This *looks* correct to me, but is it fixed here or was it already working and just needed the correct assertion? Do we have an integration test for this query?

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
```diff
@@ -1523,15 +1523,39 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
...
+                Call(ident("remove"),
+                  List(ident("x"), jscore.Literal(Js.Str("_id")))))))),
+            ListMap()),
+          $group(
+            grouped(),
+            \/-($field("__tmp4"))),
+          $group(
+            grouped("0" -> $sum($literal(Bson.Int32(1)))),
+            \/-($literal(Bson.Null)))))
+    }
```
mossprescott: Ditto.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -1830,12 +1793,13 @@ object WorkflowBuilder {
       case (_, ArrayBuilderF(_, _)) => delegate
 
       case _ =>
+        // scala.sys.error("failed merge")
```
mossprescott: debug code

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -1681,13 +1656,40 @@ object WorkflowBuilder {
...
+        spb2 @ ShapePreservingBuilderF(src2, inputs2, _))
+          if inputs1 === inputs2 && ShapePreservingBuilder.dummyOp(spb1) == ShapePreservingBuilder.dummyOp(spb2) =>
+        merge(src1, src2).map { case (lbase, rbase, wb) =>
+          (lbase, rbase, ShapePreservingBuilder(wb, inputs1, op1))
+        }
+      case (ShapePreservingBuilderF(src, inputs, op), _) =>
+        merge(src, right).map { case (lbase, rbase, wb) =>
+          (lbase, rbase, ShapePreservingBuilder(wb, inputs, op))
+        }
+      case (_, ShapePreservingBuilderF(src, inputs, op)) => delegate
```
mossprescott: If these cases *need* to be higher than they were, maybe there should be a comment explaining why. At least until I replace all this with my new approach which isn't sensitive to ordering :)

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -1481,196 +1547,105 @@ object WorkflowBuilder {
...
-        case _                                  => None
-      }
+  @tailrec
+  def findKeys(wb: WorkflowBuilder): Option[List[BsonField.Name]] = {
+    wb.unFix match {
+      case CollectionBuilderF(_, _, s2)       => s2
+      case DocBuilderF(_, shape)              => shape.keys.toList.some
+      case GroupBuilderF(_, _, Doc(obj), _)   => obj.keys.toList.some
+      case ShapePreservingBuilderF(src, _, _) => findKeys(src)
+      case ExprBuilderF(_, _)                 => Nil.some
```
mossprescott: I struggled to explain this result to @jedesah. An ExprBuilder *can* produce a document-shaped result; we just don't know for sure what it is. Doesn't that make this `None` (meaning you can simply omit the case?

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -1481,196 +1547,105 @@ object WorkflowBuilder {
...
-      }
-      val op = keyPairs.toNel.map { keyPairs =>
-        group.copy(graph = chain(group.graph, $sort(keyPairs)))
-      }.getOrElse(group)
-
-      Fix[WorkflowBuilderF](op)
+      case _ => distincting(src).point[M]
     }
+
+    findSort(src)
```
mossprescott: I have nothing clever to say about this except that I'm very happy all of that code got deleted. :)

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -1481,196 +1547,105 @@ object WorkflowBuilder {
...
-
-    for {
-      t1 <- foldBuilders(src, keys)
-      (merged, value, fields) = t1
+  def distinct(src: WorkflowBuilder): M[WorkflowBuilder] = {
 
-      b2 <- toCollectionBuilder(merged)
-      CollectionBuilderF(graph, base, struct) = b2
+    findKeys(src).fold(
+      // NB: only need to delete from the key, but this cleans up the plan
```
mossprescott: If I understand correctly, you mean the result would be correct if `_id` was deleted from the key but preserved in the value? But it makes the plan more complex? And presumably the simpler plan is more efficient, too, and will be even better when we can project the result out of the $group op's `_id`. So that sounds good to me.

jdegoes:
If we don't already have adequate distinct AND distinct by regression tests that verify results exactly (not just `contains`), this would be a good PR to add them.

core/src/main/scala/slamdata/engine/physical/mongodb/bson.scala
sellout: Ah, of course 😑

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: It’s (unfortunately) never necessary. We just try to prefer `===` (or `≗`) whenever it’s available. Need to get something into WartRemover that prevents us from using `==` and forces us to derive `Equal` instances for everything.

So, here I should really derive `Equal` for `BsonField` and keep the `===`.

core/src/test/scala/slamdata/engine/physical/mongodb/planner.scala
sellout: This was already working, but the missing assertion meant it didn’t get triggered when it was fixed.

I can add an integration test if there’s not one.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: Looking forward to that. But, yes, basically these need to appear above the `(ExprBuilder, _)` cases, because SPBs should be closer to the root than most other builders.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: This may deserve a custom type. You wanted this pulled out from where it was in `distinct` for some dark purpose, and this meant changing the shape a bit from what I originally had.

So, `None` means there’s we couldn’t deduce the shape, so we do the slow thing of just deleting `_id` in a `$simpleMap`. A non-empty `List` means we found all the fields we need for the grouping. And `Nil` … well that means the key is just a single expression, but whereas a one-element `List` holds its single expression in some field, `Nil` means that the single expression is at `$$ROOT`.

By the time we get to `Workflow`, that expression must be in some field, but in Builders, we often have things at $$ROOT until we convert the source to `Workflow` and discover the field that our expression is actually relative to.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: Yep, exactly. initially I implemented it as `distinctBy(src, List(del))`, but this required us to pass `$$ROOT` through the `$simpleMap` along with the cleaned-up version. The implementation here actually generates the same `Workflow` as the old version already was.

sellout:
@jdegoes Good point. I’ll check our distinct tests. I’m sure some disappeared when we fixed `group by`, so I wouldn’t be surprised if we needed some new ones.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
wemrysi: +1 for favoring `Equal[A]` over `Object#equals`.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: @mossprescott What if I have `findKeys` return `Option[Base]`? (where `Base` is the type I added elsewhere in this PR to fix `mergeContents`)

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: And … `Equal[A]` found a place in WFB where we were comparing a `BsonField` to a `String`. Thanks, heterogeneous equality!

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
mossprescott: That gives you *four* possible results, right? `None` (unknown), `Root` (for `Expr`), `List[Field]` (for docs), and `Field` (meaning what?)

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
mossprescott: I'm happy about that but iI don't think you really need the comment (except as an expression of our joy and pride :smile:)

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: Well, it would never generate `Field`, but at least where I _call_ it, Field is a valid result, meaning a single key whose expression is at that field … `Root` and `Field` are basically `DocVar` decomposed.

mossprescott: That sounds better to me than `Option[List[...]`, given that all of the values have a well-defined meaning.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -800,36 +828,74 @@ object WorkflowBuilder {
...
 
+  sealed trait Base {
+    def \ (that: Base): Base = this match {
+      case Root()      => that
+      case Field(name) => that match {
+        case Root()       => this
+        case Field(name2) => Field(name \ name2)
+        case Subset(fields) => this // NB: can we do better?
+      }
+      case Subset(_)   => this match {
```
mossprescott: This can't be right, can it? `this match { ... this match { ...`

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -800,36 +828,74 @@ object WorkflowBuilder {
...
+        case Root()       => this
+        case Field(name2) => that // NB: can we do better?
+        case Subset(fields) => that // NB: can we do better?
```
mossprescott: These aren't really `NB`s, are they? More like `TODO`s, I think. So should get cleaned up?

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
sellout: No … no it can not.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
```diff
@@ -800,36 +828,74 @@ object WorkflowBuilder {
     case _ => ArrayBuilder(wb, List(\/-($$ROOT)))
   }
 
+  sealed trait Base {
```
mossprescott: Because `Base` isn't very evocative, and after pairing with @jedesah and trying to help him follow this stuff, I think this trait and the case classes could all use descriptive comments.

core/src/main/scala/slamdata/engine/physical/mongodb/workflowbuilder.scala
mossprescott: Wouldn't it be clearer to use the old `(this, that) match { ... }`?

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -1484,8 +1484,8 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
           $group(
             grouped("__tmp0" -> $first($$ROOT)),
             -\/(reshape(
-              "city"  -> $field("city"),
-              "state" -> $field("state")))),
+              "0"  -> $field("city"),
+              "1" -> $field("state")))),
```
jedesah: Why do the field names disappear?

jedesah:
Looks good to me at this point. `anaM` and `apoM` lack tests, but I will address that in a separate PR, no sense holding up this one for something like that. Will merge unless someone else has any comments soon.

core/src/test/scala/quasar/physical/mongodb/planner.scala
jdegoes: Also the following project seems to refer to the old field names not `0` and `1`.

sellout: All that’s important here is the set of values that we group by. However, MongoDB does require that that set is presented as either a single expression or a document. Using the names of the fields as the keys was done by [this](https://github.com/slamdata/quasar/pull/920/files#diff-1d2c982f39cd60cd821f0028a32ad1cbL1577), but that isn’t general enough to handle all cases, so we no longer do it at all.

@jdegoes This just affects the fields we group by. The shape of the grouped content is unchanged.

jdegoes: :+1:

PR #921, opened 2015-09-14
rename example config file and update the script that refers to it

mossprescott:


wemrysi:
Not really related to the rename, but should we change the contents of the example config to not point at the mongolab instance?

sellout:
I think it’s fine if it points at MongoLab … at least it means it’s a real working config, and we don’t want to point at our actual test/demo instance.

mossprescott:
OK by me. Not sure it actually matters to the test, which just starts the server and would probably accept any well-formed config. If we want to get rid of mongolab at some point we could switch it to the new shared environment and just point it at slamengine-test-01 over there, which is not used by the demo or CI.

wemrysi:
Ah, yeah, the test. I was thinking more from the POV of users who make use of it as a default, not sure what will happen to that instance as Quasar's popularity increases ;)

mossprescott:
The nice thing about that instance is that when it fills up it just gives an error and doesn't start charging us more money, unlike the other environment.

wemrysi:
True, though then the "real working config" really isn't...

wemrysi:
It looks like this built successfully but I think the status widget is still pointing at "slamengine" somehow, so it's wedged. If others agree then I'll merge.

scripts/testJar
```diff
@@ -7,42 +7,42 @@ source "$(dirname $0)/constants"
...
-SLAMDATA_LOG_FILE="$TEMP_DIR/slamout.log"
+QUASAR_SUCCESS_OUTPUT="Server started listening on port"
+QUASAR_LOG_FILE="$TEMP_DIR/slamout.log"
 
-echo "Launching SlamData JAR..."
+echo "Launching Quasar JAR..."
 
-java -jar "$SLAM_WEB_JAR_PATH" &> "$SLAMDATA_LOG_FILE" &
-SLAM_PID=$!
+java -jar "$SLAM_WEB_JAR_PATH" &> "$QUASAR_LOG_FILE" &
```
jdegoes: This is a reference to `SLAM_WEB_JAR_PATH`. Which should really be `QUASAR_WEB_JAR_PATH`, updated here and wherever it originates (plus all references). Aside from copyright, there should be no references to `Slam` in the project.

wemrysi:
@jdegoes yeah, it looks like we missed most of `scripts/` do we want to expand this PR to include the rest of that or file a separate issue? We'll need to rename environment variables, the default config directory on POSIX (`.config/SlamData`) and even one of the scripts themselves (`startSlamEngine`).

jdegoes:
@wemrysi I'd prefer it in this PR if you dont' mind.

mossprescott:
I have another round of changes that I could add to this PR, including those scripts. I wanted to get all the builds building first and We seem to be good there as of last night.

mossprescott:
Latest commit updates all the scripts and also updates the program name in the REPL's "usage" message.

The _only_ references to slam-anything that should be left are copyright strings, a couple of places where we refer to SlamData, the overall product (including the README and the config directory), and the fallback config location, and a couple of mentions of "SlamSQL" in the README.

jdegoes:
Can you change SlamSQL -> SQL<sup>2</sup>?

wemrysi:
One last place that I found: we append "/slamdata" to the content path that may be specified as a command-line option to the web jar in addition to mounting the `staticFileService` under the `/slamdata` endpoint.

The idea of including a way to serve these files seems rather SlamData, the company, specific (I wouldn't expect an API server like quasar to provide such a feature). However, keeping the feature but finding an unbranded name seems ok for the short-term.

jdegoes:
> One last place that I found: we append "/slamdata" to the content path that may be specified as a command-line option to the web jar in addition to mounting the staticFileService under the /slamdata endpoint.

I think it could be generally useful to serve some static files, if nothing else an `index.html` that says, 'Quasar is running' or something. But yeah, the `slamdata` part needs to go. Possibly you could specify the subpath as a command-line option. 

Any preference for sticking that in here versus tracking it in a separate issue / PR?

wemrysi:
> I think it could be generally useful to serve some static files

True, I just think it is outside the scope of quasar.

Anyhow, I think a separate issue for this bit is a good idea as it will likely involve semantic changes, that will allow us to keep this to just renames.

wemrysi:
On that note, :+1:  from me.

mossprescott:
I'm adding a sub-task.

PR #924, opened 2015-09-15
Add documentation and rename stuff

jedesah:


core/src/main/scala/quasar/physical/mongodb/expression/package.scala
```diff
@@ -27,9 +27,12 @@ import quasar.jscore, jscore.{JsCore, JsFn}
 import scalaz._, Scalaz._
 
 package object expression {
-  type Expression = Fix[ExprOp]
+  /**
+   * Expression that can evaluated in the aggregation pipeline
+   */
+  type PipelineExpression = Fix[ExprOp]
```
sellout: Nope, this is just `Expression` – it’s Mongo’s terminology, not ours. Also, it’s not limited to the aggregation pipeline (although that’s the only place we currently support using them).

jedesah: Ok, then can we call it `MongoExpression` or something? I believe there is a semantic problem with the sentence an Expression is either an Expression or a JSFunction.

core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -36,10 +36,21 @@ object WorkflowBuilder {
...
   import IdHandling._
 
+  /**
+   * A partial description of a query that can be run on an instance of MongoDB
+   */
   type WorkflowBuilder = Fix[WorkflowBuilderF]
+  /**
+   * If we know what the shape is, represents the list of Fields.
+   */
   type Schema = Option[List[String]]
```
wemrysi: What does a value of `Some(Nil)` signify? That we know we have an empty object?

core/src/main/scala/quasar/physical/mongodb/expression/package.scala
jedesah: I don't like `MongoExpression` because we are already in Mongo land, no sense repeating, but I like it more then an Expression is an Expression or...

core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
jedesah: Presumably. @mossprescott or @sellout might have a better idea if `NonEmptyList` would be more appropriate here.

core/src/main/scala/quasar/physical/mongodb/expression/package.scala
sellout: The name that needs to be improved there is `Expr` (assuming you’re referring to one of the WorkflowBuilder `Expr`s).

Another option is that hopefully Expression (and Workflow//Accumulator/etc.) will eventually be in a separate repo, which is the context in which the name `Expression` is correct. WorkflowBuilder, however, is Quasar-specific, so in WorkflowBuilder, you could `import Expression => MongoExpression` or something.

core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
sellout: @wemrysi Yeah – it might be useful to eliminate that. An empty set of projections isn’t even valid for Mongo, so restricting this to NonEmptyList sounds reasonable. I don’t _think_ we create empty projections anywhere, so it shouldn’t be complicated.

core/src/main/scala/quasar/physical/mongodb/expression/package.scala
mossprescott: It's already in a different package, so I think importing `Expression => MongoExpression` is currently an option.

But I tend to agree that `Expr` is the problematic name, especially since there is also `Contents.Expr`. Doesn't `Expr` ultimately become `SlamScript` (or whatever we're calling it now)?

sellout: Yeah, `Expr` will eventually be a SlamScript ADT that becomes either Expression or JsFn further down the line. And we’ll no longer convert Expression->JsFn (because maintaining the semantics there is terrible, among other reasons).

And I agree that the aliasing is currently an option – was just trying to draw the line between our high-level Mongo driver and Quasar’s interface to that driver.

PR #925, opened 2015-09-16
Remove `Backend#defaultPath`

wemrysi:
This removes `Backend#defaultPath` which had dubious semantics and was only used in tests. Test usage was replaced with the ability to explicitly specify the prefix path for test backends via an environment variable, defaulting to `/quasar-test/` when unspecified.

### DIFF
--- a/it/src/test/scala/quasar/backendtest.scala
+++ b/it/src/test/scala/quasar/backendtest.scala
@@ -16,49 +16,78 @@ import quasar.fs._
 
 object TestConfig {
 
+  /**
+   * The path prefix under which test data may be found as well as where tests
+   * can write test data/output.
+   */
+  val DefaultTestPathPrefix = Path("/quasar-test/")
+
+  /**
+   * The environment variable used to externally specify the test path prefix.
+   *
+   * NB: The same path prefix is used for all backends under test.
+   */
+  val TestPathPrefixEnvName = "QUASAR_TEST_PATH_PREFIX"
+
   lazy val backendNames: List[String] = List("mongodb_2_6", "mongodb_3_0")
 
   private def fail[A](msg: String): Task[A] = Task.fail(new RuntimeException(msg))
 
-  def envName(backend: String): String = "SLAMDATA_" + backend.toUpperCase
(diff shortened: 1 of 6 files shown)


it/src/test/scala/quasar/backendtest.scala
```diff
@@ -68,19 +97,23 @@ trait BackendTest extends Specification {
...
-    }).map(_ => ()).run
-  }
+  def backendShould(runTests: (Path, Backend, String) => Unit): Unit =
+    (for {
+      dir   <- TestConfig.testDataPathPrefix
+      bknds <- AllBackends
+      _     <- bknds.traverse_ { case (n, b) =>
+                 Task.delay(n should runTests(dir, b, n)).void
+               }
+    } yield ()).run
```
sellout: Could you use applicative style here? It’s one of the WartRemover checks we have disabled because we have some other cases to clean up.

it/src/test/scala/quasar/backendtest.scala
```diff
@@ -16,49 +16,78 @@ import quasar.fs._
 
 object TestConfig {
 
+  /**
+   * The path prefix under which test data may be found as well as where tests
+   * can write test data/output.
+   */
```
sellout: I only mention it because I asked you to change something else, but … can we use the Scala Guide style for comments? http://docs.scala-lang.org/style/scaladoc.html

it/src/test/scala/quasar/backendtest.scala
wemrysi: I can, though `Monad` is required to `join` the final value, but it will probably still look nicer ;)

sellout: Ah right … dat join. Well, up to you, then.

it/src/test/scala/quasar/backendtest.scala
wemrysi: Sure thing.

sellout:
Ah crap – conflicts.

wemrysi:
Yeah, just resolved them, should be good to go assuming travis agrees.

PR #928, opened 2015-09-17
Make importTestData script more flexible

jedesah:
MongoDB can be used without credentials, this makes this script take advantage of that.

Authored with @wemrysi


mossprescott:
I'm going to let Travis finish before merging, but this seems helpful.

Note: we really need to replace this script with one that runs the backend and handles our JSON properly. `mongoImport` works for simple data, but fails for any of our interesting datasets, such as `days.data`.

mossprescott:
Coveralls. <sigh>

PR #929, opened 2015-09-18
make static file serving optional and parameterize the location

mossprescott:
SD-959 #done

Adds a new parameter, `-L`, when launching the API server, which specifies the URL path static files should be served from. This will be used to configure the path to the front-end's files, without needing any knowledge hard-coded in the server.

Adds a new page at `/welcome`, to which `/` is redirected if no static files are being served. This will be a default landing page for users starting the bare server, and would be a good place to host some simple tips about how to get started, or links to documentation, etc. For now, it's just a tiny page showing the version.

Fixed an infinite HTTP redirect loop when static files aren't present.


web/src/main/scala/quasar/api/fs.scala
```diff
@@ -617,6 +620,21 @@ final case class FileSystemApi(
         .run.flatMap(_.as(Ok("")) valueOr handlePathError)
   }
 
+  def welcomeService(config: Task[Config]) = {
```
wemrysi: I think the `config` argument can be elided as it isn't used anywhere in the body of the function.

web/src/main/scala/quasar/api/fs.scala
```diff
@@ -136,7 +137,9 @@ object ResponseFormat {
  * @param configChanged Expected to persist a Config when called. Called whenever the Config changes.
  */
 final case class FileSystemApi(
-  contentPath: String, initialConfig: Config,
+  contentLoc: Option[String],
```
wemrysi: What do you think about a new datatype for the static content settings, say `case class StaticContent(path: String, loc: String)` and change the argument to accept an `Option[StaticContent]`? This keeps the two paths related and also pushes the choice of default location (`/files` in this case) out of the API itself and into whatever is creating it or, better yet, reading the configuration.

web/src/main/scala/quasar/api/fs.scala
mossprescott: Yes, I didn't end up using it and I'll remove it.

web/src/main/scala/quasar/api/fs.scala
mossprescott: Actually I think it makes sense to go further and pull this stuff (`welcome`, the redirect, and any static files) entirely out of `FileSystemApi`.

wemrysi: :+1:

jedesah: +1

mossprescott:
@wemrysi: updated. That ugly logic with the two Options ended up in `servers`, where it can do less damage.

web/src/main/scala/quasar/api/server.scala
```diff
@@ -93,18 +93,30 @@ object Server {
...
-  def servers(contentPath: String, idleTimeout: Duration, tester: BackendConfig => EnvTask[Unit],
+  def servers(contentLoc: Option[String], contentPath: Option[String],
+              idleTimeout: Duration, tester: BackendConfig => EnvTask[Unit],
               mounter: Config => EnvTask[Backend], configWriter: Config => Task[Unit])
              : (Process[Task, (Int, Http4sServer)], Option[(Int, Config)] => Task[Unit]) = {
 
     val configQ = async.boundedQueue[Option[(Int, Config)]](2)(Strategy.DefaultStrategy)
     val reload = (cfg: Config) => configQ.enqueueOne(Some((cfg.server.port, cfg)))
 
+    val fileSvcs = contentPath match {
```
wemrysi: I think moving this here is nice, though I still think a single datatype for the static config would be cleaner, for example it isn't obvious that `contentLoc` is ignored for the following combination of arguments `contentLoc = Some("/foo"), contentPath = None`, but with a single datatype this scenario is not even representable.

web/src/main/scala/quasar/api/server.scala
```diff
@@ -93,18 +93,30 @@ object Server {
...
+      case None =>
+        ListMap(
+          "/" -> redirectService("/welcome"))
+      case Some(fsPath) =>
+        val urlPath = contentLoc.getOrElse("/files")
```
wemrysi: I think moving the default content location value (`/files`) into the `Options` class below would make things clearer by moving the value closer to the option it is the default for.

mossprescott:
I put the location and path together as suggested, and broke out redirection as a separate param. This ends up making the "/files" path never arise, because that was only needed in the case we don't use (static content, but no location provided).

I also made the method accept a list of StaticContent, because it seemed reasonable and cost nothing.

wemrysi:
I think we should either keep the default for `--content-location` or error when `--content-path` specified without a corresponding `--content-location`. As-is, if one only specifies the `--content-path` it is silently ignored.

Also, do we still need the `jarPath` as a default for `--content-path`?

sellout:
Ooooh, getting rid of the terrible `jarPath` would be great. But don’t we need that to discover where the content lives on different platforms, in our distributed app?

mossprescott:
`jarPath` is used by the installer, where we have to get information about the install location from somewhere and, sadly, that means content-path is always present, so we *need* it to be ignored if content-location isn't specified (so that we'll redirect to /welcome).

The other use case is the demo, where we need to explicitly specify a path on the file system, since we haven't assembled the app in the same way as the installer does.

This is pretty *ad hoc* for sure, but now that it's been pushed all the way out to `main` I was hoping you'd give it a pass.

mossprescott:
@sellout: I think we do, at least that's what you decided when you wrote that code :-)

wemrysi:
Do you mean the installers assume the default content path is `/the/path/to/the/dir/the/jar/is/in/slamdata`? If so, is it broken now anyways since we no longer append `/slamdata` to the content path?

Anyhow, can't we still error if `--content-path` is specified without a corresponding `--content-location` on the command line? One could still specify `--content-location` by itself, which should support the installer use-case.

sellout:
Sounds like the launcher in the installer project needs to specify `--content-location slamdata`? Please don’t leave that broken for me come Monday 😉

mossprescott:
@sellout: There's a PR adding that parameter to the launcher. Of course, I don't know if it works.

@estewei: Hmm. `jarPath` puts in `/docroot`, but looking the last bundle I installed, the files are actually under `docroot/slamdata`. So you may be right. Arg.

sellout:
@mossprescott Yeah, `jarPath` has “docroot”, but then the staticFileService append[s/ed] “slamdata” … which I assume that parameter is now doing.

mossprescott:
@sellout: no, the paramater is *not* doing it, and can't because reasons.

sellout:
Ok, then the installer can probably put the static content directly in docroot. The idea was that there might be _multiple_ Web apps in an install (EG, OEM), and so we needed to put them in directories.

mossprescott:
@sellout: with this latest commit, and the one I just added in another installer PR (slamdata/installers#5), the `docroot/slamdata` path is supplied on the command line, so no need to change the way it's deployed.

@wemrysi: this adds the new param we discussed, cleans up the logic and pulls it out to a separate function with unit tests. It does *not* use pathy, and I'm inclined to let that go for the time being.

web/src/main/scala/quasar/api/server.scala
```diff
@@ -93,18 +95,22 @@ object Server {
    * Pass [[None]] to the returned function to shutdown any running server and
    * prevent new ones from being started.
    */
-  def servers(contentPath: String, idleTimeout: Duration, tester: BackendConfig => EnvTask[Unit],
+  def servers(staticContent: List[StaticContent], redirect: Option[String],
```
wemrysi: How about `defaultLocation` or `rootLocation` in place of `redirect`, i.e. something that describes what the parameter is rather than how it is being achieved.

This also doesn't need to be an `Option` as we're handling the default value in `main`.

web/src/main/scala/quasar/api/server.scala
```diff
@@ -147,19 +153,40 @@ object Server {
...
-    opt[String]('C', "content-path") action { (x, c) => c.copy(contentPath = x) } text("path where static content lives")
+    opt[String]('L', "content-location") action { (x, c) => c.copy(contentLoc = Some(x)) } text("location where static content is hosted")
+    opt[String]('C', "content-path") action { (x, c) => c.copy(contentPath = Some(x)) } text("path where static content lives")
+    opt[Unit]('r', "content-path-relative") action { (_, c) => c.copy(contentPathRelative = true) } text("specifies that the content-path is relative to the install directory (not the current dir)")
     opt[Unit]('o', "open-client") action { (_, c) => c.copy(openClient = true) } text("opens a browser window to the client on startup")
     opt[Int]('p', "port") action { (x, c) => c.copy(port = Some(x)) } text("the port to run Quasar on")
     help("help") text("prints this usage text")
   }
 
+  def interpretPaths(options: Options): EnvTask[List[StaticContent]] = {
```
wemrysi: How about returning `Option` here instead of `List` to more accurately represent that this will return at most one `StaticContent`?

web/src/main/scala/quasar/api/server.scala
```diff
@@ -147,19 +153,40 @@ object Server {
...
+    val defaultLoc = "/files"
+
+    def path(p: String): EnvTask[String] =
+      liftE(
+        if (options.contentPathRelative) jarPath.map(_ + p)
+        else Task.now(p))
+
+    (options.contentLoc, options.contentPath) match {
+      case (None, None) => liftE(Task.now(Nil))
```
wemrysi: I think `point[EnvTask]` would be clearer here.

web/src/main/scala/quasar/api/server.scala
mossprescott: It's an Option for the benefit of any caller who doesn't need to redirect root, including specifically the integration test that calls it with None.

As for the name, I guess I see this as "the path to which root is redirected", so I'm open to any way of abbreviating that as it seems a bit long ;-) I don't really think there's a more accurate way to describe it than that.

mossprescott:
@wemrysi: I've got fixes for those two small issues queued up but wondering if there's anything else before I push.

web/src/main/scala/quasar/api/server.scala
wemrysi: Hm, I think this would be clearer, maybe at the loss of some theoretical flexibility, if the existing `staticContent` and `redirect` options were consolidated into a single `staticContent: Option[StaticContent]` option. If it is `Some` then static content serving is enabled and root is redirected to the location specified in `StaticContent` otherwise redirected to the welcome page.

As-is when would one define `redirect` and not define `staticContent`?

mossprescott: You could redirect `/` to `http://google.com/?q=slamdata`, or really anything else at all.

And note that the first param is still `List[StaticContent]`, because the caller might (theoretically) want to serve more than one directory of static files, in which case it would not be clear which one should be redirected to.

wemrysi: > You could redirect / to http://google.com/?q=slamdata, or really anything else at all.

True, but is supporting that really within the scope of quasar?

> And note that the first param is still List[StaticContent], because the caller might (theoretically) want to serve more than one directory of static files, in which case it would not be clear which one should be redirected to.

Also true, but it currently isn't possible for a user to configure more than one directory, so we're not limiting any actual functionality with the change I proposed.

Anyhow, I think think serving static files also falls into the "convenient for SlamData deploys" bucket, so I'm ok with leaving this alone for now with the hope that we eventually remove this feature completely (or at least move it into a SlamData-specific server main or somesuch).

PR #930, opened 2015-09-18
[SD-895] Don't require the MongoDB connection string to include an authentication database

wemrysi:
For certain operations in the MongoDB backend (like checking the `mongod` version) and when faced with a query that doesn't read from a path, we need to be able to supply a database in order to proceed. Previously, we used the authentication db from the mongo connection URI for this, but this doesn't work when using mongo without authentication (or if the user doesn't have permission to write to the authentication db).

This removes the restriction by attempting to find a database the user has access to when needed. For checking the version, we just attempt the `buildinfo` command on the databases the user has access to until one works and error otherwise. For the case when we need a database for temp collections, we try and find the first one that we can create a collection in, again erroring if this fails.

Backend validation is also moved into `BackendDefinition` eliminating the need for `Backend#checkCompatibility`.

### DIFF
--- a/core/src/main/scala/quasar/backend.scala
+++ b/core/src/main/scala/quasar/backend.scala
@@ -349,16 +349,15 @@ object Backend {
       _       <- testWrite(backend)
     } yield ()
 
-  private def testWrite(backend: Backend): ETask[EnvironmentError, Unit] = {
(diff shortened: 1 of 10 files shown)


wemrysi:
Gah, forgot to ready branch, closing and will reopen.

PR #931, opened 2015-09-18
[SD-895] Don't require auth db in mongo connection string

wemrysi:
For certain operations in the MongoDB backend (like checking the `mongod` version) and when faced with a query that doesn't read from a path, we need to be able to supply a database in order to proceed. Previously, we used the authentication db from the mongo connection URI for this, but this doesn't work when using mongo without authentication (or if the user doesn't have permission to write to the authentication db).

This removes the restriction by attempting to find a database the user has access to when needed. For checking the version, we just attempt the `buildinfo` command on the databases the user has access to until one works and error otherwise. For the case when we need a database for temp collections, we try and find the first one that we can create a collection in, again erroring if this fails.

Backend validation is also moved into `BackendDefinition` eliminating the need for `Backend#checkCompatibility`.

SD-895 #done

### DIFF
--- a/core/src/main/scala/quasar/backend.scala
+++ b/core/src/main/scala/quasar/backend.scala
@@ -63,8 +63,6 @@ sealed trait Backend { self =>
   import Path._
   import Planner._
 
-  def checkCompatibility: ETask[EnvironmentError, Unit]
-
   /**
    * Executes a query, producing a compilation log and the path where the result
    * can be found.
@@ -226,8 +224,6 @@ trait PlannerBackend[PhysicalPlan] extends Backend {
 
   lazy val queryPlanner = planner.queryPlanner(evaluator.compile(_))
 
-  def checkCompatibility = evaluator.checkCompatibility
-
   def run0(req: QueryRequest) = {
     queryPlanner(req).map(plan => for {
       rez    <- evaluator.execute(plan)
@@ -341,24 +337,20 @@ object Backend {
 
   def test(config: BackendConfig): ETask[EnvironmentError, Unit] =
     for {
-      backend <- BackendDefinitions.All(config).fold[ETask[EnvironmentError, Backend]](
-        EitherT.left(Task.now(MissingBackend("no backend in config: " + config))))(
-        EitherT.right(_))
-      _       <- backend.checkCompatibility
+      backend <- BackendDefinitions.All(config)
       _       <- trap(backend.ls.leftMap(EnvPathError(_)), err => InsufficientPermissions(err.toString))
       _       <- testWrite(backend)
     } yield ()
 
-  private def testWrite(backend: Backend): ETask[EnvironmentError, Unit] = {
-    val data = Data.Obj(ListMap("a" -> Data.Int(1)))
+  private def testWrite(backend: Backend): ETask[EnvironmentError, Unit] =
     for {
       files <- backend.ls.leftMap(EnvPathError(_))
-      dir = files.map(_.path).find(_.pureDir).getOrElse(Path.Root)
-      tmp = dir ++ Path(".quasar_tmp_connection_test")
-      _ <- backend.save(tmp, Process.emit(data)).leftMap(EnvWriteError(_))
-      _ <- backend.delete(tmp).leftMap(EnvPathError(_))
+      dir   =  files.map(_.path).find(_.pureDir).getOrElse(Path.Root)
+      tmp   =  dir ++ Path(".quasar_tmp_connection_test")
+      data  =  Data.Obj(ListMap("a" -> Data.Int(1)))
+      _     <- backend.save(tmp, Process.emit(data)).leftMap(EnvWriteError(_))
+      _     <- backend.delete(tmp).leftMap(EnvPathError(_))
     } yield ()
-  }
 
   private def wrap(description: String)(e: Throwable): EnvironmentError =
     EnvEvalError(CommandFailed(description + ": " + e.getMessage))
@@ -381,9 +373,6 @@ final case class NestedBackend(sourceMounts: Map[DirNode, Backend]) extends Back
   private var mounts = sourceMounts
 
   private def nodePath(node: DirNode) = Path(List(DirNode.Current, node), None)
-
-  def checkCompatibility: ETask[EnvironmentError, Unit] =
-    mounts.values.toList.map(_.checkCompatibility).sequenceU.map(κ(()))
 
   def run0(req: QueryRequest) = {
     mounts.map { case (mountDir, backend) =>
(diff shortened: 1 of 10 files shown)


core/src/main/scala/quasar/backend.scala
```diff
@@ -472,15 +461,22 @@ final case class NestedBackend(sourceMounts: Map[DirNode, Backend]) extends Back
         b => Process.eval[ETask[E, ?], Path](EitherT(Task.now(path.rebase(nodePath(node)).leftMap(ef)))).flatMap[ETask[E, ?], A](f(b, _))))
 }
 
-final case class BackendDefinition(create: PartialFunction[BackendConfig, Task[Backend]]) extends (BackendConfig => Option[Task[Backend]]) {
-  def apply(config: BackendConfig): Option[Task[Backend]] = create.lift(config)
+final case class BackendDefinition(run: BackendConfig => EnvTask[Backend]) {
+  def apply(config: BackendConfig): EnvTask[Backend] = run(config)
```
sellout: 👍🏾

core/src/main/scala/quasar/evaluator.scala
```diff
@@ -104,6 +97,7 @@ object Evaluator {
       def format(message: String, detail: Option[String]) =
         Json(("error" := message) :: detail.toList.map("errorDetail" := _): _*)
 
+      // TODO: These seem to be rather MongoDB-specific
```
sellout: ConnectionFailed seems generic enough, and the _idea_ of MissingDatabase seems generic. Maybe add a message parameter to MissingDatabase and have the caller mention that it’s an auth db that’s missing?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -136,10 +162,11 @@ trait MongoDbEvaluatorImpl[F[_]] {
         val dbName = physical.op.foldMap {
           case Fix($Read(Collection(dbName, _))) => List(dbName)
           case _ => Nil
-        }.headOption.orElse(defaultDb)
+        }.headOption.orElse(executor.defaultWritableDB)
```
sellout: I know this is existing behavior, but this seems to use the defaultWritableDB more often than it needs to.

When queries are submitted, it’s relative to some directory (at least in the Web and REPL interfaces), any relative directories in the query have that directory prepended.

So … if there is a Mongo mount at `/` and a query with no `Read ` is submitted via `GET http://foo.com/query/fs/myDb?query=…` then temp collections should end up in `myDb`, _not_ `defaultWritableDB`.

And, if there is no context from which we can determine the DB, should we really be writing anywhere? I know we clean up (fingers crossed), but it still seems like bad behavior.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -274,25 +309,37 @@ class MongoDbExecutor[S](client: MongoClient, nameGen: NameGenerator[State[S, ?]
...
-  def version(dbName: String) =
-    liftMongo(
-      Option(client.getDatabase(dbName).runCommand(
-        Bson.Doc(ListMap("buildinfo" -> Bson.Int32(1))).repr)
-        .getString("version")).fold(
-        List[Int]())(
-        _.split('.').toList.map(_.toInt)))
+    def attemptVersion(dbNames: Set[String]): EvaluationTask[List[Int]] =
+      EitherT(dbNames.toList.toNel.toRightDisjunction(NoDatabase()).point[Task])
+        .flatMap(_.traverseU(dbName => lookupVersion(dbName).swap).map(_.head).swap)
```
sellout: What about `Bitraverse.leftTraverse` rather than the `swap`s here?

core/src/main/scala/quasar/evaluator.scala
wemrysi: Aye, I think adding a message parameter to `MissingDatabase` will help. `ConnectionFailed` is a generic concept but the error message makes assumptions about the type of connection and possible reasons for failure (server/port). When we rename `Backend` to `Filesystem` we may want to replace the notion of "database" with something from the filesystem domain.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
wemrysi: > When queries are submitted, it’s relative to some directory (at least in the Web and REPL interfaces), any relative directories in the query have that directory prepended.

Ah, I didn't realize that was the behavior, so does that mean by the time the `Workflow` is crystallized that a `Read` from the parent directory will exist?

> And, if there is no context from which we can determine the DB, should we really be writing anywhere? I know we clean up (fingers crossed), but it still seems like bad behavior.

I agree completely, I'd much prefer not to write anywhere but a database referenced in the query. We have [tests that expect a workflow not involving a database](https://github.com/wemrysi/quasar/blob/ready/SD-895/core/src/test/scala/quasar/physical/mongodb/evaluator.scala#L35) to succeed, which passed before due to the presence of the dubious `defaultDb`. Are these something that we still expect to work?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
wemrysi: Wouldn't that be a traversal over `F[A] = EitherT[Task, A, List[Int]]` and not the list of database names?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: The problem is that if the only reference to a DB is the URI path (IE, it’s a GET with no Read), then that reference is currently lost right after parsing.

But, I’m rethinking this … if there’s a query with no Read, it’s pretty much evaluated via constant folding. The problem is, the Evaluator always returns a path to a collection, so even if something can be evaluated entirely in Quasar, we need to write it to the DB and then return the collection it was written to. SD-958 proposes to fix this. So, I think we already have a way around this planned. IE, it’s fine for now.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: Ah right … you want to traverse over the right, but stop at the first one. Nevermind.

core/src/main/scala/quasar/evaluator.scala
sellout: So, you want to try improving these a bit? I would say that for `ConnectionFailed`, the error could just be “connection failed”, and (as is already the case) the message be the detail.

This is now the only thing to change on this PR 😃

wemrysi:
@sellout Right you are! Deleted.

PR #932, opened 2015-09-19
Remove 'SlamData' from the configuration file path

wemrysi:
Removes another instance of "SlamData" that slipped by, the name of the parent dir of the config file.

### DIFF
--- a/core/src/test/scala/quasar/config/config.scala
+++ b/core/src/test/scala/quasar/config/config.scala
@@ -114,7 +114,7 @@ class ConfigSpec extends Specification with DisjunctionMatchers {
   // NB: Not possible to test windows deterministically at this point as cannot
   //     programatically set environment variables like we can with properties.
   "defaultPath" should {
-    val comp = "SlamData/quasar-config.json"
+    val comp = "quasar-config.json"
     val macp = "Library/Application Support"
     val posixp = ".config"
 
--- a/core/src/main/scala/quasar/config/config.scala
+++ b/core/src/main/scala/quasar/config/config.scala
@@ -122,8 +122,7 @@ object Config {
     val dirPath: RelDir[Sandboxed] = os.fold(
       currentDir,
       dir("Library") </> dir("Application Support"),
-      dir(".config")
-    ) </> dir("SlamData")
+      dir(".config"))
 
     val baseDir = OptionT.some[Task, Boolean](os.isWin)
       .ifM(localAppData, OptionT.none)
@@ -140,10 +139,10 @@ object Config {
    * NB: Paths read from environment/props are assumed to be absolute.
    */
   private def defaultPath: Task[FsPath[File, Sandboxed]] =
-    OS.currentOS >>= defaultPathForOS(file("quasar-config.json"))
+    OS.currentOS >>= defaultPathForOS(dir("quasar") </> file("quasar-config.json"))
 
   private def alternatePath: Task[FsPath[File, Sandboxed]] =
-    OS.currentOS >>= defaultPathForOS(file("slamengine-config.json"))
+    OS.currentOS >>= defaultPathForOS(dir("SlamData") </> file("slamengine-config.json"))
 
   def fromFile(path: FsPath[File, Sandboxed]): EnvTask[Config] = {
     import java.nio.file._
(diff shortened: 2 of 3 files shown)


sellout:
Oh, hrmm, this one is interesting …

In an application environment, I think we _want_ the config to be under the application name. Does this become a new command-line option, defaulting to “quasar”, but available for an application (like SlamData’s launchers) to configure as it may?

core/src/test/scala/quasar/config/config.scala
```diff
@@ -114,7 +114,7 @@ class ConfigSpec extends Specification with DisjunctionMatchers {
   // NB: Not possible to test windows deterministically at this point as cannot
   //     programatically set environment variables like we can with properties.
   "defaultPath" should {
-    val comp = "SlamData/quasar-config.json"
+    val comp = "quasar-config.json"
```
sellout: Do these tests pass? It seems like `comp` should be `"quasar/quasar-config.json"`.

wemrysi: Yes, `defaultPathForOS` used to include the parent dir for the config file, now it does not and only includes the OS-specific portion.

wemrysi:
> In an application environment, I think we want the config to be under the application name. Does this become a new command-line option, defaulting to “quasar”, but available for an application (like SlamData’s launchers) to configure as it may?

Couldn't said application use the `--config` command-line option to specify its custom location?

mossprescott:
@wemrysi's point is a good one, except that there would be no way to specify the app name but still get the config file stored in the same platform-appropriate location.

Furthermore, I think there's something to be said for writing the file in the same location (under `quasar`) either way, if only for our own benefit. I don't think users will particularly care one way or the other since the files are already buried in a more or less hidden location.

sellout:
Right – specifying the app-name while still getting platform-specific locations seems desirable.

It’s fine to just put things in “quasar/” for now, but I _think_ that would cause problems for apps that, say, want to be in the Apple App Store (including us, eventually?), which only have access to “~/Library/Application Support/<app name>”, so we should at least have an issue for an app-name flag.

wemrysi:
@mossprescott @sellout You both raise legitimate concerns for apps that are going to bundle quasar in some form. However, I question whether quasar should even attempt to pick a platform-specific default in the first place as this seems to have converted a deploy-time decision into a compile-time one. It is also unlikely to be correct in many situations, for example if quasar is packaged up for a particular Linux distribution the configuration will likely be in a distribution-specific location and thus specified via `--config`. Similarly for FreeBSD, Solaris, etc.

In the case where it is known, a priori, that the configuration _must_ live in a particular location (like in the Apple App Store scenario), does it make sense for an application bundling quasar to rely on the defaults to get that correct?

That said, I'm getting a bit off topic. For this PR if it helps ease the transition from SlamEngine -> Quasar while enabling us to remove references to "SlamData" in the codebase, I'm ok with adding the (hidden?) "app name" configuration option as a stopgap until we're able to come to a consensus on this issue.

mossprescott:
Sigh. This really comes down to some slightly ugly hacks that simplify deployment while creating confusion.

The Quasar jar should probably only serve the API, and should probably default to reading its config from `.../quasar/`. I don't think it really makes sense to complicate the Quasar executable with too many options and special support for any particular deployment. That jar really should serve *only* as a simple way to experiment with the API.

Instead, the server process bundled with the SlamData installer should probably include that jar as a dependency, but have its own main method that runs code to figure out the config location and to set up the serving of static files. That would mean building a little bit of Scala to run there, something we don't have at this point.

For the time being, I don't think we need to worry all that much about any of this. We could always migrate the file again in the future if need be. The main goal is just to get rid of unexpected references to SlamData from the project (and its behavior), and I think the PR as is accomplishes that.

:+1:

wemrysi:
> Sigh. This really comes down to some slightly ugly hacks that simplify deployment while creating confusion.

Agreed. I'm fine with this PR as-is if ya'll are.

sellout:
We probably need a JIRA “story” or something to cover the generalizations @mossprescott describes. Then this looks good.

wemrysi:
I've created a [JIRA issue](https://slamdata.atlassian.net/browse/SD-1027) to extract the SlamData-specific functionality.

PR #936, opened 2015-09-22
[SD-837] Fix handling of constants in library aggregation functions

jedesah:
Authored with @wemrysi

### DIFF
--- a/core/src/main/scala/quasar/std/agg.scala
+++ b/core/src/main/scala/quasar/std/agg.scala
@@ -30,6 +30,7 @@ trait AggLib extends Library {
     noSimplification,
     partialTyper {
       case List(Type.Const(Data.Set(Nil))) => Type.Const(Data.Int(0))
+      case List(Type.Const(_))             => Type.Const(Data.Int(1))
       case List(_)                         => Type.Int
     },
     κ(success(Type.Top :: Nil)))
@@ -45,22 +46,25 @@ trait AggLib extends Library {
 
   val Min = Reduction("MIN", "Finds the minimum in a set of values", Type.Comparable :: Nil,
     noSimplification,
-    reflexiveTyper,
+    identityTyper,
     reflexiveUnary(Type.Comparable))
 
   val Max = Reduction("MAX", "Finds the maximum in a set of values", Type.Comparable :: Nil,
     noSimplification,
-    reflexiveTyper,
+    identityTyper,
     reflexiveUnary(Type.Comparable))
 
   val Avg = Reduction("AVG", "Finds the average in a set of numeric values", Type.Numeric :: Nil,
     noSimplification,
-    constTyper(Type.Dec),
+    partialTyper {
+      case List(Type.Const(t))  => Type.Const(t)
+      case _                    => Type.Dec
+    },
     NumericUnary)
 
   val Arbitrary = Reduction("ARBITRARY", "Returns an arbitrary value from a set", Type.Top :: Nil,
     noSimplification,
-    reflexiveTyper,
+    identityTyper,
     reflexiveUnary(Type.Top))
 
   def functions = Count :: Sum :: Min :: Max :: Avg :: Arbitrary :: Nil
--- a/core/src/main/scala/quasar/std/library.scala
+++ b/core/src/main/scala/quasar/std/library.scala
@@ -67,10 +67,9 @@ trait Library {
       Func.Untyper =
     partialUntyperOV(codomain)(f.lift(_).map(success))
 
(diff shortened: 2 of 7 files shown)


mossprescott:
The ticket mentioned several error cases that aren't addressed here as far as I can tell. Can we get tests for some of the queries with multiple, mixed projections? I suspect some of those bugs are still there. Ideally, these cases would be covered by the query generator in the planner tests, but I suspect that adding some of these aggregation functions will confuse those tests' assertions.

core/src/main/scala/quasar/std/agg.scala
```diff
@@ -30,6 +30,7 @@ trait AggLib extends Library {
     noSimplification,
     partialTyper {
       case List(Type.Const(Data.Set(Nil))) => Type.Const(Data.Int(0))
+      case List(Type.Const(_))             => Type.Const(Data.Int(1))
```
sellout: This case isn’t actually correct. You could have `Type.Const(Data.Set(set))`, in which case you want to return `Type.Const(Data.Int(set.length))`.

core/src/main/scala/quasar/std/agg.scala
```diff
@@ -45,22 +46,25 @@ trait AggLib extends Library {
...
     noSimplification,
-    reflexiveTyper,
+    identityTyper,
     reflexiveUnary(Type.Comparable))
 
   val Avg = Reduction("AVG", "Finds the average in a set of numeric values", Type.Numeric :: Nil,
     noSimplification,
-    constTyper(Type.Dec),
+    partialTyper {
+      case List(Type.Const(t))  => Type.Const(t)
```
sellout: Same in this case.

wemrysi:
> This is definitely not correct according to the SQL standard. It should count the rows in zips (technically, it is the number of rows for which the expression 7 does not evaluate to null). Are we consciously deviating from the standard?

Ah, good catch, this wasn't intended to deviate from the standard.

sellout:
So, how do we reconcile this with the expectation that `sum(1) = 1`?

wemrysi:
Are you saying that `sum(1) = 1` _is_ the expected behavior? If so, then I guess we're deviating from the standard.

sellout:
I thought that was mentioned in the ticket, which @jdegoes did say deviates from SQL.

jdegoes:
> I thought that was mentioned in the ticket, which @jdegoes did say deviates from SQL.

Under MRA, which has local / compositional semantics, yes, `SUM(1) = 1`. That's sort of independent from SQL<sup>2</sup>, however, since we can "fix up" these sorts of things quite easily (in MRA, emulation can be done by a map of the `*` set to a constant, namely `1`, which is then aggregated). 

In any case, the fix for ANSI SQL compatibility would be in the _logical plan_, after compilation. The planner should definitely be compiling `SUM(1)` to something that evaluates to `1`.

The question of whether or not we want to "fix up" (re-break?) SQL for backward compatibility is a good one. Since we are introducing syntactic extensions to satisfy all characteristics of NoSQL analytics systems (rather than emulating flat tables or some such), the SQL people will be using in practice is not ANSI SQL; they can't just wholesale re-use existing queries. So I think we have flexibility to break with ANSI SQL in a few key places for consistency, or possibly introduce compilation flags (non-local mode versus local mode) in which various transformations are used to "fix up" a compiled logical plan before sending it off to the planner.

sellout:
Ok, so the current fix is correct (at least in concept) for the MRA semantics. For the SQL semantics, you need to insert the Constantly function I mentioned the other day. I _think_ this currently needs to be done in compile0, because that's where we have the appropriate table context to get our set from.

core/src/main/scala/quasar/std/agg.scala
wemrysi: How much actual computation is appropriate to do in these typer functions? I don't disagree with what you're saying here, but this would also imply that we should compute the average of a `Const(Data.Set(xs))` in the `Avg` function as well, no?

wemrysi: Ah, nevermind, I guess that's what you meant by the comment below.

mossprescott:
Build failed with everyone's favorite intermittent API test, only this time it was *way* over the expected size. How can gzipped size differ by more than a factor of 2? It makes you wonder if the data actually got written twice.

I suggest we either submit another PR to straighten that out and merge that, or else fix it here.

wemrysi:
Closing, as it makes more sense to base this work on the result of merging #934 as the latter contains many of the same changes as this PR as well as ones that are required to finish this PR. Once that PR has been finished and merged, a new one for this work will be opened based on it.

PR #941, opened 2015-09-25
Remove Java property in favor of build-info plugin

jedesah:


mossprescott:
Looks like the plugin failed when it tried to push coverage to coveralls: https://travis-ci.org/slamdata/quasar/builds/82195465#L3776.

This currently doesn't fail the build; not sure if that's what we want or not.

Anyway, I think it's worth looking into.

mossprescott:
Contrast with this successful build on master: https://travis-ci.org/slamdata/quasar/builds/82002395#L4012

jedesah:
Okay, I looked into this and it appears to be a bug with the latest version of `sbt-coveralls` interacting with the latest version of `sbt-scoverage`. Unfortunately, downgrading scoverage is not possible because `sbt-coverage` it did not expose `enabled` through an sbt Key before 1.3.x.

This PR will have to wait on this which may be fixed in as little as a few days if the project can get a new version out: https://github.com/scoverage/sbt-coveralls/issues/62

jedesah:
Closing because it's blocked. Tracked here: https://slamdata.atlassian.net/browse/SD-1042

PR #943, opened 2015-10-01
Separate `limit` and `offset` operators from `select`.

sellout:
#done SD-1001, #done SD-1002

Also, fix `TransformSelect` to work on nested selects.

### DIFF
--- a/core/src/main/scala/quasar/std/set.scala
+++ b/core/src/main/scala/quasar/std/set.scala
@@ -24,7 +24,7 @@ import scalaz._, NonEmptyList.nel, Validation.{success, failure}
 import scalaz.syntax.applicative._
 
 trait SetLib extends Library {
-  val Take = Sifting("TAKE", "Takes the first N elements from a set", Type.Top :: Type.Int :: Nil,
+  val Take = Sifting("(LIMIT)", "Takes the first N elements from a set", Type.Top :: Type.Int :: Nil,
     noSimplification,
     partialTyper {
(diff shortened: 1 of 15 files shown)


sellout:
Assigned to @jedesah, but bring in @drostron on the review.

core/src/main/scala/quasar/std/set.scala
```diff
@@ -24,7 +24,7 @@ import scalaz._, NonEmptyList.nel, Validation.{success, failure}
 import scalaz.syntax.applicative._
 
 trait SetLib extends Library {
-  val Take = Sifting("TAKE", "Takes the first N elements from a set", Type.Top :: Type.Int :: Nil,
+  val Take = Sifting("(LIMIT)", "Takes the first N elements from a set", Type.Top :: Type.Int :: Nil,
```
jedesah: Why the parenthesis around LIMIT?

jedesah: Can we add documentation explaining what `Sifting` represents.

sellout: The name of Funcs is currently an unfortunate conflation of its name and the name of the corresponding SQL function/operator. Basically we look up Funcs from the name of the SQL operation and the name has to match. It's something that needs to be fixed. And operators are displayed with parens.

sellout: I'll add docs for Sifting, but it's the same operations as ShapePreservingBuilder – they take a set and operate on the set itself, leaving individual elements unchanged – eg, sort, filter, take.

PR #946, opened 2015-10-02
Add codecov for trial use

jedesah:
- Update travis Scala version
- minor refactoring of travis build steps


wemrysi:
What advantages does codecov bring over coveralls? Or is the point of this PR to try it out and see if there are any?

What is benefit of inlining `buildSuccess` into the travis config and deleting `buildFailure`? The travis config still references `buildFailure` in the `after_failure` option.

I also don't think codecov is reporting accurate coverage, at least it doesn't match with what we report via scoverage or coveralls, actually all three report different numbers:

* scoverage: 79.75%
* coveralls: 81.381%
* codecov: 82.25%

we should try and get codecov (and coveralls) to match what scoverage reports if they're going to be useful (as scoverage is what we currently use to fail the build).

jedesah:
@wemrysi It's not 100% clear, but supposedly codecov might bring coverage diff to the table. Aside from that, just a trial to see how it goes.

I am not sure why the numbers are all different. I will have a look and see if I can figure it out in a reasonable amount of time.

sellout:
`buildFailure` was an empty script … but yeah, travis.yml shouldn’t reference it any more. The entire `buildSuccess` script was wrapped in a `if [ $TRAVIS ]`, and Travis has nice handling of output when you run commands in it directly, so inlining it seemed reasonable.

As far as the different numbers, scoverage does expression-level coverage, so that is the most accurate and what we actually trust for our numbers. I’m not sure why coveralls and codecov differ. Coveralls is line-based, and I assume codecov is as well

We do also fail the build if coveralls drops at all (it would be nice if we could get scoverage to fail like that, but ¯\_(ツ)_/¯). I think the potential to diff coverage is the main benefit of codecov.

wemrysi:
> As far as the different numbers, scoverage does expression-level coverage, so that is the most accurate and what we actually trust for our numbers. I’m not sure why coveralls and codecov differ. Coveralls is line-based, and I assume codecov is as well

Aye, though that feature implies state of some sort, so not sure if it is possible with our stateless scoverage build.

sellout:
Yeah, really I want coveralls or codecov or whatever to actually use the expr-level info that scoverage gives it. Then the service has the right numbers and has the state for diffing. And then we can disable failing travis when scoverage doesn’t meet minimum (since coveralls will fail in that case).

mossprescott:
Agree on disabling the scoverage limit. I think we'd be much better of just having codecov highlighting the coverage in diffs (assuming it works), to help catch missing tests during review. The number was never all that meaningful anyway.

PR #947, opened 2015-10-06
Remove GroupId.

sellout:
Since `GroupBuilder` started holding separate `WorkflowBuilder`s for its
keys, maintaining a `GroupId` created with the original set of keys is
no longer necessary. We still generate the ID in `render` to make it
easy to eyeball groups that should align.

Also removed a no-longer-used `merge` case.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
@@ -180,20 +180,15 @@ object WorkflowBuilder {
   type GroupValue[A] = AccumOp[A] \/ A
   type GroupContents = DocContents[GroupValue[Expression]]
 
-  final case class GroupId(srcs: List[WorkflowBuilder]) {
-    override def toString = hashCode.toHexString
-  }
-
   final case class GroupBuilderF[A](
-    src: A, keys: List[A], contents: GroupContents, id: GroupId)
+    src: A, keys: List[A], contents: GroupContents)
       extends WorkflowBuilderF[A]
   object GroupBuilder {
     def apply(
       src: WorkflowBuilder,
       keys: List[WorkflowBuilder],
-      contents: GroupContents,
-      id: GroupId) =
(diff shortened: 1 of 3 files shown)


sellout:
Now I _definitely_ prefer codecov over coveralls 😉

wemrysi:
Aye, that is nice.

core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -1022,45 +1016,45 @@ object WorkflowBuilder {
...
+        case (GroupBuilderF(_, _, Doc(_)), ValueBuilderF(_)) => delegate
 
         case (
-          GroupBuilderF(src1, keys, Expr(\/-($var(DocVar.ROOT(_)))), id1),
-          GroupBuilderF(src2, _,    Expr(\/-($var(DocVar.ROOT(_)))), id2))
-            if id1 == id2 =>
-          impl(src1, src2, combine).map(GroupBuilder(_, keys, Expr(\/-($$ROOT)), id1))
+          GroupBuilderF(src1, keys1, Expr(\/-($var(DocVar.ROOT(_))))),
+          GroupBuilderF(src2, keys2, Expr(\/-($var(DocVar.ROOT(_))))))
+            if keys1 ≟ keys2 =>
```
wemrysi: We're replacing an equality comparison on ids with one on keys, but the `GroupId` used to contain the `src` of the `GroupBuilderF` (based on the previous implementation of `groupBy(...)`), should we also be including the `src` in the new comparisons? Does omitting it change the semantics? Same goes for the other instances where we make the same change.

sellout: The src doesn’t have to be the same. It did before, when the keys were expressions that pointed into the `src` (IE, the keys were dependent on the src), but now keys are their own WorkflowBuilders and have no reliance on the `src`.

PR #948, opened 2015-10-06
Demonstrates that SD-892 is not a bug

jedesah:
SD-426 still needs investigation as it would appear SD-892 was not the issue.

I can rebase this PR is that is deemed appropriate (not sure because these commits were in a previous ready PR)


jdegoes:
This does not demonstrate that SD-892 is not a bug, only that this test fails to reproduce it. I feel like the test has minimal utility, although the larger concern that would be worth testing is that every query returns identical results whether the query is executed via GET or POST, albeit that would require quite a substantial change and this series of related tickets is already long overdue for a fix.

jedesah:
@jdegoes I did not realize that SD-892 was the parent issue, so I was going to close this one and continue work as SD-426, but indeed now realizing that SD-892 is the parent issue, this PR name does not make sense.

jedesah:
Closing in favor of #949 which does not contain the test that failed to reproduce SD-892

PR #949, opened 2015-10-06
[SD-1043] Collection of test infrastructure improvements

jedesah:
- Rename asAbsolute to removeCurrentDir and add documentation
- Add documentation on scan and scan0
- Flesh out interactive package and put it in main
- Add tests for interactive package
- regression tests now use the interactive package to load dataResources
- Add version of backendShould that accepts dataResources and performs cleanup
- remove code in build script that loads in zip data and remove notion of ambient MongoDB


jedesah:
@mossprescott This PR is ready and should be pretty quick to review as I have already addressed most of the issues that you brought up during a previous review. I am not sure why codecov is saying this change decreases coverage, I would be inclined to trust coveralls.

core/src/main/scala/quasar/fs/path.scala
```diff
@@ -48,7 +48,10 @@ final case class Path(dir: List[DirNode], file: Option[FileNode]) {
     case _=> this
   }
 
-  def asAbsolute: Path = dir match {
+  /**
+   * If this path contains [[DirNode.Current]] as the outmost node, return this path minus that node.
+   */
+  def removeCurrentDir: Path = dir match {
```
mossprescott: How is this an improvement? `DirNode.Current` seems like an implementation detail, not something we want to make users *more* aware of.

Anyway, once we convert to pathy, absolute/relative is represented in a clearer way and in the type), so meh.

core/src/test/scala/quasar/fs/path.scala
```diff
@@ -143,39 +143,39 @@ class PathSpecs extends Specification with DisjunctionMatchers {
 
   "Path.asAbsolute" should {
```
mossprescott: If we're renaming the method, this should get updated too.

mossprescott:
No other comments. I guess I would prefer to leave that method as is, but I don't feel super strongly about it.

Also, disappointed that codecov won't show me the diff, but I guess this isn't the best PR to test it with, since a lot of the changes are just moving non-code files around.

core/src/main/scala/quasar/fs/path.scala
jedesah: The reason I changed it is because my understanding is that this does not implement what one might expect from "asAbsolute". So in the absence of a semantically correct name, I just gave it a name saying what it does.

mossprescott: I see your point, but I don't think that name is any good, since it exposes a confusing aspect of the implementation. Maybe there's a simpler fix. How about replacing uses with `Path.Root ++ _` or `_.relativeTo(Path.Root)`? I think one or the other of those would have the same affect and be easier to understand.

PR #950, opened 2015-10-07
A few improvments on the SQL side.

sellout:
* A better error when faced with an unbound SQL var,
* avoid wrapping/unwrapping a string in `parse`/`parseExpr`,
* a test to show that we can parse `primary_expr` at the top level.

This doesn’t fix any issues, but hopefully provides some clarity on
SD-1044 and SD-1046.

### DIFF
--- a/core/src/main/scala/quasar/compiler.scala
+++ b/core/src/main/scala/quasar/compiler.scala
@@ -587,7 +587,7 @@ trait Compiler[F[_]] {
 
           case NullLiteral() => emit(LogicalPlan.Constant(Data.Null))
           case Vari(name) =>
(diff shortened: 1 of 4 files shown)


core/src/main/scala/quasar/compiler.scala
```diff
@@ -586,6 +586,8 @@ trait Compiler[F[_]] {
           case BoolLiteral(value) => emit(LogicalPlan.Constant(Data.Bool(value)))
 
           case NullLiteral() => emit(LogicalPlan.Constant(Data.Null))
+          case Vari(name) =>
```
drostron: Do we want to add a test exercising this?

PR #951, opened 2015-10-07
Move MongoDB backend tasks to a per-mount thread pool

mossprescott:
http4s expects all Tasks to be non-blocking, and runs all requests on a thread pool which has only one thread per CPU core/thread. Since we use the synchronous MongoDB API, all of our backend tasks were blocking that thread, meaning that web API requests could be blocked by a small number of long-running MongoDB queries, say. In particular, the original bug report currently being tracked by [SD-896](https://slamdata.atlassian.net/browse/SD-896), which happened on a VM with only a single CPU core.

This PR creates a thread pool for each MongoDB mount, so that API requests for separate mounts (and requests that don't hit MongoDB) can run fully independently. Each mount's pool has a small number of threads, because it appears that running more than one or two simultaneous requests against MongoDB has a large impact on performance (all of the queries take *much* longer than usual, and all are delayed until all complete). In any case, having a separate pool for each mount will make it easy to tune the level of parallelism in the future, or even to make it a configuration item.

Also, refine the MapReduce ADT to reflect the permitted options in MongoDB.


jdegoes:
> Each mount's pool has a small number of threads, because it appears that running more than one or two simultaneous requests against MongoDB has a large impact on performance (all of the queries take much longer than usual, and all are delayed until all complete). 

This does not sound good (i.e. it doesn't sound like it will address user issues) or right (Mongo, like most NoSQL dbs, is designed for large number of concurrent reads, not stacking up the reads as is described here).

Can you characterize these queries? Are they all map/reduce queries?

mossprescott:
@jdegoes: *this* PR is meant to address the specific complaint, which was about the server being unable to serve static content after a long-running query was submitted. That was definitely a problem, especially in scenarios where that pool consisted of a single thread, and that won't happen anymore now that those queries don't block in http4s's thread pool.

There may be more to do to figure out if things are not performing as expected on the MongoDB side. My most repeatable testing involved running multiple identical map-reduce queries against my local MongoDB instance while also making other non-MongoDB requests to the API. I also tested with the slowest aggregation queries I could come up with, and saw a similar effect.

core/src/main/scala/quasar/backend.scala
```diff
@@ -252,6 +252,49 @@ trait PlannerBackend[PhysicalPlan] extends Backend {
...
+final case class ThreadPoolAdapterBackend(
+    backend: Backend,
+    pool: java.util.concurrent.ExecutorService) extends Backend {
+
+  private def fork[E] = new (ETask[E, ?] ~> ETask[E, ?]) {
+    def apply[A](t: ETask[E, A]): ETask[E, A] =
+      EitherT(Task.fork(t.run)(pool))
+  }
+
+  def count0(path: Path)      = fork(backend.count0(path))
```
jdegoes: @wemrysi's free refactoring is going to totally clean this up. :smile:

core/src/main/scala/quasar/physical/mongodb/mapreduce.scala
```diff
@@ -61,30 +61,33 @@ object MapReduce {
     def bson(dst: Collection): Bson
   }
 
-  sealed trait Action
+  sealed trait Action {
+    def nonAtomic: Option[Boolean]
```
jdegoes: The semantics of `Option[Boolean]` are not obvious, how about a 3-way ADT instead?

jdegoes:
> this PR is meant to address the specific complaint, which was about the server being unable to serve static content after a long-running query was submitted

Which ticket are you referring to? SD-896 is about queries being executed sequentially instead of in parallel. If that's fixed by this issue, then :+1: , although it sounds like we've got some digging to do in order to understand why query performance is O(n), where n = number of queries. On multi-core and distributed systems, MongoDB should be performing much better up to some bound (= paralellization factor).

jdegoes:
How about we merge this in (after review) and create a separate ticket for the remaining issues you saw?

core/src/main/scala/quasar/physical/mongodb/mapreduce.scala
mossprescott: Indeed. This is a shallow mapping of the driver API, which accepts a boolean but doesn't require it to be present. Since the default is documented, I think we can get away with a 2-way ADT.

mossprescott:
SD-896 talks about parallel execution, but was based on speculation about causes of the behavior that was seen in the front-end. When I dug into the original front-end ticket, it was clear that the original report was about the UI being blocked by a long query, so that's what I fixed first.

I don't know much about the limits of map-reduce performance, but I'm not all that surprised that my current setup doesn't perform well (single machine, no index, map-reduce without a preceding filter, and running the same query over the same data). We might need to set up a more realistic test to get a feel for whether we're getting in the way at all when MongoDB is performing well.

I'll write up another ticket for that.

wemrysi:
:+1:, @mossprescott did you write up the ticket for the other issue(s)? I'll merge once you have (just so we don't forget to do it).

bwmcadams:
@jdegoes  asked me about this... and I'd have to dig into code in a lot more detail still. However,  there are generally a **lot** of issues w/ the official MongoDB approach to connection management. Not just that they basically *force* a connection pool upon you, that they synchronize and lock way too many places in what seems like sloppy/heavy-handed/lazy/scared ways... but the official approach to how to deal with connections and thread pinning. Many of these can probably cause the kinds of issues my brief skim of the issue shows me.

You may find my writeup (which got me some flak when I wrote it)  on [Connection Pinning](https://github.com/bwmcadams/hammersmith/blob/no-sleep-til/ConnectionPinning.md) enlightening.

The hope is to finish Hammersmith to help with this, but without help no ETA.

wemrysi:
@mossprescott assures me he will create a new ticket, so merging this.

PR #953, opened 2015-10-07
bump http4s to 0.10.1

drostron:
- #done SD-857 (fixed via update)
- preparation for SD-884

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: scala
 scala:
 - 2.11.7
+jdk: oraclejdk8
 sudo: false
 before_install:
 - "./scripts/installMongo mongodb-linux-x86_64-2.6.9 mongo2.6 27018"
--- a/web/src/main/scala/quasar/api/fs.scala
+++ b/web/src/main/scala/quasar/api/fs.scala
@@ -24,14 +24,15 @@ import quasar.fs._, Path.{Root => _, _}
 import quasar.sql._
 
 import scala.collection.immutable.TreeSet
+import scala.concurrent.duration.DurationInt
 import java.nio.charset.StandardCharsets
 
 import argonaut.{DecodeResult => _, _ }, Argonaut._
 import org.http4s.{Query => HQuery, _}, EntityEncoder._
 import org.http4s.argonaut._
-import org.http4s.dsl.{Path => HPath, listInstance => _, _}
+import org.http4s.dsl.{Path => HPath, _}
 import org.http4s.headers._
-import org.http4s.server._
+import org.http4s.server._, syntax.ServiceOps
 import org.http4s.util.{CaseInsensitiveString, Renderable}
 import scalaz._, Scalaz._
 import scalaz.concurrent._
@@ -571,7 +572,7 @@ final case class FileSystemApi(
             })
             val ct = `Content-Type`(MediaType.`application/zip`)
             val disp = ResponseFormat.fromAccept(accept).disposition
-            linesResponse(bytes).map(_.withHeaders((ct :: disp.toList): _*))
+            linesResponse(bytes).map(_.replaceAllHeaders((ct :: disp.toList): _*))
           }).join)
       } else {
         backend.flatMap(bknd => responseStream(accept, scan(path, bknd)))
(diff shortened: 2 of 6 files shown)


web/src/main/scala/quasar/api/package.scala
```diff
@@ -61,34 +54,7 @@ package object api {
...
-    def apply(service: HttpService): HttpService =
-      Service.lift { req =>
-        service.run(req).handleWith {
-          case err => InternalServerError(Json("error" := ("uncaught: " + err.toString))).map(Some(_))
-        }
-      }
-  }
-
-  object HeaderParam extends Middleware {
+  object HeaderParam {
```
jedesah: As discussed, we can extend HttpMiddleware

web/src/main/scala/quasar/api/package.scala
```diff
@@ -61,34 +54,7 @@ package object api {
...
-      Service.lift { req =>
-        service(req).flatMap {
-          case None if req.method == OPTIONS => Ok().map(Some(_))
-          case r => Task.now(r)
-        }.map(_.map(_.putHeaders(corsHeaders: _*)))
-      }
-  }
-
-  /** Handle failure in Task by returning a 500. Otherwise http4s hangs for 30 seconds and then returns 200. */
-  object FailSafe extends Middleware {
```
jedesah: Was this not being used before?

web/src/main/scala/quasar/api/fs.scala
```diff
@@ -625,6 +626,15 @@ final case class FileSystemApi(
     }
   }
 
+  def corsMiddleware(svc: HttpService): HttpService = middleware.CORS(
```
jedesah: I think this method should either be renamed or the GZip `MiddleWare` should be applied outside of it.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: scala
 scala:
 - 2.11.7
+jdk: oraclejdk8
```
sellout: Do we need Java 8? Previously our minimum version was 7. I think it’s good to be explicit about it in Travis, but we should use 7 if we can, and if not, we have some other things to update (docs, installers, etc.)

drostron: The latest http4s requires 8. Web would be limited to 8 but we could potentially utilize a Travis build matrix to build core on 7 and 8 if necessary. Do we think going full 8 is acceptable if docs, installers, ... are updated accordingly?

web/src/main/scala/quasar/api/fs.scala
drostron: Agreed. I'll move it out.

web/src/main/scala/quasar/api/package.scala
mossprescott: This was a workaround for a bug in the Jetty backend, and hasn't been used since we switched to Blaze.

drostron: Correct.

drostron: Ah, good to know the background — was curious.

web/src/main/scala/quasar/api/fs.scala
```diff
@@ -625,6 +626,15 @@ final case class FileSystemApi(
...
+    middleware.GZip(HeaderParam(svc)),
+    middleware.CORSConfig(
```
mossprescott: I'm in favor of switching to the http4s-provided CORS implementation, but I suggest testing this with the front-end, using multiple browsers (at least Chrome and not-Chrome). We have very minimal test coverage for CORS, and we've broken it a couple of times.

drostron: Agreed. It would be good to know it's still working as expected. Is this test setup documented anywhere? If not it would be good to work with someone that's set up it up before.

mossprescott: The usual approach is to install the latest release and swap out the jar. I know, sophisticated, right?

.travis.yml
jdegoes: We can move to Java 8. Just needs a ticket to track all the crap that has to be done (installers, build files, website docs, readmes, etc.).

drostron: I created SD-1051 to track.

drostron:
I just realized this is from a `wip` rather than `ready` branch. Once current comments are addressed I'll close/reopen from a `ready` branch.

web/src/main/scala/quasar/api/fs.scala
drostron: I simulated an environment where CORS was triggered and verified that SlamData can successfully communicate with Quasar. Both Chrome and Firefox were used.

drostron:
superseded by #954

PR #954, opened 2015-10-08
bump http4s to 0.10.1

drostron:
- supersedes #953
- #done SD-857 (fixed via http4s update)
- preparation for SD-884

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ See the instructions below for running and configuring these JARs.
 
 ## Building from Source
 
-**Note**: This requires Java 7 and Bash (Linux, Mac, or Cygwin on Windows).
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ See the instructions below for running and configuring these JARs.
 
 ## Building from Source
 
-**Note**: This requires Java 7 and Bash (Linux, Mac, or Cygwin on Windows).
+**Note**: This requires Java 8 and Bash (Linux, Mac, or Cygwin on Windows).
```
jedesah: How true is it anymore that Bash is required to build from source? Not suggesting you change it, just curious.

sellout: Yeah, it really looks like any shell should be fine. Bash is necessary for running the build script, but that’s not used here.

PR #955, opened 2015-10-13
[SD-1057] Extract mongoDrive and fp subproject

jedesah:


jedesah:
I am a little concerned that this change might break the oneJar, and not sure how to test that quickly locally. The Travis build should validate though that it works correctly.

jdegoes:
Up to @wemrysi but my 2 cents:

 * :-1: On pulling out mongodriver, because (a) the thing being pulled out is *not* a mongo driver (just two more or less random files), and (b) this organization does not reflect any planned future organization (e.g. factoring out each backend into its own subproject, or pulling out the type safe MongoDB wrapper into a MongoDB algebra + data structures, etc.).
 * Neutral on factoring out the `Predef` into `fp` package (it's not generally useful, but it could have value to eventual subprojects).

jedesah:
@jdegoes The `Predef` was factored out as a necessary step to factor out "mongo-driver". I agree that the current "mongoDriver" is small and crappy, but I didn't want to spend a whole lot of time actually extracting everything right now, thought it could be an ongoing process. I think what is there right now will probably totally disappear if we write a real mongo-driver but right now it's what we are actually using and this actually has the immediate benefit of unbreaking IntelliJ and reducing the amount of files that constantly getting recompiled because of our incremental build issues.

jedesah:
In general, I am not a huge fan of changes to code in order to satisfy tooling, but in this case I find that the spirit of the change (the intention to eventually create a standalone mongoDriver) combined with the practical benefit of unbreaking tooling is probably worth it.

mossprescott:
You could make this a bigger step towards a real driver by also pulling out MongoWrapper from filesystem.scala. That's where we wrap raw driver calls in `Task`, and the idea was to eventually also use it from the evaluator.

What's the vision here? Will this sub-project/repo eventually have:
1. a "low-level" Mongo free algebra and interpreter? I guess that's one of the questions @jdegoes raised above.
2. the "concrete" `WorkflowOp` ADT, and the more tree-ish `Workflow`?

wemrysi:
I agree with @jdegoes' reasons and :-1: for not creating the `mongo-driver` sub project.

@jedesah What is the issue with IntelliJ that this solves? Is creating the sub-project the only way of solving it?

jedesah:
@wemrysi IntelliJ tends to crash on certain files (typically large ones). Creating sub-projects/separate projects as well as keeping sources files relatively short in general always helps. In this particular case I was hitting a 100% crash situation when trying to edit this particular part of a file, so by making a sub-project and working on things in there I got a snappy IDE experience.

What about @mossprescott's suggestion to bring in `MongoWrapper` which is where I should have put the `exists` function anyway, does that improve things? If we are not creating a "mongo-driver" sub-project right now with this PR, then what is the plan to go about extracting the general purpose mongo functionality into a separate project and streamlining our codebase which has all kinds of benefits including making crappy (but useful) IDEs better?

jdegoes:
IMO it's not good practice to reorganize projects and files just to satisfy the requirements of a particular development environment. From an organizational perspective, this PR pulls in two seemingly random files that do not constitute a "mongo driver" into a project called "mongo driver" to make IntelliJ Idea happy. :smile: 

It's sort of like not using fixplate because IntelliJ cannot understand complex, partially applied, higher-kinded types. When the tools fail us, we shouldn't change the code, but rather change our tools!

There is no business reason to create a "mongo driver" project. I think some of our Mongo wrappers, and maybe our future Mongo API algebra, may be useful to others (and hopefully encourage third-party contributions), but factoring that out into a separate project should (a) definitely be a 10% time project (it's not something that will ever be scheduled in the SlamData roadmap), and (b) is quite different than pulling two random files into a separate subproject to prevent an IDE crasher.

PR #956, opened 2015-10-13
Insert runtime expression guards.

sellout:
The primary purpose of this is to avoid operations that can cause MongoDB to error when executed on the wrong types. This is an overall simpler type checker that will be applied to any interface that uses the LogicalPlan. Also, since MongoDB has total ordering, a test like `foo > 10` would pass if foo was almost any non-numeric value, so now we do both the original comparison as well as verify the type.

Here are the high-level changes (roughly in diff order):
* look up Funcs directly in compiler (rather than having an annotation);
* remove constant folding and `ConcatOp` decision from the compiler;
* clean up `Type.Set.dataType`;
* add `codomain` to Func (to simplify untypers);
* eliminate safe JS (typechecker now handles inserting fewer and more precise checks);
* improve JsCore simplification;
* add TypecheckF to the LogicalPlan;
* add type inference, type checking (fixes SD-779), constant folding, runtime check insertion (fixes SD-763 and SD-852), and `ConcatOp` decision to LogicalPlan;
* add mongodb Expression simplification;
* add Expression → JsCore typecheck conversion;
* add variadic apply to Selector.And and .Or;
* add Typecheck → WorkflowBuilder compilation;
* add an optional fold for removing `Type.Obj` checks after Read (EG, in Mongo, we know the result of a read is an Obj);
* use Js pprint rather than BSON repr when printing Reshapes;
* add cases to WorkflowBuilder normalization;
* remove manual type checks from `flattenObject` and `flattenArray`;
* remove `FunctionBind`, `TypeInfer`, and `TypeCheck` analyses from the SQL AST;
* many fixes to the std library types;
* remove the `Product` type (fixes SD-574);
* fix a couple other issues in Type.typecheck (fixes SD-862);
* move `Interval` out of `Temporal`;
* add `cofCataM` and `hyloM` to recursionschemes;
* add `expandBindings` to Binder;
* use `State` for LP freshName;
* fix `Js.BinOp` parens;
* a bunch of `WorkflowBuilder.normalize` improvements;
* added some previously-unseen `objectConcat` cases;
* added a simplifier for `ProjectField`;
* modified rendering of GroupBuilder to make `id` more obvious;
* make the property test check for pushing $$ROOT a bit stricter;
* fix projectField on simple GroupBuilders;
* elide Doc GroupBuilder with no accumulators;
* loosen an overly-restrictive merge case;
* improve regression test error reporting; and
* make StructureType a functor.


core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -95,6 +103,8 @@ object LogicalPlan {
       case (FreeF(n1), FreeF(n2)) => n1 ≟ n2
       case (LetF(ident1, form1, in1), LetF(ident2, form2, in2)) =>
         ident1 ≟ ident2 && form1 ≟ form2 && in1 ≟ in2
+      case (TypecheckF(expr1, typ1, cont1, fb1), TypecheckF(expr2, typ2, cont2, fb2)) =>
+        expr1 ≟ expr2 && typ1 == typ2 && cont1 ≟ cont2 && fb1 ≟ fb2
```
drostron: Should `typ1 == typ2` be `typ1 ≟ typ2`?

sellout: We don’t have `Equal[Type]` yet, because it requires `Equal[Data]`, which requires `Equal` on various threeten data types, etc.

We should eventually be deriving these, but we’re not yet.

drostron: Ah, ok.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -137,6 +147,18 @@ object LogicalPlan {
       Fix[LogicalPlan](LetF(let, form, in))
   }
 
+  // NB: This should only be inserted by the type checker. In future, this
+  //     should only exist in SlamScript – the checker will annotate nodes
+  //     where runtime checks are necessary, then they will be added during
+  //     compilation to SlamScript.
```
wemrysi: Probably want to rename `SlamScript` to `BlackShield` or whatever the current name for this is.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -197,9 +205,145 @@ object LogicalPlan {
...
+  type SemDisj[A] = NonEmptyList[SemanticError] \/ A
+
+  def inferTypes(typ: Type, term: Fix[LogicalPlan]):
+      SemValidation[TypeFree[LogicalPlan]] =
+    (term.unFix match {
+      case ReadF(c)          => success(ReadF[TypeFree[LogicalPlan]](c))
+      case ConstantF(d)      => success(ConstantF[TypeFree[LogicalPlan]](d))
+      case InvokeF(f, args)  => for {
+        types <- f.untype(typ)
+        args0 <- types.zip(args).map((inferTypes(_, _)).tupled).sequenceU
```
wemrysi: Could simplify to `traverseU`

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -197,9 +205,145 @@ object LogicalPlan {
     case InvokeF(Distinct, List(src, _)) => src._2
     case InvokeF(DistinctBy, List(src, _)) => src._2
     case InvokeF(Squash, List(src)) => src._2
+    case TypecheckF(_, _, cont, _) => cont._2
     case _ => None
   }
 
+  type TypeFree[F[_]] = Cofree[F, Type]
```
wemrysi: Would `Typed` be a more evocative name here? `Typed[LogicalPlan]` reads nicer to me, does this complicate any semantics?

sellout: Yeah, I like that.

sellout: Especially because it’s something that’ll be exposed to the backends.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -197,9 +205,145 @@ object LogicalPlan {
...
+    if (inf.contains(poss))
+      emit((poss, Nil, poss match {
+        case Type.Const(d) => Constant(d)
+        case _ => term
+      }))
+    else if (poss.contains(inf)) {
+      emitName(freshName("check").map(name =>
+        (inf, List((name, inf, term)), Free(name))))
+    }
+    else lift(-\/(NonEmptyList(SemanticError.GenericError(s"couldn’t unify inferred (${inf}) and possible (${poss}) types in $term"))))
```
wemrysi: Might want to consider the postfix `left` (and `right`) operator in place of using the `\/` constructors for better inference, in general. Though it doesn't look like it would matter in this case. There is also the postfix `wrapNel` which can be a bit tidier:
```scala
SemanticError.GeneralError("...").wrapNel.left
```

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -197,9 +205,145 @@ object LogicalPlan {
     case InvokeF(Distinct, List(src, _)) => src._2
     case InvokeF(DistinctBy, List(src, _)) => src._2
     case InvokeF(Squash, List(src)) => src._2
+    case TypecheckF(_, _, cont, _) => cont._2
     case _ => None
   }
 
+  type TypeFree[F[_]] = Cofree[F, Type]
+  type NamedConstraint = (Symbol, Type, Fix[LogicalPlan])
+  type ConstrainedPlan = (Type, List[NamedConstraint], Fix[LogicalPlan])
```
wemrysi: Might want to consider lenses or case classes for these types as it would make the code that follows easier to read.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -197,9 +205,145 @@ object LogicalPlan {
...
+    quasar.namegen.lift[SemDisj](v)
+
+  /** This function compares the inferred (required) type with the possible type
+    * from the collection.
+    * • if it’s a const type, replace the node with a constant
+    * • if the possible is a subtype of the inferred, we’re good
+    * • if the inferred is a subtype of the possible, we need a runtime check
+    * • otherwise, we fail
+    */
+  private def maybeWrap(inf: Type, poss: Type, term: Fix[LogicalPlan]):
```
wemrysi: Can we come up with a better name for this? `attemptToUnify`? `unifyOrCheck`?

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -66,6 +67,38 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
...
+              generateTypeCheck(or)(f)(Type.AnyArray)
+            case Type.Timestamp
+               | Type.Coproduct(Type.Timestamp, Type.Date)
+               | Type.Coproduct(Type.Coproduct(Type.Timestamp, Type.Date), Type.Time) =>
+              generateTypeCheck(or)(f)(Type.Date)
+            case Type.Coproduct(Type.Coproduct(Type.Coproduct(Type.Timestamp, Type.Date), Type.Time), Type.Interval) =>
+              // Just repartition to match the right cases
+              generateTypeCheck(or)(f)(
+                Type.Coproduct(Type.Interval, Type.Date))
+            case Type.Coproduct(Type.Coproduct(Type.Coproduct(Type.Coproduct(Type.Coproduct(Type.Int, Type.Dec), Type.Interval), Type.Str), Type.Coproduct(Type.Coproduct(Type.Timestamp, Type.Date), Type.Time)), Type.Bool) =>
```
wemrysi: Adding a symbolic extractor for `Type.Coproduct` would reduce noise here.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -801,6 +915,90 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
       case FreeF(name) =>
         state(-\/(InternalError("variable " + name + " is unbound")))
       case LetF(_, _, in) => state(in.head._2)
+      case TypecheckF(expr, typ, cont, fallback) =>
+        // NB: Even if certain checks aren’t needed by ExprOps, we have to
+        //     maintain them because we may convert ExprOps to JS.
+        //     Hopefully SlamScript will eliminate the need for this.
```
wemrysi: Another `SlamScript` reference

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -801,6 +915,90 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
...
+        def exprCheck: Type => Option[Expression => Expression] =
```
wemrysi: What do you think about creating a type that represents these runtime typechecks along with how to "undo" them back into javascript? The undoing is dependent on the structure of the operations defined here and having them both in one place would make it easier to maintain. I'm thinking something that pairs the `Type => Option[Fix[ExprOp] => Fix[ExprOp]]` and `ExprOp[JsFn] => JsFn` functions together, or similar, for each kind of check.

core/src/main/scala/quasar/types.scala
```diff
@@ -476,44 +415,26 @@ final case object Type extends TypeInstances {
...
-  }
-
-  private def forall(expected: Type, actuals: Seq[Type]): ValidationNel[TypeError, Unit] = {
-    actuals.headOption match {
-      case Some(head) => typecheck(expected, head) +++ forall(expected, actuals.tail)
-      case None => Validation.success(())
-    }
-  }
+  private def checkAll(actuals: Seq[Type])(check: Type => ValidationNel[TypeError, Unit]) =
+    actuals.foldLeft[ValidationNel[TypeError, Unit]](
```
wemrysi: I think you can simplify this to `actuals foldMap check`.

it/src/main/resources/tests/time.test
```diff
@@ -8,7 +8,7 @@
 
   "predicate": "containsExactly",
   "expected": [
-      { "day": "Tuesday",   "tod": "08:00:00.000", "not a date": null, "missing": null },
-      { "day": "Wednesday", "tod": "09:00:00.000", "not a date": null, "missing": null },
-      { "day": "Thursday",  "tod": "10:00:00.000", "not a date": null, "missing": null }]
+      { "day": "Tuesday",   "tod": "08:00:00.000", "not a date": {"$na": null}, "missing": {"$na": null} },
```
wemrysi: Why does the output change from `null` to `{"$na": null}`?

sellout: `{"$na": null}` is our precise encoding of “undefined”. Previously, depending on where something failed in Mongo, we might end up with undefined or null. But Quasar’s semantics say a failed expression is undefined.

wemrysi: Ah, ok, so `{"$na": null}` is the JSON encoding of Quasar's `undefined`, not JavaScript's, is that correct?

core/src/main/scala/quasar/physical/mongodb/planner.scala
sellout: I’m sympathetic to this, but I’m not sure. Currently the typecheck planners for the three phases all parallel each other, and also the Expression ⇒ JsFn conversion is with other non-obvious conversions, so bringing them together in a new type would break those other connections.

Really, we just have to get away from converting Expressions to JS.

PR #958, opened 2015-10-14
SD-884: utilize http4s FileService

drostron:


web/src/main/scala/quasar/api/package.scala
```diff
@@ -126,17 +128,19 @@ package object api {
...
+    def pathCollector(file: File, config: Config, req: Request): Task[Option[Response]] = Task.now {
+      if (file.isDirectory) StaticFile.fromFile(new File(file, "index.html"), Some(req))
+      else if (!file.isFile) None
+      else StaticFile.fromFile(file, Some(req))
+    }
+
+    def apply(basePath: String): HttpService =
+      fileService(Config(
+        systemPath = basePath,
+        pathCollector = pathCollector))
```
sellout: This looks good – but why make it an object+apply rather than a method?

drostron:
Just noticed I submitted my wip branch again. I even created a ready branch this time before submitting. I'll retarget after discussion is finished.

web/src/main/scala/quasar/api/package.scala
```diff
@@ -126,17 +128,19 @@ package object api {
...
-        StaticFile.fromString(filePath + "/index.html").fold(
-          NotFound("Couldn’t find page " + path.toString))(
-          Task.now))(
-        resp => path.file.flatMap(f => fileMediaType(f.value)).fold(
-          Task.now(resp))(
-          mt => Task.delay(resp.withContentType(Some(`Content-Type`(mt))))))
+  object staticFileService {
+    import FileService.Config
+
+    def pathCollector(file: File, config: Config, req: Request): Task[Option[Response]] = Task.now {
```
wemrysi: Is `Task.delay` more appropriate here? The docs for `File#isDirectory` and `File#isFile` seem to imply they aren't pure, though I'm not exactly sure when the existence check is actually performed.

jdegoes: Definitely not pure.

drostron: I was following http4s as in: https://github.com/http4s/http4s/blob/master/server/src/main/scala/org/http4s/server/staticcontent/FileService.scala#L49. Happy to update if there are thoughts on an improved implementation.

sellout: Yeah, I missed this, but we should almost never use `Task.now`, regardless of whether we think it might be pure. The exception is primitive values … often `()`.

web/src/main/scala/quasar/api/package.scala
drostron: If we wanted to test `pathCollector` it would be available, but it seems like having it within the outer method would be sufficient here. I'd be interested in learning what others think on the merits of the two approaches.

web/src/main/scala/quasar/api/package.scala
drostron: I've now familiarized with `Task.now` and `Task.delay` and agree the `Task.delay` is what we want here.

drostron:
superseded by #961

PR #959, opened 2015-10-14
Restrict coverage to core subproject.

sellout:
**DO NOT MERGE**

Hopefully stop Codecov.io from getting confused about what to process.


.travis.yml
```diff
@@ -9,10 +9,9 @@ before_install:
 script: "./scripts/build"
 after_success:
 - "./scripts/publishJarIfMaster"
-# Upload coverage information to coveralls.io
-- "./sbt 'project core' coveralls"
-# Upload coverage information to codecov.io (BETA)
-# Unsure about whether it will do coverage for all the code base as opposed to just core
+- "./sbt coverageAggregate"
+- rm -rf core/target it/target web/target
```
jedesah: Can we put a comment here explaining why we need to delete target folders? I only know why because you told me, and there is no way I would have guessed it.

jedesah:
Just want to make sure I understand. The current title for this PR is incorrect now right? This actually enables coverage for the whole project.

.travis.yml
sellout: We’re not going to delete target folders – it didn’t work.

sellout:
Yeah, after the first attempt failed, this basically became a testing ground for how to fix coverage. This PR is far too messy to be merged now, and still hasn’t solved the problem. But I have talked to @codecov-io on their Gitter, and have some more stuff to try at some point.

Then, once it works, nice new PR.

jedesah:
:+1:

PR #960, opened 2015-10-14
[SD-919] Error out in a consistent way in the face of a missing collection

jedesah:
authored with @drostron


core/src/main/scala/quasar/physical/mongodb/collection.scala
```diff
@@ -126,10 +126,9 @@ object Collection {
 
     def char: Parser[String] = substitute(CollectionNameEscapes) | "(?s).".r
 
-    def apply(input: String): PathError \/ String = parseAll(seg, input) match {
-      case Success(seg, _) => \/-(seg)
-      case failure : NoSuccess =>
-        -\/(InvalidPathError("failed to parse ‘" + input + "’: " + failure.msg))
+    def apply(input: String): String = parseAll(seg, input) match {
+      case Success(seg, _) => seg
+      case failure : NoSuccess => scala.sys.error("doesn't happen")
```
jdegoes: `sys.error` is not allowed in code anymore (there was a time :cry: )— all failures should be captured in the return type.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -190,7 +191,7 @@ trait MongoDbEvaluatorImpl[F[_]] {
           emit(fail(UnableToStore("MongoDB cannot store anything except documents inside collections: " + v)))
 
         case ReadTask(value) =>
-          emit((Col.User(value): Col).point[F])
+          emit(executor.exists(value).map(_ => Col.User(value): Col))
```
jdegoes: More idiomatic:

`executor.exists(value) *> (Col.User(value): Col).point[F]`

core/src/main/scala/quasar/physical/mongodb/collection.scala
jedesah: Even for an assertion/assumption? This file has a bunch of them already?

jdegoes: If I had a dollar for every time I've heard a programmer say, "That's impossible," well, I'd be self-funding SlamData. :wink:

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -332,6 +333,15 @@ class MongoDbExecutor[S](client: MongoClient,
...
 
+  override def exists(coll: Collection): M[Unit] = {
+    import quasar.fs.Path.PathError.NonexistentPathError
+    import quasar.fs.Path
+    MongoWrapper.liftTask(MongoWrapper(client).exists(coll)).liftM[MT].flatMap {
+      case true => MongoWrapper.liftTask(Task.now(())).liftM[MT]
+      case false => fail(EvalPathError(NonexistentPathError(coll.asPath.asAbsolute, None)))
+    }
+  }
+
```
jdegoes: Consider factoring out some duplication & cleaning up:

```scala
// Can be used elsewhere in the file:
def errM[A](taskErr: ???[A]): M[A] = MongoWrapper.liftTask(taskErr).liftM[MT] 

for {
  exists <- errM(MongoWrapper(client).exists(coll))
  v      <- if (exists) then ().point[M] else fail(...)
} yield v
```

If the `().point[M]` logic is used more than once, you could also consider introducing:

```scala
def succeed: M[Unit] = ().point[M]
```

to make the code:

```scala
for {
  exists <- errM(MongoWrapper(client).exists(coll))
  v      <- if (exists) then succeed else fail(??)
} yield v
```

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -385,6 +395,9 @@ class JSExecutor[F[_]](nameGen: NameGenerator[F])(implicit mf: Monad[F])
 
   val defaultWritableDB = Some("default")
 
+  // Not sure what should be done here...
+  def exists(coll: Collection): LoggerT[F]#Rec[Unit] = succeed(None, ().point[F])
```
jdegoes: Someone who knows something should clarify what should be done here so we can delete the comment. :smile:

it/src/main/scala/quasar/Interactive.scala
```diff
@@ -85,15 +86,27 @@ package object interactive {
...
+  }
+
+  /**
+   * Execute an SQL query on the backend and return the result as a stream.
+   * @param backend The `Backend` on which to run this query.
+   * @param query The SQL query to evaluate
+   * @return A Stream of `Data` representing the result of the query along with the [[Vector]] of [[PhaseResult]]
+   * @throws Exception if the query [[String]] cannot be parsed
+   */
+  def evalLog(backend: Backend, query: String): EitherT[(Vector[PhaseResult], ?), CompilationError, Process[ProcessingTask, Data]] = {
```
jdegoes: :-1:  This code has gone from not throwing an exception, to throwing an exception.

it/src/test/scala/quasar/ErrorConditions.scala
```diff
@@ -0,0 +1,39 @@
...
+
+class ErrorConditions extends BackendTest with NoTimeConversions with DisjunctionMatchers {
+
+  backendShould { (prefix, backend, backendName) =>
+    "Backend" should {
+      "error out consistently" in {
+        def testQueryOnMissingCollection(produceQueryFromCollectionPath: String => String) = {
+          val missingCollectionPath = prefix ++ Path("IDoNotExist")
+          backend.exists(missingCollectionPath).run.run.toOption.get should beFalse
+          println(missingCollectionPath.simplePathname)
```
jdegoes: Debug code.

jdegoes:
@sellout You're still on the hook for reviewing, btw. :smile:

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
wemrysi: Or even `executor.exists(value).as(Col.User(value))`

jdegoes: :+1:

core/src/main/scala/quasar/physical/mongodb/collection.scala
jedesah: If I remove all the `sys.error`s in this file, `asPath` will have to become a `Disjunction` and that will be a huge change...not to mention out of scope for this PR. I will just revert the one I changed even though that creates inconsistency in the file. The ideal solution would a be a parser library that has a notion of a partial or total parser and would provide the correct return type accordingly. I tend to dislike inconsistency and think it's reasonable to error out here due to the shortcomings of this parser library (and the Scala language really), but I won't argue any further...

it/src/main/scala/quasar/Interactive.scala
jedesah: It's test code. I thought we said that throwing exceptions in test code was acceptable. It's certainly more convenient to use and not sure what draw back it has.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
jedesah: @sellout For the `MongoDBExecutor` I can make `exists` return `F[Boolean]`. But you had mentioned that for bringing the `JSExecutor` into line one day, we might be forced to make it `F[Unit]`. Thoughts?

jedesah: @sellout Disregard previous email, was confused I think.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -332,9 +333,20 @@ class MongoDbExecutor[S](client: MongoClient,
...
+    }
+  }
+
   private def mongoCol(col: Collection) =
     client.getDatabase(col.databaseName).getCollection(col.collectionName)
 
+  private def liftTask[A](a: Task[A]): M[A] =
+    MongoWrapper.liftTask(a).liftM[MT]
+
   private def liftMongo[A](a: => A): M[A] =
```
jedesah: Kinda wanted to call `liftTask` from `liftMongo` but oh well...

core/src/main/scala/quasar/physical/mongodb/MongoWrapper.scala
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  def exists(path: Collection): Task[Boolean] = for {
+    names <- MongoWrapper.databaseNamesIterator(client)
+    databaseExists = names.exists(name => name == path.databaseName)
+    collectionNames = MongoWrapper.collectionNamesIterator(path.databaseName, client)
+    collectionExists = collectionNames.map(_.exists(name => name == path.collectionName))
+    exists <- if (databaseExists) collectionExists
+              else Task.now(false)
+  } yield exists
```
sellout: I think this is legitimately causing the tests to fail – loading a config must do a ReadTask somewhere?

It does seem possible that a user will want to execute a query on a cluster where they don’t have permissions to `listDatabases`. Rather than using `databaseNamesIterator`, we could just handle failure of `collectionNamesIterator` – if the DB doesn’t exist, then we know the collection doesn’t exist.

And if `collectionNamesIterator` fails because of authorization (is this possible in the Mongo security model?), then we should perhaps treat an error as “success”, and make a note that we should emit a warning that we might not be able to identify missing collections.

And finally, if it’s possible for exists to fail with some “I don’t know” state, we should handle the “no such ns” error on `MapReduceTask`. In fact, we probably want to do that anyway, just to try to catch some race conditions.

wemrysi: While not a complete solution, the current implementation in master attempts to use any authentication databases as a fallback if `listDatabases` fails. This helps a bit because for `listDatabases` to fail due to a permissions issue, you know that the connection uses authentication and, in the common case, we can extract a db name from that which is likely, though not guaranteed, to be useful.

Eventually, I think we're going to have to add the option to declaratively specify databases for a mongo mount, for these scenarios. Especially if an authenticated user has access to more than one database as you're only able to authenticate to one (and thus only able to deduce one from credentials info).

wemrysi: > It does seem possible that a user will want to execute a query on a cluster where they don’t have permissions to listDatabases. Rather than using databaseNamesIterator, we could just handle failure of collectionNamesIterator – if the DB doesn’t exist, then we know the collection doesn’t exist.

This will likely be the common case when it comes to any cloud hosted MongoDB.

sellout: Well, in this case, we already have the name of the database. We just need to determine if that name refers to an actual DB. So, I think `client.getDatabase(dbName)` and handling any error is the way to go.

Including the error here, rather than having to find it in the build every time:
```
[error]    '-\/(EnvEvalError(CommandFailed(Command failed with error 13: 'not authorized on admin to execute command { listDatabases: 1 }' on server ds045089.mongolab.com:45089. The full response is { "ok" : 0.0, "errmsg" : "not authorized on admin to execute command { listDatabases: 1 }", "code" : 13 })))' is not \/- with value'Config(SDServerConfig(Some(92)),Map(/ -> MongoDbConfig(mongodb://slamengine:slamengine@ds045089.mongolab.com:45089/slamengine-test-01)))' (config.scala:173)
```

`databaseNamesIterator` does still seem to handle `MongoCommandException`, but we end up with a `CommandFailed`, so I’m not sure what kind of exception is being raised, exactly.

wemrysi: > databaseNamesIterator does still seem to handle MongoCommandException

Ah, indeed it does, I overlooked that, sorry for the red-herring. It is odd the exception isn't caught, though.

> So, I think client.getDatabase(dbName) and handling any error is the way to go.

:+1:

jedesah:
I was wondering why I hadn't seen this failure when developing. It's because I assumed all tests that actually hit MongoDB were in it, but this Config test in core also hits a cloud MongoDB.

it/src/main/scala/quasar/Interactive.scala
wemrysi: I'm not sure it is more convenient from the perspective of a caller of the function as, in addition to handling `CompilationError` and possible `ProcessingError`s (which are reflected in the return type), one also has to know to catch ambient exceptions. I don't think this method should throw.

core/src/main/scala/quasar/physical/mongodb/MongoWrapper.scala
wemrysi: @jedesah Are you planning on updating the implementation of `exists` as @sellout suggested?

it/src/main/scala/quasar/Interactive.scala
jedesah: The intent is that test code and scripts (which is what interactive is meant to stand for) do not want to have to "handle" all the things that can go wrong. If something is unexpected, it throws and tests fail.

Granted the ideal solution would be a macro based string interpolator that could validate at compile time that the string in the test in indeed valid SQL and type it as such. Hopefully when we rewrite the parser we can add that, although it seems unlikely at this point. In the meantime, this seems like the better solution to me.

core/src/main/scala/quasar/physical/mongodb/MongoWrapper.scala
jedesah: In the process of doing so, but running into some weird behavior.

jedesah:
The build failure was actually not being caused by a call to `exists`. It was being caused by the innocuous looking change that separated the `databaseNames` and `collections` functions into versions that returned iterators instead of immutable `List` or `Set`. The MongoDB driver was only throwing an exception when first using the `iterator` which was outside of the try-catch block.

Anyway, updated `exists` to access the database directly, which removed the need for those Iterable versions. This should fix the build issue and generally provide the behavior we are looking for.

Still looking into whether we should catch the CommandFailed Exception and how to do that (need to familiarise myself with `StateT`). I think we would have to do a string matching in order to catch that specific exception which seems rather sketchy.

sellout:
Ah yeah – it’s a problem that that test hits MongoDB. I think it used to not, because we didn’t try to validate a config beyond well-formedness. But now we make sure we can connect.

As far as handling the `Throwable`, in a couple places we do this terrible thing, (`CommandFailed(e.getMessage)`) that just turns any `Throwable` into an `EvaluationError`. Hopefully, Mongo is giving us reasonably distinct exceptions that we are unreasonably mashing together.

Really, `CommandFailed` should probably go away in favor of sharper tools.

jedesah:
@sellout As you suggested, I (with the help of @wemrysi) added logic to catch the `CommandFailed` error. But now I have to craft a way to hit that code (for coverage) even though we guard against it!! :rage:

sellout:
You just need a free algebra that deletes the collection after an exists call 😉 … or something.

At some point we should get rid of string matching on errors … or at least push it down to where we reify the Throwable.

jedesah:
@sellout Kind of cheated by adding some trivial coverage for Json Codecs because hitting the "after" check code would require dealing with a lot of stuff that is going to change with the Free Architecture and I would rather do that when that is ready.

PR #963, opened 2015-10-15
[SD-961] Support uploading a file containing a single json Array

jedesah:


sellout:
Do we want to care at all about Content-Type here? Seems reasonable enough to just try parsing things until something works, but figured it should at least be mentioned, since we do define Content-Types for lines vs array.

mossprescott:
That's a good question. We do use `Content-Type` to recognize CSV uploads, so I think it makes sense to require it for JSON as well. Also there's the Readable/Precise distinction, which cannot be inferred.

I don't think it's safe to infer the style, because other backends will likely support storing arrays as root values (unlike MongoDB). In any case, I think we could use some more test coverage (e.g. error cases such as uploading multiple lines containing arrays).

jedesah:
> We do use Content-Type to recognize CSV uploads

@mossprescott Where do we do that?

jdegoes:
> I don't think it's safe to infer the style, because other backends will likely support storing arrays as root values (unlike MongoDB).

This is correct. It is *not* safe to infer the style and content types and dispositions (including custom settings) should be used as appropriate.

jedesah:
I think I have addressed all the issues brought forward. Please do pay close attention to the specification around accept valid json to make sure the specification reflects desired behavior.

jedesah:
Is it possible that coveralls compares the coverage level to the commit on which the PR is based instead of master? Before merging with master, I was getting a whole bunch of noise in the coveralls coverage report (https://coveralls.io/builds/3878619), but now it appears to be fine.

jedesah:
Assigning to @wemrysi because @mossprescott is out for the first part of this week.

jedesah:
I think that possibly codecov is doing a better job of this by comparing the merge commit to master.

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -0,0 +1,121 @@
...
+
+sealed trait MessageFormat {
+  def mediaType: MediaType
+  def disposition: Option[`Content-Disposition`]
+}
+sealed trait JsonFormat extends MessageFormat {
+  def codec: DataCodec
+}
+object MessageFormat {
+  final case class JsonStream private[MessageFormat](codec: DataCodec, mode: String, disposition: Option[`Content-Disposition`]) extends JsonFormat {
```
wemrysi: What do you think about introducing an ADT to encode the `Readable`/`Precise` distinction, the cases could each contain the appropriate `DataCodec` itself, i.e.

```scala
sealed trait JsonMode {
  def codec: DataCodec
}
case object Readable {
  val codec 
final case class Precise(codec: DataCodec) extends JsonMode
```
this helps to keep the codec and mode related and also allows us to avoid string matching

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -0,0 +1,121 @@
...
+    (for {
+      acc       <- accept
+      // TODO: MediaRange needs an Order instance – combining QValue ordering
+      //       with specificity (EG, application/json sorts before
+      //       application/* if they have the same q-value).
+      mediaType <- acc.values.sortBy(_.qValue).list.find(a => mediaTypes.list.exists(a.satisfies(_)))
+    } yield {
+      import org.http4s.parser.HttpHeaderParser.parseHeader
+      val disposition = mediaType.extensions.get("disposition").flatMap { str =>
+        parseHeader(Header.Raw(CaseInsensitiveString("Content-Disposition"), str)).toOption.map(_.asInstanceOf[`Content-Disposition`])
```
wemrysi: Can we pattern match on the result instead of resorting to `asInstanceOf`?

web/src/main/scala/quasar/api/MessageFormat.scala
wemrysi: I guess this actually wouldn't help us avoid string matching, now that I reread `fromAccept`, but I still think its worth doing.

web/src/main/scala/quasar/api/MessageFormat.scala
jdegoes: :+1: `asInstanceOf` and `instanceOf` should not be used in production code.

jedesah: My ego wants me to highlight the fact that I did not introduce this `asInstanceOf` but rather it shows up as added simply because I extracted `MessageFormat` into it's own file! :stuck_out_tongue: 

But I am happy to report that I managed to get rid of it without resorting to a pattern match! :smiley:

wemrysi: I did realize that after reading through the rest of the file, thanks for tidying up anyways ;) I like the solution!

PR #965, opened 2015-10-22
SD-1068 : update scaladoc to build cleanly

drostron:
The `doc` task was failing and therefore `publishLocal` was failing. This fixes the `doc` failures.

### DIFF
--- a/core/src/main/scala/quasar/fp/package.scala
+++ b/core/src/main/scala/quasar/fp/package.scala
@@ -339,7 +339,7 @@ package object fp extends TreeInstances with ListMapInstances with ToCatchableOp
     def apply[A](io: IO[A]): Task[A] = Task.delay(io.unsafePerformIO())
   }
 
-  /** Wrapper around [[IORef]] to operate in [[Task]] */
+  /** Wrapper around `IORef` to operate in `Task` */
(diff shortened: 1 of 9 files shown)


core/src/main/scala/quasar/fp/package.scala
```diff
@@ -339,7 +339,7 @@ package object fp extends TreeInstances with ListMapInstances with ToCatchableOp
     def apply[A](io: IO[A]): Task[A] = Task.delay(io.unsafePerformIO())
   }
 
-  /** Wrapper around [[IORef]] to operate in [[Task]] */
+  /** Wrapper around `IORef` to operate in `Task` */
```
drostron: I was unable to get these links recognized via autoAPIMappings even with full reference paths.

it/src/main/scala/quasar/Interactive.scala
```diff
@@ -143,7 +143,7 @@ package object interactive {
 
   /**
    * Loads a collection of data into the provided backend if not already there
-   * Same as [[loadData]] but the source description is used for the collection name.
+   * Same as `loadData` but the source description is used for the collection name.
```
drostron: I was unable to get link disambiguation working here.

PR #966, opened 2015-10-26
new "view" mount type

mossprescott:
Rename `BackendConfig` to `MountConfig`, since a view is a mount, but not a backend. New subtype, `ViewConfig`, holding a (parsed) query and optional temporary path. The codec translates between that and the URI syntax, which looks like `sql2:///foo/bar?q=select+*+...` (note: always three slashes, whether or not there is a temp path present, because a name after the second slash is interpreted as a "host" by standard parsers.)

The `/mount` API now handles reading, validating, and writing view mounts and will persist them in the config file, but they are otherwise ignored, pending SD-977 and SD-978.

A fairly dumb generator for configurations, and a simple test of round-trip encoding/decoding.


mossprescott:
Review by @wemrysi (I can't seem to assign the PR to you at the moment).

core/src/main/scala/quasar/config/config.scala
```diff
@@ -70,28 +70,70 @@ object MongoDbConfig {
   implicit def Codec = casecodec1(MongoDbConfig.apply, MongoDbConfig.unapply)("connectionUri")
 }
 
-object BackendConfig {
-  implicit def BackendConfig = CodecJson[BackendConfig](
+final case class ViewConfig(tempPath: Option[EnginePath], query: sql.Expr) extends MountConfig {
```
wemrysi: Does it make sense to also include the `String` version of the query to be able to reparse it in the case that `Expr` changes?

wemrysi: Ah, nevermind, it is being serialized using the pretty-printer, carry on ;)

core/src/main/scala/quasar/config/config.scala
```diff
@@ -70,28 +70,70 @@ object MongoDbConfig {
...
+  def validate(path: EnginePath) = for {
+    _ <- if (path.relative) -\/(InvalidConfig("Not an absolute path: " + path)) else \/-(())
+    _ <- if (path.pureDir) -\/(InvalidConfig("Not a file path: " + path)) else \/-(())
+  } yield ()
+}
+object ViewConfig {
+  private val UriPattern = "([a-z][a-z0-9+-.]*):///([^?]*)\\?q=(.+)".r
```
wemrysi: Is it possible to use `java.net.URI` for parsing?

mossprescott: I tried it and found it wasn't that helpful because it doesn't parse out query parameters at all (you just get `"q=%26select+..."`), so I was going to have to deal with that myself (and eventually additional params), and just ended up skipping it. What you need is something like a HTTP-URL parser. I even looked at http4s's parser, but it seemed like too much trouble to be worthwhile.

wemrysi:
More generally, what is the benefit of merging this now, with stubbed implementations and `TODO`s, vs just waiting to include the changes in this PR with the rest of the implementation?

mossprescott:
That's a fair question. In fact, I held back on updating the README because this commit provides essentially no useful functionality. The only reason to merge it as is would be for the satisfaction of closing [SD-977](https://slamdata.atlassian.net/browse/SD-977) (a sub-task).

If you think it makes more sense, I'll close this PR and submit one soon with another commit closing 976, which will eliminate the TODOs and be useful enough to document.

wemrysi:
I think I'd prefer that, if you don't mind, just to try and keep master as tidy/release-ready as possible (not that this commit is necessarily harmful).

mossprescott:
New PR coming up soon.

PR #967, opened 2015-10-27
[SD-1068] preparation for external configuration extension

drostron:


core/src/main/scala/quasar/config/config.scala
```diff
@@ -185,13 +210,63 @@ object Config {
...
+    implicit val CC = _CC
+    implicit val LGC = _LGC
+    implicit val LGS = _LGS
+    implicit val SM = _SM
+    implicit val SS = _SS
+    implicit val SP = _SP
+    implicit val US = _US
+    implicit val UM = _UM
+    implicit val UP = _UP
+    }
```
mossprescott: Wow.

There must be an easier way to get some genericity in the parsing of the config file, which is all this is really accomplishing, right? It's cool that shapeless can do this, but this is not something I ever want to see or have to fix. Even worse, some of it leaks into the rest of the code base, as seen in Mounter, etc.

I could imagine breaking down some of the config-file handling functions so that they can operate on any config type given a suitable encoder (while providing the "standard" type and its encoder). I don't suppose that deals with the problem of updating configuration in the API, so maybe that's how you ended up here.

I don't have alternative solution to offer, but just felt compelled to speak for the poor future programmer who has to make sense of this.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -185,13 +210,63 @@ object Config {
...
 
-  implicit val ShowConfig = new Show[Config] {
-    override def shows(f: Config) = Config.toString(f)
-  }
+  val ev = Ev.construct(Empty[Config].empty, Empty[SDServerConfig].empty)
 }
+
+trait Ev[C, S, LC <: HList, LS <: HList] {
```
mossprescott: If we have to have something like this, I really think we need documentation of `Ev` and all of the type parameters, which I can only figure out by reading the shapeless types that follow, which I can't just barely read.

core/src/main/scala/quasar/generic/package.scala
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar
+
+import shapeless._, labelled._, record._, ops.record._
+
+package object generic {
```
mossprescott: [Ack!](https://en.wikipedia.org/wiki/Bill_the_Cat)

What is this for?! Javadoc, paper references, and/or a PR to shapeless, please!

core/src/main/scala/quasar/config/config.scala
jdegoes: This was also my reaction. Seems like the cure is worse than the disease. :smile:

drostron: I pushed a commit addressing some of the noise but the original concerns will likely remain.

core/src/main/scala/quasar/config/config.scala
drostron: Agreed, documenting this would be useful.

core/src/main/scala/quasar/generic/package.scala
drostron: It is a touch of sugar to reduce code that was repeated in many places. I originally named the package `shapeless` but the import story ended up messy. I'm open to ideas around these helper methods.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -185,13 +210,35 @@ object Config {
...
+  val LGC: LabelledGeneric.Aux[C, LC],
+  val LGS: LabelledGeneric.Aux[S, LS],
+  val SM: Selector.Aux[LC, Ev.mountingsWitness.T, Map[EnginePath, BackendConfig]],
+  val SS: Selector.Aux[LC, Ev.serverWitness.T, S],
+  val SP: Selector.Aux[LS, Ev.portWitness.T, Int],
+  val US: Updater.Aux[LC, FieldType[Ev.serverWitness.T, S], LC],
+  val UM: Updater.Aux[LC, FieldType[Ev.mountingsWitness.T, Map[EnginePath, BackendConfig]], LC],
+  val UP: Updater.Aux[LS, FieldType[Ev.portWitness.T, Int], LS])
+
+object configConfOps extends ConfigOps(Ev.ev)
```
sellout: Maybe rename this to `BasicConfig`?

drostron: Sure, additionally I think there's value in separating the Server config from Config and moving it to web as we discussed. I might take a quick pass at it for this PR assuming it doesn't end up taking too much time.

drostron:
superseded by #975

PR #968, opened 2015-10-27
Implement case insensitive (and negated) regex.

sellout:
Also bump minimum scoverage req to 80%.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/planner.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/planner.scala
@@ -112,8 +112,15 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
     def invoke(func: Func, args: List[Output]): Output = {
       val HasStr: Output => OutputM[String] = _.flatMap {
         _._1(Nil)(ident("_")) match {
-          case Literal(Js.Str(str)) => \/-(str)
-          case x => -\/(FuncApply(func, "JS string", x.toString))
+          case Literal(Js.Str(str)) => str.right
+          case x => FuncApply(func, "JS string", x.toString).left
+        }
+      }
+
+      val HasBool: Output => OutputM[Boolean] = _.flatMap {
+        _._1(Nil)(ident("_")) match {
+          case Literal(Js.Bool(b)) => b.right
+          case x => FuncApply(func, "JS boolean", x.toString).left
         }
       }
 
@@ -203,9 +210,23 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
           Arity3((field, start, len) =>
             Call(Select(field, "substr"), List(start, len)))
         case Search =>
-          Arity2((field, pattern) =>
-            Call(Select(New(Name("RegExp"), List(pattern)), "test"),
-              List(field)))
+          args match {
+            case List(a1, a2, a3) => (HasJs(a1) |@| HasJs(a2) |@| HasBool(a3)) {
+              case ((fieldF, fieldP), (patternF, patternP), insen) =>
+                ({ case list => JsFn(JsFn.defaultName, Call(
+                  Select(
+                    New(Name("RegExp"),
+                      if (insen)
+                        List(patternF(list.drop(fieldP.size))(Ident(JsFn.defaultName)), Literal(Js.Str("i")))
+                      else
+                        List(patternF(list.drop(fieldP.size))(Ident(JsFn.defaultName)))),
+                    "test"),
+                  List(fieldF(list.take(fieldP.size))(Ident(JsFn.defaultName)))))
+                },
+                  fieldP.map(there(0, _)) ++ patternP.map(there(1, _)))
+            }
+            case _               => -\/(FuncArity(func, args.length))
+          }
         case Extract =>
           args match {
           case a1 :: a2 :: Nil => (HasStr(a1) |@| HasJs(a2)) {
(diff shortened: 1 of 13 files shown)


jonsterling:
Heh, thanks for getting this in!

mossprescott:
We're adding three new, unguessable operators (`~*`, `!~`, and `!~*`), which are at least common with Postgres, and we're adding a new argument to our own `SEARCH` function. Where is this query language documented? For one thing, the cheatsheet should get updated (see https://github.com/slamdata/slamdata.github.io/blob/master/support/cheatsheet.pdf and https://slamdata.atlassian.net/browse/SD-1075).

One possible answer would be to punt that to SQL^2 (and maybe write an issue for creating that documentation).

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -203,9 +210,23 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
           Arity3((field, start, len) =>
             Call(Select(field, "substr"), List(start, len)))
         case Search =>
-          Arity2((field, pattern) =>
-            Call(Select(New(Name("RegExp"), List(pattern)), "test"),
-              List(field)))
+          args match {
+            case List(a1, a2, a3) => (HasJs(a1) |@| HasJs(a2) |@| HasBool(a3)) {
```
mossprescott: Losing me here. Can't this just be `List(HasJs(a1), ...)` or `HasJs(a1) :: ... :: Nil`?

mossprescott: More importantly, since we don't seem to require the pattern to be a literal, why require the case-insensitivity flag to be a literal? It seems like we should avoid that kind of restriction in general if we can.

sellout: Nope, this isn’t `workflowƒ`. And `Has*` aren’t extractors in `jsExprƒ`.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -496,9 +526,7 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
         case (And, _)      => invoke2Nel(Selector.And.apply _)
         case (Or, _)       => invoke2Nel(Selector.Or.apply _)
         case (Not, (_, v) :: Nil) =>
-          v.map { case (sel, loc) =>
-            (sel andThen (s => s.negate)) -> loc.map(there(0, _))
-          }
+          v.map(_.bimap(_ andThen (_.negate), _.map(there(0, _))))
```
mossprescott: Thanks for cleaning up my mess :)

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -545,13 +573,16 @@ object MongoDbPlanner extends Planner[Crystallized] with Conversions {
...
         selCheck(typ).fold[OutputM[PartialSelector[B]]](
           -\/(UnsupportedPlan(node, None)))(
-          f =>
-          \/-(cont._2.fold[PartialSelector[B]](
-            κ(({ case List(field) => f(field) }, List(there(0, here)))),
-            { case (f2, p2) =>
-              ({ case head :: tail => Selector.And(f(head), f2(tail)) },
-                there[B](0, here) :: p2.map(there(1, _)))
-            })))
+          f => cont._2.flatMap { case (f2, p2) =>
```
mossprescott: How does this relate to the rest of the changes?

core/src/main/scala/quasar/physical/mongodb/planner.scala
sellout: That’s a good point. I was was mimicking what I did for Selectors, but clearly we have more flexibility in JS.

core/src/main/scala/quasar/std/string.scala
```diff
@@ -56,18 +56,19 @@ trait StringLib extends Library {
     },
     basicUntyper)
 
-  def matchAnywhere(str: String, pattern: String) = java.util.regex.Pattern.compile(pattern).matcher(str).find()
+  def matchAnywhere(str: String, pattern: String, insen: Boolean) =
+    java.util.regex.Pattern.compile(if (insen) "(?i)" ⊹ pattern else pattern).matcher(str).find()
 
   val Search = Mapping(
     "search",
     "Determines if a string value matches a regular expresssion.",
```
mossprescott: This is probably the place to include some text about the parameters.

mossprescott:
I think we're missing some integration tests:
* One or two for `!~` and `!~*` for completeness (the point being that every feature is represented by a test at that level, for documentation and future backend verification).
* One for `~*` used in non-selector position, as in `select city ~* 'bou.*, city from zips'. Can I assume we have a corresponding test for `~`?

core/src/main/scala/quasar/physical/mongodb/planner.scala
sellout: This was an accidental commit – should be in a separate PR.

it/src/main/resources/tests/guardedExpression.test
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "regex on non-string field",
+  "data": "zips.data",
```
mossprescott: Here's where you broke the build. :-)

sellout:
@mossprescott Updated (since you get no other notification).

mossprescott:
Looking good now, and I like the cleanup of `jsExpr`. :+1:

PR #969, opened 2015-10-28
Sort out codecov coverage.

sellout:
DO NOT MERGE

Hopefully stop Codecov.io from getting confused about what to process.


wemrysi:
Do we still need this PR around? Some similar changes exist in #986, along with many others to enable parallel builds, dropping support for coveralls, etc.

sellout:
I think the only thing here that needs to be persisted (and may already be in #986?) is moving the scoverage highlighting from core/build.sbt to build.sbt.

PR #971, opened 2015-10-29
Upgrading WartRemover to 0.14.

sellout:
New warts:

* Enumeration,
* ExplicitImplicitTypes (disabled due to mpilquist/simulacrum#35),
* Option2Iterable, and
* ToString (currently disabled).

Also, enable NoNeedForMonad, because the false positives have been fixed.


sellout:
I punted on fixing all the ToString failures. Got the easy ones, but we really want to move to Show (preferably with Shapeless deriving), and there were just too many walls I was running into along the way.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -63,10 +63,10 @@ trait Compiler[F[_]] {
...
 
   private object CompilerState {
-    /**
-     * Runs a computation inside a table context, which contains compilation
-     * data for the tables in scope.
-     */
+    /** Runs a computation inside a table context, which contains compilation
+      * data for the tables in scope.
+      */
+    @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NoNeedForMonad"))
```
wemrysi: If we want to avoid the supress

```scala
mod((s: CompilerState) => s.copy(tableContext = t :: s.tableContext)) *>
f                                                                     <*
mod((s: CompilerState) => s.copy(tableContext = s.tableContext.drop(1)))
```
We can also create `scalaz.Lens` for `tableContext` which have nice conversions to `State` actions and would make this less verbose.

sellout: Ah, of course.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -115,9 +115,8 @@ trait Compiler[F[_]] {
         case None    => fail(CompiledTableMissing)
       }
 
-    /**
-     * Generates a fresh name for use as an identifier, e.g. tmp321.
-     */
+    /** Generates a fresh name for use as an identifier, e.g. tmp321. */
+    @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NoNeedForMonad"))
```
wemrysi: Similarly

```scala
read[CompilerState, Int](_.nameGen).map(n => prefix + n.toString) <*
mod((s: CompilerState) => s.copy(nameGen = s.nameGen + 1))
```

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -317,6 +318,7 @@ class MongoDbExecutor[S](client: MongoClient,
   def fail[A](e: EvaluationError): M[A] =
     MonadError[ETask, EvaluationError].raiseError[A](e).liftM[MT]
 
+  @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NoNeedForMonad"))
```
wemrysi: Interesting, what was it complaining about here? The `flatMapF`?

sellout: No, it was in the expression I changed – I think the `flatMap` down there. But I’m not sure why. Seemed necessary to me. False positive?

wemrysi: Huh, so moving the `map` inside the `liftTask` made it go away, that's odd. Might be a false-positive, or might it have had something to do with the `Option` -> `Set` conversion?

sellout: No, moving the `map` didn’t` make it go away, that’s why I needed the Suppress. I _think_ the error was on the flatMap, so it shouldn’t have involved the Set, it would have been using the Monad of EvaluationTask.

I could see if it were `κ(attemptVersion)`, then we should use `*>`, so maybe there’s _something_ in one of those inner functions that’s triggering it.

wemrysi: > No, moving the map didn’t` make it go away, that’s why I needed the Suppress

Haha, silly me, of course it didn't. Anyhow, yeah I don't see why it thinks the `flatMap` isn't necessary.

sellout:
So, Travis failed in what looks like a case of puffnfresh/wartremover#159. But it didn’t fail on my machine (although I should try building the branch again).

PR #972, opened 2015-10-29
Fix a bug in converting LIKE patterns to regexes.

sellout:
This is probably not a full fix to SD-1081, but this bug would be hit
once the other bug gets fixed.

### DIFF
--- a/core/src/main/scala/quasar/compiler.scala
+++ b/core/src/main/scala/quasar/compiler.scala
@@ -193,10 +193,10 @@ trait Compiler[F[_]] {
         String = {
       def sansEscape(pat: List[Char]): List[Char] = pat match {
         case '_' :: t =>         '.' +: escape(t)
-        case '%' :: t => ".*".toList ++ escape(t)
-        case c :: t   =>
+        case '%' :: t => ".*".toList ⊹ escape(t)
+        case c   :: t =>
           if ("\\^$.|?*+()[{".contains(c))
-            '\\' +: escape(t)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/quasar/compiler.scala
```diff
@@ -193,10 +193,10 @@ trait Compiler[F[_]] {
         String = {
       def sansEscape(pat: List[Char]): List[Char] = pat match {
         case '_' :: t =>         '.' +: escape(t)
-        case '%' :: t => ".*".toList ++ escape(t)
-        case c :: t   =>
+        case '%' :: t => ".*".toList ⊹ escape(t)
```
mossprescott: First time I've seen this, although I see we have a definition for `SemigroupF` (which we'e apparently not using). Any particular reason to switch? Just curious.

mossprescott:
Sorry, just noticed this PR hanging out there. Just the one question, otherwise obviously this is a good fix for an embarrassing bug.

core/src/main/scala/quasar/compiler.scala
sellout: I just tend to prefer more general operations. In this case, I think the function could avoid the conversions to/from List and work directly on String, but I didn’t take that step. So it’s just a little flag like “oh, this just needs a semigroup, and we have one of those for strings …”

We do use it in a few other places, I think, but maybe not the most compelling reason.

jdegoes:
This would be good to get in 2.2.2

PR #975, opened 2015-11-03
[SD-1068] preparation for external configuration extension

drostron:
supersedes #967

### DIFF
--- a/core/src/main/scala/quasar/config/config.scala
+++ b/core/src/main/scala/quasar/config/config.scala
@@ -18,8 +18,10 @@ package quasar.config
 
 import com.mongodb.ConnectionString
 import quasar.Predef._
+import quasar.config.FsPath.NonexistentFileError
 import quasar.fp._
 import quasar._, Evaluator._, Errors._
+import quasar.Evaluator.EnvironmentError.EnvFsPathError
 import quasar.fs.{Path => EnginePath}
 
 import java.io.{File => JFile}
@@ -67,10 +69,6 @@ object BackendConfig {
       decoder = _.get[MongoDbConfig]("mongodb").map(v => v: BackendConfig))
 }
 
-@typeclass trait Empty[A] {
-  def empty: A
-}
-
 trait ConfigOps[C] {
   import FsPath._
 
@@ -110,7 +108,7 @@ trait ConfigOps[C] {
   private def alternatePath: Task[FsPath[File, Sandboxed]] =
     OS.currentOS >>= defaultPathForOS(dir("SlamData") </> file("slamengine-config.json"))
 
-  def fromFile(path: FsPath[File, Sandboxed])(implicit ev: DecodeJson[C]): EnvTask[C] = {
+  def fromFile(path: FsPath[File, Sandboxed])(implicit D: DecodeJson[C]): EnvTask[C] = {
     import java.nio.file._
     import java.nio.charset._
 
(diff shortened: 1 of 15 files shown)


core/src/main/scala/quasar/mounter.scala
```diff
@@ -25,10 +25,10 @@ import scalaz._, Scalaz._
 import scalaz.concurrent._
 
 object Mounter {
-  def defaultMount(config: Config): EnvTask[Backend] =
-    mount(config, BackendDefinitions.All)
+  def defaultMount(mountings:  Map[Path, BackendConfig]): EnvTask[Backend] =
```
sellout: I like this not depending on Config anymore … but maybe we can make `Mountings` at least a type alias?

drostron: It already is, just need to update it here.
https://github.com/drostron/quasar/blob/ready/SD-1068-external-conf-extension/core/src/main/scala/quasar/config/package.scala#L25

web/src/main/scala/quasar/config/WebConfig.scala
```diff
@@ -0,0 +1,40 @@
+package quasar.config
+
+import quasar.Predef._
+import quasar.config.implicits._
+
+import argonaut._, Argonaut._
+import monocle._, macros.Lenses
+
+@Lenses final case class ServerConfig(port: Int)
```
sellout: The reason the port was optional here before is that we write the config file from this structure, and if a user has a config with no port and adds a mounting, we want to maintain the config file having no port specified. And if they have the default port explicitly specified, we want to keep it in the config file.

However … we have since moved toward having configs pretty much only updated via the Quasar, so perhaps we no longer care about the distinction? @jdegoes, @mossprescott?

sellout:
I like this new approach a lot, and like that ServerConfig has been pulled out of core.

So, just minor stuff, and waiting on some input re: non-optional port number.

web/src/main/scala/quasar/api/fs.scala
```diff
@@ -273,12 +275,12 @@ final case class FileSystemApi(
             e => NotFound(e.getMessage),
             // TODO: If the requested port is unavailable the server will restart
             //       on a random one, thus this response text may not be accurate.
-            i => updateServerConfig(SDServerConfig(Some(i)), config) *>
+            i => updateServerConfig(scPort.set(i)(Empty[SC].empty), config) *>
```
wemrysi: This assumes that `Port` is the only configuration parameter for `SC`, does it not? This doesn't seem safe for an abstract `SC`, though it was fine for the concrete `SDServerConfig`, as we knew it was the only parameter.

sellout: Ah, yeah, this should be using the existing SC, not an empty one. Good catch.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -150,16 +126,17 @@ object Config {
     } yield config
   }
 
-  def fromFileOrEmpty(path: Option[FsPath[File, Sandboxed]]): EnvTask[Config] = {
-    def loadOr(path: FsPath[File, Sandboxed], alt: EnvTask[Config]): EnvTask[Config] = 
-      handleWith(fromFile(path)) { 
+  def fromFileOrEmpty(path: Option[FsPath[File, Sandboxed]])(implicit ev1: DecodeJson[C], ev2: Empty[C])
```
wemrysi: I think this function should return `OptionT[EnvTask, C]` rather than rely on the dubious `Empty` instance. This also pushes the decision of choosing an "empty" or "default" configuration as far out to the edge of the application as possible, where it belongs, imho.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -83,22 +67,14 @@ object BackendConfig {
       decoder = _.get[MongoDbConfig]("mongodb").map(v => v: BackendConfig))
 }
 
-final case class Config(
-  server:    SDServerConfig,
-  mountings: Map[EnginePath, BackendConfig])
+@typeclass trait Empty[A] {
```
wemrysi: I don't think we need this class, the places where it is currently required can be refactored to no longer need it. Having default values for concrete configuration types should be sufficient, I think.

core/src/main/scala/quasar/config/config.scala
wemrysi: We could also return an `EitherT` if we want to expose the failure reason for logging or similar.

web/src/main/scala/quasar/api/fs.scala
drostron: Yes, good catch — will update.

core/src/main/scala/quasar/config/config.scala
drostron: Agreed.

core/src/main/scala/quasar/config/config.scala
drostron: Likely so — will give it a go.

web/src/main/scala/quasar/config/WebConfig.scala
jdegoes: Probably doesn't matter so long as we can start even if the port is taken.

mossprescott: I think it still makes sense to model the attribute as optional. The API provides DELETE /server/port, and we need a way to preserve the resulting state.

drostron: Makes sense. I'll change it back to optional.

drostron:
Pushed commit addressing comments and rebased onto master.

sellout:
Looks good, just update that test name.

drostron:
Pushed a commit updating the test name.

PR #977, opened 2015-11-03
broaden wartremover application

drostron:


web/src/main/scala/quasar/api/server.scala
```diff
@@ -159,6 +157,7 @@ object Server {
     openClient: Boolean,
     port: Option[Int])
 
+  @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NonUnitStatements"))
```
sellout: Is this also due to puffnfresh/wartremover#149? We should have a comment here explaining why we need the suppression.

web/src/main/scala/quasar/api/zip.scala
```diff
@@ -54,6 +54,7 @@ object Zip {
     // mutating operations: one to accept an Op to be processed, and another
     // to poll for data that's been written.
     class Buffer {
+      @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.Var"))
```
sellout: And a comment here explaining why we need a var.

Suppressions should come with explanations, basically.

web/src/main/scala/quasar/api/server.scala
sellout: Oh, no … not 149, but still.

mossprescott:
This is a good idea. Any idea if it has any impact on compile time, or on how often we get a total recompile?

drostron:
I'll add comments as suggested by @sellout.

@mossprescott, the build times appear to be within the standard range and the incremental compilation situation appears to be similar. _(tough to be precise on those at the moment though)_

drostron:
Speaking of the incremental compilation situation, I spent a bit of time on it and appear to have found a solution. The reason it works eludes me unfortunately. Hopefully it address the issue for others as well.

sellout:
I’m ok with that as a workaround, but we should open an issue against wartremover for it, too, so we can go back to the other way eventually.

drostron:
Agreed on opening an issue against wartremover. I'll look into getting a small repo case setup and submit an issue.

drostron:
incremental compilation wartremover issue: puffnfresh/wartremover/issues/202

drostron:
Pushed commit adding comments for bare @SuppressWarnings.

mossprescott:
Since your PR was merged already, do we want to wait for a wart-remover release as opposed to merging this workaround? Otherwise, we'll need to remember to come back and clean this up, right?

drostron:
It could be a month or two before the next wartremover release and it might be nice to have incremental compilation working before then. We will have to remember to come back to it, but I imagine the associated comment will remind whomever makes the next wartremover settings update.

PR #978, opened 2015-11-03
Improvements to folds.

sellout:
I made these changes in the service of SD-1069, but ended up solving it
with a different approach. However, these changes are still useful.

Primarily, this is to fix a case where we ended up generating multiple
`MapReduceTask`s instead of a single one (see the changes in
`Workflow.crush`). However, I made some other changes …

* Functorized WorkflowTask;
* added `repeatedly` and `simply` algebra transforms;
* converted various folds to use the new transforms;
* fixed `reorderOps` to handle cases where the result was reorderable; and
* removed a few `Recursive` operations:
  * `descend` was non-recursive and enabled manual ana-esque functions,
  * `rewrite` was replaced by `simply`, and
  * `restructure` was just an odd `cata`.


core/src/main/scala/quasar/jscore/package.scala
```diff
@@ -72,6 +72,47 @@ package object jscore {
   def unsafeAssign(lhs: JsCore, rhs: => JsCore): Js.Expr =
     Js.BinOp("=", lhs.toJs, rhs.toJs)
 
+  def replaceSolitary(oldForm: JsCore, newForm: JsCore, in: JsCore) =
```
wemrysi: Mind adding return types to this and `maybeReplace`?

wemrysi:
Other than it not compiling (=P) and the minor comment I had, :+1:

PR #979, opened 2015-11-06
Introduce FunctorT and TraverseT.

sellout:
This makes the compiler tests more robust to compiler changes. On failure, we
display both the normalized diff and the original (un-normalized)
result.

`TraverseT` (and `FunctorT`) is a generalization of
`Recursive with Corecursive` providing `traverse` (and `map`) for types
that can’t implement one of `Recursive` or `Corecursive` (EG, `Free` and `Cofree`).

Transformations have been moved from the `Recursive` type class to the
new type classes, and some existing algebras have been generalized from
`Fix` to `T[_[_]]`.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/workflowop.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/workflowop.scala
@@ -18,7 +18,7 @@ package quasar.physical.mongodb
 
 import quasar.Predef._
 import quasar.{RenderTree, RenderedTree, Terminal, NonTerminal}, RenderTree.ops._
-import quasar.recursionschemes._, Recursive.ops._
+import quasar.recursionschemes._, Recursive.ops._, FunctorT.ops._
 import quasar.fp._
 import quasar.fs.Path
 import optimize.pipeline._
@@ -169,17 +169,17 @@ object Workflow {
   def task(fop: Crystallized): WorkflowTask =
     (finish(_, _)).tupled(fop.op.para(crush))._2
 
-  val coalesceƒ: WorkflowF[Workflow] => Option[Workflow] = {
+  val coalesceƒ: WorkflowF[Workflow] => Option[WorkflowF[Workflow]] = {
     case $Match(src, selector) => src.unFix match {
       case $Sort(src0, value) =>
-        Fix($Sort(Fix($Match(src0, selector)), value)).some
+        $Sort(Fix($Match(src0, selector)), value).some
       case $Match(src0, sel0) =>
-        Fix($Match(src0, sel0 ⊹ selector)).some
+        $Match(src0, sel0 ⊹ selector).some
       case _ => None
     }
     case p @ $Project(src, shape, id) => src.unFix match {
       case $Project(src0, shape0, id0) =>
-        Fix($Project(src0, inlineProject(p, List(shape0)), id0 |+| id)).some
+        $Project(src0, inlineProject(p, List(shape0)), id0 |+| id).some
       // Would like to inline a $project into a preceding $simpleMap, but
       // This is not safe, because sometimes a $project is inserted after
       // $simpleMap specifically to pull fields out of `value`, and those
@@ -195,31 +195,31 @@ object Workflow {
       //             jsShape(jscore.Ident("__tmp")))),
       //         flatten, scope)))
       case $Group(src, grouped, by) if id != ExcludeId =>
-        inlineProjectGroup(shape, grouped).map(g => Fix($Group(src, g, by)))
+        inlineProjectGroup(shape, grouped).map($Group(src, _, by))
       case $Unwind(Fix($Group(src, grouped, by)), unwound)
           if id != ExcludeId =>
         inlineProjectUnwindGroup(shape, unwound, grouped).map { case (unwound, grouped) =>
-          Fix($Unwind(Fix($Group(src, grouped, by)), unwound))
+          $Unwind(Fix($Group(src, grouped, by)), unwound)
         }
       case _ => None
     }
     case $Sort(Fix($Sort(src, sort1)), sort2) =>
-      Fix($Sort(src, sort2 ⊹ sort1)).some
+      $Sort(src, sort2 ⊹ sort1).some
     case $Limit(src, count) => src.unFix match {
       case $Limit(src0, count0) =>
-        Fix($Limit(src0, count0 min count)).some
+        $Limit(src0, count0 min count).some
       case $Skip(src0, count0) =>
-        Fix($Skip(Fix($Limit(src0, count0 + count)), count0)).some
+        $Skip(Fix($Limit(src0, count0 + count)), count0).some
       case _ => None
     }
     case $Skip(src, count) => src.unFix match {
-      case $Skip(src0, count0) => Fix($Skip(src0, count0 + count)).some
+      case $Skip(src0, count0) => $Skip(src0, count0 + count).some
       case _                   => None
     }
     case $Group(src, grouped, \/-($literal(bson))) if bson != Bson.Null =>
-      Fix($Group(src, grouped, \/-($literal(Bson.Null)))).some
+      $Group(src, grouped, \/-($literal(Bson.Null))).some
     case op0 @ $Group(_, _, _) =>
-      inlineGroupProjects(op0).map(tup => Fix(($Group[Workflow](_, _, _)).tupled(tup)))
+      inlineGroupProjects(op0).map(($Group[Workflow](_, _, _)).tupled)
     case $GeoNear(src, _, _, _, _, _, _, _, _, _) => src.unFix match {
       // FIXME: merge the params
       case $GeoNear(_, _, _, _, _, _, _, _, _, _) => None
@@ -228,33 +228,33 @@ object Workflow {
     case $Map(src, fn, scope) => src.unFix match {
       case $Map(src0, fn0, scope0) =>
         Reshape.mergeMaps(scope0, scope).map(
-          s => Fix($Map(src0, $Map.compose(fn, fn0), s)))
+          $Map(src0, $Map.compose(fn, fn0), _))
       case $FlatMap(src0, fn0, scope0) =>
         Reshape.mergeMaps(scope0, scope).map(
-          s => Fix($FlatMap(src0, $FlatMap.mapCompose(fn, fn0), s)))
+          $FlatMap(src0, $FlatMap.mapCompose(fn, fn0), _))
       case _                   => None
     }
     case $FlatMap(src, fn, scope) => src.unFix match {
       case $Map(src0, fn0, scope0)     =>
-        Reshape.mergeMaps(scope0, scope).map(map =>
-          Fix($FlatMap(src0, $Map.compose(fn, fn0), map)))
+        Reshape.mergeMaps(scope0, scope).map(
+          $FlatMap(src0, $Map.compose(fn, fn0), _))
       case $FlatMap(src0, fn0, scope0) =>
-        Reshape.mergeMaps(scope0, scope).map(map =>
-          Fix($FlatMap(src0, $FlatMap.kleisliCompose(fn, fn0), map)))
+        Reshape.mergeMaps(scope0, scope).map(
+          $FlatMap(src0, $FlatMap.kleisliCompose(fn, fn0), _))
       case _                   => None
     }
     case sm @ $SimpleMap(src, _, _) => src.unFix match {
-      case sm0 @ $SimpleMap(_, _, _) => Fix(sm0 >>> sm).some
+      case sm0 @ $SimpleMap(_, _, _) => (sm0 >>> sm).some
       case _                         => None
     }
     case $FoldLeft(head, tail) => head.unFix match {
       case $FoldLeft(head0, tail0) =>
-        Fix($FoldLeft(head0, tail0 ⊹ tail)).some
+        $FoldLeft(head0, tail0 ⊹ tail).some
       case _                       => None
     }
     case $Out(src, _) => src.unFix match {
-      case $Read(_) => src.some
-      case _        => None
+      case r @ $Read(_) => r.some
+      case _            => None
     }
     case _ => None
   }
@@ -582,7 +582,7 @@ object Workflow {
     }
 
     Crystallized(
-      hylo(promoteKnownShape(finished))(coalesce, crystallizeƒ)
+      Corecursive[Fix].ana(promoteKnownShape(finished))(crystallizeƒ).transCata[WorkflowF](coalesce)
         // TODO: this can coalesce more cases, but hasn’t been done thus far and
         //       requires rewriting many tests in a much less readable way.
         // .cata[Workflow](x => coalesce(uncleanƒ(x).unFix))
@@ -590,10 +590,10 @@ object Workflow {
   }
 
   final case class $Pure(value: Bson) extends SourceOp
-  def $pure(value: Bson) = coalesce($Pure(value))
+  def $pure(value: Bson) = Fix(coalesce($Pure(value)))
 
   final case class $Read(coll: Collection) extends SourceOp
-  def $read(coll: Collection) = coalesce($Read(coll))
+  def $read(coll: Collection) = Fix(coalesce($Read(coll)))
 
   final case class $Match[A](src: A, selector: Selector)
       extends ShapePreservingF[A]("$match") {
@@ -602,7 +602,7 @@ object Workflow {
   }
   object $Match {
     def make(selector: Selector)(src: Workflow): Workflow =
-      coalesce($Match(src, selector))
+      Fix(coalesce($Match(src, selector)))
   }
   val $match = $Match.make _
 
@@ -700,7 +700,7 @@ object Workflow {
   }
   object $Project {
     def make(shape: Reshape, id: IdHandling)(src: Workflow): Workflow =
-      coalesce($Project(src, shape, id))
+      Fix(coalesce($Project(src, shape, id)))
 
     def EmptyDoc[A](src: A) = $Project(src, Reshape.EmptyDoc, ExcludeId)
   }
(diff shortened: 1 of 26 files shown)


core/src/main/scala/quasar/recursionschemes/FunctorT.scala
```diff
@@ -0,0 +1,57 @@
...
+
+package quasar.recursionschemes
+
+import quasar.Predef._
+import quasar.fp._
+
+import scalaz._, Scalaz._
+import simulacrum.{typeclass, op}
+
+@typeclass trait FunctorT[T[_[_]]] {
```
sellout: There are four closely related functions here that differ only in when `f` is applied. They are (from earliest to latest application):

* topDownTransform – before anything
* transAna – after “projecting”
* transCata – before “embedding”
* transform – after everything
 
The first two work top-down, while the last two are bottom up. The distinction between the inner two and the outer two is more subtle, but important. The inner ones (cata and ana) have the result of `f` wrapped in the context of the input, whereas the outer ones provide the context as part of the result. This is currently only obvious with `Cofree`. When we simplify an LP – EG, `Let(Free('tmp0), Constant(5), Free('tmp0))` → `Constant(5)` – using `transform`, we’d end up with<sup><a name="rt1">[1](#ft1)</a></sup>
```scala
Cofree(Constant(5), ConstantF(5))
```
but using `transCata`, we’d end up with
```scala
Cofree(Let(Free('tmp0), Constant(5), Free('tmp0)), ConstantF(5))
```
because transCata’s `f` only returns `ConstantF(5)`, and leaves it up to `transCata` to provide the context (and the only context it has available is the one that it unwrapped just before calling `f`).

<a name="ft1">[1](#rt1)</a>: since transform’s `f` receives the argument with its context, it might appear as though it provides `f` the option returning either of the results, but the only two ways to do that are to either have the function passed as `f` constrained on both `Recursive` and `Corecursive`, which prohibits `Cofree`, or to fix `T` _as_ `Cofree`, which eliminates the flexibility.

core/src/main/scala/quasar/recursionschemes/FunctorT.scala
```diff
@@ -0,0 +1,57 @@
...
+  // NB: this could perhaps also be called `transCata`
```
sellout: This comment is false. I’ll remove it.

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -238,97 +260,67 @@ package object recursionschemes {
...
+  def zipPara[T[_[_]], F[_]] = new ZipParaPAT[T, F]
 
   /** Repeatedly applies the function to the result as long as it returns Some.
     * Finally returns the last non-None value (which may be the initial input).
     */
-  def repeatedly[T[_[_]]: Recursive: Corecursive, F[_]: Functor](
-    f: F[T[F]] => Option[T[F]]):
-      F[T[F]] => T[F] =
-    expr => f(expr).fold(Corecursive[T].embed(expr))(repeatedly(f) <<< (_.project))
+  def repeatedly[T[_[_]], F[_]](f: F[T[F]] => Option[F[T[F]]]):
```
sellout: This (and `simply`) can actually now have the type `(A => Option[A]) => A => A`.

core/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -582,7 +582,7 @@ object Workflow {
     }
 
     Crystallized(
-      hylo(promoteKnownShape(finished))(coalesce, crystallizeƒ)
+      Corecursive[Fix].ana(promoteKnownShape(finished))(crystallizeƒ).transCata[WorkflowF](coalesce)
         // TODO: this can coalesce more cases, but hasn’t been done thus far and
         //       requires rewriting many tests in a much less readable way.
         // .cata[Workflow](x => coalesce(uncleanƒ(x).unFix))
```
mossprescott: Needs updating?

core/src/main/scala/quasar/recursionschemes/FunctorT.scala
mossprescott: I appreciate the explanation; how about pasting it into the file as a comment? Seriously, I think it's about time we attempted to explain what all these folds do.

I like how you laid out the 4 "transforms" in your comment, but I wish the commonality was more obvious in the code. Could they all get names that somehow parallel each other? Maybe `transform` is `transCataF` and `topDownTransform` is `transAnaF`? Not great, but maybe there's some other suffix that suggests the more restricted signature.

core/src/main/scala/quasar/recursionschemes/FunctorT.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  def transPara[F[_]: Functor, G[_]](t: T[F])(f: F[(T[F], T[G])] => G[T[G]]):
+      T[G] =
+    map(t)(ft => f(ft.map(tf => (tf, transPara(tf)(f)))))
+
+  def transApo[F[_], G[_]: Functor](t: T[F])(f: F[T[F]] => G[T[G] \/ T[F]]):
+      T[G] =
+    map(t)(ft => f(ft).map(_.fold(ι, transApo(_)(f))))
+
+  def trans[F[_], G[_]: Functor](t: T[F])(f: F ~> G): T[G] =
```
mossprescott: Is this not called `translate` to avoid a name collision, or is it something different?

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
```diff
@@ -0,0 +1,45 @@
...
+    t: T[F], a: A)(
+    f: (A, T[F]) => M[(A, T[F])]):
+      M[T[F]] =
+    f(a, t).flatMap { case (a, tf) =>
+      traverse(tf)(_.traverse(topDownCataM(_, a)(f)))
+    }
+
+  def transCataM[M[_]: Monad, F[_]: Traverse, G[_]](t: T[F])(f: F[T[G]] => M[G[T[G]]]): M[T[G]] =
+    traverse(t)(_.traverse(transCataM(_)(f)).flatMap(f))
+}
```
mossprescott: Is `transAnaM` missing deliberately? Maybe it's impossible to write?

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import scalaz._, Scalaz._
+import simulacrum.typeclass
+
+@typeclass trait TraverseT[T[_[_]]] extends FunctorT[T] {
+  def traverse[M[_]: Applicative, F[_], G[_]](t: T[F])(f: F[T[F]] => M[G[T[G]]]): M[T[G]]
+
+  def map[F[_], G[_]](t: T[F])(f: F[T[F]] => G[T[G]]) = traverse[Id, F, G](t)(f)
+
+  def transformM[F[_]: Traverse, M[_]: Monad](t: T[F])(f: T[F] => M[T[F]]): M[T[F]] =
```
mossprescott: As a general thing, I'd love to see laws that say `transform[F] === transformM[Id, F]`, and tests to verify. Otherwise we need to write coverage for each manually?

core/src/main/scala/quasar/recursionschemes/FunctorT.scala
sellout: Yeah, I agree on both counts. I’ve been hesitant to change the names, because they came from fixplate, and I’m not sure if they have some connection to papers, etc. But showing these relations is more important.

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -28,6 +28,12 @@ import simulacrum.typeclass
 
 /** Generalized folds, unfolds, and refolds. */
 package object recursionschemes {
+  implicit def RecCorecTraverseT[T[_[_]]: Recursive: Corecursive]: TraverseT[T] =
+    new TraverseT[T] {
```
mossprescott: How much would it buy us to override `map` in these implementations? On stack frame? Nothing at all? Who cares, we'll profile later?

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -49,6 +60,13 @@ package object recursionschemes {
   //   new Corecursive[Free [?[_], A]] {
   //     def embed[F[_]: Functor](t: F[Free[F, A]]) = Free.liftF(t).join
   //   }
+  // implicit def FreeTraverseT[A]: TraverseT[Free[?[_], A]] =
```
mossprescott: Are we keeping this around because it's going to become real sometime soon?

core/src/main/scala/quasar/physical/mongodb/workflowop.scala
sellout: Not yet … I’ve only updated the compiler-related algebras to be `T`-agnostic. We’ll get around to the planner ones eventually, but even so, this is about `JsCore` being converted to `Js`, so it’ll be nasty regardless.

core/src/main/scala/quasar/recursionschemes/FunctorT.scala
sellout: I think it was originally short for “transform”, but that’s not so great. It could be `translate` now 😃

core/src/main/scala/quasar/recursionschemes/package.scala
mossprescott: Indeed, so why keep them as is? Actually, these seem like very general-purpose functions at this point (except that `simply` isn't a very good name, maybe it should be `once`). I wonder if they already exist somewhere? Hmm, hoogling `(a -> Maybe a) -> a -> a` gets me nothing and I'm out of ideas.

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
sellout: Nope, there are a lot of [un]folds missing. I’d like to get them all in, but I’d also like it to be a separate repo.

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
sellout: Yeah, good call. There’s a library for that, right? Can’t recall.

core/src/main/scala/quasar/std/relations.scala
```diff
@@ -118,9 +118,18 @@ trait RelationsLib extends Library {
...
+        orig match {
+          case InvokeF(_, List(l, r)) => l.project match {
+            case ConstantF(Data.True) => r.project.some
+            case _                    => r.project match {
+              case ConstantF(Data.True) => l.project.some
+              case _                    => None
+            }
+          }
+          case _ => None
+        }
```
mossprescott: Well, that's a bitch. What we had *was* nice and readable.

You need to be able to inspect the recursive structure and then rewrite using the children, whose actual shape is unknown. Well, mostly I'm impressed that it works, but it's not good to read at all.

How about an extractor for `InvokeF`:
```scala
object IsInvoke {
  def unapply(lp: LogicalPlan[T[LogicalPlan]]): Option[(Func, List[LogicalPlan[T[LogicalPlan])] = ...
}
```

You know, because fancy extractors are *awesome*.

If that doesn't work, let's find something that does, because these simplifications are important, there will be more of them, and we mostly test them by staring at them.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: I’m in the camp of “Who cares, we'll profile later.”

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: Yeah … I’m debating seeing if I should just try to add the Functor constraint to embed. But I think @wemrysi mentioned that the Functor constraint may disappear from `Free.liftF` at some point?

core/src/test/scala/quasar/optimizer.scala
```diff
@@ -31,39 +31,40 @@ class OptimizerSpec extends Specification with CompilerHelpers with TreeMatchers
     }
 
     "completely inline stupid lets" in {
-      Optimizer.simplifyƒ(
+      Optimizer.simplifyƒ[Fix].apply(
         LetF('tmp0, read("foo"), Let('tmp1, Free('tmp0), Free('tmp1)))) must
-        beSome(Let('tmp1, read("foo"), Free('tmp1)))
+        beSome(LetF('tmp1, read("foo"), Free('tmp1)))
```
mossprescott: Surely at one time this was meant to expect just `read("foo")`. If this is the test we want now, can we call it "inline one level of stupid lets" or something?

core/src/test/scala/quasar/optimizer.scala
```diff
@@ -31,39 +31,40 @@ class OptimizerSpec extends Specification with CompilerHelpers with TreeMatchers
...
+      Optimizer.simplifyƒ[Fix].apply(
         LetF('tmp0, read("foo"),
           Let('tmp0, read("bar"),
             makeObj(
               "bar" -> ObjectProject(Free('tmp0), Constant(Data.Str("bar"))))))) must
         beSome(
-          Let('tmp0, read("bar"),
+          LetF('tmp0, read("bar"),
             makeObj(
               "bar" -> ObjectProject(Free('tmp0), Constant(Data.Str("bar"))))))
```
mossprescott: Another test that's no longer testing what it was meant to. This should either be called "eliminate unused binding" or get fixed to do what it says.

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
mossprescott: +1 on the separate repo.

I just missed `transAnaM` because it completes the foursome. You mean you don't want writing these folds to be your full-time job?

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
mossprescott: Um, I wouldn't be the one to ask. **cats** comes with a `Laws` trait for each type, possibly others as well.

core/src/main/scala/quasar/recursionschemes/package.scala
mossprescott: Low-hanging fruit for the inevitable performance project. I like it.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: I think we’ll want Free to have these instances because I see `planƒ` eventually having a type something like `LogicalPlan[PhysicalPlan] => Free[LogicalPlan, PhysicalPlan]` and having folds (like evaluation) over that result.

mossprescott:
My number one issue here is how much worse the simplifiers are to read and write now. I really think that's worth addressing.

I'd also like to see an attempt to rationalize and explain the folds and their names. I know it's an impossible task, but I think it's worth saying *something* about the less obvious connections.

Finally, it looks like the optimizer tests have drifted from their original purpose and could use some review, although you're not creating the problem here. Maybe there are also missing tests?

Other than that, this looks great and I'm impressed :)

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
sellout: Ah, https://github.com/typelevel/discipline … I was looking in the wrong dependency list for cats.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: Yeah, I just realized it, so I made a note to remind me for the PR update. I like the rename, and I agree that they could maybe live somewhere else (assuming they don’t already).

core/src/test/scala/quasar/optimizer.scala
sellout: Do you have a preference between the two? At some point I switched a bunch of tests from using the folds to calling the algebra directly, to take some noise out of the equation. On the other hand, the way the [co]algebra is written defines how it can be used – EG, `T[F] => F[T[F]]` can only be run top-down, and a test of just the algebra may work fine, but a test of the fold may show that it needs to be rewritten to run bottom-up.

core/src/main/scala/quasar/recursionschemes/package.scala
wemrysi: Yes, the `Functor` constraint is no longer necessary in the [7.2.x branch](https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Free.scala#L73)

core/src/main/scala/quasar/std/relations.scala
sellout: I tried to write `type Simplifer = forall t. FunctorT t => LogicalPlan (t LogicalPlan) -> LogicalPlan (t LogicalPlan)`, but Scala wasn’t happy with it for some reason.

That extractor seems reasonable enough.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: Should we switch? I recall there was something else … specs2 v3? … that also required scalaz 7.2.

wemrysi: I'm not sure, though it doesn't look like 7.2 is officially released yet, might want to wait until that happens.

sellout:
This is failing during the `sbt doc`, and I have no idea why. It fails locally, but if I don’t run `doc` (like the `ready` script doesn’t), compilation and all tests succeed.

jedesah:
Weird, it doesn't look like a failure that is specific to documentation, it just looks like the code does not compile.

sellout:
@jedesah What seems to be happening is the bad documentation keeps that function from being added to the trait. Then, when we try to use that function, we get an error that it doesn't exist.

I'm not at my computer, but just before I left, I think I might have determined that it's caused by starting a line with a backtick.

mossprescott:
If this is what we get for writing comments, screw it, let's go back to the old policy.

PR #980, opened 2015-11-07
views; not exposed to users yet

mossprescott:
Completes SD-977, in terms of what the implementation is capable of, but currently views cannot be added through the API and are ignored if found in the configuration file, due to bugs that make them confusingly broken (see SD-1101).

A fairly dumb generator for configurations, and a simple test of round-trip encoding/decoding.

A new Backend implementation, which interprets all requests involving views by transforming them into queries where the view queries are sub-selects. This means embedding ASTs inside each other, which works surprisingly well but did expose some ways in which the AST is not composable.

Fix ScopeTables to treat sub-selects independently, so that it's not always necessary to qualify all field references when a sub-select is present. This relates to the more subtle problem that's now tracked in SD-1101.

Slight improvements to RenderTree for AnnotatedTree and Provenance annotations in particular.Rename `BackendConfig` to `MountConfig`, since a view is a mount, but not a backend. New subtype, `ViewConfig`, holding a (parsed) query and optional temporary path. The codec translates between that and the URI syntax, which looks like `sql2:///foo/bar?q=select+*+...` (note: always three slashes, whether or not there is a temp path present, because a name after the second slash is interpreted as a "host" by standard parsers.)

Several new error embeddings are added, and the signature for `Backend.count` is changed to account for errors that can occur when various operations end up executing a queries due to views.


core/src/main/scala/quasar/backend.scala
```diff
@@ -281,6 +281,98 @@ object ThreadPoolAdapterBackend {
...
+    views0.map { case (p, ViewConfig(tp, q)) =>
+      p.asRelative -> sql.relativizePaths(q, p.parent.asRelative).valueOr(κ(q))
+    }
+
+  def tableName(path: Path) = "__sd__" + path.filename
+
+  def expand(query: sql.Expr): sql.Expr = sql.rewriteRelations(query) {
+    case sql.TableRelationAST(name, alias) =>
+      val p = Path(name)
+      views.get(p).fold[Option[sql.ExprRelationAST[sql.Expr]]](None) { query =>
```
wemrysi: I think this `fold` can be replaced by a `map`.

core/src/main/scala/quasar/backend.scala
```diff
@@ -281,6 +281,98 @@ object ThreadPoolAdapterBackend {
...
+        Some(sql.ExprRelationAST(query, alias.getOrElse(tableName(p))))
+      }
+    case _ => None
+  }
+
+  def ls0(dir: Path) = {
+    val vs = (views.keys.map { p =>
```
wemrysi: I think we can rewrite this as `views.keys foldMap (_.rebase(dir).toSet map { ... })` (or `views.keys foldMap (_.rebase(dir).toSet) map {...}`, whichever you prefer) to make it a bit more straightforward.

wemrysi:
Looks good, though we'll have to convert it to use the new architecture soon =( I'll probably ask for some help on that if you have time.

mossprescott:
Yes, I'l definitely be able to help with the conversion/re-implementation. The main thing is to get the tests in.

mossprescott:
@estewei: thanks, I cleaned up those folds and now I need to resolve conflicts, so might ask you to take another look.

mossprescott:
@estewei: resolved and ready for you.

@drostron: you might want to take a look at the changes to the config stuff, since there were a fair number of conflicting changes I just whipped through.

mossprescott:
Of course I meant to @-mention @wemrysi :)

core/src/main/scala/quasar/config/config.scala
```diff
@@ -60,13 +60,54 @@ object MongoDbConfig {
...
 }
 
-object BackendConfig {
-  implicit def BackendConfig: CodecJson[BackendConfig] =
-    CodecJson[BackendConfig](
-      encoder = _ match {
-        case x @ MongoDbConfig(_) => ("mongodb", MongoDbConfig.Codec.encode(x)) ->: jEmptyObject
-      },
-      decoder = _.get[MongoDbConfig]("mongodb").map(v => v: BackendConfig))
+final case class ViewConfig(tempPath: Option[EnginePath], query: sql.Expr) extends MountConfig {
```
wemrysi: What is `tempPath` for? Is it used anywhere (I couldn't seem to find a usage)?

mossprescott: It's mentioned in the "spec" (the JIRA epic), as the location for any temp files that might be created. I guess if we were going to do any caching of results that's where they would go, but at present we just execute the query every time.

So practically speaking at this point it's only present so that it can be written to the config file and read back.

web/src/main/scala/quasar/api/fs.scala
```diff
@@ -43,17 +43,18 @@ import scodec.bits.ByteVector
...
  */
 final case class FileSystemApi[WC, SC](
   initialConfig: WC,
   createBackend: WC => EnvTask[Backend],
-  validateConfig: BackendConfig => EnvTask[Unit],
+  validateConfig: MountConfig => EnvTask[Unit],
   restartServer: WC => Task[Unit],
   configChanged: WC => Task[Unit],
   webConfigLens: WebConfigLens[WC, SC]) {
+// >>>>>>> upstream/master
```
drostron: Conflict resolution leftover.

drostron:
@mossprescott, the config related changes look ok.

core/src/main/scala/quasar/config/config.scala
wemrysi: Ok, does it make sense to drop it until we actually have a use for it? If we add it later, we can always read a missing `tempPath` back as `None`.

mossprescott: Sure, I don't see why not. It fails the "at least have tests for it" test, after all.

mossprescott:
@wemrysi: ready for you.

wemrysi:
At the risk of being fickle, if the temp path is going to remain part of the URI, then let's just leave it in, I didn't realize that. Better to be consistent than have all the comments saying we're ignoring part of the specified URI.

mossprescott:
@wemrysi: See the latest commit.

Also, after hearing jdegoes talk about how the front-end might use these in the short-term I decided it makes sense to use the same key for the URI for both mongodb and view config, so I made that switch.

wemrysi:
:+1: looks good to merge

PR #982, opened 2015-11-10
re-add mistakenly removed warts

drostron:
Warts were mistakenly removed in #975.


sellout:
Hoist with your own petard!

drostron:
Indeed, was caught by my own merge crossfire.

PR #983, opened 2015-11-10
Add instructions in order to run integration tests successfully

jedesah:


README.md
```diff
@@ -38,6 +38,27 @@ To compile the project and run tests, execute the following command:
...
+you probably want to install MongoDB locally and point Quasar to that one. Installing MongoDB locally is probably a good idea as it will
+allow you to run the integration tests offline as well as make the tests run as fast as possible.
+
+Visit the MongoDB website and follow the installation instructions in order to install MongoDB locally. Once that is done we need to set a few
+environment variables in order to inform Quasar about where to find the backends required in order to run the integration tests.
+
+Set the following enviroment variables:
+
+```bash
+QUASAR_MONGODB_3_0="{\"mongodb\":{\"connectionUri\":\"mongodb://`mongoURL`\"}}"
```
syves: I'm not sure what default mongoURL should be replaced with 'localhost/test' or 'localhost/db'

syves: add export to beginning of 53 and 54

jedesah: `export` is specific to bash, one, such as @drostron might be on fish and if I am not mistaken `set -x` is what is used to "export" an environment variable in that shell. I think it's safe to assume that a developer knows how to export an environment variable using their shell of choice. That being said since most people probably stick with bash, export would not hurt and non bash users can surely translate.

jedesah: @syves For the first comment, if I understand you correctly, you are suggesting adding an example, such as `localhost:27027`. Sure.

syves: well I don't think I'd ever exported an env variable, but perhaps there won't be any more interns.
Is it good practice to be specific and detailed in docs? maybe [export] or some way of denoting a default like within the JSON itself.

jedesah: @syves I think it is certainly good practice to be specific and detailed in docs. That being said I do not think it is good practice to be repetitious (in general) if a particular process or piece of information is well documented elsewhere.

Although, exporting environment variables is well documented elsewhere, I do think adding `export` is reasonable in this case.

PR #984, opened 2015-11-11
Kill AnnotatedTree.

sellout:
Reimplement analyses and compilation as [un]folds.

Also, added Elgot folds and generalized a few other recursionscheme
operations; took some advantage of @tpolecat’s kinda-curried type
parameters approach to make the type annotations less onerous; and
generalized RenderTree instanecs to work over any type, not just
recursive ones.


core/src/main/scala/quasar/variables.scala
```diff
@@ -34,47 +32,13 @@ final case class VarValue(value: String)
...
-        unchanged
-    ).run(Nil).run.run.map {
-      case (tuples, root) =>
-        val map1 = tuples.foldLeft(new java.util.IdentityHashMap[Expr, A]) { // TODO: Use ordinary map when AnnotatedTree has been off'd
-          case (map, (k, v)) => ignore(map.put(k, v)); map
-        }
-
-        Tree[Expr](root, _.children).annotate(map1.get(_))
-    }
+  // TODO: handle this later. Definitely no sooner than LogicalPlan.
```
mossprescott: Do we need a ticket to justify keeping this TODO? What does it buy us to handle them later?

sellout: This explains it a little: https://slamdata.atlassian.net/browse/SD-1008

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -71,9 +72,10 @@ trait StructuralLib extends Library {
     noSimplification,
     partialTyperV {
       case List(Const(Data.Obj(map1)), Const(Data.Obj(map2))) =>
-        success(Const(Data.Obj(map1 ++ map2)))
-      case List(Const(Data.Obj(map1)), o2) if map1.isEmpty => success(o2)
-      case List(o1, Const(Data.Obj(map2))) if map2.isEmpty => success(o1)
+        success(Const(Data.Obj(
+          if (map1.isEmpty)      map2
+          else if (map2.isEmpty) map1
+          else                   map1 ++ map2)))
```
mossprescott: This won't simplify the case of const-empty on one side and non-const on the other, right? Not that it seems like an important case. Do we even generate empty `Obj`s?

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -326,66 +266,75 @@ package object sql {
...
-      case Unop(expr, op) => NonTerminal("Unop" :: astType, Some(op.sql), render(expr) :: Nil)
-
-      case Splice(expr) => NonTerminal("Splice" :: astType, None, expr.toList.map(render(_)))
-
-      case Ident(name) => Terminal("Ident" :: astType, Some(name))
-
-      case Vari(name) => Terminal("Variable" :: astType, Some(":" + name))
-
-      case x => Terminal("LiteralExpr" :: astType, Some(pprint(x)))
+  implicit val ExprFRenderTree: RenderTree ~> λ[α => RenderTree[ExprF[α]]] =
```
mossprescott: ??? Should this type be making any sense to me?

I would have expected `def ExprFRenderTree[A]: RenderTree[ExprF[A]]`. I guess I'm missing something, or is this doing more than that would?

core/src/main/scala/quasar/std/structural.scala
sellout: We do now 😉 Simplified `MakeObjectN` to a foldLeft, and this eliminates the empty `Obj` introduced there.

It does simplify empty/non-const, but not until type-checking (vs. in `simplifyƒ`).

core/src/main/scala/quasar/semantics.scala
```diff
@@ -320,67 +312,50 @@ trait SemanticAnalysis {
...
-      ScopeTables.second >>>
-      ProvenanceInfer.second
+  val AllPhases: Expr => NonEmptyList[SemanticError] \/ Cofree[ExprF, Annotations] = expr =>
+   coelgotM[NonEmptyList[SemanticError] \/ ?]((TableScope(Map()), expr.cata(projectSortKeysƒ)))(
+     (a, node: ExprF[Cofree[ExprF, Annotations]]) =>
+       attributeCoelgotM[ValidSem](
+         zipCoelgotM[ValidSem](
+           generalizeCoelgot(identifySyntheticsƒ)(_: TableScope, _: ExprF[List[Option[Synthetic]]]).point[ValidSem],
+           inferProvenanceƒ(_: TableScope, _: ExprF[Provenance]))).apply(a._1, node).disjunction,
+     (scopeTablesƒ).tupled(_).disjunction)
```
mossprescott: So, um, yeah. How about a comment that says what this is doing, because otherwise I'm stumped. Actually I guess the type says it all. Maybe just a comment that says "the type says it all".

Is this such a disaster because you're wrapping all the annotations up into one fold? Because I would much prefer a series of simpler annotation steps, even it's less efficient.

What happens to this when the input is `Cofree[ExprF, SourceLocation]`?

mossprescott: Also, why not only `ValidSem` here?

core/src/main/scala/quasar/semantics.scala
```diff
@@ -320,67 +312,50 @@ trait SemanticAnalysis {
...
+    case InvokeFunctionF(_, args) => success(Provenance.allOf(args))
+    case MatchF(_, cases, _) =>
+      success(cases.map(_.expr).concatenate(Provenance.ProvenanceAndMonoid))
+    case SwitchF(cases, _) =>
+      success(cases.map(_.expr).concatenate(Provenance.ProvenanceAndMonoid))
+    case IntLiteralF(_) => success(Provenance.Value)
+    case FloatLiteralF(_) => success(Provenance.Value)
+    case StringLiteralF(_) => success(Provenance.Value)
+    case BoolLiteralF(_) => success(Provenance.Value)
+    case NullLiteralF() => success(Provenance.Value)
```
mossprescott: Surprised not to see the `=>`s aligned here, which I guess I would prefer.

core/src/main/scala/quasar/semantics.scala
sellout: Sorry, why not only `ValidSem` where?

core/src/main/scala/quasar/semantics.scala
```diff
@@ -162,44 +157,40 @@ trait SemanticAnalysis {
...
+          }
 
-        case _ => success(parentScope(node))
-      }
-    })
-  }
+        relations.fold[ValidSem[Map[String, SqlRelation[Expr]]]](
+          success(Map[String, SqlRelation[Expr]]()))(
+          findRelations).map(m =>
+          sel.map((TableScope(parentScope.scope ++ m), _)))
```
mossprescott: Line-breaking disaster :)

I wonder if pattern matching on `Some(relations)` vs `None` would help here, as heretical as that might sound.

core/src/main/scala/quasar/sql/package.scala
sellout: This is akin to `ExprFRenderTree[A: RenderTree]: RenderTree[ExprF[A]] `, except that with, say `Fix[ExprF]` you can’t define it that way, because of the recursion, so the natural transformation defers the resolution of `A: RenderTree`.

The ways we had done this before:
1. creating only Terminals and having Fix/Cofree/Free/whatever fix it up, or
2. creating RenderTree[Fix[ExprF]] directly

were both problematic (the former required the RenderTrees for Cofree, etc. to be written in a restrictive way & wouldn’t render, say, `ExprF[Int]` and the latter doesn’t even work for Cofree or anything).

This way is general, and is the approach used by Scalaz (vs our other attempt, which is type classes like EqualF).

So, now we have two approaches, down from three, and the EqualF approach will probably be replaced with this, too.

core/src/main/scala/quasar/semantics.scala
```diff
@@ -102,58 +101,54 @@ trait SemanticAnalysis {
...
+          } (
+            kExpr => (projs, (orderType, kExpr) :: keys, index))
+      }
+      Select(d, projections ⊹ projs2, r, f, g, Some(sql.OrderBy(keys2)))
     }
+    case node => Fix(node)
+  }
 
-    tree1 => ann(tree(tree1.root.cata(transformƒ)))
+  val identifySyntheticsƒ:
```
mossprescott: Did the hack to identify synthetics by name go from internal to public here? Because that would be a drag.

Could we rewrite the Select and also annotate the projections at the same time?

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -213,62 +235,77 @@ package object recursionschemes {
...
-  def zipPara[T[_[_]], F[_]] = new ZipParaPAT[T, F]
+  // NB: There are potentially two versions of this function – one as below, and
+  //     one where the annotation parameter should become a tuple, too.
+  def zipCoelgot[F[_]: Unzip, A, B, C](f: (A, F[B]) => B, g: (A, F[C]) => C):
+      (A, F[(B, C)]) => (B, C) =
+    (ann, node) => node.unfzip.bimap(f(ann, _), g(ann, _))
+
+  // def zipCoelgotM[F[_]: Unzip, M[_]: Applicative, A, B, C](f: (A, F[B]) => M[B], g: (A, F[C]) => M[C]):
+  //     (A, F[(B, C)]) => M[(B, C)] =
+  //   (ann, node) => node.unfzip.bitraverse(f(ann, _), g(ann, _))
```
mossprescott: Dead code.

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -167,12 +190,11 @@ package object recursionschemes {
...
     }
 
-  implicit def FixRenderTree[F[_]: Foldable](implicit RF: RenderTree[F[_]]) = new RenderTree[Fix[F]] {
-    def render(v: Fix[F]) = {
-      val t = RF.render(v.unFix)
-        NonTerminal(t.nodeType, t.label, v.children.map(render(_)))
+  implicit def FixRenderTree[F[_]](implicit RF: RenderTree ~> λ[α => RenderTree[F[α]]]):
+      RenderTree[Fix[F]] =
+    new RenderTree[Fix[F]] {
+      def render(v: Fix[F]) = RF(FixRenderTree[F]).render(v.unFix)
```
mossprescott: Alright, obviously this is what I need to absorb. <sigh>

core/src/main/scala/quasar/semantics.scala
sellout: Yeah, I should make that private – just needed it available to both folds.

Originally, I had them in a single fold, and eliminated the prefix checking, but it makes it difficult to zip the annotations.

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -69,7 +68,13 @@ package object recursionschemes {
   //   }
 
   def cofCataM[S[_]: Traverse, M[_]: Monad, A, B](t: Cofree[S, A])(f: (A, S[B]) => M[B]): M[B] =
-    t.tail.map(cofCataM(_)(f)).sequence.flatMap(f(t.head, _))
+    t.tail.traverse(cofCataM(_)(f)).flatMap(f(t.head, _))
+
+  def cofParaM[M[_]] = new CofParaMPAT[M]
+  final class CofParaMPAT[M[_]] { self =>
+    def apply[T[_[_]]: Corecursive, S[_]: Traverse, A, B](t: Cofree[S, A])(f: (A, S[(T[S], B)]) => M[B])(implicit M: Monad[M]): M[B] =
```
mossprescott: I like this as a concept and I think it's probably worth it if it cleans up even a few annotations when these folds are used. However I notice that it obfuscates the type for the reader, since you have to drill down into the corresponding `apply`. That will be even a bigger deal for IDE users (not me at the moment), who can see types/declarations by hovering over a reference.

I can't think of any simple solution. Repeating the full type in scaladoc might help, but would be tedious and of course unchecked. Maybe we just live with it.

sellout: How about a macro that expands `cofParaM[M[_]: Monad][[T[_[_]]: Corecursive, S[_]: Traverse, A, B]](…` into this? 😆

core/src/main/scala/quasar/fp/package.scala
```diff
@@ -106,6 +106,18 @@ sealed trait TreeInstances extends LowPriorityTreeInstances {
   implicit val StringRenderTree: RenderTree[String] =
     RenderTree.fromToString[String]("String")
 
+  implicit val RU: RenderTree[Unit] = new RenderTree[Unit] {
+    def render(v: Unit) = Terminal(List("Unit"), Some("()"))
+    }
```
mossprescott: Whitespace violation ;)

core/src/test/scala/quasar/recursionschemes/spec.scala
```diff
@@ -620,31 +610,31 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
...
 
       "selves" ! Prop.forAll(expGen) { exp =>
-        Foldable[Cofree[Exp, ?]].foldMap(attrSelf(exp))(_ :: Nil) must_== exp.universe
+        Foldable[Cofree[Exp, ?]].foldMap(exp.cata(attrSelf))(_ :: Nil) must_== exp.universe
       }
     }
 
     "RenderTree" should {
       "render simple nested expr" in {
         implicit def RU = new RenderTree[Unit] { def render(v: Unit) = Terminal(List("()"), None) }
```
mossprescott: Duplicated code now?

sellout: Good catch. It was already duplicated – I just moved the main one from the Mongo planner to the fp package.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -329,45 +292,39 @@ trait Compiler[F[_]] {
...
-                  (CompilerState.rootTableReq ⊛
-                    groupBy.keys.traverseU(compile0))((src, keys) =>
-                    GroupBy(src, MakeArrayN(keys: _*))))
+                  CompilerState.rootTableReq.map(GroupBy(_, MakeArrayN(groupBy.keys.map(_._2): _*))))
 
                 stepBuilder(grouped) {
                   val having = groupBy.flatMap(_.having).map(having =>
-                    (CompilerState.rootTableReq ⊛ compile0(having))(
-                      Filter(_, _)))
+                    CompilerState.rootTableReq.map(Filter(_, having._2)))
```
mossprescott: Not crazy about all the `._1` and `._2` now. How about `def exprOf` and `def lpOf` (or maybe `source` and `compiled`) or something like that?

core/src/main/scala/quasar/variables.scala
mossprescott: Given that ticket, I don't really see the need for the TODO. It's a new feature, not some un-finished implementation detail. It's not like this code is going to be hard to find.

core/src/main/scala/quasar/sql/package.scala
mossprescott: OK, you convinced me. At some point I'd like to understand what `RenderTree ~> λ[α => RenderTree[ExprF[α]]]` means, but not necessarily today.

core/src/main/scala/quasar/semantics.scala
mossprescott: Just that it's not clear why ValidSem (`ValidationNel[SemanticError]`) is used on the inside, while the somewhat equivalent `NEL[SemErr] \/ _` is used in the outer-most fold.

core/src/main/scala/quasar/semantics.scala
mossprescott: OK, that's a bummer, but making it private would be good enough for me.

core/src/main/scala/quasar/recursionschemes/package.scala
mossprescott: Sounds great, as long as I/we don't have to debug it.

core/src/main/scala/quasar/semantics.scala
sellout: Ah, because Validation isn’t a Monad.

mossprescott:
Obviously I'm psyched to see the end of AnnotatedTree, so this is a great step even if it does come with a few warts.

Also if regression tests are failing but not planner/compiler tests, that suggests that we're missing some tests, or maybe I'm misinterpreting the Travis failure?

Can't wait to get this merged!

sellout:
Compiler tests are never being run (because of our sequential coverage stuff). They do fail.

sellout:
@mossprescott In review!

core/src/main/scala/quasar/recursionschemes/package.scala
mossprescott: By the way, we've repeatedly been bitten by this rendering not indicating fixed or not for the root, most recently last night. Maybe we should think about decorating the label? Something like:
```scala
def render(v: Fix[F]) = RF(FixRenderTree[F]).render(v.unFix).relabel("Fix(" + _ + ")")
```

Realizing that it's going to be harder to read. :-/

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -29,6 +28,37 @@ package object sql {
...
+  //             case SpliceF(_) => None
+  //             case _ =>
+  //               (alias <+> extractName(expr)).orElse(index.toString().some)
+  //           })
+  //         },
+  //         r, f, g, o)
+  //     case x => x
+  //   }
+  // }
+
```
mossprescott: Dead code.

mossprescott:
Other than the one chunk of commented code, I don't see any new issues, so there are just the various little things I pointed out earlier.

I'd love to see this merged with this version of `compile`, and come back to making it a fold in a future PR, if you're on board with that.

Setting it back to in-progress so you can respond. 

Also, be advised: when you merge from master you'll get a couple of pending tests which may or may not be fixed.

core/src/main/scala/quasar/sql/package.scala
sellout: Damn – knew I was going to forget to delete that.

sellout:
Oh yeah – definitely on board with using this version of compile0 for now.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: Ah yeah – I waffled here on whether constraining F on Unzip or Functor was better. I’m guessing people are more likely to implement Functor, so that wins, this dies.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: I think that’s a win … I was trying to figure out how to just say “hey, this whole tree is Fix” or not, but didn’t come up with anything reasonable.

core/src/main/scala/quasar/sql/package.scala
sellout: So, to kind of means “Given a `RenderTree` for some arbitrary type, I can give you a `RenderTree` for `ExprF[that arbitrary type]`.” Which sounds a lot like what the more obvious version says. (`~>` is a natural transformation, which is like a function, but between Functors instead of first-order types.) The difference is that for `FixRenderTree`, we can’t add an implicit for `RenderTree[ExprF[Fix[ExprF]]]`, because until FixRenderTree has been defined, there is no RenderTree for `Fix[ExprF]`. So instead, we make the implicit a function (well, natural transformation) that when given `FixRenderTree[ExprF]` (once we have it), will give us the `RenderTree` instance we want.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: Aww, you tricked me – I thought relabel already existed 😉

sellout: Actually, do we want to change the label, or the head of the nodeType? Changing the label gives us something like `Num(Fix(1))` whereas changing the nodeType does `Fix(Num)(1)`

sellout:
Was going to rebase this, but I’ll wait until you’ve reviewed the current commits.

core/src/main/scala/quasar/sql/package.scala
mossprescott: Thanks, that helps a bit. I understood what it was doing, but not exactly why it was needed, and that existential-type/type-lambda/kind-projector business is eluding me at the moment. I might try to desugar it just to make sure I'm not still confused.

Ah, I know why I'm confused—it's not the same `α`! This would be the same, right?
```scala
implicit val ExprFRenderTree: RenderTree ~> λ[α => RenderTree[ExprF[α]]] =
  new (RenderTree ~> λ[β => RenderTree[ExprF[β]]]) {
    def apply[A](ra: RenderTree[A]): RenderTree[ExprF[A]] = new RenderTree[ExprF[A]] {
```
(but `λ(... => ...)` is kind-projector syntax). I think I've got it.

sellout: Yep, that’s it.

core/src/main/scala/quasar/recursionschemes/package.scala
slamdata-build: You're right, we want to decorate the node-type. And `Fix(Num)(1)` actually
reads better, I think.

And we *did* have `relabel` at one point, but might have deleted it.
Obviously we need lenses for RenderedTree :)

core/src/main/scala/quasar/semantics.scala
mossprescott: Did you miss this or did it end up needing to be non-private?

core/src/main/scala/quasar/variables.scala
```diff
@@ -34,47 +32,12 @@ final case class VarValue(value: String)
...
-
-        Tree[Expr](root, _.children).annotate(map1.get(_))
-    }
+  def substVarsƒ[A](vars: Variables): ExprF[Expr] => SemanticError \/ Expr = {
+    case VariF(name) =>
+      vars.value.get(VarName(name)).fold[SemanticError \/ Expr](
+        UnboundVariable(VarName(name)).left)(
+        varValue => (new SQLParser()).parseExpr(varValue.value)
+          .leftMap(VariableParseError(VarName(name), varValue, _)))
+    case x => Fix(x).right
```
mossprescott: Sweet. -55 lines and -1 embarrassing TODO. I like it. This would have been a great example for the tech talk, except I wonder if the old implementtion was just being dumb :laughing:

core/src/test/scala/quasar/recursionschemes/spec.scala
```diff
@@ -620,47 +610,46 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
...
-            |│  ╰─ ()
-            |├─ Num(0)
-            |│  ╰─ Annotation
-            |│     ╰─ ()
-            |╰─ Num(1)
-            |   ╰─ Annotation
-            |      ╰─ ()""".stripMargin
+        mul(num(0), num(1)).cata(attrK(())).shows must_==
+          """Cofree
+            |├─ Unit(())
```
mossprescott: Really? How about just `Unit`, or better yet, just `()`?

core/src/main/scala/quasar/variables.scala
sellout: Nah – the old implementation was just written for the non-Expr-based AST, back when we had Node with Expr being one of the subtypes, so we had the complicated `exprMapUp` stuff, which was kind of like a custom paramorphism(?) for that.

core/src/main/scala/quasar/fp/package.scala
```diff
@@ -106,6 +106,18 @@ sealed trait TreeInstances extends LowPriorityTreeInstances {
   implicit val StringRenderTree: RenderTree[String] =
     RenderTree.fromToString[String]("String")
 
+  implicit val RU: RenderTree[Unit] = new RenderTree[Unit] {
+    def render(v: Unit) = Terminal(List("Unit"), Some("()"))
```
mossprescott: Yes, let's do `Terminal(List("()"), None)` here. The convention is that you only have a label if the node carries some unique value. If you want to be pedantic, you could do `Terminal(List("()", "Unit"), None)`, which would be printed the same way.

core/src/main/scala/quasar/variables.scala
mossprescott: Right, well I won't tell if you don't.

core/src/test/scala/quasar/recursionschemes/spec.scala
sellout: Hey, I didn’t write it, I just deleted the duplicated version that was shadowing the real one. But yeah, I’m on board with keeping this version instead (just in the right place).

core/src/main/scala/quasar/semantics.scala
sellout: I did miss it … I made the prefix string private, and forgot about the algebras.

mossprescott:
OK, no other comments from me. Sorry about the delay.

sellout:
So, I accidentally rebased before these most recent changes. I didn’t squash, so you should still be able to see just the most recent changes. But it does mark a lot of the comments “outdated.”

mossprescott:
👍🏾

sellout:
You’re supposed to relabel it to “ready” 😉

But, I’m also going to rename the classes with the `PAT`suffix to have a `PartiallyApplied` suffix, to match the style of non/cats#577.

mossprescott:
Seems like an improvement. :)

PR #985, opened 2015-11-11
Add new set operators to the compiler.

sellout:
SD-1012 #done

* UNION,
* UNION ALL,
* INTERSECT,
* INTERSECT ALL, and
* EXCEPT.

In the LogicalPlan, UNION and INTERSECT are implemented with the ALL
versions combined with DISTINCT.


core/src/main/scala/quasar/std/set.scala
```diff
@@ -165,6 +165,36 @@ trait SetLib extends Library {
...
+    noSimplification,
+    setTyper(partialTyper {
+      case List(Type.Const(Data.Set(Nil)), s2) => s2
+      case List(s1, Type.Const(Data.Set(Nil))) => s1
+      case List(s1, s2)                        => s1 ⨿ s2
+    }),
+    setUntyper(t => success(t :: Nil)))
+
+  val Intersect = Transformation("(INTERSECT ALL)",
+    "Creates a new set with only the elements that exist in both input sets, keeping duplicates.",
```
jedesah: So if an element is present once in the first set and twice in the second set, it will be found one in the resulting set?

core/src/main/scala/quasar/std/set.scala
```diff
@@ -165,6 +165,36 @@ trait SetLib extends Library {
...
+    Type.Top, Type.Top :: Type.Top :: Nil,
+    noSimplification,
+    setTyper(partialTyper {
+      case List(s1, s2) => if (s1 == s2) s1 else Type.Const(Data.Set(Nil))
```
jedesah: Is this really the correct type in the case where they are not equal? Is it not implemented or something?

core/src/main/scala/quasar/std/set.scala
```diff
@@ -165,6 +165,36 @@ trait SetLib extends Library {
...
+    }),
+    setUntyper(t => success(t :: Nil)))
+
+  val Except = Transformation("(EXCEPT)",
+    "Removes the elements of the second set from the first set.",
+    Type.Top, Type.Top :: Type.Top :: Nil,
+    noSimplification,
+    setTyper(partialTyper {
+      case List(s1, Type.Const(Data.Set(Nil))) => s1
+      case List(s1, s2) => if (s1 == s2) s1 else Type.Const(Data.Set(Nil))
```
jedesah: Could be the same misunderstanding as above, could we not say here that the return type would be `s1` no matter what?

core/src/main/scala/quasar/std/set.scala
sellout: This is a very good question. Right now, this isn’t implemented one way or the other, but we should settle the issue.

Since a non-ALL INTERSECT exists, we know that if it’s duplicated in both, it should appear twice. Not sure about duplicated in one.

core/src/main/scala/quasar/std/set.scala
```diff
@@ -178,6 +208,6 @@ trait SetLib extends Library {
     },
     setUntyper(t => success(t :: Type.Top :: Nil)))
 
-  def functions = Take :: Drop :: OrderBy :: Filter :: InnerJoin :: LeftOuterJoin :: RightOuterJoin :: FullOuterJoin :: GroupBy :: Distinct :: DistinctBy :: Nil
+  def functions = Take :: Drop :: OrderBy :: Filter :: InnerJoin :: LeftOuterJoin :: RightOuterJoin :: FullOuterJoin :: GroupBy :: Distinct :: DistinctBy :: Union :: Intersect :: Except :: Constantly :: Nil
```
jedesah: Where does `Constantly` come from?

core/src/main/scala/quasar/std/set.scala
jedesah: Okay, I got it.

core/src/main/scala/quasar/std/set.scala
jedesah: idem.

sellout: Yeah, I think you’re right – `Except` should always return the type of the first argument.

core/src/test/scala/quasar/compiler.scala
```diff
@@ -508,6 +508,69 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
                 Constant(Data.Int(100))))))
     }
 
+    "compile union" in {
```
jedesah: Can we extract the invariants in these tests into predefined values? It creates a lot of noise and makes it hard to review :stuck_out_tongue:

core/src/main/scala/quasar/std/set.scala
sellout: It already existed, just wasn’t exported. Whether or not it needs to be exported (for use as a SQL function) is debatable, but thus far we’ve been exporting everything, even if it doesn’t reasonably get used from SQL.

core/src/main/scala/quasar/std/set.scala
sellout: @jdegoes: Do you know what the semantics here should be?

jdegoes:
```
X = [1, 2, 3]
Y = [1, 9, 2]

X INTERSECT ALL Y = [1, 1, 2]
```

In reality this is not quite precise, because I am showing values and not identities. The identities of the two separate occurrences of '1' are different, even if their values are the same.

jdegoes: Oh, and one more:

```
X = [1, 1, 2, 3]
Y = [1, 9, 2]

X INTERSECT ALL Y = [1, 1, 1, 2]
```

These are not all the same `1`'s, of course.

sellout:
updated.

sellout:
Sorry, taking this back to in-progress to change the semantics of INTERSECT so that implementing it the way John describes is possible.

jedesah:
LGTM.

PR #986, opened 2015-11-12
[SD-983] Filesystem algebras and tests

wemrysi:
This is a mostly additive PR containing the new filesystem algebras, their MongoDB interpreters, and all the refactored tests to go along with them. Neither the rest API or REPL are using the new interpreters yet and this PR is primarily for integration and to try and break up the changes a bit so that there isn't a single enormous PR for all of SD-983 (though I know this one is pretty big).

Once the API and REPL PRs are submitted, a final PR will be submitted that deletes all the infrastructure being replaced and renames the `XXX2` variants to `XXX`.

Also, after reviewing @mossprescott 's views PR some concepts clarified a bit and as a result I am going to change the distribution of algebra terms a bit, merging the read-only bits of `ManageFile` with `ExecutePlan` into a new algebra called `QueryFile`. This doesn't change anything in this PR in a major semantic way, other than shuffling of where tests and some functions live and will tidy up some of the types a bit. So it doesn't need to hold up this PR, which is why I'm not waiting.

A rough outline of what's in here
* `ReadFile`, `WriteFile`, `ManageFile` and `ExecutePlan` algebras, describing the filesytem operations
* `MongoDb` a thin wrapper around the async `mongo-java-driver`.
* Interpreters for all the filesystem algebras in terms of `MongoDb`.
* Conversion to pathy paths in all public filesystem apis (`quasar.fs.Path` is still used internally in a few places)
* An in-memory filesystem interpreter useful for tests
* Unit tests for all derived filesystem operations
* A file system specification in the form of implementation agnostic integration tests, the idea being that if an implementation passes these tests, it should be a valid filesystem.
* Implementation-specific filesystem tests for the MongoDb implementation
* A side-by-side reimplementation of the regression test runner that is a bit more performant (though MongoDB is still the bottleneck here, I believe)
* A thread-safe `TaskRef` cribbed from oncue/remotely
* Various supporting functions in `quasar.fp` for interpreting free monads based on coproducts of functors
* Broke apart regression test classes into own package and files for easier reuse.


wemrysi:
@sellout got the short straw of reviewing this, but @mossprescott, @drostron, @jedesah, @syves I'd appreciate any feedback you have as well, given the scope of the changes.

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
```diff
@@ -0,0 +1,241 @@
...
+
+import scalaz.{Optional => _, _}, Id._
+import scalaz.stream._
+import scalaz.std.vector._
+import scalaz.syntax.monad._
+import scalaz.syntax.show._
+import scalaz.concurrent.Task
+
+/** Unit tests for the MongoDB filesystem implementation. */
+class MongoDbFileSystemSpec
```
mossprescott: Like all the "it" tests, these were meant be backend-agnostic. The fact that one of them currently assumes knowledge MongoDB semantic and of the way queries are compiled for MongoDB is temporary. As we add backends, we'll expand the tested queries to cover the interesting cases. The point is to have a suite of tests to verify that we have consistent behavior across all the backends, and across each release.

So I think this test should be in the common package, and run on all backends.

.travis.yml
```diff
@@ -3,10 +3,7 @@ scala:
 - 2.11.7
 jdk: oraclejdk8
 sudo: false
-before_install:
-- "./scripts/installMongo mongodb-linux-x86_64-2.6.9 mongo2.6 27018"
-- "./scripts/installMongo mongodb-linux-x86_64-3.0.4 mongo3.0 27019"
-script: "./scripts/build"
+script: "export MONGO_LINUX=\"MONGO_${MONGO_RELEASE}_LINUX\"; ./scripts/installMongo mongodb-linux-x86_64-${!MONGO_LINUX} mongo 27018 && ./scripts/build $MONGO_RELEASE"
```
sellout: This could be written
```bash
MONGO_LINUX=\"MONGO_${MONGO_RELEASE}_LINUX\" ./scripts/installMongo mongodb-linux-x86_64-${!MONGO_LINUX} mongo 27018
```
(no export).

Also, I think it makes sense to keep the install of Mongo in `before_install` rather than merging it into the build script. I think failures get handled differently, and also the output is collapsed during the build.

core/src/main/scala/quasar/EnvironmentError.scala
```diff
@@ -0,0 +1,138 @@
...
+import argonaut._, Argonaut._
+
+import monocle._
+
+import scalaz._
+import scalaz.syntax.show._
+
+sealed trait EnvironmentError2
+
+object EnvironmentError2 {
```
sellout: So, as we discussed in the hangout (just for the record) …

the case classes should be in a deeper object, possibly `Cases` as mentioned in the wiki. Then the wrapper objects could live in the companion object.

However … I think I’m also content with just getting rid of the wrapper objects and only having the prisms around. But also please write up the prism approach for the “Faking Sum Types” wiki page. Still so much boilerplate 😢

core/src/main/scala/quasar/NameGenerator.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import scalaz._
+import scalaz.concurrent.Task
+import scalaz.syntax.functor._
+
+/** A source of strings unique within `F[_]`, an implementation must have the
+  * property that, if Applicative[F], then (freshName |@| freshName)(_ != _).
+  */
+@typeclass
+trait NameGenerator[F[_]] {
```
sellout: We have at least one other (and probably more) NameGen implementations floating around. I tried to unify them at some point, but this looks better. Have you replaced the other(s)?

core/src/main/scala/quasar/Predef.scala
```diff
@@ -28,6 +28,8 @@ object Predef extends LowPriorityImplicits {
   type tailrec = scala.annotation.tailrec
   type SuppressWarnings = java.lang.SuppressWarnings
 
+  type AnyVal = scala.AnyVal
```
sellout: I’m cautiously positive about this … I know it’s used for having unboxed classes, but are we going to get bit by people using it as a parameter type or anything? This is why we exclude `Any` and `AnyRef`, requiring them to be fully qualified if used.

core/src/main/scala/quasar/backend.scala
```diff
@@ -23,36 +23,14 @@ import quasar.Errors._
...
-    override def toString = name + "\n" + value
-  }
-
-  implicit def PhaseResultEncodeJson: EncodeJson[PhaseResult] = EncodeJson {
-    case Tree(name, value)   => Json.obj("name" := name, "tree" := value)
-    case Detail(name, value) => Json.obj("name" := name, "detail" := value)
-  }
-}
-
 sealed trait Backend { self =>
```
sellout: I saw you mark the Backend → FileSystem rename as done or in review or something … is that in an upcoming PR?

core/src/main/scala/quasar/fp/TaskRef.scala
```diff
@@ -0,0 +1,41 @@
...
+import scalaz.syntax.id._
+import scalaz.concurrent.Task
+
+/** A thread-safe, atomically updatable mutable reference.
+  *
+  * Cribbed from the `IORef` defined in oncue/remotely, an Apache 2 licensed
+  * project: https://github.com/oncue/remotely
+  *
+  */
+sealed abstract class TaskRef[A] {
```
sellout: Hahaha – I stole this from OnCue a couple months ago, to use in server.scala. Ended up not needing it (or, I probably did need it, but tried to avoid it anyway). Looks like it’s inevitable, though.

core/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -0,0 +1,98 @@
...
+  private final case class PlannerError0(lp: Fix[LogicalPlan], err: PlannerErr)
+    extends FileSystemError
+  private final case class UnknownReadHandle0(h: ReadHandle)
+    extends FileSystemError
+  private final case class UnknownWriteHandle0(h: WriteHandle)
+    extends FileSystemError
+  private final case class PartialWrite0(numFailed: Int)
+    extends FileSystemError
+  private final case class WriteFailed0(data: Data, reason: String)
+    extends FileSystemError
```
sellout: Don’t make these private, so we can derive instances.

core/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -0,0 +1,98 @@
...
+import scalaz.syntax.std.option._
+import scalaz.syntax.show._
+
+import ReadFile.ReadHandle
+import WriteFile.WriteHandle
+
+sealed trait FileSystemError {
+  import FileSystemError._
+
+  def fold[X](
```
sellout: This seems worse than pattern matching to me. It loses all context about which case is which. `fold` with more than two cases is meh.

jedesah: Agreed! And I believe @wemrysi agrees as well after a conversation we had the other week. Just never got around to changing it.

core/src/main/scala/quasar/fs/ManageFile.scala
```diff
@@ -0,0 +1,155 @@
...
+
+import scalaz._, Scalaz._
+
+sealed trait ManageFile[A]
+
+object ManageFile {
+  sealed trait MoveSemantics {
+    import MoveSemantics._
+
+    def fold[X](overwrite: => X, failIfExists: => X, failIfMissing: => X): X =
```
sellout: Same here.

core/src/main/scala/quasar/fs/ManageFile.scala
```diff
@@ -0,0 +1,155 @@
...
+    /** Indicates the move should (atomically, if possible) fail unless
+      * the destination exists, overwriting it otherwise.
+      */
+    val FailIfMissing: MoveSemantics = FailIfMissing0
+  }
+
+  sealed trait MoveScenario {
+    import MoveScenario._
+
+    def fold[X](
```
sellout: And here.

I know I said “two is ok”, but since it doesn’t scale, I don’t really see a reason to use it here.

core/src/main/scala/quasar/fs/ManageFile.scala
```diff
@@ -0,0 +1,155 @@
...
+        case DirToDir0(sd, dd)   => d2d(sd, dd)
+        case FileToFile0(sf, df) => f2f(sf, df)
+      }
+  }
+
+  object MoveScenario {
+    private final case class DirToDir0(src: AbsDir[Sandboxed], dst: AbsDir[Sandboxed])
+      extends MoveScenario
+    private final case class FileToFile0(src: AbsFile[Sandboxed], dst: AbsFile[Sandboxed])
+      extends MoveScenario
```
sellout: Should be non-private but nested … I’m guessing comments like this, `fold`, etc. are going to crop up repeatedly, so just try to catch all of them, and I’ll try not to repeat myself anymore.

.travis.yml
wemrysi: I tried it without the `export` previously and the `MONGO_LINUX` variable didn't seem to be available for indirect expansion with `!` in the same line. Though, I may have also added the semicolon at the same time, so I'll try leaving that and removing export and see if it still works.

I would also prefer to use `before_install`, but it doesn't seem like the `MONGO_RELEASE` variable is made available during the `before_install` phase (it dereferenced to empty in my attempts) so I had to move it into the `script` phase.

core/src/main/scala/quasar/fp/TaskRef.scala
wemrysi: Yeah, I was one of the proponents of avoiding it, but after trying to use what we previously had, I realized it wasn't thread-safe =( Which I'm guessing was what motivated this for the oncue folks as well.

core/src/main/scala/quasar/fs/FileSystemError.scala
wemrysi: Fair enough, though in this particular instance the types are different for every case, so you can tell pretty easily. For types where the cases share many types, I'd agree that relying on argument position isn't as nice as having the constructor name.

.travis.yml
jedesah: I would postulate that whenever a reviewer suggests a change and the author explains why that change is not possible, the action item should be to add a similar explanation in the form of documentation so as to avoid future maintainers from asking themselves the same question when looking at a particular piece of code as it has proved already to be at least somewhat deceiving. :smile:

I would even go so far as saying that explaining the reason why in the PR comments should be avoided so as to make sure the comment in the code is clear enough so that the reviewer can understand the reason from that alone.

core/src/main/scala/quasar/fs/ManageFile.scala
wemrysi: Aye, I have a task to go through and normalize all of these based on our discussion from 11/13/15.

.travis.yml
sellout: Mostly agreed, but I think doing it in PR comments first is good, because we can arrive at the right explanation, or come up with a solution that doesn’t need an explanation, rather than (more slowly) creating a series of commits that refine/remove the comment. Maybe something like:

> TBI:
> ```
> Should be in `before_install`, but `MONGO_RELEASE` is only set during `script` 😝
> ```

Where “TBI” means “to be inserted”. There may already be a convention for this kind of thing that I’ unfamiliar with?

jedesah: @sellout That's a good point. I retract my second paragraph.

core/src/main/scala/quasar/EnvironmentError.scala
wemrysi: Will do, though how about `Case` for the object name? It seems to read better to me, i.e. `Foo.Case.Bar` vs `Foo.Cases.Bar`?

core/src/main/scala/quasar/NameGenerator.scala
wemrysi: Not yet as some will be deleted alltogether after the rest of the client code refactoring is done, however my plan is to convert any usages that remain once that's done.

core/src/main/scala/quasar/Predef.scala
wemrysi: I don't mind omitting it from `Predef` and just fully-qualifying at use-sites. I wasn't sure if it was missing from `Predef` intentionally or just because we hadn't needed it yet.

core/src/main/scala/quasar/backend.scala
wemrysi: I marked it due to https://github.com/quasar-analytics/quasar/pull/986/files#diff-6b7c93201e526ada666ff5b87e56b24cR15. `Backend` will be deleted, but I'll wait to close the issue until that actually happens.

.travis.yml
wemrysi: I agree with both of you, this particular change is still in progress, which is why I haven't documented anything yet, as it doesn't actually work =(. @jedesah had the good idea of opening a separate PR to mess with the build to reduce the churn on this one (see #988). Once I've got that working, I'll make a single commit with the correct config back to this PR and close the other one.

wemrysi: On that note, any idea why we're using `before_install` to install stuff rather than the `install` phase?

sellout: Ignorance?

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -0,0 +1,32 @@
...
+  type WriteFileF[A]  = Coyoneda[WriteFile, A]
+  type ManageFileF[A] = Coyoneda[ManageFile, A]
+  type QueryFileF[A]  = Coyoneda[QueryFile, A]
+
+  type FileSystem0[A] = Coproduct[WriteFileF, ManageFileF, A]
+  type FileSystem1[A] = Coproduct[ReadFileF, FileSystem0, A]
+  type FileSystem[A]  = Coproduct[QueryFileF, FileSystem1, A]
+
+  type RelPath[S] = RelDir[S] \/ RelFile[S]
+  type AbsPath[S] = AbsDir[S] \/ AbsFile[S]
```
sellout: I think this came up before, but why are these not `type AbsPath[S] = Path[Abs, _, S]`?

core/src/main/scala/quasar/fs/ManageFile.scala
```diff
@@ -0,0 +1,155 @@
...
+  final case class Move(scenario: MoveScenario, semantics: MoveSemantics)
+    extends ManageFile[FileSystemError \/ Unit]
+
+  final case class Delete(path: AbsPath[Sandboxed])
+    extends ManageFile[FileSystemError \/ Unit]
+
+  final case class TempFile(nearTo: Option[AbsFile[Sandboxed]])
+    extends ManageFile[AbsFile[Sandboxed]]
+
+  final class Ops[S[_]](implicit S0: Functor[S], S1: ManageFileF :<: S) {
```
sellout: Is it possible to move the Functor constraint down to operations that require it? Although, at a glimpse, it looks like all of them, so … ¯\_(ツ)_/¯

wemrysi: It is possible, but every method would end up requiring it, so I factored it out. I think we'll be able to get rid of it entirely once we've updated to scalaz 7.2.

core/src/main/scala/quasar/fs/Natural.scala
```diff
@@ -0,0 +1,51 @@
...
+  val _8: Natural = new Natural(8)
+  val _9: Natural = new Natural(9)
+
+  def fromPositive(n: Positive): Natural =
+    new Natural(n.run)
+
+  implicit val naturalAddition: Monoid[Natural] =
+    Monoid.instance(_ + _, _0)
+
+  implicit val naturalMultiplication: Monoid[Natural @@ Mult] =
```
sellout: This is the kind of thing that makes me want to use non/spire. We had it as a dependency once, but then law checking informed me that my Ring wasn’t a Ring, so I removed the one use and dependency. But Natural is a Rig.

jedesah: I also want to explore the use of https://github.com/fthomas/refined to replace this as well.

sellout: Yeah, it’d be nice to not have to create type-level naturals in the first place 😄

core/src/main/scala/quasar/fs/Node.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def file: Option[RelFile[Sandboxed]] =
+    fold(κ(none[RelFile[Sandboxed]]), _.toOption)
+
+  def path: RelPath[Sandboxed] =
+    fold(\/.left, ι)
+}
+
+object Node {
+  private final case class Mount0(d: RelDir[Sandboxed]) extends Node
```
sellout: Don’t views require us to generalize this to RelPath[Sandboxed]? (@mossprescott)

core/src/main/scala/quasar/fs/Natural.scala
```diff
@@ -0,0 +1,51 @@
+package quasar
+package fs
+
+import quasar.Predef._
+
+import scalaz._
+import scalaz.Tags.{Multiplication => Mult}
+
+final class Natural private (val run: Long) {
```
sellout: Shouldn’t this extend AnyVal?

core/src/main/scala/quasar/fs/Positive.scala
```diff
@@ -0,0 +1,47 @@
+package quasar
+package fs
+
+import quasar.Predef._
+
+import scalaz._
+import scalaz.Tags.{Multiplication => Mult}
+
+final class Positive private (val run: Long) {
```
sellout: I really like that you’re making more smaller files, including breaking apart recursionschemes. We need to be better about this in general.

sellout: Also, yes, I’m more than sold on trying refined at this point in the PR 😉

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -0,0 +1,194 @@
...
+        val cleanup = (hoistFS(M.deleteFile(tmp)): ExecM[Unit])
+                        .liftM[Process].drain
+        values(tmp) onComplete cleanup
+      }
+
+      execute_(plan).liftM[Process] flatMap (_.fold(values, handleTemp))
+    }
+
+    /** Returns the path to the result of executing the given SQL^2 query. */
+    def executeQuery(query: sql.Expr, vars: Variables, out: AbsFile[Sandboxed])
```
sellout: I don’t think it’s feasible yet, but this is something we eventually want to get rid of, right? (Once variables are applied to LogicalPlan, etc.)

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -0,0 +1,194 @@
...
+          .flatMap(_.path.fold(
+            d => lsR(desc </> d),
+            f => Node.File(desc </> f).point[S]))
+
+      lsR(currentDir).foldLeft(Set.empty[Node])(_ + _)
+    }
+
+    /** Returns whether the given file exists. */
+    def fileExists(file: AbsFile[Sandboxed]): F[Boolean] = {
+      // TODO: Add fileParent[B, S](f: Path[B, File, S]): Path[B, Dir, S] to pathy
```
sellout: Add an issue to pathy for this if you haven’t already (and reference it here).

core/src/main/scala/quasar/fs/ReadFile.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  object ReadHandle {
+    implicit val readHandleShow: Show[ReadHandle] =
+      Show.showFromToString
+
+    implicit val readHandleOrder: Order[ReadHandle] =
+      Order.orderBy(_.run)
+  }
+
+  final case class Open(file: AbsFile[Sandboxed], offset: Natural, limit: Option[Positive])
```
sellout: Why can’t the limit be zero? I mean, I don’t know why you’d want to use a limit of zero, but it seems at least legit. If there is a practical reason, it’d be great to see a comment, especially since Positive is new code we have to maintain.

core/src/main/scala/quasar/fs/ReadFile.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    /** Returns a read handle for the given file, positioned at the given
+      * zero-indexed offset, that may be used to read chunks of data from the
+      * file. An optional limit may be supplied to restrict the total amount of
+      * data able to be read using the handle.
+      *
+      * Care must be taken to `close` the returned handle in order to avoid
+      * potential resource leaks.
+      */
+    def open(file: AbsFile[Sandboxed], offset: Natural, limit: Option[Positive]): M[ReadHandle] =
```
sellout: These are going to be privatized, right?

core/src/main/scala/quasar/fs/Node.scala
mossprescott: Yes, views are (currently) Mount nodes and have file paths, so this will not work at the moment.

Actually, we need to figure something out here because the front-end currently treats all "mount" nodes as a type of directory, so we need to have a fourth type (or else make mount/plain and file/dir independent flags.) Just using `RelPath` is probably fine for now.

core/src/main/scala/quasar/fs/ReadFile.scala
wemrysi: Sort of, I'm actually moving them to another ops class called `Unsafe`, which `Ops` makes use of. The reason being is that these are the cases we actually need to test for real backends as the rest of `Ops` are all derived operations. However, I think this still solves the problem of delineating between the nice operations and the "dangerous" ones.

core/src/main/scala/quasar/fs/ReadFile.scala
jedesah: I like `Positive` here. I certainly prefer it over this: https://github.com/quasar-analytics/quasar/blob/master/core/src/main/scala/quasar/backend.scala#L126

core/src/main/scala/quasar/fs/Node.scala
sellout: Yeah, this was discussed at least between @jdegoes and me a while back – they should be orthogonal. This will affect the API for at least `ls`, and the front-end will need to make changes.

core/src/main/scala/quasar/fs/ReadFile.scala
sellout: Yes, it’s definitely better than that – the alternative to Positive is Natural.

wemrysi: As for why nonzero, see https://github.com/quasar-analytics/quasar/blob/3b61cc5f7f1b25cbb17b39dc1fd5000d13c93c74/core/src/main/scala/quasar/backend.scala#L126

I was not seeking to change too many semantics (i.e. the existing implementation was already limited to positive numbers), but it looks like maybe the existing reason isn't a very good one (as it is mongodb-specific).

Even if we do agree that it doesn't need to be positive, I think I'd prefer changing it after this PR as it is an orthogonal concern since it isn't deviating from the status-quo.

sellout: Yeah, I’m fine with postponing the decision. I didn’t realize the driver required it, so that does make it a bit more complicated.

mossprescott: What is the semantics if limit is 0? Run the query, capture errors, and always produce no results? Seems perverse at best. We could support it by adding code to explicitly discard the results, if we thought that was useful to someone.

The point of catching it is mostly to be able to produce precise errors in the API, but I believe the API separately checks at the moment.

core/src/main/scala/quasar/fs/inmemory.scala
```diff
@@ -0,0 +1,242 @@
...
+import scalaz.concurrent.Task
+import pathy.Path._
+
+/** In-Memory Read/Write/ManageFile interpreters, useful for testing/stubbing
+  * when a "real" interpreter isn't needed or desired.
+  *
+  * NB: Since this is in-memory, careful with writing large amounts of data to
+  *     the file system.
+  */
+object inmemory {
```
sellout: Can you re-capitalize this? `inMemory`, maybe?

core/src/main/scala/quasar/package.scala
```diff
@@ -0,0 +1,47 @@
...
+import quasar.recursionschemes._, Fix._, FunctorT.ops._
+
+import scalaz._
+import scalaz.syntax.monad._
+import scalaz.syntax.either._
+import scalaz.syntax.writer._
+import scalaz.syntax.nel._
+import scalaz.std.vector._
+
+package object quasar {
```
sellout: I think this is fine for now, but we really need to have more packages inside `quasar` and most (all?) of these should be moved down one level.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -0,0 +1,194 @@
...
+
+    /** Returns the path to the result of executing the given SQL^2 query. */
+    def executeQuery(query: sql.Expr, vars: Variables, out: AbsFile[Sandboxed])
+                    : CompExecM[ResultFile] = {
+
+      compileAnd(query, vars)(execute(_, out))
+    }
+
+    /** Returns the path to the result of executing the given SQL^2 query
+      * using the given output file if possible.
```
mossprescott: Scaladoc got mixed up? This is the one that _doesn't_ take an output path.

core/src/main/scala/quasar/fs/WriteFile.scala
```diff
@@ -0,0 +1,240 @@
...
+    /** Create the given file with the contents of `src`. Fails if already exists. */
+    def create(dst: AbsFile[Sandboxed], src: Process[F, Data])
+              (implicit QF: QueryFile.Ops[S], MF: ManageFile.Ops[S])
+              : Process[M, FileSystemError] = {
+
+      createChunked(dst, src map (Vector(_)))
+    }
+
+    /** Same as `create` but accepts a `Foldable` of [[Data]]. */
+    def createF[H[_]: Foldable](dst: AbsFile[Sandboxed], data: H[Data])
```
mossprescott: This seems like about one or two more variants of each method than we probably really need. Are we planning to write tests to cover every variant?

It seems like we could provide only the chunked version and leave it up to the caller to use `chunked`, `processF`, or whatever.

core/src/main/scala/quasar/fs/WriteFile.scala
```diff
@@ -0,0 +1,240 @@
...
+    def open(file: AbsFile[Sandboxed]): M[WriteHandle] =
+      EitherT(lift(Open(file)))
+
+    /** Write a chunk of data to the file represented by the write handle.
+      *
+      * Attempts to write as much of the chunk as possible, even if parts of
+      * it fail, any such failures will be returned in the output `Vector`
+      * An empty `Vector` means the entire chunk was written successfully.
+      */
+    def write(h: WriteHandle, chunk: Vector[Data]): F[Vector[FileSystemError]] =
```
mossprescott: This has always been kind of an unsatisfying way of returning errors. I can think of two things that can come back: trying to save non-obj values in MongoDB, and random runtime errors. Those seem like two very different sorts of problems, one being either a client error or a server bug, depending on how you look at it, and the other suggesting that the whole request maybe should be treated as a failure. Unfortunately I don't have a better idea.

core/src/main/scala/quasar/fs/inmemory.scala
```diff
@@ -0,0 +1,242 @@
...
+    def apply[A](wf: WriteFile[A]) = wf match {
+      case WriteFile.Open(f) =>
+        for {
+          i <- nextSeq
+          h =  WriteHandle(i)
+          _ <- wFileL(h) := Some(f)
+          _ <- fileL(f) %= (_ orElse Some(Vector()))
+        } yield h.right
+
+      case WriteFile.Write(h, xs) =>
```
mossprescott: Presumably this doesn't do anything clever to simulate backend semantics around concurrent writes, correct? Put another way, is this a correct implementation of the semantics assumed by append/save/replace, etc.?

What about edge cases like appending to a file and another request deletes it half-way?

core/src/main/scala/quasar/physical/mongodb/MongoDb.scala
```diff
@@ -0,0 +1,351 @@
...
+import com.mongodb.bulk.BulkWriteResult
+import com.mongodb.client.model._
+import com.mongodb.async._
+import com.mongodb.async.client._
+
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+import scalaz.stream._
+
+final class MongoDb[A] private (protected val r: ReaderT[Task, MongoClient, A]) {
```
mossprescott: Not sure `MongoDb` is the right name for this, just because there are so many things called "mongo(db)" already.  However, thinking of it as the monad in which MongoDB effects are captured, the name is attractive.

What would we want this to be called in the separate project that eventually contains our stand-alone FP interface to MongoDB? That will have `Workflow`, `WorkflowTask`, `Bson`, and the workflow executor, but not `WorkflowBuilder` or anything that knows about LogicalPlan.

core/src/main/scala/quasar/package.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  type CompileM[A] = SemanticErrsT[PhaseResultW, A]
+
+  type EnvErr2T[F[_], A] = EitherT[F, EnvironmentError2, A]
+
+  type SeqNameGeneratorT[F[_], A] = StateT[F, Long, A]
+  type SaltedSeqNameGeneratorT[F[_], A] = ReaderT[SeqNameGeneratorT[F, ?], String, A]
+
+  /** Returns the `LogicalPlan` for the given SQL^2 query. */
+  def queryPlan(query: sql.Expr, vars: Variables): CompileM[Fix[LogicalPlan]] = {
```
sellout: This should really live in `quasar.sql`. I know you’re just creating new files that maintain structure that already exists, but it makes it noticeable 😊

sellout: Actually, sorry, this should be split … the first part (including `logical <- …`) should be in `quasar.sql`. And the last two operations (simplifying and typechecking) are the primary function of the compiler core.

core/src/main/scala/quasar/physical/mongodb/MongoDb.scala
sellout: Agreed here. BTW, my name for that project (which I’ve never said out loud) is “Mongflow”. Quick, come up with something less terrible.

core/src/main/scala/quasar/physical/mongodb/mapreduce.scala
```diff
@@ -17,10 +17,15 @@
 package quasar.physical.mongodb
 
 import quasar.Predef._
+import quasar.fs.Positive
 import quasar.javascript._
 
 import com.mongodb._
+
 import monocle.macros.{GenLens}
+
```
sellout: We don’t put line-breaks between 3rd-party libs. We have three blocks: quasar, java/scala, 3rd-party.

Minor, but since you’re explicitly adding them, figured I’d clarify (it’s in the wiki).

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -0,0 +1,194 @@
...
+
+      def comp = compToCompExec(queryPlan(query, vars)).liftM[Process]
+
+      comp flatMap (lp => evaluate(lp).translate[CompExecM](execToCompExec))
+    }
+
+    /** Returns immediate children of the given directory, fails if the
+      * directory does not exist.
+      */
+    def ls(dir: AbsDir[Sandboxed]): M[Set[Node]] =
```
mossprescott: Bit of a stretch to consider this part of the Query API, I think. I would have looked for it in ManageFile, unless we imagine that query and `ls` will have more in common in the future.

core/src/main/scala/quasar/physical/mongodb/mapreduce.scala
```diff
@@ -96,6 +111,58 @@ object MapReduce {
...
+    * @param inputSortCriteria Criteria to use to sort the input documents.
+    * @param useJsMode Whether to avoid converting intermediate values to
+    *                  BSON, leaving them as JavaScript objects instead.
+    *                  Setting this to `true` has implications on the size of
+    *                  the input, see the MongoDB `mapReduce` documentation
+    *                  for details.
+    * @param verboseResults Whether to include additional information, such
+    *                       as timing, in the results.
+    */
+  final case class Config(
```
sellout: I’m confused about this class – it seems to duplicate `MapReduce`, but it can’t replace it as types don’t match up with ones we can manipulate – e.g., using third-party BSON types instead of our Js and Selector types.

It seems we can convert directly from MapReduce to what the underlying driver wants (as we’ve been doing thus far), and if we _do_ need some IR, it should be scoped to the evaluator (which should eventually have its own package), not next to MapReduce.

core/src/main/scala/quasar/physical/mongodb/workflowtask/ast.scala
```diff
@@ -76,8 +80,13 @@ object WorkflowTaskF {
         }
     }
 
-  implicit def WorkflowTaskRenderTree(implicit RC: RenderTree[Collection], RO: RenderTree[WorkflowF[Unit]], RJ: RenderTree[Js], RS: RenderTree[Selector]) =
+  implicit def WorkflowTaskRenderTree: RenderTree[WorkflowTaskF[_]] =
```
sellout: 👍🏾

core/src/main/scala/quasar/recursionschemes/cofree.scala
```diff
@@ -0,0 +1,50 @@
...
+package quasar
+package recursionschemes
+
+import quasar.Predef._
+import RenderTree.ops._
+
+import scalaz._
+import scalaz.syntax.functor._
+
+object cofree {
```
sellout: I like that this is separated out, but not that we have to import it explicitly. Can’t we do `package object recursionschemes extends cofree`? And same with `object free`.

core/src/main/scala/quasar/recursionschemes/cofree.scala
```diff
@@ -0,0 +1,50 @@
...
+  implicit def cofreeRecursive[A]: Recursive[Cofree[?[_], A]] =
```
sellout: Style point/question: We’ve been naming instances using UpperCamelCase, so the trait name is unmodified (e.g., `CofreeRecursive`). I’m not opposed to changing it, but 1. we should document our preferred style, 2. hopefully with a reference to where that style came from (Scalaz, Cats, etc.). If it’s described in the official Scala style guide, then we don’t have to document it, and you can just tell me I’m wrong.

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -28,45 +27,6 @@ import simulacrum.typeclass
 
 /** Generalized folds, unfolds, and refolds. */
 package object recursionschemes {
```
sellout: I very much appreciate this reorg.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -0,0 +1,148 @@
...
+import scalaz.std.vector._
+import scalaz.syntax.monad._
+import scalaz.syntax.std.option._
+import scalaz.stream._
+import scalaz.concurrent.Task
+
+trait FileSystemFixture {
+  import FileSystemFixture._, inmemory._
+
+  // NB: These are mostly to make composition/implicit search nicer
```
sellout: 😢 I feel like this deserves a full :iron_eyes_cody:, but that doesn’t seem to be part of Unicode yet.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -0,0 +1,148 @@
...
+  type F[A]   = Free[FileSystem, A]
+  type RW[A]  = ReadWriteT[InMemoryFs, A]
+  type G[A]   = StateT[Task, InMemState, A]
+  type RWG[A] = ReadWriteT[G, A]
+  // NB: These two are the only ones that appear in return types
+  type M[A]   = FileSystemErrT[G, A]
```
sellout: Also, with single-char type aliases, it’s confusing to read these – e.g., `FileSystemErrT[G, A]` looks like it takes two arguments. One thing we’ve done in some places is to use Greek for the type variables.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -0,0 +1,148 @@
...
+import scalaz.stream._
+import scalaz.concurrent.Task
+
+trait FileSystemFixture {
+  import FileSystemFixture._, inmemory._
+
+  // NB: These are mostly to make composition/implicit search nicer
+  type F[A]   = Free[FileSystem, A]
+  type RW[A]  = ReadWriteT[InMemoryFs, A]
+  type G[A]   = StateT[Task, InMemState, A]
```
sellout: Yeah, these single-char names are quite confusing – what about `FreeFS`, `InMemRW`, `InMemStateTask`, etc.?

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -0,0 +1,148 @@
...
+  type RWG[A] = ReadWriteT[G, A]
+  // NB: These two are the only ones that appear in return types
```
sellout: This is false – at least `G` and `RWG` are used in return types (and `F` and `RW` depending on what part of `F ~> G` you consider a return type).

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  val hoistFs: InMemoryFs ~> G =
+    Hoist[StateT[?[_], InMemState, ?]].hoist(pointNT[Task])
+
+  val hoistRW: RW ~> RWG =
+    Hoist[StateT[?[_], ReadWrites, ?]].hoist(hoistFs)
+
+  val run: F ~> G =
+    hoistFs compose[F]
+    hoistFree(interpretFileSystem(queryFile, readFile, writeFile, manageFile))
```
sellout: This line needs to be indented to show that it’s part of the expression on the previous line.

core/src/test/scala/quasar/fs/PathyGen.scala
```diff
@@ -0,0 +1,49 @@
...
+package fs
+
+import quasar.Predef._
+
+import scalaz.syntax.functor._
+import scalaz.std.list._
+import org.scalacheck._
+import pathy.Path._
+
+object PathyGen {
```
sellout: Ideally this would be made available to us from scala-pathy, but I’m fine with it being here in the short-term. We need a pathy issue pointing at this, though.

core/src/test/scala/quasar/specs2/ExclusiveExecution.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import org.specs2.mutable.Specification
+import org.specs2.specification.{SpecificationStructure, Fragments}
+
+/** Trait that tags all examples in a spec for exclusive execution. Examples
+  * will be executed sequentially and parallel execution will be disabled.
+  *
+  * Use this when you have tests that muck with global state.
+  */
+trait ExclusiveExecution extends SpecificationStructure { self: Specification =>
```
sellout: 👍🏾

sellout:
Ok, nap time 😫

Very excited about this approach and looking forward to all the deletion coming up. Also very excited to have a (partial) second filesystem that we can test without an external dependency – that’ll help keep us honest in terms of not having Mongo semantics leak out of the filesystem.

Also, good luck when you eventually merge/rebase this on master.

core/src/test/scala/quasar/fs/PathyGen.scala
jedesah: https://github.com/slamdata/scala-pathy/pull/16

sellout: Great! So that just needs a reviewer and to get merged so @wemrysi can use it here!

core/src/main/scala/quasar/recursionschemes/cofree.scala
wemrysi: scalaz/cats/shapeless use lowercase, spire seems to use upper case. I've grown accustomed to lower case because of the former. I don't think there is an "official" stance other than the general convention that methods/functions start with lowercase (but even that we deviate from at times for ctor functions).

core/src/test/scala/quasar/fs/FileSystemFixture.scala
mossprescott: :+1: for giving real names to anything warranting an alias, and :-1: on greek for the variables. I think one-letter names are great for types that appear only in one line or a small function, say, and that's certainly the universal Scala convention.

We've been guilty of using `M` for "the monad of interest", even when it's not a variable (`Compiler.M`, `WorkflowBuilder.M`), and I think it's a really bad habit. Whatever that type represents, it's worth expressing in some kind of name.

I know some of these aliases might be there only to aid inference, and never get used directly, in which case I'd say let's have a convention for that. The best example would be the types building up the Coproducts elsewhere in this PR, which get called `FileSystem0`, etc. I think it would be great if they were prefixed with something indicating that they're an implementation detail. Maybe `InternalFilesystem0` or `__Filesystem1` (if that's legal) but hopefully someone has a better idea.

core/src/main/scala/quasar/EnvironmentError.scala
wemrysi: > However … I think I’m also content with just getting rid of the wrapper objects and only having the prisms around.

Ok, this is what I'm refactoring to, so if I've misunderstood you, let me know asap!

For each sum type, there will be
* Sealed trait with no `fold` method
* Cases in a `Case` object inside the trait's companion
* Capitalized constructor function values inside the companion that return the type of the trait (rather than the type of the constructor class)
* `Prism[Trait, (v1, v2, ..., vN)]` where `(v1, v2, ..., vN)` is a tuple of the arguments to a constructor, in the trait's companion, for each of the cases in `Case`.

core/src/main/scala/quasar/fs/Natural.scala
wemrysi: Aye it should.

core/src/main/scala/quasar/fs/Node.scala
wemrysi: Can we deal with these issues when we actually refactor views to use the algebras?

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: Indeed, will be corrected.

core/src/main/scala/quasar/fs/Node.scala
mossprescott: I think so, yes. There should be a little more clarity by then.

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: I agree it isn't a perfect fit, but maybe it means that `QueryFile` is a poor choice of name as, in terms of capability, one would like to be able to implement querying and listing the contents of the filesystem without having to also implement mutating said filesystem. Or maybe it means we just need two, single-term algebras. I didn't think it was too much of a stretch to group the two operations that represent questions about the filesystem into the same algebra (but it is a bit of a stretch, I'll admit). Any suggestions?

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: This method has actually already made its way into pathy, so we just need to release a new snapshot and update this code to use it.

core/src/main/scala/quasar/fs/WriteFile.scala
wemrysi: I could be convinced to remove the `XXXF` variants in favor of a `fromFoldable` syntax op for `Process`, but I think the non-chunked versions are too convenient, and obviously correct, to bother removing. The tests are implemented almost exclusively in terms of them so they've got implicit coverage.

core/src/main/scala/quasar/fs/inmemory.scala
wemrysi: > Presumably this doesn't do anything clever to simulate backend semantics around concurrent writes, correct? Put another way, is this a correct implementation of the semantics assumed by append/save/replace, etc.?

I'm not sure what would need simulating? As for correctness, it passes the same specifications that all filesystem implementations are expected to pass, so we should make sure they form a complete specification of the filesystem.

> What about edge cases like appending to a file and another request deletes it half-way?

I don't think this particular scenario is represented in the specs. There appear to be at least two versions of it, one where a delete happens during a particular `Write` and one where a delete happens in-between two `Writes` to the same handle.

Something else worth considering is what behavior is appropriate to prescribe and what are we content to allow to remain implementation-dependent?

core/src/main/scala/quasar/EnvironmentError.scala
sellout: 👍🏾

core/src/main/scala/quasar/fs/ManageFile.scala
sellout: 👍🏾 (can you add a “TODO: Remove Functor constraint after upgrading to Scalaz 7.2” comment?)

core/src/main/scala/quasar/fs/inmemory.scala
mossprescott: It sounds like we have more test coverage than I was aware of, so that's good. I think we might want to beef it up in the future, based on what we run into with other backends, but let's not worry about it for now.

core/src/main/scala/quasar/fs/package.scala
wemrysi: Probably because `pathy.Path.refineType` didn't exist at the time.

core/src/main/scala/quasar/physical/mongodb/MongoDb.scala
wemrysi: > What would we want this to be called in the separate project that eventually contains our stand-alone FP interface to MongoDB?

Not sure. Also unsure if it would even exist in said project. This really just exists as light insulation from the java driver, it isn't very full-featured, only what was necessary for our immediate needs.

As for the name, how unsure are you that it isn't right ;) Any suggestions for alternatives?

core/src/main/scala/quasar/physical/mongodb/mapreduce.scala
wemrysi: Yes, the wiki is clear on this. I think I just mis-remembered newlines between each lib, vs between each category. I'll add it to the list of things to cleanup.

core/src/main/scala/quasar/recursionschemes/cofree.scala
wemrysi: Yeah, that's fine. It will likely change again with further tidying to `recursionschemes` as it would be nice to be able to import the types without polluting implicit scope, but we don't have to worry about that now.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
wemrysi: It looks like `_` is a valid character to start an alias with, so for some final type `T[_]`, we could do `_T0[_]`, `_T1[_]`, etc for intermediate types that exist purely for the purposes of compiler handholding.

I also agree, in general, about trying to come up with useful names for aliases.

core/src/test/scala/quasar/specs2/ExclusiveExecution.scala
```diff
@@ -0,0 +1,28 @@
...
+  import SerialExecution._
+
+  sequential
+
+  override def map(fs: => Fragments) =
+    section(ExclusiveExecutionTag) ^ super.map(fs) ^ section(ExclusiveExecutionTag)
+}
+
+object SerialExecution {
```
wemrysi: Whoops, artifact of incomplete rename, will fix.

core/src/main/scala/quasar/physical/mongodb/MongoDb.scala
mossprescott: Fair enough; I realize most of my comments were really questions and if I could be more definitive, I would.

I could live with something like `MongoDbEffect` or `MongoDbIO` or `MongoDbTask`. I don't think any of those are great, but at least they would clearly identify this thing.

core/src/main/scala/quasar/fs/ManageFile.scala
wemrysi: Sure, but I'm only going to add it on one of these (as it is also remove `Coyoneda`, remove all the `XXXF` type aliases, possibly other things).

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: What is it that we want to get rid of? This method?

sellout: Any method in this part of the code that knows about SQL.

wemrysi: Ah, ok, sure, if we can get rid of this at some point, fine with me.

core/src/main/scala/quasar/physical/mongodb/MongoDb.scala
wemrysi: Ok, I can live with `MongoDbIO` as well, I'll use that.

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
wemrysi: @mossprescott As we discussed via hangout, I see your point, particularly regarding the deletion tests. The error check for saving non-objects is MongoDB-specific, though, and we hope to get rid of the map-reduce/aggregation specific tests as well.

However, the deletion ones also assume knowledge of MongoDB currently in order to ignore failures based on permission issues (i.e. when connected to MongoDB as a user that doesn't have permission to delete everything). If we want to keep this check then I propose we leave these tests here for now (with additional comments on where they should eventually move to and why they haven't yet) until we can come up with a more general way of ignoring these backend-specific failures within the backend-agnostic tests.

Another possibility is to omit these mongo-specific checks and just require that, if you specify the isolated environment flag when running the tests then you also must connect with a user that has the proper permissions, otherwise don't expect tests to work.

mossprescott: That sounds fine for now. When we get around to adding a second (real) backend, we'll take another look at everything MongoDB-specific and see what makes sense to generalize and how.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  val query  = QueryFile.Ops[FileSystem]
+  val read   = ReadFile.Ops[FileSystem]
+  val write  = WriteFile.Ops[FileSystem]
+  val manage = ManageFile.Ops[FileSystem]
+
+  val emptyMem = InMemState.empty
+
+  val hoistInMem: InMemoryFs ~> InMemIO =
+    Hoist[StateT[?[_], InMemState, ?]].hoist(pointNT[Task])
```
jedesah: Why are we putting `InMemoryFs` in `Task`? I don't think anything effectful is happening. The `State` monad seems like a perfectly good monad into which to translate the `Free` algebra in the case of the InMemory interpreter, at least as far as the tests for the derived functions are concerned. I think this would help reduce some of the noise in those tests.

The `web` project probably wants to provide a natural transformation into `Task` in order to co-mingle with `http4s`.

wemrysi: You're absolutely right, however we lift into `Task` preemptively for convenience with `Process` and it's `Catchable` requirement. We could provide a natural transformation for lifting only when `Process` is involved, but it almost always is and I think that would be noisier than having `Task` in the mix.

I wonder if a `Try`-like monad is a feasible alternative to `Task` for computations that are impure only due to the possibility of exceptions, but perform no other side-effects?

wemrysi: Actually, I guess that is pretty much what we currently do as the lifting is done in the `runXXX` methods.

jedesah: Ah, that makes sense. But it would be nice for a tests like `fileExists` or `descendantFiles` in `QueryFileSpec` that do not require `Process` to be able to deal directly with `State`.

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: @mossprescott @jdegoes I think the two of you are on to something with the additional dimensions. Ideally, we'd only need the "streaming" version of `QueryFile` and use it combined with `WriteFile` to provide the non-streaming version, but that doesn't seem possible for performance reasons, even if it is cleaner, conceptually.

What do you think, for the purposes of this PR, of leaving `ls` where it is for now, annotated with some comments (and a link to this PR), and revisiting the organization when we get to adding views? I know the web API PR that @jedesah is making some additions to the algebra, so we could postpone the tidying until that is done as well so we get it all at once?

jdegoes: :+1:

mossprescott: 👍🏾

core/src/main/scala/quasar/physical/mongodb/mapreduce.scala
wemrysi: The reason it exists is to separate the configuration/specification of the map-reduce operation from where it should be output, which allowed for nicer methods on `MongoDbIO`. It also avoids the ambiguity of `MapReduce(out = None)`, which I don't think is valid as some output destination (even if it is 'inline') is required. Similarly with `MapReduce.bson` which accepts a destination collection, but doesn't always use it, which isn't obvious.

In the existing `MongoDbEvaluatorImpl`, when handling a `MapReduceTask` we create a temp collection and then hand that to `executor.mapReduce` but we don't even know if it was used (well maybe _we_ do, but the types don't), even though we then return that temp collection as the output.

However, you're right that there is a lot of duplication. I'll see about trying to tighten up with existing `MapReduce` to get rid of the ambiguities and switch over to it in the new code as well.

wemrysi:
@sellout @mossprescott I think I've addressed all of your comments/concerns (aside from the map-reduce bit, which is in-progress), let me know if you agree when you have some time.

wemrysi:
@sellout Ok, map-reduce changes are done. I was able to get rid of the `MapReduce.Config` class by making some adjustments to `MapReduce` itself, gaining some precision in the process.

Ready for final review.

sellout:
Looks good – just some questions about prism usage and Travis/coverage stuff. Also, I agree with @jedesah’s comments on Task/State.

sellout:
Oh yeah, and maybe pull the scoverage highlighting line from #969, since you’re doing coverage stuff, then I can kill that PR that has long outlived its usefulness 😉

wemrysi:
Regarding the `Task`/`State` comments, I believe @jedesah has cleaned that up as part the API changes.

jedesah:
@wemrysi @sellout I did.

wemrysi:
@sellout I've pushed another commit that addresses all of your build comments, let me know if there is anything else.

mossprescott:
Just catching up with this thread: My recollection is we didn't use coverageAggregate before because it was broken on scala 2.11, so that may have since been fixed.

.travis.yml
```diff
@@ -22,6 +17,13 @@ notifications:
   slack:
     secure: KPCXtCAjtlpRQcH7TQtprrf67yIqPVpht9L6/+wbbNd14Tj3x53fiZ3OTabHVx3SL2ZWxbWq+FJGxdIFMyxC/fC0bV9flB7PlRtqHJV1Q7g49odAMAZL7syWr7vFP5bJt+TXuxHvMBihDHFk1Cehy4a1cnrOMReDgqGIFCwQtbI=
 env:
+  matrix:
```
mossprescott: As far as I can tell from [this build](https://travis-ci.org/quasar-analytics/quasar/builds/92807935), Travis runs the two builds sequentially and takes roughly twice as long. Or is it just reporting the time stupidly?

sellout: I think it’s the latter. It just sums the build times rather than telling you the actual elapsed time.

mossprescott: That figures. In that case since we're not paying for the time, why not?

sellout:
👍🏾but 😨

.travis.yml
wemrysi: Part of the prompting for this was that something caused the non-parallel version to timeout in travis (i.e. travis decided something was taking too long and killed it, it seemed).

PR #987, opened 2015-11-13
[SD-986] support listening on multiple ports

drostron:


web/src/main/scala/quasar/api/server.scala
```diff
@@ -124,30 +139,33 @@ class ServerOps[WC: CodecJson, SC](
...
-
-    def shutdown(srv: Option[(Int, Http4sServer)], log: Boolean): Task[Unit] =
-      srv.traverse_ { case (p, s) =>
-        s.shutdown *> (if (log) stdout("Stopped server listening on port " + p) else Task.now(()))
+        servers <- createServers(updCfg, idleTimeout, fsApi.AllServices.map(_ ++ fileSvcs ++ redirSvc))
+        _       <- stdout(s"Server started listening on port ${portsString(servers)}").liftM[EnvErrT]
+      } yield servers
+
+    def shutdown(srv: Option[Servers], log: Boolean): Task[Unit] =
+      srv.traverse_ { case m =>
```
wemrysi: If you like, you can leverage the fact that the composition of `Foldable` is also `Foldable` to tidy this a bit

```scala
Foldable[Option].compose[List].traverse_(srv) { case (p, s) => ...}
```

drostron: I like that and wonder if there's a convenient way to get a reference to the `Servers` value for the call to `portsString`.

wemrysi: Hm, good point. Though, do we only want to log if all succeed? Or do we want to log each one individually as it succeeds?

drostron: Individually seems like what we want. I'll update the logging on startup to be per server as well.

web/src/main/scala/quasar/api/server.scala
```diff
@@ -88,16 +92,27 @@ class ServerOps[WC: CodecJson, SC](
                        anyAvailablePort)
       .getOrElse(requested)
 
-  def createServer(config: WC, idleTimeout: Duration, svcs: EnvTask[ListMap[String, HttpService]])
-    : EnvTask[Http4sServer] = {
+  def builders(config: WC, idleTimeout: Duration): Task[Builders] = Task.now(List(
```
wemrysi: Why not just return an `(Int, BlazeBuilder)`? There is only one and, if it is pure, no need to point into `Task`.

web/src/main/scala/quasar/api/server.scala
```diff
@@ -88,16 +92,27 @@ class ServerOps[WC: CodecJson, SC](
...
+    wcPort.get(config) -> BlazeBuilder
+      .withIdleTimeout(idleTimeout)
+      .bindHttp(wcPort.get(config), "0.0.0.0")))
+
+  def createServers(config: WC, idleTimeout: Duration, svcs: EnvTask[ListMap[String, HttpService]])
```
wemrysi: Any reason to require `ListMap` in the signature instead of `Map`?

web/src/main/scala/quasar/api/server.scala
drostron: `List[(Int, BlazeBuilder)]` is in preparation for extension where more the one port and server will be required. True about `Task` — slated for pruning.

web/src/main/scala/quasar/api/server.scala
drostron: That requirement didn't change. There's a ordering to the services in the handling of requests.

wemrysi: Maybe we should just change the type of `svcs` to `EnvTask[List[(String, HttpService)]]` then? We don't rely on the fact that `svcs` is a `Map` anywhere in the body of this function.

web/src/main/scala/quasar/api/server.scala
drostron: Oh right, `Task` is in the mix due to a downstream project overriding `builders` and being effectful. I'm sure there's another way to handle this, but the ways that come to mind involve more changes than I want to incur at the moment. Maybe there's another way to handle this that isn't too refactor heavy?

web/src/main/scala/quasar/api/server.scala
```diff
@@ -88,16 +93,27 @@ class ServerOps[WC: CodecJson, SC](
...
+  def builders(config: WC, idleTimeout: Duration): Task[Builders] = Task.now(List(
+    wcPort.get(config) -> BlazeBuilder
+      .withIdleTimeout(idleTimeout)
+      .bindHttp(wcPort.get(config), "0.0.0.0")))
+
+  def createServers(config: WC, idleTimeout: Duration, svcs: EnvTask[Services])
+    : EnvTask[ServersErrors] = {
+
+    def servicesBuilder(services: Services, builders: Builders): Builders =
+      services.toList.reverse
```
wemrysi: I'm pretty sure this is equivalent to `services.foldRight`.

web/src/main/scala/quasar/api/server.scala
wemrysi: One possibility is to parameterize `createServers` on a `List` of builders, i.e. `createServers(..., builders: Builders)` rather than calling `builders` directly from within `createServers`. You could also be a bit more abstract, if you like, and just parameterize on any `F[_]: Traverse` of builder.

web/src/main/scala/quasar/api/server.scala
drostron: A copy relic — will update.

web/src/main/scala/quasar/api/server.scala
drostron: I'll ended up making `builders` abstract _(one of the options we discussed)_ since it meshes more easily with the downstream extension _(for the moment at least)_.

drostron:
@wemrysi: comments addressed

wemrysi:
Thanks, LGTM :+1:

PR #988, opened 2015-11-16
Enabling concurrent builds in travis

wemrysi:
A temporary PR to test out tweaks to travis config to get tests for multiple backends running in parallel.


wemrysi:
Hm, not sure what the deal is with coverage, it is uploaded twice, once per matrix entry. Maybe we need to figure out how to only upload once.

wemrysi:
This has served it's purpose, find the squashed changes in #986 .

PR #990, opened 2015-11-16
enable pending tests for sub-select scoping with views

mossprescott:
Two tests were marked pending and their assertions were broken; one was just wrong and the other had become out-dated since last week.

This commit provides the proof that getting rid of AnnotatedTree fixed SD-1101.

### DIFF
--- a/core/src/test/scala/quasar/compiler.scala
+++ b/core/src/test/scala/quasar/compiler.scala
@@ -954,29 +954,21 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
               "size" -> ObjectProject(Free('tmp0), Constant(Data.Str("size")))))))
     }
 
-    "compile sub-select with ambiguous names" in {
+    "compile sub-select with same un-qualified names" in {
       testLogicalPlanCompile(
         "select city, pop from (select city, pop from zips) temp",
         Let('tmp0,
           Let('tmp1,
             read("zips"),
-            Let('tmp2,
+            Squash(
               makeObj(
                 "city" -> ObjectProject(Free('tmp1), Constant(Data.Str("city"))),
-                "pop" -> ObjectProject(Free('tmp1), Constant(Data.Str("pop")))
-              ),
-              Let('tmp3,
-                Squash(Free('tmp2)),
-                Free('tmp3)))),
-          Let('tmp4,
+                "pop" -> ObjectProject(Free('tmp1), Constant(Data.Str("pop")))))),
+          Squash(
             makeObj(
               "city" -> ObjectProject(Free('tmp0), Constant(Data.Str("city"))),
-              "pop" -> ObjectProject(Free('tmp0), Constant(Data.Str("pop")))
-            ),
-            Let('tmp5,
-              Squash(Free('tmp4)),
-              Free('tmp5)))))
-    }.pendingUntilFixed("SD-1101")
+              "pop" -> ObjectProject(Free('tmp0), Constant(Data.Str("pop")))))))
+    }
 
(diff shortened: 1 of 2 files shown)


core/src/test/scala/quasar/compiler.scala
```diff
@@ -954,29 +954,21 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
...
 
-    "compile sub-select with ambiguous names" in {
+    "compile sub-select with same un-qualified names" in {
       testLogicalPlanCompile(
         "select city, pop from (select city, pop from zips) temp",
         Let('tmp0,
           Let('tmp1,
             read("zips"),
-            Let('tmp2,
+            Squash(
```
jdegoes: All these spurious `Squash`es are semantically incorrect. Although I realized that happened in an earlier PR in which I was not paying close enough attention, so doesn't need fixing here.

sellout: Oh, this is good to know, because the SQL² stuff still maintains a Squash per Select.

mossprescott: `Squash` still a no-op in the planner, so I'm tempted to suggest we just don't generate it for now if we aren't getting it right.

jdegoes: `Squash` is needed so the `ORDER BY` clause will operate on the full set, rather than be limited to the last dimension of data (e.g. generated by a `GROUP BY`, array or map shift, etc.). If you use it arbitrarily on all expressions, the semantics are effectively cross (because you're reducing all the expressions to single-dimensional sets with unrelated provenance).

jdegoes: > Squash still a no-op in the planner, so I'm tempted to suggest we just don't generate it for now if we aren't getting it right.

Fine by me. Better than generating it incorrectly.

sellout: So, `Squash` happens before `OrderBy`, not after `Select`. Does that mean all the cases where we don’t use an `OrderBy` in the query we should never Squash? I guess we need to Squash before distinct as well, right? And two squashes separated by any number of non-identity altering ops is identical to just having the first Squash?

jdegoes: Yeah, `Squash` happens before `OrderBy`. An alternative is to just define `OrderBy` to "internally" do a squash, i.e. it's really a composite of some `Sort` function and `Squash`. 

An alternative is just to ignore this issue entirely and not insert the squashes at all, which may result in more predictable behavior. One can argue the semantics of ANSI SQL are preserved, because the `SELECT` clause must be fully "pre-squashed"; i.e. there is no left dimensional shift and any groups must be aggregated.

The tricky thing there would be a query like:

```sql
SELECT user, age, COUNT(*) FROM people GROUP BY age ORDER BY age
```

Without the auto-squashing, this (non-ANSI SQL) query would end up ordering inside each group, because the set being ordered is 2 dimensional.

`DISTINCT` has similar issues that require resolution. I'd just ignore `Squash` for now and we can figure out what we want to do as part of the MRA / qScript layer.

sellout: 👍🏾

mossprescott:
@jdedoes: do we need a ticket to straighten out SQUASH?

@sellout: so is this good to merge?

mossprescott:
I meant @jdegoes, obviously.

jdegoes:
No, it will be subsumed by the MRA / qScript epic.

sellout:
Ready.

mossprescott:
Thanks!

PR #991, opened 2015-11-17
Enable view creation and interpretation

mossprescott:
- construct the ViewBackend wrapping the regular filesystem and interpreting all requests involving views
- allow views to be created through the API
- update the README with details on view configuration and affected APIs

This completes SD-976 and SD-977.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ you probably want to install MongoDB locally and point Quasar to that one. Insta
 allow you to run the integration tests offline as well as make the tests run as fast as possible.
 
 In order to install MongoDB locally you can either use something like Homebrew or simply go to the MongDB website and follow the
-instructions that can be found there. 
+instructions that can be found there.
 
 Once we have a MongoDB instance handy, we need to set a few
 environment variables in order to inform Quasar about where to find the backends required in order to run the integration tests.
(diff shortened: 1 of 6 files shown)


mossprescott:
@wemrysi: I assigned you just because you reviewed the previous PR in which we decided to disable these features through the API. I'd be happy to pick on someone else if you have enough on your plate right now.

mossprescott:
I pulled this back to "in progress" while I sort out the behavior of `ls`.

README.md
```diff
@@ -139,6 +139,25 @@ Then the filesystem will contain the paths `/local/test/` and `/local/students/c
...
+If the mount's key is "view" then the mount represents a "virtual" file, defined by a SQL query. When the file's contents are read or referred to, the query is executed to generate the current result on-demand. A view can be used to create dynamic data that combines analysis and formatting of existing files without creating temporary results that need to be manually regenerated when sources are updated.
+
+For example, given the above MongoDB mount, an additional view could be defined in this way:
+
+```json
+  "mountings": {
+    ...,
+    "/simpleZips": {
+      "view": {
+        "connectionUri": "sql2:///?q=select+_id+as+zip,+city,+state+from+\"/local/test/zips\""
```
sellout: I thought we only used `connectionUri` with Mongo because that is how you connect to the DB. For views, why couldn’t we just do `"view": { "sql2": "select id as zip, city, state from \"/local/test/zips\"" }`?

wemrysi: :+1:

mossprescott: @jdegoes can speak to the bigger picture, but the idea is to make every configuration string a URI, so you can enter then in the same way using any tool that supports URIs. For example, a generic config UI (even though we're moving towards custom UI for each type in our own front-end).

web/src/main/scala/quasar/api/fs.scala
```diff
@@ -384,14 +384,18 @@ final case class FileSystemApi[WC, SC](
...
+        x =>
+          if (!x) NotFound()
+          else
+            path.file.fold[Task[Response]]({
+              bknd.ls(path).fold[Task[Response]](
+                handlePathError,
+                // NB: we sort only for deterministic results, since JSON lacks `Set`
+                paths => Ok(Json("children" := paths.toList.sorted))).join
+            })(
+            _ => Ok(Json.obj()))).join
```
sellout: Indentation is a bit off on these two lines. Also, why not `κ`?

sellout:
Good, minor comments (and really the first one is just a question, and can be addressed in a different PR, since the change here is just describing behavior that already exists).

README.md
jdegoes: This will soon be stuffed in a relational database (2.5) so the schema for all mount configs has to be the same, anyway. 

Given the number of backends that are already expressed in URI (e.g. pretty much any JDBC database, as well as HDFS and lots of others), makes sense to standardize on that, although of course we don't have to use it at the API boundary. But my vote would just be keep it as it is, makes it easy to do things with mounts programmatically if the structure of every mount is the same.

web/src/main/scala/quasar/api/fs.scala
mossprescott: Ah, let me clean that up. I think I can get rid of some curlies and a type annotation or two as well.

sellout:
👍🏾

PR #992, opened 2015-11-17
Add array and map operators to the compiler.

sellout:
SD-995 #done, SD-996 #done

Unshift and shift of array and map keys and values.

Also rename “object” to “map” in affected places (more renaming to be
done as part of SD-1018).


drostron:
👍🏽

Couple of items related to my lack of familiarity:
* I would like gain a better understanding on what's missing from mongodb that's necessary to support FlattenArrayIndices, ShiftArray, FlattenMapKeys.
* `foo{"bar"}` doesn't seem to be supported. What am I missing?
* Am I missing map concat?

sellout:
I don’t think anything is missing from Mongo to support these – it’s just not part of this issue. The SQL² epic is intended to implement the full parser/compiler (IE, the frontend) without worrying about implementation for any particular backend.

`foo{bar}` should already be supported as `FieldDeref` / `structural.ObjectProject`, not new in this PR.

We don’t have any kind of concat reduction yet, so I don’t think mapconcat is possible. I could be overlooking something, though.

jdegoes:
Array concat is `||`, inherited from Postgres, map concat could be anything we want. I'm partial to `++`, although you can call the function manually as for all operators (`MAP_CONCAT`?). 

The corresponding reductions have syntax, e.g. `[foo ...]` reduces the last dimensional set into an array, while `{ keys : values ...}` does the same for maps.

sellout:
Ohhhhh, concatting two maps, not `mapconcat`. Got it!

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -220,10 +220,98 @@ trait StructuralLib extends Library {
     },
     untyper(tpe => success(List(FlexArr(0, None, tpe)))))
 
+  val FlattenMapKeys = Expansion(
+    "{*:}",
+    "Flattens a map into a set",
```
wemrysi: This just flattens the keys, no? Not the entire map?

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -220,10 +220,98 @@ trait StructuralLib extends Library {
...
+    Set(Top), AnyObject :: Nil,
+    noSimplification,
+    partialTyper { case List(x) if x.objectLike => Str },
+    untyper(tpe => success(List(Obj(Map(), Some(Top))))))
+
+  val FlattenArrayIndices = Expansion(
+    "[*:]",
+    "Flattens an array into a set",
```
wemrysi: Same here no, this doesn't flatten the contents, just the indices, right?

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -220,10 +220,98 @@ trait StructuralLib extends Library {
...
+    Set(Int), AnyArray :: Nil,
+    noSimplification,
+    partialTyper { case List(x) if x.arrayLike => Int },
+    partialUntyper {
+      case Set(Int) | Int => List(FlexArr(0, None, Top))
+    })
+
+  val ShiftMap = Expansion(
+    "SHIFT_MAP",
+    "Shifts a map into a set",
```
wemrysi: This (and the rest of the help text on these new functions) leaves a bit to be desired, especially if this is ever displayed to a user. I'd prefer something a bit more descriptive, this doesn't really convey any more information than the name of the function itself (the fact that it returns a set, but that's about it).

sellout: Got any suggestions as to what to say? If we had some MRA docs, I think referencing the definitions of “shift”, “unshift”, and “flatten” there would be a good idea.

wemrysi: Heh, well my comment was prompted by the fact that I don't remember what "shift" and "unshift" mean ;) At the risk of being verbose, would something like "Shifts a map into a set, <explain what that actually means>" work?

wemrysi: And I agree regarding referencing some docs, should they exist.

jdegoes: 'Zooms into the array elements, looking at the indices,' etc.

PR #995, opened 2015-11-19
Split /metadata's "type" field into "type" and "mount".

mossprescott:
SD-978 #done

This allows a client to identify views as both a kind of mount and a kind of file.

Also fixed the listing of directories which are mounts but also contain overlayed views, and where a view appears at the same path as an actual file.

### DIFF
--- a/web/src/main/scala/quasar/api/fs.scala
+++ b/web/src/main/scala/quasar/api/fs.scala
@@ -184,11 +184,9 @@ final case class FileSystemApi[WC, SC](
 
   implicit def FilesystemNodeEncodeJson = EncodeJson[FilesystemNode] { fsn =>
     Json(
-      "name" := fsn.path.simplePathname,
-      "type" := (fsn.typ match {
-        case Mount => "mount"
-        case Plain => fsn.path.file.fold("directory")(κ("file"))
-      }))
+      (("name" := fsn.path.simplePathname) ::
+        ("type" := fsn.path.file.fold("directory")(κ("file"))) ::
+        fsn.mountType.toList.map("mount" := _)): _*)
   }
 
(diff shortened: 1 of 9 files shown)


web/src/main/scala/quasar/api/fs.scala
```diff
@@ -184,11 +184,10 @@ final case class FileSystemApi[WC, SC](
...
     Json(
-      "name" := fsn.path.simplePathname,
-      "type" := (fsn.typ match {
-        case Mount => "mount"
-        case Plain => fsn.path.file.fold("directory")(κ("file"))
-      }))
+      (("name" := fsn.path.simplePathname) ::
+        ("type" := fsn.path.file.fold("directory")(κ("file"))) ::
+        Nil) ++
+        fsn.mountType.toList.map("mount" := _): _*)
```
sellout: Can’t this be written without the Nil?
```scala
Json(
  ("name" := fsn.path.simplePathname) ::
    ("type" := fsn.path.file.fold("directory")(κ("file"))) ++
    fsn.mountType.toList.map("mount" := _): _*)
```

sellout:
That’s it.

web/src/main/scala/quasar/api/fs.scala
mossprescott: <facepalm>

Kinda disappointed that's all you could come up with, though.

sellout:
👍🏾

I did also like how you included the mount type in the result. I hadn’t considered that before.

web/src/main/scala/quasar/api/fs.scala
mossprescott: Hey, github ate my attempt at cuteness, which went something like this: `<facepalm>`.

It's hard to be facetious in this medium.

PR #996, opened 2015-11-20
Improvements to handling set literals.

sellout:
Progress toward SD-1045.

* Parse `()` as an empty set,
* constant-fold `x IN ()` to `false`, and
* don’t accept arrays for `IN`.

This doesn’t solve the final problem of single-element sets, which would be easy with SD-1078 fixed.

### DIFF
--- a/core/src/main/scala/quasar/std/set.scala
+++ b/core/src/main/scala/quasar/std/set.scala
@@ -197,6 +197,21 @@ trait SetLib extends Library {
     setTyper(partialTyper { case List(s1, _) => s1 }),
     setUntyper(t => success(t :: Type.Top :: Nil)))
 
+  val In = Mapping(
+    "(in)",
+    "Determines whether a value is in a given set.",
+    Type.Bool, Type.Top :: Type.AnySet :: Nil,
+    noSimplification,
+    partialTyper {
+      case List(_,             Type.Const(Data.Set(Nil))) =>
+        Type.Const(Data.Bool(false))
+      case List(Type.Const(x), Type.Const(Data.Set(set))) =>
+        Type.Const(Data.Bool(set.contains(x)))
+      case List(_,             Type.Const(Data.Set(_)))   => Type.Bool
+      case List(_,             _)                         => Type.Bool
+    },
+    basicUntyper)
+
   val Constantly = Mapping("CONSTANTLY", "Always return the same value",
     Type.Bottom, Type.Top :: Type.Top :: Nil,
     noSimplification,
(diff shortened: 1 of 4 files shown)


core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -656,6 +656,11 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
             Selector.In(Bson.Arr(List(Bson.Text("AZ"), Bson.Text("CO"))))))))
     }
 
+    "plan filter with field in empty array" in {
+      plan("select * from zips where state in ()") must
+        beWorkflow($pure(Bson.Arr(Nil)))
+    }
+
     "plan filter with field containing constant value" in {
```
sellout: This test and the next one should really change. Currently, MongoDB sees Sets as Arrays, so this ends up still working, but really, we should have to do something like `where 43 in loc[:_]` (where `[:_]` is an array shift – currently being added in another PR).

jdegoes: This is probably correct. It depends on one little detail: what are the semantics of set construction when the thing inside the set is already a set? Sets cannot be nested in MRA, because they're not part of the data model, but if we add an extra dimension for nested sets, we may be able to define a quite useful semantic for `IN` here.

jdegoes: But yeah, we do need to change that, if a user wants to do in an array, it will probably take the above proposed semantic together with the array shift operator `[_]` / `[:_]`.

sellout:
@mossprescott Ping.

core/src/main/scala/quasar/std/set.scala
```diff
@@ -197,6 +197,21 @@ trait SetLib extends Library {
...
+    "Determines whether a value is in a given set.",
+    Type.Bool, Type.Top :: Type.AnySet :: Nil,
+    noSimplification,
+    partialTyper {
+      case List(_,             Type.Const(Data.Set(Nil))) =>
+        Type.Const(Data.Bool(false))
+      case List(Type.Const(x), Type.Const(Data.Set(set))) =>
+        Type.Const(Data.Bool(set.contains(x)))
+      case List(_,             Type.Const(Data.Set(_)))   => Type.Bool
+      case List(_,             _)                         => Type.Bool
```
mossprescott: Redundant.

mossprescott:
Just the one tiny thing. Seems simple enough. Sorry about the delay.

sellout:
Feels a bit silly, but … back in your court.

mossprescott:
One more thing... no, actually this looks fine.

PR #997, opened 2015-11-20
fix /metadata 404 response body

mossprescott:
Was empty, now has valid JSON with `{ "error": "There is no file/directory at path ..."}`

Note: this fixes only the request which was changed to a 404 in a recent commit. A quick search shows several other occurrences of `NotFound()` which we might want to address, probably in our existing "improve error messages" ticket ([SD-824](https://slamdata.atlassian.net/browse/SD-824)).


jonsterling:
Do these changes set the content-type on the response?

jedesah:
@jonsterling Yes, http4s should take care of that.

jedesah:
LGTM. Can merge when build completes.

PR #998, opened 2015-11-21
allow views to be nested inside other mounts

mossprescott:
SD-1151 #done

This adds two simple tests, and the simplest, quickest possible fix. It's beginning to look like making ViewConfig a subtype of MountConfig along with MongoDbConfig was a mistake, with more and more code ending up either pattern matching or inspecting paths to decide how to handle mounts. But in the interest of getting a fix in quickly, this commit makes it a bit worse. I'm going to write a "technical debt" ticket with a proposal for cleaning this up.


mossprescott:
Seems to work in some simple manual testing of the API.

sellout:
This seems reasonable for now – are you thinking we’ll distinguish between views and mounts in future, or just between the configs?

👍🏾

mossprescott:
Thanks, just waiting for Travis now.

See https://slamdata.atlassian.net/browse/SD-1152.

PR #1000, opened 2015-11-24
Streaming query evaluation

mossprescott:
This probably all gets squashed down to three commits before all's said and done (the first two here, followed by everything else squashed), but I think the way they're broken out here may aid understanding.

The upshot is about half of our regression test queries can now run on a read-only connection; the remainder are marked as "pending" for the new MONGO_READ_ONLY test configuration.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/evaluator.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/evaluator.scala
@@ -17,18 +17,19 @@
 package quasar.physical.mongodb
 
 import quasar.Predef._
+import quasar.fp._
 import quasar.recursionschemes._, Recursive.ops._
 import quasar._, Errors._, Evaluator._
-import quasar.fs.Path.PathError.NonexistentPathError
+import quasar.fs.Path, Path.PathError.NonexistentPathError
 import quasar.javascript._
 import Workflow._
 
 import com.mongodb._
-import com.mongodb.client.model._
 import scalaz.{Free => FreeM, _}, Scalaz._
 import scalaz.concurrent._
+import scalaz.stream._
 
-trait Executor[F[_]] {
+trait Executor[F[_], C] {
   def generateTempName(dbName: String): F[Collection]
 
   def version: F[List[Int]]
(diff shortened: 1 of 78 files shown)


mossprescott:
Coverage dropped from 80.1 to 79.9, causing the build to fail. I'm trying to figure out if there's something I can easily pick up.

mossprescott:
+47, -12, and decreased by 0.01%?

core/src/test/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -57,7 +54,7 @@ class EvaluatorSpec extends Specification with DisjunctionMatchers {
       val wf = $pure(Bson.Text("foo"))
 
       MongoDbEvaluator.toJS(crystallize(wf)) must beLeftDisjunction
-    }
+    }.pendingUntilFixed("not a failure when streamed, unless we want to enforce that limitation")
```
sellout: We should definitely not enforce that limitation. The fact that `$pure` operations that are the source of another op will still fail in MongoDB is a bug – we should work around those cases either by writing the value into a document and rewriting the next task to take that into account, or (I think more likely) not having `$pure` exist at all, and rather inline that data into whatever references it.

core/src/main/scala/quasar/physical/mongodb/workflowtask/ast.scala
```diff
@@ -60,21 +60,22 @@ object WorkflowTaskF {
...
-          case ReadTaskF(coll) => G.point(ReadTaskF[B](coll))
-          case QueryTaskF(src, query, skip, limit) =>
-            f(src).map(QueryTaskF(_, query, skip, limit))
-          case PipelineTaskF(src, pipe) => f(src).map(PipelineTaskF(_, pipe))
-          case MapReduceTaskF(src, mr) => f(src).map(MapReduceTaskF(_, mr))
-          case FoldLeftTaskF(h, t) =>
-            (f(h) |@| t.traverse(f))(FoldLeftTaskF(_, _))
-        }
-    }
+  // NB: currently unused, and hurts coverage
```
sellout: 😢

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -387,14 +505,20 @@ class JSExecutor[F[_]: Monad](nameGen: NameGenerator[F])
 
   def mapReduce(source: Collection, dst: Collection, mr: MapReduce) =
     write(Call(Select(toJsRef(source), "mapReduce"),
-      List(mr.map, mr.reduce, mr.bson(dst).toJs)))
+      List(mr.map, mr.reduce, mr.bson(Some(dst)).toJs)))
```
sellout: Not that it matters in this case, but @wemrysi has pointed that that using `dst.some` is typed properly, whereas `Some(dst)` has a subtype.

it/src/test/scala/quasar/fs/filesystem.scala
```diff
@@ -31,11 +31,17 @@ class FileSystemSpecs extends BackendTest with DisjunctionMatchers with SkippedO
     case x => x must beNull
   }
 
-  backendShould(interactive.zips.run) { (prefix, fs, _, files) =>
+  backendShould(interactive.zips.run) { (prefix, insertFs, fs, name, files) =>
     val relPrefix = prefix.asRelative
     val TestDir = relPrefix ++ testRootDir ++ genTempDir.run
     val ZipsDir = files.head
 
+    implicit class SkippedOnReadOnly[T: org.specs2.execute.AsResult](t: => T) {
```
sellout: I don’t think it should hold up this PR, but I want to discuss this class. Not opposed to it, it just raises some questions.

sellout:
So, just fix those pending tests to pass with the good results, and that’s pretty much it. Sorry it’s so late tonight.

mossprescott:
Note that I also corrected the semantics of PureTask when streaming to match the non-streaming impl, which is that an array value is broken down into its elements. I'm not sure that's behavior we want in the long term, but I definitely think they need to match.

it/src/test/scala/quasar/fs/filesystem.scala
mossprescott: Certainly. The whole approach of attempting to run everything on read-only and then flagging stuff that's known to break seems bogus, but was the best way to get it done with the least merging nightmare, I think. In some ways, this might be simpler when we've broken down the backend trait into Read and Write algebras, and it will be natural to test them separately.

sellout: Read/Write algebras was my first thought, too. But _all_ queries are part of the Read algebra. The eventual solution for read-only, I think, is to do as much as possible in a single task, then handle the rest in the LP evaluator, so we should never have to fail on read-only. This is the same kind of approach we’ll need for handling other limited data sources, like the Twitter API or whatever.

sellout: Oh, sorry – so, this class does mostly (exclusively?) mark stuff that would be in the write algebra – it’s the integration tests that are skipped on `mongodb_read_only` that my previous comment really applies to.

mossprescott: Absolutely, which is why all the queries are "pending", meaning (in my mind) that they all _should_ work and the implementation just isn't there yet. This `skippedOnReadOnly` thing is for tests that explicitly verify write behavior, so would never be expected to work.

jdegoes: > This is the same kind of approach we’ll need for handling other limited data sources, like the Twitter API or whatever.

Exactly. Eventually all queries must succeed regardless of whether the underlying source of data supports writeback or is "query complete". Just a question of how long it will take to get the answer in cases where a query complete data source only has read capabilities and the query involved requires writeback to fully evaluate.

mossprescott:
@sellout: Waiting for your OK and Travis's. I'll have my squashed commits ready to go.

sellout:
Yeah, the problem with breaking down arrays is that we’re dealing with Bson there, and Bson has no distinction between an array and a set, so we don’t know whether it _should_ be an array or a set. Another reason to not push pure values through the planner.

sellout:
👍🏾

mossprescott:
Squashed, coverage looks good locally, and building now. By the way, I was on the fence about how much squashing to do here: Two commits are really separate fixes (commenting a Traverse instance and changing some logging), and then the other three are all distinct changes, but really go together in the sense that some tests would fail if they weren't all merged together.

Ultimately I started to feel like I was pressing my luck pushing the commits around, so I just called it.

jdegoes:
> Yeah, the problem with breaking down arrays is that we’re dealing with Bson there, and Bson has no distinction between an array and a set, so we don’t know whether it should be an array or a set. Another reason to not push pure values through the planner.

Well, MRA makes this easy: Sets can only be introduced by operations on values, since they don't exist in the data model, only the MRA calculus. The one place we have to cheat is in collections, since a collection represents a value (a map) that's already been left-shifted, so we have to treat it as a set.

But certainly, no value inside a document inside a collection can be interpreted as a set.

sellout:
I think that’s orthogonal – the issue here is that these two LPs:
```scala
Constant(Data.Arr(List("a", "b", "c")))
```
and
```scala
FlattenArray(Constant(Data.Arr(List("a", "b", "c"))))
→ Constant(Data.Set(List("a", "b", "c")))
```
currently end up as `PureTask(Bson.Arr(List("a", "b", "c")))`, which is also weird, because all we do once we get to that point is is convert the Bson back into Data to return it to the user.

PR #1001, opened 2015-11-24
Add test for evaluate on QueryFile

jedesah:
- Create an Interpreter abstraction to simplify interpreting various Monad transformers and helps avoid using Task when we don't need to


jedesah:
@wemrysi Probably spent a bit more time on this then I could have but I am pretty proud of myself for coming up with this. This way we can either interpret into a `Task[State]` or just `State` depending on the situation without creating endless natural transformations. It also reduces the amount of code needed interpretation with `ReadWrite`. Hopefully you don't spot some fundamental flaw with it :smile:

core/src/main/scala/quasar/fp/interpret.scala
```diff
@@ -36,4 +36,49 @@ object interpret {
...
+
+    def interpretT2[T1[_[_],_]: Hoist, T2[_[_],_]: Hoist](implicit functor:Functor[ALGEBRA], monad: Monad[M]): T1[T2[F,?],?] ~> T1[T2[M,?],?] =
+      Hoist[T1].hoist[T2[F,?],T2[M,?]](interpretT[T2])(Hoist[T2].apply[F])
+
+    def runLog[A](p: Process[F,A])(implicit functor: Functor[ALGEBRA], monad:Monad[M], catchable: Catchable[M]): M[IndexedSeq[A]] =
+      p.translate(interpret).runLog
+
+    // Would be possible if `MonadTrans` had a notion of `Catchable`. Instead we need to copy paste and
+    // replace the value of T with a particular Monad Transformer
+    // def runLogT[T[_[_],_]:Hoist,A](p: Process[T[F,?],A]): T[TaskMemState,IndexedSeq[A]] = {
```
wemrysi: Would something like `def runLogT[T[_[_], _]: Hoist, A](p: Process[T[F, ?], A])(implicit T: Catchable[T[M, ?]]): T[M, IndexedSeq[A]]` work?

core/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -23,7 +23,7 @@ object InMemory {
   type RM = Map[ReadHandle, Reading]
   type WM = Map[WriteHandle, AFile]
 
-  type InMemoryFs[A]  = State[InMemState, A]
+  type MemState[A]  = State[InMemState, A]
```
wemrysi: I don't think I'm a fan of this, I'd prefer the alias to be what the concrete type means, not how it is implemented. i.e. if we decided to add another layer I wouldn't want to have to rename the monad again.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -77,60 +71,52 @@ trait FileSystemFixture {
...
+  type F[A]            = Free[FileSystem, A]
+  type InMemFix[A]     = ReadWriteT[MemState, A]
+  type MemStateTask[A] = StateT[Task, InMemState,A]
+  type MemStateFix[A]  = ReadWriteT[MemStateTask,A]
+
+  object Mem extends Interpreter[FileSystem,MemState] {
+    val interpretTerm: FileSystem ~> MemState =
+      interpretFileSystem(queryFile, readFile, writeFile, manageFile)
+
+    val interpret: F ~> MemState = super.interpret
```
wemrysi: Why is this needed?

core/src/main/scala/quasar/fs/InMemory.scala
jedesah: I guess I am not convinced that the return type for an interpreter should be called Filesystem. The Filesystem is the interpreter not the the type produced by interpretation. But that's an unrelated concern, I can revert this and revisit at a later time.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
jedesah: It makes it easier to use (without using apply), because it does not have any implicit parameters.

core/src/main/scala/quasar/fs/InMemory.scala
wemrysi: That's fine, if you have a better name for what's there, I'm open to it, I just wasn't a fan of this particular choice ;)

jedesah: Ya, was not 100% sold on `MemState` anyway as it could be confusing whether it's the actual state of the InMemory interpreter or the `State` Monad fixed to `InMemState`.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
wemrysi: Ah, maybe it makes sense to factor the `Functor[F]` and `Monad[M]` constraints to the "module" level given how this type will be used and that they appear in every method. Could even change to `abstract class Interpret[F[_]: Functor, M[_]: Monad)(val interpretTerm: F ~> M) { ... }`.

jedesah: Oh ya, that's what I wanted to do last night, but I was at a lost on how to encode that in Scala.

jedesah: You prefer `F[_]` over `ALGEBRA[_]`? :smile: Currently we are using `F[A] = Free[ALGEBRA,A]` though?

jdegoes: By convention, `F[_]` should probably be the algebra. The other thing could be called `Dsl[A]` or `Program[A]` or whatever.

jedesah:
@wemrysi Even more generalized. Thx for the suggestions.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -77,60 +71,46 @@ trait FileSystemFixture {
...
+  val hoistTask: InMemoryFs ~> MemStateTask =
+    Hoist[StateT[?[_], InMemState, ?]].hoist(pointNT[Task])
 
-  def runLog[A](p: Process[FileSystemErrT[F, ?], A]): InMemResult[IndexedSeq[A]] =
-    p.translate[InMemResult](runResult).runLog
+  object MemTask extends Interpreter[FileSystem, MemStateTask](
+    interpretTerm = hoistTask compose Mem.interpretTerm
+  )
 
-  def evalLogZero[A](p: Process[FileSystemErrT[F, ?], A]): Task[FileSystemError \/ IndexedSeq[A]] =
```
wemrysi: Any particular reason not to preserve this? It tidies up a fairly common use-case...

core/src/main/scala/quasar/fp/interpret.scala
```diff
@@ -36,4 +36,51 @@ object interpret {
       def apply[A](fa: Coproduct5[F, G, H, I, J, A]) =
         fa.run.fold(f, interpret4(g, h, i, j)(_))
     }
+
+  class Interpreter[F[_]: Functor, M[_]: Monad](val interpretTerm: F ~> M) {
```
wemrysi: I'd also say go ahead and move this into its own file. We might even want to create a new package under `fp` called `free` to put this and other bits related to working with `Free`, though we can do that part later.

core/src/main/scala/quasar/fp/interpret.scala
```diff
@@ -36,4 +36,51 @@ object interpret {
...
+
+    def runLogT2[T1[_[_],_]: Hoist, T2[_[_],_]: Hoist,A](p: Process[T1[T2[Program,?],?],A])(implicit catchable: Catchable[T1[T2[M,?],?]]): T1[T2[M,?],IndexedSeq[A]] = {
+      type ResultT[A] = T1[T2[M,?],A]
+      val monadR: Monad[ResultT] = Hoist[T1].apply[T2[M,?]](Hoist[T2].apply)
+      p.translate[ResultT](interpretT2[T1,T2]).runLog[ResultT,A](monadR,catchable)
+    }
+
+    // Specialized for type inference
+
+    def runLog[E,A](p: Process[EitherT[Program,E,?],A])(implicit catchable: Catchable[M]): EitherT[M,E,IndexedSeq[A]] = {
```
wemrysi: Can we move these into the particular implementations in `FileSystemFixture`? You could even create a new subclass there that contains these if you need them both in `Mem` and `MemTask`.

wemrysi:
@jedesah This looks pretty good, I had a few more mostly organizational comments, then we're probably good to merge.

core/src/test/scala/quasar/fs/FileSystemFixture.scala
jedesah: Not really, will restore in the `MemTask` object

core/src/main/scala/quasar/fp/interpret.scala
jedesah: Sure. good idea.

core/src/main/scala/quasar/fp/interpret.scala
jedesah: It's unfortunate that we need those and of course they are specialized to our needs and we would need to remove them if we were to contribute this to some external project. But for now, I think it's likely to be generally applicable to anywhere in `quasar` where we might need to interpret `Free`. But I don't feel too strongly about this, if you really believe they are only applicable in `FileSystemFixture`, I am happy to create a subclass in there that adds these from which `Mem`, `MemTask` and `MemFixTask` can inherit from.

wemrysi: How about sibling class to `Interpret` that includes these? Then things that need them can use the specialized class, keeping the general one focused?

jedesah: sure, good idea.

jedesah:
@wemrysi Updated.

wemrysi:
:+1: Looks good

wemrysi:
Not that this isn't building, but it looks like the issue might be due to my PR that you merged. Not sure if you want to fix it here or I can just submit another PR to fix that bit.

jedesah:
@wemrysi Ya, sure, that would be great, the fix is not immediately obvious to me.

PR #1003, opened 2015-11-24
[SD-1146] Implement Explain for MongoDB interpreter

wemrysi:
Makes some minor modifications to the `Explain` constructor and adds the implementation for MongoDB.

An output file is added to the `Explain` constructor as, in this PR, there is no support for streaming queries and it seemed to make sense to handle this in one spot rather than have each backend have to do it. Note that the client API does not require an output file as either the first one from the LP is used or a placeholder is used.

### DIFF
--- a/core/src/main/scala/quasar/fs/QueryFile.scala
+++ b/core/src/main/scala/quasar/fs/QueryFile.scala
@@ -16,7 +16,8 @@ object QueryFile {
   final case class ExecutePlan(lp: Fix[LogicalPlan], out: AFile)
     extends QueryFile[(PhaseResults, FileSystemError \/ ResultFile)]
 
-  final case class Explain(lp: Fix[LogicalPlan]) extends QueryFile[PhaseResults]
+  final case class Explain(lp: Fix[LogicalPlan], out: AFile)
+    extends QueryFile[(PhaseResults, Option[FileSystemError])]
 
   /** TODO: While this is a bit better in one dimension here in `QueryFile`,
     *       @mossprescott points out it is still a bit of a stretch to include
@@ -29,6 +30,11 @@ object QueryFile {
     */
   final case class ListContents(dir: ADir)
     extends QueryFile[FileSystemError \/ Set[Node]]
+
+  /** File used in explain output when a `LogicalPlan` does not contain any
+    * file references.
+    */
+  val DefaultExplainOut: AFile = rootDir </> dir("explain") </> file("out")
 
   final class Ops[S[_]](implicit S0: Functor[S], S1: QueryFileF :<: S) {
     import ResultFile._
@@ -49,25 +55,28 @@ object QueryFile {
     def execute(plan: Fix[LogicalPlan], out: AFile): ExecM[ResultFile] =
       EitherT(WriterT(lift(ExecutePlan(plan, out))): G[FileSystemError \/ ResultFile])
 
-    def explain(plan: Fix[LogicalPlan]): F[PhaseResults] =
-      lift(Explain(plan))
-
     /** Returns the path to the result of executing the given [[LogicalPlan]] */
     def execute_(plan: Fix[LogicalPlan])
                 (implicit M: ManageFile.Ops[S]): ExecM[ResultFile] = {
-
-      val outFile = plan.foldMap {
-        case Fix(LogicalPlan.ReadF(p)) => Vector(p)
-        case _                         => Vector.empty
-      }.headOption flatMap pathToAbsFile cata (M.tempFileNear, M.anyTempFile)
-
       for {
-        out <- toExec(outFile)
+        out <- toExec(firstFile(plan) cata (M.tempFileNear, M.anyTempFile))
         rf0 <- execute(plan, out)
         rf1 =  user.getOrModify(rf0)
                  .map(usr => if (usr == out) Temp(usr) else User(usr))
                  .merge
       } yield rf1
+    }
+
+    /** Returns a description of how the the given logical plan will be executed
+      * along with any error encountered during planning.
+      */
+    def explain(plan: Fix[LogicalPlan])
+               (implicit M: ManageFile.Ops[S])
+               : F[(PhaseResults, Option[FileSystemError])] = {
+
+      firstFile(plan)
+        .cata(M.tempFileNear, M.anyTempFile)
+        .flatMap(f => lift(Explain(plan, f)))
     }
 
     /** Returns the source of values from the result of executing the given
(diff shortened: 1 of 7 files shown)


web/src/main/scala/quasar/api/server.scala
```diff
@@ -53,7 +53,6 @@ class ServerOps[WC: CodecJson, SC](
   configOps: ConfigOps[WC],
   defaultWC: WC,
   val webConfigLens: WebConfigLens[WC, SC]) {
-  import webConfigLens._
```
jedesah: Oh ya, I left that in there, because I am scared it means that I did not merge Dave's changes to server properly. But we will have to sit down together anyway, and make sure server.scala is doing everything it should be doing. So good thing to get rid of that warning (which would eventually become an error)

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
```diff
@@ -183,7 +183,7 @@ class CompileAndQueryServiceSpec extends Specification with FileSystemFixture wi
         query = Some(selectAll(filesystem.filename)),
         state = filesystem.state,
         status = Status.Ok,
-        response = "InMemory\nPlan(logical: Squash(Read(Path(\"" + pathString + "\"))))"
```
jedesah: I am curious what the rationale for this change is?

jedesah:
I am not sure I understand the link between not having streaming support in this PR and having to add a file parameter to the `Explain` term. Do we need to hit MongoDB in order to come up with the `PhaseResults`?

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
wemrysi: The in-memory interpreter doesn't currently support `ExecutePlan` or `Explain`, returning the `LogicalPlan` as the plan for executing itself isn't really accurate. Better to be explicit and just return a message stating that it is unsupported.

jedesah: Fair enough.

wemrysi:
The link is that the only way to currently execute a `LogicalPlan` requires the backend to provide the path out the output in the filesystem, thus every execution plan must have someplace to write the output to. It doesn't seem correct to then also ask the backend to explain how it plans on executing a query without also providing it someplace to write the planned results to.

The choice of whether the results of a plan should be written to the filesystem or streamed may also affect the plan itself. i.e. you get a different plan based on which you choose or maybe a particular `LogicalPlan` isn't executable in streaming form, etc.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -16,7 +16,8 @@ object QueryFile {
   final case class ExecutePlan(lp: Fix[LogicalPlan], out: AFile)
     extends QueryFile[(PhaseResults, FileSystemError \/ ResultFile)]
 
-  final case class Explain(lp: Fix[LogicalPlan]) extends QueryFile[PhaseResults]
+  final case class Explain(lp: Fix[LogicalPlan], out: AFile)
+    extends QueryFile[(PhaseResults, FileSystemError \/ Unit)]
```
jedesah: I am not sure what I think about this style where the result is stuffed in the Monad stack. It's certainly weird for me to have a function return `Unit`, then again, I can see how it might be useful to have the result in the monad stack itself for ease of composition.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -70,6 +67,12 @@ object QueryFile {
       } yield rf1
     }
 
+    /** Returns a description of how the the given logical plan will be executed. */
+    def explain(plan: Fix[LogicalPlan]): ExecM[Unit] = {
+      val outFile = firstFile(plan) getOrElse DefaultExplainOut
```
jedesah: I am not sure I totally grasp this. We are are grabbing the first file in the plan and setting that as the destination file? Why don't we do the same thing as `execute_` and use a `tempFileNear` or `anyTempFile`?

jedesah:
What are we going to do then when we do have streaming of results? Make the destination parameter optional in `Explain`?

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: You're right, not sure what I was thinking there, I'll fix it.

wemrysi:
Aye, probably something like that.

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: I see your point, we could try returning `(PhaseResults, Option[FileSystemError])` and not lift it into the stack.

wemrysi:
@jedesah Ok, I fixed the issues you raised, thanks for the feedback!

core/src/main/scala/quasar/fs/QueryFile.scala
jedesah: Btw, just out of curiosity, did you try this out and revert back to this for some reason?

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -70,6 +67,18 @@ object QueryFile {
       } yield rf1
     }
 
+    /** Returns a description of how the the given logical plan will be executed
+      * along with any error encountered during planning.
+      */
+    def explain(plan: Fix[LogicalPlan])
+               (implicit M: ManageFile.Ops[S])
+               : F[(PhaseResults, Option[FileSystemError])] = {
```
jedesah: Ah, I see you changed it here instead of at the Algebra level.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -122,11 +131,22 @@ object QueryFile {
       comp flatMap (lp => evaluate(lp).translate[CompExecM](execToCompExec))
     }
 
-    def explainQuery(query: sql.Expr, vars: Variables): SemanticErrsT[F,PhaseResults] = {
-      val writer = queryPlan(query, vars).run
-      val phases = writer.written
-      EitherT[F,SemanticErrors,PhaseResults](
-        writer.value.map(lp => explain(lp).map(physicalPhases => phases ++ physicalPhases)).sequenceU)
+    /** Returns a description of how the the given SQL^2 query will be executed
```
jedesah: there is a small typo here. the is repeated twice.

jedesah:
Not a big deal, but would have been easier to review changes if the commit was not rewritten. I will merge and we can fix the typo at some point :smile:

PR #1005, opened 2015-11-24
disable writing a temp file when checking connections

mossprescott:
We no longer strictly require the ability to write, now that some
queries can be run in "streaming" mode.

SD-1163 #done


mossprescott:
Merging based on @jedesah's thumbs up in chat. Not even waiting for Travis, because what could go wrong?

PR #1006, opened 2015-11-25
fix handling of views containing absolute paths

mossprescott:
SD-1165 #done

When the ViewBackend is created, it normalizes all paths in every view
query so that they can be composed with each other. Previously, this
normalization ignored absolute paths, leaving us with a mixture of
absolute and relative paths, and if an absolute path happened to refer
to another view, it would not be properly expanded when it came time to
evaluate the (outer) view's query.

Now all paths in view queries are converted to absolute, just to make
it easier to understand/explain. The locations of the views themselves
are still stored as relative and interpreted relative to '/', just
because that's how the Backend trait works. That means the fix here is
really just to convert paths in queries to relative when doing the view
lookup.

### DIFF
--- a/core/src/main/scala/quasar/backend.scala
+++ b/core/src/main/scala/quasar/backend.scala
@@ -280,9 +280,13 @@ object ThreadPoolAdapterBackend {
 final case class ViewBackend(backend: Backend, views0: Map[Path, ViewConfig]) extends Backend {
   import Backend._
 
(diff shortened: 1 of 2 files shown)


sellout:
This makes sense, and the test changes look good. Seems like pathy will help us avoid issues like this in future.

mossprescott:
Yes, I think pathy will help some, and I think switching to free algebras will, too.

PR #1007, opened 2015-11-25
Optimize $group operations.

sellout:
SD-953 #done

This rewrites $groups that duplicate keys as fields to remove the
duplication.

Some other changes included here:

* object keys that used the empty string now use “0”,
* BsonField.Leaf and .Index have been removed (we’ve only been using
  .Name for a while now),
* added Traverse1 and Comonad instances to AccumOp,
* fixed a bug in `reorderOps`, where we didn’t rewrite selectors across
  `$include`,
* added `Reshape.find`, which returns the path to the provided
  `Expression`, and
* fixed `simpleShape` to include “_id” when appropriate.


it/src/main/resources/tests/aliasedFieldSortedByOriginal.test
```diff
@@ -3,14 +3,14 @@
...
-                 { "Result Alias": "AL", "1":  567 },
-                 { "Result Alias": "AR", "1":  578 },
-                 { "Result Alias": "AZ", "1":  270 },
-                 { "Result Alias": "CA", "1": 1516 },
-                 { "Result Alias": "CO", "1":  414 },
-                 { "Result Alias": "CT", "1":  263 },
-                 { "Result Alias": "DC", "1":   24 },
-                 { "Result Alias": "DE", "1":   53 },
-                 { "Result Alias": "FL", "1":  804 }]
+    "expected": [{ "1":  195, "Result Alias": "AK" },
```
jdegoes: This is unfortunate. If we know when this will happen and it isn't too frequent, we can always introduce 2 new projects: one to name them something like `__t1`', etc., and then one to rename them to their final names, to force MongoDB to get the order right. I know the JIRA ticket is still under consideration.

it/src/main/resources/tests/groupedJoin.test
```diff
@@ -4,7 +4,7 @@
     "data": "slamengine_commits.data",
     "query": "SELECT p.author.login, COUNT(*) FROM slamengine_commits p INNER JOIN slamengine_commits c ON p.sha = c.sha GROUP BY p.author.login",
     "predicate": "containsExactly",
-    "expected": [{ "login": "mossprescott", "1": 15 },
-                 { "login": "sellout",      "1":  9 },
-                 { "login": "jdegoes",      "1":  6 }]
+    "expected": [{ "1": 15, "login": "mossprescott" },
+                 { "1":  9, "login": "sellout"      },
```
jdegoes: Hmm, looks like this is consistently in the wrong order -- are you sure there's no bug being introduced? If a shape has new fields, as one would expect after a `$group`, then MongoDB will respect the order. It's only existing fields that order is not respected (plus or minus).

sellout: If you look at the planner tests that changed, you can see that the rewritten projections keep the same order as the original `$group`. I also checked some of the integration tests (and just (re?)checked this one now) to make sure.

The final task is
```js
db.tmp.gen_0.aggregate(
  [
    {
      "$group": {
        "1": { "$sum": { "$literal": NumberInt(1) } },
        "_id": { "0": "$value.__tmp23" }
      }
    },
    { "$project": { "login": "$_id.0", "1": "$1" } }],
  { "allowDiskUse": true });
```
You can see the `$project` is ordered properly, and considering the grouping, plus the nesting/projecting of `login`, I don’t know how this ordering possibly comes from non-altered fields in the temp collection.

it/src/main/resources/tests/aliasedFieldSortedByOriginal.test
sellout: I don’t think renaming them is enough to fix the ordering issue. And I would _hope_ that Mongo would simplify stuff like that before executing the pipeline anyway.

Also, I have no idea what triggers this behavior – at least the cases introduced here, that don’t seem to meet the criteria of that MongoDB issue.

it/src/main/resources/tests/groupedJoin.test
mossprescott: This is consistent with what I observed when I was looking into this a while back. Existing field names from any previous pipeline stage always show up before new field names. In the $project, `1` is preserved and `login` is added, so it goes after. In these cases, I guess that means anything projected form the grouping key will end up last.

See https://slamdata.atlassian.net/browse/SD-600.

it/src/main/resources/tests/aliasedFieldSortedByOriginal.test
jdegoes: MongoDB mutably updates documents in place in the BSON format, which is probably why fields left over from a previous pipeline are not recreated, but merely reused. Asya @ MongoDB indicated forcing a new document structure (via introducing all new names) and then redoing the old structure would solve the problem.

sellout: You’re right – just aliasing the fields is enough to fix the ordering. So, perhaps this boosts the priority of SD-600?

jdegoes: > You’re right – just aliasing the fields is enough to fix the ordering. So, perhaps this boosts the priority of SD-600?

The problem is it will decrease performance, so I'd rather push off a bit.

I'd suggest, however, that our tests have an option (maybe they already do?) to ignore column order, so we can stop "infecting" these tests with MongoDB implementation details.

core/src/main/scala/quasar/physical/mongodb/bson.scala
```diff
@@ -239,18 +239,23 @@ sealed trait BsonField {
...
-    case (Path(x), y: Leaf) => Path(NonEmptyList.nel(x.head, x.tail :+ y))
-    case (y: Leaf, Path(x)) => Path(NonEmptyList.nel(y, x.list))
-    case (x: Leaf, y: Leaf) => Path(NonEmptyList.nels(x, y))
-  }
-
-  def \\ (tail: List[BsonField]): BsonField = if (tail.isEmpty) this else this match {
-    case Path(p) => Path(NonEmptyList.nel(p.head, p.tail ::: tail.flatMap(_.flatten.toList)))
-    case l: Leaf => Path(NonEmptyList.nel(l, tail.flatMap(_.flatten.toList)))
+    case (Path(x),     Path(y))     => Path(x ⊹ y)
+    case (Path(x),     y @ Name(_)) => Path(x ⊹ NonEmptyList(y))
```
jedesah: Why not `y: Name`?

sellout: Because `:` pattern matches use runtime type checks (`isInstanceOf`).

jedesah: I am not sure I understand the problem as `Name.unapply` is also a runtime typecheck

sellout: Ok, so I don’t know a lot about scalac internals and may very well be wrong here, but I would hope that `case class` destructuring (as `Name` doesn’t have an explicit `unapply`) wouldn’t require a runtime type check. Something like `Path(y)` doesn’t make wartremover complain about `isInstanceOf` like `:` does). Although, I don’t know how wartremover feels about `@`.

jedesah: It seems to me either way the jvm needs to check the type of the object. I mean, it's kind of the whole point of this particular pattern match, to check the type of `this` and `that` at runtime. I am not sure why we would care exactly how scalac desugars the pattern match into a runtime type check. But if the old way triggers a wart, then I guess we should write it the new way in order to work around what I would consider a WartRemover bug.

wemrysi: Not sure if this helps, but you could just use `that`, i.e.

`case (Path(x), Name(_)) => Path(x ⊹ NonEmptyList(that))`

jdegoes:
@jedesah I'd like this in 2.3.3 (tomorrow), so please review when you can.

core/src/main/scala/quasar/physical/mongodb/bson.scala
mossprescott: As I understand it, we're avoiding `:` in pattern matching because it can be used in ways that generate warnings from either scalac or wartremover, but also because it has some power that we don't need. That is, you only *need* to use a `:` if you have an "open" type and want to match values whose specific types you don't know about. Our traits are always `sealed` and therefore you can use extractors and should never have to do an explicit type check.

On the question of whether or not it's a runtime type-check, I think the right way to think about that is that something like `case y @ Name(_)` is a type-check for the JVM, but another implementation would just be checking the tag (that is, it only has to differentiate between cases of the sum type, not any possible type). The fact that Scala on the JVM (probably) conflates the two is a detail we don't have to worry about. On the other hand, `case x: Name` might be a true runtime test, if Name is a sub-type of whatever is being matched, and/or if Name itself has sub-types.

I'm not sure I've got all that quite right, but it's along those lines. Anyone have a different understanding and can enlighten us?

jdegoes: Type checking exposes the fact that ADTs are implemented with inheritance, whereas pattern matching keeps that implementation detail secret -- indeed, pattern matching can work even if you choose not to implement your ADTs with inheritance (e.g. fold functions).

sellout: @wemrysi In this case, we can’t use `that`, because the NonEmptyList can only contain `Name`s, not `BsonField`. I.e., `Name` is doing double-duty as a component of the BsonField sum type and as a type in its own right.

sellout: @wemrysi I do want to separate those at some point, where we either have a `Name` containing one `Field`, or a `Path` containing at least two `Field`s.

jedesah: @mossprescott Thx for the detailed explanation which comes to similar conclusions as an offline conversation I had with @sellout right before lunch. It's too bad about the numerous `_` required for a case class with many fields, but I guess we'll just have to live with that for now as there is no better alternative.

core/src/main/scala/quasar/physical/mongodb/optimize/optimize.scala
```diff
@@ -78,6 +78,45 @@ package object optimize {
...
+    val simplifyGroupƒ:
+        WorkflowF[Workflow] => Option[WorkflowF[Workflow]] = {
+      case $Group(src, Grouped(cont), id) =>
+        val (newCont, proj) =
+          cont.foldLeft[(ListMap[BsonField.Name, Accumulator], ListMap[BsonField.Name, Reshape.Shape])](
+            (ListMap(), ListMap())) {
+            case ((newCont, proj), (k, v)) =>
+              lazy val default =
+                (newCont + (k -> v), proj + (k -> $include().right))
+              def meh(loc: Option[DocVar]) =
```
jedesah: Does `meh` stand for something? or is it just like "meh, don't want to come up with a better name"?

core/src/main/scala/quasar/physical/mongodb/optimize/optimize.scala
```diff
@@ -78,6 +78,45 @@ package object optimize {
 
     def deleteUnusedFields(op: Workflow) = deleteUnusedFields0(op, None)
 
+    /** Converts a \$group with fields that duplicate keys into a \$group
+      * without those fields followed by a \$project that projects those fields
+      * from the key.
+      */
+    val simplifyGroupƒ:
```
jedesah: Not gonna lie, this function is really hard for me to follow, but that might be because I am not yet familiar with this part of the code base.

I think a brief description of `$Group` and or `Grouped` might be helpful, because it's not obvious for me to figure out what exactly is `$Group`, `Grouped`, `cont`, `Shape` and why the shape is called `id` in this context?

core/src/main/scala/quasar/physical/mongodb/optimize/optimize.scala
sellout: Ahahaha – it’s “I haven’t figured out what this function does yet” – should have been renamed before committing.

core/src/main/scala/quasar/physical/mongodb/bson.scala
wemrysi: @sellout Ah, thanks for pointing that out. Separating them eventually sounds like a good idea.

core/src/main/scala/quasar/physical/mongodb/optimize/optimize.scala
sellout: `$Group` is the MongoDB `$group` pipeline operator (aggregates values that are grouped by some expression).

`Grouped` is a newtype for a Map of fields to aggregations (e.g., `select count(*) as n` becomes an entry in that map like `BsonField.Name("n") -> $sum($literal(1))`).

`cont` is short for “contents” – what we’re grouping.

And finally, the way grouping works is to create a new `_id` field, and all documents with the same value in that field have their contents aggregated together as directed by the `Grouped` value. `Shape` is the type of that `_id` field, which is `Reshape \/ Expression` – where `Expression` is some simple-valued MongoDB expression operator and `Reshape` is a nested document recursively containing `Shape`s.

core/src/test/scala/quasar/physical/mongodb/optimize.scala
```diff
@@ -18,6 +15,48 @@ class OptimizeSpecs extends Specification with TreeMatchers {
...
+      chain(
+        $read(Collection("db", "zips")),
+        $group(
+          Grouped(ListMap(
+            BsonField.Name("city") -> $push($field("city")))),
+          Reshape(ListMap(
+            BsonField.Name("0") -> $field("city").right)).left))
+        .transCata(once(simplifyGroupƒ)) must
+      beTree(chain(
+        $read(Collection("db", "zips")),
```
jedesah: I think we can avoid duplication and make it easier on the reader here by saying that this plan should not change, ie. `lp.transCata(once(simplifyGroupƒ)) must beTree(lp)` where `lp` is this particular `LogicalPlan`

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -3233,6 +3263,17 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
         case _ => Nil
       }) aka "group ops pushing $$ROOT"
 
+    def appropriateColumns0(wf: Workflow, q: Query) = {
+      val fields = fieldNames(wf).map(_.filterNot(_ ≟ "_id"))
+      fields aka "column order" must beSome(columnNames(q))
```
jedesah: huh, neat, never saw this `aka` before.

core/src/test/scala/quasar/physical/mongodb/optimize.scala
sellout: 👍🏾

core/src/test/scala/quasar/physical/mongodb/optimize.scala
```diff
@@ -18,6 +15,48 @@ class OptimizeSpecs extends Specification with TreeMatchers {
...
 
+  "simplifyGroupƒ" should {
+    "elide useless reduction" in {
+      chain(
+        $read(Collection("db", "zips")),
+        $group(
+          Grouped(ListMap(
+            BsonField.Name("city") -> $sum($field("city")))),
+          Reshape(ListMap(
+            BsonField.Name("0") -> $field("city").right)).left))
```
jedesah: As discussed, probably don't need this MongoDB specific Bson wrapper for arrays.

core/src/test/scala/quasar/physical/mongodb/optimize.scala
```diff
@@ -18,6 +15,48 @@ class OptimizeSpecs extends Specification with TreeMatchers {
 
   import optimize.pipeline._
 
+  "simplifyGroupƒ" should {
+    "elide useless reduction" in {
+      chain(
+        $read(Collection("db", "zips")),
+        $group(
+          Grouped(ListMap(
+            BsonField.Name("city") -> $sum($field("city")))),
```
jedesah: "population" might be a better sample name here

jedesah:
Otherwise, LGTM, but some of the code here is a little too unfamiliar to me in order for me to approve with a high level of confidence without spending a considerable amount more time reviewing.

it/src/main/resources/tests/aliasedFieldSortedByOriginal.test
sellout: I thought this issue existed, but I couldn’t find it, so this is now SD-1174.

PR #1009, opened 2015-11-25
Fixes for recent PRs

wemrysi:
A couple of fixes that slipped through the last PR to the web feature


wemrysi:
@jedesah This seems to work, but we're getting a weird compile error for one of the build jobs, it doesn't happen for me locally (and one of the jobs worked) so I'm ok to merge this. Have you ever experienced the error that travis reports?

jedesah:
@wemrysi Just had a look, looks like a compiler crash to me. Let's merge it.

wemrysi:
@jedesah Ah, it looks `doc` related, which is why it only fails the first build. I think one of the issues is fixed upstream, the other two were introduced in `Interpreter`.

PR #1010, opened 2015-11-26
Fix test failures in CompileAndQueryServiceSpec

jedesah:
- InMemState now has pre-canned responses for queries
- other minor improvements
- Remove scaladoc links in Interpreter causing doc generation failures
- Various fixes in CompileAndQueryServiceSpec in order to make all tests pass

### DIFF
--- a/core/src/main/scala/quasar/fs/InMemory.scala
+++ b/core/src/main/scala/quasar/fs/InMemory.scala
@@ -4,6 +4,7 @@ package fs
 import quasar.Predef._
 import quasar.fp._
 import quasar.recursionschemes.Fix
+import quasar.Planner.UnsupportedPlan
 
 import scalaz._, Scalaz._
 import scalaz.concurrent.Task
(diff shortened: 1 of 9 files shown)


core/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -182,6 +203,12 @@ object InMemory {
   private val rmL: InMemState @> RM =
     Lens.lensg(s => m => s.copy(rm = m), _.rm)
 
+  private val contentsL: InMemState @> FileMap =
```
wemrysi: This is duplicating `fmL` above, probably just want to rename that one.

core/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -125,20 +133,33 @@ object InMemory {
 
   val queryFile: QueryFile ~> InMemoryFs = new (QueryFile ~> InMemoryFs) {
     def apply[A](qf: QueryFile[A]) = qf match {
-      case ExecutePlan(_, out) =>
-        (Vector(unsupported), ResultFile.User(out).right[FileSystemError])
-          .point[InMemoryFs]
+      case ExecutePlan(lp, out) =>
```
wemrysi: If we add another lens, we can rewrite this a bit more succinctly:

```scala
def queryResponseL(lp: Fix[LogicalPlan]): InMemState @> Option[Vector[Data]] =
  Lens.mapVLens(lp) <=< queryResponsesL

for {
  rFile  <- fileL(out).st map (_ ? ResultFile.User(out) | ResultFile.Temp(out))
  lpData <- queryResponseL(lp)
  noSupportMsg   = "In Memory interpreter does not currently support this plan"
  noSupportError = PlannerError(lp, UnsupportedPlan(lp.unFix, Some(noSupportMsg)))
  result <- lpData cata (data =>
              fileL(out).set(Some(data)).as(rFile.right[FileSystemError]),
              noSupportError.left[ResultFile].point[InMemoryFs])
  phase  <- queryResponsesL.st map (lookup(lp, _))
} yield (Vector(phase), result)
```

jedesah: This doesn't quite compile, but I did use lenses to make the implementation more concise, ended up not having to add that other lens for now.

core/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -125,20 +133,29 @@ object InMemory {
...
+        val noSupportMsg = "In Memory interpreter does not currently support this plan"
+        queryResponsesL.st.flatMap { queryResponses =>
+          val result = queryResponses.get(lp).cata(
+            data => fileL(out).st.bimap(
```
wemrysi: Can reduce this to `fileL(out) assigno Some(data) map (previous => ...)`

wemrysi:
Other than my last minor comment, :+1:

core/src/main/scala/quasar/fs/InMemory.scala
jedesah: Awesome! Was looking for something like that but did not find it!

PR #1011, opened 2015-11-26
Have Mongo’s JS return Ints when appropriate.

sellout:
Both `Length` and `Count` are defined to return Int, but we’ve just been
using the native JS numbers. This uses Mongo’s `NumberLong` to make sure
the results are integers.


drostron:
👍🏿

PR #989, opened 2015-11-16
Translate web project for Free Architecture

jedesah:

PR #1012, opened 2015-11-26
handle variables, offset, and limit in GET /query

mossprescott:
SD-1150 #done

`GET /query` requests now accept `offset`, `limit`, and `var.` parameters, and `POST` now expects variables to have the `var.` prefix as well.

**Note:** this is a breaking API change. Clients need to be updated to prefix variable names when calling `POST /query`.

Also implemented variables in `/compile` (it was already mentioned in the README, but apparently never implemented).

Tests for all of the above that actually inspect the resulting plans.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,9 +219,13 @@ Type `help` for information on other commands.
 
 The server provides a simple JSON API.
 
-### GET /query/fs/[path]?q=[query]
+### GET /query/fs/[path]?q=[query]&offset=[offset]&limit=[limit]&var.[foo]=[value]
 
-Executes a SQL query, contained in the single, required query parameter, on the backend responsible for the request path.
+Executes a SQL query, contained in the required `q` parameter, on the backend responsible for the request path.
+
+Optional `offset` and `limit` parameters can be specified to page through the results, and are interpreted the same way as in `GET /data`.
+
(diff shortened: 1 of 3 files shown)


mossprescott:
@jedesah: You got this one because you've been porting the API and I figured you might be interested to see what changes need to get merged.

Also, I'd like to talk to you about how these tests could be written in the free algebra world. Currently, we have the whole backend stubbed and I had to jump through some hoops to get the query captured in a way that I could inspect in the tests. Ideally, I would construct a req, invoke the service and then inspect the resulting `QueryFile` term, or maybe interpret it to a log of "here's the query that would get run". But from a quick look at #989 and it looks like we might be there quite yet (or at least we weren't when I looked).

jdegoes:
@jedesah I'd like this in 2.3.3 (tomorrow), so please review when you can.

jedesah:
@mossprescott 

> Ideally, I would construct a req, invoke the service and then inspect the resulting QueryFile term, or maybe interpret it to a log of "here's the query that would get run".

Unfortunately, we have to interpret the algebra into the `Task` monad in order to create an `HttpService` which means we can't do any fancy custom interpretation in the web api tests. However a similar approach to what you have taken in this PR would be possible in the new tests.

jedesah:
LGTM.

PR #1013, opened 2015-11-27
Attempt to make Collection test less volatile

wemrysi:
Scalacheck tests for creating a Collection from a pathy path were
occasionally failing due to scalacheck not being able to generate enough
inputs satisfying the test constraints, this reformulates the test to hopefully reduce the likelyhood
of this happening.

### DIFF
--- a/core/src/test/scala/quasar/physical/mongodb/collection.scala
+++ b/core/src/test/scala/quasar/physical/mongodb/collection.scala
@@ -287,13 +287,12 @@ class CollectionSpec extends Specification with ScalaCheck with DisjunctionMatch
         beRightDisjunction
     }
 
-    "never emit an invalid db name" ! prop { f: AbsFileOf[SpecialStr] =>
-      val file = f.path
-      val notInRootDir = parentDir(file).fold(false)(_ != rootDir)
-      val notTooLong = posixCodec.printPath(file).length < 30
+    "never emit an invalid db name" ! prop { (db: SpecialStr, c: SpecialStr) =>
+      val f = rootDir </> dir(db.str) </> file(c.str)
+      val notTooLong = posixCodec.printPath(f).length < 30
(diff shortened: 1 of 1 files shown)


core/src/test/scala/quasar/physical/mongodb/collection.scala
```diff
@@ -287,13 +287,12 @@ class CollectionSpec extends Specification with ScalaCheck with DisjunctionMatch
         beRightDisjunction
     }
 
-    "never emit an invalid db name" ! prop { f: AbsFileOf[SpecialStr] =>
-      val file = f.path
-      val notInRootDir = parentDir(file).fold(false)(_ != rootDir)
-      val notTooLong = posixCodec.printPath(file).length < 30
+    "never emit an invalid db name" ! prop { (db: SpecialStr, c: SpecialStr) =>
+      val f = rootDir </> dir(db.str) </> file(c.str)
```
sellout: Shouldn’t `c` be `RelFileOf[SpecialStr]`, so we preserve the semantics of not having files at different depths?

wemrysi: Would you explain what you mean by "not having files at different depths", I don't understand.

sellout: Oh, first of all, I meant to write “preserve the semantics of having files at different depths” (no “not”).

Given the previous `notInRootDir` constraint, it seems like `AbsFileOf` would create files like `/someFile`, `/someDir/someFile`, `/someDir/anotherDir/someFile`, etc. whereas the new approach always generates `/someDir/someFile`.

sellout: And … now that I read the name of the test, I see that that isn’t important. We only care about the db name. So … nevermind.

PR #1014, opened 2015-11-30
avoid excessive backtracking when parsing deeply-nested parens

mossprescott:
SD-1169 #done

Formerly, the parser would always attempt to parse a parenthesized
expression as a set literal, then fail when it turned out to contain
only one expression, and then re-parse it as a single expression. So it
was O(2^n) for certain patterns, including the way the pretty printer
renders long sequences of "or".

A simple fix is just to parse once and then decide based on the number
of sub-expressions whether to parse to SetLiteral or just an expression.

Three new tests (which pass before the fix) just verify the expected
behavior for 0-, 1-, and 2-element parenthesized expressions, and one
new test is the query from the bug report, which used to take at least
many minutes to parse, and now takes well under a second.

### DIFF
--- a/core/src/main/scala/quasar/sql/parser.scala
+++ b/core/src/main/scala/quasar/sql/parser.scala
@@ -185,17 +185,8 @@ class SQLParser extends StandardTokenParsers {
     }
   }
 
-  def rep2sep[T, U](p: => Parser[T], s: => Parser[U]) =
-    p ~ rep1(s ~> p) ^^ { case x ~ y => x :: y }
-
-  def set_literal: Parser[Expr] =
-    op("(") ~ op(")") ^^^ SetLiteral(Nil) |
-    (op("(") ~> rep2sep(expr, op(",")) <~ op(")")) ^^ (SetLiteral(_))
-
   def array_literal: Parser[Expr] =
     (op("[") ~> repsep(expr, op(",")) <~ op("]")) ^^ (ArrayLiteral(_))
-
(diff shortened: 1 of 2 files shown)


jdegoes:
@drostron I'd like this in 2.3.3 (tomorrow), so please review when you can.

core/src/main/scala/quasar/sql/parser.scala
```diff
@@ -185,18 +185,9 @@ class SQLParser extends StandardTokenParsers {
...
-    p ~ rep1(s ~> p) ^^ { case x ~ y => x :: y }
-
-  def set_literal: Parser[Expr] =
-    op("(") ~ op(")") ^^^ SetLiteral(Nil) |
-    (op("(") ~> rep2sep(expr, op(",")) <~ op(")")) ^^ (SetLiteral(_))
-
   def array_literal: Parser[Expr] =
     (op("[") ~> repsep(expr, op(",")) <~ op("]")) ^^ (ArrayLiteral(_))
 
-  def set_expr: Parser[Expr] = select | set_literal
```
drostron: Was the select part of this unnecessary? I don't see it carried over.

mossprescott: So it seems. I'm pretty sure a select in that position will always be surrounded by parens anyway, so it probably wasn't even correct to parse it without.

drostron:
👍🏽

PR #1015, opened 2015-11-30
[SD-1170] Mounting algebra

wemrysi:
Adds a `Mounting` algebra describing the operations of managing filesystem mounts. A spec is included, implemented in terms of the algebra that describes how an interpreter of the algebra must behave.

A single interpreter is also defined, that maintains mount state in memory, allowing effectful mounting via a callback function.

### DIFF
--- a/core/src/main/scala/quasar/fp/prism.scala
+++ b/core/src/main/scala/quasar/fp/prism.scala
@@ -1,0 +1,24 @@
+package quasar.fp
+
+import monocle.Prism
+import scalaz._
+
+trait PrismInstances {
+  import Liskov.<~<
+
+  // TODO: See if we can implement this once for all tuples using shapeless.
+  implicit class PrismOps[A, B](prism: Prism[A, B]) {
+    def apply(b: B): A = prism.reverseGet(b)
+
+    def apply[C, D](c: C, d: D)(implicit ev: (C, D) <~< B): A =
+      apply(ev((c, d)))
+
+    def apply[C, D, E](c: C, d: D, e: E)(implicit ev: (C, D, E) <~< B): A =
+      apply(ev((c, d, e)))
+
+    def apply[C, D, E, F](c: C, d: D, e: E, f: F)(implicit ev: (C, D, E, F) <~< B): A =
+      apply(ev((c, d, e, f)))
+  }
+}
+
+object prism extends PrismInstances
--- a/core/src/main/scala/quasar/mount/Mounting.scala
+++ b/core/src/main/scala/quasar/mount/Mounting.scala
@@ -1,0 +1,123 @@
+package quasar.mount
+
+import quasar.Predef._
+import quasar.Variables
+import quasar.fp._
+import quasar.fs.{Path => _, _}
+import quasar.sql._
+
+import monocle.Prism
+import monocle.std.{disjunction => D}
+import pathy._, Path._
+import scalaz._
+import scalaz.syntax.std.option._
+import scalaz.syntax.apply._
+import scalaz.syntax.monadError._
+
+sealed trait Mounting[A]
+
+object Mounting {
+  final case class Lookup(path: APath)
+    extends Mounting[Option[MountConfig2]]
+
+  final case class MountView(loc: AFile, query: Expr, vars: Variables)
+    extends Mounting[MountingError \/ Unit]
+
+  final case class MountFileSystem(loc: ADir, typ: FileSystemType, uri: ConnectionUri)
+    extends Mounting[MountingError \/ Unit]
+
+  final case class Unmount(path: APath)
+    extends Mounting[MountingError \/ Unit]
+
+  final class Ops[S[_]](implicit S0: Functor[S], S1: MountingF :<: S) {
+    import MountConfig2._
+
+    type F[A] = Free[S, A]
+    type M[A] = EitherT[F, MountingError, A]
+
+    /** Returns the mount configuration for the given mount path or nothing
+      * if the path does not refer to a mount.
+      */
+    def lookup(path: APath): OptionT[F, MountConfig2] =
+      OptionT(lift(Lookup(path)))
+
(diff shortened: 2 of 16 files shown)


wemrysi:
@sellout @mossprescott @jedesah @drostron This is the direction I'm currently taking with the mounting algebra, if you have a few minutes I'd appreciate any comments, reservations, etc.

core/src/main/scala/quasar/mount/Mounting.scala
```diff
@@ -0,0 +1,124 @@
...
+sealed trait Mounting[A]
+
+object Mounting {
+  final case class Lookup(path: APath)
+    extends Mounting[Option[MountConfig2]]
+
+  final case class MountView(loc: AFile, query: Expr)
+    extends Mounting[MountingError \/ Unit]
+
+  final case class MountFileSystem(loc: ADir, typ: FileSystemType, config: Json)
```
mossprescott: Given what we know, it's probably safe to make this take just the type and a `connectionUri`.

core/src/main/scala/quasar/mount/Mounting.scala
```diff
@@ -0,0 +1,124 @@
...
+import scalaz.syntax.apply._
+import scalaz.syntax.monadError._
+
+sealed trait Mounting[A]
+
+object Mounting {
+  final case class Lookup(path: APath)
+    extends Mounting[Option[MountConfig2]]
+
+  final case class MountView(loc: AFile, query: Expr)
```
mossprescott: Just a heads up that this becomes `query: Expr, variables: Variables` in my PR coming up probably today.

core/src/main/scala/quasar/fp/prism.scala
```diff
@@ -0,0 +1,25 @@
+package quasar
```
mossprescott: By the way, why are these new files not getting the Apache license head comment? Did we break that plugin?

core/src/main/scala/quasar/fp/prism.scala
```diff
@@ -0,0 +1,25 @@
...
+package fp
```
mossprescott: We haven'y used these "exploded" package declarations, as far a I know. I'd say let's be consistent and use the on-liner unless there's a reason to switch.

core/src/main/scala/quasar/fp/prism.scala
wemrysi: Hm, it would seem so.

core/src/main/scala/quasar/fp/prism.scala
wemrysi: Less explicit importing, afaik.

jdegoes: > Less explicit importing, afaik.

Anti-feature. :smiling_imp:

sellout: Yeah, I agree with @jdegoes. Especially in this case, we don’t want anything else to leak into FP. It should be at the top of the dependency graph.

mossprescott:
Will there be a way to enumerate mounts in this algebra? Because they will certainly show up in QueryFile's `ls`, which makes these two algebras oddly asymmetrical views on the same object. I wonder if they should be more closely integrated.

@jdegoes has talked about a "metadata store" (see [SD-979](https://slamdata.atlassian.net/browse/SD-979)), which would seem to encompass both.

core/src/main/scala/quasar/fp/prism.scala
wemrysi: In general, I agree, though for a type in `a.b.c` having members of `a` and `b` in scope seems reasonable given the hierarchical structure. However, I don't hold this strongly so am fine with changing it.

mossprescott: Ah, I see. Amazing how there's a new syntax trap to learn about every day in Scala. :-P

sellout: I think the useful imports tends to go the other way … a package is most likely to use its subpackages, and any sibling/parent package imports need to be more explicit. Ideally, siblings would be independent, and parent packages would contain the parts that tie them together.

wemrysi:
@mossprescott There could be a way to enumerate mounts, though not sure that is needed by anything at the moment. Though you raise an interesting point as to whether `ls` should be concerned with mounts, or just list paths, especially since the filesystem algebras have no notion of mounting.

core/src/main/scala/quasar/fp/prism.scala
wemrysi: That's true, though you also tend to see abstractions further up in the hierarchy and then implementations of said abstractions further down.

core/src/main/scala/quasar/mount/Mounting.scala
wemrysi: Sounds reasonable, if that's the case, then what about changing the JSON representation as well:

```json
{
  "/some/path": { "mongodb": "mongodb://..." },
  "/another/path": { "view": "sql2:///..." }
}
```

i.e., do we still need the intermediate object with the `connectionUri` key?

wemrysi: Or even

```json
{
  "/some/path": ["mongodb", "mongodb://..."],
  "/another/path": ["view", "sql2:///..."]
}
```

mossprescott: Sure, except that it's a breaking change for the front-end to deal with, as well as making every existing user's configuration invalid, so I guess I don't see the point unless we're making some significant improvement, which I don't think this is.

wemrysi: True, and the existing one is more forward-compatible.

wemrysi:
Looking into that compile failure...

drostron:
👍🏽

PR #1016, opened 2015-11-30
import Fix._

mossprescott:
This fixes the compile on master.

Obviously, it's not safe to merge without rebasing, but github said it was fine, so I got lazy. Not sure why this additional import is now necessary, but it certainly fixes the compile.

### DIFF
--- a/web/src/test/scala/quasar/api/fs.scala
+++ b/web/src/test/scala/quasar/api/fs.scala
@@ -3,7 +3,7 @@ package quasar.api
 import quasar.Predef._
 import quasar.RenderTree.ops._
 import quasar.api.Mock.ActionType
-import quasar.recursionschemes._, Recursive.ops._, FunctorT.ops._
+import quasar.recursionschemes._, Fix._, Recursive.ops._, FunctorT.ops._
 import quasar.fp._
 import quasar._, Backend._, Evaluator._
 import quasar.config._


wemrysi:
:+1: assuming travis agrees

mossprescott:
What do you think? Should we always rebase on current master (and re-run the tests) before merging? We are definitely not in that habit, even before we started rewriting history. Or is this just one of those things that happens sometimes?

wemrysi:
Hm, if there aren't any conflicts, then maybe just merging master into the branch to test integration is good enough?

jedesah:
@mossprescott I think it's just something that happens from time to time. I think we can try and solve this with tooling if we think it's an important problem, but I am not sure that requiring manual rebasing on master in the absence of merge conflicts is going to be worth the overhead.

sellout:
I think this is the argument for a DMZ branch.

mossprescott:
Right, a DMZ is a better fix. I guess it probably isn't any more likely than it ever was, so never mind.

sellout:
So, fine that this got merged as a hot fix … but we should _not_ have to import `Fix._`.

This seems to be an implicit lookup issue, since those are the only things in `object Fix`. In general, it seems that lookup does succeed in finding implicits in un-imported companion objects. Why didn’t it work in this case?

mossprescott:
Beats me. I just went to the file I copied the offending code out of, compared the imports, and added the missing one. I got no compile error and no unused import, so /shrug.

Does unused import detection even work when implicit resolution is involved? I vaguely remember seeing that mentioned in somebody's release notes.

PR #1017, opened 2015-12-01
handle variable bindings in views

mossprescott:
SD-1143 #done

Accept variable bindings in the connectionUri for a view mount. These are parsed out to a Variables map in ViewConfig (so they can be re-serialied later), and substituted into the view query when it is mounted. Unbound variables lead to errors 1) when a view mount is added through the API and 2) whenever a view query is evaluated through Backend's methods.

A few new bits and pieces:
- Yet another error wrapper to embed CompilationError in EnvironmentError.
- A convenience `substVars()` which doesn't need type annotations, since it's now being used in a third place.
- Replace `rewriteRelations()` with `rewriteRelationsM()`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,13 +1,9 @@
 import sbt._
 import Keys._
+import CustomKeys._
 import de.heikoseeberger.sbtheader.license.Apache2_0
 import de.heikoseeberger.sbtheader.HeaderPlugin
 import scoverage._
-
-val scalazVersion  = "7.1.4"
-val slcVersion     = "0.4"
-val monocleVersion = "1.1.1"
-val pathyVersion   = "0.0.3"
 
 // Exclusive execution settings
 lazy val ExclusiveTests = config("exclusive") extend Test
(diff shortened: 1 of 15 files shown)


sellout:
😱

core/src/main/scala/quasar/config/config.scala
```diff
@@ -61,14 +62,18 @@ object MongoDbConfig {
...
+  def validate(path: EnginePath) = for {
+    _ <- if (path.relative) -\/(InvalidConfig("Not an absolute path: " + path))
+          else if (path.pureDir) -\/(InvalidConfig("Not a file path: " + path))
+          else \/-(())
+    _ <- Variables.substVars(query, variables).leftMap(e => EnvCompError(CSemanticError(e)))
+  } yield ()
 }
 object ViewConfig {
-  private val UriPattern = "([a-z][a-z0-9+-.]*):///\\?q=(.+)".r
+  private val VarPatternStr = "var\\.([^=]+)=([^&]+)"
```
wemrysi: Can we switch to using `org.http4s.Uri` for the parsing here?

core/src/main/scala/quasar/variables.scala
```diff
@@ -40,4 +40,7 @@ object Variables {
           .leftMap(VariableParseError(VarName(name), varValue, _)))
     case x => Fix(x).right
   }
+
+  def substVars(expr: Expr, variables: Variables) =
```
wemrysi: Mind adding a return type here?

core/src/main/scala/quasar/config/config.scala
mossprescott: I looked at it before and it seemed like it wasn't very helpful but now that the regex's have multiplied I think it's worth giving it another try.

core/src/main/scala/quasar/variables.scala
mossprescott: Sure.

core/src/main/scala/quasar/config/config.scala
wemrysi: I tried parsing some view uris with it at the repl, seemed to work ok on the ones I tried. You can also get a nice `Map` of the query parameters.

core/src/main/scala/quasar/backend.scala
```diff
@@ -262,18 +262,21 @@ final case class ViewBackend(backend: Backend, views0: Map[Path, ViewConfig]) ex
   // arrive to the `0` methods of the trait), and all paths inside the
   // queries are interpreted relative to each view's parent dir (and as
   // absolute, because at least that way they're all consistent).
-  val views: Map[Path, sql.Expr] =
-    views0.map { case (p, ViewConfig(q)) =>
-      p.asRelative -> sql.relativizePaths(q, p.parent.asAbsolute).valueOr(κ(q))
+  val views: Map[Path, SemanticError \/ sql.Expr] =
```
wemrysi: Hm, what is the rationale for allowing a mount that results in an error? I'd prefer said mount be rejected outright rather than create a mount that we know will never work.

mossprescott: The API does reject a bad mount, but we still have to handle the case that an invalid view is found in the config file (which could happen if it's edited manually or if we maybe if we change the format down the road). Reporting the problem when the view is used seems more user-friendly than, say, logging and dropping it at load time.

core/src/main/scala/quasar/config/config.scala
```diff
@@ -61,14 +62,18 @@ object MongoDbConfig {
     casecodec1(MongoDbConfig.apply, MongoDbConfig.unapply)("connectionUri")
 }
 
-final case class ViewConfig(query: sql.Expr) extends MountConfig {
-  def validate(path: EnginePath) =
-    if (path.relative) -\/(InvalidConfig("Not an absolute path: " + path))
-    else if (path.pureDir) -\/(InvalidConfig("Not a file path: " + path))
-    else \/-(())
+final case class ViewConfig(query: sql.Expr, variables: Variables) extends MountConfig {
```
wemrysi: What do you think about `final case class ViewConfig(rawQuery: sql.Expr, variables: Variables, substQuery: sql.Expr)` where `substQuery` (is there a better name for this?) is the successful result of binding `rawQuery` using `variables`?

You could even make the constructor private and replace it with `apply(expr: sql.Expr, vars: Variables): SemanticError \/ ViewConfig` which guarantees correct configs.

core/src/main/scala/quasar/backend.scala
wemrysi: Hm, well you could also have a malformed URI in the config, which has a different error behavior than this, right?

Ideally, we wouldn't create any mounts that we know to be invalid. There is an argument to be made for lazy evaluation of mounts loaded from the config file for the sake of speed, and if we were going to do this then I think lazily evaluating view mounts (or just all mounts) would be fine. But even then, I'm not sure actually creating the mount if the one we found is invalid, we'd need a way to communicate this in some sort of admin interface or whatnot, maybe even allow the user to correct it there and update the config file or "metadata store".

wemrysi: Whoops, wasn't finished with that previous one, more to come...

core/src/main/scala/quasar/config/config.scala
mossprescott: Are you trying to get rid of the duplication between this method and the ViewBackend impl., or trying to eliminate invalid ViewConfigs from existence? If the former, then I would suggest adding `substQuery` as a lazy val. That would make it impossible to construct self-contradictory configs. But it think the type is still `SemanticError \/ Expr`....

 If it's the latter, see my reply to your other comment (above).

core/src/main/scala/quasar/config/config.scala
mossprescott: Yes, I'm getting further with it now after giving it anther look. I'm not crazy about having the dependency on http4s in the core project, but as long as we're using URIs and paths this way it makes sense to have something there.

mossprescott: Thanks for the nudge!

core/src/main/scala/quasar/backend.scala
wemrysi: That being said, its out of the scope of this issue. Do we currently validate all mounts on startup? If so then I think it is appropriate to validate this as well. If we still want to communicate via the rest api that a mount is configured but is broken, then I think this should be done in a more general way, and not in the `ViewBackend`. i.e. another type of backend that just holds broken configurations and returns errors when any of them are accessed. I think `ViewConfig`s should be correct on construction and the `ViewBackend` closed to the fact that a bad config may have existed in the config file.

mossprescott: That's a good idea and would nicely isolate the backend/interpreter from worrying about bad configs, while also allowing the API to represent them coherently and allow them to be reviewed and fixed by the user.

For the record, an unparsable URI, or even an unparsable query is treated as a config file format error, and prevents the server from starting at all. Both of which would be better handled in the way you're suggesting.

There is a ticket out there to deal better with bad mounts (specifically, connection failures): [SD-1131](https://slamdata.atlassian.net/browse/SD-1131). I suggest we come back to both of these issues after the dust settles on the free algebra impl.

For this PR, I'd be willing to constrain ViewConfig to (statically) valid views, but I would want to go ahead and implement "InvalidMount" so that we're not making the "dead server" problem worse. And that's going to involve more changes to the Backend stuff that need to be re-implemented on the free algebra side. Hmm. Any ideas on a middle path?

wemrysi: I sympathize with not wanting to increase the chance that the server doesn't start, but if you did validate them on startup it would at least be consistent. At the moment if the server starts, then you know everything works. The price for this assurance is, admittedly, steep, but at least you get something for your troubles. I'm not sure which is worse, increasing the chances that the server wont start, or having it start and there still be broken things lurking.

mossprescott: No contest. The server not starting is worse, because users can't figure out what's going on and they email support.

Now you can say that we should produce a nice error message in that case, and that the bundled app should somehow surface that message, and I won't argue with you, but I don't think any of that is in scope here.

mossprescott: And furthermore, I still think we want the server to start if at all possible, so it can serve the API and the front-end which provides the interface to fix the problem. Otherwise we're asking users to find a config file buried somewhere on disk and manually correct it, which is just not friendly.

wemrysi: Alright, well given that we're reimplementing this for the new architecture right now, might as well just leave it as-is. The discussion has been valuable as we have a decent idea of how to want to proceed for the new implementation.

web/build.sbt
```diff
@@ -2,10 +2,11 @@ name := "Web"
 
 mainClass in Compile := Some("quasar.api.Server")
 
+// NB: also defined in /build.sbt
```
wemrysi: That's unfortunate, should probably either ditch these sub-build files or move our version variables into an object that we can share across them.

build.sbt
```diff
@@ -97,6 +98,7 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
     "org.typelevel"     %% "shapeless-scalaz"          % slcVersion     % "compile, test",
     "com.slamdata"      %% "pathy-core"                % pathyVersion   % "compile",
     "com.github.mpilquist" %% "simulacrum"             % "0.4.0"        % "compile, test",
+    "org.http4s"        %% "http4s-dsl"                % http4sVersion  % "compile",
```
wemrysi: It didn't look like you used anything from `dsl` (correct me if I'm wrong), is `http4s-core` sufficient?

mossprescott:
@wemrysi: I believe I've addressed everything, unless you want me to figure out how to share that `http4sVersion` definition.

jonsterling:
Huzzah, it built!

wemrysi:
:+1: merge away

sellout:
Don’t we need more tests for this? It dropped coverage by half a percent.

mossprescott:
Thanks, @wemrysi.

@sellout: Actually, coverage is up by 0.05% since the preceding build on master, according to scoverage. If you can make sense of that codecov summary, please do explain it to me.

Squashing now.

sellout:
Cool 👍🏾

jonsterling:
thanks!!

PR #1018, opened 2015-12-01
Revisit DataServiceSpec tests and add checks for server state after certain operations

jedesah:
- Just generally try and tackle as many dangling TODOs as possible

### DIFF
--- a/web/src/test/scala/quasar/api/services/DataServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/DataServiceSpec.scala
@@ -2,6 +2,7 @@ package quasar
 package api
 package services
 
+import org.http4s.Uri.Authority
 import org.http4s.server.middleware.GZip
 import org.specs2.ScalaCheck
 import org.specs2.mutable.Specification
@@ -13,7 +14,7 @@ import JsonPrecision._
 import JsonFormat._
 import quasar.DataCodec
 
-import argonaut.{Argonaut, JsonObject, JsonNumber, Json}
+import argonaut.Json
 import argonaut.Argonaut._
 import org.http4s._
 import org.http4s.headers._
@@ -22,7 +23,9 @@ import quasar.fs.{Path => _, _}
 import pathy.Path
 import pathy.Path._
 import pathy.scalacheck.PathyArbitrary._
+import scalaz.scalacheck.ScalazArbitrary._
 import quasar.fs.NumericGen._
+import quasar.DataGen._
 
 import scalaz._, Scalaz._
 import scalaz.concurrent.Task
@@ -30,7 +33,7 @@ import scalaz.stream.Process
 
 import quasar.api.MessageFormatGen._
 
-import org.scalacheck.Arbitrary
+import org.scalacheck.{Arbitrary, Gen}
 
 import Fixture._
 
@@ -40,7 +43,12 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
   def service(mem: InMemState): HttpService =
     data.service[FileSystem](runStatefully(mem).run.compose(fileSystem))
 
-  implicit val arbJson: Arbitrary[Json] = Arbitrary(Arbitrary.arbitrary[String].map(jString(_)))
+  def serviceRef(mem: InMemState): (HttpService, Task[InMemState]) = {
+    val (inter, ref) = runInspect(mem).run
+    (data.service[FileSystem](inter compose fileSystem), ref)
+  }
+
+  implicit val arbFileName: Arbitrary[FileName] = Arbitrary(Gen.alphaStr.filter(_.nonEmpty).map(FileName(_)))
 
   val csv = MediaType.`text/csv`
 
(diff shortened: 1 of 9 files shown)


web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -395,7 +434,8 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
           to = destFile,
           state = emptyMem,
           status = Status.NotFound,
-          expectedBody = s"${printPath(file)}: doesn't exist")
+          expectedBody = s"${printPath(file)}: doesn't exist",
+          newState = Unchanged)
       }
       "be 404 if attempting to move a dir into a file" ! prop {(fs: NonEmptyDir, file: AbsFile[Sandboxed]) =>
```
wemrysi: I know this isn't a part of this PR, but thought I'd mention it since I happened to notice; I wouldn't expect a 404 in this situation, I think a 400 or maybe even a 409 would be more appropriate. If we're using a 404 to maintain parity with the existing implementation, then that's fine, maybe just add a comment to revisit.

wemrysi:
Have you been able to reproduce the compile issue locally? It'd be good to get that solved. Other than that and my minor comment, :+1:

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
jedesah: Actually, I am the one who added those tests, as the lack of a clear specification for such a situation became glaringly apparent thanks to the new strongly typed paths. And actually if you take a closer look, the test itself is fine, only the description text is incorrect. Fixed.

wemrysi: Ah, so they are, thanks.

jedesah:
@wemrysi It's weird because the Mongo 2.6 Travis build seems to consistently encounter a compiler crash (which is probably not due to `doc` as the `doc` command succeeds locally). This time around the failure for the build of Mongo 3.0 was probably caused by two unrelated tests having the same hardcoded port number. I fixed that so hopefully that should disappear.

Really unclear about what the compiler crash is about though, and furthermore perplexed as to why it only happens with the build for Mongo 2.6. What are the differences between the two builds, do they both test `core`? Because if so, that sounds like a bit of a waste but if one of them does not, then maybe sbt is getting into a weird state for some reason and that is causing the compiler crash.

wemrysi:
That build does a bit more than the other two just because it happens to be the first job. The first job builds the docs, does a coverage run and runs the jar test, whereas jobs `2, 3, ..., N` just run the tests. Yes, all the tests are run for all environments, but the non-it tests run so fast that it doesn't really matter and the build doesn't complete until all jobs have completed, so not running them on the other jobs isn't really going to speed up the overall build time.

jedesah:
hmm. Do we build the docs for the tests on Travis? because I just ran `test:doc` and that comes up with a few failures so that could explain the crash I guess.

jedesah:
But the things it's complaining about are also in `master`, so unlikely to be the culprit.

jedesah:
Well, at least seems to have fixed the issue with the 3.0 build.

@wemrysi Unless you have some ideas for possible solutions to the first build failure, I am inclined to merge and figure out what is up with that failure later.

wemrysi:
No, we don't run `test:doc` on travis. Have you tried running the same command that is run on travis locally to see if you can reproduce?

jedesah:
nope, let me try that (duh...)

jedesah:
@wemrysi Okay, ran the build script on my local machine and was able to reproduce. It appears that we are running into a `scoverage` issue. Compiling core without coverage works fine, but with coverage, it crashes almost immediately. It appears the file triggering the crash is `QueryFile.scala`.

Not sure what to do about this, I can submit a crash report on the `scoverage` project github but what do we do in quasar, disable coverage for that branch? I can also have a look at QueryFile.scala and see if I might not be able to refactor some things in hopes that that would magically aligns the bits in the right order and we would sidestep the crash...

jedesah:
Looking at the diff with master, no change in `QueryFile.scala` pops up as a possible culprit.

wemrysi:
From the error message, it looks like it involves line 196 in `QueryFile.scala`, which is in this function

```scala
private def firstFile(lp: Fix[LogicalPlan]): Option[AFile] =
  Tag.unwrap(lp foldMap[FirstOption[QPath]] {
    case Fix(LogicalPlan.ReadF(p)) => Tag(Some(p))
    case _                         => Tag(None)
  }) flatMap pathToAbsFile
```

jedesah:
@wemrysi Ya, it looks like it does not like the `TypeApply` so presumably it's not a huge fan of `scalaz.Tag`. If I change the implementation to this: `lp.collect{case Fix(LogicalPlan.ReadF(p)) => p}.headOption.flatMap(pathToAbsFile)` then it does not trigger the compilation crash. Even though it is fewer characters, I am not a huge fan of this alternative implementation because it builds up the whole list of paths as opposed to simply picking the first one, but I guess it's a reasonable compromise in order for coverage to work. I will comment out the current implementation and write a comment explaining why it can't be used right now and create an issue on `scoverage` explaining the issue.

wemrysi:
Out of curiousity, did you try specifying the types in `Tag.unwrap` and `Tag` explicitly, rather than having them inferred? Or maybe even ascribing `Option[QPath]` as the type of `Tag.unwrap(...)`?

jedesah:
Just tried all those things, and I still get the crash.

wemrysi:
Boo, thanks for trying. If this build succeeds, then :+1:

jedesah:
Well now it's that other server restart failure for which we have a sub-task. Weird that it seems to consistently trigger on the 2.6 branch but not on the other. You said that we were running tests for the `web` project on both builds right? Anyway, I say we merge this for now and revisit that later.

wemrysi:
:+1:

jedesah:
Oops, meant to flatten out those commits first. Oh well. I guess we will have to flatten out the free branch before merging in master anyway, so not a huge deal.

PR #1019, opened 2015-12-01
Implement find() and count().

sellout:


core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -278,6 +300,118 @@ trait MongoDbEvaluatorImpl[F[_], C] {
...
+      case List($Match((), sel), $Skip((), skip), $Limit((), limit)) =>
+        executor.find(source, sel.some, None, None, skip.toInt.some, limit.toInt.some)
+      case List($Match((), sel), Projectable(bson)) =>
+        executor.find(source, sel.some, bson.some, None, None, None)
+      case List($Skip((), count), Projectable(bson)) =>
+        executor.find(source, None, bson.some, None, count.toInt.some, None)
+      case List($Limit((), count), Projectable(bson)) =>
+        executor.find(source, None, bson.some, None, None, count.toInt.some)
+      case List($Match((), sel), $Skip((), count), Projectable(bson)) =>
+        executor.find(source, sel.some, bson.some, None, count.toInt.some, None)
```
mossprescott: I think you could collapse all the cases for combinations of Skip/Limit by first peeling off those ops (with a four-way pattern match) and then matching on `(pipeline, skipOpt, limitOpt)`.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -278,6 +300,118 @@ trait MongoDbEvaluatorImpl[F[_], C] {
...
+        executor.wrappedCount(field, source, None, count.toInt.some, None)
+      case List($Limit((), count), Countable(field)) =>
+        executor.wrappedCount(field, source, None, None, count.toInt.some)
+      case List($Match((), sel), $Skip((), count), Countable(field)) =>
+        executor.wrappedCount(field, source, sel.some, count.toInt.some, None)
+      case List($Match((), sel), $Limit((), count), Countable(field)) =>
+        executor.wrappedCount(field, source, sel.some, None, count.toInt.some)
+      case List($Skip((), skip), $Limit((), limit), Countable(field)) =>
+        executor.wrappedCount(field, source, None, skip.toInt.some, limit.toInt.some)
+      case List($Match((), sel), $Skip((), skip), $Limit((), limit), Countable(field)) =>
```
mossprescott: Probably apply the same approach to Countable: first strip off a trailing Countable op, then match `(pipeline, skipOpt, limitOpt, countFieldOpt)` That leaves you with only the rest of the pipeline to match.

core/src/test/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -67,20 +68,53 @@ class EvaluatorSpec extends Specification with DisjunctionMatchers {
...
+        $read(Collection("db", "zips")),
+        $match(Selector.Doc(
+          BsonField.Name("pop") -> Selector.Gte(Bson.Int64(1000)))),
+        $group(
+          Grouped(ListMap(
+            BsonField.Name("num") -> $sum($literal(Bson.Int64(1))))),
+          $literal(Bson.Null).right))
+
+      MongoDbEvaluator.toJS(crystallize(wf)) must beRightDisjunction(
+        """{ "num": db.zips.count({ "pop": { "$gte": NumberLong(1000) } }) };
```
mossprescott: I avoided this syntax (which doesn't actually mean what you think it does in JS) and wrapped the value in `[..]`, which works and sort of makes sense if the result is meant to be analogous to a cursor.

mossprescott: Correction: it may or may not be correct JS, but it doesn't work in the mongo shell:

```
> { a: 1 }
1
> foo = { a: 1 }
{ "a" : 1 }
> [ { a: 1 } ]
[ { "a" : 1 } ]
```

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -48,7 +50,27 @@ trait Executor[F[_], C] {
   def exists(coll: Collection): F[Unit]
 
   def pureCursor(values: List[Bson]): F[C]
-  def readCursor(source: Col): F[C]
+  def count(
```
mossprescott: As far as I can tell, this is actually unused.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -48,7 +50,27 @@ trait Executor[F[_], C] {
...
+    query: Option[Selector],
+    skip: Option[Int],
+    limit: Option[Int]):
+      F[C]
+  def wrappedCount(
+    field: BsonField.Name,
+    source: Col,
+    query: Option[Selector],
+    skip: Option[Int],
+    limit: Option[Int]):
```
mossprescott: It seems like we should wrap up these parameters in a `Count` case class, alongside Pipeline and MapReduce. Same goes for `Find`.

I think it would be less confusing to call the Selector here `sel` or `filter`.

And all of these could probably use documentation somewhere about when the various decorations are applied. For example, MapReduce sorts before mapping, while Find sorts after, but before skip/limit, etc.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: So, skip/limit don’t always happen at the end of the pipeline, but maybe with their re-orderability, it’s still ok to separate them from the rest of the pipeline?

core/src/test/scala/quasar/physical/mongodb/evaluator.scala
sellout: Oh … interesting. What the hell?!

But yeah – it should be wrapped in an array anyway. That makes sense.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: It is currently, and I can get rid of it for the time being.

We will eventually use it for things like `select value count(*) …` and optimizing mapReduce (both for `jsMode` and for θJoins), and once we have a BlackShield evaluator, the wrapping will be handled there and `count` will be the only version provided by the evaluator.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: I agree – we actually do have a `FindQuery` class already, which oddly is where `Selector` is defined.

`query` is the name MongoDB uses for this parameter, so I tried to keep their terminology. I agree it’s a bit confusing. `filter` sounds good.

I tried to order the parameters in the order the operations are applied, but I don’t think the MapReduce class does that.

sellout:
Also, the terrible `evalPipeline` should be handled in `Workflow.task`, but we don’t currently know at that point whether we want a collection or a cursor, so while the logic should basically be the same, it’s later in the process. This also means that `$where` selectors, field exclusions, etc. still end up in `mapReduce` rather than find/count.

core/src/test/scala/quasar/physical/mongodb/evaluator.scala
sellout: 👍🏾

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -278,6 +281,72 @@ trait MongoDbEvaluatorImpl[F[_], C] {
...
+        proj.rhs.some
+      case _ => None
+    }
+  }
+
+  private def extractRange(pipeline: workflowtask.Pipeline):
+      (workflowtask.Pipeline, (Option[Long], Option[Long])) =
+    pipeline match {
+      case Nil                                => (Nil, (None,   None))
+      case $Skip((), s) :: $Limit((), l) :: t => (t,   (s.some, l.some))
```
mossprescott: This obviously works now, because SQL only supports skip/limit at the (logical) end of the pipeline, but in general you can write LogicalPlans where this will go wrong by moving skip/limit after match, for example. I think you need to look for these only at the end (since that's the semantics of `find`).

What was the other op that can follow skip/limit?

mossprescott: By the way, I *think* you could break this with a sub-select, e.g. `select * from (select city from zips order by city asc skip 10 limit 10) order by city desc) t`

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -410,18 +480,49 @@ class MongoDbExecutor[S](client: MongoClient,
   def pureCursor(values: List[Bson]) =
     liftTask(Task.now(Process.emitAll(values.map(BsonCodec.toData))))
 
-  def readCursor(source: Col) =
-    fromCursor(Task.delay { mongoCol(source.collection).find() }, source)
+  def count0(
+    source: Col, count: Count) =
+    Process.emit(
+      mongoCol(source.collection).count(
+        count.query.fold(???)(_.bson.repr),
```
mossprescott: Unfinished?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -410,18 +480,49 @@ class MongoDbExecutor[S](client: MongoClient,
...
+            find.sort.foldLeft(
+              find.projection.foldLeft(
+                find.query.foldLeft(
+                  mongoCol(source.collection).find())(
+                  (c, sel) => c.filter(sel.bson.repr)))(
+                (c, proj) => c.projection(proj.repr)))(
+              (c, sort) => c.sort($Sort.keyBson(sort).repr)))(
+            (c, count) => c.skip(count.toInt)))(
+          (c, count) => c.limit(count.toInt))),
+      source)
```
mossprescott: Cute. Might want to take a look at what I did in `mapReduceCursor` a few lines lower down and pick one way or the other.

core/src/main/scala/quasar/physical/mongodb/findquery.scala
```diff
@@ -19,274 +19,52 @@ package quasar.physical.mongodb
...
+      comment.map(            "$comment"     -> Bson.Text(_)),
+      explain.map(exp      => "$explain"     -> Bson.Int32(if (exp) 1 else 0)),
+      hint.map(               "$hint"        -> _),
+      maxScan.map(            "$maxScan"     -> Bson.Int64(_)),
+      max.map(max          => "$max"         -> Bson.Doc(max.map(_.leftMap(_.asText)))),
+      min.map(min          => "$min"         -> Bson.Doc(min.map(_.leftMap(_.asText)))),
+      orderby.map(ts       => "$orderby"     -> Bson.Doc(ts.map(_.bimap(_.asText, _.bson)).list.toListMap)),
+      returnKey.map(retK   => "$returnKey"   -> Bson.Int32(if (retK) 1 else 0)),
+      showDiskLoc.map(show => "$showDiskLoc" -> Bson.Int32(if (show) 1 else 0)),
+      snapshot.map(snap    => "$snapshot"    -> Bson.Int32(if (snap) 1 else 0))
```
mossprescott: How about adding a function for this repeated code?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: Yeah, I think you’re right, but $Project (if there’s no $Sort) also comes after skip/limit. I could maintain the preceding and following parts of the pipeline separately and match against both. That doesn’t seem any more complicated.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: Uhh … *cough* yeah. Whoops. That needs to be an empty doc.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: Yeah, I think I like yours better.

core/src/main/scala/quasar/physical/mongodb/findquery.scala
sellout: 👍🏾

core/src/main/scala/quasar/physical/mongodb/findquery.scala
```diff
@@ -19,274 +19,52 @@ package quasar.physical.mongodb
...
+  comment:     Option[String],
+  explain:     Option[Boolean],
+  hint:        Option[Bson],
+  maxScan:     Option[Long],
+  max:         Option[ListMap[BsonField, Bson]],
+  min:         Option[ListMap[BsonField, Bson]],
+  orderby:     Option[NonEmptyList[(BsonField, SortType)]],
+  returnKey:   Option[Boolean],
+  showDiskLoc: Option[Boolean],
+  snapshot:    Option[Boolean]) {
```
mossprescott: It's unclear to me how these parameters relate to Find. As far as I can tell, these are all modifiers that get applied to the *cursor*, not really part of $find per se.

What is the relationship between `orderBy` here and `Find.sort` below? 

Why single out `skip` and `limit` in `Find`, if they also appear in `Count`? Do they apply to `Count`?

core/src/main/scala/quasar/recursionschemes/FunctorT.scala
```diff
@@ -41,20 +41,30 @@ import simulacrum.{typeclass, op}
...
   */
 @typeclass trait FunctorT[T[_[_]]] {
+  /** very roughly like Uniplate’s `descend` */
   @op("∘") def map[F[_], G[_]](t: T[F])(f: F[T[F]] => G[T[G]]): T[G]
 
+  /** akin to Uniplate’s `transform` */
   def transCataT[F[_]: Functor](t: T[F])(f: T[F] => T[F]): T[F] =
     f(map(t)(_.map(transCataT(_)(f))))
 
+  /** akin to Uniplate’s `topDownTransform` */
```
mossprescott: I like it 👍🏾

core/src/main/scala/quasar/physical/mongodb/findquery.scala
sellout: FindQuery is a model of this structure: https://docs.mongodb.org/manual/reference/operator/query-modifier/

core/src/test/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -99,7 +157,8 @@ class EvaluatorSpec extends Specification with DisjunctionMatchers {
           |          { "pop": { "$gte": NumberLong(100) } }]
           |      }
           |    },
-          |    { "$sort": { "city": NumberInt(1) } }],
+          |    { "$group": { "pop": { "$sum": "$pop" }, "_id": "$city" } },
+          |    { "$sort": { "_id": NumberInt(1) } }],
```
mossprescott: I recognize and appreciate the hard work you've done here to update these tests, but do they assure us that these plans are correct?

The more variations we have on this, including streaming and not, the more I think it would be better to actually execute these commands. If we could do that through the Java driver, that might be worth doing sometime soon, but I suspect we would need to launch a `mongo` shell process, which seems like a pain. But there would be no need at that point to write these tests manually. What do you think?

sellout: I think these are distinct from tests that execute the plan. We want to ensure that certain cases use the efficient operations, which we don’t learn by checking the results.

I think it would be great if the integration tests ran both evaluators.

mossprescott:
Looks pretty good, just some confusing bits now that we're exposing more of the complexity of cursors, a couple of questions, and one `???` left behind.

With regard to your comment on `where` selectors in `find`, my reading of SD-954 is that it does call for that, so is there still work to do to push some of this into the planner?

core/src/test/scala/quasar/physical/mongodb/evaluator.scala
sellout: Oh, but also there are definitely cases where the Mongo and Js evaluators don’t currently line up. E.g., `exists` does nothing on the JsEvaluator.

sellout:
Ok, this should be ready.

It still doesn’t handle things like `$where` – we need a new ticket for exposing `dest: Option[Collection]` to the planner, so we can use it in `Workflow.task`, as well as restructuring WorkflowTask to distinguish between ones that can write to a temp collection and ones that can’t.

Also, the separation of Pipeline head/tail isn’t ideal, but it should at least be correct.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -278,6 +281,79 @@ trait MongoDbEvaluatorImpl[F[_], C] {
...
+      case (Nil, List(Projectable(bson))) =>
+        executor.find(source, Find(None, bson.some, None, skip, limit))
+      case (List($Sort((), keys)), Nil) =>
+        executor.find(source, Find(None, None, keys.some, skip, limit))
+      case (List($Match((), sel)), List(Projectable(bson))) =>
+        executor.find(source, Find(FindQuery.selector(sel).some, bson.some, None, skip, limit))
+      case (List($Match((), sel), $Sort((), keys)), Nil) =>
+        executor.find(source, Find(FindQuery.selector(sel).some, None, keys.some, skip, limit))
+      case (List(Projectable(bson), $Sort((), keys)), Nil) =>
+        executor.find(source, Find(None, bson.some, keys.some, skip, limit))
```
mossprescott: Are you sure this is safe? The [docs](https://docs.mongodb.org/manual/reference/method/cursor.sort/#cursor.sort) say sorting is always done before projection. Maybe this is OK because `Projectable` only matches `$include`, so you can't rename the fields in a way that would break the sort?

sellout: Yeah, exactly. Although also handling the reverse might cover more cases – IE, where we don’t project the sort field.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -533,8 +634,29 @@ class JSExecutor[F[_]: Monad](nameGen: NameGenerator[F])
...
     write(Bson.Arr(values).toJs)
-  def readCursor(source: Col) =
-    write(Call(Select(toJsRef(source.collection), "find"), Nil))
+
+  def count0(source: Col, count: Count) =
+    count.limit.foldLeft(
+      count.skip.foldLeft(
+        Call(Select(toJsRef(source.collection), "count"), count.query.toList.map(_.bson.toJs)))(
+        (c, count) => Call(Select(c, "skip"), List(Num(count, false)))))(
+      (c, count) => Call(Select(c, "limit"), List(Num(count, false))))
```
mossprescott: Missed opportunity to use CursorOp.app?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -533,8 +634,29 @@ class JSExecutor[F[_]: Monad](nameGen: NameGenerator[F])
...
+    write(
+      find.limit.foldLeft(
+        find.skip.foldLeft(
+          find.sort.foldLeft(Call(Select(toJsRef(source.collection), "find"),
+            find.query.foldRight(
+              find.projection.foldRight[List[Js.Expr]](Nil)(_.toJs :: _))(
+              _.bson.toJs :: _)))(
+            (c, keys) => Call(Select(c, "sort"), List($Sort.keyBson(keys).toJs))))(
+          (c, count) => Call(Select(c, "skip"), List(Num(count, false)))))(
+        (c, count) => Call(Select(c, "limit"), List(Num(count, false)))))
```
mossprescott: Missed opportunity to use CursorOp.app?

core/src/main/scala/quasar/physical/mongodb/findquery.scala
mossprescott: Interesting. I'm not sure all of these are exposed the Java driver. I would still prefer to remove anything that we're not actually using, which is most of this, isn't it?

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: Indeed! I had the “cursor” part of it stuck in my head, forgetting that it’s just `A`.

core/src/main/scala/quasar/physical/mongodb/findquery.scala
mossprescott: Arg. Just noticed it's all translated by `bson`. So it's implemented, but not tested. I guess I can live with it.

sellout: Yes – we currently actually use none of it. It used to be used by $GeoNear, which we also didn’t use, so it also wasn’t used then. I’m happy to get rid of it altogether, and just use `Selector` in its place.

mossprescott:
Other than minor comments, my only concern here is how do we know we have regression coverage for all the cases? Can you run `./sbt coverage it/test` (and *not* the JSExecutor test) and see how many get hit, maybe? I imagine at least a few of those cases are not in the suite.

core/src/main/scala/quasar/physical/mongodb/findquery.scala
mossprescott: :+1:

sellout:
Ok, regression tests now cover all cases.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -222,11 +241,11 @@ trait MongoDbEvaluatorImpl[F[_], C] {
         emit(executor.exists(value).as(Col.User(value): Col))
 
       case QueryTask(source, query, skip, limit) =>
-        // TODO: This is an approximation since we're ignoring all fields of "Query" except the selector.
+        // TODO: This is an approximation since we're ignoring all fields of "FindQuery" except the selector.
```
mossprescott: Can delete this TODO now, I think :)

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -410,25 +505,45 @@ class MongoDbExecutor[S](client: MongoClient,
   def pureCursor(values: List[Bson]) =
     liftTask(Task.now(Process.emitAll(values.map(BsonCodec.toData))))
 
-  def readCursor(source: Col) =
-    fromCursor(Task.delay { mongoCol(source.collection).find() }, source)
+  def count0(
+    source: Col, count: Count) =
+    Process.emit(
```
mossprescott: Weird line wrapping.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
```diff
@@ -410,25 +505,45 @@ class MongoDbExecutor[S](client: MongoClient,
...
+          .app(count.skip)((c, count) => c.skip(count.toInt))
+          .app(count.limit)((c, count) => c.limit(count.toInt))))
+
+  def wrappedCount(field: BsonField.Name, source: Col, count: Count) =
+    liftTask(
+      Task.delay(count0(source, count).map(n =>
+        Data.Obj(ListMap(field.asText -> Data.Int(n))))))
+
+  def find(
+    source: Col, find: Find) =
```
mossprescott: Seriously?

sellout: Oh, that’s just left over from when there was no `Find`, just a ton of parameters.

core/src/main/scala/quasar/physical/mongodb/evaluator.scala
sellout: Same here.

core/src/main/scala/quasar/physical/mongodb/workflowtask/ast.scala
```diff
@@ -65,22 +65,21 @@ object WorkflowTaskF {
...
-  //         case ReadTaskF(coll) => G.point(ReadTaskF[B](coll))
-  //         case QueryTaskF(src, query, skip, limit) =>
-  //           f(src).map(QueryTaskF(_, query, skip, limit))
-  //         case PipelineTaskF(src, pipe) => f(src).map(PipelineTaskF(_, pipe))
-  //         case MapReduceTaskF(src, mr, oa) => f(src).map(MapReduceTaskF(_, mr, oa))
-  //         case FoldLeftTaskF(h, t) =>
-  //           (f(h) |@| t.traverse(f))(FoldLeftTaskF(_, _))
-  //       }
-  //   }
+  implicit def WorkflowTaskFTraverse: Traverse[WorkflowTaskF] =
```
mossprescott: :smile:

sellout: Yeah – we use it now. Imagine that 😉

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -2318,6 +2318,8 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
             \/-($literal(Bson.Null)))))
     }
 
+    // TODO: We should be able to write this distinct without the initial
+    //       $project. All of the values can be projected from the id.
```
mossprescott: Is this just something you noticed along the way? Seems like this falls under the earlier story for projecting from group keys, if that wasn't already applied to `distinct`?

I guess I'm saying this TODO should be referencing an open issue.

it/src/main/resources/tests/caseWithGroup.test
```diff
@@ -17,7 +17,9 @@
                 group by state
                 order by funnel",
 
-    "predicate": "equalsInitial",
+    "note": "should be `equalsInitial`, with a separate `containsAtLeast` with
+             the IL entry (and perhaps others).",
```
mossprescott: Funny. This is the one I had to fix for 3.2 (which I did by adding `state` to the sort).

core/src/test/scala/quasar/physical/mongodb/planner.scala
sellout: I’m not going to change it – I fixed it and removed that comment in the `distinct()` PR coming up (basically, identifying streamable `distinct`s was too difficult without fixing that).

it/src/main/resources/tests/fieldSubset.test
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "some entries from a collection",
+
+    "data": "zips.data",
+
+    "query": "select city from zips limit 10",
+
+    "predicate": "containsExactly",
+    "expected": [
```
mossprescott: This requires that the backend maintain insertion order, which isn't something we want to do. Can you make the result well-defined while still covering the right case in the evaluator?

sellout: I don’t know that I can.

sellout: This specifically needs to do only projection and limit.

it/src/main/resources/tests/filterSkipLimitCount.test
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "filter, skip, limit, and count",
+
+    "data": "zips.data",
+
+    "query": "select count(*) from (select * from zips where city like 'BOU%' offset 15 limit 10) x",
```
mossprescott: Same issue here. Any `offset`/`limit` with no `order by` will be ill-defined).

it/src/main/resources/tests/fieldSubset.test
sellout: Well, and/or skip.

core/src/main/scala/quasar/physical/mongodb/workflowtask/ast.scala
wemrysi: Ha, #1027 also uncommented this ;)

it/src/main/resources/tests/filterSkipLimitCount.test
sellout: We could perhaps just use evaluator unit tests for these, and assume that integration coverage on the other cases is good enough for these as well.

mossprescott:
I'd like to see these tests run on 3.2 before we merge this, and I think we might need to come up with another predicate for those tests (say, "should contain at least *n* of these values"), because I don't think we can afford to pretend that backends preserve insertion order.

Also, a couple of stray TODOs.

Otherwise looks good!

core/src/main/scala/quasar/physical/mongodb/workflowtask/ast.scala
mossprescott: Awesome, because I felt seriously dirty commenting it but needed to get that build over the coverage bar in a hurry ;) Glad to see it come back and pulling its weight.

core/src/test/scala/quasar/physical/mongodb/planner.scala
mossprescott: Good enough for me.

core/src/main/scala/quasar/physical/mongodb/workflowtask/ast.scala
sellout: If it weren’t so difficult to derive Arbitrary at the moment, then we would have just run the traverse laws over it rather than commenting in the first place.

it/src/main/resources/tests/filterSkipLimitCount.test
mossprescott: Hmm. Actually I think my comment was misplaced on this one, which is actually well-defined.

sellout:
Ok, only more-deterministic result sets now.

mossprescott:
LGTM

PR #1021, opened 2015-12-03
Regression test improvements

mossprescott:
The first commit fixes a bug where the destination collection was not deleted if 1) the result is streamed, 2) the last stage is a map-reduce, and 3) the result isn't read all the way to the end. All of which happened to be true of our regression tests for joins, with the result that my test DB was gradually filling up with orphaned temporaries, making one of the *other* tests hang eventually.

The second commit just reduces the data size for the map-reducing regression tests, so that they run in no more than few seconds each (down from more than a minute in some cases). The new dataset, `largeZips`, contains only large individual zip codes, including at least one from each state, plus all the zips in Colorado, and every city that resembles `BOULDER` (because some of the tests expected to see those cities).

The result is the build runs about 5 minutes faster on Travis, and the regression tests specifically take less than a minute locally, down from several minutes.


mossprescott:
Odd that one of the three builds timed out last night, but Travis seemed to be having capacity issues at the time, so I restarted it this morning and it ran fine. Restarted the whole shebang this morning to get a second opinion.

sellout:
👍🏾

PR #1022, opened 2015-12-03
[SD-986] service extension

drostron:


drostron:
Ended up being unnecessary.

PR #1024, opened 2015-12-08
[SD-1177] Port over addition of offset and limit to query API

jedesah:
- Add show for Positive
- Implement response for SemanticError
- Add new test in CompileAndQuerySpec and refactor a little to get it in there nicely
- Add simple in-memory query evaluator

### DIFF
--- a/web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
@@ -3,6 +3,7 @@ package api
 package services
 
 import Predef._
+import quasar.fp._
 
 import argonaut._
 
@@ -18,12 +19,16 @@ import pathy.scalacheck.AbsFileOf
 import pathy.scalacheck.PathyArbitrary._
 import quasar.fs._
 import quasar.fs.InMemory._
+import quasar.recursionschemes._
 
 import Fixture._
+import NumericGen._
 import quasar.std.IdentityLib
 
 import scalaz.concurrent.Task
 import scalaz.stream.Process
+
+import scalaz._, Scalaz._
 
 import query._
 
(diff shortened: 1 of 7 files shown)


jedesah:
@wemrysi Assigned @mossprescott to this PR as he did the original work on master, but feel free to have a look. I am issuing this PR against the feature branch even though it contains some changes to the in-memory interpreter, but once this gets merged. I will issue a PR against master with all the additive changes to `core` that are contained in this feature branch.

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
```diff
@@ -39,36 +44,66 @@ class CompileAndQueryServiceSpec extends Specification with FileSystemFixture wi
...
+  def selectAll(from: File) = "select * from \"" + printPath(from) + "\""
+  def selectAllWithVar(from: File, varName: String) = selectAll(from) + " where pop < :" + varName
+
+  def toLP(q: String): Fix[LogicalPlan] = (for {
+    ast        <- (new sql.SQLParser()).parse(sql.Query(q)).toOption
+    annotated  <- SemanticAnalysis.AllPhases(ast).toOption
+    logical    <- Compiler.compile(annotated).toOption
+    simplified <- logical.transCata(repeatedly(Optimizer.simplifyƒ)).toOption
+    checked    <- LogicalPlan.ensureCorrectTypes(simplified).toOption
+  } yield checked).getOrElse(scala.sys.error("could not compile: " + q))
```
mossprescott: Why not use `quasar.queryPlan()`? This makes the third copy of this code.

jedesah: Oh, sure, let me look into using `queryPlan` to see if I can re-use that. I guess I just translated your original commit a little too faithfully :smile:

mossprescott:
I'm having a hard time following these tests. One of the things that was added in that PR was that the tests actually check the query that is submitted to the Backend, which had paths interpreted according to the request path, etc., as well as the rewriting of skip/limit into the query. Are these tests actually verifying the queries or not? AS far as I can tell, they're looking only at the response status and body.

mossprescott:
Wait, hold on, I think I see it now: you're recording the evaluated LP in the state. So that's OK I think.

jedesah:
So in the case of `GET`. offset and limit are being tested by looking at the output to see if the proper data was returned, given the property based nature of the tests, this should give us a pretty high degree of confidence that they are working.

As for POST, the situation is a little more complicated I guess because we don't get back the results. I mean, we could read the results from the output file I guess to have a higher degree of confidence in offset and limit.

For the variable, since we are using `queryPlan`, we are testing that the variable is being substituted properly, or else it would not find the pre-canned response in the state.

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
```diff
@@ -185,12 +249,23 @@ class CompileAndQueryServiceSpec extends Specification with FileSystemFixture wi
     "plan simple query" ! prop { filesystem: SingleFileMemState =>
       // Representation of the directory as a string without the leading slash
       val pathString = printPath(filesystem.file).drop(1)
-      get(compileService)(
+      get[String](compileService)(
+        path = filesystem.parent,
+        query = Some(Query(selectAll(file(filesystem.filename.value)))),
+        state = filesystem.state,
+        status = Status.Ok,
+        response = κ(ok)
```
mossprescott: The old test was verifying variable substitution here, for whatever that's worth. You should be able to do the same by inspecting the response, right?

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
```diff
@@ -185,12 +249,23 @@ class CompileAndQueryServiceSpec extends Specification with FileSystemFixture wi
...
+      val pathString = printPath(filesystem.file).drop(1)
+      val query = selectAllWithVar(file(filesystem.filename.value),varName.value)
+      get[String](compileService)(
         path = filesystem.parent,
-        query = Some(selectAll(filesystem.filename)),
+        query = Some(Query(query,varNameAndValue = Some((varName.value, var_.toString)))),
         state = filesystem.state,
         status = Status.Ok,
-        response = "Lookup in Memory\nLookup Squash(Read(Path(\"" + pathString + "\"))) in Map()"
+        response = κ(ok)
```
mossprescott: Similar here.

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
jedesah: I could. After trying to do that, I decided not too because it would have involved copying a large piece of hardcoded text and it felt like it was filesystem specific. Do you think it's fine if it's backend specific?

mossprescott: I don't understand. This is running only against the InMemory filesystem, right?

jedesah: Yes, but ideally the web api specifications should not be tied to a particular filesystem. We use the in-memory filesystem because it is fast and pure, but we should strive to specify the behavior of the web api independant of a particular filesystem.

mossprescott:
On POST: what you want to test is always just that the right query was evaluated; in the old tests, I had to do all kinds of tricks to verify that by inspecting the response because I wanted to avoid state in the (stubbed) backend. In this case, you could inspect the log that comes back long with the output path.

But it might be easier now to just record the query when it's evaluated and inspect it directly, or at least that was my hope. :)

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
jedesah: And I think this is true of the vast majority of tests, however in this case, it may be perfectly reasonable to say that the `compile` route needs to explain the compilation phases which is hard to do without hardcoding the test to a particular filesystem and assuming it will work with others.

mossprescott: Ah, I see what you're saying. Yes, ideally we would specify the behavior only up to the point that the LP is handed over to the filesystem, and not actually have to run it. Until then, whatever is least painful is what we want to do I think.

mossprescott:
I think this covers the new functionality, so I say LGTM and let's plan to work more on improving this stuff after we cut over to it.

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
jedesah: You're right. In a perfect world we might be able to write the api as `Free` and inspect that the right operations appear in the right order but currently we are forced to translate `Free` into `Task` pretty early and are stuck with the InMemory interpreter in order to test the api. Not sure where this leaves us for this particular test, let's just stick the huge string as the expected result?

jdegoes: The long-term goal, obviously, should be to *not* translate `Free` into `Task`, but translate the composite algebra into tasks, so that interpreters can be composed. Blog post coming soon. :)

wemrysi: @jdegoes What do you mean by "tasks" here?

PR #1026, opened 2015-12-09
add MongoDB 3.2 to the test matrix

mossprescott:
SD-1141 #done

Note: A couple of queries are not matching, so this isn't ready to go yet.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -46,12 +46,12 @@ if [[ ${LOCAL_MONGODB:-} == "true" ]] ; then
   if [[ "$QUASAR_MONGODB_RELEASE" == "2_6" ]]; then
     export QUASAR_MONGODB_2_6="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_2_6}\"}}"
   elif [[ "$QUASAR_MONGODB_RELEASE" == "3_0" ]]; then
-    export QUASAR_MONGODB_3_0="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_0}\"}}"
+    export QUASAR_MONGODB_3_0="{\"mongodb\": {\"connectionUri\": \"mongodb://quasar-dbOwner:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\"}}"
   elif [[ "$QUASAR_MONGODB_RELEASE" == "3_0_RO" ]]; then
     export QUASAR_MONGODB_READ_ONLY="{\"mongodb\": {\"connectionUri\": \"mongodb://quasar-read:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\"}}"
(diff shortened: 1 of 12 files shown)


mossprescott:
One test was failing just because the query was unsorted but the test expected a particular result, so that's fixed. Now there's one more failure that I haven't been able to figure out.

sellout:
I see a few tests pending – are they related to field ordering? SD-1174 proposes being able to mark a backend as “unordered” (like “pending”) if it doesn’t respect field ordering.

mossprescott:
I don't remember what's keeping each of those tests in pending (they've mostly been around for quite a while), but my sense is that it's mostly unresolved Mongo weirdness, like projecting `_id` and the behavior around missing values.

sellout:
Ah, right – those are pending everywhere, not just 3.2

mossprescott:
Why don't we just fix the field ordering, fer crissakes? We just need to insert a final (pair of) project(s), similar to what we do to eliminate _id. Er, unless we want to stream from a map-reduce, I guess, but in that case we should do it in memory.

jdegoes:
> Why don't we just fix the field ordering, fer crissakes? 

Yeah, let's just bite the bullet. It would be ideal if we can detect some cases when this is not needed, just to avoid the performance penalty, but that's not strictly required.

mossprescott:
@jedesah, you got this one just because I don't think I've given you anything recently.

scripts/build
```diff
@@ -45,15 +46,18 @@ if [[ ${LOCAL_MONGODB:-} == "true" ]] ; then
   if [[ "$QUASAR_MONGODB_RELEASE" == "2_6" ]]; then
     export QUASAR_MONGODB_2_6="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_2_6}\"}}"
   elif [[ "$QUASAR_MONGODB_RELEASE" == "3_0" ]]; then
-    export QUASAR_MONGODB_3_0="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_0}\"}}"
+    export QUASAR_MONGODB_3_0="{\"mongodb\": {\"connectionUri\": \"mongodb://quasar-dbOwner:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\"}}"
```
jedesah: What's this `quasar-dbOwner`?

jedesah:
I know this was not added by this PR, but what is Mongo 3.0 RO?

jedesah:
I am under the impression this PR is "in progress" as opposed to "in review", but @mossprescott let me know if my assumption is wrong.

wemrysi:
Mongo 3.0 RO is a read-only connection to a mongodb 3.0.x instance used to test that we can execute a subset of queries using the streaming functionality on a read-only connection.

jedesah:
Ah, duh, should have figured that one out on my own.

mossprescott:
@jedesah: this was meant to be ready, but now it's failing in Travis and it looks like I need to do a series of test builds, so I'll close this and open a new PR when I have it working.

PR #1027, opened 2015-12-10
[SD-1155] Streaming query responses for 'Free' architecture

wemrysi:
Adds the `EvaluatePlan` term to the `QueryFile` algebra along with terms describing a cursor of query results, providing a streaming interface to evaluating `LogicalPlan`.

I've also refactored the new architecture's query regression tests to run twice, once for providing query results via a file path and another via streaming. I think this is useful from the point of view of the abstraction as these are both primitive operations.


jdegoes:
Due to performance reasons, eventually we will want this to be the moral equivalent of, `ContentType => Array[Byte]` (not really `Array`, but whatever the safe / just-as-fast version of that is).

The `Vector[Data]` is a nasty source of performance degradation. Wherever possible, we want to go direct from whatever a driver provides, to whatever the client requests, without having to perform all the intermediate boxing and translation that we do now.

The other issue is that, this is a perfectly fine, if imperative, `Free` encoding for streaming data. Albeit, you might consider something like `Cofree[Free[F, ?], Vector[Data]]`, which models a stream of data produced by snippets of sequential computation in `F` (maybe not in this PR!).

wemrysi:
@jdegoes Good point regarding performance, I like the idea of `ContentType => Bytes` or something similar. If we can schedule that as an improvement on its own, I think that would be best as we also need to deal with the `ReadFile` algebra and probably the inverse for the `WriteFile` algebra.

Your comment on `Cofree` is interesting, though I'm not sure I fully understand what you meant. It seems like an alternative to our use of `Process`. Would the algebra itself need to change in order to use it? And how does one deal with early termination by the client with `Cofree`?

jdegoes:
>  If we can schedule that as an improvement on its own, I think that would be best as we also need to deal with the ReadFile algebra and probably the inverse for the WriteFile algebra.

:+1: 

> Your comment on Cofree is interesting, though I'm not sure I fully understand what you meant.

I'm still working it out, but embedding what is essentially a coinductive process into `Free` will always result in an imperative API (open, next chunk, close, etc.). Ignore my comment for now and I'll get back to you if I figure out the correct encoding.

wemrysi:
> embedding what is essentially a coinductive process into Free will always result in an imperative API (open, next chunk, close, etc.)

@jdegoes That has been my conclusion as well, but definitely interested in any discoveries you make!

core/src/main/scala/quasar/physical/mongodb/WorkflowExecutor.scala
```diff
@@ -170,24 +246,45 @@ trait WorkflowExecutor[F[_]] {
 object WorkflowExecutor {
   import Workflow.Crystallized
 
+  /** A cursor to results from evaluating a `Workflow`, optionally paired with
```
mossprescott: Should be "a cursor *that*"?

Hey, I'm easing into it ;)

core/src/main/scala/quasar/physical/mongodb/fs/bsoncursor.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import org.bson.Document
+import scalaz._, Id._
+import scalaz.concurrent.Task
+import scalaz.syntax.monad._
+import scalaz.syntax.std.option._
+import scalaz.std.vector._
+
+object bsoncursor {
+  implicit val bsonCursorDataCursor: DataCursor[MongoDbIO, BsonCursor] =
```
mossprescott: I think the names are a bit confusing here. First of all, `BsonCursor` is actually a cursor of `Document`, and secondly I find it mostly obfuscatory to define a synonym for the mongo driver's type, although I suppose `AsyncBatchCursor[Document]` would be a mouthful seeing how much it shows up.

But it you're going to have the type alias, why is the object not called `BsonCursor`? Does it need to not be the companion for some reason?

it/src/test/scala/quasar/ResultFileQueryRegressionSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+      execToCompExec compose[M] Hoist[FileSystemErrT].hoist[F, G](liftMT[F, PhaseResultT])
+
+    for {
+      tmpFile <- toCompExec(manage.tempFileNear(tmpPath)).liftM[Process]
+      result  <- query.executeQuery(expr, vars, tmpFile).liftM[Process]
+      cleanup =  hoistM(manage.delete(tmpFile))
+      data    <- read.scanAll(resultFile get result)
+                   .translate(hoistM)
+                   .onComplete(Process.eval_(cleanup))
+    } yield data
```
mossprescott: Super glad to see this! We talked about wanting to run them both ways on the old implementation but punted.

wemrysi:
@mossprescott I finished rebasing onto master, so should be ready for a final review.

core/src/main/scala/quasar/physical/mongodb/fs/bsoncursor.scala
wemrysi: `BsonCursor` is now an alias for `AsyncBatchCursor[BsonDocument]`, is that any better? As for the object, as it only contains typeclass instances (or that is the intent) I used the lower-cased name following the defacto convention from other libraries (scalaz, cats, monocle) for objects that contain instances for externally defined types.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -62,59 +80,42 @@ object QueryFile {
       * Execution of certain plans may return a result file other than the
       * requested file if it is more efficient to do so (i.e. to avoid copying
       * lots of data for a plan consisting of a single `ReadF(...)`).
+      *
+      * TODO: It looks like we probably no longer need the `ResultFile` type
+      *       as the result will either be a file from the plan or `out`, but never
+      *       a temp file.
```
mossprescott: It looks like the same is true for the old Col/ResultPath type, and if so I'd love to see it deleted, at least in the new implementation. It should be easy to verify: just eliminate the `Temp` constructor and see if any test fails.

mossprescott:
Just a few comments on names and style. I don't see anything concerning. I'm taking your word that you got all the tests ported over from @sellout's PR(s), because I'm too fried to do check them at the moment.

Sorry about the comments ending up on the commits.

Looks great, can't wait to rebase on this.

wemrysi:
@mossprescott thanks, added a few more commits to address the rest of your comments. If things look good, mark it "ready" and I'll squash and merge.

mossprescott:
:+1:

PR #1028, opened 2015-12-11
Optimize apparent cross joins

mossprescott:
Includes a handful of small fixes, and then the main event:

Rewrite InnerJoin/Filter sequences, putting conditions before, in, or after the join.

The idea is to identify all filtering associated with an inner join in the source plan, breaking down all components into four categories:
- conditions referring to only the left or only the right source, which can be moved ahead of the join (SD-910)
- conditions referring to both sources using `=`, which can be rewritten into the join (SD-1063)
- all others, which have to be done after the join

This is implemented in a new rewriteCrossJoins fold, which uses boundParaM (which is new). Because the new fold introduces new temp names which made it difficult for tests to compare plans, names are now rewritten to a compact sequence starting with `__tmp0` after rewriting cross joins. Nested lets are re-ordered to a consistent sequence. Also, the plan has to be simplified prior to the new fold, so all these steps are now wrapped up in `Optimizer.optimize`, which makes sure to run things in the proper sequence. This replaces the fold with `simplifyƒ` that was being called from several places before.

Added test for two queries which "worked" before but are now massively more efficient. Added an explicit time limit to the regression tests so these will fail definitively if the optimization stops working at some point. These tests (and all the rest) run in a few seconds locally, but the limit is quite high for now (45s) because some take as long as 25s in Travis.

### DIFF
--- a/core/src/main/scala/quasar/logicalplan.scala
+++ b/core/src/main/scala/quasar/logicalplan.scala
@@ -184,13 +184,13 @@ object LogicalPlan {
 
       def initial[A] = Map[Symbol, A]()
 
-      def bindings[A](t: LogicalPlan[Fix[LogicalPlan]], b: G[A])(f: LogicalPlan[Fix[LogicalPlan]] => A): G[A] =
+      def bindings[T[_[_]]: Recursive, A](t: LogicalPlan[T[LogicalPlan]], b: G[A])(f: LogicalPlan[T[LogicalPlan]] => A): G[A] =
         t match {
-          case LetF(ident, form, _) => b + (ident -> f(form.unFix))
+          case LetF(ident, form, _) => b + (ident -> f(form.project))
           case _                    => b
         }
 
-      def subst[A](t: LogicalPlan[Fix[LogicalPlan]], b: G[A]): Option[A] =
+      def subst[T[_[_]]: Recursive, A](t: LogicalPlan[T[LogicalPlan]], b: G[A]): Option[A] =
         t match {
           case FreeF(symbol) => b.get(symbol)
           case _             => None
--- a/core/src/main/scala/quasar/logicalplan.scala
+++ b/core/src/main/scala/quasar/logicalplan.scala
@@ -17,7 +17,7 @@
 package quasar
 
 import quasar.Predef._
-import quasar.recursionschemes._, cofree._, Recursive.ops._
+import quasar.recursionschemes._, cofree._, Recursive.ops._, FunctorT.ops._
 import quasar.fp._
 import quasar.fs.Path
 import quasar.namegen._
@@ -180,6 +180,7 @@ object LogicalPlan {
 
   implicit val LogicalPlanBinder: Binder[LogicalPlan] = new Binder[LogicalPlan] {
       type G[A] = Map[Symbol, A]
+      val G = implicitly[Traverse[G]]
 
       def initial[A] = Map[Symbol, A]()
 
@@ -198,6 +199,40 @@ object LogicalPlan {
 
   def freshName(prefix: String): State[NameGen, Symbol] =
     quasar.namegen.freshName(prefix).map(Symbol(_))
+
+  // NB: this can't currently be generalized to Binder, because the key type isn't exposed there.
+  def renameƒ[M[_]: Monad](f: Symbol => M[Symbol]):
+      ((Map[Symbol, Symbol], Fix[LogicalPlan])) =>
+        M[LogicalPlan[(Map[Symbol, Symbol], Fix[LogicalPlan])]] =
+  { case (bound, t) =>
+    t.unFix match {
+      case LetF(sym, expr, body) =>
+        f(sym).map(sym1 =>
+          LetF(sym1,
+            (bound, expr),
+            (bound + (sym -> sym1), body)))
+      case FreeF(sym) =>
+        val v: LogicalPlan[(Map[Symbol, Symbol], Fix[LogicalPlan])] =
+          FreeF(bound.get(sym).getOrElse(sym))
+        v.point[M]
+      case t =>
+        t.strengthL(bound).point[M]
+    }
+  }
+
+  def rename[M[_]: Monad](f: Symbol => M[Symbol])(t: Fix[LogicalPlan]): M[Fix[LogicalPlan]] =
+    Corecursive[Fix].anaM[LogicalPlan, M, (Map[Symbol, Symbol], Fix[LogicalPlan])](
+      (Map(), t))(renameƒ(f))
+
+  def normalizeTempNames(t: Fix[LogicalPlan]) =
+    rename[State[NameGen, ?]](κ(freshName("tmp")))(t).evalZero
+
+  val normalizeLetsƒ: LogicalPlan[Fix[LogicalPlan]] => Option[LogicalPlan[Fix[LogicalPlan]]] = {
+      case LetF(b, Fix(LetF(a, x1, x2)), x3) => LetF(a, x1, Let(b, x2, x3)).some
+      case t => None
+  }
+
+  def normalizeLets(t: Fix[LogicalPlan]) = t.transAna(repeatedly(normalizeLetsƒ))
 
   type Typed[F[_]] = Cofree[F, Type]
   final case class NamedConstraint(name: Symbol, inferred: Type, term: Fix[LogicalPlan])
--- a/core/src/main/scala/quasar/logicalplan.scala
+++ b/core/src/main/scala/quasar/logicalplan.scala
@@ -88,11 +88,20 @@ object LogicalPlan {
           val nodeType = "LogicalPlan" :: Nil
 
           def render(v: LogicalPlan[α]) = v match {
-            case ReadF(name)             => Terminal("Read" :: nodeType, Some(name.pathname))
-            case ConstantF(data)         => Terminal("Constant" :: nodeType, Some(data.toString))
-            case InvokeF(func, args)     => NonTerminal("Invoke" :: nodeType, Some(func.name), args.map(ra.render))
-            case FreeF(name)             => Terminal("Free" :: nodeType, Some(name.toString))
-            case LetF(ident, form, body) => NonTerminal("Let" :: nodeType, Some(ident.toString), List(ra.render(form), ra.render(body)))
+            // NB: a couple of special cases for readability
+            case ConstantF(Data.Str(str)) => Terminal("Str" :: "Constant" :: nodeType, Some(str.shows))
+            case InvokeF(structural.ObjectProject, expr :: name :: Nil) =>
+              (ra.render(expr), ra.render(name)) match {
+                case (RenderedTree(_, Some(x), Nil), RenderedTree(_, Some(n), Nil)) =>
+                  Terminal("ObjectProject" :: nodeType, Some(x + "[" + n + "]"))
+                case (x, n) => NonTerminal("Invoke" :: nodeType, Some(ObjectProject.name), x :: n :: Nil)
+              }
+
+            case ReadF(name)              => Terminal("Read" :: nodeType, Some(name.pathname))
+            case ConstantF(data)          => Terminal("Constant" :: nodeType, Some(data.toString))
+            case InvokeF(func, args)      => NonTerminal("Invoke" :: nodeType, Some(func.name), args.map(ra.render))
+            case FreeF(name)              => Terminal("Free" :: nodeType, Some(name.toString))
+            case LetF(ident, form, body)  => NonTerminal("Let" :: nodeType, Some(ident.toString), List(ra.render(form), ra.render(body)))
             case TypecheckF(expr, typ, cont, fallback) =>
               NonTerminal("Typecheck" :: nodeType, Some(typ.shows),
                 List(ra.render(expr), ra.render(cont), ra.render(fallback)))
--- a/core/src/main/scala/quasar/logicalplan.scala
+++ b/core/src/main/scala/quasar/logicalplan.scala
@@ -221,8 +221,7 @@ object LogicalPlan {
   }
 
   def rename[M[_]: Monad](f: Symbol => M[Symbol])(t: Fix[LogicalPlan]): M[Fix[LogicalPlan]] =
-    Corecursive[Fix].anaM[LogicalPlan, M, (Map[Symbol, Symbol], Fix[LogicalPlan])](
-      (Map(), t))(renameƒ(f))
+    (Map[Symbol, Symbol](), t).anaM(renameƒ(f))
 
   def normalizeTempNames(t: Fix[LogicalPlan]) =
     rename[State[NameGen, ?]](κ(freshName("tmp")))(t).evalZero
--- a/core/src/main/scala/quasar/optimizer.scala
+++ b/core/src/main/scala/quasar/optimizer.scala
@@ -135,7 +135,7 @@ object Optimizer {
   }
 
   def preferProjections(t: Fix[LogicalPlan]): Fix[LogicalPlan] =
-    boundPara(t)(preferProjectionsƒ)._1.transCata(repeatedly(simplifyƒ))
+    t.boundPara(preferProjectionsƒ)._1.transCata(repeatedly(simplifyƒ))
 
   val elideTypeCheckƒ: LogicalPlan[Fix[LogicalPlan]] => Fix[LogicalPlan] = {
     case LetF(n, b, Fix(TypecheckF(Fix(FreeF(nf)), _, cont, _)))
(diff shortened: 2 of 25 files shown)


jdegoes:
John's law of testing indeterminism: any timeout, no matter how generous, will eventually produce a false negative that causes developers who did not write the timeout to curse the one who did. :smiling_imp:

mossprescott:
I'm open to suggestions of other ways to verify that the optimization is effective. If we don't explicitly time out those tests, the failure mode is hanging, filling the DB with garbage, and then eventually returning the correct result. This was just meant to give a more decipherable error.

core/src/main/scala/quasar/recursionschemes/package.scala
```diff
@@ -321,6 +322,21 @@ package object recursionschemes extends CofreeInstances with FreeInstances {
...
+      Applicative[M].sequence(B.bindings[M[A]](t, B.G.map(b)(_.point[M]))(s => loop(s, b)))(B.G).flatMap { newB =>
+        B.subst(t, newB).cata[M[A]](
+          _.point[M],
+          t.traverse(x => loop(x.unFix, newB).map((x, _))).flatMap(f))
+      }
+    }
+
+    loop(t.unFix, B.initial)
+  }
+
```
mossprescott: @sellout: not very happy about how cumbersome this all is. Maybe you can suggest some way to simplify?

it/src/test/scala/quasar/RegressionSpec.scala
```diff
@@ -16,16 +17,19 @@ import argonaut._, Argonaut._
 import org.specs2.execute._
 import org.specs2.matcher._
 import org.specs2.specification.{Example}
+import org.specs2.time.NoTimeConversions
```
mossprescott: This ends up not being necessary. I'll remove it.

it/src/test/scala/quasar/RegressionSpec2.scala
```diff
@@ -77,7 +78,10 @@ class RegressionSpec2
       _    <- Task.delay(println(test.query))
       _    <- run(test.data.cata(verifyDataExists, success.point[F]))
       data =  execQuery(test.query, test.variables)
-      res  <- verifyResults(test.expected, data, run)
+      // FIXME: would prefer to time evaluation, but not verification (see the old RegressionSpec)
+      t    <- timed(verifyResults(test.expected, data, run))
```
mossprescott: @estewei: I could not figure out how to put Tasks around the first element of stream to time the execution of the query, so had to settle for this, which is less precise (it won't be clear if the query is slow or if verifying the result is, and we happen to know that verifying result *can* be slow). Any ideas?

jdegoes:
I think the regression tests should verify correctness, not necessarily performance (though performance metrics could be useful so long as they are not a criteria to fail the build), while unit tests should verify the correct (e.g. "performant") plans are being generated.

I like completely deterministic builds. Either they always succeed, or they always fail.

wemrysi:
Aye, I think performance testing is definitely useful, but it would seem like doing this out-of-band via benchmarking or similar with results reported in a way that we can then alert on trends might be a better way to measure than as part of the test suite. Especially given that we can't really hold the testing environment constant on travis.

mossprescott:
Given the emerging consensus, I'll back out the timeout.

sellout:
Yeah, I wouldn’t even really call this an optimization, so much as normalization. It happens in the LP, which _has_ no performance characteristics. We’re basically picking one out of a set of semantically-equivalent LPs.<sup><a name="rt1">[1](#ft1)</a></sup>

I am sympathetic to getting more useful feedback than “the tests took an hour, but everything succeeded in the end.” Using benchmarks helps, but getting that info before we merge into master would help more. So, we should probably add an issue for getting a dedicated benchmark machine set up (Travis is way too inconsistent), and it should also be simple to ask the machine to benchmark a particular fork/branch so we can debug performance regressions. But, obviously, not a part of this PR.

<a name="ft1">[1](#rt1)</a>: Once we have BlackShield, I could see LP having only `CrossJoin` (and outer joins), with this pass split into two pieces – one that pushes some conditions before the Cross (applied during LP normalization), and another that converts Crosses to Inners by moving some conditions into the join clause (applied to BlackShield). The elimination of Inners from LP makes it more orthogonal, and then we add in the “hints” to the backend when we convert to BlackShield.

mossprescott:
Updated and rebased.

core/src/main/scala/quasar/recursionschemes/package.scala
sellout: I don’t really have anything substantive here – a few things that could be simplified, `Fix` could be generalized to `T: Recursive`, and the whole thing could move to the `Recursive` type class. But that’s all just pushing food around the plate.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -190,6 +200,37 @@ object LogicalPlan {
...
+        val v: LogicalPlan[(Map[Symbol, Symbol], Fix[LogicalPlan])] =
+          FreeF(bound.get(sym).getOrElse(sym))
+        v.point[M]
+      case t =>
+        t.strengthL(bound).point[M]
+    }
+
+  def rename[M[_]: Monad](f: Symbol => M[Symbol])(t: Fix[LogicalPlan]): M[Fix[LogicalPlan]] =
+    Corecursive[Fix].anaM[LogicalPlan, M, (Map[Symbol, Symbol], Fix[LogicalPlan])](
+      (Map(), t))((renameƒ(f) _).tupled)
```
sellout: Ugh, we really gotta do the IdOps thing for Corecursive, so this can be written as
```scala
(Map(), t).anaM[LogicalPlan](renameƒ(f).tupled)
```

Oh, and the function rewrite I mentioned above should allow `(renameƒ(f) _).tupled` to be `renameƒ(f).tupled`.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -190,6 +200,37 @@ object LogicalPlan {
...
+
+  def normalizeTempNames(t: Fix[LogicalPlan]) =
+    rename[State[NameGen, ?]](κ(freshName("tmp")))(t).evalZero
+
+  val normalizeLetsƒ: LogicalPlan[Fix[LogicalPlan]] => Option[LogicalPlan[Fix[LogicalPlan]]] = {
+      case LetF(b, Fix(LetF(a, x1, x2)), x3) => Some(LetF(a, x1, normalizeLets(Let(b, x2, x3))))
```
sellout: You’re calling `normalizeLets` inside `normalizeLetsƒ`, which seems nasty. Thankfully, the types of FunctorT are very flexible. Are you sure you don’t just want to write this as `LetF(a, x1, Let(b, x2, x3)).some` and then use `transAna` instead of `transCata`?

core/src/main/scala/quasar/optimizer.scala
```diff
@@ -161,4 +168,134 @@ object Optimizer {
...
+    // A condition that refers to left and right sources using equality, so may
+    // be rewritten into the join condition:
+    final case class EquiCond[A](run: (Fix[LogicalPlan], Fix[LogicalPlan]) => A) extends Component[A]
+    // A condition which refers only to the left source:
+    final case class LeftCond[A](run: Fix[LogicalPlan] => A) extends Component[A]
+    // A condition which refers only to the right source:
+    final case class RightCond[A](run: Fix[LogicalPlan] => A) extends Component[A]
+    // A condition which refers to both sources but doesn't have the right shape
+    // to become the join condition:
+    final case class OtherCond[A](run: (Fix[LogicalPlan], Fix[LogicalPlan]) => A) extends Component[A]
```
sellout: Could you add a comment/issue that we want an optional pre-backend pass to extract `OtherCond`s from the join? Then, in the LP, we can put _all_ conditions that reference both sources into the join clause, and let backends that don’t support full θJoins fall back to equijoin+filter.

This is already a step in the right direction there, because previously if an “`OtherCond`” was written into a join condition by a user, the Mongo planner would error on that query, but now you extract it into a Filter and we end up with a happy Mongo. But it still happens a bit too soon (although it won’t matter until we have another backend or get around to supporting θJoins in Mongo).

core/src/main/scala/quasar/optimizer.scala
```diff
@@ -161,4 +168,134 @@ object Optimizer {
...
+    }
+
+
+    node match {
+      case Filter((src, Fix(InnerJoin(joinL :: joinR :: joinCond :: Nil))) :: (cond, _) :: Nil) =>
+        val comps = flattenAnd(joinCond).map(toComp(joinL, joinR)) ++
+                    flattenAnd(cond).map(toComp(JoinDir.Left.projectFrom(src), JoinDir.Right.projectFrom(src)))
+        newJoin(joinL, joinR, comps)
+
+      // TODO: enable this when it can be made to not conflict (see tests pending SD-1190)
```
sellout: What does this mean: “when it can be made to not conflict”? I get that this is the case where we’d pull “bad” conditions _out_ of a pre-existing join, but I don’t see where the conflict is.

core/src/main/scala/quasar/recursionschemes/TraverseT.scala
```diff
@@ -55,4 +55,9 @@ object TraverseT {
       def traverse[M[_]: Applicative, F[_], G[_]](t: T[F])(f: F[T[F]] => M[G[T[G]]]) =
         f(Recursive[T].project(t)).map(Corecursive[T].embed)
     }
+
+  /** Import from this object instead of `ops._` to get just ops for the
+    * methods of TraverseT, and not those inherited from FunctorT.
+    */
```
sellout: I would reference the simulacrum issue in this comment.

core/src/test/scala/quasar/helpers.scala
```diff
@@ -24,13 +24,14 @@ trait CompilerHelpers extends Specification with TermLogicalPlanMatchers {
...
+  // NB: this plan is simplified and normalized, but not optimized
   def compileExp(query: String): Fix[LogicalPlan] =
     compile(query).fold(
       e => throw new RuntimeException("could not compile query for expected value: " + query + "; " + e),
-      _.transCata(repeatedly(Optimizer.simplifyƒ)))
+      lp => (LogicalPlan.normalizeLets _ >>> LogicalPlan.normalizeTempNames _)(Optimizer.simplify(lp)))
 
   def testLogicalPlanCompile(query: String, expected: Fix[LogicalPlan]) = {
-    compile(query).toEither must beRight(equalToPlan(expected))
+    compile(query).map(Optimizer.optimize).toEither must beRight(equalToPlan(expected))
```
sellout: So, why is the expected value (from `compileExp`) _not_ optimized, but the actual one is?

I think the comment should explain the why.

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -3229,18 +3229,32 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
...
                     $field("right", "city"),
                     $literal(Bson.Undefined)),
-                "__tmp7" -> $field("right"),
-                "__tmp8" -> $field("left"),
-                "__tmp9" -> $eq($field("left", "_id"), $field("right", "_id"))),
+                "__tmp11" -> $field("right"),
+                "__tmp12" -> $field("right", "pop"),
+                "__tmp13" -> $field("left"),
+                "__tmp14" -> $field("left", "pop"),
+                "__tmp15" -> $lt($field("left", "pop"), $field("right", "pop"))),
```
sellout: Ugh, this is a broken $cond case. We do the comparison in this $project, then type check in the following $match. Not your problem, but sad to see it.

core/src/main/scala/quasar/optimizer.scala
mossprescott: What happens is the InnerJoin gets rewritten  first, as a Let, and then when the original Filter is encountered it isn't in a shape that can be recognized and it isn't rewritten. The idea of that ticket is to come back and make this smarter and also pick up the theta-join case. I like your idea about making that an optional rewrite also.

sellout: Cool, can you add something like that to this comment or the issue?

sellout:
Looks pretty good. A couple suggestions re: comments/explanations and recursion, but nothing serious.

core/src/test/scala/quasar/helpers.scala
mossprescott: The idea is that you give`compileExp` the query that you happen to know compiles directly to the desired LP. This constrains the result more precisely than if both queries were just required to optimize to the same plan. But of course it also constrains the compiler to behave that way, which we might not want. 

I would be ok with the alternative: optimize both plans and just make sure they end up the same. Might even be able to do a property-based teat that way.

sellout: Ok, that makes sense to me – since currently optimize only covers things that can be explicitly controlled in the query. But add that to the comment.

core/src/main/scala/quasar/recursionschemes/package.scala
mossprescott: Agree boundParam/Cata should be in Recursive. But boundAttribute has to stay here, correct?

sellout: Hrmm, is `boundAttribute(f)` not equivalent to `boundCata(attribute(f))`? If so, it can be deleted.

core/src/main/scala/quasar/logicalplan.scala
mossprescott: This *is* the rewrite you asked for earlier, and it still needs `_`, just because it made `renameƒ` a little easier to write, but I could do it the other way, if you feel strongly.

I count a couple of dozen places we have to do `Corecursive[...].`, so I'd be glad to do the `IdOps` thing. I'll just put it in `fp.package`?

core/src/main/scala/quasar/logicalplan.scala
mossprescott: No, I'm not sure, and I didn't feel good about this either (I think I @mentioned you on the commit a while back). I'll give it a try.

mossprescott: `@mentioned`. Awesome.

You're a genius. `transAna` is all I needed.

core/src/main/scala/quasar/logicalplan.scala
sellout: I don’t feel too strongly about it. Just took me a little longer to read the `renameƒ` definition, because I didn’t see the function shape I was expecting.

I think `IdOps` should be in `recursionschemes.package` … but maybe that’s what you meant?

core/src/main/scala/quasar/logicalplan.scala
sellout: Cool. Sometimes I think “oh, I just need to go top-down instead of bottom-up” (or vice versa), but then I run into some other issue and need to switch to a different algebra type. But at least with the `trans` stuff, it’s easier to experiment.

core/src/main/scala/quasar/optimizer.scala
mossprescott: Between [SD-1190](https://slamdata.atlassian.net/browse/SD-1190) and [SD-1054](https://slamdata.atlassian.net/browse/SD-1054) I feel like we've got this pretty well covered. I'll add a comment there, OK?

core/src/main/scala/quasar/logicalplan.scala
mossprescott: I'll try switching it; I don't think it will be bad.

And, no, that's not what I meant, but obviously you're correct and it seems like we must be importing `recursionschemes._` in the relevant places, so that makes sense.

core/src/main/scala/quasar/optimizer.scala
sellout: Yeah, that looks good.

mossprescott:
Probably will squash some of those commits before merging, but this should be ready to review again.

sellout:
Yeah, so … just delete that giant file and you’re good.

mossprescott:
By the way, I just noticed I didn't address your comment about `boundAttribute`. The answer is it's not *trivially* equivalent to `boundCata(attribute(f))`. The fixed-ness of the types doesn't line up. I could give it a closer look tomorrow if it's important to you. It's used in exactly one place, and it might be easy to use boundPara instead (or, maybe better, to implement boundAttribute that way).

sellout:
I don’t think changing `boundAttribute` is too important. I tried the trivial conversion myself and realized it didn’t line up, so I didn’t push it.

mossprescott:
Alright, I'll do a little squashing and merge. Thanks!

PR #1030, opened 2015-12-14
[SD-1042] Ready/remove java property

jedesah:


jedesah:
@sellout Can you have a look at this? Should take about 30 seconds to review :smile:

sellout:
I already did … but apparently never submitted my comment(?). I did update the label, though.

jedesah:
Ah, ya what's the protocol again? we comment because there is no way to get notified of a label change?

Anyway, thx for reviewing! :smile:

sellout:
Yeah, exactly.

PR #1031, opened 2015-12-15
Implement streaming distinct() on MongoDB.

sellout:
SD-956 #done

A few other changes done to support this:

* Replaced the `java.lang.Object` BSON representation with the new
  `BsonValue` one added in the 3.0 driver. This is a bit more principled
  and gets rid of a few places where we had to suppress WartRemover
  warnings.
* Implemented `org.bson.conversions.Bson` for our Bson type, which
  eliminates some cases where we previously had to explicitly convert to
  BsonDocument.
* Modified `Workflow.distinct` to distinct in a way that can be
  optimized by `simplifyGroupƒ`, which also makes it easier to match in
  the evaluator.

Otherwise, this follows the same form as the `find()` and `count()`
implementations.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/filesystem.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/filesystem.scala
@@ -21,6 +21,8 @@ import quasar.fp._
 import quasar._, Backend._, Errors._
 import quasar.fs._, Path._
 
+import com.mongodb.client.FindIterable
+import org.bson.BsonDocument
 import scalaz._, Scalaz._
 import scalaz.concurrent._
 import scalaz.stream._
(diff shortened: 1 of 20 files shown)


sellout:
And, @drostron, since @mossprescott just reviewed other similar changes and also has experience with this code, you may want to ask him about anything that’s not obvious.

core/src/main/scala/quasar/physical/mongodb/filesystem.scala
```diff
@@ -32,14 +34,19 @@ trait MongoDbFileSystem extends PlannerBackend[Workflow.Crystallized] {
...
       Process[ETask[ResultError, ?], Data] = {
-    val skipper = (it: com.mongodb.client.FindIterable[org.bson.Document]) => it.skip(offset.toInt)
-    // NB As per the MondoDB documentation, we inverse the value of limit in order to retrieve a single batch.
-    // The documentation around this is very confusing, but it would appear that proceeding this way allows us
-    // to retrieve the data in a single batch. Behavior of the limit parameter on large datasets is tested in
-    // FileSystemSpecs and appears to work as expected.
-    val limiter = (it: com.mongodb.client.FindIterable[org.bson.Document]) => limit.map(v => it.limit(-v.toInt)).getOrElse(it)
-
-    val skipperAndLimiter = skipper andThen limiter
+    // In another world, these would be NatTrans.
```
drostron: Why not this world?

sellout: If they weren’t restricted to `scan0`, it’d probably be worthwhile, but in Scala it adds a lot of noise, so without being able to use these more generally, it’s not worth the tradeoff.

drostron: Ok.

drostron:
👍🏿

PR #1032, opened 2015-12-15
add MongoDB 3.2 to the test matrix

mossprescott:


README.md
```diff
@@ -55,15 +55,16 @@ Set the following environment variables:
 For bash this would like such,
 
 ```bash
+export QUASAR_MONGODB_3_2="{\"mongodb\":{\"connectionUri\":\"mongodb://`mongoURL`\"}}"
```
jedesah: I am not an expert at bash, so maybe I am missing something, but would it not be possible to only have `QUASAR_MONGODB`, since they all have the same value?

README.md
```diff
@@ -55,15 +55,16 @@ Set the following environment variables:
...
 ```
 
 where \`mongoURL\` is the url at which one can find a Mongo database. For example \`mongoURL\` would probably look
 something like `localhost:27017` for a local installation. This means the integration tests will be run against
-both MongoDB version 2.6 and version 3.0. Alternatively, you can choose to install only one of these and run the integration
-tests against only that one database. Simply omit one of the two in order to avoid testing against it. On the integration
-server, the tests are run against both supported versions of MongoDB.
+both MongoDB versions 2.6, 3.0, and 3.2. Alternatively, you can choose to install only one of these and run the integration
+tests against only that one database. Simply omit a version in order to avoid testing against it. On the integration
+server, the tests are run against all supported versions of MongoDB, as well as read-only configurations.
```
jedesah: Not a big deal, but we only have one read-only configuration.

README.md
jedesah: Nevermind, did not realize this was the README :smile:

mossprescott: Haha. I was confused by the same thing and tried to make them different values, but found that it ended up even more confusing so :shrug:

README.md
mossprescott: True for now, but if it was less painful to add them I would be tempted to add configurations for 3.2 read-only and for the old storage engine. In principle, we could have lots of them and I don't think this description needs to stay strictly in sync.

jedesah:
LGTM! :smile:

README.md
jedesah: Agreed.

wemrysi: Re pain, how about moving all this test config to a file of some sort? Not necessarily in this PR, but I think we're convinced the env-var approach doesn't really scale.

mossprescott: I think that makes sense, certainly at the point that we want to expand the number of variables to include things like multiple MongoDB configurations, platforms, etc.

See [SD-1198](https://slamdata.atlassian.net/browse/SD-1198).

PR #1033, opened 2015-12-16
SD-1006 – Implement the pow (^) operator.

sellout:
SD-1006 #done


core/src/main/scala/quasar/std/math.scala
```diff
@@ -74,6 +74,23 @@ trait MathLib extends Library {
     }) ||| numericWidening,
     biReflexiveUnapply)
 
+  val Power = Mapping("(^)", "Raises the first argument to the power of the second",
+    Type.Numeric, Type.Numeric :: Type.Numeric :: Nil,
+    noSimplification,
+    (partialTyper {
+      case _ :: Type.Const(Data.Number(v2)) :: Nil if v2 == 0 =>
```
jedesah: I am wondering if it might be ever so slightly simpler if we defined `val zero = Type.Const(Data.Number(0))` and the the case pattern could be ```case _ :: `zero` :: Nil =>```

Then we could re-use `zero` again for ```case `zero` :: _ :: Nil =>```. There might be other places in this file where `zero` might be useful, it seems like a constant value that is likely to lead to simplification in many numeric functions.

Anyway, maybe it's not worth, just crossed my mind when reading the code.

sellout: Yeah, I like that … we use it in a bunch of other places in the file, too.

core/src/main/scala/quasar/std/math.scala
```diff
@@ -74,6 +74,23 @@ trait MathLib extends Library {
...
+        Type.Const(Data.Int(1))
+      case v1 :: Type.Const(Data.Number(v2)) :: Nil if v2 == 1 => v1
+      case (zero @ Type.Const(Data.Number(v1))) :: _ :: Nil if v1 == 0 => zero
+      case Type.Const(Data.Int(v1)) :: Type.Const(Data.Int(v2)) :: Nil
+          if v2.isValidInt =>
+        Type.Const(Data.Int(v1.pow(v2.toInt)))
+      case Type.Const(Data.Number(v1)) :: Type.Const(Data.Int(v2)) :: Nil
+          if v2.isValidInt =>
```
jedesah: Is there no function in Scala/Java that calculates `pow` over non-Integers? It's too bad, it would be nice if we could pre-compute any `pow` expression operating over constants as opposed to only in the case of an Integer in the exponent position.

core/src/test/scala/quasar/std/math.scala
```diff
@@ -141,6 +141,22 @@ class MathSpec extends Specification with ScalaCheck with ValidationMatchers wit
...
 
+    "simplify number raised to 0th power" in {
+      Power(Type.Int, Const(Int(0))) should beSuccessful(Const(Int(1)))
+    }
+
+    "simplify 0 raised to any (non-zero) power" in {
+      Power(Const(Int(0)), Type.Int) should beSuccessful(Const(Int(0)))
+    }
+
+    "simplify any number raised to 1st power" in {
```
jedesah: Ideally, it would be nice if this was a property, but maybe it's not for pragmatism.

sellout: Actually, I should change how this is done – this also needs a simplification along with it – we don’t just want the type of the result to be the type of the first argument, but rather, if the type (well, currently just the literal expression) of the second argument is constantly `1`, then we should elide the call to `Power` completely.

jedesah: Fair enough.

core/src/main/scala/quasar/std/math.scala
sellout: There’s not that I could find – only BigNum raised to an `int`. When we have a BlackShield evaluator, we’ll need to handle any BigNum exponent, but right now, I think this covers the useful stuff.

jedesah: Fair enough.

core/src/test/scala/quasar/std/math.scala
```diff
@@ -141,6 +141,33 @@ class MathSpec extends Specification with ScalaCheck with ValidationMatchers wit
       expr must beFailing
     }
 
+    "typecheck number raised to 0th power" ! prop { (t: Type) =>
+      if (Type.Numeric.contains(t))
+        Power(t, Const(Int(0))) should beSuccessful(Const(Int(1)))
+      else 1 must_== 1
```
sellout: In these tests, the if/else is terrible – what is a better way to restrict `t` to be a subtype of `Type.Numeric`?

mossprescott: Certainly we could add an Arbitrary for Numeric, if we don't have that already.

And by the way, the right way to do this sort of thing when it can't be avoided is:
```scala
Type.Numeric.contains(t) ==>
  ...
```
That forces scalacheck to keep generating values until the test runs the expected number of times. And if you screw it up, it will complain that it *couldn't* generate enough values.

sellout: I’d prefer something more general than Arbitrary for Numeric – e.g., a function that takes a type and returns an Arbitrary[Type] that is a subtype of that type – or, just how do we filter Arbitraries … which it looks like is feasible with `==>`, so I can do something with that.

mossprescott: You can apply the condition to the generator with `Gen.suchThat`, which basically works the same way as `==>` but transforms the generator.

The issue either way is that if your condition is particularly specific, you end up generating and throwing away a lot of values to find ones that match. Seems like these things should be written in Prolog, maybe?

jedesah: @sellout Do you think there is a particular problem with `(t: Type.Numeric)`?

Seems to me to be the nicest solution as far as the property is concerned. Then we can possibly implement a generic Generator that filters on a particular subtype like you described.

sellout: Well, Numeric is a val, not a type. Is there a way to do that?

wemrysi: You could create a newtype representing the subset of values, i.e. `final case class NumericType(t: Type) extends AnyVal` and then an `Arbitrary[NumericType]`.

jedesah: Well, one way would be to define `Coproduct` as `Coproduct[L <: Type, R <: Type](left: L, right: R)` instead of `Coproduct(left: Type, right: Type)` and then say that `type Numeric = Coproduct[Int,Dec]`

sellout: Ended up just creating `arbitraryNumeric: Arbitrary[Type]`, rather than modifying the Type AST (which should eventually become a functor, anyway).

So, this is ready for review again.

core/src/main/scala/quasar/std/math.scala
```diff
@@ -38,8 +42,8 @@ trait MathLib extends Library {
     MathAbs, MathAbs :: MathRel :: Nil,
     noSimplification,
     (partialTyper {
-      case Type.Const(Data.Number(v1)) :: t2 :: Nil if (v1.signum == 0) && (Type.Numeric contains t2) => t2
-      case t1 :: Type.Const(Data.Number(v2)) :: Nil if (Type.Numeric contains t1) && (v2.signum == 0) => t1
+      case Zero :: t2 :: Nil if Type.Numeric.contains(t2) => t2
```
jedesah: This went from `Data.Number` to `Data.Int`. This means we will no longer match on `Data.Dec`. Is that intended?

Also, I am surprised that tests pass. It seems to me what we want here is: ```case `Zero` :: t2 :: ...``` or else `Zero` will be binded to any value no matter its type.

sellout: UpperCamel identifier is treated as a value, while lowerCamel is treated as a binding (unless you backtick it).

But you’re right about `Data.Int`. I don’t know if we ever end up with `Data.Dec(0.0)`, but it does seem like we should handle it. Unfortunately, `Data.Number` is just an object with `apply`/`unapply`, handling both Int and Dec cases. I suppose I could make `Zero` an object with an unapply that calls Number’s unapply.

jedesah: > UpperCamel identifier is treated as a value, while lowerCamel is treated as a binding

What!? :confused: I actually thought about that for a sec earlier, but I was like no, that can't be possible...

core/src/test/scala/quasar/std/math.scala
```diff
@@ -141,6 +141,28 @@ class MathSpec extends Specification with ScalaCheck with ValidationMatchers wit
       expr must beFailing
     }
 
+    "typecheck number raised to 0th power" ! arbitraryNumeric { (t: Type) =>
```
jedesah: What? is `arbitraryNumeric` the same one as above? If so, I did not know `specs2` (or is it scalacheck?) had support for such a notation. What does it look like for a property that takes a tuple? You need to use a custom `Arbitrary[Tuple2[A.B]]`?

core/src/test/scala/quasar/std/math.scala
```diff
@@ -141,6 +141,28 @@ class MathSpec extends Specification with ScalaCheck with ValidationMatchers wit
...
+    "typecheck 0 raised to any (non-zero) power" ! arbitraryNumeric { (t: Type) =>
+      (t != Const(Int(0))) ==>
+        (Power(Const(Int(0)), t) should beSuccessful(Const(Int(0))))
+    }
+
+    "typecheck any number raised to 1st power" ! arbitraryNumeric { (t: Type) =>
+      Power(t, Const(Int(1))) should beSuccessful(t)
+    }
+
+    "typecheck constant raised to int constant" in {
```
jedesah: I think this could be made into a property although with a guard on the size of the second argument.

core/src/main/scala/quasar/std/math.scala
```diff
@@ -74,14 +78,36 @@ trait MathLib extends Library {
     }) ||| numericWidening,
     biReflexiveUnapply)
 
-  /**
-   * Subtracts one value from another, promoting to decimal if either operand is decimal.
-   */
+  val Power = Mapping("(^)", "Raises the first argument to the power of the second",
+    Type.Numeric, Type.Numeric :: Type.Numeric :: Nil,
+    new Func.Simplifier {
```
jedesah: Why do we have a `Func.Simplifier` on `Power` but not on `Multiply` and `Add`?

core/src/test/scala/quasar/std/math.scala
sellout: It’s something I saw in `test/…/types.scala`. With multiple parameters, it looks like `! (arbitrarySimpleType, arbitrarySimpleType) { (t1: Type, t2: Type) => … }`.

jedesah: I like it, just surprised to see it and it does not seem to be documented on the specs2 website.

core/src/main/scala/quasar/std/math.scala
sellout: Yeah, we should have simplifiers on Add and Multiply for 0 and 1 respectively.

jedesah: Ah, okay, a separate concern, but at least I am less confused.

jedesah: Why doesn't the simplify logic mirror the `partialTyper` logic entirely except for `numericWidening` of course?

sellout: Simplifier handles cases that evaluation/constant-folding can’t. In `Power(Add(a, b), 1)`, the typer can’t return a Const type, so it can’t rewrite the term. But the simplifier rewrites that to `Add(a, b)`. However, in `Power(Add(a, b), 0)`, the first argument is irrelevant, and the whole term can be constant-folded to `Const(Data.Int(1))`.

More generally, constant folding (one of the things the typers gives us) only works when all the parts we care about (the result, plus any relevant arguments) are constants.

So, they really shouldn’t overlap at all in most cases. The simplifier _could_ duplicate all the constant-folding cases, but not vice versa.

jedesah: Ah, okay, I understand better now.

sellout: I'll add docs to that effect to functions.scala.

jedesah: :+1:

core/src/test/scala/quasar/std/math.scala
sellout: I tried to make this into a property, but using `(d: Double, i: Int)` (which is _more_ constrained than `BigDecimal.pow`) resulted in various issues, like underflow, and constraining the types led to the test being abandoned for too many discarded cases.

sellout:
Ready again.

core/src/main/scala/quasar/std/math.scala
jedesah: Did you look into making `Zero` an object with an unapply?

jedesah:
I am a little confused. I posted a question on whether you had tried making `Zero` an object with an unapply, but I can't seem to find it anymore. I posted as a reply to one of our threads, not sure what happened to it.

sellout:
Yeah, I see the comment in Slack, but not on the commit it claims to be attached to … weird.

In any case, yes, and I just forgot to commit it before pushing the previous changes. Pushed now.

jedesah:
Maybe it's worth doing the same thing for `Func.Simplifier`, there seems to be a few instances of `0` and `1` that could be made slightly easier to read. There is the challenge of coming up with a slightly different name for `Zero` and `One` though.

sellout:
Added the test and LP versions of Zero/One.

jedesah:
LGTM :+1:

PR #1034, opened 2015-12-22
General improvements to Free in core

jedesah:
This commits brings over the general purpose improvements to core that were done in the web branch (which is currently blocked on more work that needs to be done on Free in core) so that they can be available as soon as possible for projects that depend on Quasar as a library. Also to avoid conflicts with current work on Free in master.

This commit aggregation is co-authored with @wemrysi, specifically the part on the addition of the `Explain` term to the `QueryFile` algebra.

### DIFF
--- a/core/src/main/scala/quasar/fs/Natural.scala
+++ b/core/src/main/scala/quasar/fs/Natural.scala
@@ -22,26 +22,28 @@ import quasar.Predef._
 import scalaz._
 import scalaz.Tags.{Multiplication => Mult}
 
-final class Natural private (val run: Long) extends scala.AnyVal {
+final class Natural private[fs] (val value: Long) extends scala.AnyVal {
   def plus(other: Natural): Natural =
-    new Natural(run + other.run)
+    new Natural(value + other.value)
 
   def + (other: Natural): Natural =
     plus(other)
 
   def times(other: Natural): Natural =
-    new Natural(run * other.run)
+    new Natural(value * other.value)
 
   def * (other: Natural): Natural =
     times(other)
 
   def toInt: Int =
-    run.toInt
+    value.toInt
 }
 
 object Natural {
   def apply(n: Long): Option[Natural] =
     Some(n).filter(_ >= 0).map(new Natural(_))
+
+  def unapply(n: Natural): Option[Long] = Some(n.value)
 
   val _0: Natural = new Natural(0)
   val _1: Natural = new Natural(1)
@@ -55,7 +57,7 @@ object Natural {
   val _9: Natural = new Natural(9)
 
   def fromPositive(n: Positive): Natural =
-    new Natural(n.run)
+    new Natural(n.value)
 
   implicit val naturalAddition: Monoid[Natural] =
     Monoid.instance(_ + _, _0)
@@ -64,4 +66,8 @@ object Natural {
     Monoid.instance(
       (x, y) => Mult(Mult.unwrap(x) * Mult.unwrap(y)),
       Mult(_1))
+
+  implicit val equal: Equal[Natural] = Equal.equalA
+
+  implicit val show: Show[Natural] = Show.shows(_.value.toString)
 }
--- a/core/src/main/scala/quasar/fs/QueryFile.scala
+++ b/core/src/main/scala/quasar/fs/QueryFile.scala
@@ -32,6 +32,9 @@ object QueryFile {
   final case class ExecutePlan(lp: Fix[LogicalPlan], out: AFile)
     extends QueryFile[(PhaseResults, FileSystemError \/ ResultFile)]
 
+  final case class Explain(lp: Fix[LogicalPlan], out: AFile)
+    extends QueryFile[(PhaseResults, FileSystemError \/ PhaseResult)]
+
   /** TODO: While this is a bit better in one dimension here in `QueryFile`,
     *       `@mossprescott` points out it is still a bit of a stretch to include
     *       in this algebra. We need to revisit this and probably add algebras
@@ -66,19 +69,25 @@ object QueryFile {
     /** Returns the path to the result of executing the given [[LogicalPlan]] */
     def execute_(plan: Fix[LogicalPlan])
                 (implicit M: ManageFile.Ops[S]): ExecM[ResultFile] = {
-
-      val outFile = plan.foldMap {
-        case Fix(LogicalPlan.ReadF(p)) => Vector(p)
-        case _                         => Vector.empty
-      }.headOption flatMap pathToAbsFile cata (M.tempFileNear, M.anyTempFile)
-
       for {
-        out <- toExec(outFile)
+        out <- toExec(firstFile(plan) cata (M.tempFileNear, M.anyTempFile))
         rf0 <- execute(plan, out)
         rf1 =  user.getOrModify(rf0)
                  .map(usr => if (usr == out) Temp(usr) else User(usr))
                  .merge
       } yield rf1
+    }
+
+    /** Returns a description of how the the given logical plan will be executed
+      * along with any error encountered during planning.
+      */
+    def explain(plan: Fix[LogicalPlan])
+               (implicit M: ManageFile.Ops[S])
+               : F[(PhaseResults, FileSystemError \/ PhaseResult)] = {
+
+      firstFile(plan)
+        .cata(M.tempFileNear, M.anyTempFile)
+        .flatMap(f => lift(Explain(plan, f)))
     }
 
     /** Returns the source of values from the result of executing the given
@@ -133,6 +142,24 @@ object QueryFile {
       comp flatMap (lp => evaluate(lp).translate[CompExecM](execToCompExec))
     }
 
+    /** Returns a description of how the the given SQL^2 query will be executed
+      * along with any error encountered during planning.
+      */
+    def explainQuery(query: sql.Expr, vars: Variables)
+                    (implicit M: ManageFile.Ops[S])
+                    : SemanticErrsT[F, (PhaseResults, FileSystemError \/ PhaseResult)] = {
+
+      type E[A, B] = EitherT[F, A, B]
+
+      queryPlan(query, vars).run.run match {
+        case (prs, \/-(lp)) =>
+          explain(lp).map(_.leftMap(prs ++ _)).liftM[SemanticErrsT]
+
+        case (_, -\/(semErrs)) =>
+          semErrs.raiseError[E, (PhaseResults, FileSystemError \/ PhaseResult)]
+      }
+    }
+
     /** Returns immediate children of the given directory, fails if the
       * directory does not exist.
       */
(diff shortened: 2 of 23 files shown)


jedesah:
@drostron Sorry about that, forgot to look at the results of testing `it` before pushing. Just addressed that, so hopefully should build cleanly.

jedesah:
@mossprescott Assigning you as I am currently pairing with @drostron and thus stopping him from reviewing the PR :smile:. Plus I have another PR for him to review coming soon.

core/src/main/scala/quasar/fs/Natural.scala
```diff
@@ -6,27 +6,29 @@ import quasar.Predef._
 import scalaz._
 import scalaz.Tags.{Multiplication => Mult}
 
-final class Natural private (val run: Long) extends scala.AnyVal {
+final class Natural private[fs] (val value: Long) extends scala.AnyVal {
```
mossprescott: Making this private to the package seems unfortunate, since there's a lot of other code in this package that isn't part of this implementation. Maybe these types should go to `quasar.fp.numeric` or something like that.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -16,6 +16,9 @@ object QueryFile {
   final case class ExecutePlan(lp: Fix[LogicalPlan], out: AFile)
     extends QueryFile[(PhaseResults, FileSystemError \/ ResultFile)]
 
+  final case class Explain(lp: Fix[LogicalPlan], out: AFile)
+    extends QueryFile[(PhaseResults, Option[FileSystemError])]
```
mossprescott: I've had some back and forth with @wemrysi on this before, but I would prefer `FileSystemError \/ Unit` for this type, which would parallel the other terms better. Probably even better would be to return the actual plan on the right, but we have no good type for it at the moment (other than `String`/`PhaseResult`). As it is, it's somewhat unclear how to interpret the result here: if there's an error, then the last `PhaseResult` probably describes the error, if `None`, then you have to pick out the last `PhaseResult`.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -28,6 +31,11 @@ object QueryFile {
   final case class ListContents(dir: ADir)
     extends QueryFile[FileSystemError \/ Set[Node]]
 
+  /** File used in explain output when a `LogicalPlan` does not contain any
+    * file references.
+    */
+  val DefaultExplainOut: AFile = rootDir </> dir("explain") </> file("out")
```
mossprescott: Seems to be unused. Can we get rid of this?

core/src/main/scala/quasar/fs/Natural.scala
jedesah: Yes, really these types should not be defined here. They get deleted in #1025 which I hope to update and get merged soonish as my 10% time for this week. I would almost be inclined to hold off on making changes that would create annoying conflicts, but I can change if you think it's best to only merge this in the best state possible.

core/src/main/scala/quasar/fs/QueryFile.scala
jedesah: I am not a huge fan of a function that returns `Unit` unless it's main purpose is an effect which is not the case for `Explain`. As you said, I think it would probably be best to return the plan, we can use a weak type like `String` or `PhaseResult` for now and consider using a stronger type later?

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -12,6 +13,7 @@ package object fs {
 
   type FileSystem0[A] = Coproduct[WriteFileF, ManageFileF, A]
   type FileSystem1[A] = Coproduct[ReadFileF, FileSystem0, A]
+  /** FileSystem[A] = [[ReadFileF]] \/ [[WriteFileF]] \/ [[ManageFileF]] \/ [[QueryFileF]] */
```
mossprescott: I appreciate the comment, but I think the use of double-brackets and `\/` are confusing here. Also, isn't the order actually Query, Read, Write, Manage? Not sure if that ever matters, but it seems like it could be confusing at some point.

I guess I would prefer something like `/** FileSystem = QueryFileF or ReadFileF or WriteFileF or ManageFileF  */`, and not try too hard to make it look like actual code.

core/src/main/scala/quasar/fs/QueryFile.scala
jedesah: That seems fair enough.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -30,7 +32,11 @@ package object fs {
     w: WriteFile ~> M,
     m: ManageFile ~> M
   ): FileSystem ~> M =
-    interpret.interpret4[QueryFileF, ReadFileF, WriteFileF, ManageFileF, M](
+    interpret4[QueryFileF, ReadFileF, WriteFileF, ManageFileF, M](
       Coyoneda.liftTF(q), Coyoneda.liftTF(r), Coyoneda.liftTF(w), Coyoneda.liftTF(m))
+
+  def convert(path: pathy.Path[_,_,Sandboxed]): fs.Path = fs.Path(posixCodec.printPath(path))
```
mossprescott: Presumably this and the one after are temporary until we port everything, so maybe a comment indicating that?

core/src/main/scala/quasar/fs/package.scala
jedesah: I am okay with that.

core/src/main/scala/quasar/fs/package.scala
jedesah: Good idea.

core/src/main/scala/quasar/fs/package.scala
jedesah: btw, the order does not matter that much in the sense that we would typically use `Inject` to interact with `FileSystem`, but best to put it in the right order if one needs to work with it "manually".

core/src/test/scala/quasar/fs/QueryFileSpec.scala
```diff
@@ -2,60 +2,76 @@ package quasar
...
-      "return false when file doesn't exist" ! prop { (f1: AFile, f2: AFile) =>
-        run(query.fileExists(f1))
-          .eval(InMemState fromFiles Map(f2 -> Vector(Data.Int(1))))
-          .run must beFalse
+      "return false when file doesn't exist" ! prop { (absentFile: AFile, s: SingleFileMemState) =>
+        Mem.interpret(query.fileExists(absentFile)).eval(s.state) must beFalse
       }
 
-      "return false when dir exists with same name as file" ! prop { f: AFile =>
+      "return false when dir exists with same name as file" ! prop { (f: AFile, data: Vector[Data]) =>
```
mossprescott: What does it prove to range over arbitrary file contents here? Seems like just burning cycles if all we're looking at is whether the file exists. The only potentially interesting case I can think of is empty Vector vs. non-empty.

core/src/test/scala/quasar/fs/QueryFileSpec.scala
```diff
@@ -2,60 +2,76 @@ package quasar
...
+      "return true when file exists" ! prop { s: SingleFileMemState =>
+        Mem.interpret(query.fileExists(s.file)).eval(s.state) must beTrue
       }
 
-      "return false when file doesn't exist" ! prop { (f1: AFile, f2: AFile) =>
-        run(query.fileExists(f1))
-          .eval(InMemState fromFiles Map(f2 -> Vector(Data.Int(1))))
-          .run must beFalse
+      "return false when file doesn't exist" ! prop { (absentFile: AFile, s: SingleFileMemState) =>
+        Mem.interpret(query.fileExists(absentFile)).eval(s.state) must beFalse
```
mossprescott: What's the guarantee that `absentFile` isn't the same as `s.file` here?

core/src/test/scala/quasar/fs/ReadFileSpec.scala
```diff
@@ -20,7 +18,7 @@ class ReadFileSpec extends Specification with ScalaCheck with FileSystemFixture
 
       val p = write.append(f, xs.toProcess).drain ++ read.scanAll(f)
 
-      evalLogZero(p).run.toEither must beRight(xs)
+      MemTask.runLogEmpty(p).run must_== \/-(xs)
```
mossprescott: Why not `beRightDisjunction`?

core/src/test/scala/quasar/fs/QueryFileSpec.scala
jedesah: We already have a property, so, unless I am missing something, I think the extra cycles are negligible. Also, like you said, there are probably two interesting cases that we are testing here as opposed to previously, the case of an empty file (that exists) and a non-empty file.

core/src/main/scala/quasar/fs/Natural.scala
mossprescott: That's what I suspected. If we think the Refined stuff is going to get merged soon then definitely don't bother cleaning this up, but in that case why are we fiddling with the implementation?

core/src/test/scala/quasar/fs/QueryFileSpec.scala
jedesah: That's unrelated to this change, but a nice catch nonetheless!

core/src/test/scala/quasar/fs/ReadFileSpec.scala
jedesah: Some kind of bizarre implicit nonsense. It's why it was being converted to `scala.Either`. I preferred just constructing the disjunction as I don't really see the point of `beRight` or `beRightDisjunction` or `beTrue` anyway. It's not shorter and it does not provide a better error message either.

core/src/main/scala/quasar/fs/Natural.scala
jedesah: This PR are all the changes in the branch that translates the `web` sub-project to Free that don't need to live there. I made this change a long time ago in that branch in order to widen the type of `Posite` to `Natural`.

core/src/main/scala/quasar/fs/QueryFile.scala
mossprescott: Well, I would argue that `Err \/ Unit` isn't any more Unit-like than `Option[Err]`, but it's obviously a question of style. I like the idea of returning the plan, and I suppose `PhaseResult` is the best choice. What that leaves unclear is how it relates to `._1`. Is the final plan in the vector as well?

In a previous incarnation, a `PhaseResult` could represent an error, so all you would get from this method would be the vector and you could inspect the last phase to see the result. That was better in some ways, but awkward in others. I think we haven't figured out the right way to represent this yet:
- a non-empty log of phases, each with a name and result.
- the last phase may be either an error or an "executable" plan.
- the phases cross all architectural layers, so there's no common type for describing their results. Plus, we want the client to be able to render them nicely, so we capture RenderedTrees with "type"-decorated nodes when we can, String otherwise.

core/src/test/scala/quasar/fs/ReadFileSpec.scala
mossprescott: You're not gonna get an argument from me on that, but I would rather we were consistent.

core/src/test/scala/quasar/fs/QueryFileSpec.scala
mossprescott: I suppose you have a point.

If this kind of testing was (computationally) free, and if we could run every possible input, I would agree every input that doesn't need to be fixed should be arbitrary. But given that our builds are already painfully slow, I think we need to be judicious about how we deploy property-based tests. In this case, I think the interesting cases might be: 
1. file paths of various depths
2. non-empty vs. empty data

It is simply wasteful to test many different file names at the same level and many different examples of non-empty contents. Ideally we'd cover the interesting cases very well with only a half-dozen or so inputs, but of course there's no concise way to express that, so instead we'll run the default number of combinations and hope for the best.

At this point I'm just musing, so you can consider the comment retracted ;)

jedesah: I agree with your reasoning. I guess we should decide what our objective is. Currently tests for `core` take 40s and `web` take 34s. For me, as long as those numbers stay below one minute I am pretty happy and I have done some tweaking of property properties in the past to make sure that stays that way.

jedesah:
@mossprescott I think I addressed your concerns.

mossprescott:
LGTM. Sorry for the delay.

jedesah:
Not a problem! To be expected when you open a PR on Christmas eve! :stuck_out_tongue:

PR #1035, opened 2015-12-22
Update license headers, and verify them going forward

mossprescott:
Adds a step which will fail the build if any file does not have the correct license header.

I took this approach because I could not figure out why `createHeaders` is not being run before `compile`, as it seems like it was intended to be. In any case, failing the build gives us an actual verification that the code is in the right state, as opposed to just hoping the automation did its job.


drostron:
Yeah, strange that `createHeaders` was not being run as part of `compile`. A quick glance says it should be. Verification is a good addition.

👍🏼

PR #1037, opened 2015-12-31
Modify test for descendantFiles in order to avoid a spurious failure

jedesah:


wemrysi:
@jedesah Mind explaining what the failure was and how this change fixes it?

jedesah:
It's a failure related to a very particular kind of "nasty" path. I did not spend too much time trying to understand the exact nature of the problem as it's well known that we currently only support `PathOf[AlphaCharacters]`. I would like to see us understand what exactly is going on that makes "nasty" paths fail the test, (my first hypothesis is that it has something to do with converting to `fs.Path` and back) so the intent of this change is to make the build more stable and explicit the implicit fact that we do not support truly arbitrary paths right now. I will create a ticket if there isn't one already to track investigating arbitrary path issues.

jedesah:
For reference: https://slamdata.atlassian.net/browse/SD-1238

mossprescott:
I wouldn't say that we only support alpha. We went to some trouble to ensure that we could translate *any* Path to a properly-escaped collection name, for one thing. If conversion between pathy and old-style paths is leading to new limitations, I think that's something we want to know about. We might be able to live with it, but I think we want to know about it. For example, we might not be escaping paths with slashes in the nodes, and I think that would be OK for now. But if we can't handle paths with spaces or quotes, then that would not be OK to ship.

wemrysi:
I think it is worth spending some time to try and analyze one of the errors produced (be good to have that output and any analysis in this PR as well) as it might be that it is just `descendantFiles` or the test itself that has a defect.

jedesah:
Here is the output from the failure:

```
returns all descendants of the given directory
[error]      A counter-example is [DirIn(Root,DirName(.)), DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(/)),DirName(암腲嗐怞ꐫ)),DirName(넽匘䃒䃰坈)),DirName(ᯌ)),DirName(שּׁ軇ᨮ聿ﻃ㽠)),DirName(칁İ)),DirName(Ṓ뮰.ㄙ搢M.慨)),DirName(Ⱍ盂콷튇)), DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(./..溋캉秣)),DirName(⿤ե虒)),DirName(/₥顉鯮酩՗)),DirName(ᮎ.危뻉棽镀ꚵ/)),DirName(ࣴ.)),DirName(텸)),DirName(䘺ꆹ剺醫倸晵/✩蠻鮸)),DirName(㔅)),DirName(픺/竄멲)),DirName(椸륁촙仓)), DirIn(DirIn(Root,DirName(.)),DirName(ꁵꥇﰍ)), NonEmptyList(Ά㕜钑뱤諲ﰾ嘵በ횄)] (after 41 tries)
[error]      'Right(Set(FileIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(/)),DirName(암腲嗐怞ꐫ)),DirName(넽匘䃒䃰坈)),DirName(ᯌ)),DirName(שּׁ軇ᨮ聿ﻃ㽠)),DirName(칁İ)),DirName(Ṓ뮰.ㄙ搢M.慨)),DirName(Ⱍ盂콷튇)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄)), FileIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(./..溋캉秣)),DirName(⿤ե虒)),DirName(/₥顉鯮酩՗)),DirName(ᮎ.危뻉棽镀ꚵ/)),DirName(ࣴ.)),DirName(텸)),DirName(䘺ꆹ剺醫倸晵/✩蠻鮸)),DirName(㔅)),DirName(픺/竄멲)),DirName(椸륁촙仓)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄)), FileIn(DirIn(Current,DirName(ꁵꥇﰍ)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄))))' is Right but Set(FileIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(/)),DirName(암腲嗐怞ꐫ)),DirName(넽匘䃒䃰坈)),DirName(ᯌ)),DirName(שּׁ軇ᨮ聿ﻃ㽠)),DirName(칁İ)),DirName(Ṓ뮰.ㄙ搢M.慨)),DirName(Ⱍ盂콷튇)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄)), FileIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(./..溋캉秣)),DirName(⿤ե虒)),DirName(/₥顉鯮酩՗)),DirName(ᮎ.危뻉棽镀ꚵ/)),DirName(ࣴ.)),DirName(텸)),DirName(䘺ꆹ剺醫倸晵/✩蠻鮸)),DirName(㔅)),DirName(픺/竄멲)),DirName(椸륁촙仓)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄)), FileIn(DirIn(Current,DirName(ꁵꥇﰍ)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄)))
[error]        must not contain: FileIn(DirIn(Current,DirName(ꁵꥇﰍ)),FileName(Ά㕜钑뱤諲ﰾ嘵በ횄)) (QueryFileSpec.scala:15)
```

jedesah:
I don't think it's a problem just with `descendantFiles` because I have had to do the same thing with many tests in web in order to make them consistently pass.

If we think it's worth spending time on, I am happy to try and figure out the cause. Originally, adding property based testing to the web project, I thought it was just pre-existing behavior that we never exercised in our tests, but @mossprescott's comment is leading me to think it may have been introduced in our migration to Free.

mossprescott:
Not sure what to make of this. There are definitely `/`s in a couple of those paths, which might be problematic for the old `Path` used in the MongoDB backend. But this test is only using the in-memory interpreter, right? I wonder if it's finding bugs in pathy.

wemrysi:
The problem is with the test itself. In the error you posted, `dp` is a prefix of `od` and thus finding all the descendant files in `dp` will include those in `od`. Our "not equal" check isn't strong enough, we need to assert that `dp` and `od` aren't prefixes of each other, I think.

wemrysi:
@jedesah I realize I didn't tag you in my comment yesterday, doing so now for notification purposes.

jedesah:
Superseded by #1041.

PR #1038, opened 2016-01-01
fix ammonite shell

mossprescott:
The dependency was missing. Also updated the start command for the
current version.

I'm sure this was working at one point, but I could not find the dependency anywhere in the history. I find it useful occasionally, so here's a fix that also brings it up to the current version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -77,8 +77,8 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
   concurrentRestrictions in Global := Seq(
     Tags.exclusive(ExclusiveTest)
   ),
+
   console <<= console in Test, // console alias test:console
-  initialCommands in (Test, console) := """ammonite.repl.Repl.run("prompt.update(\"λ \")")""",
 
   scalazVersion  := "7.1.4",
   slcVersion     := "0.4",


drostron:
Ammonite was completely removed in #975, but was partially and likely accidentally reintroduced in #986. There was a quick discussion in slack regarding the original removal. I prefer not having to exit both the ammonite repl followed by the std repl to get back to the sbt console. If this could be fixed then ammonite could be a fine choice. @mossprescott, are you referring to ammonite specific features when you mention its usefulness or would the standard repl suffice?

mossprescott:
I just noticed the warning when starting the console. I'll update to remove it since that's what was decided. Thanks!

wemrysi:
Yeah, I thought I fixed it in a PR and then it was removed in master and probably screwed up in the merge of said PR, but I never really use the repl and didn't notice.

wemrysi:
@mossprescott When you said "I'll update to remove it", does "it" refer to the warning or ammonite itself? i.e. did we decide to keep ammonite or not?

mossprescott:
Finally got around to updating this branch. I just removed the offending initialization, so ammonite is now history. Per discussion, the sbt-ammonite plugin seems like a better option and works well when configured from ~/.sbt.

wemrysi:
:+1:

PR #1039, opened 2016-01-05
Add convenience methods on MoveScenario

jedesah:


jedesah:
@sellout I don't know if it's worth "fixing" coverage here as it is so trivial. But if you feel strongly that it's the correct thing to do, I will do it.

sellout:
I’m fine with the coverage – scoverage numbers are actually looking pretty good, so no worries.

sellout:
Ah sorry, didn’t realize I was assigned here. Thought you were just asking about coverage because I’m the one who gets pedantic about that 😆

Anyway. It’s good.

PR #1043, opened 2016-01-07
Additional effect algebras

wemrysi:
Adds algebras for three new, generic, effects: monotonic numeric sequence source, key-value store and failure, mostly extracted from @mossprescott's work in #1040. This is just to make the new algebras available to others and avoid blocking on that PR.

### DIFF
--- a/core/src/main/scala/quasar/effect/Failure.scala
+++ b/core/src/main/scala/quasar/effect/Failure.scala
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2014 - 2015 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


mossprescott:
Seems fine.

The only thing I'd like to see is a comment for each algebra and term explaining what they're for, and in particular the requirements for implementations. I realize some of them seem painfully obvious, but I still think it's worth a few words to give context for these deliberately abstract definitions.

wemrysi:
I agree, oversight on my part for omitting them. Also, headers. Will get all that fixed up shortly.

wemrysi:
@mossprescott updated to address your concerns, build should also succeed now.

mossprescott:
LGTM

PR #1044, opened 2016-01-07
Add FileExists term to QueryFile algebra

jedesah:
authored with @drostron

### DIFF
--- a/core/src/main/scala/quasar/fs/QueryFile.scala
+++ b/core/src/main/scala/quasar/fs/QueryFile.scala
@@ -67,6 +67,9 @@ object QueryFile {
   final case class ListContents(dir: ADir)
     extends QueryFile[FileSystemError \/ Set[Node]]
 
+  final case class FileExists(file: AFile)
+    extends QueryFile[FileSystemError \/ Boolean]
+
   @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NonUnitStatements"))
   final class Ops[S[_]](implicit S0: Functor[S], S1: QueryFileF :<: S)
     extends LiftedOps[QueryFile, S] {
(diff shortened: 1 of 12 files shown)


core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -67,6 +67,9 @@ object QueryFile {
   final case class ListContents(dir: ADir)
     extends QueryFile[FileSystemError \/ Set[Node]]
 
+  final case class FileExists(file: AFile)
+    extends QueryFile[FileSystemError \/ Boolean]
```
wemrysi: Can this just return a `Boolean`? It seems like it should be total over all `AFile` inputs.

jedesah: Not every `AFile` is valid as far as certain backends are concerned. For instance, MongoDB puts a limit of 120 on the length of a Path. We need this type in order to convey that kind of an error to the client of the API.

wemrysi: Can we conveying that fact in the output type of the interpreter for mongo, rather than here in the abstraction?

jedesah: We could, and I see how that would be more elegant in some ways. At the same time, we would have to have the same semantic for `ListContents` which currently returns Mongo specific errors in the term result. Also, by adding this possibility for an error at the MongoDB interpreter level, we are adding an extra layer of error handling that may or may not apply to certain other terms of the algebra.

drostron:
👍🏽

drostron:
Ult once @wemrysi signs off.

core/src/main/scala/quasar/fs/QueryFile.scala
wemrysi: Another way to look at this, if a path is invalid for a particular implementation, then could a file ever exist at that path?

jedesah: No. But as a user I would prefer to be informed that the path I supplied is an invalid MongoDB path and will never ever have anything that lives there rather than a `false` which only conveys that there is nothing there right now.

wemrysi:
:+1:

PR #1045, opened 2016-01-08
Add file path to ReadHandle and WriteHandle

wemrysi:
As suggested by @jedesah, this adds some useful context to the handles which is useful both for debugging and for nicer log/error messages.


jedesah:
LGTM :+1:

PR #1046, opened 2016-01-09
Add knowledge of FileExists to view interpreter

jedesah:


jedesah:
Also, in theory, should fix the build.

core/src/test/scala/quasar/fs/TraceFS.scala
```diff
@@ -28,7 +30,11 @@ object TraceFS {
           case More(handle)         => \/-(Vector.empty)
           case Close(handle)        => ()
           case Explain(lp)          => (Vector.empty, \/-(ExecutionPlan(FsType, lp.toString)))
-          case ListContents(dir)    => \/-(nodes.get(dir).getOrElse(Set()))
+          case ListContents(dir)    => \/-(ls(dir))
+          case FileExists(file)     =>
+            val relevantNodes = ls(fileParent(file))
+            val rFile = file1[Sandboxed](fileName(file))
```
mossprescott: This is kind of a drag; when I needed it in ViewSpec I got away with `p(parentDir(p).get).get`, which I wouldn't recommend here, obviously.

I think we should add something like `relativeToParent[S]: Path[_, File, S] => Path[Rel, File, S]` to pathy, but I won't hold up this fix for it.

core/src/test/scala/quasar/fs/view.scala
```diff
@@ -275,11 +279,39 @@ class ViewFSSpec extends Specification {
...
+    "return true is there is a view at that path" ! prop { (file: AFile, lp: Fix[LogicalPlan]) =>
+      val program = query.fileExists(file).run
+
+      val ops = Vector(Inject[QueryFile, FSAction].inj(QueryFile.FileExists(file)))
+
+      val expected = (ops, true.right)
+
+      viewInterp(Views(Map(file -> lp)), Map(),program) must_== expected
+    }
+  }
```
mossprescott: I'm OK with either this property-based style or the other style, but I'd rather we were consistent. Again, not worth holding up the fix.

mossprescott:
LGTM. Sorry I didn't notice this was up until just now.

core/src/test/scala/quasar/fs/TraceFS.scala
jedesah: Agreed. We should add a function for this since it does seem to come up.

core/src/test/scala/quasar/fs/view.scala
jedesah: I think it's okay to mix and match properties and one off tests as most appropriate but it is true that probably most tests in this file could be expressed in terms of a property, so we should choose to do so or not. Unsurprisingly, my vote is to write as many property based tests as possible and make sure and tweak them so that they do not add significant runtime to the test suite unless it's for a very good reason.

jedesah:
Btw, I have no idea why codecov is saying -0.50% code coverage on this one. I looked at the "diff" and I do not see where the supposed additional 42 miss'  are coming from.

PR #1047, opened 2016-01-11
Readable TraceFS logs

mossprescott:
Note: this is based on @jedesah's PR to fix the build. Will rebase on master as soon as that's merged, just to be safe.

After burning a bunch of time trying to get RenderTree to work with the quantified type we were using for `Trace`, I realized it gets a lot simpler if we just render *before* accumulating the values in Writer, so here's an implementation of that, plus some slight improvements to the view tests following on JR's addition.

### DIFF
--- a/core/src/main/scala/quasar/fs/WriteFile.scala
+++ b/core/src/main/scala/quasar/fs/WriteFile.scala
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package quasar
-package fs
+package quasar.fs
 
 import quasar.Predef._
+import quasar._, RenderTree.ops._
 import quasar.effect.LiftedOps
 import quasar.fp._
 
@@ -229,4 +229,15 @@ object WriteFile {
     implicit def apply[S[_]](implicit S0: Functor[S], S1: WriteFileF :<: S): Unsafe[S] =
       new Unsafe[S]
   }
+
+  implicit def RenderWriteFile[A] =
+    new RenderTree[WriteFile[A]] {
+      def render(wf: WriteFile[A]) = wf match {
+        case Open(file)           => NonTerminal(List("Open"), None, List(file.render))
+        case Write(handle, chunk) =>
+          NonTerminal(List("Read"), handle.toString.some,
+            chunk.map(d => Terminal(List("Data"), d.toString.some)).toList)
+        case Close(handle)        => Terminal(List("Close"), handle.toString.some)
+      }
+    }
 }
--- a/core/src/test/scala/quasar/fs/view.scala
+++ b/core/src/test/scala/quasar/fs/view.scala
@@ -72,7 +72,7 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
                 EitherT.right(query.unsafe.close(h)))
       } yield ()).run.run
 
-      viewInterp(views, DefaultNodes, f)._1 must beTree(traceInterp(exp)._1)
+      viewInterp(views, Map(), f)._1 must beTree(traceInterp(exp, Map())._1)
     }
 
     "translate limited read to query" in {
@@ -101,7 +101,7 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
                 EitherT.right(query.unsafe.close(h)))
       } yield ()).run.run
 
-      viewInterp(views, DefaultNodes, f)._1 must beTree(traceInterp(exp)._1)
+      viewInterp(views, Map(), f)._1 must beTree(traceInterp(exp, Map())._1)
     }
 
     "read from closed handle (error)" in {
@@ -116,7 +116,7 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
         _ <- read.unsafe.read(h)
       } yield ()).run
 
-      viewInterp(views, DefaultNodes, f)._2 must_== -\/(UnknownReadHandle(ReadFile.ReadHandle(p, 0)))
+      viewInterp(views, Map(), f)._2 must_== -\/(UnknownReadHandle(ReadFile.ReadHandle(p, 0)))
     }
 
     "double close (no-op)" in {
@@ -131,7 +131,7 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
         _ <- EitherT.right(read.unsafe.close(h))
       } yield ()).run
 
-      viewInterp(views, DefaultNodes, f)._2 must_== \/-(())
+      viewInterp(views, Map(), f)._2 must_== \/-(())
     }
   }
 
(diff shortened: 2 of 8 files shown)


core/src/main/scala/quasar/fs/WriteFile.scala
```diff
@@ -229,4 +229,15 @@ object WriteFile {
...
       new Unsafe[S]
   }
+
+  implicit def RenderWriteFile[A] =
+    new RenderTree[WriteFile[A]] {
+      def render(wf: WriteFile[A]) = wf match {
+        case Open(file)           => NonTerminal(List("Open"), None, List(file.render))
+        case Write(handle, chunk) =>
+          NonTerminal(List("Read"), handle.toString.some,
+            chunk.map(d => Terminal(List("Data"), d.toString.some)).toList)
```
jedesah: This seems a little odd to me that we would create a `Terminal` for each `Data` in the `Write` term, but I can't really think of a better way to do it.

core/src/test/scala/quasar/fs/view.scala
```diff
@@ -28,24 +30,27 @@ class ViewFSSpec extends Specification {
...
+        interpretFileSystem[Trace](qfTrace(nodes), rfTrace, wfTrace, mfTrace))
 
-  def viewInterp[A](views: Views, t: Free[FileSystem, A]): (Vector[FSAction[_]], A) =
+  def viewInterp[A](views: Views, nodes: Map[ADir, Set[Node]], t: Free[FileSystem, A]): (Vector[RenderedTree], A) =
     (t flatMapSuspension view.fileSystem[ViewFileSystem](views))
-      .foldMap(traceViewFs)
+      .foldMap(traceViewFs(nodes))
       .eval(VS(0, Map.empty)).run
 
+  implicit val RenderedTreeRenderTree = new RenderTree[RenderedTree] {
```
jedesah: I am one to talk since I added a `arbLogicalPlan` just in this file, but that was because I was not sure where to put it and I did not feel like I had the time to make it more generic. I think though that this `RenderTreeRenderTree` should probably live in `RenderTreeInstances`.

core/src/test/scala/quasar/fs/view.scala
```diff
@@ -157,7 +162,7 @@ class ViewFSSpec extends Specification {
 
       val f = manage.move(FileToFile(viewPath, otherPath), Overwrite).run
 
-      viewInterp(views, f) must_==(
+      viewInterp(views, Map(), f) must_==(
         (Vector.empty,
           -\/(FileSystemError.PathError(PathError2.InvalidPath(viewPath, "cannot move view")))))
```
jedesah: Just curious, why can't we move a view?

core/src/test/scala/quasar/fs/view.scala
```diff
@@ -171,7 +176,7 @@ class ViewFSSpec extends Specification {
 
       val f = manage.move(FileToFile(otherPath, viewPath), Overwrite).run
 
-      viewInterp(views, f) must_==(
+      viewInterp(views, Map(), f) must_==(
         (Vector.empty,
           -\/(FileSystemError.PathError(PathError2.InvalidPath(viewPath, "cannot move file to view location")))))
```
jedesah: Same question I guess, why is view so special?

jedesah: Oh I think I might know the answer. Is it because all operations that modify a view in some fashion should be done through the mount algebra?

jedesah:
A few general comments that this PR made me think about.

First off, @mossprescott seeing your approach to testing `Free` made me realize that the `Interpreter` class that I added a while back is much less useful than I first thought. Instead of using it, you just call `run` as many times as necessary to get out of all of the `MonadTransformers` and then just `foldMap` on that. It seems obvious now, but I did not think of that which is why most of the functions in the `Interpreter` class "work around" the monad transformers and produce the result back in the original monad transformer stack. I think this could be relatively useful outside of tests, but in tests, you typically end up decomposing the monad transformer stack after that anyway, so what's the point?

There is one way that the `Interpreter` class seems still rather useful for tests and that's when dealing with `Process`. I noticed that most of the tests for views use the unsafe api. I wonder if that is to avoid the annoyance of `Process` with `Free` or is it to have more control over which underlying operations are being performed since we are opting for a tracing approach. If it's the first reason, it might be worth considering using the safe api along with the `Interpreter` class.

Speaking of the approach to "trace" the operations that are actually happening, I think it's totally fine, but it's worth noting that it is not the approach that we have taken so far in quasar-advanced for testing interpreters that "augment" a program. Instead we have been simply checking that the result is the same as produced by the original interpreter. I think both approaches are fine and have their strenghts, but it's worth being aware that we are taking these two separate approaches to solving a similar problem.

core/src/main/scala/quasar/fs/WriteFile.scala
mossprescott: For the purposes of the current tests, it could as easily be something like "Data(10 rows)", but this seemed like it might be useful at some point.

core/src/test/scala/quasar/fs/view.scala
mossprescott: This is a pretty weird thing to need, and just results from the unusual `Vector[RenderedTree]` type. We don't put these trees inside collections anywhere else, so I don't think there's much point on making it available. On top of that, I worry a little about confusion or even circularities in the conversions we make available, since this thing is effectively RenderTeee.identity.

core/src/test/scala/quasar/fs/view.scala
mossprescott: Right. A view can't be moved via `ManageFile`, in the same way that you can't move a directory which is a mount point. The only difference is that view paths *can* be nested inside other mount paths, but (for now) non-view mounts can't be.

mossprescott:
I think Interpreter's methods might still be useful in general, but in these tests I happen to know the exact stack and can run the values directly. One approach doesn't invalidate the other, I think.

Similarly, I think both ways of writing assertions are useful. Because the view interpreter for the most part just modifies the *inputs* (queries and paths) and not the *outputs* (results), it made sense to me to test by verifying the rewritten terms and not bothering to actually handle them in a serious way. This came out of the experience of writing the old view tests, which had no choice but to run against a real backend and verify the results, which turned out to be cumbersome. Anyway, I think the result is good, in that the tests take the form "operations x, y, z on these views are equivalent to operations a, b, c, d on the filesystem", which corresponds well to the way I think about views being defined.

core/src/test/scala/quasar/fs/view.scala
jedesah: I think this is a pretty standard thing with typeclasses when it makes sense. For instance Argonaut has a `EncodeJson[Json]`. I am not too worried about circularities unless we provide an implicit conversion, which I don't think we do.

That being said, I am totally okay with waiting to see if we need it somewhere else (hopefully we will remember it's lying around here).

jedesah:
LGTM :+1:

PR #1048, opened 2016-01-11
SD-1181 Hierarchical FileSystem

wemrysi:
Adds a filesystem interpreter that, given a set of mounts, routes filesystem requests to underlying interpreters based on the paths in the requests.

### DIFF
--- a/core/src/main/scala/quasar/effect/KeyValueStore.scala
+++ b/core/src/main/scala/quasar/effect/KeyValueStore.scala
@@ -20,7 +20,7 @@ import quasar.Predef._
 import quasar.fp.TaskRef
 
 import monocle.Lens
-import scalaz.{Lens => _, _}, Scalaz._
+import scalaz.{Lens => _, _}
 import scalaz.concurrent.Task
 
 /** Provides the ability to read, write and delete from a store of values
@@ -63,8 +63,10 @@ object KeyValueStore {
       new Ops[K, V, S]
   }
 
-  def taskRefKeyValueStore[A, B](initial: Map[A, B]): Task[KeyValueStore[A, B, ?] ~> Task] = {
-    TaskRef(initial).map(ref =>
+  def taskRefKeyValueStore[A, B](
+    initial: Map[A, B]
+  ): Task[KeyValueStore[A, B, ?] ~> Task] =
+    TaskRef(initial) map { ref =>
       new (KeyValueStore[A, B, ?] ~> Task) {
         def apply[C](fa: KeyValueStore[A, B, C]): Task[C] = fa match {
           case Put(key, value) =>
@@ -74,20 +76,39 @@ object KeyValueStore {
           case Delete(key) =>
             ref.modifyS(m => ((m - key), ()))
         }
-      })
-  }
-
-  def stateKeyValueStore[F[_]: Monad, K, V, S](l: Lens[S, Map[K, V]]) =
-    new (KeyValueStore[K, V, ?] ~> StateT[F, S, ?]) {
-      def apply[A](fa: KeyValueStore[K, V, A]): StateT[F, S, A] = fa match {
-        case Put(key, value) =>
-          StateT[F, S, A](s => (l.modify(_ + (key -> value))(s), ()).point[F])
-
-        case Get(key) =>
-          StateT.stateTMonadState[S, F].gets(s => l.get(s).get(key))
-
-        case Delete(key) =>
-          StateT.stateTMonadState[S, F].modify(s => l.modify(_ - key)(s))
       }
     }
+
+  /** Returns an interpreter of `KeyValueStore[K, V, ?]` into `StateT[F, S, ?]`,
+    * given a `Lens[S, Map[K, V]]`.
+    *
+    * NB: Uses partial application of `F[_]` for better type inference, usage:
+    *
+    *   `stateKeyValueStore[F](lens)`
+    */
+  object stateKeyValueStore {
+    def apply[F[_]]: Aux[F] =
+      new Aux[F]
+
+    final class Aux[F[_]] {
+      type ST[S, A] = StateT[F, S, A]
+
+      def apply[K, V, S](l: Lens[S, Map[K, V]])(implicit F: Monad[F])
+                        : KeyValueStore[K, V, ?] ~> ST[S, ?] = {
+        new(KeyValueStore[K, V, ?] ~> ST[S, ?]) {
+          def apply[A](fa: KeyValueStore[K, V, A]): ST[S, A] = fa match {
+            case Put(key, value) =>
+              st.modify(l.modify(_ + (key -> value)))
+
+            case Get(key) =>
+              st.gets(s => l.get(s).get(key))
+
+            case Delete(key) =>
+              st.modify(l.modify(_ - key))
+          }
+          val st = MonadState[ST, S]
+        }
+      }
+    }
+  }
 }
--- a/core/src/main/scala/quasar/effect/MonotonicSeq.scala
+++ b/core/src/main/scala/quasar/effect/MonotonicSeq.scala
@@ -20,8 +20,9 @@ import quasar.Predef._
 import quasar.fp.TaskRef
 
 import monocle.Lens
-import scalaz.{Lens => _, _}, Scalaz._
+import scalaz.{Lens => _, _}
 import scalaz.concurrent.Task
+import scalaz.syntax.applicative._
 
 /** Provides the ability to request the next element of a monotonically
   * increasing numeric sequence.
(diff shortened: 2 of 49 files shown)


core/src/main/scala/quasar/effect/KeyValueStore.scala
```diff
@@ -74,20 +76,32 @@ object KeyValueStore {
           case Delete(key) =>
             ref.modifyS(m => ((m - key), ()))
         }
-      })
-  }
+      }
+    }
+
+  object stateKeyValueStore {
+    def apply[F[_]]: Aux[F] =
```
mossprescott: I like this trick, but I also think it might be worth a comment to clarify how it works. As in, "call it like this: `stateKeyValueStore[F](mapL)`"

core/src/main/scala/quasar/effect/LiftedOps.scala
```diff
@@ -25,5 +27,5 @@ abstract class LiftedOps[G[_], S[_]: Functor](implicit S: Coyoneda[G, ?] :<: S)
   type F[A] = Free[S, A]
```
mossprescott: Since I got confused by this earlier, I wonder if a longer name would be more clear. `FreeS[A]`?

core/src/main/scala/quasar/effect/MonotonicSeq.scala
```diff
@@ -57,17 +58,24 @@ object MonotonicSeq {
...
       })
   }
 
-  def stateMonotonicSeq[F[_]: Applicative, S](l: Lens[S, Long]) =
-    new (MonotonicSeq ~> StateT[F, S, ?]) {
-      def apply[A](fa: MonotonicSeq[A]): StateT[F, S, A] = fa match {
-        case Next =>
-          StateT[F, S, A](s => (l.modify(_+1)(s), l.get(s)).point[F])
+  object stateMonotonicSeq {
+    def apply[F[_]]: Aux[F] =
```
mossprescott: As above.

core/src/main/scala/quasar/fp/package.scala
```diff
@@ -467,4 +468,19 @@ package object fp extends TreeInstances with ListMapInstances with EitherTInstan
     new (F ~> G) {
       def apply[A](fa: F[A]) = I inj fa
     }
+
+  /** Lift a `State` computation to operate over a "larger" state given a `Lens`. */
+  object zoomNT {
```
mossprescott: Here again, and just clarify, the problem I have with this is that you use it like a function `Lens[A, B] => (StateT[F, A, ?] ~> StateT[F, B, ?])`, but that signature is hard to pick out of this series of definitions.

core/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -14,121 +14,98 @@
...
+  final case class UnknownResultHandle private[fs] (h: ResultHandle)
+    extends FileSystemError
+  final case class UnknownReadHandle private[fs] (h: ReadHandle)
+    extends FileSystemError
+  final case class UnknownWriteHandle private[fs] (h: WriteHandle)
+    extends FileSystemError
+  final case class PartialWrite private[fs] (numFailed: Int)
+    extends FileSystemError
+  final case class WriteFailed private[fs] (data: Data, reason: String)
+    extends FileSystemError
```
mossprescott: So we're just defining Prisms now and not these capitalize constructors? Sounds good to me.

core/src/main/scala/quasar/fs/mounted.scala
```diff
@@ -0,0 +1,209 @@
...
+
+import monocle.Optional
+import monocle.function.Field1
+import monocle.std.tuple2._
+import pathy.{Path => PPath}, PPath._
+import scalaz._
+import scalaz.std.tuple._
+import scalaz.syntax.functor._
+
+object mounted {
```
mossprescott: What exactly is the relationship between this and `chroot`? Is it a different mapping of paths? Could that be factored out? Or is it more subtle than that?

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -18,19 +18,18 @@ package quasar
 
 import quasar.Predef._
 import quasar.fp.free._
-import quasar.fp.TaskRef
 import quasar.effect._
 
-import scalaz._
+import monocle.Lens
+import pathy.{Path => PPath}, PPath._
```
mossprescott: I guess I'd rather see pathy's `Path` used as is and rename `quasar.Path` to `QPath` as we've done elsewhere (not a huge fan of `EnginePath`). That will mean fewer changes down the road when we rip it the rest of the way out.

mossprescott:
I see no major concerns with this. I would be OK with merging this as is in the interest of moving on to the next steps.

core/src/main/scala/quasar/fs/mounted.scala
wemrysi: They're inverses of each other, `mounted(dir) compose chroot(dir) == id`. And yes, I think we could factor out the common code into something both could reuse. Something like a set of path transformation functions, one per algebra.

core/src/main/scala/quasar/fs/FileSystemError.scala
wemrysi: Yeah, this is where we (well, @sellout and I at least, with no objections from anyone else) last landed on sum-type encodings.

core/src/main/scala/quasar/effect/KeyValueStore.scala
wemrysi: Yeah, that's a good point. The partial application does obfuscate things, I'll add comments.

core/src/main/scala/quasar/mount/Mounts.scala
```diff
@@ -0,0 +1,89 @@
...
+import pathy.Path._
+import scalaz._, Scalaz._
+
+/** A mapping of values to directory paths, maintains the invariant that no
+  * path is a prefix of any other path.
+  *
+  * The current implementation is linear in the number of mounts, might be able
+  * to do better using a bintree given an `Order` on `ADir`.
+  */
+final class Mounts[A](val toMap: Map[ADir, A]) {
```
mossprescott: The invariant is only maintained if you use `empty` and `add` to build up a `Mounts` instance. since the case class is public, it would be easy to create one that violates it, even accidentally.

If we think this invariant is important, should we make it a trait and provide *only* `empty`?

And I still think there's a more generic ADT hiding in here ("NonconflictingMap"?), where the predicate on the keys is a parameter.

core/src/main/scala/quasar/effect/LiftedOps.scala
wemrysi: Something more evocative would probably be more clear, though I'd prefer it to be more "what" and less "how". i.e. what do we call the free monad generated by the functor `S[_]`, which includes the effect `G[_]`?

core/src/main/scala/quasar/mount/Mounts.scala
wemrysi: Good catch! It should be `final class Mounts[A] private (val toMap: Map[ADir, A]) { ... }`, that was my intent, just forgot.

And yes, this type more-or-less wraps `(Map[A, B], (A, A) => String \/ Unit)` where `A = ADir`.

core/src/main/scala/quasar/effect/LiftedOps.scala
mossprescott: How about `Lifted`? That would make it `def lift[A](ga: G[A]): Lifted[A]`, but I admit it's not so great when you see it in `Ops` (e.g. `def get(k: K): OptionT[Lifted, V]`).

core/src/main/scala/quasar/fs/package.scala
wemrysi: I agree with you, but I'm not sure how to do so in this instance as the code is in the `fs` package object. I'm not sure if it is possible to rename the import in the package object itself as the `Path` type is just available from the ambient environment, we didn't import it explicitly.

core/src/main/scala/quasar/effect/LiftedOps.scala
wemrysi: I dunno, that's not very satisfying either. Mind if we move this to its own PR/discussion/issue? It affects all of our `Ops` classes and, while I think it is valuable to resolve, isn't really related to this PR and any change is going to affect types across the codebase.

core/src/main/scala/quasar/fs/package.scala
mossprescott: I see. I suppose we could move `Path` to `quasar.deprecated` or something like that, but I'd rather we put that energy into deleting it.

core/src/main/scala/quasar/effect/LiftedOps.scala
mossprescott: Agreed.

core/src/main/scala/quasar/fs/mounted.scala
mossprescott: Isn't it just two functions (here, `stripPrefixA` and `rebasePath`), assuming mapping over errors is a separate issue? 

It looks that way, in which case I'd say it improves clarity to pull them out, whether or not the two interpreters get refactored into one.

wemrysi:
@mossprescott Thanks for the push to refactor `chroot` and `mounted`, much cleaner now.

PR #1049, opened 2016-01-12
Feature/translate web api for free

mossprescott:
As far as I can tell, all the changes affecting the moved/split files had already been merged, so this PR just brings the branch up-to-date with master, and includes a couple of very minor fixes and comments.

This will adds *all* the commits from master to the branch; I assume we'll just squash it all down to one commit before merging back to master. How does that sound?


mossprescott:
Hang on, let me try that again from a proper topic branch...

PR #1050, opened 2016-01-12
Update to latest

mossprescott:
As far as I can tell, all the changes affecting the moved/split files had already been merged, so this PR just brings the branch up-to-date with master, and includes a couple of very minor fixes and comments.

This will add *all* the commits from master to the branch; I assume we'll just squash it all down to one commit before merging back to master. How does that sound?


wemrysi:
Aye, that sounds fine.

:+1:

mossprescott:
Gonna let this build again for a couple more missing headers. Looks like the "restart" tests are still failing only in Travis, but I'll probably merge anyway if it otherwise looks good.

wemrysi:
Ah, that must be what SD-1167 refers to.

PR #1051, opened 2016-01-12
Reorganize mount-related files into 'mount' package under 'fs'

wemrysi:
This is an organizational PR that

1. Moves the `mount` package under the `fs` package
2. Moves view-related files to the `mount` package
3. Moves hierarchical-related files to the `mount` package


mossprescott:
LGTM

PR #1052, opened 2016-01-14
Path related additions

jedesah:
- Move function that extracts paths from LogicalPlan from MongoDB interpreter into fs.package
- Add functions to convert from fs.Path to path.Path and vice-versa


wemrysi:
@jedesah There are some similar conversion functions on `quasar.fs.Path`, is it possible to use those? Or at least to add whatever you need as methods on `Path` or members of its companion? Then we don't have to remember to delete them when we delete that file.

wemrysi:
@jedesah The functions added to the `Interpreter` class don't seem to have anything to do with paths. Might they be better served by their own PR?

jedesah:
@wemrysi Ok, I have created #1054 for the `Interpreter` functions and @drostron will open a new PR with the pathy related changes along with more pathy related changes we had to work through today.

PR #1056, opened 2016-01-18
SD-1244 Handle mount requests

wemrysi:
Adds mount/unmount handlers that ultimately manage a composite `FileSystem` interpreter that composes views and mounts of physical interpreters together, making the result available via the `AtomicRef` effect.


core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  val fsMounter = FileSystemMounter(fsDef)
+  val testType = FileSystemType("test")
+  val testUri = ConnectionUri("https://test.example.com")
+  def mount(d: ADir) = fsMounter.mount[Eff](d, testType, testUri)
+  val unmount = fsMounter.unmount[Eff] _
+
+  "FileSystemMounter" should {
+    "mounting" >> {
+      "fails when mounting above an exsiting mount" >> {
```
drostron: Interesting that this differs from `MountingSpec`. What is the background and would it be desirable to have a `ViewMountingSpec`?

wemrysi: It differs as the implementation is more restrictive than the abstraction. There is no reason at the `Mounting` abstraction layer that one filesystem mount point can't be a prefix of another, but the current implementation imposes this additional restriction.

As for a `ViewMountingSpec`, it's probably desirable in order to test that views are added/removed from the managed `Views` object, but all the really interesting work is handled by `quasar.queryPlan` which is tested elsewhere.

core/src/main/scala/quasar/fs/mount/Mounter.scala
```diff
@@ -0,0 +1,106 @@
...
+    def unmount0(req: MountRequest): Mnt[Unit] =
+      free.lift(unmount(req)).into[S]
+
+    def handleRequest(req: MountRequest): MntE[Unit] = {
+      val failIfExisting: MntE[Unit] =
+        mountConfigs.get(req.path).as(pathExists(req.path)).toLeft(())
+
+      failIfExisting                                        *>
+      mount0(req)                                           *>
+      merr.handleError(failIfExisting)(e =>
```
jedesah: I don't understand what is going on here. Why are we checking if the path exists after mounting? Don't we expect it to exist?

wemrysi: The check is an attempt to check that nothing else successfully mounted at the path that is in the process of being mounted here, i.e.

1. We check that path, P, doesn't exist in the current set of mounts
2. We call the mount handler to perform the mount
3. On success, we update the set of mounts with P

what if somebody else mounts P in between 1 and 3? If we don't check, we'll just overwrite their mount (though this still isn't perfect, as the following comment mentions).

core/src/main/scala/quasar/fs/mount/Mounter.scala
```diff
@@ -0,0 +1,106 @@
...
+  def apply[F[_], S[_]](
+    mount: MountRequest => F[MountingError \/ Unit],
+    unmount: MountRequest => F[Unit]
+  )(implicit
+    S0: Functor[S],
+    S1: F :<: S,
+    S2: MountConfigsF :<: S
+  ): Mounting ~> Free[S, ?] = {
+
+    type Mnt[A]    = Free[S, A]
```
jedesah: Can we come up with some kind of convention for a type alias for `Free[S,?]`. Most places it's `F` but here it's `Mnt`. My vote would be to use `FreeS` everywhere because if it were not for Scala's imperfect inference, I would vote against a type alias for this, so `FreeS` kind of conveys that it's only there for type inference.

core/src/main/scala/quasar/fs/mount/Mounter.scala
```diff
@@ -0,0 +1,106 @@
...
+    * supplied functions to handle mount and unmount requests.
+    *
+    * @param mount   Called to handle a mount request.
+    * @param unmount Called to handle the removal/undoing of a mount request.
+    */
+  def apply[F[_], S[_]](
+    mount: MountRequest => F[MountingError \/ Unit],
+    unmount: MountRequest => F[Unit]
+  )(implicit
+    S0: Functor[S],
```
jedesah: I forget the name for this in Scala, but I think `Functor` should be on line 36. Makes it easier to read IMHO.

drostron:
👍🏽

wemrysi:
@drostron @jedesah Pushed a couple commits to address your comments.

drostron:
👍🏽on the additional commits.

core/src/main/scala/quasar/effect/KeyValueStore.scala
```diff
@@ -63,52 +104,57 @@ object KeyValueStore {
...
+    final class Aux[ST[_, _]] {
+      def apply[K, V, S](l: Lens[S, Map[K, V]])(implicit ST: MonadState[ST, S])
+                        : KeyValueStore[K, V, ?] ~> ST[S, ?] =
         new(KeyValueStore[K, V, ?] ~> ST[S, ?]) {
           def apply[A](fa: KeyValueStore[K, V, A]): ST[S, A] = fa match {
-            case Put(key, value) =>
-              st.modify(l.modify(_ + (key -> value)))
+            case CompareAndPut(k, expect, update) =>
+              lookup(k) flatMap { cur =>
+                if (cur == expect)
```
jedesah: Are we not using `===` here to avoid asking for an `Equal` on `V`?

core/src/main/scala/quasar/effect/KeyValueStore.scala
```diff
@@ -63,52 +104,57 @@ object KeyValueStore {
...
-        }
-      }
+  /** Returns an interpreter of `KeyValueStore[K, V, ?]` into `Task`, given a
+    * `TaskRef[Map[K, V]]`.
+    */
+  def fromTaskRef[K, V](ref: TaskRef[Map[K, V]]): KeyValueStore[K, V, ?] ~> Task =
+    new (KeyValueStore[K, V, ?] ~> Task) {
+      val toST = toState[State](Lens.id[Map[K,V]])
+      def apply[C](fa: KeyValueStore[K, V, C]): Task[C] =
+        ref.modifyS(toST(fa).run)
```
jedesah: I love the reusability here! I guess it's not a huge issue that we have two layers attempting to guarantee atomicity, it just means `State` is doing a little extra work for no reason.

core/src/main/scala/quasar/effect/KeyValueStore.scala
wemrysi: Yeah, and for consistency with our use of `Map`. i.e. if we're going to require `Equal[V]` we should also require `Order[K]` or `Hashable[K]`. We probably want to do this eventually, but we're using `Map` everywhere else so I just stuck with it here.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
```diff
@@ -0,0 +1,140 @@
...
+/** Handles mount requests, managing a composite `FileSystem` interpreter
+  * into a base effect type `F`, adding support for views and hierarchical
+  * mounts.
+  *
+  * @tparam F the base effect that `FileSystem` operations are translated into
+  * @tparam S the composite effect, supporting the base, view and hierarchical effects
+  */
+final class EvaluatorMounter[F[_], S[_]](
+  fsDef: FileSystemDef[F]
+)(implicit S0: Functor[S],
```
jedesah: I really do think this should a context bound. If it's not, at least, it should not be called `S0`. It makes it hard to distinguish it from the rest of the implicits which are listing the required components of `S`. IMHO.

wemrysi: Aye, that's fine, I fixed the other one you mentioned, this one just slipped by.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
```diff
@@ -0,0 +1,140 @@
...
+           S1: F :<: S,
+           S2: MonotonicSeqF :<: S,
+           S3: ViewStateF :<: S,
+           S4: MountedResultHF :<: S,
+           S5: HFSFailureF :<: S) {
+
+  import MountRequest._, FileSystemDef.DefinitionResult
+
+  type EvalFS[A]     = Free[S, A]
```
jedesah: What's the rationale behind naming it this way as opposed to some other name that would be uniform across our codebase?

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
jedesah: :+1:

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
wemrysi: "What" vs "how". The fact that this happens to be a free monad isn't needed/used in the rest of the file, we only use the fact that it is a monad. I.e. we could change the rhs to another monad and the rest of the code would not change.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  def mount[T[_]: Functor]
+      (req: MountRequest)
+      (implicit T0: F :<: T,
+                T1: fsMounter.MountedFsF :<: T,
+                T2: MountedViewsF :<: T,
+                T3: EvalFSRefF :<: T)
+      : Free[T, MountingError \/ Unit] = {
+
+    type EvalM[A]    = Free[T, A]
```
jedesah: Granted, this certainly precludes calling it just `F` like we do elsewhere, but I think I would still vote for `FreeS` and `FreeT`.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
jedesah: Okay, I can get behind that. We just need to come up with really good names for these things, or else it makes it very tedious to understand the code by constantly looking up aliases. In this case, I am not clear why it is called `Eval` but that may very well be because I haven't completely wrapped my head around what this class accomplishes, as opposed to a bad name. But still I think we should try and maintain a good balance between concise names for type alias' and names that do not require constantly looking up what they actually are because the name is not sufficient to convey what it is.

jedesah:
Okay, I did my best to try and find any flaws. I am not happy with how little I understand certain parts of this PR, but I am not sure it's worth investing more time right now to try and understand them better to maybe find some things that could be slightly improved.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
wemrysi: I agree on trying to use descriptive names where appropriate. In this case, the alias is named `EvalFS` as it is the monad we evaluate filesystem effects with.

wemrysi:
@jedesah Thanks for your efforts, just pushed one more commit that moves that `Functor` constraint to a context bound and adds a comment to `EvaluatorMounter#updateComposite` as I think that function warrants one.

jedesah:
:+1:

PR #1058, opened 2016-01-20
Add normalize function for Free

jedesah:


wemrysi:
@jedesah Could you give (or link to) some motivation for this function? Curious as to what sort of situations one would use it in.

wemrysi:
Reposting the motivation here for completeness: https://github.com/slamdata/quasar-advanced/pull/12/files#diff-f42ccc5a52742c3ac48baf1630279c0aR44

wemrysi:
Given `quasar.fp.pointNT` is `normalize` necessary?

jedesah:
@wemrysi I don't believe it is necessary anymore. Thx for pointing out the alternative!

PR #1059, opened 2016-01-20
Refactor effect interpreters to use MTL classes when available

wemrysi:
Instead of the default interpreters for effects in `quasar.effect` interpreting into specific monads, we can use the "mtl" classes, like `MonadState` and `MonadError`, where possible for more generality.

### DIFF
--- a/core/src/main/scala/quasar/effect/Failure.scala
+++ b/core/src/main/scala/quasar/effect/Failure.scala
@@ -90,20 +90,12 @@ object Failure {
       new Ops[E, S]
   }
 
-  def toDisjunction[E]: Failure[E, ?] ~> (E \/ ?) =
-    new (Failure[E, ?] ~> (E \/ ?)) {
-      def apply[A](fa: Failure[E, A]): E \/ A = fa match {
-        case Fail(e) => \/.left(e)
+  def toError[F[_, _], E](implicit F: MonadError[F, E]): Failure[E, ?] ~> F[E, ?] =
+    new (Failure[E, ?] ~> F[E, ?]) {
+      def apply[A](fa: Failure[E, A]) = fa match {
+        case Fail(e) => F.raiseError(e)
       }
     }
-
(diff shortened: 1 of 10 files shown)


core/src/main/scala/quasar/effect/Failure.scala
```diff
@@ -90,21 +90,13 @@ object Failure {
...
   }
 
-  def toDisjunction[E]: Failure[E, ?] ~> (E \/ ?) =
-    new (Failure[E, ?] ~> (E \/ ?)) {
-      def apply[A](fa: Failure[E, A]): E \/ A = fa match {
-        case Fail(e) => \/.left(e)
+  def toError[F[_, _], E](implicit F: MonadError[F, E]): Failure[E, ?] ~> F[E, ?] =
+    new (Failure[E, ?] ~> F[E, ?]) {
+      def apply[A](fa: Failure[E, A]) = fa match {
+        case Fail(e) => F.raiseError(e)
```
mossprescott: Nice.

core/src/test/scala/quasar/fs/mount/HierarchicalFileSystemSpec.scala
```diff
@@ -65,11 +66,11 @@ class HierarchicalFileSystemSpec extends mutable.Specification with FileSystemFi
   def runH: F ~> HFSM = {
     val seqNT: MonotonicSeqF ~> HFSM =
       liftMT[MountedFs, HFSErrT].compose[MonotonicSeqF](
-        Coyoneda.liftTF[MonotonicSeq, MountedFs](MonotonicSeq.stateMonotonicSeq[Id](seq)))
+        Coyoneda.liftTF[MonotonicSeq, MountedFs](MonotonicSeq.toState[State](seq)))
```
mossprescott: Really nice.

mossprescott:
LGTM.

mossprescott:
The build failure here is the config test timing out on monogolab, by the way. I'll restart just so we can get a clean report.

wemrysi:
Thanks!

PR #1060, opened 2016-01-20
Organize scalacheck.Arbitrary instances in quasar-core

wemrysi:
Where appropriate, extracts/renames `Arbitrary` instances into their own files for easy a la carte importing and adds aggregate `Arbitraries` objects to enable easy importing of all the instances in a package.

Also adds a few arbitrary instances for types in `quasar.fs.mount`


sellout:
Cool, I like this. It all looks good, but I’d like you to add something about this (an `Arbitraries` trait & object in each package, etc.) to the wiki.

👍🏽

wemrysi:
@sellout I've added a section to the wiki, good idea!

PR #1061, opened 2016-01-21
Implement the /mount API given a `Mounting ~> Task`

mossprescott:
See SD-1142

In `Mounting.Ops.mount`, return errors that result from the ambiguity in that method's argument in a separate error type (PathTypeMismatch), so that MountingErrors will only represent errors from actually interpreting the algebra.

New common stuff:
- `errorResponse()` to format simple error messages in the expected JSON syntax.
- Use that in generating responses for FileSystemError and PathError2.
- `requiredHeader()` to extract headers in a convenient EitherT.
- `resandbox()` for the previously duplicated code to fix up the type of `Abs` paths parsed by the services.

Renamed FileName to XFileName to hopefully reduce confusion and to avoid collisions.

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/Mounting.scala
+++ b/core/src/main/scala/quasar/fs/mount/Mounting.scala
@@ -26,10 +26,7 @@ import quasar.sql._
 import monocle.Prism
 import monocle.std.{disjunction => D}
 import pathy._, Path._
-import scalaz._
-import scalaz.syntax.std.option._
-import scalaz.syntax.apply._
-import scalaz.syntax.monadError._
+import scalaz._, Scalaz._
 
 sealed trait Mounting[A]
 
@@ -45,6 +42,11 @@ object Mounting {
 
   final case class Unmount(path: APath)
     extends Mounting[MountingError \/ Unit]
+
+  /** Indicates the wrong type of path (file vs. dir) was supplied to the `mount`
+    * convenience function.
+    */
+  final case class PathTypeMismatch(path: APath)
 
   @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NonUnitStatements"))
   final class Ops[S[_]](implicit S0: Functor[S], S1: MountingF :<: S)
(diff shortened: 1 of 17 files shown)


core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -46,6 +43,11 @@ object Mounting {
   final case class Unmount(path: APath)
     extends Mounting[MountingError \/ Unit]
 
+  /** Indicates the wrong type of path (file vs. dir) was supplied to the `mount`
+    * convenience function.
+    */
+  final case class PathTypeMismatch(path: APath)
```
wemrysi: Worth extending `AnyVal`?

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -104,10 +104,16 @@ object Mounting {
 
     ////
 
-    private type E[A, B] = EitherT[F, A, B]
+    private type ErrF[A, B] = EitherT[F, A, B]
+
+    def bifold[G[_]: Functor, E, A, EE, AA](v: EitherT[G,E,A])(e: E => EE \/ AA, a: A => EE \/ AA): EitherT[G, EE, AA] =
```
wemrysi: Can we make these helpers private?

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -0,0 +1,111 @@
...
+                    mountingErrorResponse(_).left,
+                    _.leftMap(pathTypeMismatchResponse))))
+      } yield exists
+
+    HttpService {
+      case GET -> AsPath(path) =>
+        M.lookup(path).fold(
+          cfg => Ok(EncodeJson.of[MountConfig2].encode(cfg)),
+          errorResponse(NotFound, "There is no mount point at " + posixCodec.printPath(path)))
+        .foldMap(f).run
```
wemrysi: Is that `Task.run`? If so, can we change to `join`?

core/src/main/scala/quasar/fs/mount/package.scala
```diff
@@ -28,6 +28,14 @@ package object mount {
   type MountingF[A] = Coyoneda[Mounting, A]
   type MntErrT[F[_], A] = EitherT[F, MountingError, A]
 
+  type MountingFileSystem[A] = Coproduct[MountingF, FileSystem, A]
+
+  def interpretMountingFileSystem[M[_]: Functor](
+      m: Mounting ~> M,
```
wemrysi: Looks like some indenting weirdness here.

mossprescott: I was trying to be consistent with what I used in the view interpreters, for example, but it does come off a bit oddly here. Sigh.

web/src/main/scala/quasar/api/services/mount.scala
mossprescott: I suppose it is and I will.

core/src/main/scala/quasar/fs/mount/Mounting.scala
mossprescott: It's worth it if we have a convention that says we should for these newtypes.

core/src/main/scala/quasar/fs/mount/Mounting.scala
mossprescott: Okey doke.

wemrysi:
:+1:

mossprescott:
Rebased and ready to for a final review, but I'm seeing occasional test failures when paths contain certain characters, so I'll be taking a look at that before merging.

wemrysi:
LGTM, :+1:

mossprescott:
That was a fun exercise. Some part of this is a bug in http4s, but really we need to figure out path escaping (which is already broken in master, see [SD-1322](https://slamdata.atlassian.net/browse/SD-1322)). Anyway I was able to put a workaround in the test so it shouldn't fail intermittently.

PR #1064, opened 2016-01-22
separate tests for MOVE file/dir to same path

mossprescott:
Currently 409 Conflict, but should possible be some kind of 200 and a
no-op. Anyway this documents the current behavior and fixes the
intermittent failure.

The diff is unfortunate here. I had to make `expectedBody` a function so that the test for moving a directory to the same path could match the error which can refer to any file in the directory.

### DIFF
--- a/web/src/test/scala/quasar/api/services/DataServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/DataServiceSpec.scala
@@ -5,6 +5,7 @@ package services
 import org.http4s.Uri.Authority
 import org.http4s.server.middleware.GZip
 import org.specs2.ScalaCheck
+import org.specs2.execute.AsResult
 import org.specs2.mutable.Specification
 import pathy.scalacheck.AbsFileOf
 import quasar.Data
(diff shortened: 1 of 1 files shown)


web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -452,26 +454,49 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
...
-          to = dir,
+          to = fs.dir,
           state = fs.state,
-          status = Status.Created,
-          expectedBody = "",
-          newState = Changed(fs.filesInDir.map{ case (relFile,data) => (dir </> relFile.path, data)}.list.toMap))
+          status = Status.Conflict,
+          body = (json: Json) => json must beOneOf(
+            fs.filesInDir.list.map { case (p, _) =>
+              Json("error" := s"${printPath(fs.dir </> p.path)} already exists")
```
jedesah: Did you consider changing the spec to say fs.dir already exists? I know the `InMemory` interpreter does not understand directories, and maybe that's why this is the error being produced, but it might be a better error message to complain about the directory itself. Unless we allow to merge the content of the directory being moved with the directory that already exists, in which case this would be the appropriate error message. Okay, I guess I just convinced myself. But I would be curious to get your thoughts on this.

jedesah:
Aside from concerns about the error message being produced for conflicts on a directory move, LGTM. Like you said though, your intent is to better document the current behavior and we can change it to what the stakeholder wants later on so I think this can merged as is.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
mossprescott: Actually, MongoDB doesn't really understand directories either, in a very similar way ;)

I don't think this error is very good, but since the correct behavior from the web service point of view is probably to return 200 after, say, just checking that the directory exists, I don't think there's much point in worrying over it right now.

A better question is what happens when you MOVE a directory to a *different* directory that also exists (and has files in it), as you're suggesting, and I agree in that case this message might be reasonable but there are more important questions to resolve there I think. What happens if some files are moved and then one conflicts? Should we try to pre-validate, and then give the user a chance to decide what to do? What if we do that and then conditions change before "commit"? All of these are things I don't think we want to worry about right now.

PR #1066, opened 2016-01-28
Create async mongo client from a ConnectionString instead of settings

wemrysi:
Fixes a bug where we weren't copying the credentials from the connection string into the settings object when creating the mongo client. We were only using the settings object in order to configure a timeout that appears to be no longer needed given that we're monitoring the connection time via a different means for this driver.

Given this, we can just switch to creating the mongo client directly from the connection string.


mossprescott:
LGTM

PR #1068, opened 2016-01-28
Allow arbitrary expressions in the REPL.

sellout:
Previously, the REPL only recognized expressions that began with
`select`, however, the parser has long been able to handle expressions
like `4 + 3`, etc. that don’t use `select`. This gets rid of the
`Unknown` REPL command, and tries to treat anything that doesn’t match
some other pattern as an expression.

This also adds literal map/object syntax to the parser & SQL AST (part
of SD-990), short-circuits Constant LPs before calling the backend
planner, and fixes a few other issues I ran into when testing the REPL:

* `select 4 + 3 offset 1` would parse incorrectly as `select (4 + 3 offset 1)`,
* flattening didn’t constant fold when it could, and
* `Take`/`Drop`/`MakeArray`/`MakeObject` didn’t operate on Data.Set correctly.

### DIFF
--- a/core/src/main/scala/quasar/planner.scala
+++ b/core/src/main/scala/quasar/planner.scala
@@ -35,8 +35,8 @@ trait Planner[PhysicalPlan] {
     EitherT[WriterResult, E, A]((Vector(PhaseResult.Detail(name, plan.toString)), \/-(a)))
   }
 
-  def queryPlanner(showNative: PhysicalPlan => (String, Cord))(implicit RA: RenderTree[PhysicalPlan]):
-      QueryRequest => EitherT[(Vector[quasar.PhaseResult], ?), CompilationError, PhysicalPlan] = { req =>
+  def compileToLP(implicit RA: RenderTree[PhysicalPlan]):
+      QueryRequest => EitherT[(Vector[quasar.PhaseResult], ?), CompilationError, Fix[LogicalPlan]] = { req =>
     // TODO: Factor these things out as individual WriterT functions that can be composed.
     for {
       select     <- withTree("SQL AST")(\/-(req.query))
@@ -45,7 +45,14 @@ trait Planner[PhysicalPlan] {
       logical    <- withTree("Logical Plan")(Compiler.compile(tree).leftMap(CSemanticError(_)))
       optimized  <- withTree("Optimized")(\/-(Optimizer.optimize(logical)))
       checked    <- withTree("Typechecked")(LogicalPlan.ensureCorrectTypes(optimized).disjunction.leftMap(ManyErrors(_)))
-      physical   <- plan(checked).leftMap(CPlannerError(_))
+    } yield checked
+  }
+
+  def backendPlanner(showNative: PhysicalPlan => (String, Cord))(implicit RA: RenderTree[PhysicalPlan]):
+      Fix[LogicalPlan] => EitherT[(Vector[quasar.PhaseResult], ?), CompilationError, PhysicalPlan] = { lp =>
+    // TODO: Factor these things out as individual WriterT functions that can be composed.
+    for {
+      physical   <- plan(lp).leftMap(CPlannerError(_))
       _          <- withTree("Physical Plan")(\/-(physical))
       _          <- withString(physical)(showNative)
     } yield physical
--- a/core/src/main/scala/quasar/sql/ast.scala
+++ b/core/src/main/scala/quasar/sql/ast.scala
@@ -39,6 +39,8 @@ object ExprF {
   final case class VariF[A](symbol: String) extends ExprF[A]
   final case class SetLiteralF[A](exprs: List[A]) extends ExprF[A]
   final case class ArrayLiteralF[A](exprs: List[A]) extends ExprF[A]
+  /** Can’t be a Map, because we need to arbitrarily transform the key */
+  final case class MapLiteralF[A](exprs: List[(A, A)]) extends ExprF[A]
   final case class SpliceF[A](expr: Option[A]) extends ExprF[A]
   final case class BinopF[A](lhs: A, rhs: A, op: BinaryOperator)
       extends ExprF[A]
@@ -229,6 +231,13 @@ object ArrayLiteralF {
   def unapply[A](obj: ExprF[A]): Option[List[A]] = obj match {
     case ExprF.ArrayLiteralF(exprs) => Some(exprs)
     case _                          => None
+  }
+}
+object MapLiteralF {
+  def apply[A](exprs: List[(A, A)]): ExprF[A] = ExprF.MapLiteralF(exprs)
+  def unapply[A](obj: ExprF[A]): Option[List[(A, A)]] = obj match {
+    case ExprF.MapLiteralF(exprs) => Some(exprs)
+    case _                        => None
   }
 }
 object SpliceF {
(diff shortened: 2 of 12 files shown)


core/src/main/scala/quasar/planner.scala
```diff
@@ -35,8 +35,8 @@ trait Planner[PhysicalPlan] {
     EitherT[WriterResult, E, A]((Vector(PhaseResult.Detail(name, plan.toString)), \/-(a)))
   }
 
-  def queryPlanner(showNative: PhysicalPlan => (String, Cord))(implicit RA: RenderTree[PhysicalPlan]):
-      QueryRequest => EitherT[(Vector[quasar.PhaseResult], ?), CompilationError, PhysicalPlan] = { req =>
+  def queryPlanner(implicit RA: RenderTree[PhysicalPlan]):
```
jedesah: I wonder if we might be able to come up with a better name for this? If I understand correctly, it is compiling the sql string into `LogicalPlan`. I wonder if this even needs to be inside `Planner`? Maybe something like `compileToLP` or something.

core/src/main/scala/quasar/sql/ast.scala
```diff
@@ -39,6 +39,8 @@ object ExprF {
   final case class VariF[A](symbol: String) extends ExprF[A]
   final case class SetLiteralF[A](exprs: List[A]) extends ExprF[A]
   final case class ArrayLiteralF[A](exprs: List[A]) extends ExprF[A]
+  /** Can’t be a Map, because we need to arbitrarily transform the key */
```
jedesah: I am not sure I fully understand this comment.

core/src/main/scala/quasar/std/set.scala
```diff
@@ -45,6 +45,9 @@ trait SetLib extends Library {
     setTyper(partialTyper {
       case _ :: Type.Const(Data.Int(n)) :: Nil if n == 0 =>
         Type.Const(Data.Set(Nil))
+      case Type.Const(Data.Set(s)) :: Type.Const(Data.Int(n)) :: Nil
+          if n.isValidInt =>
+        Type.Const(Data.Set(s.take(n.intValue)))
       case Type.Set(t) :: _ :: Nil => t
```
jedesah: Totally unrelated, but I am confused by the typer again. Taking from a `Set` of `A` returns `A`? I would have expected still a `Set` of `A`. The addition looks fine, it's the line below that confuses me.

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -33,6 +33,8 @@ trait StructuralLib extends Library {
     AnyObject, Str :: Top :: Nil,
     noSimplification,
     partialTyper {
+      case List(Const(Data.Str(name)), Const(Data.Set(data))) =>
+        Const(Data.Set(data.map(d => Data.Obj(Map(name -> d)))))
```
jedesah: A `Set` is `AnyObject`?

jedesah: Oh, this is the thing where a `Set` of a single element has the same type as that element?

core/src/main/scala/quasar/sql/ast.scala
sellout: We need approx `Map[A, A]`, and you can’t implement Functor on that, because the key is constrained (or, at least _should_ be – wouldn’t be surprised if Scala’s wasn’t). But it definitely will be when we switch to Cats.

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -56,8 +58,10 @@ trait StructuralLib extends Library {
     AnyArray, Top :: Nil,
     noSimplification,
     partialTyper {
-      case Const(data) :: Nil => Const(Data.Arr(data :: Nil))
-      case valueType :: Nil   => Arr(List(valueType))
+      case Const(Data.Set(data)) :: Nil =>
```
jedesah: I kind of wonder if it might make sense to have some kind of pass elsewhere that removes the `Set` type from a singleton set as opposed to having to do this in every single function (and remember to do it to new functions in the future).

core/src/main/scala/quasar/std/set.scala
sellout: Yeah, basically, until we have MRA, our handling of Sets is inconsistent. We don’t always distinguish between `Set(x)` and `x`. But in certain cases Sets do show up, and we “simplify” them. I don’t really have a good defense. it’s a problem of dimension tracking.

core/src/main/scala/quasar/std/structural.scala
sellout: Yep, basically that same thing. With MRA, the codomain should be `Set(AnyObject)`, and an `AnyObject` is equivalent to a singleton Set of AnyObject.

core/src/main/scala/quasar/sql/ast.scala
jedesah: Ah, that makes sense.

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -49,7 +49,9 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
     }
   }
 
-  val queryPlanner = MongoDbPlanner.queryPlanner(κ("Mongo" -> Cord.empty))
+  def queryPlanner(qr: QueryRequest) =
+    MongoDbPlanner.queryPlanner.apply(qr)
```
jedesah: Ha! This is conclusive proof that there is something weird with the names of these things. I am not craZY!

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -84,6 +86,11 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
         beWorkflow($pure(Bson.Doc(ListMap("0" -> Bson.Int64(1)))))
     }
 
+    "plan complex constant" in {
+      plan("[1, 2, 3, 4, 5][*] limit 3 offset 1") must
```
jedesah: What does `[1,2,3,4,5][*]` mean?

sellout: `[1 … 5]` is just literal array syntax, then `[*]` is array flatten. so this expression builds up an array, flattens it into a set with 5 int entries. Then we take/drop elements from that set.

jedesah: Okay, forgot about the `*` thing even though I now remember John presenting it at SlamFest. Then, it took me a bit more time to remember that the ordering of `limit` and `offset` matters. Definitely in this post-fix notation, it's a little surprising at first.

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -225,7 +232,11 @@ trait StructuralLib extends Library {
     "Zooms in on the keys of a map, also extending the current dimension with the keys",
     Set(Top), AnyObject :: Nil,
     noSimplification,
-    partialTyper { case List(x) if x.objectLike => Str },
+    partialTyper {
+      case List(Const(Data.Obj(map))) =>
+        Const(Data.Set(map.toList.map(p => Data.Str(p._1))))
```
jedesah: Might be slightly nicer here to do `map.keys.map(Data.Str).toList`

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -200,6 +204,8 @@ trait StructuralLib extends Library {
     Set(Top), AnyObject :: Nil,
     noSimplification,
     partialTyperV {
+      case List(Const(Data.Obj(map))) =>
+        success(Const(Data.Set(map.toList.map(_._2))))
```
jedesah: Same here. Perhaps, `map.values.toList`.

sellout:
@jedesah Back in your court.

jedesah:
LGTM.

PR #1069, opened 2016-01-28
in the view FS interpreter, handle PathNotFound errors from the underlying FS

mossprescott:
...which by the way are now generated by TraceFS.

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/view.scala
+++ b/core/src/main/scala/quasar/fs/mount/view.scala
@@ -177,10 +177,18 @@ object view {
           query.explain(views.rewrite(lp)).run.run
 
         case ListContents(dir) =>
-          query.ls(dir).map(_ ++ views.ls(dir)).run
+          query.ls(dir).run.map(_ match {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -177,10 +184,12 @@ object view {
           query.explain(views.rewrite(lp)).run.run
 
         case ListContents(dir) =>
-          query.ls(dir).map(_ ++ views.ls(dir)).run
+          handlePathError(query.ls(dir), Set[PathName]())
```
wemrysi: Hm, don't we want to return the result of `views.ls(dir)` in the error case? And if that result is empty, we want to error with a `PathNotFound`, unless `dir === rootDir`, in which case we can return the empty set.

wemrysi: It seems the integration tests would agree ;)

wemrysi:
:+1:

PR #1070, opened 2016-01-29
Convert the REST API and server to use the 'Free' architecture

wemrysi:
Converts the configuration, rest api, server main and their dependencies to use the `Free`-based architecture, replacing the previous rest api implementation.

NB: Given the number of commits and the amount of times we've merged master into this branch, I think we'll either want to merge into master as-is, or squash into a single monolithic commit.


wemrysi:
Looks like there is an issue with renaming a mongo database, the server responds with a `500` and an empty body. Though renaming a collection seems to work.

wemrysi:
Also looks like there is an issue with creating a folder at the top-level where the new folder would result in a new database. Currently the frontend attempts to create a database called "Untitled+Folder" which, once it is turned parsed by our `Collection` type becomes "Untitled$addFolder" and then add a collection to it, but mongodb 3.0.x fails with

`2016-01-29T04:26:36.743+0000 I -        [conn2805] Assertion: 17320:cannot do createCollection on namespace with a $ in it: Untitled$addFolder.__quasar\.tmp_388bb331_1`

I'm not sure if this works on master either, or how different versions of mongo behave given a db name like this.

wemrysi:
However, creating folders inside a database and moving files/collections into them appears to work fine.

sellout:
Are you sure that’s not a URL-decoding failure? I think “Untitled Folder” is a more likely name than “Untitled+Folder”. Not that the “+” handling isn’t a bug, but the regression may be in decoding.

wemrysi:
Aye it might be, though if something with a "+" in it ever made it through intentionally, we'd still hit the other issue.

sellout:
Yeah, and I think we encode all kinds of symbols with `$<something>`, so plenty of ways to run into that.

wemrysi:
Odd, so the "+" only shows up when "Untitled%20Folder" is decoded for the database name, when creating a new folder inside a database it shows up as "Untitled Folder" in the collection name.

mossprescott:
That's odd. That's our own encoding for database names using `$add` to escape the `+`, so maybe we need to revisit that choice.

sellout:
@mossprescott Yeah, I remember the discussion of names to use. I can’t seem to find any documentation on restrictions on “$” in names.

@wemrysi So, we should be URL-decoding (or rather, http4s should be) _everything_ before we look at anything. Is the broken decoding happening in the URL path, or in a path embedded in the query? Is it getting double-encoded somewhere? Weird that %20 would _become_ + during decoding. Sounds like an http4s bug?

mossprescott:
Oops.

```cpp
inline bool NamespaceString::normal(StringData ns) {
    if (ns.find('$') == std::string::npos)
        return true;
    return oplog(ns);
}
```
https://github.com/mongodb/mongo/blob/bee7144f59bc8782c3c8e76c45a9f9e0e39cb1e6/src/mongo/db/namespace_string-inl.h#L43-L47

mossprescott:
We are definitely not handling encoding properly in the new web stuff, but I punted on fixing it because many of the issues already exist in master (see [SD-1322](https://slamdata.atlassian.net/browse/SD-1322)). But this problem with spaces in folder names seems like it needs to be addressed.

mossprescott:
Also see https://github.com/mongodb/mongo/blob/bee7144f59bc8782c3c8e76c45a9f9e0e39cb1e6/src/mongo/db/namespace_string.h#L207: "A "$" is used for namespaces holding index data, which do not contain BSON objects in their records. ("oplog.$main" is the exception)"

sellout:
Alright, so we have to find another symbol to prefix our encoded symbols with, and then additional encode that symbol … right? Can we use Unicode in names? Would be nice to pick something a bit more obscure :)

sellout:
Like, � seems to be a good bet.

mossprescott:
Right, I think it's fair to say that this encoding has never worked, so we should just pick something else. Unicode seems fine on my system but I wouldn't trust it everywhere. I suggest we pick something in 7-bit ASCII but not any of these: 
```scala
  val DatabaseNameEscapes = List(
    " "  -> "+",
    "."  -> "~",
    "$"  -> "$$",
    "+"  -> "$add",
    "~"  -> "$tilde",
    "/"  -> "$div",
    "\\" -> "$esc",
    "\"" -> "$quot",
    "*"  -> "$mul",
    "<"  -> "$lt",
    ">"  -> "$gt",
    ":"  -> "$colon",
    "|"  -> "$bar",
    "?"  -> "$qmark")
```

sellout:
Oh, looks like we need to re-encode $, then, too. Ohhhhhh! and that’s where we’re adding the plus!

sellout:
Oh, maybe not – because it ends up as $add … any chance we’re double-encoding the db name somehow?

PR #1071, opened 2016-01-29
Path encoding/escaping issues

mossprescott:
Really not happy about all the silly filtering this requires in the tests but didn't want to make this wait for a fix in pathy. On the other hand, switching the escape char for DB's has hardly any fallout.

With this in place, I can add folders at the DB-level, through the API, with names like `Untitled Folder.slam` and `foo/bar.slam`:
```
> show databases
...
Untitled+Folder~slam  0.078GB
admin                 0.078GB
foo%divbar~slam       0.078GB
quasar-test           0.453GB
```

### DIFF
--- a/web/src/test/scala/quasar/api/services/MountServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/MountServiceSpec.scala
@@ -20,18 +20,12 @@ import pathy.Path, Path._
 import pathy.scalacheck._
 import pathy.scalacheck.PathyArbitrary._
 
-class MountServiceSpec extends Specification with ScalaCheck with Http4s {
+class MountServiceSpec extends Specification with ScalaCheck with Http4s with PathUtils {
   import posixCodec.printPath
 
   val StubFs = FileSystemType("stub")
 
-  // NB: workaround for apparent bug in http4s's decoding of paths, where
-  // a '%' followed by '/' or a high code-point is dropped. This won't directly
-  // affect the production code path, because it parses the path from the
-  // request stream, where these characters should be escaped anyway.
-  // However, `posixCodec` is certainly not the right way to escape paths
-  // for HTTP and we will want to sort that out at some point. See SD-1322.
-  def pathUri(path: APath): Uri = Uri(path = printPath(path).replaceAll("%(?![0-9a-fA-F])", "%25"))
+  def pathUri(path: APath): Uri = Uri(path = UriPathCodec.printPath(path))
 
   def viewConfig(q: String, vars: (String, String)*): (Expr, Variables) =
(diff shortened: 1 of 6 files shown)


mossprescott:
And, on the other side:
```json
{
    "children": [
        {
            "type": "directory",
            "name": "Untitled Folder.slam"
        },
        {
            "type": "directory",
            "name": "admin"
        },
        {
            "type": "directory",
            "name": "foo/bar.slam"
        },
        {
            "type": "directory",
            "name": "quasar-test"
        },
...
```

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -291,83 +300,91 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s {
...
+              }
             }
           }
         }
 
         // TODO: escaped paths do not survive being embedded in error messages
         "succeed with view 'above' existing fs path" ! prop { (d: AbsDirOf[AlphaCharacters], view: RFile, fsSuffix: RDir) =>
-          runTest { service =>
-            val fsCfg = ()
+          (!hasDot(d.path) && !hasDot(view)) ==> {
```
jedesah: Kind of curious why here we call `hasDot` twice, but on line 172 we create a single path and call `hasDot` once. I guess not a very big deal at all, just seems a little inconsistent.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -506,16 +531,18 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s {
...
-        runTest { service =>
-          for {
-            resp <- service(Request(
-                      method = POST,
-                      uri = pathUri(parent))
-                    .withBody("""{ "stub": { "connectionUri": "foo" } }""").run)
-          } yield {
-            resp.as[Json].run must_== Json("error" := "The 'X-File-Name' header must be specified")
-            resp.status must_== Status.BadRequest
+      "be 400 with missing X-File-Name header" ! prop { (parent: ADir) =>
```
jedesah: huh, nice catch on getting rid of pointless `fsDir`.

jedesah:
LGTM!

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
mossprescott: Because I meant to clean that up and missed a couple of places, thanks!

mossprescott:
Fixed, rebased, merged, and feeling pretty confident about it.

PR #1073, opened 2016-02-01
Add folder for Changelogs and create tentative changelog for 3.0.0

jedesah:


sellout:
Here’s a possible rephrasing/expansion of the flexible SQL bit:

* can use expressions that don’t begin with `select` as standalone queries, e.g. `5 + 1`
* added the `^` (power) operator and literal map syntax to SQL²: `{4: "four", 3: "three"}`

PR #1074, opened 2016-02-03
Various improvements to our SBT config.

sellout:
This was triggered by newer versions of Ensime not working with our
repository. The two changes that were necessary to fix that were

* Don’t explicitly include `Defaults.defaultSettings` in our settings –
  it tramples various plugin settings, and is deprecated.
* Include our settings on the root project, so we don’t also fetch scala
  2.10 libraries.

Other changes:

* added a dir-locals file to get Emacs tools to use the local `sbt`
  rather than some other one on the system,
* added compiler options to be used by the meta-project build (these
  would have caught one of the primary issues above),
* updated the copyright notice to include 2016,
* made the Java version check failure include the version it _did_ find,
* replaced `CustomKeys` with the simpler `Versions` object,
* removed the `NextVersion` steps in the release process, and
* upgraded a couple dependencies.

There is still more to be done here. Ideally, we’d almost completely
eliminate .sbt files and do everything in .scala files, but this is an
improvement.


mossprescott:
:-1: 

screen shot 2016-02-02 at 16 32 21

mossprescott:
More seriously, I'd prefer if the license comments were in a separate commit.

sellout:
Yeah, I can get on board with the copyright thing. I just updated it in the build.sbt as I was going through, then boom – 300 files changed.

build.sbt
```diff
@@ -11,23 +19,24 @@ lazy val ExclusiveTests = config("exclusive") extend Test
...
-lazy val checkHeaders = taskKey[Unit]("Fail the build if createHeaders is not up-to-date")
+lazy val checkHeaders =
+  taskKey[Unit]("Fail the build if createHeaders is not up-to-date")
 
-lazy val standardSettings = Defaults.defaultSettings ++ Seq(
+lazy val commonSettings = Seq(
   headers := Map(
-    "scala" -> Apache2_0("2014 - 2015", "SlamData Inc."),
-    "java"  -> Apache2_0("2014 - 2015", "SlamData Inc.")),
+    ("scala", Apache2_0("2014 - 2015", "SlamData Inc.")),
```
jedesah: What is the rationale here for ordinary tuple syntax over `->`?

build.sbt
```diff
@@ -38,8 +47,8 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
     "JBoss repository" at "https://repository.jboss.org/nexus/content/repositories/",
     "Scalaz Bintray Repo" at "http://dl.bintray.com/scalaz/releases",
     "bintray/non" at "http://dl.bintray.com/non/maven"),
-  addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.5.4"),
```
jedesah: I wonder if this could help with the issues we run into using kind-projector. I didn't find any release notes though... Man, what kind of open source project does not have release notes in their git repo... :smile:

project/Versions.scala
```diff
@@ -0,0 +1,10 @@
+import sbt._, Keys._
+
+object Versions {
```
jedesah: I don't like this. I think it should be defined next to `libraryDependencies`.

build.sbt
```diff
@@ -62,61 +71,49 @@ lazy val standardSettings = Defaults.defaultSettings ++ Seq(
...
+    "com.github.scopt"  %% "scopt"                     % "3.3.0"        % "compile, test",
+    "org.threeten"      %  "threetenbp"                % "1.2"          % "compile, test",
+    "org.mongodb"       %  "mongo-java-driver"         % mongoVersion   % "compile, test",
+    "org.mongodb"       %  "mongodb-driver-async"      % mongoVersion   % "compile, test",
+    "io.argonaut"       %% "argonaut"                  % "6.1"          % "compile, test",
+    "org.jboss.aesh"    %  "aesh"                      % "0.55"         % "compile, test",
+    "org.typelevel"     %% "shapeless-scalaz"          % slcVersion     % "compile, test",
+    "com.slamdata"      %% "pathy-core"                % pathyVersion   % "compile",
+    "com.github.mpilquist" %% "simulacrum"             % "0.4.0"        % "compile, test",
+    "org.http4s"        %% "http4s-core"               % http4sVersion  % "compile",
```
jedesah: Why do we need `http4s-core` in core?

sellout: I actually tried to remove it, but we use `Uri` and possibly other bits.

build.sbt
sellout: Basically, inability to import the `ArrowAssoc` implicit class directly. We solve this in the main project in our Predef object, but that seemed like overkill for the couple instances here.

project/Versions.scala
sellout: Yeah, so then I should probably move libraryDependencies to a .scala file.

build.sbt
jedesah: Ah, okay, that's right.

project/Versions.scala
jedesah: Nah, you don't need to do that. You should be able to write:

```
libraryDependencies ++= {
    val scalazVersion = ...
    ...
    Seq(
      ...)}
```

.dir-locals.el
```diff
@@ -0,0 +1 @@
+((nil . ((sbt:program-name . "./sbt"))))
```
drostron: Wished this wasn't necessary. Out of curiosity, what are we getting out of the sbt script? I haven't been using it for quasar and it's been quite awhile since I've used it in other projects.

project/Versions.scala
sellout: Well, at least one of these version numbers is shared between the dependencies here and the web dependencies, so we can’t just use a val there. I had originally added these as vals, and I’m guessing that whoever added the web dependencies changed the versions to settings so they could be shared.

.dir-locals.el
sellout: I think the main thing this launcher gets us (although may there’s other stuff) is that it makes quasar more self contained. You don’t need scala or sbt installed on your system. And since we use the same code to _get_ sbt, hopefully a bit more consistent/deterministic than devs having their own versions.

sellout: That said, I’m not opposed to getting rid of the launcher.

drostron: Fair enough. The default sbt installs do a good job of ensuring that the sbt version specified in build.properties is used. I do see the advantage of having a script install everything necessary to get a dev environment up and running though.

mossprescott: IIRC, the sbt script was there to increase the PermGen size which used to be a pain, but obviously that's no longer an issue.

If we don't have our own `sbt` script, does that mean we don't need `.dir-locals.el`?

I think getting rid of it is a good idea.

sellout: Yep, I’d happily remove .dir-locals.el in that case.

sellout:
@mossprescott Re: license header changes – do you want that in a separate PR, or just a non-squashed commit in this PR?

mossprescott:
A non-squashed commit would be fine.

project/Versions.scala
jedesah: I don't think there is any link between the use of `http4s-core` in quasar `core` subproject and its use in `web` subproject. So I might argue that they don't need to be "linked" together. If you can make the argument that they are linked, I think I would still argue for only having that defined in some external/common place and have the values specific to the main build file defined as locally as possible.

sellout: Don’t they need to be linked in that we can only have one version as a dependency?

jedesah: Not technically. `core` can declare the version it depends on like any scala lib would and it would be up to `web` to use that version or a newer one if it wants. In practice, I see little reason not to use the same version, but one could argue that they do not "fundamentally" need to be the same. Really, I am just attempting to fabricate an argument for why this particular dependency can be duplicated and we can declare these variables locally.

sellout: @jedesah What do you think of the newer approach (Versions is gone, there’s now Dependencies).

sellout:
@drostron Whenever you think this looks good, I’ll squash everything, then add another commit that updates the license headers everywhere, and give you another chance to look before merging.

project/Versions.scala
jedesah: I like it.

drostron:
👍🏽

drostron:
👍🏾

PR #1076, opened 2016-02-05
Add compiler flags.

sellout:
* Turn on the long-disabled `-Ywarn-numeric-widen` after eliminating the
  warnings and
* add options to take advantage of Java 8 features (e.g., invoke dynamic
  & lambdas).

### DIFF
--- a/core/src/main/scala/quasar/fs/ManageFile.scala
+++ b/core/src/main/scala/quasar/fs/ManageFile.scala
@@ -42,44 +42,32 @@ object ManageFile {
     *     additional primitive operations for the conditional write operations.
     */
   object MoveSemantics {
-    object Case {
-      case object Overwrite extends MoveSemantics
-      case object FailIfExists extends MoveSemantics
-      case object FailIfMissing extends MoveSemantics
-    }
-
     /** Indicates the move operation should overwrite anything at the
       * destination, creating it if it doesn't exist.
       */
-    val Overwrite: MoveSemantics = Case.Overwrite
+    case object Overwrite extends MoveSemantics
 
     /** Indicates the move should (atomically, if possible) fail if the
       * destination exists.
       */
-    val FailIfExists: MoveSemantics = Case.FailIfExists
+    case object FailIfExists extends MoveSemantics
 
     /** Indicates the move should (atomically, if possible) fail unless
       * the destination exists, overwriting it otherwise.
       */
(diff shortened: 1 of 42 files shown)


core/src/main/scala/quasar/fs/ManageFile.scala
```diff
@@ -100,18 +91,12 @@ object ManageFile {
...
   object MoveScenario {
-    object Case {
-      final case class DirToDir(src: ADir, dst: ADir)
+    // FIXME: These constructors should be private, but that causes errors in
+    //        transformPaths.scala.
+    final case class DirToDir(src: ADir, dst: ADir)
         extends MoveScenario
-      final case class FileToFile(src: AFile, dst: AFile)
+    final case class FileToFile(src: AFile, dst: AFile)
         extends MoveScenario
```
sellout: @wemrysi When I make these constructors private (and convert the calls in transformPaths.scala to use the prisms), I get errors like
```
monocle.Prism[quasar.fs.ManageFile.MoveScenario,(quasar.fs.ADir, quasar.fs.ADir)] does not take parameters
              (src, dst) => dirToDir(inPath(src), inPath(dst)),
                                    ^
```
yet, I managed to convert other types in this commit to that same style without problem. Any idea what I’m doing wrong?

wemrysi: You'll need to import `quasar.fp.prism._` (or `quasar.fp._`) to get the implicits that provide the syntax.

sellout: Ah, thanks. I was hoping it was something like that. I’ll add that to the Wiki page. Also, I don’t think we need `private[pkg]`, right, just `private` is fine. The only thing that needs to see the constructors are the prisms in the same object.

wemrysi: Unfortunately, you do need `private[pkg]` as the prisms are in the _trait_'s object, not the subtype introduced by the constructor =(

wemrysi: Though, and I didn't think about this before, it looks like `private[MoveScenario]` should work, which is a bit better.

sellout:
Ok, this is ready for review, finally. Especially check the last changes to the build script.

wemrysi:
:+1:

PR #1077, opened 2016-02-05
Convert data service to a QHttpService

wemrysi:
* Added Catchable[Free[S, ?]] when Task :<: S
* Changed QHttpService to a be a partial function to allow for taking
  advantage of Http4s fallthrough behavior

### DIFF
--- a/core/src/main/scala/quasar/fp/free/package.scala
+++ b/core/src/main/scala/quasar/fp/free/package.scala
@@ -16,7 +16,12 @@
 
 package quasar.fp
 
+import quasar.Predef._
+
 import scalaz._
+import scalaz.syntax.either._
+import scalaz.syntax.monad._
+import scalaz.concurrent.Task
 
 package object free {
   type Coproduct3[F[_], G[_], H[_], A] = Coproduct[F, Coproduct[G, H, ?], A]
@@ -72,5 +77,36 @@ package object free {
     new (Coproduct5[F, G, H, I, J, ?] ~> M) {
       def apply[A](fa: Coproduct5[F, G, H, I, J, A]) =
         fa.run.fold(f, interpret4(g, h, i, j)(_))
+    }
+
+  /** A `Catchable` instance for `Free[S, ?]` when `Task` can be injected into `S`. */
+  implicit def freeCatchable[S[_]: Functor](implicit S: Task :<: S): Catchable[Free[S, ?]] =
+    new Catchable[Free[S, ?]] {
+      type G[A] = Free[S, A]
+      private val injFT: Task ~> G = injectFT[Task, S]
+      private val lftFT: S ~> G = liftFT[S]
+
+      def attempt[A](fa: Free[S, A]): Free[S, Throwable \/ A] =
+        injFT(Task.delay(fa.resume match {
+          case \/-(a) =>
+            a.right[Throwable].point[G]
+
+          case -\/(sa) => S.prj(sa) match {
+            case Some(t) =>
+              injFT(t.attempt) flatMap {
+                case -\/(t)   => t.left[A].point[G]
+                case \/-(fa0) => attempt(fa0)
+              }
+
+            case None =>
+              lftFT(sa).flatMap(attempt)
+          }
+        }).attempt map {
+          case \/-(a) => a
+          case -\/(t) => t.left[A].point[G]
+        }).join
+
+      def fail[A](t: Throwable): Free[S, A] =
+        injFT(Task.fail(t))
     }
 }
--- a/web/src/main/scala/quasar/api/ServerOps.scala
+++ b/web/src/main/scala/quasar/api/ServerOps.scala
@@ -23,7 +23,7 @@ import quasar.console._
 import quasar.config._
 import quasar.effect._
 import quasar.fp._
-import quasar.fs.{FileSystem, QueryFile, APath, ADir, Empty}
+import quasar.fs._
 import quasar.fs.mount._
 import Http4sUtils._
 import quasar.physical.mongodb._
@@ -33,10 +33,10 @@ import java.io.File
 import java.lang.System
 
 import argonaut.{CodecJson, EncodeJson}
+import monocle.Lens
 import org.http4s
 import org.http4s.server.{Server => Http4sServer, HttpService}
 import org.http4s.server.blaze.BlazeBuilder
-import monocle.Lens
 import scalaz.{Failure => _, Lens => _, _}
 import scalaz.syntax.monad._
 import scalaz.syntax.show._
@@ -139,6 +139,7 @@ object Server {
         Coyoneda.liftTF[WorkflowExecErr, Task](Failure.toTaskFailure[WorkflowExecutionError]))))
   }
 
+  // TODO: Interpret HFSFailureF into `ResponseOr`
   def fsEffToTask(
     seqRef: TaskRef[Long],
     viewHandlesRef: TaskRef[ViewHandles],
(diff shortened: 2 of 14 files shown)


core/src/main/scala/quasar/fp/free/package.scala
```diff
@@ -73,4 +78,35 @@ package object free {
       def apply[A](fa: Coproduct5[F, G, H, I, J, A]) =
         fa.run.fold(f, interpret4(g, h, i, j)(_))
     }
+
+  /** A `Catchable` instance for `Free[S, ?]` when `Task` can be injected into `S`. */
+  implicit def freeCatchable[S[_]: Functor](implicit S: Task :<: S): Catchable[Free[S, ?]] =
```
jedesah: Out of curiosity, do you think it would be possible to do the same thing with `Failure`?

web/src/main/scala/quasar/api/ServerOps.scala
```diff
@@ -227,7 +232,15 @@ object Server {
       val mounting: MountingF ~> TaskAndConfigsM =
         Coyoneda.liftTF[Mounting, TaskAndConfigsM](free.foldMapNT(mnt) compose mounter)
 
-      free.interpret2(mounting, liftTask compose evalFromRef(evalFsRef, f))
+      val throwFailure: FileSystemFailureF ~> Task =
```
jedesah: does not technically "throw", but I think that's fine.

web/src/main/scala/quasar/api/package.scala
```diff
@@ -17,29 +17,52 @@
...
 import pathy.Path, Path._
+import scalaz.{Failure => _, _}, Scalaz._
+import scalaz.concurrent.Task
 
 package object api {
   // TODO: Names
   type ResponseT[F[_], A] = EitherT[F, Response, A]
   type ResponseOr[A] = ResponseT[Task, A]
 
+  /** Interpret a `Faliure` effect into `ResponseOr` given evidence the
```
jedesah: typo in Faliure

core/src/main/scala/quasar/fp/free/package.scala
wemrysi: What do you mean? Implement `Catchable[Free[S, ?]]` given `Failure[Throwable, ?] :<: S`?

jedesah: Ya, that's what I meant.

wemrysi: Actually, it might. I originally started to write it for `MonadError[_[_, _], Throwable]` but was concerned about being able to capture exceptions in the monad, however we might be able to just use `try`/`catch` directly rather than rely on the monad to.

It'd be worth exploring, for sure.

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -55,22 +56,23 @@ final case class RestApi(defaultPort: Int, restart: Int => Task[Unit]) {
...
       //"/query/fs"     -> query.service(f),
       //"/server"       -> server.service(defaultPort, restart),
       //"/welcome"      -> welcome.service
     )
-  }
 }
 
 object RestApi {
-  def wrap(service: HttpService) =
+  def withDefaultMiddleware(service: HttpService): HttpService =
```
drostron: Rename appreciated.

core/src/main/scala/quasar/fp/free/package.scala
jedesah: Also, if we have  a `Catchable` for `Free` in such a circumstance, then it strikes me that we could `run` the `Process[Free[S,?],A]` and simplify `QHttpService` to `Request => Free[S, Response]` where `Task :<: S` or `Failure[Throwable,?] :<: S`.

web/src/main/scala/quasar/api/services/RestApi.scala
jedesah: `wrap` is SOO much better. Have no idea what you guys are talking about! :stuck_out_tongue_closed_eyes:

core/src/main/scala/quasar/fp/free/package.scala
wemrysi: I'm not sure that would allow for streaming responses.

web/src/main/scala/quasar/api/services/package.scala
```diff
@@ -33,12 +33,66 @@ import scalaz.concurrent.Task
...
+
+  def requiredHeader2[F[_]](key: HeaderKey.Extractable, request: Request): QuasarResponse[F] \/ key.HeaderT =
+    request.headers.get(key) \/> QuasarResponse.error(BadRequest, s"The '${key.name}' header must be specified")
 
   def respond[S[_], A, F[_]](a: Free[S, A])(implicit ev: ToQuasarResponse[A, F]): Free[S, QuasarResponse[F]] =
     a.map(ev.toResponse)
 
-  def requiredHeader2[F[_]](key: HeaderKey.Extractable, request: Request): QuasarResponse[F] \/ key.HeaderT =
-    request.headers.get(key) \/> QuasarResponse.error(BadRequest, s"The '${key.name}' header must be specified")
+  def withOffsetAndLimit[S[_]](
```
jedesah: Not sure I am a huge fan of `withOffsetAndLimit` vs `handleOffsetAndLimit`, what was your reasoning for making the change?

core/src/main/scala/quasar/fp/free/package.scala
jedesah: I can't come up with any particular reason why it would not, what makes you think it would not work?

wemrysi: Maybe I've misunderstood what you're suggesting, but once you `process.runLog` you're no longer streaming, no?

jedesah: yaya, okay nevermind :smile:.

web/src/main/scala/quasar/api/services/package.scala
wemrysi: Actually, it'd probably be better to change the function to return `QuasarResponse[S] \/ (Natural, Positive)` rather than accept a function as the latter behavior is what makes the naming odd. Without that we could call it 'validatedOffsetAndLimit` or somesuch.

wemrysi: Possibly even better yet, have two separate functions, though we usually use these together so I'd be ok with the one.

jedesah: Ya, that does sound better and it strikes me that this kind of thing might benefit from framework support as it seems pretty common to want to associate a "Response" with an invalid parameter and do the normal thing if the parameter is present and valid. It's likely the Response would be similar wherever the parameter is expected but not found or invalid.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -499,19 +510,19 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
       }
     }
     "DELETE" >> {
-      "be 200 with existing file" ! prop { filesystem: SingleFileMemState =>
+      "be 204 with existing file" ! prop { filesystem: SingleFileMemState =>
         val request = Request(uri = Uri(path = filesystem.path), method = Method.DELETE)
         val (service, ref) = serviceRef(filesystem.state)
         val response = service(request).run
-        response.status must_== Status.Ok
+        response.status must_== Status.NoContent
```
jedesah: Just learned about the HTTP `No Content` response code

drostron:
👍🏼

web/src/main/scala/quasar/api/services/package.scala
wemrysi: @jedesah @drostron What do we think about something like `offsetOr400` as a name for the function (just for offset, limit will have an analogous one)? Any better suggestions?

jedesah: How about `validate[A](a: ValidationNel[ParseFailure,A]): QuasarResponse[S] \/ A`

Then we could just map over the options and `validate`

wemrysi: See `orBadRequest` in my latest commit, it asks for a bit more (`Traverse`) but gives you `QuasarResponse[S] \/ F[A]`

wemrysi:
@jedesah @drostron Ok, I think I've addressed everything. Let me know if so and I'll squash and merge.

drostron:
LGTM

jedesah:
:+1:

PR #1078, opened 2016-02-05
Releasing 3.0.1

mossprescott:
This time *not* using `sbt release`, just for giggles.

### DIFF
--- a/CHANGELOG/3.0.1.md
+++ b/CHANGELOG/3.0.1.md
@@ -1,0 +1,1 @@
+- Fix bug where GET /data would fail if the target path was a view containing a relative path.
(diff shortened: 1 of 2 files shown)


mossprescott:
By the way, the rationale for just bumping the patch number is that this is not new API. The equivalent request did work in 2.3.3; the bug was a regression in 3.0.0. Fixing the bug just makes the existing API work correctly.

jedesah:
LGTM. I don't even know if it's useful to wait for the travis PR build... Anyway, will merge once that is green unless someone else gets to it first.

mossprescott:
Agreed, With no code changes, what could go wrong? Merging.

PR #1080, opened 2016-02-06
Update scalaz to 7.1.7

wemrysi:
Other than staying current, the main reason for upgrading is the
potential performance benefits for `Free`, given the `fastFlatMap`
optimization backported in 7.1.7.

See https://github.com/scalaz/scalaz/pull/1085

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -5,7 +5,7 @@ import scala.collection.Seq
 import sbt._, Keys._
 
 object Dependencies {
-  private val scalazVersion  = "7.1.4"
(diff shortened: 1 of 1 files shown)


sellout:
Why not upgrade to 7.2? Still waiting on dependencies to do that?

wemrysi:
Probably, though I haven't checked in awhile. However, the work required to upgrade to 7.1.7 is what you see in this PR whereas upgrading to 7.2 looks like it will involve more code changes, even if all the dependencies are available, as it makes breaking changes. This PR just seeks to benefit from the easy win in 7.1.7.

PR #1081, opened 2016-02-06
Convert query service to QHttpService

drostron:
- delete stale code in services package
- add ToQuasarResponse instances for ParsingError and SemanticErrors

_paired with @jedesah on this work_


web/src/main/scala/quasar/api/services/package.scala
```diff
@@ -104,79 +100,4 @@ package object services {
       )
   }
 
-  //----- TO DELETE -----//
```
wemrysi: Now that all of this is gone, mind renaming `requiredHeader2` to `requiredHeader`?

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
+    val Q = QueryFile.Ops[S]
+
+    QHttpService {
       case req @ GET -> AsDirPath(path) :? QueryParam(query) +& Offset(offset) +& Limit(limit) => {
-        handleOffsetLimitParams(offset, limit) { (offset, limit) =>
-          SQLParser.parseInContext(query, QPath.fromAPath(path)).fold(
-            formatParsingError,
-            expr => queryPlan(addOffsetLimit(expr, offset, limit), vars(req)).run.value.fold(
-              errs => translateSemanticErrors(errs),
+        (offsetOrInvalid[S](offset) |@| limitOrInvalid[S](limit)) { (offset, limit) =>
```
wemrysi: I _think_ you could change this to
```scala
respond((offsetOrInvalid[S](offset) |@| limitOrInvalid[S](limit)) { (offset, limit) => ... }.point[Free[S, ?]])
```
getting rid of the final `toResponse[S]` and `merge`.

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
+          SQLParser.parseInContext(query, QPath.fromAPath(path)).map(
+            expr => queryPlan(addOffsetLimit(expr, offset, limit), vars(req)).run.value.map(
               logicalPlan => {
                 val requestedFormat = MessageFormat.fromAccept(req.headers.get(Accept))
-                formatQuasarDataStreamAsHttpResponse(f)(
-                  data = Q.evaluate(logicalPlan).translate[FileSystemErrT[Free[S, ?], ?]](removePhaseResults),
-                  format = requestedFormat)
+                val formattedData = requestedFormat.encode[FileSystemErrT[Free[S,?], ?]](
```
wemrysi: These 5 lines can be changed to
```scala
formattedDataResponse(
  requestedFormat,
  Q.evaluate(logicalPlan).translate[FileSystemErrT[Free[S, ?], ?]](removePhaseResults)))
```

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
                 absFile => \/-(-\/(absFile)),
-                relDir => -\/(BadRequest("Destination must be a file")),
-                absDir => -\/(BadRequest("Destination must be a file")))(destination.value)
+                relDir => -\/(QuasarResponse.error[S](BadRequest, "Destination must be a file")),
+                absDir => -\/(QuasarResponse.error[S](BadRequest, "Destination must be a file")))(destination.value)
               // Add path of query if destination is a relative file or else just jump through Sandbox hoop
               val absDestination = destinationFile.flatMap(_.bimap(
-                absFile => sandbox(rootDir, absFile).map(file0 => rootDir </> file0) \/> InternalServerError("Could not sandbox file"),
-                relFile => sandbox(currentDir, relFile).map(file0 => currentDir </> file0).map(file1 => path </> file1) \/> InternalServerError("Could not sandbox file")
+                absFile => sandbox(rootDir, absFile).map(file0 => rootDir </> file0) \/>
```
wemrysi: Can remove partiality with `quasar.fs.sandboxAbs`

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
+                  QuasarResponse.error[S](InternalServerError, "Could not sandbox file"),
+                relFile => sandbox(currentDir, relFile).map(file0 =>
```
wemrysi: I'd support adding an analogous `sandboxRel` that sandboxes a relative path onto `currentDir` in a total manner so that we don't have to handle an error that can't happen.

wemrysi: Oh, nevermind, it isn't the same situation as the `relFile` could start with a `../`. Though, in that case, I don't think an `InternalServerError` is appropriate as the user gave a path that is unsupported and can correct it, so maybe a `BadRequest` instead?

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
+                .leftMap(_.toResponse[S])
               val destinationFile = posixCodec.parsePath(
                 relFile => \/-(\/-(relFile)),
                 absFile => \/-(-\/(absFile)),
-                relDir => -\/(BadRequest("Destination must be a file")),
-                absDir => -\/(BadRequest("Destination must be a file")))(destination.value)
+                relDir => -\/(QuasarResponse.error[S](BadRequest, "Destination must be a file")),
+                absDir => -\/(QuasarResponse.error[S](BadRequest, "Destination must be a file")))(destination.value)
               // Add path of query if destination is a relative file or else just jump through Sandbox hoop
               val absDestination = destinationFile.flatMap(_.bimap(
```
wemrysi: I know this wasn't added in this PR, but could `_.fold(...)` here instead of `_.bimap(...).merge`

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
+                absFile => sandbox(rootDir, absFile).map(file0 => rootDir </> file0) \/>
+                  QuasarResponse.error[S](InternalServerError, "Could not sandbox file"),
+                relFile => sandbox(currentDir, relFile).map(file0 =>
+                  currentDir </> file0).map(file1 => path </> file1) \/>
+                    QuasarResponse.error[S](InternalServerError, "Could not sandbox file")
               ).merge)
               val resultOrError = (parseRes |@| absDestination)((expr, out) => {
-                // Unwrap the 3 level Monad Transformer, convert from Free to Task
-                Q.executeQuery(expr, vars(req), out).run.run.run.foldMap(f).flatMap { case (phases, result) =>
+                // Unwrap the 3 level Monad Transformer
```
wemrysi: The rest of the restult should be able to be shortened to something like
```scala
respond(parseRes.tuple(absDestination) traverse { case (expr, out) =>
  Q.executeQuery(expr, vars(req), out).run.run.run map { case (phases, result) =>
    result.map(_.map(rfile => Json(
      "out"    := posixCodec.printPath(rfile),
      "phases" := phases)))
  }
})
```

web/src/main/scala/quasar/api/services/query.scala
wemrysi: Or actually, don't we just want to `sandbox(path, relFile)`?

web/src/main/scala/quasar/api/services/query.scala
```diff
@@ -64,104 +60,103 @@ object query {
...
+              .getOrElse(QuasarResponse.error(InternalServerError,
+                s"No explain output for plan: \n\n" + RenderTree[Fix[LogicalPlan]].render(lp).shows))
+          case (_, -\/(fsErr)) => fsErr.toResponse[S]
         })
 
-    HttpService {
+    QHttpService {
       case req @ GET -> AsDirPath(path) :? QueryParam(query) +& Offset(offset) +& Limit(limit) =>
-        handleOffsetLimitParams(offset, limit) { (offset, limit) =>
+        (offsetOrInvalid[S](offset) |@| limitOrInvalid[S](limit)) { (offset, limit) =>
```
wemrysi: Similar to above, I think you can avoid the `Free.pure` and `_.toResponse` calls by `traverse`-ing twice and wrapping the result in `respond(...)`.

wemrysi:
I think I'd prefer the compile and query services to be in their own files (same for tests) for consistency with the rest of the services. Thoughts?

drostron:
Having compile and query services and respective tests in their own files sounds good.

drostron:
@wemrysi, thanks for the suggestions and comments! I'll put up a commit to address soon.

web/src/main/scala/quasar/api/services/query.scala
jedesah: So I am pretty sure our current implementation is buggy, but the true fix is non trivial and I think it might be out of the scope of this PR. @wemrysi What do you think of adding a TODO? Because short of adding a version `</>` which accepts a non-sandboxed path and returns a non-sandboxed path and fixing `canonicalize` to return an `Option` (and accept a non-sandboxed), I am not sure what is the best way to fix this.

wemrysi: Current implementation of what?

jedesah: Current implementation of the code being commented on. Ah, good point, so then I wonder if there is a function to "unsandbox"

wemrysi: Ah, ok I wasn't sure if you were referring to this code or something in `scala-pathy`.

wemrysi: What is makes the true fix "non-trivial"? (Also, `unsandbox` appears to exist)

jedesah: Well, because we would have to change `canonicalize`.

jedesah: I am okay, with fixing it, just will take a bit more time than adding a TODO and a Jira ticket

jedesah: Okay, nvm, don't need to change `canonicalize`, I think we found an alternative fix

wemrysi: Ok, a TODO is fine then, please also file an issue at the scala-pathy project.

drostron:
@wemrysi, what do you think of placing `query` and `compile` services into a `query` package under `quasar.api.service`?

drostron:
Further, @jedesah just suggested that `query.query` would be `query.execute`.

sellout:
Yay to more packages 👍🏽

wemrysi:
@drostron Sure, sounds good

drostron:
@wemrysi, pushed a couple commits addressing your comments.

web/src/main/scala/quasar/api/services/query/compile.scala
```diff
@@ -0,0 +1,64 @@
...
+    def phaseResultsResponse(prs: PhaseResults): Option[QuasarResponse[S]] =
+      prs.lastOption map {
+        case PhaseResult.Tree(name, value)   => Json(name := value).toResponse
+        case PhaseResult.Detail(name, value) => QuasarResponse.string(Ok, name + "\n" + value)
+      }
+
+    def explainQuery(
+      expr: sql.Expr, offset: Option[Natural], limit: Option[Positive], vars: Variables): Free[S, QuasarResponse[S]] =
+      queryPlan(addOffsetLimit(expr, offset, limit), vars).run.value.fold(
+        _.toResponse[S].point[Free[S, ?]],
```
wemrysi: Could get rid of this line by `traverse`-ing instead of `fold`.

web/src/main/scala/quasar/api/services/query/execute.scala
```diff
@@ -0,0 +1,99 @@
...
+            )
+          )
+        }.point[Free[S, ?]]
+      )
+      case GET -> _ => queryParameterMustContainQuery[S]
+      case req @ POST -> AsDirPath(path) =>
+        Free.liftF(S1.inj(EntityDecoder.decodeString(req))).flatMap { query =>
+          if (query == "") postContentMustContainQuery[S]
+          else {
+            respond(requiredHeader[S](Destination, req).map { destination =>
```
wemrysi: I'd prefer `traverse(...)` to `map(...).sequence(...)`

wemrysi:
A couple of minor comments, but nothing that holds up the merge, so I'll defer to you on them. LGTM :+1:

wemrysi:
Also, any idea where coverage went?

drostron:
I see [some coverage](https://github.com/quasar-analytics/quasar/pull/1081#issuecomment-180911936) info. Did there used to be something additional?

wemrysi:
Aye, I see that, but the comments say there is no coverage info for this branch, i.e. no info on how it changes coverage. Not worth holding up the merge, but was just curious why it wasn't working.

drostron:
Ah, indeed. I'm also now curious. Seems like there isn't any difference from our normal PRs other than PRing into a topic branch. No idea if that effects it or not (_would have to look into it_).

web/src/main/scala/quasar/api/services/query/compile.scala
drostron: Done

web/src/main/scala/quasar/api/services/query/execute.scala
drostron: Converted to `traverse`.

wemrysi:
Good point, that might be it. I don't remember if it ever worked for other PRs to topic branches.

PR #1082, opened 2016-02-07
Replace homegrown Natural and Positive with refinement types provided by refined

jedesah:


jedesah:
@wemrysi and @drostron Finally got around to rebasing this work on master and addressing some of @wemrysi's final review comments!! :smile:

jedesah:
@wemrysi btw, fixed the issues with the CI build.

wemrysi:
@jedesah I haven't forgotten about this, but mind if it waits until next week after 2.5?

jedesah:
@wemrysi No problem! I will have to resolve conflicts anyway now that we finally merged the other PR. Also, if scala 2.11.8 gets release in the meantime (probably not) that would allow me to get rid of many workarounds and comments.

core/src/main/scala/quasar/fp/numeric/package.scala
```diff
@@ -0,0 +1,52 @@
...
+  implicit def widenPositive[F[_,_],N](a: F[Int,Greater[N]])(implicit rt: RefType[F]): F[Long,Greater[N]] =
+    rt.unsafeWrap(rt.unwrap(a).toLong)
+
+  implicit def widenNatural[F[_,_]](a: F[Int, NonNegative])(implicit rt: RefType[F]): F[Long,NonNegative] =
+    rt.unsafeWrap(rt.unwrap(a).toLong)
+
+  implicit def positiveToNatural[F[_,_], A](a: F[A,RPositive])(implicit rt: RefType[F]): F[A, NonNegative] =
+    rt.unsafeWrap(rt.unwrap(a))
+
+  implicit def monoid[F[_,_],T](implicit rt: RefType[F], num: scala.Numeric[T]): Monoid[F[T,NonNegative]] =
```
wemrysi: I the following instance names should be prefixed since they're in the package object and have a much greater chance of conflicting with other methods of the same name when imported.

core/src/main/scala/quasar/fp/numeric/package.scala
```diff
@@ -0,0 +1,52 @@
...
+import eu.timepit.refined.api.{RefType, Refined}
+import scalaz.{Equal, Show, Monoid}
+import scalaz.syntax.show._
+
+package object numeric {
+
+  type Natural = Long Refined NonNegative
+  type Positive = Long Refined RPositive
+
+  def Positive(a: Long): Option[Positive] = refineV[RPositive](a).right.toOption
```
wemrysi: What do you think about having these be aliases for the "fully applied" macro class and then adding an ops class that allows us to call the `refineV` version on them? Something like
```scala
val Natural = new RefineMFullyApplied[Refined, Long, NonNegative]

implicit final case class RefineMFullyAppliedOps[F[_, _], T, P](rmf: RefineMFullyApplied[F, T, P]) extends scala.AnyVal {
  // Could come up with a different name here, or not
  val refineV: RefinePartiallyApplied[F, P] = RefType[F].refine[P]
}
```
This should allow for using the macro for literals, `Natural(42)`, and the "value" method otherwise, `Natural.refineV(someParam)` (assuming the ops class works).

core/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -213,10 +214,18 @@ object InMemory {
         import quasar.LogicalPlan._
         import quasar.std.StdLib.set.{Drop, Take}
         import quasar.std.StdLib.identity.Squash
+        // Side-step https://issues.scala-lang.org/browse/SI-9581 by avoiding values of limit
```
wemrysi: This seems like an odd place to workaround this issue. Is the problem that we're generating values of `Int.MaxValue` in tests via scalacheck? If so I'd just limit the size there, if that's not it then what is being guarded against here?

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
```diff
@@ -193,17 +197,21 @@ class CompileAndQueryServiceSpec extends Specification with FileSystemFixture wi
...
-          path = filesystem.parent,
-          query = Some(Query(query, offset = Some(offset), limit = Some(limit), varNameAndValue = Some((varName.value,var_.toString)))),
-          state = filesystem.state.copy(queryResps = Map(lp -> filesystem.contents)),
-          status = Status.Ok,
-          response = (a: String) => a must_==
-            jsonReadableLine.encode(Process.emitAll(filesystem.contents): Process[Task, Data]).runLog.run
-              .drop(offset.value.toInt).take(limit.value.toInt).mkString("")
-        )
+      "GET" ! prop { (filesystem: SingleFileMemState, varName: AlphaCharacters, var_ : Int, offset: Int Refined NonNegative, limit: Int Refined RPositive) =>
+        // Side-step https://issues.scala-lang.org/browse/SI-9581 by avoiding values of limit
```
wemrysi: Seems like maybe we want a newtyped `Int` that we can limit generation for? Then we could just document it once and use it everywhere. Also will make it easier to revert once the bug in collections is fixed.

core/src/main/scala/quasar/fp/numeric/package.scala
jedesah: Good point.

core/src/main/scala/quasar/fp/numeric/package.scala
jedesah: I don't think it's all that useful because refined automatically lifts a literal into a refined type and so this just makes lifting a non-literal more verbose. But I made the change just to test it out so if you feel strongly that this is a better approach I can make the change.

core/src/main/scala/quasar/fs/InMemory.scala
jedesah: I think this is the right place to address this issue. However your comment made me realize that it would be more correct to return a `FileSystemError` instead of silently using a smaller value.

web/src/test/scala/quasar/api/services/CompileAndQueryServiceSpec.scala
jedesah: That's a good idea.

jedesah: Well. Actually there is only two places. I say it's not worth the effort for something that is scheduled to be released in 1 week.

core/src/main/scala/quasar/fp/numeric/package.scala
wemrysi: Ah, yeah you're right, ok.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
```diff
@@ -16,20 +16,29 @@
 
 package quasar.fs
 
+import org.specs2.ScalaCheck
```
wemrysi: Move down to our conventional import section

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
```diff
@@ -108,44 +116,42 @@ class ReadFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT) {
...
-        val off = (Natural._5 * Natural._5 * Natural._4) + Natural._1
-        val r = runLogT(run, read.scan(smallFile.file, off, None))
+        val r = runLogT(run, read.scan(smallFile.file, smallFileSize |+| 1L, None))
         r.runEither must beRight((xs: scala.collection.IndexedSeq[Data]) => xs must beEmpty)
       }
 
-      "scan with offset k > 0 and no limit skips first k data" >> {
-        val k = Natural._9 * Natural._2
-        val r = runLogT(run, read.scan(smallFile.file, k, None))
+      "scan with offset k > 0 and no limit skips first k data" ! prop { k: Int Refined RPositive =>
```
wemrysi: Why not just generate a `Positive`?

wemrysi: Oh, nm I see you're converting positive to natural? I thought refined already supported using a stronger condition where a weaker one was expected

jedesah: So `Positive` is a `Long` but due to how we are testing the property (using scala collections), we can't really test all values of `Long` since the functions on `List` accept `Int`s. So I ask for an `Int` that is `Positive`(`RPositive`) and `widenPositive` into a `Positive` that `read.scan` accepts. This happens often in the tests and was an important reason why I thought it would be nice to use refined.

jedesah: Refined does have a mechanism to automatically convert a `refined.numeric.Positive` into a `refined.numeric.Natural`.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
```diff
@@ -108,44 +116,42 @@ class ReadFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT) {
...
+                  .dropWhile(_._2 < k.get).map(_._1)
 
         r.runEither must beRight(d.toIndexedSeq)
-      }
+      }.set(minTestsOk = 10)
 
-      "scan with offset zero and limit j stops after j data" >> {
-        val j = Positive._5
-        val r = runLogT(run, read.scan(smallFile.file, Natural._0, Some(j)))
+      "scan with offset zero and limit j stops after j data" ! prop { j: Int Refined RPositive =>
```
wemrysi: Can you limit `j` to a number only slightly larger than `smallFileSize`? The chance of generation actually getting a number in `[1, smallFileSize]` is pretty small.

jedesah: What's so special about this test though, doesn't the same logic apply to other tests in this file?

jedesah: Ah,I guess it's because of the wording of the example.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
wemrysi: Hm, it doesn't seem like we should be modifying the spec because of a particular implementation. Also, how likely is it that we get an interesting case where `0 < k < smallFilesize`? Won't the vast majority of cases be `smallFileSize < k`?

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
wemrysi: Yeah, I think it does. We can't use arbitrarily sized files for performance reasons so using arbitrarily sized indexes are going to be not very interesting most of the time, I think.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
jedesah: > it doesn't seem like we should be modifying the spec because of a particular implementation

I don't think we should modify the specification text (if applicable) just because of a limitation in expressing a property, but I am not sure what you mean about not modifying the spec, there is no easy way to express a property with `Long`s, I am not sure how to solve that. Just to be clear, we are asking for an Int, not because the `InMemory` interpreter does not support values of `Long` (although it doesn't) but because the code expressing our property does not support values of `Long`.

jedesah: > Also, how likely is it that we get an interesting case where 0 < k < smallFilesize?

I would say just as likely as `k > smallFilesize`. I think the interesting cases here are:
- `k = 1`
- `0 < k < smallFilesize`
- `k = smallFilesize`
- `k > smallFilesize`
- `k = max of domain`

As long as we stand a good chance of hitting each of these at least once, I think we are good. My understanding of scalacheck is that we do stand a good chance of hitting each one of these, but of course I am not sure.

wemrysi: Ah! Ok, I see what you mean now, thanks for the explanation.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
jedesah: > We can't use arbitrarily sized files for performance reasons

I think we have discussed this previously but my personal preference would be to express the property in a "pure" way, i.e. an arbitrary file and provide a generator that takes into account the practicalities of performance. I won't make the change as part of this PR but eventually it would be nice to move towards that in this file.

>so using arbitrarily sized indexes are going to be not very interesting most of the time`

I am not sure that is true. I am pretty sure scalacheck's default generators for range and such don't just generate values equally distributed within the range. I think it will favor values close to the extremities which are interesting cases for us.

core/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -208,18 +209,24 @@ object InMemory {
...
         import quasar.std.StdLib.set.{Drop, Take}
         import quasar.std.StdLib.identity.Squash
-        Recursive[Fix].para[LogicalPlan, Option[Vector[Data]]](lp) {
-          case ReadF(path) => path.asAFile.flatMap(pathyPath => fileL(pathyPath).get(mem))
-          case InvokeF(Drop, (_,src) :: (Fix(ConstantF(Data.Int(skip))),_) :: Nil) => src.map(_.drop(skip.toInt))
-          case InvokeF(Take, (_,src) :: (Fix(ConstantF(Data.Int(limit))),_) :: Nil) => src.map(_.take(limit.toInt))
+        // Side-step https://issues.scala-lang.org/browse/SI-9581 by avoiding values of limit
+        // that are close to Int.MaxValue
+        // TODO: Change back to Int.MaxValue after upgrade to Scala 2.11.8
+        val safeIntMax = 10000000
```
wemrysi: I still think this should be handled via a newtype + centralized constant as this particular constant appears in three different places in the code (also, is there a reason for this particular constant? If so mind documenting it?). If we had a `SafeIntForVector` type or somesuch, we could use it to generate safe values.

jedesah: I think we are spending too much time wondering about something that will be fixed in a week and that I fully intend to clean up immediately. Frankly I would just wait for scala 2.11.8 before merging if I wasn't so tired of constantly rebasing this PR. If you feel strongly this effort is worth it I will do it in order to get this PR finally merged. Where do you suggest I define `SafeIntForVector`?

There is no particular reason for `10000000` (is that the right amount of zeros?) aside from the fact that it seems sufficiently far away from `Int.MaxValue` to not trigger the bug even if we add some reasonable number to it and yet far enough away from `0` to useful.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
wemrysi: I think you're right about the `k = 1`, `k > smallFileSize` and `k = max of domain` cases, it's the other two that concern me as the generator doesn't know anything about `smallFileSize`. If you're confident that we'll get sufficient coverage there, then sounds good.

jedesah: Ah, okay, I understand what you mean now. `o < k < smallFilesize` is probably okay because it will be hit by the values close to `1` but it's rather unlikely we will hit `k = smallFilesize`. I wish scalacheck had a way for us to say "here is the general property and here are some interesting cases I can come up with myself, please test them out along with other values of your choosing in the domain of my input types".

core/src/main/scala/quasar/fs/InMemory.scala
wemrysi: I hear you and my goal isn't to hold this up unecessarily. How about we just define the constant itself someplace (maybe in `quasar.fp.numeric`?) with a comment of why it exists and the link to the issue and then just reference it everywhere you're using the literal? Then it is at least in one place and we'll see a big ugly name everywhere it is used.

jedesah: That sounds good.

jedesah:
@wemrysi I think this is finally ready to be merged!

wemrysi:
:+1:

PR #1084, opened 2016-02-09
SD-1327 Refactor quasar web to expose routes as a program in Free

drostron:


core/src/main/scala/quasar/errors.scala
```diff
@@ -51,6 +52,12 @@ object Errors {
     liftMT[Task, G]
   }
 
+  /** Flatten by inserting the `E` into the failure case of `Task` */
+  def flatten[E:Show] = new (ETask[E,?] ~> Task) {
+    def apply[A](t: ETask[E,A]): Task[A] =
+      t.fold(e => Task.fail(new RuntimeException(e.shows)), Task.now).join
+  }
+
```
sellout: This scares me. Why do we need this?

core/src/test/scala/quasar/fs/InMemoryArbitrary.scala
```diff
@@ -30,3 +30,5 @@ trait InMemoryArbitrary {
   implicit val arbitraryInMemState: Arbitrary[InMemState] =
     Arbitrary(arb[Map[AFile, Vector[Data]]].map(InMemState(0, _, Map(), Map(), Map(), Map())))
 }
+
+object InMemoryArbitrary extends InMemoryArbitrary
```
sellout: How is it even possible to make a file without a final newline?

web/src/main/scala/quasar/api/Http4sUtils.scala
```diff
@@ -0,0 +1,174 @@
...
+import shapeless._
+import shapeless.Nat
+import shapeless.ops.nat._
+import shapeless.nat._
+
+object Http4sUtils {
+
+//  sealed trait BindingType
+//  case object HTTP extends BindingType
+//  case object Local extends BindingType
```
sellout: Are these obsolete or aspirational?

core/src/test/scala/quasar/fs/InMemoryArbitrary.scala
drostron: Haha, exactly!

core/src/main/scala/quasar/errors.scala
jedesah: Basically, we need this at the edge of the world. We need it for Http Streaming and maybe somewhere else. I think it's probably a good idea to move it to `web` as I don't see a good reason why it should be defined (or used!) in `core`.

web/src/main/scala/quasar/api/Http4sUtils.scala
jedesah: We will probably need them if we ever want to support restarting the server on a new port again in quasar-advanced.

jedesah: But I think we can get rid of it for now.

drostron:
@sellout, pushed a few commits addressing comments.

sellout:
Ok, this looks good. The new approach is nice.

The review was a bit more cursory than I like, but with three of you already collaborating, I’m not too concerned about that.

PR #1085, opened 2016-02-10
Fix handling of `IN` for single-element sets.

sellout:
SD-1344 #done

In SQL²/MRA,  a value and a single-element set have the same
semantics. This makes that happen for the `IN` operator, which has been
a long-standing problem for the frontend.

### DIFF
--- a/core/src/main/scala/quasar/std/set.scala
+++ b/core/src/main/scala/quasar/std/set.scala
@@ -18,7 +18,7 @@ package quasar.std
 
 import quasar.Predef._
 import quasar.fp._
-import quasar.recursionschemes._
+import quasar.recursionschemes._, Recursive.ops._
 import quasar._, LogicalPlan._
 
 import scalaz._, Scalaz._, NonEmptyList.nel, Validation.{success, failure}
@@ -56,7 +56,7 @@ trait SetLib extends Library {
   val Drop = Sifting("(OFFSET)", "Drops the first N elements from a set",
     Type.Top, Type.Top :: Type.Int :: Nil,
     new Func.Simplifier {
-      def apply[T[_[_]]: Recursive: FunctorT](orig: LogicalPlan[T[LogicalPlan]]) = orig match {
+      def apply[T[_[_]]: Recursive: Corecursive](orig: LogicalPlan[T[LogicalPlan]]) = orig match {
         case IsInvoke(_, List(set, ConstantF(Data.Int(n)))) if n == 0 =>
           set.some
         case _ => None
@@ -80,7 +80,7 @@ trait SetLib extends Library {
   val Filter = Sifting("WHERE", "Filters a set to include only elements where a projection is true",
     Type.Top, Type.Top :: Type.Bool :: Nil,
     new Func.Simplifier {
-      def apply[T[_[_]]: Recursive: FunctorT](orig: LogicalPlan[T[LogicalPlan]]) =
+      def apply[T[_[_]]: Recursive: Corecursive](orig: LogicalPlan[T[LogicalPlan]]) =
         orig match {
           case IsInvoke(_, List(set, ConstantF(Data.True))) => set.some
           case _                                            => None
@@ -195,7 +195,7 @@ trait SetLib extends Library {
     "Removes the elements of the second set from the first set.",
     Type.Top, Type.Top :: Type.Top :: Nil,
     new Func.Simplifier {
-      def apply[T[_[_]]: Recursive: FunctorT](orig: LogicalPlan[T[LogicalPlan]]) = orig match {
+      def apply[T[_[_]]: Recursive: Corecursive](orig: LogicalPlan[T[LogicalPlan]]) = orig match {
         case IsInvoke(_, List(set, ConstantF(Data.Set(Nil)))) => set.some
         case _                                                => None
       }
@@ -203,16 +203,46 @@ trait SetLib extends Library {
     setTyper(partialTyper { case List(s1, _) => s1 }),
     setUntyper(t => success(t :: Type.Top :: Nil)))
 
+  // TODO: Handle “normal” functions without creating Funcs. They should be in
+  //       a separate functor and inlined prior to getting this far. It will
+  //       also allow us to make simplification non-Corecursive and ∴ operate
+  //       on Cofree.
   val In = Mapping(
     "(in)",
     "Determines whether a value is in a given set.",
     Type.Bool, Type.Top :: Type.AnySet :: Nil,
-    noSimplification,
+    new Func.Simplifier {
+      def apply[T[_[_]]: Recursive](orig: LogicalPlan[T[LogicalPlan]])(implicit T: Corecursive[T]) =
+        orig match {
+          case InvokeF(_, List(item, set)) => set.project match {
+            case ConstantF(Data.Set(_)) => Within(item, T.embed(StructuralLib.UnshiftArray(set))).some
+            case ConstantF(_)           => RelationsLib.Eq(item, set).some
+            case _                      => None
+          }
+          case _ => None
+        }
+    },
     partialTyper {
       case List(_,             Type.Const(Data.Set(Nil))) =>
         Type.Const(Data.Bool(false))
       case List(Type.Const(x), Type.Const(Data.Set(set))) =>
         Type.Const(Data.Bool(set.contains(x)))
+      case List(Type.Const(x), Type.Const(y))             =>
+        Type.Const(Data.Bool(x == y))
+      case List(_,             _)                         => Type.Bool
+    },
+    basicUntyper)
+
+  val Within = Mapping(
+    "within",
+    "Determines whether a value is in a given array.",
+    Type.Bool, Type.Top :: Type.AnyArray :: Nil,
+    noSimplification,
+    partialTyper {
+      case List(_,             Type.Const(Data.Arr(Nil))) =>
+        Type.Const(Data.Bool(false))
+      case List(Type.Const(x), Type.Const(Data.Arr(arr))) =>
+        Type.Const(Data.Bool(arr.contains(x)))
       case List(_,             _)                         => Type.Bool
     },
     basicUntyper)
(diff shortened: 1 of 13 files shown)


jedesah:
@sellout Does this need to get into `2.5`?

sellout:
Yes.

jedesah:
ah man..., okay I will have a look at it.

jedesah:
@sellout You sure this is SD-1344? It says jonsterling is assigned and progress has not started, not to mention it does not seem obviously related.

core/src/main/scala/quasar/std/set.scala
```diff
@@ -203,16 +203,46 @@ trait SetLib extends Library {
...
+      case List(Type.Const(x), Type.Const(y))             =>
+        Type.Const(Data.Bool(x == y))
+      case List(_,             _)                         => Type.Bool
+    },
+    basicUntyper)
+
+  val Within = Mapping(
+    "within",
+    "Determines whether a value is in a given array.",
+    Type.Bool, Type.Top :: Type.AnyArray :: Nil,
```
jedesah: Is `Any` in `Array[Any]` ?!? :scream:

core/src/main/scala/quasar/std/set.scala
```diff
@@ -203,16 +203,46 @@ trait SetLib extends Library {
...
+    noSimplification,
+    partialTyper {
+      case List(_,             Type.Const(Data.Arr(Nil))) =>
+        Type.Const(Data.Bool(false))
```
jedesah: Oh man, this is the exact same logic as `IN`, I wonder if there might be a way define the logic once and reuse it for `IN` and `WITHIN`. I kind of wonder why we need `IN` in the first place, seems like it would be nice to have `IN` work on both arrays and sets. Maybe this is explained in the spec for this ticket which I have reason to believe is incorrect in the PR description.

core/src/main/scala/quasar/std/set.scala
sellout: This is not Scala’s `Any`, this is just the name in Quasar’s type system for an Array containing anything (which has to be heterogeneous, like Scala’s `Array[Any]`, because that is the terrible decision made by NoSQL stores).

sellout:
That is the right ticket. Look at @jdegoes comments.

core/src/main/scala/quasar/std/set.scala
jedesah: I know (well not about the terrible NoSQL stores), just giving you a hard time! :smile:

core/src/main/scala/quasar/std/set.scala
sellout: Sets are special and have distinct semantics. E.g., a bare value and a single-element set have the same semantics, but that’s not true for a single-element array. Also, sets are part of the MRA model, and not a value type to be stored in a collection.

The duplication here isn’t ideal, but it’s a consequence of the order things currently happen in. We don’t have type information at the simplification stage, and so there has to be some duplication here. The comment above `In` (which should be added as an issue) roughly describes the solution.

core/src/main/scala/quasar/std/set.scala
```diff
@@ -203,16 +203,46 @@ trait SetLib extends Library {
     setTyper(partialTyper { case List(s1, _) => s1 }),
     setUntyper(t => success(t :: Type.Top :: Nil)))
 
+  // TODO: Handle “normal” functions without creating Funcs. They should be in
```
jedesah: Not sure I totally understand this comment. Do you mean that we would want to eventually create a new "AST" Node type (do we still call it an AST even in fixed point style?) for a user-defined function?

sellout: Yeah, something we would `Coproduct` with `LogicalPlan` for desugaring. with an algebra like
```
UserFunc :+: LogicalPlan ~> LogicalPlan
```
that inlines those callsites.

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -310,7 +310,12 @@ trait StructuralLib extends Library {
     "Unshifts an integral dimension from the set identity, creating an array with the dimensional values as the indices.",
     AnyArray, Set(Top) :: Nil,
     noSimplification,
-    partialTyper { case List(tpe) => FlexArr(0, None, tpe) },
+    partialTyper {
+      case List(Const(Data.Set(vs))) => Const(Data.Arr(vs))
+      case List(Const(v))            => Const(Data.Arr(List(v)))
+      case List(Set(tpe))            => FlexArr(0, None, tpe)
```
jedesah: We don't really need a fourth case here but I guess it doesn't hurt.

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -310,7 +310,12 @@ trait StructuralLib extends Library {
     "Unshifts an integral dimension from the set identity, creating an array with the dimensional values as the indices.",
     AnyArray, Set(Top) :: Nil,
     noSimplification,
-    partialTyper { case List(tpe) => FlexArr(0, None, tpe) },
+    partialTyper {
+      case List(Const(Data.Set(vs))) => Const(Data.Arr(vs))
+      case List(Const(v))            => Const(Data.Arr(List(v)))
```
jedesah: I have mentioned this before and I think you reassured me that we would change how we handle this in the future, but this approach to making an single element equivalent to a single element set does not scale.

core/src/main/scala/quasar/std/structural.scala
sellout: We do need the `List(Set(tpe))` case – it extracts the type from the set.

core/src/main/scala/quasar/std/structural.scala
sellout: Agreed. I hope I didn’t reassure you too much 😉 Handling it correctly just relies on having MRA.

core/src/test/scala/quasar/matchers.scala
```diff
@@ -40,13 +40,13 @@ trait TermLogicalPlanMatchers {
   case class equalToPlan(expected: Fix[LogicalPlan])
       extends Matcher[Fix[LogicalPlan]] {
     def apply[S <: Fix[LogicalPlan]](s: Expectable[S]) = {
-      val normed = FunctorT[Cofree[?[_], Fix[LogicalPlan]]].transCata(Recursive[Fix].cata(s.value)(attrSelf))(repeatedly(Optimizer.simplifyƒ[Cofree[?[_], Fix[LogicalPlan]]]))
-      val diff = (Recursive[Cofree[?[_], Fix[LogicalPlan]]].convertTo(normed).render diff expected.render).shows
+      val normed = s.value.transCata(repeatedly(Optimizer.simplifyƒ))
```
jedesah: Haha, I like this way better!

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -703,7 +703,7 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
             Selector.In(Bson.Arr(List(Bson.Text("AZ"), Bson.Text("CO"))))))))
     }
 
-    "plan filter with field in empty array" in {
```
jedesah: Is it possible line 698 should be set as well?

core/src/test/scala/quasar/matchers.scala
sellout: The issue is that it loses the original de-normalized form, so failure reporting doesn’t necessarily match up with what’s written in the test expectation. I’d like to get Cofree back, but it relies on that change re `UserFunc` described earlier.

core/src/test/scala/quasar/physical/mongodb/planner.scala
sellout: Yep!

core/src/main/scala/quasar/std/structural.scala
jedesah: Ah, okay, how did it work before?

jedesah:
LGTM.

core/src/main/scala/quasar/std/structural.scala
sellout: It didn’t. This operation is currently not implemented on any backend (read: Mongo). So, the constant folding “evaluator” was the first test case.

jedesah: So I am not crazy! Good to know.

sellout: ((a → b) ∧ ¬a) ↛ ¬b

😉

jedesah: I thought math was my friend! :sob:

sellout: I think there’s a joke here about the axiom of choice …

core/src/main/scala/quasar/std/structural.scala
jdegoes: Post-MRA, sets are deleted from the type system, literals are auto-lifted into singleton sets, and so forth, which should clean up a lot of things.

it/src/main/resources/tests/singleElementSet.test
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "filter field in single-element set",
+    "data": "zips.data",
+    "query": "select * from zips where state in ('MA') and pop < 10",
```
jdegoes: This is so nice!

it/src/main/resources/tests/inBareElement.test
```diff
@@ -0,0 +1,9 @@
+{
```
jdegoes: I am just curious: is this expected to work in this PR: `WHERE foo IN [1, 2, 3][*]`?

sellout: It does (as it already had) … but through careful omission. If I can’t constant-fold the `[set...]`, I don’t rewrite to `within`, and Mongo currently treats sets as arrays. This isn’t ideal, but `[...]` isn’t implemented on Mongo yet. I have a hunch as to how to, but I didn’t when I wrote this code.

core/src/main/scala/quasar/std/structural.scala
jedesah: :+1:

PR #1086, opened 2016-02-10
Simplify build and avoid a thrid re-compilation

jedesah:


sellout:
I like this change anyway, but I only see two builds triggered by each. The nice bit is that the new one should fail faster – by testing before we do the second build – and also avoids moving the JAR around.

However, I have two concerns. When I tried removing `coverageReport` before, I ended up getting no comment from Codecov. Also, are you sure that the oneJar build after having generated coverage info doesn’t contain anything extra (e.g., the coverage report).

jedesah:
@sellout My understanding observing the previous build is that we were compiling a third time because `githubRelease` was recompiling (because last time it compiled it was with coverage enabled). This way, we recompile twice instead of three times. Do you know something I don't?

jedesah:
@sellout I am not absolutely sure that the jar does not contain extra information, but we know that the code should not contain instrumentation as sbt feels the need to re-compile.

jdegoes:
Initially, the one jar wouldn't even run when compiled with coverage. I'm not sure if that's changed. In any case, we may not inflate the size of the end-user onejar anymore than it already is. In fact, soon we will introduce ProGuard to shrink the "production jar" further.

sellout:
Yeah, agreed that the jar doesn’t contain instrumentation. I know you had checked for the recompile.

If `githubRelease` was triggering a rebuild, that was definitely a bug somewhere. The intent was that the jar created by `oneJar` (and then tested by `testJar`) was the same artifact uploaded by `githubRelease`. Granted, very possible that that broke at some point, but if we’re testing the jar and then throwing it away, definitely a bug.

jedesah:
Can't disagree with that.

wemrysi:
RIP coverage =(

drostron:
Haha, I think @sellout foretold of this.

wemrysi:
:+1:, merge at-will once coverage is restored.

PR #1087, opened 2016-02-11
Refactor how "finalization" of qservices is done

wemrysi:
By "finalization" I mean the collapsing of the various individual api services into a single service that routes requests via the path and adds our standard middleware, feel free to suggest a better name for this.

* Separates choice of what to finalize from finalization itself
* Move api mounting out of the server, update ServerBlueprint
* Take advantage of composition to construct default middleware

Most of these changes are to make existing "server" functionality easier to reuse in quasar-advanced.

I think there is more we could do to this end, but I didn't want to risk disrupting too much before the 2.5 release. For example, I'm not sure we need the `Prefix` middleware we're currently using and could instead use `org.http4s.server.Router`, but wasn't sure if we had anything that depended on the way `Prefix` mangles paths.

### DIFF
--- a/web/src/main/scala/quasar/api/Http4sUtils.scala
+++ b/web/src/main/scala/quasar/api/Http4sUtils.scala
@@ -36,9 +36,7 @@ import shapeless.nat._
 
 object Http4sUtils {
 
-  final case class ServerBlueprint(port: Int,
-                                   idleTimeout: Duration,
-                                   svcs: Map[String, HttpService])
+  final case class ServerBlueprint(port: Int, idleTimeout: Duration, svc: HttpService)
 
   // Lifted from unfiltered.
   // NB: available() returns 0 when the stream is closed, meaning the server
@@ -90,24 +88,12 @@ object Http4sUtils {
     * @param flexibleOnPort Whether or not to choose an alternative port if requested port is not available
     * @return Server that has been started along with the port on which it was started
     */
-  def startServerFromBlueprint(blueprint: ServerBlueprint, flexibleOnPort: Boolean): Task[(Http4sServer, Int)] = {
+  def startServer(blueprint: ServerBlueprint, flexibleOnPort: Boolean): Task[(Http4sServer, Int)] = {
     for {
       actualPort <- if (flexibleOnPort) choosePort(blueprint.port) else Task.now(blueprint.port)
-      builder = BlazeBuilder.withIdleTimeout(blueprint.idleTimeout).bindHttp(actualPort, "0.0.0.0")
-      server <- startServer(blueprint.svcs, builder)
+      builder    =  BlazeBuilder.withIdleTimeout(blueprint.idleTimeout).bindHttp(actualPort, "0.0.0.0")
+      server     <- builder.mountService(blueprint.svc).start
     } yield (server, actualPort)
-  }
-
-  /** Start `Server` of services with supplied `ServerBuilder`
-    * @param services Map of `HttpService` indexed by where to mount them
-    * @return A running `org.http4s.server.Server`
-    */
-  def startServer(services: Map[String, HttpService], builder: ServerBuilder): Task[Http4sServer] = {
-    val builderAll = services.toList.reverse.foldLeft(builder) {
-      // TODO: Consider getting rid of Prefix in favor of http4s prefix
-      case (b, (path, svc)) => b.mountService(Prefix(path)(svc))
-    }
-    Task.delay(builderAll.run)
   }
 
   /** Given a `Process` of [[ServerBlueprint]], returns a `Process` of `Server`.
@@ -123,7 +109,7 @@ object Http4sUtils {
   def servers(configurations: Process[Task, ServerBlueprint], flexibleOnPort: Boolean): Process[Task, (Http4sServer,Int)] = {
 
     val serversAndPort = configurations.evalMap(conf =>
-      startServerFromBlueprint(conf, flexibleOnPort).onSuccess { case (_, port) =>
+      startServer(conf, flexibleOnPort).onSuccess { case (_, port) =>
         stdout("Server started listening on port " + port) })
 
     serversAndPort.evalScan1 { case ((oldServer, oldPort), newServerAndPort) =>
@@ -138,18 +124,20 @@ object Http4sUtils {
 
   /** Produce a stream of servers that can be restarted on a supplied port
     * @param initialPort The port on which to start the initial server
-    * @param produceRoutes A function that given a function to restart a server on a new port, supplies a server mapping
-    *                      from path to `Server`
+    * @param produceService A function that given a function to restart a server on a new port,
+    *                       returns an `HttpService`
     * @return The `Task` will start the first server and provide a function to shutdown the active server.
     *         It will also return a process of servers and ports. This `Process` must be run in order for servers
     *         to actually be started and stopped. The `Process` must be run to completion in order for appropriate
     *         clean up to occur.
     */
-  def startServers(initialPort: Int,
-                   produceRoutes: (Int => Task[Unit]) => Map[String, HttpService]): Task[(Process[Task, (Http4sServer,Int)], Task[Unit])] = {
+  def startServers(
+    initialPort: Int,
+    produceService: (Int => Task[Unit]) => HttpService
+  ): Task[(Process[Task, (Http4sServer,Int)], Task[Unit])] = {
     val configQ = async.boundedQueue[ServerBlueprint](1)
     def startNew(port: Int): Task[Unit] = {
-      val conf = ServerBlueprint(port, idleTimeout = Duration.Inf, produceRoutes(startNew))
+      val conf = ServerBlueprint(port, idleTimeout = Duration.Inf, produceService(startNew))
       configQ.enqueueOne(conf)
     }
     startNew(initialPort).flatMap(_ => servers(configQ.dequeue, false).unconsOption.map {
@@ -158,8 +146,8 @@ object Http4sUtils {
     })
   }
 
-  def startAndWait(port: Int, routes: (Int => Task[Unit]) => Map[String, HttpService], openClient: Boolean): Task[Unit] = for {
-    result <- startServers(port, routes)
+  def startAndWait(port: Int, service: (Int => Task[Unit]) => HttpService, openClient: Boolean): Task[Unit] = for {
+    result <- startServers(port, service)
     (servers, shutdown) = result
     _ <- if(openClient) openBrowser(port) else Task.now(())
     _ <- stdout("Press Enter to stop.")
--- a/web/src/main/scala/quasar/api/services/RestApi.scala
+++ b/web/src/main/scala/quasar/api/services/RestApi.scala
@@ -29,16 +29,14 @@ import scala.collection.immutable.ListMap
 import org.http4s
 import org.http4s.Request
 import org.http4s.dsl._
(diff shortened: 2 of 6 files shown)


web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -47,44 +45,60 @@ final case class RestApi(defaultPort: Int, restart: Int => Task[Unit]) {
...
-object RestApi {
-  def withDefaultMiddleware(service: HttpService): HttpService =
-    cors(GZip(HeaderParam(service orElse HttpService {
-      case req if req.method == OPTIONS => Ok()
-    })))
+  /** Converts `QHttpService`s into `HttpService`s and mounts them along with
+    * any additional `HttpService`s provided, applying the default middleware
+    * to the result.
+    *
+    * TODO: Is using `Prefix` necessary? Can we replace with
```
drostron: 👍🏽

jedesah:
@wemrysi If I remember correctly, it's likely that not only do we not need the `Prefix` middleware, but we probably don't even need `Router` as the version of http4s we are now using has an optional `prefix` argument when mounting an `HttpService`.

I really do wish though that we could create an `HttpService` which represents the whole server instead of carrying around `Routes` or  `Map[String, HttpService]`. It seems like it would be more elegant, but maybe there are performance implications.

drostron:
"finalization" seems sufficiently clear.

web/src/main/scala/quasar/api/Http4sUtils.scala
```diff
@@ -90,26 +88,14 @@ object Http4sUtils {
...
+      builder    =  BlazeBuilder.withIdleTimeout(blueprint.idleTimeout).bindHttp(actualPort, "0.0.0.0")
+      server     <- builder.mountService(blueprint.svc).start
     } yield (server, actualPort)
   }
 
-  /** Start `Server` of services with supplied `ServerBuilder`
-    * @param services Map of `HttpService` indexed by where to mount them
-    * @return A running `org.http4s.server.Server`
-    */
-  def startServer(services: Map[String, HttpService], builder: ServerBuilder): Task[Http4sServer] = {
```
jedesah: How are you planning on starting the server in quasar-advanced?

wemrysi:
@jedesah That's what `Router` does, you give it a bunch of routes and you get back a single `HttpService`, which I think is better than using the `mountService` method on the `ServerBuilder` anyhow (there is some talk of deprecating that method, see https://github.com/http4s/http4s/issues/369).

jedesah:
> That's what Router does

Yes, I just got to the part of the PR where I was realizing that. That's cool, I had been looking for something like that. Happy to see you found it!

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -47,44 +45,60 @@ final case class RestApi(defaultPort: Int, restart: Int => Task[Unit]) {
...
+    */
+  def finalizeServices[S[_]: Functor](
+    f: S ~> ResponseOr)(
+    qsvcs: Map[String, QHttpService[S]],
+    hsvcs: Map[String, HttpService]
+  ): HttpService = {
+    val allSvcs = qsvcs.mapValues(_.toHttpService(f)) ++ hsvcs
+    // Sort by prefix length so that foldLeft results in routes procesed in
+    // descending order (longest first).
+    defaultMiddleware(allSvcs.toList.sortBy(_._1.length).foldLeft(HttpService.empty) {
```
jedesah: What is the benefit of it being mounted in descending order?

wemrysi: So that `/a/b/c` will be consulted before `/a/b`

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -47,44 +45,60 @@ final case class RestApi(defaultPort: Int, restart: Int => Task[Unit]) {
...
+      case (acc, (path, svc)) => Prefix(path)(svc) orElse acc
```
jedesah: Does using `orElse` here increase the latency of handling a request as opposed to using `ServerBuilder.mount`?

I definitely prefer the composability of this approach and would have done it myself had I not been scared of affecting performance of the overall server.

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -47,44 +45,60 @@ final case class RestApi(defaultPort: Int, restart: Int => Task[Unit]) {
...
+    })
+  }
+
+  def defaultMiddleware: HttpMiddleware =
+    cors <<< gzip <<< HeaderParam <<< passOptions
```
jedesah: :+1:

drostron:
👍🏿

web/src/main/scala/quasar/api/services/RestApi.scala
wemrysi: This is pretty much what happens in the implementation of `BlazeBuilder.mountService`, see https://github.com/http4s/http4s/blob/v0.10.1/blaze-server/src/main/scala/org/http4s/server/blaze/BlazeServer.scala#L103 as they use collect everything into a `Router` (which folds over the inputs using what amounts to `orElse`).

jedesah:
@wemrysi Aside from my concern about the performance of mounting services (which hopefully if it is an issue could be resolved with `Router`) and wondering what quasar-advanced looks like with the disappearance of `startServer`, I think this is a super positive change!

web/src/main/scala/quasar/api/Http4sUtils.scala
wemrysi: I'm not as familiar with how the server is started in quasar-advanced, is there something in particular this change is preventing or making more difficult?

web/src/main/scala/quasar/api/services/RestApi.scala
jedesah: Ah, that makes sense, might want to add that in the comment. Seems obvious now that you mention it, but I think it's worth mentioning anyway. Even better would be to not use a `String` to represent a `Path`, but if that's what `http4s` wants...

web/src/main/scala/quasar/api/services/RestApi.scala
jedesah: Hmm... I wonder if this is just a fundamental drawback of their approach. It seems as you scale to a web server with many routes you would want some kind of way to use a `HashMap` or something to match against routes as opposed to just pattern matching on every single possible route that the server accepts. Anyway, if this is how it's done internally, then by all means let's build these things in a more compositional way!

web/src/main/scala/quasar/api/Http4sUtils.scala
wemrysi: Also, notice that `startServer` still exists, it just has a different signature (accepts a `ServerBlueprint`). Since we aren't mounting services via the builder any longer, if quasar-advanced doesn't want to, or can't, use `ServerBlueprint` then it can just `builder.mountService(advancedService)`.

jedesah: Yes. In quasar-advanced we need to supply the `ServerBuilder` because we need to build one server with ssl and one server without. Also, the ssl builder needs to `.bindHttp()` and the non-ssl needs to `.bindLocalhost()` in order to get the desired behavior.

jedesah: It's a good point that now that don't have this stupid `Map[String, HttpService]` this version of `startServer` is not doing very much for us anymore, I get we can avoid using it in quasar-advanced, just use the builder directly.

web/src/main/scala/quasar/api/services/RestApi.scala
wemrysi: I'm planning on leaving `Prefix` alone for now until @mossprescott can weigh in (as I think he added it). It deals with paths differently than how `Router` does and I wanted to understand why before making any changes.

drostron: Sounds good.

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -47,44 +45,60 @@ final case class RestApi(defaultPort: Int, restart: Int => Task[Unit]) {
...
-      case req if req.method == OPTIONS => Ok()
-    })))
+  /** Converts `QHttpService`s into `HttpService`s and mounts them along with
+    * any additional `HttpService`s provided, applying the default middleware
+    * to the result.
+    *
+    * TODO: Is using `Prefix` necessary? Can we replace with
+    *       `org.http4s.server.Router` instead?
+    */
+  def finalizeServices[S[_]: Functor](
```
jedesah: This function is doing two separate things and the first one is as simple as `qsvs.mapValues(_.toHttpService(f))` which if we were to go the `Router` approach would become `service.toHttpService` so I would be inclined to pull it out of here (maybe even just get rid of it) as just mapping over something might not deserve it's own function.

jedesah: But also, okay with it as is while we wait to make the change to `Router` after the release.

jedesah:
@wemrysi Thx for answering all my questions. LGTM!

web/src/main/scala/quasar/api/services/RestApi.scala
wemrysi: Yeah, I went back and forth on this as well, I agree it's a bit smelly, but it is useful for now as a way to mount and apply the middleware in one spot. It will likely disappear when we do switch to `Router` as I think there is still more organization and separating of concerns to be had with how we construct the composite service, when we decide to apply middleware (i.e. deciding what is appropriate for all routes vs specific routes), when/how to combine our free services with normal `HttpServices`, this is definitely just a first attempt.

jedesah: :+1:

wemrysi:
Expanded the comment about why the mounts are sorted before folding, let me know if there is anything else needed to get this merged.

jedesah:
No. LGTM.

wemrysi:
Ok, we still doing the "ready" tagging protocol?

jedesah:
It's just one commit, so @drostron can click the merge button himself.

wemrysi:
What about the version and the changelog?

jedesah:
As yes. Well technically, the version and changelog should be reviewed.

wemrysi:
Really? I thought they would be added in the merge commit itself?

drostron:
@wemrysi, I forgot to switch the label to Ready. Done.

jedesah:
I think in general we should review which number is bumped as well as the documentation in the changelog as part of the PR.

wemrysi:
I guess I can put that in the comments here, but we don't know the version until it is actually merged.

jedesah:
Well. You change the version accordingly and hope no other PR get merged before :smile: This is why it might make sense to setup those 3 branches (one for each kind of version bump) and open PRs against the appropriate branch. That way we can signal intent and automatic the actually version bumping and the addition to the changelog (based on the commit message?).

wemrysi:
@jedesah @drostron So merging this will bump us to 3.0.3, changelog is "Refactor api service 'finalization' to be more decoupled for better reuse as a library". Sound good?

drostron:
Sounds good.

jedesah:
@wemrysi Ya, that sounds good.

PR #1088, opened 2016-02-11
Remove use of mongolabs in sample config file

jedesah:
- It gets used as part of our test suite that runs on Travis and it is too flaky. Hardcode to localhost instead.

### DIFF
--- a/example-quasar-config.json
+++ b/example-quasar-config.json
@@ -2,7 +2,7 @@
     "mountings": {
         "/": {
             "mongodb": {
-                "connectionUri": "mongodb://slamengine:slamengine@ds045089.mongolab.com:45089/slamengine-test-01"
+                "connectionUri": "mongodb://localhost:27018"
             }
         }
     },


sellout:
👍🏽

jdegoes:
Eventually we should also change SBT to pull down the correct MongoDB to use and start it locally so the tests can succeed on fresh checkout.

sellout:
@jdegoes They already should succeed on a fresh checkout – I’m ok with this change because this config gets tested only on Travis before releasing the jar. Running the tests locally still doesn’t require a particular Mongo instance, it just requires you to have _some_ Mongo instance (≥ 2.6) somewhere.

jdegoes:
:+1:

jedesah:
@sellout I think what @jdegoes meant is that the sbt build should take care of everything. i.e. the tests should not fail even if the developer has not previously gone through the steps of installing an instance of MongoDB and configuring things so that quasar knows where to find it.

Anyway, I have thoughts on how to make that happen, but one step at a time.

jdegoes:
> @sellout I think what @jdegoes meant is that the sbt build should take care of everything. i.e. the tests should not fail even if the developer has not previously gone through the steps of installing an instance of MongoDB and configuring things so that quasar knows where to find it.

:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

PR #1089, opened 2016-02-11
Switch quoting style to SQL².

sellout:
**NB**: This is a breaking change

### DIFF
--- a/it/src/test/scala/quasar/view.scala
+++ b/it/src/test/scala/quasar/view.scala
@@ -192,7 +192,7 @@ class ViewSpecs extends BackendTest with DisjunctionMatchers with SkippedOnUserE
     }
 
     "query with view referencing a view" in {
-      val query = """select max("1") from `/view/a/smallCityCounts1`"""
+      val query = """select max(`1`) from `/view/a/smallCityCounts1`"""
       root.evalResults(QueryRequest(parse(query), Variables(Map.empty))).fold[Result](
         e => failure(e.toString),
         _.runLog.run.run must beRightDisjunction(Vector(Data.Obj(ListMap(
@@ -200,7 +200,7 @@ class ViewSpecs extends BackendTest with DisjunctionMatchers with SkippedOnUserE
     }
 
     "query with view referencing a view (absolute)" in {
-      val query = """select max("1") from `/view/b/smallCityCounts2`"""
+      val query = """select max(`1`) from `/view/b/smallCityCounts2`"""
       root.evalResults(QueryRequest(parse(query), Variables(Map.empty))).fold[Result](
         e => failure(e.toString),
         _.runLog.run.run must beRightDisjunction(Vector(Data.Obj(ListMap(
--- a/it/src/test/scala/quasar/view.scala
+++ b/it/src/test/scala/quasar/view.scala
@@ -51,10 +51,10 @@ class ViewSpecs extends BackendTest with DisjunctionMatchers with SkippedOnUserE
 
     val views = Map(
       Path("/view/simpleZips") -> config.ViewConfig(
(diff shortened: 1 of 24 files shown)


jdegoes:
To summarize to make sure the front-end team can make this change:

1. Strings are now represented using JSON-compatible double-quotes (including escape syntax for special characters, unicode, etc.).
2. Identifiers with special characters are now escaped using back ticks.
3. Single characters, including unicode characters escaped using JSON syntax, are represented with single quotes (`'a'`).

Anything else?

sellout:
@jdegoes Yep, that’s basically it. The Unicode escape is `\u1234` (exactly 4 hex digits). Eventually it would be nice to support the `\u{123456}` (1+ hex digits) style, but that will require some other changes, I think. E.g., a character won’t fit in a scala Char, etc.

it/src/test/scala/quasar/view.scala
```diff
@@ -51,10 +51,10 @@ class ViewSpecs extends BackendTest with DisjunctionMatchers with SkippedOnUserE
 
     val views = Map(
       Path("/view/simpleZips") -> config.ViewConfig(
-        parse("select _id as zip, city, state from  \"/" + ZipsPath.simplePathname + "\""),
+        parse("select _id as zip, city, state from  `/" + ZipsPath.simplePathname + "`"),
```
drostron: Could be
```
s"select _id as zip, city, state from  `/${ZipsPath.simplePathname}`"
```
and similar for a few below

it/src/test/scala/quasar/view.scala
```diff
@@ -63,16 +63,16 @@ class ViewSpecs extends BackendTest with DisjunctionMatchers with SkippedOnUserE
         Variables(Map())),
       // ...and this one uses an absolute path, and lives in a different dir:
       Path("/view/b/smallCityCounts2") -> config.ViewConfig(
-        parse("select St, count(*) from \"/view/a/smallCities\" group by St order by count(*) desc, St"),
+        parse("select St, count(*) from `/view/a/smallCities` group by St order by count(*) desc, St"),
         Variables(Map())),
       Path("/view/badRef") -> config.ViewConfig(
-        parse("""select foo from "/mnt/test/nonexistent""""),
+        parse("""select foo from `/mnt/test/nonexistent`"""),
```
drostron: Single `"`s would suffice — similar for a few below, or I suppose triple quoting everything within the file would be consistent.

sellout: Yeah, I think we should probably just move toward using `"""` for all queries. I’ll have to add a keybinding for that, though, because otherwise Emacs is too “smart” to do it.

it/src/test/scala/quasar/view.scala
sellout: I thought we were anti-interp … although I could be wrong and don’t have anything to back that up.

drostron: Interesting, I haven't heard anything negative about them before and like the way they read. I'd be curious to know the reasoning behind anit-interp if anyone knows.

drostron:
👍🏽

it/src/test/scala/quasar/view.scala
sellout: @jdegoes / @mossprescott I’d swear I heard this from one of you. Am I mistaken? If we do have a stance, we should probably add it to the style guide.

jdegoes: I don't like the fact that Scala *has* string interpolation (this is the epitome of bloat). But given that it *does* have it, if people want to use it consistently in preference to string concatenation, that's fine by me. :+1:

sellout:
This is now in a new PR against a different branch (as it’s a breaking change) – #1093.

PR #1091, opened 2016-02-13
Implement the `delete` expression.

sellout:
This is an expression, not a statement – it simply returns a new set
that’s a subset of the original set.

### DIFF
--- a/core/src/main/scala/quasar/sql/parser.scala
+++ b/core/src/main/scala/quasar/sql/parser.scala
@@ -94,11 +94,11 @@ class SQLParser extends StandardTokenParsers {
   def floatLit: Parser[String] = elem("decimal", _.isInstanceOf[lexical.FloatLit]) ^^ (_.chars)
 
   ignore(lexical.reserved += (
-    "all", "and", "as", "asc", "between", "by", "case", "cross", "date", "desc", "distinct",
+    "all", "and", "as", "asc", "between", "by", "case", "cross", "date", "delete", "desc", "distinct",
     "else", "end", "escape", "except", "exists", "false", "for", "from", "full", "group", "having", "in",
-    "inner", "intersect", "interval", "is", "join", "left", "like", "limit", "not", "null",
+    "inner", "insert", "intersect", "interval", "into", "is", "join", "left", "like", "limit", "not", "null",
     "offset", "oid", "on", "or", "order", "outer", "right", "select", "then", "time",
-    "timestamp", "true", "union", "when", "where"
+    "timestamp", "true", "union", "update", "when", "where"
   ))
 
   ignore(lexical.delimiters += (
@@ -118,11 +118,18 @@ class SQLParser extends StandardTokenParsers {
 
   def select: Parser[Expr] =
     keyword("select") ~> opt(keyword("distinct")) ~ projections ~
-      opt(relations) ~ opt(filter) ~
+      opt(from) ~ opt(filter) ~
       opt(group_by) ~ opt(order_by) ^^ {
         case d ~ p ~ r ~ f ~ g ~ o =>
           Select(d.map(κ(SelectDistinct)).getOrElse(SelectAll), p, r.join, f, g, o)
       }
+
+  def delete: Parser[Expr] =
+    keyword("delete") ~> from ~ filter ^^ { case r ~ f =>
+      Select(SelectAll, List(Proj(Splice(None), None)), r, Not(f).some, None, None)
+    }
+
+  def query = delete | select
 
   def projections: Parser[List[Proj[Expr]]] =
     repsep(projection, op(",")).map(_.toList)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/quasar/sql/parser.scala
```diff
@@ -94,11 +94,11 @@ class SQLParser extends StandardTokenParsers {
   def floatLit: Parser[String] = elem("decimal", _.isInstanceOf[lexical.FloatLit]) ^^ (_.chars)
 
   ignore(lexical.reserved += (
-    "all", "and", "as", "asc", "between", "by", "case", "cross", "date", "desc", "distinct",
+    "all", "and", "as", "asc", "between", "by", "case", "cross", "date", "delete", "desc", "distinct",
```
jdegoes: Do we have to add more reserved words? In fact, when can we delete all these reserved words and rely on context for disambiguation?

sellout: We _might_ be able to achieve that with the current parser, but it requires careful rewriting of the rules, because the Scala combinators don’t have backtracking. Even that might not be enough, because if someone writes something like (and this is a terribly contrived example) `select (delete …) from zips`, the parser has already decided if `delete` refers to a bare identifier or the beginning of a delete expression before it processes the `…`.

sellout: Oh, _but_, even with that last issue, if we make the “keyword” cases attempt to match first, we can get around it by explicitly quoting the identifier, so that should allow users to get away without quotes _more_ often, if not as much as we’d like.

jdegoes: Scala's parser combinators do support backtracking via `|`. You just have to list the longest match first, e.g. always try to match for the "statements" before matching for identifiers, etc.

Sounds like we may have to refactor the grammar though so may as well push out.

sellout: But `|` isn’t enough … maybe backtracking isn’t the right word, or maybe `|` backtracking is too weak? For example,
```sql
select * from zips offset 6
```
without keywords, it parses a complete `select` expression at `select * from zips offset` (with “offset” being the alias for the collection), then fails a full parse, because there is still a token left in the string.

But perhaps we could at least move toward _fewer_ keywords.

jdegoes: We can fix that by being less aggressive in parsing aliases, i.e. succeeding without the alias and then later backtracking only if all other parses fail.

Albeit, maybe this is an argument to support `select * from zips as offset`, i.e. unify aliases in projection with alias in selection (see [here](http://www.w3schools.com/sql/sql_alias.asp)). Makes the grammar less locally ambiguous.

sellout: I have a change that allows for the `as` to be optional for both projections and tables, but I’m happy to go the other direction with it as well.

jdegoes: Yeah, I think I'd prefer going the other direction, as this simple change would have a profoundly disambiguating effect on the grammar. I'm not sure if you succeeded in parsing `SELECT` as `SELECT <expr> <expr>`, but if so, that introduces a host of ambiguities with respect to aliases. Also, I've seen users of SlamData try to use `AS` for joins / from, which doesn't work (or didn't, not sure now).

sellout: `AS` _should_ work for relations, but I never was a fan of that ambiguity in SQL. Since we’ve departed from SQL as much as we have, adding this restriction works for me.

jdegoes: > AS should work for relations, but I never was a fan of that ambiguity in SQL. Since we’ve departed from SQL as much as we have, adding this restriction works for me.

Yeah, it'd be nice if you could introduce aliases anywhere in the same consistent syntax, `foo AS bar` always introduces an alias that propagates outward (or something like that).

Anyway, I finally decided (as you can probably tell) that strict ANSI compatibility doesn't really make a lot of sense, since in order to use Quasar with nested data, you need the extensions, and if you're already writing extended SQL, we may as well clean up the little things along the way that have tripped people up for ages.

In the end, we'll have no more variation than, e.g. N1QL or AQL or DocumentDB SQL or any one of numerous other SQL documents, but we'll be more consistent, cleaner, and easier for humans and machines to generate (and parse).

sellout: Since we’re changing stuff … can we also get rid of `date foo` vs `date(foo)` syntax? Not a fan of those “operators”.

jdegoes: > Since we’re changing stuff … can we also get rid of date foo vs date(foo) syntax? Not a fan of those “operators”.

I'm definitely open to that, although, one can view `DATE "foo"` (etc.) as being a data constructor, in which case perhaps it *should* look different than a function, because there are only a few data constructors and all of them have special syntax, (`"foo"` for string, `{}` for map, `[]` for array, etc.).

sellout: Well, we don’t have variadic functions (yet), but even if we did `String('f', 'o', 'o')` is significantly more noise than `"foo"` as is `Map(Arr("foo", 1), Arr("bar", 2), …)`. I.e., those are syntactic sugar over functions to make queries more readable. `date "foo"` doesn’t seem to buy us anything over `date("foo")` … but it does increase ambiguity.

jdegoes: True. I think I'm fine with that change, especially if it simplifies the grammar. Once you have the BNF we can review and figure out if there are any other syntactic irregularities we should delete in the process of standardization.

wemrysi:
:+1:

PR #1092, opened 2016-02-14
Eliminate keywords!

sellout:
This has a few breaking changes:

* `as` is required for relation aliases
* `date`, `time`, `timestamp`, `interval`, and `oid` are now normal
  functions and thus require parens around their argument

And, non-breakingly, keywords can be used (either with or without being
quoted) as identifiers.


jdegoes:
:tada: :+1:

jdegoes:
@sellout Are all keywords gone or just most of them?

sellout:
All except `true`, `false`, and `null`, which should be treated the same regardless of context, I believe, no? I could eliminate those without changing the semantics, but it seems right to treat those tokens as special.

sellout:
I figured with the breakage, it’d be good to get this in the same build as the quotation changes.

jdegoes:
>  I could eliminate those without changing the semantics, but it seems right to treat those tokens as special.

I don't see how you would do that:

```sql
SELECT true, false FROM foo
```

Constant columns or values from the "foo" dataset?

So I think it makes sense to keep them as keywords. You can always back-tick them if you want to treat them as identifiers.

I guess all the other ex-keywords have sufficient (surrounding) structure that we can completely tell from context???

> I figured with the breakage, it’d be good to get this in the same build as the quotation changes.

Is this the last of the syntactic breakage, or do we have other stuff as well?

sellout:
@jdegoes Glad you asked! Because it turns out that something that I thought worked, didn’t.

Basically, for the non-keywords we prefer interpreting as “operators” then fall back to “identifier” if it couldn’t be parsed as an operator (eg, we found `case`, but not the other bits in the right shape to make a valid `case` expr). And, if there’s an ambiguity, you could disambiguate by quoting the identifier to say “this is not an operator”.

But that part didn’t work. Only the lexer saw the difference between quoted/unquoted identifiers, once the parser saw it, `case` and `"case"` (old syntax, still current on this branch) were identical. So, I just fixed that. But I haven’t been able to come up with a case where we need it. (The other use of quoting – non-standard ident names – still worked fine.)

Anyway, back to `select true, false from foo` … having them as true keywords or not, since they don’t need any other context to parse (trailing expr, other keywords, etc.) they always get the operator interpretation, so in both cases they would be constant columns. If you wanted to use them as field names, you need to do `select "true", "false" from foo`, which should (now) work whether or not they are true keywords.

I guess if I make them _not_ keywords, I’ll at least have a test case for the disambiguation.

sellout:
Oh, and yes, I think this and the quoting style are the only breaking changes left in SQL².

core/src/main/scala/quasar/sql/parser.scala
```diff
@@ -304,17 +288,29 @@ class SQLParser extends StandardTokenParsers {
     unary_operator ~ primary_expr ^^ {
       case op ~ expr => op(expr)
     } |
-    keyword("not")        ~> cmp_expr ^^ Not |
-    keyword("exists")     ~> cmp_expr ^^ Exists |
-    case_expr
+    name("not")        ~> cmp_expr ^^ Not |
```
jedesah: This is when you want a semantic diff.

sellout:
I think I’m going to get rid of the remaining keywords, because 1. I don’t like lexers anyway and 2. it lets me have a test (which already exists, actually) that exercises the quoting disambiguation.

jdegoes:
:+1: Sounds good.

sellout:
Oh, I came up with a good (although unrealistic) ambiguity example:
```
select case from when where then and end
```
Like that it’s simply a select with a single projection that’s a case expression (`from`, `where`, and `and` are identifiers). But quote `case`
```sql
select "case" from when where then and end
```
and now it’s a select with a relation and a filter, with (`when`, `then`, `end`, and of course `case` as identifiers).

jdegoes:
@sellout The most important thing is absolute and utter consistency with regard to *predictability*. 

If an expression which can be interpreted as a multi-part operator is always so interpreted, then, well, that's a simple enough rule to describe and use. If there's every anywhere even one exception, that's not so good!

sellout:
Right. We only ever parse an identifier as an identifier as the last case in the most-deeply-nested rule of the parser. No proof, but there shouldn’t be a way to get there without passing through every other possibility first. Quoting is the escape hatch that says _definitely_ parse this as an identifier (still matched by the same deeply-nested rule, but just no longer able to be matched earlier by the “keyword” rule). And if there’s somewhere it can match as identifier before keyword, it’s definitely a bug (not just an inconsistency in the syntax), because there’s no reverse-escape hatch quoting to say “treat this identifier as a keyword”.

sellout:
I mean, it’s possible to come up with convoluted examples like the above that make this seem indecipherable, but mostly it allows people to do reasonable things like `select date from events` without wondering why their query isn’t working.

sellout:
@jedesah: That last commit should eliminate a bit of the noise in the parser file, because now I just redefine the `keyword` combinator, rather than adding a new `name` one.

jdegoes:
@sellout :+1: 

Btw, the grammar can no longer be formalized with BNF, although something like PEG should work.

sellout:
It’s been a while since I’ve written a BNF, but why can’t it be specified as that now? I thought it could handle any CFG.

sellout:
Oh, deterministic choice. Hrmm.

jdegoes:
@sellout Yes, you need ordered choice, since the grammar is not CF (t will do unlimited backtracking). PEG defines not a generator for programs in the language, but a recognizer, and only supports ordered choice, which is sufficient for SQL<sup>2</sup>.

sellout:
This is now tracked by PR #1094, against a new branch.

PR #1093, opened 2016-02-15
Switch quoting style to SQL².

sellout:
**NB**: This is a breaking change


jdegoes:
@damonLL Can you handle the doc updates on quasar-analytics.org and slamdata.com?

@sellout We'll need a list of stuff that's been added, as opposed to change, though we can wait to doc that since it's less relevant than breaking changes.

sellout:
@jdegoes @damonLL All PRs merged into master now include adding a file to CHANGELOG/ in this case, there might be some more detail needed, but in general, looking at the files between the quasar dependency of the previous SD release and the next SD release should cover it. Maybe we can write a quick script to generate the aggregate changelog …

PR #1094, opened 2016-02-15
Eliminate keywords!

sellout:
This has a few breaking changes:

* `as` is required for relation aliases
* `date`, `time`, `timestamp`, `interval`, and `oid` are now normal
  functions and thus require parens around their argument

And, non-breakingly, keywords can be used (either with or without being
quoted) as identifiers.


web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
```diff
@@ -193,7 +193,7 @@ class ExecuteServiceSpec extends Specification with FileSystemFixture with Scala
           destination = Some(printPath(destination)),
           state = filesystem.state,
           status = Status.BadRequest,
-          response = _ must_== Json("error" := "keyword 'case' expected; `where'")
+          response = _ must_== Json("error" := "end of input; ErrorToken(illegal character)")
```
jedesah: As a side comment, might make sense at this level to simply expect an error as opposed to it being specific like this as I would wager that when this test failed, you simply copied the output from "actual" into expected.

sellout: You’re right – for some reason I was thinking there was more in the structure than just the `error` field, but … uh, nope.

jedesah:
Sorry. Couldn't find anything. LGTM.

PR #1096, opened 2016-02-17
Fix bug where failure effects were being handled via Task.fail

wemrysi:
As part of the translation to the free architecture, we handled some `Failure` effects via `Task.fail`, this caused a few classes of errors to always result in 500 responses.

This fixes the problem by properly interpreting these errors into `QuasarResponse`s.

Unfortunately this happens when constructing the final interpreters in `main` and thus there is no test coverage currently. We'll need tests for the server process itself to check that the effects are being evaluated as expected.

Authored with @mossprescott

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/WorkflowExecutionError.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/WorkflowExecutionError.scala
@@ -58,7 +58,7 @@ object WorkflowExecutionError {
         case InsertFailed(b, r) =>
           s"Failed to insert BSON, `$b`, $r"
         case NoDatabase =>
-          "No database"
+          "Unable to determine a database in which to store temporary collections"
       }
 
       s"Error executing workflow: $msg"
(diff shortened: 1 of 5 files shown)


core/src/main/scala/quasar/physical/mongodb/WorkflowExecutionError.scala
```diff
@@ -58,7 +58,7 @@ object WorkflowExecutionError {
         case InsertFailed(b, r) =>
           s"Failed to insert BSON, `$b`, $r"
         case NoDatabase =>
-          "No database"
+          "Unable to determine a database in which to store temporary collections"
```
drostron: I take it there's no situation a `NoDatabase` error can occur other than storing a temporary collection?

wemrysi: Yeah, or at least that's all we use that case for.

drostron:
👍🏼

PR #1097, opened 2016-02-17
[SD-1367] Fix issue where certain error conditions would respond with 200 and then drop connection

wemrysi:
Fixes an issue where when a `QuasarResponse` body `Process` emits some output before the first effect and/or has other `Await` cases before the first "real" effect (like those introduced by many process1 combinators) _and_ the first `F` effect fails, the resulting response becomes a `200 OK` and then drops the connection.

More concretely, this manifested when retrieving a view-backed file via the `/data/fs` endpoint in `application/json` format where the view query resulted in an error. In this case, a `[\n` is prepended to the output stream which also has process1 "effects" due to `intersperse`-ing characters between the values in the response.

Kudos to @mossprescott who actually figured out the problem and authored the solution with me.

This is related to SD-1367, but I'm not sure if it fixes the issue as the defect fixed here was preventing the actual error encountered from surfacing to the client. So even if SD-1367 isn't fixed by this, it should allow us to now see the root cause of the issue and fix it.

### DIFF
--- a/web/src/main/scala/quasar/api/QuasarResponse.scala
+++ b/web/src/main/scala/quasar/api/QuasarResponse.scala
@@ -27,6 +27,9 @@ import org.http4s._
 import org.http4s.headers._
 import org.http4s.dsl._
 import scalaz.{Optional => _, _}
+import scalaz.std.anyVal._
+import scalaz.std.iterable._
+import scalaz.syntax.foldable._
 import scalaz.syntax.monad._
 import scalaz.concurrent.Task
 import scalaz.stream.Process
@@ -34,7 +37,7 @@ import scodec.bits.ByteVector
 
 @Lenses
 final case class QuasarResponse[S[_]](status: Status, headers: Headers, body: Process[Free[S, ?], ByteVector]) {
-  import QuasarResponse.HttpResponseStreamFailureException
+  import QuasarResponse.{PROCESS_EFFECT_THRESHOLD_BYTES, HttpResponseStreamFailureException}
 
   def modifyHeaders(f: Headers => Headers): QuasarResponse[S] =
     QuasarResponse.headers.modify(f)(this)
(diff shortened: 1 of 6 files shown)


web/src/main/scala/quasar/api/QuasarResponse.scala
```diff
@@ -64,6 +67,32 @@ final case class QuasarResponse[S[_]](status: Status, headers: Headers, body: Pr
...
+    * in `process1` and the like).
+    *
+    * This leads us to the current workaround which is to define this threshold
+    * which should be, ideally, just large enough to require the first `F` to
+    * produce the bytes, but not more. We then consume the byte stream until it
+    * ends or we've consumed this many bytes. Finally we have a chance to inspect
+    * the `F` and see if anything failed before handing the rest of the process
+    * to http4s to continue streaming to the client as normal.
+    */
+  val PROCESS_EFFECT_THRESHOLD_BYTES = 100
```
sellout: Can you add something about why 100 is reasonable value for this? And will changing other bits (say, adding a response header somewhere else) require this number to be bumped?

wemrysi: Will do and no, headers do not affect this. The only reason the threshold is needed at all is because we prepend some data to certain streams and so our existing use of `unconsOption` wasn't working as it stopped after it received the first `Emit` which, if we prepended data, meant that we didn't actually run any effect and so missed any chance to check for errors.

sellout: Ah, cool. So it really is just those three bytes, and in general, any wrapping of data in a response like that should be tiny … unless we start supporting SOAP or another abomination 😉

sellout:
👍🏽

web/src/main/scala/quasar/api/QuasarResponse.scala
wemrysi: Aye, and I'll be the first to admit this isn't ideal, I'd much rather prefer a categorical way of expressing "consume the stream until the first `F` effect" (as opposed to the first `Await`), but I'm not sure how to do that (or if it is possible). If we can figure out how, then we'll definitely replace this.

PR #1098, opened 2016-02-17
Improve the error message when a file path is supplied to the /query service

mossprescott:
Fixes SD-1364

Pass file/dir path through to the queryPlanner, which will reject file paths as required, and added a test to verify the resulting error message.

Also moved the `pathUri` utility function that properly escapes paths for use in request URIs to PathUtil where it can be used in any test.


mossprescott:
Note: I don't think this is important for 2.5, since the problem has not been encountered by slamdata as far as I know.

drostron:
👍🏾

PR #1099, opened 2016-02-18
Reorganize web server code

wemrysi:
Very few semantic changes here, mostly moves and renames. The goal being to break apart the rather large `quasar.api.Server` object into more focused bits. This involved

* Creating the `quasar.server` package and moving the `Server` object,  the server "control" service, as well as some of the types used in the server (and tests) into this package.
* Moving all the concrete effect types and their interpreters into the `quasar.server.impl` object and adding documentation describing why each one exists and how it is used.
* Dealing with the resulting fallout in `Server`

While I was in there, I also added a fix for SD-1164. Now if there is a configuration related error, it will be logged to STDERR and the server will still start using the default configuration. The reason we decide to still go ahead and start is for the "desktop" use-case. In that scenario a user is less likely to know the configuration exists or where to find it, so at least they'll get a running, albeit empty, application.


mossprescott:
See comments on the commits (not exactly sure why they're not showing up here this time, except github).

mossprescott:
Nvm. Guess I just needed to refresh the page.

mossprescott:
By the way, it's pretty hard to follow the evolution here with you rebasing as you go (at least I assume that's what's happening).

wemrysi:
Ah, sorry about that!

mossprescott:
:+1:

PR #1101, opened 2016-02-19
A few quality of life improvements to Q{Response, HttpService}

wemrysi:
* Renames `QuasarResponse` -> `QResponse` for consistency with `QHttpService` (the improved brevity is nice as well)
* Removes the unnecessary `Functor` constraint on `QHttpService#apply`
* Adds instance methods to `QResponse` and `QHttpService` for transforming their effects


drostron:
Already experiencing increased happiness. 👍🏾

PR #1102, opened 2016-02-19
Improve reliability of arbitrary generation of refined values

jedesah:
- Fix bug in `Choose` instance of `SafeIntForVector`
- Add custom generators for discrete refined numerical values
- fix error in specification of "scan with offset k and limit j takes j data, starting from k" in ReadFilesSpec
- use custom generators in specs

### DIFF
--- a/core/src/test/scala/quasar/fp/numeric/SafeIntForVectorArbitrary.scala
+++ b/core/src/test/scala/quasar/fp/numeric/SafeIntForVectorArbitrary.scala
@@ -16,8 +16,6 @@
 
 package quasar.fp.numeric
 
-import quasar.Predef._
-
 import eu.timepit.refined.scalacheck.numeric.Bounded
 import org.scalacheck.Gen
 import org.scalacheck.Gen.Choose
@@ -25,7 +23,7 @@ import org.scalacheck.Gen.Choose
 object SafeIntForVectorArbitrary {
   implicit val chooseSafeIntForVector: Choose[SafeIntForVector] = new Choose[SafeIntForVector] {
     def choose(low: SafeIntForVector, high: SafeIntForVector) =
-      Gen.choose(Int.MinValue, SafeIntForVector.maxValue).map(SafeIntForVector.unsafe)
+      Gen.choose(low.value, high.value).map(SafeIntForVector.unsafe)
   }
 
   implicit val SafeIntForVectorBounded: Bounded[SafeIntForVector] =
--- a/it/src/test/scala/quasar/fs/ReadFilesSpec.scala
+++ b/it/src/test/scala/quasar/fs/ReadFilesSpec.scala
@@ -20,17 +20,19 @@ import quasar.Predef._
 import quasar.Data
 import quasar.fp._
 import quasar.fp.numeric._
+import quasar.fp.numeric.NumericArbitrary._
 
 import java.lang.RuntimeException
 import scala.annotation.tailrec
 
 import eu.timepit.refined.auto._
 import eu.timepit.refined.numeric.{Positive => RPositive,_}
-import eu.timepit.refined.scalacheck.numeric._
+import eu.timepit.refined.scalacheck.numeric.chooseRefinedNum
 import eu.timepit.refined.W
 import eu.timepit.refined.api.Refined
 import monocle.std.{disjunction => D}
 import org.scalacheck.Arbitrary
+import org.scalacheck.Prop
 import org.specs2.ScalaCheck
 import pathy.Path._
 import scalaz.{EphemeralStream => EStream, _}, Scalaz._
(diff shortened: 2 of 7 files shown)


core/src/test/scala/quasar/fp/numeric/SafeIntForVectorArbitrary.scala
```diff
@@ -25,7 +23,7 @@ import org.scalacheck.Gen.Choose
 object SafeIntForVectorArbitrary {
   implicit val chooseSafeIntForVector: Choose[SafeIntForVector] = new Choose[SafeIntForVector] {
     def choose(low: SafeIntForVector, high: SafeIntForVector) =
-      Gen.choose(Int.MinValue, SafeIntForVector.maxValue).map(SafeIntForVector.unsafe)
+      Gen.choose(low.value, high.value).map(SafeIntForVector.unsafe)
```
mossprescott: Seems like a good catch.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
```diff
@@ -143,7 +144,7 @@ class ReadFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT) w
         r.runEither must beRight(d.toIndexedSeq)
       }(implicitly, // In order to keep test execution relatively fast
         Arbitrary(chooseRefinedNum[Refined, Int, RPositive](1, 200)), implicitly,
-        Arbitrary(chooseRefinedNum[Refined, Int, NonNegative](1, 200)), implicitly).set(minTestsOk = 5)
+        Arbitrary(chooseRefinedNum[Refined, Int, NonNegative](0, 200)), implicitly).set(minTestsOk = 5)
```
mossprescott: Ok, so 0 is better than 1, but this seems like a painful way to use arbitraries, and why is `minTestsOk=5` needed if the range is restricted?

mossprescott:
I'm not really up-to-speed on this Refined stuff, so can't really review the "discrete" business. Otherwise just a couple of comments.

core/src/test/scala/quasar/fp/numeric/SafeIntForVectorArbitrary.scala
jedesah: :smile:

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
jedesah: It's a little more verbose than I would like, but I really like the separation between the "pure" property and the reality of "verifying" it in a reasonable amount of time.

jedesah:
Looks like travis has come back to life! :fireworks:

@mossprescott Your comment made me realize that it was probably worth writing up a quick summary of why `NumericArbitrary` is useful.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
mossprescott: Just realized that when I've done this sort of thing I've used something like:

```scala
"something" ! Prop.forAll(someGen(...), otherGen(...)) { (j, k) => 
 ...
}
```

Would that be a good option here?

mossprescott:
Just the comment on scalacheck style. Otherwise LGTM and hopefully will help with some of the Travis failures.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
jedesah: My personal preference is to have it at the end of the property definition rather than the beginning (because it's unfortunate computational implementation details as opposed to elegant logical statements). Is it the lack of `implicit` that makes you prefer to declare it upfront?

mossprescott: You prompted me to think this through more and I now have a more coherent position :)

With the restricted generator at the bottom, on a casual reading this test claims to prove that "the behavior should be ... for all non-negative ints" (which in fact is probably what we'd like to prove), but in fact it proves only that the behavior is correct for 0..200. If we happen to have a bug that appears only when `k` is larger than 200, then this test will never find it, the same way it wouldn't find bug for the `k = 0` case prior to this commit. I don't think it's a good idea to hide that, and I think that's what putting it at the bottom does.

By contrast, it makes sense to de-emphasize `minTestsOk` because it effects only the "non-functional" behavior of the test (how long it runs and how likely it is to find a bug on any one run), but not the actual behavior being tested (because even if minTestsOk is 1 it will still try all possible values with some probability).

jedesah: Okay, I can get behind that.

sellout:
Please, please, please get this merged! 🙏🏽

mossprescott:
LGTM.

NB: I just restarted the Travis build that failed, which was the "server restart" test. :disappointed:

PR #1103, opened 2016-02-20
fix inlining of $project ops

mossprescott:
Fixes SD-1403

Two cases were being silently dropped:

1. "dangling" references, where a project op referred to a name that was simply missing from the previous shape.
2. references where the referenced expresion was not a trival name _and_ the reference reached down into the named object. In particular, the case we were hitting is when the first projection does a type check on `$$ROOT` (i.e. `"__tmp0": $cond(..., $$ROOT, undefined)`) and then the second reaches in with something like `"__tmp3": "$__tmp0.city"`.

Both cases are now treated as non-inlinable and the two projections are left as is. That makes the dangling reference easier to spot and debug (and it will be caught in property-based tests). In the second case, it seems prudent to preserve the two projections because attempting to inline it would mean expanding the first expression more than once in the second (three times, in the case of a type-check).

Includes tests at three levels: optimizer, WorkflowBuilder, and planner, because the effect is pretty subtle.

### DIFF
--- a/core/src/test/scala/quasar/physical/mongodb/planner.scala
+++ b/core/src/test/scala/quasar/physical/mongodb/planner.scala
@@ -74,11 +74,16 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
       e => scala.sys.error("parsing error: " + e.message),
       expr => queryPlanner(QueryRequest(expr, Variables(Map()))).run._2).toEither
 
-  def plan(logical: Fix[LogicalPlan]): Either[PlannerError, Crystallized] =
-    (for {
-      simplified <- emit(Vector.empty, \/-(Optimizer.simplify(logical)))
+  def plan(logical: Fix[LogicalPlan]): Either[PlannerError, Crystallized] = {
+    val (log, wf) = (for {
+      _          <- emit(Vector(PhaseResult.Tree("Input", logical.render)), ().right)
+      simplified <- emit(Vector.empty, \/-(Optimizer.simplify(logical))): EitherWriter[PlannerError, Fix[LogicalPlan]]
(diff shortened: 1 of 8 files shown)


mossprescott:
I held the version-bump out as a separate commit for now, just in case we get something automatic in place before this gets merged. I'll squash it before merging otherwise.

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -74,11 +74,19 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
...
+    val (log, wf) = (for {
+      _          <- emit(Vector(PhaseResult.Tree("Input", logical.render)), ().right)
+      simplified <- emit(Vector.empty, \/-(Optimizer.simplify(logical))): EitherWriter[PlannerError, Fix[LogicalPlan]]
+      _          <- emit(Vector(PhaseResult.Tree("Simplified", logical.render)), ().right)
       phys       <- MongoDbPlanner.plan(simplified)
-    } yield phys).run._2.toEither
+    } yield phys).run
+
+    // import quasar.fp._
+    // println(log.map(_.show).mkString("\n\n"))
```
sellout: debugging

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "4.0.5"
+version in ThisBuild := "4.0.6"
```
sellout: Granted, we should automate this ASAP, but generally we don’t bump the version or changelog in the PR, because something else might get merged first, and then these conflict. But … maybe that’s not so bad.

sellout:
I like it. Uses Option more consistently. Just the bit of commented debugging code, really. 👍🏽

PR #1105, opened 2016-02-22
Remove custom handling of UnsupportedContentType

jedesah:
- Upgrade to http4s 0.12.1
- Upgrade to scalaz-stream 0.8
- Improve error message of response when user supplied media type is incorrect
- Add fix that was submitted upstream to http4s (for better error message)


jedesah:
I am really confused about the build failure for this PR.

The test failure is in `core` and only two files were changed in core and they are such benign changes that have nothing to do with `CoreConfigSpec`. But it seems to happen 100% of the time so it must be the change. `http4s` and `scalaz-stream` were upgraded which seems like the most likely culprit although it is not obvious to me how this would affect json serialization.

Anyone have any ideas?

mossprescott:
Not sure what's going on there but I would suggest one or more of the following:
- write a RenderTree for the config type and use `beTree` or `beRightDisjOrDiff` to get a diff that shows you where the problem is.
- write a shrinker for the config type so you get a minimized failing input. See `PlannerSpec.shrinkExpr`.
- maybe change the generator to use `size` more intelligently. I can't really tell if that's the issue, but if you're generating hundreds of mounts with hundreds of nodes in the view queries, yikes.

wemrysi:
@jedesah It likely has to do with a workaround we added b/c http4s wasn't properly escaping query param key names when rendering to string, see

https://github.com/quasar-analytics/quasar/blob/master/core/src/main/scala/quasar/fs/mount/MountConfig.scala#L121

now that it is fixed upstream, these are probably being double-escaped.

jedesah:
@wemrysi Thx, that was it!

CHANGELOG/5.0.0.md
```diff
@@ -0,0 +1 @@
+- Improve error message when client issues a POST or PUT request on `/data/fs` with an unsupported or missing Content-Type header
```
sellout: Gotta rename this file.

core/src/main/scala/quasar/fs/mount/MountConfig.scala
```diff
@@ -118,12 +118,8 @@ object MountConfig2 {
   private def viewCfgAsUri(query: Expr, vars: Variables): String = {
     import org.http4s._, util._, CaseInsensitiveString._
 
-    // TODO: Workaround for https://github.com/http4s/http4s/issues/510
-    val urlEncode: String => String =
-      UrlCodingUtils.urlEncode(_, spaceIsPlus = false, toSkip = UrlFormCodec.urlUnreserved)
-
     val qryMap = vars.value.foldLeft(Map("q" -> List(sql.pprint(query)))) {
-      case (qm, (n, v)) => qm + ((urlEncode(VarPrefix + n.value), List(v.value)))
+      case (qm, (n, v)) => qm + ((VarPrefix + n.value, List(v.value)))
```
sellout: I assume this is the fix for the double-encoding @wemrysi mentioned?

sellout:
This looks good. Just make sure that the version info is still correct when merge it.

wemrysi:
@jedesah Is updating to scalaz-stream 0.8 necessary for this PR? If so, we need to audit all the places we may be acquiring resources with streams and update accordingly, see https://github.com/functional-streams-for-scala/fs2/wiki/0.8#preemption-changes.

If it isn't necessary for this PR, then I'd wait and update scalaz-stream in a separate PR. Though, we should probably do a separate PR for scalaz-stream either way because of the code changes involved.

core/src/main/scala/quasar/fs/mount/MountConfig.scala
jedesah: Yes. That is correct.

jedesah:
@sellout Git conflict resolution should take care of making sure I don't merge an outdated version (right?)

jedesah:
@wemrysi `http4s` depends on `scalaz-stream` `0.8` so I really just made the change to squelch the warning message. I think I will make the upgrade to `0.8` in this PR as there is already many changes in here that are specific to it.

jedesah:
@sellout Why the change to "in progress"?

sellout:
@jedesah Build failure, but I’m guessing that was one of our random ones, since you restarted it. I just saw the ❌ on https://github.com/pulls?q=is%3Aopen+assignee%3Asellout, so I hadn’t been checking.

But also, you have to change the name of your CHANGELOG file and undo the manual change to version.sbt. The `merge` script will fail on this PR.

Just reviewed all the other changes. Looks good.

sellout:
Argh! The timeout failures! Do we have any idea what’s behind those?

jedesah:
@sellout Ya, I know, I think we have some timeouts that are not long enough. :stuck_out_tongue_closed_eyes:

jedesah:
@sellout Does the merge script do anything that I haven't done manually? Of course I don't plan to do them manually on my next PR, but if this build finally passes and no one else has merged, could I just merge this one?

mossprescott:
The way the merge script works now is it amends your last commit with the two version-related changes (renaming revision.md and updating version.sbt). So you can do that manually and get the same effect, as long as you're rebased on master.

If you use the merge script, you don't have to rebase, because it applies the version _after_ merging your branch onto (a copy of) master.

sellout:
@jedesah You’re right – you should be fine, I think.

@mossprescott That `git pull` onto the fresh branch currently requires a fast-forward, though, so you either have to be [re]based on master or have merged (I think). I think that part could be relaxed, but we if so, we should discuss what happens in that case.

PR #1106, opened 2016-02-23
Don’t inherit “standard” lexers.

sellout:
SD-1404 #done

Fix our tokenizer to not rely on any external parsers, and *cough* fix
the queries that needed to be updated because of this change.


mossprescott:
So `'abc'` is a parse error now? Worth adding a test?

And what is char literal syntax actually good for anyway, since we don't have a separate Char type?

sellout:
Yep, I’ll add a test.

Char (not equivalent to Scala’s Char) _will_ be a type, and Strings will be an Array of Char (at least semantically).

mossprescott:
:+1:

PR #1108, opened 2016-02-24
Push filtering ahead of joins that already have conditions

mossprescott:
A little more normalization of the incoming plan was sufficient to get the existing rewriting to recognize the additional case mentioned in SD-1418.

Specifically, we now "hoist" Lets that appear in the arguments of Invoke and Typecheck. Note: this is safe only if there is no possibility of introducing spurious shadowing, which we arrange for by first normalizing the names (which ensures they are all unique). Therefore this PR adds a couple of additional rewrites to `optimize`, and also makes the rewrites even more sensitive to being applied in the correct sequence.

The fallout was surprisingly modest:
- only 6 compiler tests needed to be updated for Lets moving around
- only 1 planner test needed updating. See my comment on the commit.

### DIFF
--- a/core/src/main/scala/quasar/optimizer.scala
+++ b/core/src/main/scala/quasar/optimizer.scala
@@ -294,11 +294,28 @@ object Optimizer {
     * input is expected to come straight from the SQL^2 compiler or
     * another source of un-optimized queries.
     */
-  def optimize(t: Fix[LogicalPlan]): Fix[LogicalPlan] = {
-    val t1 = t.transCata(repeatedly(simplifyƒ))
-    val t2 = boundParaS(t1)(rewriteCrossJoinsƒ).evalZero
-    val t3 = t2.transCata(repeatedly(simplifyƒ))
-    val t4 = (normalizeLets _ >>> normalizeTempNames _)(t3)
-    t4
-  }
+  val optimize: Fix[LogicalPlan] => Fix[LogicalPlan] =
+    NonEmptyList[Fix[LogicalPlan] => Fix[LogicalPlan]](
+      // Eliminate extraneous constants, etc.:
+      _.transCata(repeatedly(simplifyƒ)),
+
+      // NB: must precede normalizeLets to eliminate possibility of shadowing:
+      normalizeTempNames,
+
+      // NB: must precede rewriteCrossJoins to normalize Filter/Join shapes:
+      normalizeLets,
+
+      // Now for the big one:
+      boundParaS(_)(rewriteCrossJoinsƒ).evalZero,
+
+      // Eliminate trivial bindings introduced in rewriteCrossJoins:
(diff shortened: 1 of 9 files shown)


mossprescott:
Build should be clean now, and I included `revision.md` so you can review that as well. I'll rename it when I squash and bump the version (although it seems like those steps could be automated in some way...).

core/src/main/scala/quasar/optimizer.scala
```diff
@@ -295,10 +295,28 @@ object Optimizer {
...
+      // Eliminate trivial bindings introduced in rewriteCrossJoins:
+      _.transCata(repeatedly(simplifyƒ)),
+
+      // Final pass to normalize the resulting plans for better matching in tests:
+      normalizeLets,
+
+      // This time, fix the names last so they will read naturally:
+      normalizeTempNames)
+
+    fns.foldLeft(t) { case (t, f) => f(t) }
```
sellout: Heh, nice.

core/src/test/scala/quasar/compiler.scala
```diff
@@ -538,31 +536,36 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
     "compile union" in {
       testLogicalPlanCompile(
         "select loc, pop from zips union select city from zips",
-        Distinct[FLP](Union[FLP](setA, setB)))
+        Let('__tmp0, read("zips"),
+          Distinct[FLP](Union[FLP](setA, setB))))
```
sellout: These are a bit confusing, but I don’t know what to do about it.

mossprescott: Another way to approach this would be to put `sertA` and `setB` back the way they were and normalize after composing them. Would that bet better?

core/src/main/scala/quasar/optimizer.scala
mossprescott: Still not completely happy with it. I think I should define the function point-free by folding with `compose`/`>>>`. What do you think?

sellout: Yeah, I think I like that more.

core/src/test/scala/quasar/compiler.scala
sellout: Yeah, I think that’d be good – then `set*` makes more sense on their own.

sellout:
@damonLL ran into this again today, with a more significant query. I suggested he write it in “cross-style” as a workaround, but getting out a version with this in it would be great.

mossprescott:
Ready for final review. 

In updating those UNION tests, I realized Let hoisting wasn't working the way I thought it was, so I added a couple of tests (which I should have done earlier, to be honest), and a comment explaining the current behavior, which I think is good enough for now.

sellout:
👍🏽

PR #1110, opened 2016-02-26
Upgrade to sbt 0.13.11

jedesah:


jedesah:
@sellout This should be a really quick one! :smile:

sellout:
Just add a CHANGELOG/revision.md file so we can test the new script on this 😄

jedesah:
Can't do it right now, I left the office. I think this bumping the patch
version is a bit of a stretch but I guess I can convince myself it makes
sense.

sellout:
Seriously, every merge into master bumps the version. All of them. Simple, consistent. Anyone checks out master, they have a known version of quasar.

jedesah:
I like the elegance, but I would prefer if it was rephrased as every change
to quasar bumps it's version.

I'm not sure our downstream clients will appreciate downloading a whole new
identical version of quasar because we fixed a typo in the Readme.

jdegoes:
> I'm not sure our downstream clients will appreciate downloading a whole new identical version of quasar because we fixed a typo in the Readme.

That's not really relevant, however, since if we're doing things correctly, Quasar's patch version is bumped *multiple times per day*.

mossprescott:
Right, and a patch never delivers new features, so there's no pressure to upgrade unless you need a bug fix.

Anyway we made a decision, so let's at least get the script working and try it for a while before we re-open the debate.

sellout:
👍🏾

jedesah:
Ya, we probably want to figure out a way to automatically close PRs after the merge is done. But I can live with it as is until we get the dmz branch in place.

PR #1111, opened 2016-02-27
Convert the REPL to the Free architecture

mossprescott:
I will add a commit tomorrow that deletes the old REPL implementation, but wanted to get this in review while it was technically Friday.


CHANGELOG/revision.md
```diff
@@ -0,0 +1,5 @@
+Convert the REPL over to our new architecture, with a handful of improvements.
+- When launching the REPL, "-c" or "--config" should now be used on the command line to specify the location of the configuration file.
+- Better feedback about errors when launching the REPL.
+- A new setting, `format`, which allows JSON or CSV output to be selected. The default is still the nicely aligned table.
+- It's now possible to evaluate "queries" that do not refer to any data source (such as `1 + 2`), even if multiple mounts are present.
```
mossprescott: Note: I'm claiming this a "no features" release because it doesn't add any features to the web API, but I could be talked into a `minor` release because it does add features to the REPL. Opinions?

jdegoes:
Every release is at least a patch release.

mossprescott:
@jdegoes: that's what I meant, just a "revision/patch".

sellout:
@mossprescott I'd argue for minor due to now being able to make backend-independent queries, since that's also available through the Web API (and something I think the front end has been wanting to take advantage of).

CHANGELOG/minor.md
```diff
@@ -0,0 +1,6 @@
+Convert the REPL over to our new architecture, with a handful of improvements.
```
drostron: All other changelogs are fully bulletized. Is that something we want to continue?

drostron: Mostly curious, this seems fine to me.

core/src/main/scala/quasar/effect/Timing.scala
```diff
@@ -0,0 +1,59 @@
+/*
```
drostron: There is also a Timing effect in advanced. We'll likely want to consolidate the two at some point.

CHANGELOG/minor.md
mossprescott: Hmm. When we start doing something with the contents, I imagine it will just be to put together a .md with a section for each release. So it will be nice if their formatting is consistent, or at least if there's some logic to it.

In this case, I wanted to mention two separate changes and one of them had sub-items, so this is what I ended up with.

I guess now that you're mentioning it, it seems silly to me to bulletize a one-liner, which is what we've done in a lot of cases.

In any case I don't think it's big deal because we can always go back and re-format them retroactively once we decide on a common style (and cut a release with that change :smiling_imp: )

core/src/main/scala/quasar/effect/Timing.scala
mossprescott: Thanks for pointing that out. It looks like they won't be hard to unify. Should I do that in this PR so we can switch advanced to it with a single change?

drostron: If it's not too disruptive to this PR, that sounds helpful.

mossprescott:
Updated with the additional operation in Timing so we can use it in advanced.

core/src/main/scala/quasar/codec.scala
```diff
@@ -44,6 +44,9 @@ object DataEncodingError {
 
   implicit val EncoderDataEncodingError: EncodeJson[DataEncodingError] =
     EncodeJson(err => Json.obj("error" := err.message))
+
+  implicit val DataEncodingShow: Show[DataEncodingError] =
```
jedesah: Not a big deal, but I wonder why we are using uppercase for these vals. We (and the rest of the scalaz community) seems to favor lowercase for typeclass instances. Also, I was going to say I prefer having the full name of the type in the instance name, but then I realized it was in it's companion object, so I think it should just be `show` and `encodeJson`.

core/src/main/scala/quasar/console.scala
```diff
@@ -29,9 +29,9 @@ object console {
   def stderr(msg: => String): Task[Unit] =
     Task.delay(System.err.println(msg))
 
-  def logErrors[E:Show](a: quasar.Errors.ETask[E,Unit]) =
+  def logErrors(a: quasar.Errors.ETask[String, Unit]): Task[Unit] =
```
jedesah: I suppose there is a good reason to add `"Error: "`, but I don't understanding the reasoning behind making this less generic.

mossprescott: Just because the Show instance for String wraps it in quotes and printing errors with quotes around them looked dumb to me. It's only being used with MainTask, which has String on the left.

core/src/main/scala/quasar/codec.scala
mossprescott: We have more than one convention for these at this point (see, for example, the instance on line 45) :smile: 

Also, wasn't there some back and forth on another PR about using full names vs simple names like `show` (which could potentially conflict if the companion's contents are imported)?

I dont feel strongly about it one way or the other, but I'd like to be consistent, at least on a per-file basis.

core/src/main/scala/quasar/console.scala
jedesah: Fair enough.

core/src/main/scala/quasar/codec.scala
jedesah: I believe the conclusion was that it's fine for it to be short if it's in the companion object as it should never need to be imported. However it should be the full name if it's not in the companion object.

core/src/main/scala/quasar/repl/repl.scala
```diff
@@ -17,440 +17,238 @@
...
-        Mounter.defaultMount(MountingsConfig.fromMC2(config.mountings))
-          .fold(e => printErrorAndFail(e.message), Task.now _)
-          .join
-
-      for {
-        fsPath  <- args.headOption traverse parsePath
-        cfg     <- configFromPath(fsPath)
-        backend <- backendFromConfig(cfg)
-      } yield backend
+  def secondsAndTenths(s: Double): Double =
```
drostron: I believe `Double`s larger than `Int.MaxValue` will produce an incorrect result. Granted the likelihood of someone running a query for 10s of years would be quite surprising. If I understand the goal correctly, an alternative approach would be to use the `f` interpolator above like: `f"Query time: $elapsed%.1fs"`

core/src/main/scala/quasar/effect/Timing.scala
```diff
@@ -0,0 +1,82 @@
...
+    private val unsafe = Unsafe[S]
+
+    /** Clock time, convertible to time and date, but with relatively little
+      * precision.
+      */
+    def timestamp: F[Instant] =
+      lift(Timestamp)
+
+    /** Elapsed time to evaluate some term, with greater precision. */
+    def time[A](fa: F[A]): F[(A, Double)] = {
```
drostron: Thoughts on using something like `org.threeten.bp.Duration` instead of `Double`?

core/src/main/scala/quasar/repl/Command.scala
```diff
@@ -0,0 +1,103 @@
...
+      case UnsetVarPattern(name)         => UnsetVar(name)
+      case ListVarPattern()              => ListVars
+      case _                             => Select(None, Query(input))
+    }
+
+  type XDir = RelDir[Unsandboxed] \/ ADir
+  object XDir {
+    def unapply(str: String): Option[XDir] =
+      Option(str)
+        .filter(_ != "")
```
drostron: Could go with scalaz's type safe alternative: `=/=` or `≠`.

core/src/main/scala/quasar/repl/Command.scala
```diff
@@ -0,0 +1,103 @@
...
+        .flatMap { s =>
+          posixCodec.parseRelDir(s).map(_.left) orElse
+            posixCodec.parseAbsDir(s).map(sandboxAbs(_).right)
+        }
+  }
+  type XFile = RelFile[Unsandboxed] \/ AFile
+  object XFile {
+    def unapply(str: String): Option[XFile] =
+    Option(str)
+      .filter(_ != "")
```
drostron: Same here

core/src/main/scala/quasar/effect/Timing.scala
mossprescott: I wouldn't object. It looks like Duration handles nanosecond resolution.

core/src/main/scala/quasar/repl/Command.scala
mossprescott: Sure, I'll make that change.

core/src/main/scala/quasar/repl/prettify.scala
```diff
@@ -232,4 +231,11 @@ object Prettify {
         cols.map(n => flat.get(n).fold("")(render(_).value))
       })
   }
+
+  /** Pure version of `renderStream. */
+  // FIXME: "pure" in the sense that it's hiding a call to Task.run
```
mossprescott: Could use some input on whether this is a reasonable thing to do. @estewei?

core/src/main/scala/quasar/repl/prettify.scala
```diff
@@ -232,4 +231,11 @@ object Prettify {
...
+  def renderValues(src: List[Data]): List[List[String]] = {
```
mossprescott: Could use some input on whether this is a reasonable thing to do. @wemrysi?

jdegoes: No. The function claims to be total and referentially transparent, which is false since you are executing effectful code inside `Task`. You could try making `renderStream` more polymorphic (if indeed it can be so) in the `F[_]` of `Process`, such that you could run it with `Id`, or just factor out the pure logic and reuse it here and for an effectful `renderStream`.

core/src/main/scala/quasar/effect/Timing.scala
```diff
@@ -0,0 +1,82 @@
...
+    implicit def apply[S[_]: Functor](implicit S: TimingF :<: S): Ops[S] =
+      new Ops[S]
+  }
+
+  @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NonUnitStatements"))
+  final class Unsafe[S[_]: Functor](implicit S: TimingF :<: S)
+    extends LiftedOps[Timing, S] {
+
+    /** Raw nanoseconds value, which is unsafe in the sense that these values
+      * are only meaningful when compared to each other.
```
jedesah: I wonder if we can come up with a better name than `Unsafe`. Maybe `Raw`, `Internal`, `LowLevel`?

core/src/main/scala/quasar/fs/WriteFile.scala
```diff
@@ -102,6 +102,13 @@ object WriteFile {
     def append(dst: AFile, src: Process[F, Data]): Process[M, FileSystemError] =
       appendChunked(dst, src map (Vector(_)))
 
+    /** Appends the given data to the specified file, creating it if it doesn't
+      * exist. May not write every given value in the presence of errors, returns
+      * a [[FileSystemError]] for each input value that failed to write.
+      */
+    def appendThese(dst: AFile, data: Vector[Data]): M[Vector[FileSystemError]] =
+      unsafe.open(dst) flatMapF (h => unsafe.write(h, data) <* unsafe.close(h) map (_.right))
```
jedesah: I don't think this ensures we always close the handle.

We probably want something like:

`unsafe.open(dst).flatMap(h => EitherT.right(unsafe.write(h, data).onFinish(κ(unsafe.close))))`

core/src/main/scala/quasar/repl/prettify.scala
wemrysi: You should be able to `renderStream(Process.emitAll(src), src.length).toList` in this case. If that doesn't work, you may need to allow for a subtype of `F[_]` in the signature of `renderStream`.

core/src/main/scala/quasar/fs/WriteFile.scala
wemrysi: @jedesah Where does 'onFinish` come from?

core/src/main/scala/quasar/effect/Timing.scala
wemrysi: I would argue that this doesn't need to be separate from the other combinators, what is the advantage of doing so?

core/src/main/scala/quasar/fs/WriteFile.scala
jedesah: @wemrysi It's defined on `Task`. At least in `scalaz` `7.2`. Is it not in `7.1`?

jedesah: oops. Not `Task` nevermind! haha

jedesah: Don't we still have a problem though? Not immediately obvious the best to solve it though since `Free` does not have a concept of failure.

jedesah: Maybe that's why we don't do this and only expose it through `Process`?

jedesah: Ah, I see, we make the assumption there will be no untyped failure which I think is a fine assumption. Sorry for the noise.

core/src/main/scala/quasar/repl/ConsoleIO.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import quasar.Predef._
+
+import quasar.effect.LiftedOps
+
+import scalaz._
+
+sealed trait ConsoleIO[A]
+object ConsoleIO {
+  final case class PrintLn(message: String) extends ConsoleIO[Unit]
```
jedesah: I think I would probably go with `Print` and make a derived function `println` that adds `'\n'` to the `message`. We would have to add a `Print` term anyway if want that one day so probably fewer terms is better.

core/src/main/scala/quasar/effect/Timing.scala
mossprescott: I was just trying to discourage using it in the wrong way, which is a trap audit logging fell into. The point is that Ops contains the nicely encapsulated "user" API, while the actual case classes, Unsafe, and whatever else are all part of the implementation, even they _might_ be useful in some other circumstance.

In this case, I considered making `nanos` private in `Ops`, but decided that was too heavy-handed, because I wasn't sure it would _always_ make sense or even be possible to use `time[A](f: F[A])`.

core/src/main/scala/quasar/fs/WriteFile.scala
wemrysi: @jedesah If you're concerned about a particular implementation failing in such a way during the interpretation of `Write` that the `Close` is never reached, then I'd say it is the responsibility of that implementation to clean up any resources in that case. From the pov of the algebra, any errors during write should be returned as `FileSystemError`s which are then passed along after the `Close` is sequenced.

mossprescott: I'll add a comment because I think you're correct and I don't think it's obvious what's happening there.

core/src/main/scala/quasar/repl/prettify.scala
mossprescott: @wemrysi That works, although I don't understand why or how.

@jdegoes: I don't think I agree, but I'm happy to get rid of the code and not have the argument ;)

wemrysi: @mossprescott scalaz-stream has special syntax for `Process0` (see `Process0Syntax`) which is guaranteed to be pure, so you can convert it to/from collections.

core/src/main/scala/quasar/effect/Timing.scala
jedesah: I think it's reasonable to have it in a separate `Ops` class for that reason. But I do have a slight preference for a different name.

core/src/main/scala/quasar/repl/ConsoleIO.scala
mossprescott: I'm not sure it's that simple. The interpreter ends up using a Java PrintStream, which the lizard part of my brain is telling me will treat a `println` call differently than `print` with a newline. The docs suggest that's been fixed, however, and now flushing happens either way (http://docs.oracle.com/javase/7/docs/api/java/io/PrintStream.html):

> Optionally, a PrintStream can be created so as to flush automatically; this means that the flush method is automatically invoked after a byte array is written, one of the println methods is invoked, or a newline character or byte ('\n') is written.

So it seems safe to make that change, but I'm inclined to skip it because this effect is buried in the `repl` package and I don't think it's likely to get used elsewhere.

jedesah: Sounds more complicated than it initially appeared to me. Won't secretly hate you if you don't change it.

core/src/main/scala/quasar/effect/Timing.scala
wemrysi: While I'd still be fine with not separating it and just documenting thoroughly, if we are going to separate I think another name would be more appropriate as we use `Unsafe` elsewhere to mostly refer to resource safety. `Low` might be a reasonable candidate.

core/src/main/scala/quasar/repl/Main.scala
```diff
@@ -0,0 +1,152 @@
...
+      cfgPath      <- opts.config.fold(none[FsFile].point[MainTask])(cfg =>
+                        FsPath.parseSystemFile(cfg)
+                          .toRight(s"Invalid path to config file: $cfg")
+                          .map(some))
+
+      config       <- cfgPath.cata(CoreConfig.fromFile, CoreConfig.fromDefaultPaths).leftMap(_.shows)
+
+      // NB: for now, there's no way to add mounts through the REPL, so no point
+      // in starting if you can't do anything and can't correct the situation.
+      _            <- if (config.mountings.toMap.isEmpty) EitherT.left(Task.now("No mounts present"))
```
wemrysi: Could shorten to `MainTask.raiseError("No mounts present")`

core/src/main/scala/quasar/repl/Main.scala
```diff
@@ -0,0 +1,152 @@
...
+                      else ().point[MainTask]
+
+      coreApi      <- CoreEff.interpreter.liftM[MainErrT]
+      ephemeralApi =  CfgsErrsIO.toMainTask(MntCfgsIO.ephemeral) compose coreApi
+      _            <- mountAll[CoreEff](config.mountings) foldMap ephemeralApi
+
+      cfgRef       <- TaskRef(config).liftM[MainErrT]
+      durableApi   =  CfgsErrsIO.toMainTask(MntCfgsIO.durableFile(cfgRef, cfgPath)) compose coreApi
```
wemrysi: Is there a point to using the "durable" interpreter? Are there any repl operations that can affect the configuration?

core/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -0,0 +1,251 @@
...
+      } yield ()
+
+    cmd match {
+      case Help =>
+        P.println(HelpMessage)
+
+      case Debug(level) =>
+        RS.modify(_.copy(debugLevel = level)) *>
+          P.println(s"Set debug level: $level")
+      case SummaryCount(rows) =>
```
wemrysi: Very minor, but any reason not to add a newline before each `case` line? We have it before some and not others, I think it'd read nicer if they were consistent.

web/src/main/scala/quasar/api/package.scala
```diff
@@ -63,6 +67,18 @@ package object api {
           _.right[Response].point[Task])))
     }
 
+  /** Interprets errors into `Response`s, for use in web services. */
+  def toResponseOr(evalCfgsIO: MntCfgsIO ~> Task): CfgsErrsIOM ~> ResponseOr = {
```
wemrysi: I think this should still live somewhere under the `server` package. Could even go in that package object itself.

drostron:
Once comments are addressed, LGTM.

core/src/main/scala/quasar/repl/Repl.scala
mossprescott: I was doing a "grouping related cases" thing, but yeah, I agree it's a bit hard to read so I'll make it consistent.

web/src/main/scala/quasar/api/package.scala
mossprescott: Sure.

core/src/main/scala/quasar/repl/Main.scala
mossprescott: Not at this point, but I expect there will be at some point. The main reason to use it right now was just to avoid having to put together a separate interpreter providing the mounts.

core/src/main/scala/quasar/repl/ConsoleIO.scala
mossprescott: Sounds good. As long as it's out in the open :)

core/src/main/scala/quasar/repl/Main.scala
wemrysi: Could we use `ephemeralApi` until we need the durable one?

core/src/main/scala/quasar/effect/Timing.scala
mossprescott: Went with `LowLevel` as the apparent common ground.

mossprescott:
I believe I've addressed every comment, so if you're all happy I'd like to squash and rebase.

wemrysi:
LGTM, looking forward to seeing this merged!

jedesah:
:+1:

drostron:
👍🏽

mossprescott:
Rebasing got hairy so this all ended up in one commit. Thanks, everyone, for all the feedback!

mossprescott:
Merged via the script.

PR #1112, opened 2016-03-02
fix patent infinite loop in MongoDbIO's Catchable

mossprescott:
See SD-1426

One way to make this happen was to attempt to save a document at a path directly under the root of a MongoDB mount, whereas appending to the same path would give the correct error.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [SD-1426] fix hang in the MongoDB backend in certain error conditions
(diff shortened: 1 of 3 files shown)


drostron:
👍🏽

mossprescott:
@sellout: after merging with the `merge` script, this PR stills shows up as open, maybe because the commit was rewritten?

jedesah:
@mossprescott @sellout Ah, because normally Github would be smart enough to detect that the commits are now part of the target branch and the PR is no longer necessary?

PR #1113, opened 2016-03-02
Add the read-only config to the regression tests using the Free arch.

mossprescott:
See SD-1189

Put the additional "setup/insert" interpreter and some other useful stuff into FileSystemUT, modify tests to use it as needed, and mark some tests as either `skipped` with read-only backends or else explicitly run them only on read-write.

### DIFF
--- a/it/src/test/scala/quasar/fs/ManageFilesSpec.scala
+++ b/it/src/test/scala/quasar/fs/ManageFilesSpec.scala
@@ -17,7 +17,8 @@
 package quasar.fs
 
 import quasar.Predef._
-import quasar.Data
+
+import quasar.{Data, TestConfig}
 import quasar.fp._
 
 import pathy.Path._
@@ -25,7 +26,8 @@ import scalaz._, Scalaz._
 import scalaz.concurrent.Task
 import scalaz.stream._
 
-class ManageFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT) {
+class ManageFilesSpec extends FileSystemTest[FileSystem](
+  FileSystemTest.allFsUT.map(_.filterNot(fs => TestConfig.isMongoReadOnly(fs.name)))) {
   import FileSystemTest._, FileSystemError._, PathError2._
   import ManageFile._
 
@@ -39,9 +41,11 @@ class ManageFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT)
   def deleteForManage(run: Run): FsTask[Unit] =
     runT(run)(manage.delete(managePrefix))
 
-  fileSystemShould { _ => implicit run =>
+  fileSystemShould { fs =>
+    implicit val run = fs.testInterpM
+
     "Managing Files" should {
-      step(deleteForManage(run).runVoid)
+      step(deleteForManage(fs.setupInterpM).runVoid)
 
       "moving a file should make it available at the new path and not found at the old" >> {
         val f1 = managePrefix </> dir("d1") </> file("f1")
@@ -229,7 +233,7 @@ class ManageFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT)
         runLogT(run, p).runEither must beRight(anotherDoc)
       }
 
-      step(deleteForManage(run).runVoid)
+      step(deleteForManage(fs.setupInterpM).runVoid)
     }; ()
   }
 }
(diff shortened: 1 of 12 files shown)


core/src/main/scala/quasar/physical/mongodb/MongoDbIO.scala
```diff
@@ -341,7 +341,7 @@ object MongoDbIO {
       override def map[A, B](fa: MongoDbIO[A])(f: A => B) = fa map f
       def point[A](a: => A) = new MongoDbIO(Kleisli(_ => Task.now(a)))
       def bind[A, B](fa: MongoDbIO[A])(f: A => MongoDbIO[B]) = fa flatMap f
-      def fail[A](t: Throwable) = fail(t)
+      def fail[A](t: Throwable) = MongoDbIO.fail(t)
```
mossprescott: Note: this is the fix from #1112, without which _many_ of these tests would simply hang. I assume that PR will get merged first and this won't end up in this diff.

wemrysi:
@mossprescott Since the tests use the `FileSystem` api to insert data, what do you think about pushing the read-only concern into the interpreter itself? i.e. do we need to make the run/insert distinction explicit, or can we construct an interpreter from a read-only half and a write half that sends all operations we'd expect a read-only interpreter to perform to the read-only half and everything else to the write half?

mossprescott:
I would have to give that some thought. I think that _might_ be sufficient for what these tests are accomplishing, which is to verify that the behavior is the same in all cases where things do work. But it's also important to verify failure behavior in cases where things don't work, and your question makes me realize that this PR doesn't do any of that. Hmm.

wemrysi:
Hm, it would seem that we'd want to confirm that the correct subset of functionality works given a minimal set of permissions (which we should also define, I think we just assume what "read-only" means at this point). If a user has less than the minimum, we'd just assume things aren't going to work, there are many different combinations of topology and permissions that could lead to this scenario, I don't think we'll be able to test them all. I'm not even sure if we document the minimal set of permissions that are needed for the "full" slamdata experience either (i.e. the need to be able to list all databases in order to see more than one, etc).

mossprescott:
Updated and build looks good.

it/src/test/scala/quasar/fs/ManageFilesSpec.scala
```diff
@@ -39,9 +40,13 @@ class ManageFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT)
   def deleteForManage(run: Run): FsTask[Unit] =
     runT(run)(manage.delete(managePrefix))
 
-  fileSystemShould { _ => implicit run =>
+  fileSystemShould { fs =>
+    implicit val run = fs.testInterpM
+
+    if (!TestConfig.isMongoReadOnly(fs.name)) {
```
wemrysi: Can we filter the input set of filesystems instead?

it/src/test/scala/quasar/fs/WriteFilesSpec.scala
```diff
@@ -39,9 +40,13 @@ class WriteFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT)
   def deleteForWriting(run: Run): FsTask[Unit] =
     runT(run)(manage.delete(writesPrefix))
 
-  fileSystemShould { _ => implicit run =>
+  fileSystemShould { fs =>
+    implicit val run = fs.testInterpM
+
+    if (!TestConfig.isMongoReadOnly(fs.name)) {
```
wemrysi: Same question as for `ManageFilesSpec`

it/src/test/scala/quasar/fs/ManageFilesSpec.scala
mossprescott: Yes, I considered adding `fileSystemShouldFiltered(pred: FileSystemUT => Boolean)(examples: F ~> Task)` but couldn't think of a decent name, so punted. Any suggestions?

wemrysi: I was actually referring to the input to the `FileSystemTest` class above, i.e. something like
```scala
class ManageFilesSpec extends FileSystemTest[FileSystem](
  FileSystemTest.allFsUT map (_.filter(<some predicate>))
) { ... }
```

mossprescott:
Updated again.

wemrysi:
:+1: thanks for taking care of this!

mossprescott:
Merged via script.

PR #1114, opened 2016-03-03
[SD-1503] WIP

jedesah:
- Modify Regression spec framework to support requiring more than one dataset for a given regression test
- add two new datasets to mirror issue example
- add failing regression test for join with same field names


jedesah:
Superseded by #1152

PR #1116, opened 2016-03-04
[SD-1338] Fix duplicate field name leading to silent misbehaving

jedesah:
A duplicated alias or field name would lead to a column being dropped from the output instead of producing a clear compilation error message

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [SD-1338] Fix issue where a duplicated alias or fields with the same name would lead to a column being dropped from the output instead of producing a clear compilation error message
--- a/core/src/main/scala/quasar/compiler.scala
+++ b/core/src/main/scala/quasar/compiler.scala
@@ -38,7 +38,8 @@ trait Compiler[F[_]] {
   // HELPERS
   private type M[A] = EitherT[F, SemanticError, A]
 
-  private type CompilerM[A] = StateT[M, CompilerState, A]
+  private type CompilerStateT[F[_],A] = StateT[F, CompilerState, A]
+  private type CompilerM[A] = CompilerStateT[M, A]
 
   private def syntheticOf(node: CoExpr): List[Option[Synthetic]] =
     node.head._1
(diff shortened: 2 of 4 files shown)


CHANGELOG/revision.md
```diff
@@ -0,0 +1 @@
+- Fix issue where a duplicated alias or fields with the same name would lead to a column being dropped from the output instead of producing a clear compilation error message
```
mossprescott: Generally we're mentioning the issue number in these messages.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -40,6 +40,8 @@ trait Compiler[F[_]] {
 
   private type CompilerM[A] = StateT[M, CompilerState, A]
 
+  private type CompileStateT[M[_],A] = StateT[M, CompilerState, A]
```
mossprescott: This seems like a super confusing way of defining these two types. How about something like this:
```scala
private type CompilerStateT[F[_], A] = StateT[F, CompilerState, A]
private type CompilerM[A] = CompilerStateT[M, A]
```

core/src/main/scala/quasar/compiler.scala
```diff
@@ -315,82 +317,88 @@ trait Compiler[F[_]] {
...
-        val projs = projections.map(_.expr)
-
-        val syntheticNames: List[String] =
-          names.zip(syntheticOf(node)).flatMap {
-            case (Some(name), Some(_)) => List(name)
-            case (_,          _)       => Nil
-          }
+        (names.flatten(Option.option2Iterable) diff names.flatten(Option.option2Iterable).distinct).headOption.cata(
+          duplicateFieldName =>
+            EitherT.left[F, SemanticError,Fix[LogicalPlan]]((SemanticError.DuplicateFieldName(duplicateFieldName):SemanticError).point[F]).liftM[CompileStateT],
```
mossprescott: I think it would be better to capture _all_ the duplicated names, rather than just the first. Otherwise you have to fix one error at a time (scalac-style).

mossprescott:
After catching up on the discussion in the ticket, I think you should add a test(s) for cases with generated names as well. Maybe these two at least: `a.name, b.name` and `a.name, b.foo as name`. Both of those should also be errors.

jedesah:
Those tests would soon(ish) become obsolete by [SD-1503](https://slamdata.atlassian.net/browse/SD-1503) but I am not opposed to adding them now anyway to document current behavior.

core/src/main/scala/quasar/compiler.scala
jedesah: I thought about doing that but then I figured I was being too nitpicky and it would be annoying to deal with a `NonEmptyList` of duplicate names in the error class when the common case is a single duplicate.

I kind of wish the compiler was setup to return more than one `SemanticError`. I think that seems like a more general solution to this problem.

CHANGELOG/revision.md
jedesah: Can do.

core/src/main/scala/quasar/compiler.scala
jedesah: Okay with me.

jedesah:
@mossprescott Added the tests.

mossprescott:
LGTM

jedesah:
Closed via merge script.

PR #1118, opened 2016-03-05
[SD-1502] Fix Timeout exception in `ControlServiceSpec`

jedesah:


drostron:
It seems concerning that 30s isn't sufficient. Any hint as to why it isn't?

jedesah:
@drostron No idea. I don't know if you followed the slack discussion Friday afternoon. But I tried running the tests on my machine and artificially constraining the thread count to two like it should be on Travis and the tests still took `2s` to run. I can't think of anything else but Travis vms being really slow. But it is odd because I don't think Travis is that much slower then my machine for the rest of the build. Maybe they have some kind of security code around opening ports or something that makes such an operation drastically slower, that could make sense since we open more than one in rapid fire succession.

drostron:
Could the port be taken by another test? I'm able to get a timeout exception locally when that's the case. Utilizing `anyAvailablePorts` might hold promise.

jedesah:
@drostron to the rescue!

That actually makes seems like a rather plausible explanation. Let me investigate.

mossprescott:
If that's it, maybe use `ExclusiveExecution`?

jedesah:
@mossprescott I will look into it, but at this point I am thinking there may be a better solution, i.e being smarter about choosing ports, rather than throttling the build with `ExclusiveExecution`

jedesah:
This looks like it might have worked since we did not encounter a build failure. Will change the message in the changelog and see if it still passes the builds reliably.

jedesah:
Nevermind, it didn't work.

jedesah:
@drostron Do you think this is good to go if this last build passes?

drostron:
👍🏾

jedesah:
Closed via merge script.

PR #1119, opened 2016-03-05
remove the last two uses of shapeless's deriveEqual

mossprescott:
This improves incremental compile time for the two files involved (workflowbuilder and logicalplan) by from about 100s to less than 10s.

From-scratch compile time is reduced by about 50% for me.

Will try to confirm some improvement in Travis before merging.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- improve compile time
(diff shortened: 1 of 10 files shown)


sellout:
Nice. Hopefully we’ll be able to use deriving more consistently and with less performance hit with export-hook, or when Typeclassic is a thing.

mossprescott:
Looks like this is saving about a minute or so in each build in Travis. Building again after finding a line of debug code I've been meaning to clean up...

drostron:
👍🏽

mossprescott:
Merged via the script.

PR #1121, opened 2016-03-07
SD-1329 Fix moving and deleting toplevel folders (databases) on mongo mounts

wemrysi:
As reported in SD-1329 moving and deleting top-level folders in a mongo mount didn't work. This fixes the problems, which reduced to one of two issues:

* Double "db name escaped" temp file paths from mongo
* Attempts to modify objects in the mongo "system" namespace

these were fixed by, respectively:

* Unescaping database names obtained from a path before using them to construct a new path
* Explicitly handling moving of databases, skipping over objects in the "system" namespace

In attempting to debug this issue, I also discovered there are situations where mongo exceptions weren't being surfaced properly, which is now fixed so unhandled mongo error messages should be exposed via a 500 error in the api.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [SD-1329] Fix moving and deleting mongo databases via filesystem
--- a/core/src/main/scala/quasar/effect/Failure.scala
+++ b/core/src/main/scala/quasar/effect/Failure.scala
@@ -92,6 +92,13 @@ object Failure {
       new Ops[E, S]
   }
 
+  def mapError[D, E](f: D => E): Failure[D, ?] ~> Failure[E, ?] =
+    new (Failure[D, ?] ~> Failure[E, ?]) {
+      def apply[A](fa: Failure[D, A]) = fa match {
+        case Fail(d) => Fail(f(d))
+      }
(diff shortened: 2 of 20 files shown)


mossprescott:
LGTM, unless you think it's worth investigating `copyDb`.

mossprescott:
A last note: for what it's worth, it looks to me like the old impl (in `mongodb/filesystem.scala`) just moved collections one at a time, without bothering to skip the system stuff, and I vaguely recall that it would fail with a MongoDB error about that. It doesn't look like we had a test in the old `FileSystemSpecs`.

wemrysi:
I started out with `copydb` and then realized it would only work in the case where a database was effectively being renamed, i.e. `mv /foo/ /bar/`, however it doesn't work in cases like `mv /foo/ /bar/baz/quux/` when the `/bar/` db already exists (and adds more complexity to the implementation even if it doesn't). `copydb` might eventually be faster in certain situations (in the current implementation I think it does more-or-less what we're doing here) and we could decide to optimize there if it ever becomes so.

As for preservation of indicies, while it may possibly be slower, this method does appear to preserve indices when moving collections between databases.

mossprescott:
LGTM

PR #1124, opened 2016-03-08
Cleanup post Free-arch conversion

wemrysi:
A bunch of cleanup now that nothing depends on the legacy `Backend` code
* Delete all the `Backend` code and supporting files
* Remove the `2` suffix from `MountConfig2`, `MountingsConfig2` and `EnvironmentError2`
* Remove the dependency on the synchronous mongo java driver

I left `PathError2` as-is, mostly as a reminder to convert `LogicalPlan` to `scala-pathy`.

Anything else I forgot or overlooked?


sellout:
So happy to see the line counts in this PR 🎉

sellout:
This is one of those PRs where I really want to look at the _unchanged_ code more than the changed code … kind of.

sellout:
I think this looks pretty good (modulo successful builds). Removing the sync dependency alone should have triggered compilation failures that would have cascaded through most of the spots you had to clean up, right?

wemrysi:
I removed the sync mongodb dependency last, so it only triggered one issue as most of the code that used it had already been deleted.

wemrysi:
Ugh, build failures due to timeouts, #1118 can't get merged soon enough!

PR #1125, opened 2016-03-08
Support SSL connections to MongoDB

mossprescott:
This PR includes a small code change, adds a dependency, and most importantly includes a script that generates test certificates and details how to get it all to work.

@damonLL, please have a look and comment on the documentation here. Even better, pull the build and the script and see if you can get it to work by following the instructions there.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/util.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/util.scala
@@ -84,14 +84,64 @@ object util {
             Task.now(())
         }
 
-    def createClient(cs: ConnectionString) =
+    val InvalidHostNameAllowedProp = "invalidHostNameAllowed"
+
+    @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NonUnitStatements"))
+    def settings(cs: ConnectionString, invalidHostNameAllowed: Boolean): Task[MongoClientSettings] = Task.delay {
+      import com.mongodb.connection._
+
+      // NB: this is apparently the only way to get from a ConnectionString to a
+      // MongoClient while also inspecting/modifying anything in the settings.
+      // This is following `MongoClients.create(ConnectionString)`, and will have
+      // to be revisited if a driver release adds additional settings objects.
+      val settings = MongoClientSettings.builder
+
+      settings.clusterSettings(ClusterSettings.builder
+        .applyConnectionString(cs)
+        .build)
+
+      settings.connectionPoolSettings(ConnectionPoolSettings.builder
+        .applyConnectionString(cs)
+        .build)
+
+      settings.credentialList(cs.getCredentialList)
+
+      settings.serverSettings(ServerSettings.builder
+        .build)
+
+      settings.socketSettings(SocketSettings.builder
+        .applyConnectionString(cs)
+        .build)
+
+      val sslSettings = SslSettings.builder
+        .applyConnectionString(cs)
+        .invalidHostNameAllowed(invalidHostNameAllowed)
+        .build
+      settings.sslSettings(sslSettings)
+
+      // NB: Netty _must_ be used if SSL is required, but we do not use it by default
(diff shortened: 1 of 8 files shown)


core/src/main/scala/quasar/physical/mongodb/util.scala
```diff
@@ -88,14 +88,62 @@ object util {
...
+        .applyConnectionString(cs)
+        .invalidHostNameAllowed(invalidHostNameAllowed)
+        .build
+      settings.sslSettings(sslSettings)
+
+      // NB: Netty _must_ be used if SSL is required, but we do not use it by default
+      // mostly because it seems to cause the REPL to fail to exit cleanly. If necessary,
+      // it can also be forced using a system property (see the README).
+      if (sslSettings.isEnabled)
+        settings.streamFactoryFactory(new com.mongodb.connection.netty.NettyStreamFactoryFactory())
```
wemrysi: Mind wrapping this with braces since there is no `else` clause?

core/src/main/scala/quasar/physical/mongodb/util.scala
```diff
@@ -88,14 +88,63 @@ object util {
...
+        .build)
+
+      settings.socketSettings(SocketSettings.builder
+        .applyConnectionString(cs)
+        .build)
+
+      val sslSettings = SslSettings.builder
+        .applyConnectionString(cs)
+        .invalidHostNameAllowed(invalidHostNameAllowed)
+        .build
```
drostron: `.build` could throw, but since we don't run on Java 1.6 seems ok to forgo wrapping in `Task`.

core/src/main/scala/quasar/repl/Main.scala
```diff
@@ -137,7 +137,7 @@ object Main {
 
       coreApi      <- CoreEff.interpreter.liftM[MainErrT]
       ephemeralApi =  CfgsErrsIO.toMainTask(MntCfgsIO.ephemeral) compose coreApi
-      _            <- mountAll[CoreEff](config.mountings) foldMap ephemeralApi
+      _            <- (mountAll[CoreEff](config.mountings) foldMap ephemeralApi).flatMapF(_.point[Task])
```
drostron: Is this addition necessary?

mossprescott: Yes. What was happening there is `mountAll foldMap ephemeralApi` was returning `String \/ Unit`, and then we were binding that to `_`, discarding the error in the process. Flatmap/pointing puts the error on the left in MainTask, which gets it logged and stops forward progress, which is what we want for now in the case that mounting fails.

The following commit applies the same change to the api server.

core/src/main/scala/quasar/physical/mongodb/util.scala
mossprescott: Wow, that's thorough. :smile: 

I think this code is arguably effectful if only because it does `new NettyStreamFactoryFactory()`, so I'm gonna go ahead and wrap it in Task. Note: That won't at the moment change the behavior anyway because it was already being _called_ from within Task.delay, but I think it makes sense to make it clear and avoid future problems.

core/src/main/scala/quasar/repl/Main.scala
drostron: Makes sense. Thanks for the explanation.

core/src/main/scala/quasar/physical/mongodb/util.scala
jdegoes: Again, it doesn't matter *where code is being called from*. If it's effectful, it *must* be wrapped in `Task` in order to maintain referential transparency.

The default assumption with any Java code should be that (a) it's effectful, and (b) it may return `null`, unless a detailed analysis of its documentation reveals stronger guarantees.

drostron:
👍🏾, once @damonLL signs off.

mossprescott:
Merged via the script

PR #1127, opened 2016-03-10
Make the Null type comparable.

sellout:
Fixes SD-1505.

Also modify Mongo plans for the associated expression guards (especially
when converting from expressions to JS) to keep them effecient and
correct.

### DIFF
--- a/CHANGELOG/minor.md
+++ b/CHANGELOG/minor.md
@@ -1,0 +1,1 @@
+- [SD-1505] generalize `Neq` to work for `Top`, which allows `!= null` to work like `is not null`
--- a/core/src/main/scala/quasar/std/relations.scala
+++ b/core/src/main/scala/quasar/std/relations.scala
@@ -35,7 +35,7 @@ trait RelationsLib extends Library {
     basicUntyper)
 
   val Neq = Mapping("(<>)", "Determines if two values are not equal",
-    Type.Bool, Type.Comparable :: Type.Comparable :: Nil,
+    Type.Bool, Type.Top :: Type.Top :: Nil,
     noSimplification,
     partialTyper {
       case Type.Const(data1) :: Type.Const(data2) :: Nil => Type.Const(Data.Bool(data1 != data2))
--- a/core/src/test/scala/quasar/physical/mongodb/planner.scala
+++ b/core/src/test/scala/quasar/physical/mongodb/planner.scala
@@ -850,25 +850,10 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
                  Selector.Doc(BsonField.Name("baz") ->
                    Selector.Type(BsonType.Bool)))),
              Selector.And(
-               Selector.Or(
-                 Selector.Doc(BsonField.Name("bar") ->
-                   Selector.Type(BsonType.Int32)),
-                 Selector.Doc(BsonField.Name("bar") ->
-                   Selector.Type(BsonType.Int64)),
-                 Selector.Doc(BsonField.Name("bar") ->
-                   Selector.Type(BsonType.Dec)),
-                 Selector.Doc(BsonField.Name("bar") ->
-                   Selector.Type(BsonType.Text)),
-                 Selector.Or(
-                   Selector.Doc(BsonField.Name("bar") ->
-                     Selector.Type(BsonType.Date)),
-                   Selector.Doc(BsonField.Name("bar") ->
-                     Selector.Type(BsonType.Bool)))),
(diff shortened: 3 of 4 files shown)


core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -535,6 +539,13 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
                   Selector.Doc(f -> Selector.Type(BsonType.Int32)),
                   Selector.Doc(f -> Selector.Type(BsonType.Int64)),
                   Selector.Doc(f -> Selector.Type(BsonType.Dec))))
+            case Type.Null ⨿ Type.Int ⨿ Type.Dec ⨿ Type.Interval =>
```
mossprescott: Surprised to not see a case here for `Int ⨿ Dec` here (since it does appear at line 294). Isn't one other wrong?

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -974,13 +985,15 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
                 // $lt(expr, $literal(Bson.Timestamp(Instant.ofEpochMilli(0), 0)))))
                 $lt(expr, $literal(Bson.Regex("", "")))))
             // NB: Some explicit coproducts for adjacent types.
-            case Type.Int ⨿ Type.Dec ⨿ Type.Str =>
+            case Type.Int ⨿ Type.Dec ⨿ Type.Str
+               | Type.Int ⨿ Type.Dec ⨿ Type.Interval ⨿ Type.Str =>
```
mossprescott: So, `Interval` was missing here (and in the next case)?

sellout: Actually, the diff makes this look a bit confusing. I just collapsed two identical cases with `|`, then I added a new pair that includes `Type.Null` and compares with `$lte` rather than `$lt`.

mossprescott:
Many new expressions that will type check now, I believe. `foo < null`, for one. 

Also, including `null` in the typecheck for comparable means new behavior for `foo < bar` if either or both are null, does it not?

What is the expected behavior for these expressions in aggregation ops? In Selectors? I think it's worth having some tests. Ideally, we would add these expressions to our hypothesized cross-stage verification (where you specify an expression and we evaluate it in 1) constant-folding, 2) BlackShield, 3) aggregation ops, 4) selector, and 5) JavaScript.)

Since we don't have that, I guess we need more regression tests. :disappointed:

core/src/main/scala/quasar/physical/mongodb/planner.scala
sellout: No – it all depends on how that particular stage handles types (because, _of course_ it’s different for JS, selectors, and expressions). In the selector, testing for `Int` is independent of testing for `Dec`, but in JS, they overlap (ideally ints are stored using `NumberInt` or `NumberLong`, but we have to allow for them being a standard JS `Number`, which overlaps with `Dec`’s definition). Yes, it _is_ terribly confusing. But the explicit `⨿` cases just try to eliminate duplicated checks.

A better way would be simplifiers that rewrote conditions (but I’m not sure if that would work for all the expression cases), eliminating duplication.

sellout:
That’s a good point. Yeah, for Null, we do now have `<`/`<=`, etc. We could restrict `Null` to `=`/`!=`, but it’s … uh … “Comparable instance” is trivial. The trouble with having it be actually Comparable is that our type inference is not perfect, so it means that with this PR, `null` will compare as less than ints, strings, etc. and that is something that is likely to happen accidentally much more often than accidentally comparing Int and String, because of `null` used for missing values.

So now I’m leaning toward restricting it to just `=`/`!=`.

mossprescott:
Exactly. If restricting it it to `=`/`!=` is feasible, it sounds like a much better idea to me.

jdegoes:
In MRA there is a total ordering on values, which is required for a sorting operation to have any meaning on heterogeneous values. Of course, we can and will improve type inference with time.

sellout:
@jdegoes Yeah, but our ordering is not the same as Mongo’s ordering, which is what this ends up using.

sellout:
The updated version of this just changes `Neq` to work on `Top` (like `Eq` already does). This actually just simplifies various tests.

mossprescott:
:+1:

Deleting type-checks in plans is even better than deleting code :smile:

PR #1128, opened 2016-03-11
Two fixes for /metadata 

mossprescott:
These turned up in looking at SD-1331, but may not entirely account for what's going on there.

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/view.scala
+++ b/core/src/main/scala/quasar/fs/mount/view.scala
@@ -187,7 +187,10 @@ object view {
           })
 
         case FileExists(file) =>
-          query.fileExists(file).map(_ || views.contains(file)).run
+          if (views.contains(file))
+            true.point[Free[S, ?]]
+          else
+            query.fileExists(file)
       }
     }
   }
--- a/core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
@@ -160,9 +160,9 @@ private final class QueryFileInterpreter[C](
       }).liftM[QRT].liftM[WorkflowExecErrT]
 
     case FileExists(file) =>
-      collFromPathM(file)
-        .flatMap(MongoDbIO.collectionExists(_).liftM[FileSystemErrT])
-        .run.liftM[QRT].liftM[WorkflowExecErrT]
+      Collection.fromPathy(file).fold(
+        κ(false.point[MQ]),
+        coll => MongoDbIO.collectionExists(coll).liftM[QRT].liftM[WorkflowExecErrT])
   }
 
   ////
--- a/core/src/test/scala/quasar/fs/TraceFS.scala
+++ b/core/src/test/scala/quasar/fs/TraceFS.scala
@@ -23,7 +23,14 @@ import quasar.fp._
 import pathy.{Path => PPath}, PPath._
 import scalaz._, Scalaz._
 
-/** Interpreter that just records a log of the actions that are performed. */
+/** Interpreter that just records a log of the actions that are performed. Only a
+  * handful of operations actually produce results:
+  * - ManageFile.TempFile always produces a file called "tmp" either alongside the
+  *     argument (if it's a file path) or inside it (if it's a dir).
+  * - QueryFile.ListContents and .FileExists respond based on a map of directory
+  *     paths to sets of content nodes. `rootDir` implicitly exists (even if empty),
+  *     otherwise `ls` gives a pathError for any dir that does not appear in the map.
+  */
 object TraceFS {
   val FsType = FileSystemType("trace")
 
@@ -50,7 +57,7 @@ object TraceFS {
           case FileExists(file)     =>
             ls(fileParent(file)).fold(
               κ(false),
-              _.contains(fileName(file).right)).right
+              _.contains(fileName(file).right))
         }))
   }
 
--- a/core/src/test/scala/quasar/fs/mount/view.scala
+++ b/core/src/test/scala/quasar/fs/mount/view.scala
@@ -341,30 +341,24 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
 
   "QueryFile.fileExists" should {
     "behave as underlying interpreter" ! prop { file: AFile =>
-      val program = query.fileExists(file).run
+      val program = query.fileExists(file)
 
       val ops = traceInterp(program, Map())._1
 
       val hasFile = {
         val paths = Map(fileParent(file) -> Set(fileName(file).right[DirName]))
-        val expected = (ops, true.right)
-        viewInterp(Views.empty, paths, program) must_== expected
+        viewInterp(Views.empty, paths, program) must_==((ops, true))
       }
       val noFile = {
-        val expected = (ops, false.right)
-        viewInterp(Views.empty, Map(), program) must_== expected
+        viewInterp(Views.empty, Map(), program) must_==((ops, false))
       }
       hasFile and noFile
     }
 
     "return true if there is a view at that path" ! prop { (file: AFile, lp: Fix[LogicalPlan]) =>
-      val program = query.fileExists(file).run
+      val program = query.fileExists(file)
 
-      val ops = traceInterp(program, Map())._1
-
-      val expected = (ops, true.right)
-
-      viewInterp(Views(Map(file -> lp)), Map(), program) must_== expected
+      viewInterp(Views(Map(file -> lp)), Map(), program) must_==((Vector(), true))
     }
   }
 }
--- a/it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
+++ b/it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
@@ -235,7 +235,7 @@ class MongoDbFileSystemSpec
             val out = renameFile(file, κ(FileName("out")))
 
             def check0(expr: sql.Expr) =
-              (run(query.fileExists(file).run).run ==== false.right) and
+              (run(query.fileExists(file)).run ==== false) and
               (errP.getOption(
                 runExec(query.executeQuery(expr, Variables.fromMap(Map()), out))
                   .run.value.run
@@ -277,6 +277,24 @@ class MongoDbFileSystemSpec
 
           (runLogT(run, p) <* runT(run)(manage.delete(tdir)))
             .runEither must beRight(contain(FileName("foobar").right[DirName]))
+        }
+      }
+
+      "File exists" >> {
+        "for missing file at root (i.e. a database path) should succeed" >> {
+          val tfile = rootDir </> file("foo")
+
+          val p = query.fileExists(tfile)
+
+          run(p).run must_== false
+        }
+
+        "for missing file not at the root (i.e. a collection path) should succeed" >> {
+          val tfile = rootDir </> dir("foo") </> file("bar")
+
+          val p = query.fileExists(tfile)
+
+          run(p).run must_== false
         }
       }
 
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,2 @@
+- Fix /metadata for certain nonexistent files to produce `404 File Not Found` (instead of `400 BadRequest` with a confusing message).
+- Fix /metadata to succeed when views exist at those same paths, and others.
--- a/core/src/main/scala/quasar/fs/Empty.scala
+++ b/core/src/main/scala/quasar/fs/Empty.scala
@@ -102,7 +102,7 @@ object Empty {
             fsPathNotFound(d)
 
         case QueryFile.FileExists(_) =>
-          false.right.point[F]
+          false.point[F]
       }
     }
 
--- a/core/src/main/scala/quasar/fs/InMemory.scala
+++ b/core/src/main/scala/quasar/fs/InMemory.scala
@@ -193,7 +193,7 @@ object InMemory {
         ls(dir) map (r => if (dir === rootDir) r.getOrElse(Set()).right else r)
 
       case FileExists(file) =>
-        contentsL.st.map(_.contains(file).right)
+        contentsL.st.map(_.contains(file))
     }
 
     private def evalPlan[A](lp: Fix[LogicalPlan])
--- a/core/src/main/scala/quasar/fs/QueryFile.scala
+++ b/core/src/main/scala/quasar/fs/QueryFile.scala
@@ -67,7 +67,7 @@ object QueryFile {
     extends QueryFile[FileSystemError \/ Set[PathName]]
 
   final case class FileExists(file: AFile)
-    extends QueryFile[FileSystemError \/ Boolean]
+    extends QueryFile[Boolean]
 
   @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.NonUnitStatements"))
   final class Ops[S[_]](implicit S0: Functor[S], S1: QueryFileF :<: S)
@@ -173,8 +173,14 @@ object QueryFile {
     }
 
     /** Returns whether the given file exists. */
-    def fileExists(file: AFile): M[Boolean] =
-      EitherT(lift(FileExists(file)))
+    def fileExists(file: AFile): F[Boolean] =
+      lift(FileExists(file))
+
+    /** Returns whether the given file exists, lifted into the same monad as
+      * the rest of the functions here, for convenience.
+      */
+    def fileExistsM(file: AFile): M[Boolean] =
+      fileExists(file).liftM[FileSystemErrT]
 
     ////
 
--- a/core/src/main/scala/quasar/fs/WriteFile.scala
+++ b/core/src/main/scala/quasar/fs/WriteFile.scala
@@ -145,7 +145,7 @@ object WriteFile {
                      (implicit QF: QueryFile.Ops[S], MF: ManageFile.Ops[S])
                      : Process[M, FileSystemError] = {
 
-      QF.fileExists(dst).liftM[Process].ifM(
+      QF.fileExistsM(dst).liftM[Process].ifM(
         shouldNotExist(dst).liftM[Process],
         saveChunked0(dst, src, MoveSemantics.FailIfExists))
     }
@@ -163,7 +163,7 @@ object WriteFile {
                    (implicit QF: QueryFile.Ops[S], MF: ManageFile.Ops[S])
                    : M[Vector[FileSystemError]] = {
 
-      QF.fileExists(dst).ifM(
+      QF.fileExistsM(dst).ifM(
         shouldNotExist(dst) map (Vector(_)),
         saveThese0(dst, data, MoveSemantics.FailIfExists))
     }
@@ -173,7 +173,7 @@ object WriteFile {
                       (implicit QF: QueryFile.Ops[S], MF: ManageFile.Ops[S])
                       : Process[M, FileSystemError] = {
 
-      QF.fileExists(dst).liftM[Process].ifM(
+      QF.fileExistsM(dst).liftM[Process].ifM(
         saveChunked0(dst, src, MoveSemantics.FailIfMissing),
         shouldExist(dst).liftM[Process])
     }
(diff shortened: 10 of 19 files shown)


core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
```diff
@@ -161,9 +161,15 @@ private final class QueryFileInterpreter[C](
       }).liftM[QRT].liftM[WorkflowExecErrT]
 
     case FileExists(file) =>
-      collFromPathM(file)
-        .flatMap(MongoDbIO.collectionExists(_).liftM[FileSystemErrT])
-        .run.liftM[QRT].liftM[WorkflowExecErrT]
+      // NB: there can never be a collection at a path directly under the root
+      // (only a database), but it is not an error to ask the question.
```
jedesah: :+1: For a clear and concise comment!

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -347,12 +347,16 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
+        viewInterp(Views.empty, paths, program) must_==((ops, expected))
       }
       val noFile = {
-        val expected = (ops, false.right)
-        viewInterp(Views.empty, Map(), program) must_== expected
+        val expected =
+          if (fileParent(file) ≟ rootDir)
+            false.right
+          else
+            FileSystemError.pathError(PathError2.pathNotFound(fileParent(file))).left
```
jedesah: I'm getting a little tired and will probably take another fresh look at this tmr morning. But it strikes me as odd that we expect a `PathNotFound` as a return value to `FileExists`, especially since in the `MongoDbFileSystemSpec` below we seem to be expecting `false.right` even though the parent directory does not exists which leads me to believe that the desired behavior for `FileExists` is to always return `true` or `false`, but never `NotFound`.

mossprescott: I think it's a good question.

For the MongoDB FS, I think clearly the right answer is always `false` and never `pathNotFound`, because in principle any path at all _could_ contain a file, so it's not like something about the path is making it impossible to answer the question. By the way, it may not be obvious, but the change to MongoDb is to stop it producing `invalidPath`, so I still that's legit regardless.

On the other hand, for `hierarchical`, it may make sense to produce `pathNotFound` in the case that the path is outside any mounted filesystem, which is what's happening here. That would make it clear that in the current configuration, no file could _ever_ appear at that path.

On the third hand, I'm pretty sure we actually would want the API to produce a 404 in that case, and the kicker is that it does, because `false` actually gets mapped to `pathNotFound` at that level.

So at the moment it's academic. However, given that the type is `FileSystemError \/ Boolean` (and we probably need to keep that, in case of other errors), I don't think there's a good way to express that this method doesn't produce `pathNotFound`, unless of course we want to go with something like `assertFileExists(file: AFile): FileSystemError \/ Unit` (or `Option[FileSystemError]`, as you're probably thinking ;).

Also, one way or the other, we still need the view interpreter to handle these error cases, because the idea is it's possible to mount views literally anywhere, regardless of what problems those paths might cause for the other interpreters. So I think this bug fix is mostly orthogonal to the issue of whether not this particular situation is handled this way.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -189,7 +189,14 @@ object view {
           })
 
         case FileExists(file) =>
-          query.fileExists(file).map(_ || views.contains(file)).run
+          query.fileExists(file).run.map(_ match {
+            case \/-(x) =>
+              (x || views.contains(file)).right
+            case -\/(err @ PathError(PathNotFound(_))) =>
```
mossprescott: After writing my response to @jedesah's comment, I'm actually thinking this should catch _all_ errors. Regardless of what problems the underlying interpreter has with the path, if there's a view mounted, there's a file there as far as this filesystem is concerned (because all references to that path will get re-written by this interpreter.)

Therefore I'll probably change this to a simple `fold` that handles all errors the same way. Make sense?

core/src/test/scala/quasar/fs/mount/view.scala
wemrysi: I'm still not convinced the return type needs to be `FileSystemError \/ Boolean`, I think we could just return a `Boolean`. If a particular implementation wants to interpret into an effect with an implementation-specific error type, it can, but from the view of the filesystem abstraction I don't see how this could possibly fail.

jedesah: I can get behind that. This feels like deja vu. Did we already have a discussion on this topic before? Does anyone remember where that is?

wemrysi: We discussed it previously in #1044

jedesah: @wemrysi Thx for the link. That helps me refresh my mind on the issue. I think having it return just `Boolean` is probably the right answer. I do mention in that thread that we should make sure the behavior is coherent, i.e. that we should make sure all mongo specific errors are inserted in the mongodb interpreter as opposed to the term result.

core/src/main/scala/quasar/fs/mount/view.scala
jedesah: That makes sense to me.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
jedesah: I wonder though, if we add `FileSystemErrT` to the mongodb interpreter monad stack if it makes sense to put this error in there instead of returning `false` conveying in some way the fact that it is IMPOSSIBLE for there to ever be a file here.

core/src/test/scala/quasar/fs/TraceFS.scala
```diff
@@ -48,9 +56,7 @@ object TraceFS {
           case Explain(lp)          => (Vector.empty, \/-(ExecutionPlan(FsType, lp.toString)))
           case ListContents(dir)    => ls(dir)
           case FileExists(file)     =>
-            ls(fileParent(file)).fold(
-              κ(false),
-              _.contains(fileName(file).right)).right
+            ls(fileParent(file)).map(_.contains(fileName(file).right))
```
jedesah: What is the rationale behind this change?

mossprescott: It makes it possible to test the behavior of the view interpreter when the underlying interpreter fails. As it was, it was always hiding those failures by converting them to `false.right`.

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
```diff
@@ -280,6 +280,24 @@ class MongoDbFileSystemSpec
         }
       }
 
+      "File exists" >> {
+        "for missing file at root (i.e. a database path) should succeed" >> {
+          val tfile = rootDir </> file("foo")
+
+          val p = query.fileExists(tfile)
+
+          runT(run)(p).runEither must beRight(false)
```
jedesah: So, as per my comment above, we should consider whether this should yield a left since it is a MongoDB particularity. Maybe that's not a good idea, just wanted to put a comment to remember to change this if we go that route.

core/src/test/scala/quasar/fs/TraceFS.scala
```diff
@@ -23,7 +23,15 @@ import quasar.fp._
 import pathy.{Path => PPath}, PPath._
 import scalaz._, Scalaz._
 
-/** Interpreter that just records a log of the actions that are performed. */
+/** Interpreter that just records a log of the actions that are performed. Only a
```
jedesah: Total side comment, but I have wondered in the past what `TraceFS` does more than `Free` since `Free` is also something that "records" what needs to be done. Maybe `Free` records what needs to be done and `TraceFS` records what was actually done.

Just wondering if it would be possible to compare different instances of `Free[S,A]` for equality instead of interpreting them and inspecting the traces.

core/src/main/scala/quasar/fs/mount/view.scala
jedesah: Interestingly though, if we changed the result type of `FileExists` to `Boolean` instead of `FileSystemError \/ Boolean`, this would make it harder for the `view` interpreter to "catch" problems with an underlying interpreter, but return `true` as long as there is a view that's mounted there. Or would it?

core/src/test/scala/quasar/fs/TraceFS.scala
jedesah: Ah, that makes sense.

core/src/main/scala/quasar/fs/mount/view.scala
mossprescott: I was just writing a reply on that very question.

I think for FileExists there's an easy solution to that, which is that the view interpreter can simply check for a view _first_, and not call the underlying interpreter if it can return `true` from it's own information.

jedesah: That makes a lot of sense. Too much sense actually, I am jealous I did not think of that myself!

mossprescott: Does a similar argument apply to ListContents?

1) If an error occurs the prevents the mongodb interpreter from determining the contents of a directory (say, a connection failure), then it can produce an error in some Failure effect. The view interpreter wouldn't be able to intercept that, but that's OK because it would have no way of producing a correct result anyway.

2) if a path is constructed in such a way that it could never have children _in MongoDB_, then the interpreter would have to produce a value so as to allow the view interpreter to overlay any views (which can exist at _any_ path). But what value?

Right now, the answer is `pathNotFound.left: FileStsemError \/ Set[PathName]`, but another option would be `none: Option[Set[PathName]]` and yet another would be simply `Set[Path]()`. I'm not sure I like any of those all that much.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
wemrysi: If we're interested in conveying this sort of information, maybe we should encode it into the return type of `FileExists` (and probably rename `FileExists` to something else to communicate the additional behavior). I.e. return a type that has cases like "exists", "doesn't exist", "couldn't exist", etc.

I'm not sure if we need this, but we could then implement `fileExists` as a derviative of this more precise operation that just throws away information, collapsing "doesn't exist" and "couldn't exist" into `false`.

It doesn't seem like asking if a path that could never exist exists should result in an error though, when `false` is an accurate answer.

core/src/main/scala/quasar/fs/mount/view.scala
wemrysi: What is undesirable about the view filesystem handling the `pathNotFound` returned by the underlying filesystem?

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
jedesah: I guess that could be a solution. The disadvantage of it being that it does not allow different interpreters to convey whether THEY have any special cases or not and what the nature of the special cases unlike in the original solution you proposed of having it be part of the interpreter monad stack.

core/src/main/scala/quasar/fs/mount/view.scala
mossprescott: I'm just asking whether `FileSystemError \/ Set[PathName]` makes sense as a type, but I'm willing to drop it ;)

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
wemrysi: I guess what I'm trying to get at is that it seems surprising, as a client of the API, to receive an error when asking if a file exists because a particular implementation happens to know it could never exist, when `false` would answer the question asked adequately.

core/src/main/scala/quasar/fs/mount/view.scala
wemrysi: It seems like `FileSystemError \/ Set[PathName]` is an improvement over `Set[PathName]` as the latter conflates an empty directory with one that doesn't exist, whereas the former allows us to communicate that the operation is only defined for existing directories.

mossprescott: Yeah, kinda. Anyway, like I said, I'm over it.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
jdegoes: > I guess what I'm trying to get at is that it seems surprising, as a client of the API, to receive an error when asking if a file exists because a particular implementation happens to know it could never exist, when false would answer the question asked adequately.

:+1: An existence API should never error merely because a file cannot exist. 

Think of it from the perspective of the SlamData app (a prototypical Quasar app): the app wants to check if a file exists and will branch off the true/false return value. But if suddenly, some "falses" will be encoded by an error, then the app has to know that fact, and then internally capture and decode the error and treat it as equivalent to the "false" case.

core/src/test/scala/quasar/fs/TraceFS.scala
mossprescott: That is a thought-provoking question, and not one I can readily answer. I think the practical fact is that you cannot compare the two things we have in these tests, which in general are a `Free[FileSystem, A]` and a `Free[ViewFileSystem, A]`, because types (even if you could deal with the extra effects in the view program). But I'm sure there is a more subtle answer, along the lines of to be able to compare them, you would have to impose constraints X, Y, and Y…

jdegoes:
> Total side comment, but I have wondered in the past what TraceFS does more than Free since Free is also something that "records" what needs to be done. Maybe Free records what needs to be done and TraceFS records what was actually done.

`Free` describes a program in `F`. The structure of the program is dynamic, and may change as it's being executed. In order to explore its structure, you need to execute it -- which corresponds to feeding it the values that it's reading from its environment. If you feed different values, you may get different programs.

In general, `Free[F, A]` can only be compared with `Free[F, A]`, of course, you can compare with `Free[G,A]` if you have `G ~> F` (i.e. `F` is "bigger" than `G`).

However, to perform the comparison, you must actually execute the program by feeding it values. Since the values you feed it determine the program's continuation, any comparison you do would be valid *strictly* for the specific values you fed it. 

So in the best case, you can compare `Free[F, A]` with `Free[F, A]` and determine that they yield the same `A` for the same input values. But to say they represent the same program, you have to prove this **for all possible input values**. Since that is not possible, you cannot prove that `Free[F, A]` is equal to another `Free[F, A]`, at least for some arbitrary `F`.

Indeed, you should expect this, because in general, two programs cannot be compared for equality. Even more basic, two functions cannot (in general) be compared for equality, because to do so you have to perform a symbolic proof, which corresponds to traversing an infinite graph by an unbounded distance.

However, that said, for specific `F` you may be able to prove equality. In particular, an `F` which does not read any values. This corresponds to a program whose structure does not depend on the result of computation; in some sense you can think of it as a "write only" program. In the Haskell formulation of `Free`, you can actually just pattern match to perform the analysis.

Meanwhile, however, you can execute a program, log its series of operations, and compare that log with another log. This is equivalent to a unit test which compares that two programs "did the same things in the same order". Of course, it is *not* a prove that two programs are the same, only a proof that for some fixed inputs, two programs produced the same set of operations in the same order.

Hence, `TraceFS` records what a program does, as it's doing it (according to some other interpreter responsible for producing the values the program needs), while `Free` merely describes the program generically in terms of some `F`.

jedesah:
@jdegoes That makes sense. Thx for the explanation!

jdegoes:
No problem!

mossprescott:
Switching to just `Boolean` mostly simplified things a bit, but I did add a variant, `fileExistsM` that lifts the result into `FileSystemErrT`, because it was being used that way in a lot of places. This seems gross to me, but it's certainly not out of line with similar functions we have all over the place. 

Other than questions of code smell, it essentially didn't affect the APIs behavior, and all in all, I do think it's an improvement.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
```diff
@@ -161,9 +161,9 @@ private final class QueryFileInterpreter[C](
       }).liftM[QRT].liftM[WorkflowExecErrT]
 
     case FileExists(file) =>
-      collFromPathM(file)
-        .flatMap(MongoDbIO.collectionExists(_).liftM[FileSystemErrT])
-        .run.liftM[QRT].liftM[WorkflowExecErrT]
+      Collection.fromPathy(file).fold(
```
jedesah: I am still somewhat concerned with this behavior, i.e. simply returning `false` when there is something wrong with the path being provided. Either way, I think it's probably good that it's not part of the core algebra but it might make sense for the mongodb interpreter to report this as a failure in the monad stack rather than simply returning `false`. But we also already discussed this, so I can get on board with just returning `false`.

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
```diff
@@ -280,6 +280,24 @@ class MongoDbFileSystemSpec
         }
       }
 
+      "File exists" >> {
+        "for missing file at root (i.e. a database path) should succeed" >> {
+          val tfile = rootDir </> file("foo")
+
+          val p = query.fileExists(tfile)
+
+          run(p).run must_== false
```
jedesah: So, just to confirm one last time, if someone uses the `metadata` api to query such a file in a mongodb filesystem, we are content returning `false` instead of `"It is not possible to have a file at the top level of a MongoDB filesystem"`?

core/src/test/scala/quasar/fs/TraceFS.scala
```diff
@@ -50,7 +57,7 @@ object TraceFS {
           case FileExists(file)     =>
             ls(fileParent(file)).fold(
               κ(false),
-              _.contains(fileName(file).right)).right
+              _.contains(fileName(file).right))
```
jedesah: I wonder if we should just change the implementation of `ls` here. Kind of weird to be constructing a `pathNotFound` just to ignore it and I doubt it's being used elsewhere, but I could be wrong.

jedesah: Nevermind! We use it on line `56`. :facepunch:

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -341,30 +341,24 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
     "return true if there is a view at that path" ! prop { (file: AFile, lp: Fix[LogicalPlan]) =>
-      val program = query.fileExists(file).run
+      val program = query.fileExists(file)
 
-      val ops = traceInterp(program, Map())._1
-
-      val expected = (ops, true.right)
-
-      viewInterp(Views(Map(file -> lp)), Map(), program) must_== expected
+      viewInterp(Views(Map(file -> lp)), Map(), program) must_==((Vector(), true))
```
jedesah: I don't understand why we don't get the trace of `FileExists`?

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
mossprescott: That's right. And from the outsider's point of view, it's possible for a file to appear there, if it's a view.

core/src/test/scala/quasar/fs/mount/view.scala
mossprescott: When there's a view present, there's no longer any need to check the underlying FS.

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
jedesah: :+1:

core/src/test/scala/quasar/fs/mount/view.scala
jedesah: Ah, makes sense.

jedesah:
LGTM!

mossprescott:
Merged as 4.4.5 via the script.

PR #1129, opened 2016-03-11
SD-1263 Handle specific failing case of "plan multiple reducing projections (all, distinct, orderBy)"

drostron:
Thanks to @mossprescott for assistance on this.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
@@ -1549,7 +1549,7 @@ object WorkflowBuilder {
     }
   }
 
-  private def findSort(src: WorkflowBuilder)(distincting: WorkflowBuilder => WorkflowBuilder):
+  private def findSort(src: WorkflowBuilder)(distincting: WorkflowBuilder => M[WorkflowBuilder]):
       M[WorkflowBuilder] = {
     @tailrec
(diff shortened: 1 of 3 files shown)


core/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -1580,11 +1584,14 @@ object WorkflowBuilder {
...
-          val (fields, keys) = ks.toList.map(k =>
-            (k -> $first($var(DocField(k))).left[Expression],
-              Fix(normalize(ExprBuilderF(src, \/-($var(DocField(k)))))))
-          ).unzip
-          findSort(src)(GroupBuilder(_, keys, Doc(fields.toListMap)))
+          val keys = ks.toList.map(k => Fix(normalize(ExprBuilderF(src, \/-($var(DocField(k)))))))
+          findSort(src) { newSrc =>
+            val newSrcFields = findKeys(newSrc).foldMap {
+              case Subset(fields) => fields.toList.map(k => k -> $first($var(DocField(k))).left[Expression])
+              case _ => List()
```
sellout: Any more thought on whether we should error here?

drostron: Right, thanks for the reminder. An error for this case makes sense — will update accordingly.

sellout:
👍🏾

PR #1130, opened 2016-03-12
Update to scala 2.11.8 (SD-1514)

jedesah:
- Update version of `sbt-buildinfo` so that generated build info code compiles under our compiler flags
- Remove a whole bunch of unused imports that are correctly reported as unused
- Make pattern match exhaustive now that scala reports it as non-exhaustive


jedesah:
This PR is unfortunately blocked on https://github.com/sbt/sbt-buildinfo/pull/81 unless we temporarily disable `-Yno-imports` while we wait for a new version of `sbt-buildinfo` to be released.

core/src/main/scala/quasar/optimizer.scala
```diff
@@ -232,6 +232,8 @@ object Optimizer {
 
             case (LeftCond(h),    RightCond(cs))   => OtherCond((l, r) => h(l) :: cs(r))
             case (RightCond(h),   LeftCond(cs))    => OtherCond((l, r) => h(r) :: cs(l))
+
+            case _                                 => ???
```
jedesah: @sellout Tried to figure out what best to put here in order to make this pattern match total, but was unable to wrap my head around the required knowledge in a reasonable amount of time. Happy to have a look at it with you so next time I can figure out myself or to just write what you tell me to.

sellout: This is @mossprescott’s code, so he might have more to add, but it seems like we need to cover cases for OtherCond and EquiCond. Makes me wonder if there’s a monoid for `Component[A]` where `A` has a monoid. Then these cases get extracted to there and we have roughly `ts.map(_._2 ∘ (List(_))).fold` here.
Also, just as a side note, I tend to make “catchall” cases as specific as possible while still being total. In this case, that’d be `case (_, _)` rather than `case _`.

sellout: Oh, and the cases would basically be `(EquiCond, EquiCond) => EquiCond` and the others would `=> OtherCond`.

sellout:
I’d rather not disable `-Yno-imports`, but it would be nice to get in the fixes to our code that 2.11.8 identifies. Could we basically commit this _except_ for the actual Scala (and sbt-buildinfo) version change, until there’s a new sbt-buildinfo?

jedesah:
@sellout Luckily @eed3si9n was super responsive on publishing a new version of `sbt-buildinfo` with our changes so no need to separate the PR.

core/src/main/scala/quasar/optimizer.scala
jedesah: @mossprescott I think we may have forgotten the `(EquiCond, EquiCond) => EquiCond` case. Or did we consider it and I forgot?

mossprescott: That case would be something like `(l.x = r.y) != (l.z = r.t)`, so I think it's still an `OtherCond`.

jedesah: What about `(l.x = r.y) = (l.z = r.t)`? Is that non-sensical?

mossprescott: It's no _more_ non-sensical, and it's still `OtherCond`.

jedesah: Ah, okay, I think I get something that I did not fully grasp before. `EquiCond` represents a very special case that is useful for optimization, i.e `foo.id = bar.id`, but anything more involved even if it is only composed of equalities would still be `OtherCond`.

jedesah:
@sellout I think this is ready for review again.

core/src/main/scala/quasar/optimizer.scala
mossprescott: Exactly. Basically EquiCond only matches things that the planner knows how to convert to an efficient join, which is the pattern of (expr involving one source) = (expr. of the other source).

sellout:
This looks good, pending Travis.

PR #1134, opened 2016-03-16
SD-1517 – Make JS plan for join conditions optional.

sellout:
We only ever use the JS plan from either the left or right side of the
condition, this change allows one to fail, in which case we use the
other. Also handle the case where _neither_ has a JS plan.

### DIFF
--- a/core/src/main/scala/quasar/physical/mongodb/planner.scala
+++ b/core/src/main/scala/quasar/physical/mongodb/planner.scala
@@ -668,6 +668,11 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
           }
         }
 
+      def makeKeys(k: List[Ann], comp: Ann):
+          Option[List[JsFn]] =
+        k.traverse(HasJs).flatMap(k =>
+          findArgs(k, comp).map(applyPartials(k, _))).toOption
+
       func match {
         case MakeArray => lift(Arity1(HasWorkflow).map(makeArray))
         case MakeObject =>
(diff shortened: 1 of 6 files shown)


sellout:
I probably should have kept this as two separate commits in the PR (not counting the one that fixes the tests on RO dbs).

The first step was making JS plans optional in join conditions, since flattening doesn’t have a JS plan. That mostly affected the Mongo planner, but also the `(Option, Option)` pattern match in WorkflowBuilder. The second step was handling the case where _neither_ side has a JS plan, which required extracting `wbReduce` and `jsReduce` from the body of `WorkflowBuilder.join`.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -669,6 +669,11 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
           }
         }
 
+      def makeKeys(k: List[Ann], comp: Ann):
+          Option[List[JsFn]] =
+        k.traverse(HasJs).flatMap(k =>
+          findArgs(k, comp).map(applyPartials(k, _))).toOption
```
drostron: I take it `join` isn't interested in `PlannerError`s or is it somehow known there won't be any?

sellout: Yeah, basically, there are three ways to generate plans – Selectors, JS, and Workflows – and in various situations we require different ones to succeed. Normally we know which ones we need in the planner itself, but in the case of the join, we don’t.

In this case, WorkflowBuilder can take advantage of a JS plan if it exists (allowing us to fold the preparation of one side of the join into the mapReduce that handles joining the two sides), but if neither side of the condition has a JS plan, we can still succeed (therefore not caring about the PlannerError) by just preparing each side of the join in its own temp collection, then using the mapReduce to only perform the join of those two temp collections.

drostron: Ok — explanation appreciated.

drostron:
Opinion or background on not having test(s) for join in `WorkflowBuilderSpec`?

sellout:
Good point. I should add a test or two there (or somewhere near there). We really want LP => Workflow tests, but I think WorkflowBuilder is the closest we have at the moment.

sellout:
@drostron Ok, test added.

drostron:
👍🏾

PR #1135, opened 2016-03-16
SD-837 Constant folding for aggregation functions

wemrysi:
Implements constant folding for all of the functions in `AggLib`.

In order to get constant sets passed to these functions to typecheck, I modified type inference to infer a set-based type for an argument which itself was a constant set. I chose to do this during inference rather than typechecking as we perform the former top-down, which provides the necessary context to make this decision vs the bottom-up process of typechecking.

This is admittedly a bit smelly, but seemed to at least be contained since we only apply this transformation when the function being applied is a `Reduction`. I also tried just adding `Set` types to the domain of the reduction functions, but that had the unintended side-effect of generating lots of unecessary runtime typechecks for the additional types (which only exist to support constants), not to mention confusing the type signatures of those functions.

I've implemented the behavior as requested in the ticket, though after some discussion with @mossprescott I realize this behavior will likely be unexpected to those coming from SQL, hopefully the upcoming MRA spec will help to clarify this.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [SD-837] Constant folding for aggregation functions
--- a/core/src/main/scala/quasar/data.scala
+++ b/core/src/main/scala/quasar/data.scala
@@ -146,6 +146,45 @@ object Data {
     def toJs = jscore.ident(Js.Undefined.ident)
   }
 
+  final class Comparable private (val value: Data) extends scala.AnyVal
+
+  object Comparable {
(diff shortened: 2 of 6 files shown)


sellout:
This looks really nice, and there are a bunch of new tests. But somehow our coverage is still dropping!

However, 👍🏾

wemrysi:
@sellout The merge script realized that I didn't include a changelog, I added one for a revision bump as this feature was already partially implemented, just not complete. Sound good?

sellout:
Yep, definitely a bug fix for stuff that should have been working.

PR #1136, opened 2016-03-16
Allow constant-foldable expressions to be used as LIKE patterns.

sellout:
SD-1523 #done

This moves the conversion from LIKE => regex into the simplifier (where
it should be) and out of the compiler. Also, non-foldable expressions
are technically allowed now, but aren’t yet supported by the MongoDB
backend (we could implement a LIKE => regex converter in mapReduce to
add support).

There was also a small secondary bug in our handling of the generic `||`
operator, where constant strings would fail to typecheck. That is also fixed.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [SD-1523] allow expressions that fold to literals as `LIKE` patterns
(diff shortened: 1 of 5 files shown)


wemrysi:
LGTM :+1:

PR #1137, opened 2016-03-16
SD-1509 Readable JSON data codec should render NA values as null

wemrysi:
Instead of rendering `Data.NA` as `"NA"`, the "readable" `Data` codec now renders them as the more idiomatic JSON `null`.

SD-1509 #done

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,6 +502,7 @@ time      | `"10:30:05"`    | `{ "$time": "10:30:05" }` | HH:MM[:SS[:.SSS]]
 interval  | `"PT12H34M"`    | `{ "$interval": "P7DT12H34M" }` | Note: year/month not currently supported.
 binary    | `"TE1OTw=="`    | `{ "$binary": "TE1OTw==" }` | BASE64-encoded.
 object id | `"abc"`         | `{ "$oid": "abc" }` |
+error     | `null`          | `{ "$na": null }` | A runtime error occurred producing or representing this value.
 
 
 ### CSV
--- a/core/src/main/scala/quasar/repl/Repl.scala
+++ b/core/src/main/scala/quasar/repl/Repl.scala
@@ -231,7 +231,7 @@ object Repl {
     P: ConsoleIO.Ops[S]
   ): Free[S, Unit] = {
     def formatJson(codec: DataCodec)(data: Data) =
-      DataCodec.Precise.encode(data).fold(
+      codec.encode(data).fold(
         err => "error: " + err.shows,
         _.pretty(minspace))
 
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [SD-1509] Render NA values as JSON null in readable Data codec
--- a/core/src/main/scala/quasar/codec.scala
+++ b/core/src/main/scala/quasar/codec.scala
@@ -161,7 +161,7 @@ object DataCodec {
 
         case Id(value)        => \/-(jString(value))
 
-        case `NA`             => \/-(jString("NA"))
+        case `NA`             => \/-(jNull)
       }
     }
 
--- a/core/src/test/scala/quasar/codec.scala
+++ b/core/src/test/scala/quasar/codec.scala
@@ -150,7 +150,7 @@ class DataCodecSpecs extends Specification with ScalaCheck with DisjunctionMatch
       "encode set"       in { DataCodec.render(Data.Set(List(Data.Int(0), Data.Int(1), Data.Int(2)))) must beRightDisjunction("[ 0, 1, 2 ]") }
       "encode binary"    in { DataCodec.render(Data.Binary(Array[Byte](76, 77, 78, 79))) must beRightDisjunction("\"TE1OTw==\"") }
       "encode objectId"  in { DataCodec.render(Data.Id("abc")) must beRightDisjunction("\"abc\"") }
-      "encode NA"        in { DataCodec.render(Data.NA) must beRightDisjunction("\"NA\"") }
+      "encode NA"        in { DataCodec.render(Data.NA) must beRightDisjunction("null") }
     }
 
     "round-trip" ! prop { (data: Data) =>


jdegoes:
README needs updating.

wemrysi:
@jdegoes Thanks for the reminder.

On that note, given that this also affects the shape of JSON responses from the API in a non backward-compatible way, it would seem that this would require a major version bump. Agreed?

sellout:
@wemrysi Good question … do we distinguish between the readable format being informative, vs the normative precise format? If so, then I think it’s a revision. If not, then major.

jdegoes:
I don't think it warrants a major bump, since there's no way for client code to distinguish between an "NA" string and an "NA" value denoting the absence of a value; and any code that relied on the "NA" string literal would almost be broken by definition for that reason.

README.md
```diff
@@ -502,6 +502,7 @@ time      | `"10:30:05"`    | `{ "$time": "10:30:05" }` | HH:MM[:SS[:.SSS]]
 interval  | `"PT12H34M"`    | `{ "$interval": "P7DT12H34M" }` | Note: year/month not currently supported.
 binary    | `"TE1OTw=="`    | `{ "$binary": "TE1OTw==" }` | BASE64-encoded.
 object id | `"abc"`         | `{ "$oid": "abc" }` |
+undefined | `null`          | `{ "$na": null }` |
```
jedesah: What is the rationale behind `"$na"`? I was thinking maybe something like `{ "$other": undefined }`

jedesah: Actually, the data type in quasar is `NA`, so maybe it would be better to use "na" here in the type column. Maybe there is a good reason for undefined. Still think it might make sense to use something like `{ "$other": "na"}` or `{"$other": "undefined"}`

core/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -231,7 +231,7 @@ object Repl {
     P: ConsoleIO.Ops[S]
   ): Free[S, Unit] = {
     def formatJson(codec: DataCodec)(data: Data) =
-      DataCodec.Precise.encode(data).fold(
+      codec.encode(data).fold(
```
jedesah: Nice catch! Probably not worth writing a test that would expose this incorrect behavior.

wemrysi:
@jdegoes If a field was known to have a type other than `String` when present then the "NA" distinction would be observable, wouldn't it? Also, the readable format is the default used when no `Accept` header is received or if the client specifies `Accept: application/json`.

Note: I'm not necessarily trying to argue for a major bump, just trying to tease out the criteria for the decision (and future ones).

sellout:
@wemrysi On the other hand, `NA` wasn’t a documented part of the format until you just added that line. So it looks like a brand new feature (minor) to me!

jdegoes:
>  If a field was known to have a type other than String when present then the "NA" distinction would be observable, wouldn't it? 

But that's precisely the point: Quasar only ever exposes a heterogeneous model of data, in which some values may have the field and others not; and in which the type may change for every value. So there is no way to know that a field consistently exists, let alone that it has a singular type for all values.

Also, we have already documented that the readable format does not preserve type information and is merely designed to be more readable to a human than the strict format.

However, I don't care either way.

README.md
wemrysi: Good question, it's true the `Data` constructor is named `NA` but it is somewhat nebulously defined. I'd prefer to have the conceptual name in the `README` even if it differs from how it might be represented in code.

@jdegoes @sellout @mossprescott _Is_ the concept here "na"? Does that mean "not-applicable"? Or does "undefined" or maybe "unknown" better describe this?

From the comment in code it seems like `NA` is used for both missing fields as well as fields having a type unrepresentable with `Data`.

jdegoes: The concept is an error: undefineds are produce when an operation is applied to the wrong type of value.

jedesah: Where does this leave us as far as choosing a name for the "type" column as well as an appropriate "precise" representation?

wemrysi:
@sellout @jdegoes I'm on board with a revision bump, thanks for all the input!

README.md
```diff
@@ -502,6 +502,7 @@ time      | `"10:30:05"`    | `{ "$time": "10:30:05" }` | HH:MM[:SS[:.SSS]]
 interval  | `"PT12H34M"`    | `{ "$interval": "P7DT12H34M" }` | Note: year/month not currently supported.
 binary    | `"TE1OTw=="`    | `{ "$binary": "TE1OTw==" }` | BASE64-encoded.
 object id | `"abc"`         | `{ "$oid": "abc" }` |
+unsupported | `null`          | `{ "$na": null }` | Used for missing fields and fields having a type not supported by Quasar.
```
wemrysi: @jedesah What do you think of this version?

README.md
```diff
@@ -502,7 +502,7 @@ time      | `"10:30:05"`    | `{ "$time": "10:30:05" }` | HH:MM[:SS[:.SSS]]
 interval  | `"PT12H34M"`    | `{ "$interval": "P7DT12H34M" }` | Note: year/month not currently supported.
 binary    | `"TE1OTw=="`    | `{ "$binary": "TE1OTw==" }` | BASE64-encoded.
 object id | `"abc"`         | `{ "$oid": "abc" }` |
-undefined | `null`          | `{ "$na": null }` |
+unsupported | `null`          | `{ "$na": null }` | Used for missing fields and fields having a type not supported by Quasar.
```
jedesah: I like the added description and I think "unsupported" is probably a better term for the concept than "undefined", even though neither of them seems all that great. I still think the presence of `null` on the precise side is very odd since it is not `null` in any way as far as I can understand, that is simply the representation we give it in the readable format. I kind of see why `null` is a convenient value to put there, but still seems odd.

Ultimately though, I don't think it matters THAT much and I defer to the spec/stakeholders to decide what is desired.

wemrysi: I think the precise encoding of `NA` is just a way to represent a singleton/sentinel value in JSON that isn't a string and yet carries more information than `null`.

jedesah: Right, but an alternative way would be to use an "$other" escaped key to describe what type of other data it is using a `String` as value. But the current way is probably fine too and has the advantage of not requiring a change.

jdegoes:
In the readable format, `null` is indeed the best mapping. 

The terminology of `undefined` comes from Javascript, whereby `undefined` refers to a thing which does not have a definite value (for reason of being "undefined"). In MRA core, it's also called `Undefined`, although it could just as easily be called, `Error`, because there is no literal for `Undefined`, and the only way to generate them is by applying an operation to a value with the wrong structure.

The concept is somewhat co-opted to indicate values whose types are supported by the backend, but which do not have any coherent representation in Quasar, such as "Javascript with scope" and other absurd and deprecated types in BSON. This is a separate meaning, but does not warrant a separate data type in Quasar; backend-specific details should not pollute the Quasar data model, and the general meaning of "error" carries the right intention.

jedesah:
LGTM!

README.md
```diff
@@ -502,6 +502,7 @@ time      | `"10:30:05"`    | `{ "$time": "10:30:05" }` | HH:MM[:SS[:.SSS]]
 interval  | `"PT12H34M"`    | `{ "$interval": "P7DT12H34M" }` | Note: year/month not currently supported.
 binary    | `"TE1OTw=="`    | `{ "$binary": "TE1OTw==" }` | BASE64-encoded.
 object id | `"abc"`         | `{ "$oid": "abc" }` |
+error     | `null`          | `{ "$na": null }` | Used when a result could not be determined due to an input value having the wrong structure or when a field has a native type not representable by Quasar.
```
wemrysi: @jedesah Ok, I think this conveys what @jdegoes described, agree?

jdegoes: Not to obsess, but this is too strict, as the `undefined` value has nothing to do with fields.

I'd suggest something simpler and more general, such as:

```
error | `null` | `{ "$na": null }` | A runtime error occurred producing or representing this value.
```

wemrysi: Obsess away, accurate documentation is important! Thanks for the input, will correct.

wemrysi:
@jedesah Ok, I think we've arrived at a final version.

Thanks everyone for sweating the details with all your input and explanation!

PR #1138, opened 2016-03-16
preserve field order in Data.Obj

mossprescott:
This is just a simple switch to use ListMap in Data.Obj, as we do basically everywhere else in the code where we have maps of fields.

This fixes two planner tests which were marked with TODOs referring to this exact issue. Since it was so simple to make the switch, I'm not sure why we left those in and didn't just fix it.

I have not yet updated the regression tests. There are about 10 or so failures, some obviously for field order, but others also which I haven't looked at yet.

Any reason not to make this switch?

### DIFF
--- a/it/src/test/scala/quasar/ResultFileQueryRegressionSpec.scala
+++ b/it/src/test/scala/quasar/ResultFileQueryRegressionSpec.scala
@@ -21,10 +21,8 @@ import quasar.fs._
 import quasar.regression._
 import quasar.sql._
 
-import scalaz.{~>, Hoist}
+import scalaz._, Scalaz._
 import scalaz.stream.Process
-import scalaz.std.vector._
-import scalaz.syntax.monad._
 
 class ResultFileQueryRegressionSpec
   extends QueryRegressionTest[FileSystemIO](
@@ -45,7 +43,7 @@ class ResultFileQueryRegressionSpec
     for {
       tmpFile <- hoistM(manage.tempFile(DataDir)).liftM[Process]
       outFile <- query.executeQuery(expr, vars, tmpFile).liftM[Process]
-      cleanup =  hoistM(manage.delete(tmpFile))
+      cleanup =  hoistM(manage.delete(tmpFile)).whenM(outFile ≟ tmpFile)
       data    <- read.scanAll(outFile)
                    .translate(hoistM)
                    .onComplete(Process.eval_(cleanup))
--- a/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
+++ b/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
@@ -44,6 +44,9 @@ abstract class QueryRegressionTest[S[_]: Functor](
 
   import QueryRegressionTest._
 
+  // NB: forcing the examples to not overlap seems to avoid some errors in Travis
+  sequential
+
   type FsErr[A] = FileSystemErrT[F, A]
 
   val qfTransforms = QueryFile.Transforms[F]
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,3 @@
+- preserve order of fields when writing data to MongoDB
+- make the regression test matchers more robust and improve messages
+- don't "unwind" the constant in a case like `select "abc", sum(pop) from zips`
--- a/core/src/main/scala/quasar/compiler.scala
+++ b/core/src/main/scala/quasar/compiler.scala
@@ -232,7 +232,7 @@ trait Compiler[F[_]] {
       //       ObjectConcat with an empty map on one side, this could be done
       //       in a single foldLeft.
       fields match {
-        case Nil     => LogicalPlan.Constant(Data.Obj(Map()))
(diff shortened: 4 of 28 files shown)


jdegoes:
:+1:

sellout:
Looks sensible to me.

mossprescott:
Thanks. I'll take a look at the test later on and hopefully wrap this up.

it/src/main/resources/tests/constantAndGrouped.test
```diff
@@ -1,5 +1,6 @@
 {
   "name": "constant and a grouped value",
+  "backends": { "mongodb_2_6": "pending", "mongodb_3_0": "pending", "mongodb_read_only": "pending", "mongodb_3_2": "pending" },
```
mossprescott: This is a legitimate bad result that was not being caught by `containsExactly`. The result is 29k copies of the expected result, because we don't insert an `Arbitrary` reduction around a constant, but we probably should so I left the test as is and marked it pending.

I think we should go ahead and fix that, but perhaps not in this PR.

jdegoes: It will be fixed by @sellout's PR, because `Constantly` will no longer be wrapped around literals, therefore, there is no need to insert an arbitrary reduction around them.

mossprescott: Even better.

mossprescott:
Rebased so I could fix the new test @sellout added.

it/src/main/resources/tests/constantAndGrouped.test
sellout: The PR mentioned above (#1142) has been merged, so if you rebase, you can see if that fixed it.

mossprescott: Thanks, I'll rebase and unleash that test.

mossprescott: Hmm. Doesn't seem to be any fixed-er.

mossprescott:
@sellout: This is ready for a look, finally.

With the ControlService tests skipped, all four builds passed on the first try in Travis, and I just kicked it off again to make sure.

mossprescott:
@sellout: Built clean in Travis with that fix in `objectConcat`. Thanks!

sellout:
This looks great. Glad to finally fix all the issues plaguing integration tests. And happy to see some arbitraries becoming more arbitrary.

it/src/test/scala/quasar/ResultFileQueryRegressionSpec.scala
```diff
@@ -45,7 +43,7 @@ class ResultFileQueryRegressionSpec
     for {
       tmpFile <- hoistM(manage.tempFile(DataDir)).liftM[Process]
       outFile <- query.executeQuery(expr, vars, tmpFile).liftM[Process]
-      cleanup =  hoistM(manage.delete(tmpFile))
+      cleanup =  hoistM(manage.delete(tmpFile)).whenM(outFile ≟ tmpFile)
```
wemrysi: Why is this necessary?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -44,6 +44,9 @@ abstract class QueryRegressionTest[S[_]: Functor](
 
   import QueryRegressionTest._
 
+  // NB: forcing the examples to not overlap seems to avoid some errors in Travis
+  sequential
```
wemrysi: Any idea why this is, or which examples? Is this something new required as a result of other changes, or had the failures been happening all along?

it/src/test/scala/quasar/ResultFileQueryRegressionSpec.scala
sellout: Because sometimes the tmpFile never gets created.

wemrysi: Ah, right (for some reason I read that operator as "not equal to").

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
mossprescott: It's a bit mysterious, but when I changed the matchers to be more rigorous about verifying the results and capturing failures, I started seeing "137" errors in Travis. According to [this issue](https://github.com/travis-ci/travis-ci/issues/4192#issuecomment-119053081) @jr dug up, it may be related to using more memory. My best guess is that the tests, when running in parallel, are now overlapping more than they used to, and that's just too much for the under-powered Travis Docker instances in some way.

Another hypothesis would be that the matchers are no longer "memory safe" in that they somehow keep an unbounded number of values from the query result in memory until they're done reading. I convinced myself that wasn't the case, but another look wouldn't hurt.

In any case, I tried it several times and `sequential` _always_ made the error go away, and does not noticably slow them down.

sellout:
This is in ready – which I should have set it to last week, but just did this morning.

mossprescott:
Thanks!

I will revert the change to skip the ControlService tests, since I haven't noticed as many of those failures lately.

mossprescott:
Merged as 5.0.7.

PR #1139, opened 2016-03-17
fix splitting of typechecks in selectors

mossprescott:
Fixes SD-1508

Essentially, we now detect when typechecks and cond tests need to be performed before evaluating expressions and do those tests in $project/$simpleMap, rather than putting them off until $match.

A number of planner tests actually were expecting "bad" plans, which showed these delayed typechecks.

The planner's InputFinder is now a simple ADT which makes it possible to inspect its structure. This is a big help for debugging as well.

JavaScript type checks are now all factored out into a utility object in the mongodb package (one was somewhat bogusly in jscore, and the rest were partially duplicated in a couple of places).


core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -256,9 +267,9 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
                 case "year"         => \/-(x => Call(Select(x, "getFullYear"), Nil))
 
                 case _ => -\/(FuncApply(func, "valid time period", field))
-              }): PlannerError \/ (JsCore => JsCore)).map(x => source.bimap[PartialFunction[List[JsFn], JsFn], List[InputFinder[B]]](
+              }): PlannerError \/ (JsCore => JsCore)).map(x => Bitraverse[(?,?)].bimap[PartialFunction[List[JsFn], JsFn], List[InputFinder], PartialFunction[List[JsFn], JsFn], List[InputFinder]](source)(
```
sellout: Still the inexplicable implicit search fail, eh?

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -45,16 +45,26 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
...
+  object InputFinder {
+    case object Here extends InputFinder {
+      def apply[A](a: Cofree[LogicalPlan, A]): A =
+        a.head
+    }
+    final case class There(index: Int, next: InputFinder) extends InputFinder {
+      def apply[A](a: Cofree[LogicalPlan, A]): A =
+        next((Recursive[Cofree[?[_], A]].children(a).apply)(index))
+    }
+  }
```
sellout: So, I think this is a debate we haven’t fully had yet, but … we’ve been leaning toward only importing packages, not objects, and only putting things like implicits inside objects. More significant types get their own packages, but “helper” types like this, we just define the case classes at the same level as the trait.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -668,6 +680,30 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
...
         }
 
+      /** Check for any values used in a selector which reach into the
+        * "consequent" branches of Typecheck or Cond nodes, and which
+        * involve expressions that are not safe to evaluate before
+        * evaluating the typecheck/condition. Using such values in
+        * selectors causes runtime errors when the expression ends up
+        * in a $project or $simpleMap prior to $match.
+        */
+      def unsafe(ann: Cofree[LogicalPlan, OutputM[WorkflowBuilder]], f: InputFinder): Boolean = {
```
sellout: I would call this maybe `isUnsafe` or something, since I think of `unsafeX` as something that is itself unsafe to call. Like *cough* `unsafeTap` *cough*

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -683,8 +719,14 @@ object MongoDbPlanner extends Planner[Crystallized] with JsConversions {
...
-                HasSelector(a2).flatMap(s =>
-                  s._2.map(_(on)).sequence.map(filter(wf, _, s._1))) <+>
+                HasSelector(a2).flatMap { case (sel, ifs) =>
+                  if (!ifs.exists(unsafe(on, _)))
+                    ifs.map(_(on)).sequence.map(filter(wf, _, sel))
+                  else
+                    HasWorkflow(a2).map(wf2 => filter(wf, List(wf2), {
+                      case f :: Nil => Selector.Doc(f -> Selector.Eq(Bson.Bool(true)))
+                    }))
+                } <+>
```
sellout: Well, at least _this_ chunk of code didn’t get any prettier 😉

core/src/main/scala/quasar/physical/mongodb/JsConversions.scala
```diff
@@ -30,6 +30,40 @@ trait JsConversions {
...
+    jscore.BinOp(jscore.Instance, expr, jscore.ident("Date"))
+
+  def isBinary(expr: JsCore) =
+    jscore.BinOp(jscore.Instance, expr, jscore.ident("Binary"))
+
+  def isObjectId(expr: JsCore) =
+    jscore.BinOp(jscore.Instance, expr, jscore.ident("ObjectId"))
 }
 
 object JsConversions extends JsConversions
```
sellout: So, I know you didn’t create this in this PR, but this could easily be a package object.

sellout:
Minor comments. Overall, I really like both the fix and the other restructuring you did.

core/src/main/scala/quasar/physical/mongodb/planner.scala
mossprescott: Yup. I tried it a few different ways with no luck. Probably will just rewrite with `map { case (p, f)` to clean it up.

core/src/main/scala/quasar/physical/mongodb/planner.scala
wemrysi: I appreciate that this organization doesn't pollute the scope where `InputFinder` is defined with the  subtypes `Here` and `There`. Anyhow, sounds like something worth discussing as I think we have a good mix of both styles in the codebase.

sellout: Yes, but “a mix in the codebase” is not the preferred style. We need to choose one. Planner should have its own package, since we also want to split selector/js/workflow planners into their own files, too.

So, right now, `There`/`Here` would pollute `quasar.physical.mongo`, but when planner is modularized, it’ll be more restricted. More smaller packages is a worthy goal, I think.

wemrysi: > Yes, but “a mix in the codebase” is not the preferred style.

Agree wholeheartedly, "good" was not meant as qualitative, but quantitative.

core/src/main/scala/quasar/physical/mongodb/JsConversions.scala
mossprescott: `JsConversions` is not the right name for this thing, but I think we do need a home for JsCore constructions that are particular to MongoDB. What was really wrong before was having some of this code in the `jscore` package object.

What package object would this go in? `mongodb.javascript`?

core/src/main/scala/quasar/physical/mongodb/planner.scala
mossprescott: Uh huh. I'm open to suggestions as always.

core/src/main/scala/quasar/physical/mongodb/planner.scala
mossprescott: Yes, that's a good point. Even better would be a name that more clearly expresses the sense in which this is not ok.

core/src/main/scala/quasar/physical/mongodb/planner.scala
mossprescott: So, separate file, InputFinder.scala, and Here/There at package scope? I could put it in `mongodb.planner` to get the ball rolling. ;)

sellout: I would prefer to just rearrange all of that in a separate PR to keep the commits readable. I’m happy to do it myself once this one is merged.

core/src/main/scala/quasar/physical/mongodb/JsConversions.scala
sellout: Yeah, that sounds like a good name.

mossprescott:
Ready for another look.

sellout:
👍🏾

mossprescott:
Merged via the script.

PR #1141, opened 2016-03-17
Improve documentation of API surround `GET` method of `/data` resource

jedesah:


jedesah:
@beckyconning Could you have a look at this too to validate whether this would have been useful to you prior to asking about `disposition` in Slack?

jedesah:
I searched the web some to find usages of disposition in media types and have not found anything. It's kind of interesting, because the client is essentially saying: "can you please add a `Content-Disposition` header and give it this value, awesome, thx". It feels like the server could respond: "well, if you know what you want, why don't you just do it yourself?". Anyway, if it's practical for the frontend, that's fine, but it seems like we are doing something "out of the ordinary" here so I just wanted to call it out as a little odd.

README.md
```diff
@@ -384,10 +384,18 @@ Retrieves data from the specified path, formatted in JSON or CSV format. The `of
 {"id":6,"guid":"a2863ec1-9652-46d3-aa12-aa92308de055","isActive":false,"balance":"$1,621.67","picture":"http://placehold.it/32x32","age":34,"eyeColor":"blue","latitude":-83.908456,"longitude":67.190633}
 ```
 
-The output format can be selected using an `Accept` header as described above.
+The output format can be selected using an `Accept` header. Supported formats are `application/json`, `application/ldjson`/`application/x-ldjson` or `text/csv`. To supplement these standard media types, it is possible to supply additional information in the media type extensions as described below. For all acceptable formats, it is possible to specify a desired `Content-Disposition` by adding the "disposition" extensions to the `Accept` header field media range. The value of the disposition key should correspond to the the desired value to be found in the `Content-Disposition` header of the response.
```
drostron: Are people familiar enough with "disposition" extensions? I didn't end up with concrete info after a cursory search. An example and/or pointing to a reference doc might help.

wemrysi: Here's the [relevant RFC](http://tools.ietf.org/html/rfc6266).

README.md
```diff
@@ -384,10 +384,18 @@ Retrieves data from the specified path, formatted in JSON or CSV format. The `of
...
 
 Given a directory path (ending with a slash), produces a `zip` archive containing the contents of the named directory, database, etc. Each file in the archive is formatted as specified in the request query and/or `Accept` header.
 
+#### Csv format customization
+
+It is possible to use the `columnDelimiter`, `rowDelimiter` `quoteChar` and `escapeChar` extensions keys to customize the format of the csv.
```
mossprescott: There's a brief description and example of these parameters under "GET /query/...". We should probably explain in just one place and refer to it from the other.

README.md
```diff
@@ -384,10 +384,18 @@ Retrieves data from the specified path, formatted in JSON or CSV format. The `of
...
+
+#### Json format customization
+
+As an extension to any json media type, it is possible to supply a "mode". Acceptable values for the `mode` extension are `precise` and `readable`. By choosing the `precise` mode, the client will receive json that is fully unambiguous and can be re-uploaded to quasar at a later time without any loss of information. The `readable` format on the other hand is intended to be more amenable to human consumption but carries the risk of information loss if certain types of data are present in the data being requested. If no `mode` is supplied, `quasar` will default to the `readable` mode.
```
mossprescott: There's a quite lengthy description under "Data Formats" lower down. Should definitely refer to that here, and possibly just document `mode` at this point.

Also, the same applies to `GET /query`, so should have some mention there.

README.md
```diff
@@ -384,10 +384,18 @@ Retrieves data from the specified path, formatted in JSON or CSV format. The `of
 {"id":6,"guid":"a2863ec1-9652-46d3-aa12-aa92308de055","isActive":false,"balance":"$1,621.67","picture":"http://placehold.it/32x32","age":34,"eyeColor":"blue","latitude":-83.908456,"longitude":67.190633}
 ```
 
-The output format can be selected using an `Accept` header as described above.
+The output format can be selected using an `Accept` header. Supported formats are `application/json`, `application/ldjson`/`application/x-ldjson` or `text/csv`. To supplement these standard media types, it is possible to supply additional information in the media type extensions as described below. For all acceptable formats, it is possible to specify a desired `Content-Disposition` by adding the "disposition" extensions to the `Accept` header field media range. The value of the disposition key should correspond to the the desired value to be found in the `Content-Disposition` header of the response.
 
 Given a directory path (ending with a slash), produces a `zip` archive containing the contents of the named directory, database, etc. Each file in the archive is formatted as specified in the request query and/or `Accept` header.
 
+#### Csv format customization
```
mossprescott: CSV should be capitalized.

README.md
```diff
@@ -384,10 +384,18 @@ Retrieves data from the specified path, formatted in JSON or CSV format. The `of
...
+
+It is possible to use the `columnDelimiter`, `rowDelimiter` `quoteChar` and `escapeChar` extensions keys to customize the format of the csv.
+
+#### Json format customization
```
mossprescott: JSON should be capitalized.

README.md
sellout: And surrounded by ☠️

drostron:
👍🏽

PR #1142, opened 2016-03-18
Allow non-literals in SQL set literals

sellout:
SD-1516 #done

Also,

- SQL `IN` no longer (accidentally) works on arrays
- select now has MRA set semantics, rather than SQL row semantics
- eliminate `Set` from the type system

### DIFF
--- a/core/src/main/scala/quasar/std/set.scala
+++ b/core/src/main/scala/quasar/std/set.scala
@@ -29,14 +29,7 @@ trait SetLib extends Library {
   private def setTyper(f: Func.Typer): Func.Typer =
     ts => f(ts).map {
       case x @ Type.Const(Data.Set(_)) => x
-      case x @ Type.Set(_)             => x
-      case rez                         => rez // ⨿ Type.Set(rez)
-    }
-  def setUntyper(f: Type => ValidationNel[SemanticError, List[Type]]):
-      Func.Untyper =
-    untyper {
-      case Type.Set(t)             => f(t)
-      case t                       => f(t)
+      case rez                         => rez
     }
 
   val Take = Sifting("(LIMIT)", "Takes the first N elements from a set",
@@ -48,10 +41,9 @@ trait SetLib extends Library {
       case Type.Const(Data.Set(s)) :: Type.Const(Data.Int(n)) :: Nil
           if n.isValidInt =>
         Type.Const(Data.Set(s.take(n.intValue)))
-      case Type.Set(t) :: _ :: Nil => t
-      case t           :: _ :: Nil => t
+      case t :: _ :: Nil => t
     }),
-    setUntyper(t => success(t :: Type.Int :: Nil)))
+    untyper(t => success(t :: Type.Int :: Nil)))
 
   val Drop = Sifting("(OFFSET)", "Drops the first N elements from a set",
     Type.Top, Type.Top :: Type.Int :: Nil,
@@ -66,16 +58,15 @@ trait SetLib extends Library {
       case Type.Const(Data.Set(s)) :: Type.Const(Data.Int(n)) :: Nil
           if n.isValidInt =>
         Type.Const(Data.Set(s.drop(n.intValue)))
-      case Type.Set(t) :: _ :: Nil => t
-      case t           :: _ :: Nil => t
+      case t :: _ :: Nil => t
     }),
-    setUntyper(t => success(t :: Type.Int :: Nil)))
+    untyper(t => success(t :: Type.Int :: Nil)))
 
   val OrderBy = Sifting("ORDER BY", "Orders a set by the natural ordering of a projection on the set",
     Type.Top, Type.Top :: Type.Top :: Type.Top :: Nil,
     noSimplification,
     setTyper(partialTyper { case set :: _ :: _ :: Nil => set }),
-    setUntyper(t => success(t :: Type.Top :: Type.Top :: Nil)))
+    untyper(t => success(t :: Type.Top :: Type.Top :: Nil)))
 
   val Filter = Sifting("WHERE", "Filters a set to include only elements where a projection is true",
     Type.Top, Type.Top :: Type.Bool :: Nil,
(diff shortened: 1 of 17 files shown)


core/src/main/scala/quasar/std/set.scala
```diff
@@ -253,12 +244,9 @@ trait SetLib extends Library {
     partialTyper {
```
jedesah: I am confused here (sorry this part of the codebase always confuses me a little). The `in` function returns a value of type `Type.Bool`, why is the `partialTyper` producing a type that is not boolean?

jedesah: As @sellout mentioned offline, just a silly diff illusion! :smile:

core/src/main/scala/quasar/std/set.scala
```diff
@@ -253,12 +244,9 @@ trait SetLib extends Library {
...
       case Type.Const(const) :: Type.Const(Data.Set(s)) :: Nil =>
         Type.Const(Data.Set(s.map(κ(const))))
-      case const :: Type.Const(Data.Set(s)) :: Nil => Type.Set(const)
-      case const :: Type.Set(_)             :: Nil => Type.Set(const)
```
jedesah: This is me being confused, but what about if the second argument is a `Type.Array`?

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -241,18 +241,21 @@ trait StructuralLib extends Library {
...
 
   val ShiftMap = Expansion(
     "SHIFT_MAP",
     "Zooms in on the values of a map, adding the keys as a new dimension",
-    Set(Top), AnyObject :: Nil,
-    noSimplification,
+    Top, AnyObject :: Nil,
+    new Func.Simplifier {
+      def apply[T[_[_]]: Recursive: Corecursive](orig: LogicalPlan[T[LogicalPlan]]) = orig match {
+        case IsInvoke(_, List(UnshiftMap(List(set)))) => set.project.some
```
jedesah: This makes sense, is it me, or could this have been like this before? (trying to better understand this code! :smile: )

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -285,18 +294,21 @@ trait StructuralLib extends Library {
...
+    Int, AnyArray :: Nil,
     noSimplification,
     partialTyper { case List(x) if x.arrayLike => Int },
-    partialUntyper {
-      case Set(Int) | Int => List(FlexArr(0, None, Top))
-    })
+    partialUntyper { case Int => List(FlexArr(0, None, Top)) })
 
-  val UnshiftMap = Reduction(
+  val UnshiftMap: Func = Reduction(
```
jedesah: Scala being uncooperative?

core/src/main/scala/quasar/std/structural.scala
```diff
@@ -304,15 +316,19 @@ trait StructuralLib extends Library {
...
-  val UnshiftArray = Reduction(
+  val UnshiftArray: Func = Reduction(
     "[...]",
     "Unshifts an integral dimension from the set identity, creating an array with the dimensional values as the indices.",
-    AnyArray, Set(Top) :: Nil,
-    noSimplification,
+    AnyArray, Top :: Nil,
+    new Func.Simplifier {
+      def apply[T[_[_]]: Recursive: Corecursive](orig: LogicalPlan[T[LogicalPlan]]) = orig match {
+        case IsInvoke(_, List(ShiftArray(List(array)))) => array.project.some
```
jedesah: Random thought, I wonder if it would be possible to declare a function as being the dual of another and implementing an AST traversal that would simplify all function invocations where the argument is an invocation of the dual of the function being invoked. Would avoid a little bit of code duplication.

core/src/main/scala/quasar/types.scala
```diff
@@ -384,8 +373,6 @@ final case object Type extends TypeInstances {
   final case object Interval          extends Type
   final case object Id                extends Type
 
-  final case class Set(value: Type) extends Type
```
jedesah: RIP `Type.Set` :skull:

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -114,11 +114,7 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
 
     "plan simple constant from collection" in {
       plan("select 1 from zips") must
-        beWorkflow(chain(
-          $read(Collection("db", "zips")),
```
jedesah: I like this!

core/src/main/scala/quasar/std/set.scala
sellout: An array is still just a single value. A set isn’t actually a value, but rather something outside of the type system (eventually Set will disappear from `Data`, too, but we’re not there yet).

E.g., in `select 4 + pop, length(city), array_length([1, 2, 3]) from zips`, `pop` and `city` are each a set of values from zips. `4 +` and `length` are mapped over those sets implicitly, but `array_length` is not implicitly mapped over the value `[1, 2, 3]`. In fact, since we don’t have HOF, the only way to map over something is to convert it to a set (eg, via `[_]`), and then apply the operation, then convert it back. 

It is still a bit confusing since we do have Set in Data, but hopefully that will disappear in the near future, too (although, I don’t know what approach we should take. One option is using `Set :+: Data` for the parts of the code that need to be aware of it. Another is to “suspend” literal sets in something like `ShiftArray(Constant(Data.Arr(List(1, 2, 3))))`, etc.)

core/src/main/scala/quasar/std/structural.scala
sellout: Yeah, it sees the Simplifier’s pattern matching as being a recursive call, so I have to annotate these. But I don’t know why it doesn’t similarly complain about ShiftArray/Map.

core/src/main/scala/quasar/std/set.scala
jedesah: Makes sense.

core/src/main/scala/quasar/std/structural.scala
sellout: Yep, this totally could have been there. Just that now we run into some of these identities (not this one).

core/src/main/scala/quasar/std/set.scala
jdegoes: >  Another is to “suspend” literal sets in something like ShiftArray(Constant(Data.Arr(List(1, 2, 3)))), etc.)

:+1:

jedesah:
LGTM!

PR #1143, opened 2016-03-22
SD-1339 Support view mount operations in Data API

drostron:


README.md
```diff
@@ -396,8 +396,6 @@ unchanged.
 
 If an error occurs when reading data from the request body, the response contains a summary in the common `error` field, and a separate array of error messages about specific values under `details`.
 
-Fails if the path identifies a view.
```
mossprescott: I think we need to document what _does_ happen now if the path is a view (and similar for each of the next three endpoints).

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -87,9 +89,9 @@ object view {
 
   /** Intercept and fail any write to a view path; all others are passed untouched. */
```
mossprescott: This comment needs updating.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -115,29 +118,46 @@ object view {
...
+            MoveScenario.fileToFile.getOption(scenario).>>= { case (src, dst) =>
+              views.map.collectFirst { case (loc, lp) if src === loc =>
+                val dstExists = views.map.contains(dst)
+                semantics match {
+                  case FailIfExists if dstExists =>
+                    emit[S, A](-\/(pathError(pathExists(dst))))
+                  case FailIfMissing if !dstExists =>
+                    emit[S, A](-\/(pathError(pathNotFound(dst))))
+                  case _ =>
+                    (ViewMounter.unmount[S](loc) >>= κ(ViewMounter.mount[S](dst, lp))) ∘ (_.right[FileSystemError])
```
mossprescott: This won't work if the view contains any relative paths and `dst` is in a different directory. In that case I suppose we would have to rewrite to an absolute path, or else fail.

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -162,46 +187,59 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
 
       val f = write.unsafe.open(p).run
 
-      viewInterp(views, Map(), f) must_==(
-        (Vector.empty,
-          -\/(FileSystemError.pathError(PathError2.invalidPath(p, "cannot write to view")))))
+      viewInterp(views, Map(), f) must_== ViewInterpResult(
+        traceInterp(f, Map())._1,
+        VS.empty,
+        \/-(WriteFile.WriteHandle(p, 0)))
```
mossprescott: OK, but what happens if you actually write to the handle?

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -162,46 +187,59 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
-          -\/(FileSystemError.pathError(PathError2.invalidPath(viewPath, "cannot move file to view location")))))
-    }
+    "fail when destination view exists and semantic is FailIfExists" in
+      moveShouldFail(Views(Map(srcPath -> q, dstPath -> q)), FailIfExists, PathError2.pathExists(dstPath))
+
+    "succeed when destination view exists and semantic is FailIfMissing" in
+      moveShouldSucceed(Views(Map(srcPath -> q, dstPath -> q)), FailIfMissing)
+
+    "fail when destination view doesn't exist and semantic is FailIfMissing" in
+      moveShouldFail(Views(Map(srcPath -> q)), FailIfMissing, PathError2.pathNotFound(dstPath))
```
mossprescott: What about moving views to paths where actual files are present? Any special behavior there?

README.md
drostron: Agreed.

core/src/main/scala/quasar/fs/mount/view.scala
mossprescott: Hmm. But a common case might be relative paths referring to files/views in the same notebook (i.e. other results), in which case the relative path should be preserved. How would we know?

mossprescott:
Does this PR actually fix the problem SD-1339 is getting at, which is that moving or deleting a notebook leaves behind any views mounted within it? I suppose maybe it does, if that behavior happens in ManageFile.Ops, but I think it deserves a test.

Aside from that, this whole concept isn't really sitting well with me. I understand the desire to make it simple to work with views, but conflating them with ordinary files in this way will create many surprising behaviors. Off the top of my head:
- appending to a path can now result in data disappearing. I can't really imagine a scenario in which replacing a view with some data is what you actually want to do.
- a `POST` or `PUT` to `/data` can now result in a `GET /mount` on the same path changing from 200 to 404.
- moving a file will work even across mounts, but if the original mount is later changed the moved file will stop working.

On top of that, it's now less clear when the `/mount` endpoint should be used. Should we document the cases in which `DELETE /mount` is equivalent to `DELETE /data` and so on? Do we need to provide advice about which API should be used?

core/src/test/scala/quasar/fs/mount/view.scala
drostron: That could be a useful addition to this test.

core/src/main/scala/quasar/fs/mount/view.scala
drostron: Good point. This case deserves more attention and a test or two.

core/src/test/scala/quasar/fs/mount/view.scala
drostron: Worthy of a few additional test cases.

drostron:
@mossprescott, the questions you bring up are appreciated. My experience with quasar's filesystem and views aren't sufficient to provide informed answers around previous and potential design choices. Hopefully others will offer some background and opinion.

jdegoes:
As I recall, @jedesah and I had a discussion about this (which we probably should have done on the ticket...oops, sorry!), and he convinced me the data API was sufficient to manage the view API given flexibility afforded by the permissions system.

There are many reasons, but the main argument is consistency: views are *exposed* as files in the data API, in that you can use the data API to read their contents, set skip and limit, etc; therefore it is natural if the other file system operations work on views, to the maximum extent possible.

However, these are all good concerns:

> appending to a path can now result in data disappearing. I can't really imagine a scenario in which replacing a view with some data is what you actually want to do.

Perhaps this should be an error, instead?

> a POST or PUT to /data can now result in a GET /mount on the same path changing from 200 to 404.

Seems like if we don't allow this (see above), we can prevent this scenario.

> moving a file will work even across mounts, but if the original mount is later changed the moved file will stop working.

Could be an argument for updating views that reference a mount when that mount itself is changed.

drostron:
@jdegoes, thanks for background and input.

Based on feedback I'll place this In Progress and:

- Omit `PUT` and `POST` from handling views (_at least for the moment, until we have a compelling use case_)
- Add test(s) representing the required api calls for moving and deleting a notebook
- Investigate and potentially add view updates upon file moves across mounts
- Address other comments

README.md
```diff
@@ -157,7 +157,7 @@ For example, given the above MongoDB mount, an additional view could be defined
     ...,
     "/simpleZips": {
       "view": {
-        "connectionUri": "sql2:///?q=select%20_id%20as%20zip%2C%20city%2C%20state%20from%20%22%2Flocal%2Ftest%2Fzips%22%20where%20pop%20%3C%20%3Acutoff&var.cutoff=1000"
```
jedesah: Hmm... I saw this before, but what is this change about again?

drostron: Update per recentish change in quoting style, https://github.com/quasar-analytics/quasar/pull/1093.

README.md
jedesah: :+1:

core/src/main/scala/quasar/fs/mount/view.scala
jedesah: Consider using `|@|` instead of `>>=` since we are using `κ`. I don't know what `∘` means but I guess I could read up on it.

jdegoes: `*>` if you don't use the LHS.

drostron: Oh right! I keep forgetting about those handy discarding apply ops.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -115,29 +118,46 @@ object view {
...
-            emit[S,A](-\/(pathError(invalidPath(path, "cannot delete view"))))
-          else
-            manage.delete(path).run
+          ViewMounter.views[S] >>= { views =>
+            val delete = manage.delete(path).run
+            refineType(path).fold(
+              d => views.map.keySet
+                .foldMap(f => f.relativeTo(d).map(κ(f)).toList)
+                .foldMap(ViewMounter.unmount[S])
+                .>>=(κ(delete)),
```
jedesah: idem.

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -53,17 +56,39 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
       KeyValueStore.toState[TraceS](_handles),
       MonotonicSeq.toState[TraceS](_seq),
       liftMT[Trace, VSF] compose
         interpretFileSystem[Trace](qfTrace(paths), rfTrace, wfTrace, mfTrace))
 
-  def viewInterp[A](views: Views, paths: Map[ADir, Set[PathName]], t: Free[FileSystem, A]): (Vector[RenderedTree], A) =
-    (t flatMapSuspension view.fileSystem[ViewFileSystem](views))
-      .foldMap(traceViewFs(paths))
-      .eval(VS(0, Map.empty)).run
+  case class ViewInterpResult[A](renderedTrees: Vector[RenderedTree], vs: VS, result: A)
```
jedesah: :+1:

core/src/main/scala/quasar/fs/mount/view.scala
wemrysi: Unless we're sure we know what the correct behavior would be when modifying a view during a move, I'd say we leave them alone. We may end up frustrating users more than helping them (i.e. if I have a view with relative paths and move  or copy it to a destination with the same structure as the source, I might not want my view modified at all).

jdegoes: Good point.

drostron: So, does that mean we're just moving views that contain only absolute paths? If so, is it worth having move at all?

wemrysi: I think we could move any views, how they will behave at their new location is dependent on the view itself, i.e. it would be possible to move a view to a new location and it wouldn't "work" immediately, but might eventually (if the necessary mounts were added/removed, etc). This isn't any different from how views behave in general, though (as a path referred to by a view may disappear or appear at any time).

drostron: Sounds good, I will leave the behavior as is.

wemrysi:
@jdegoes 

> There are many reasons, but the main argument is consistency: views are exposed as files in the data API

True, but there is still the inconsistency of not being able to create a view via the filesystem api. Any thoughts on having views be an "application" built on top of the filesystem rather than a part of the filesystem itself? i.e. a view would be a file with a particular extension containing the metadata representing the view. So creating a view would literally be creating a file having a particular format (similar to how notebooks are handled), this file could be operated on like any other file and would just be interpreted differently when read.

To be sure, this encoding comes with its own set of tradeoffs, for example not being able to edit the view itself as one cannot read the "metatdata" therein (though one can't really do that now unless you manually edit the config file). You also wouldn't be able to create a view if the only filesystem that exists doesn't support writing files (though we might be able to solve this by mounting the "metadata store" as a "system" filesystem that could be used in cases like this). I'm sure we can come up with others.

I'm not sure if I think this encoding is necessarily better, I just started thinking about it myself. Would appreciate in others' opinions (or pointers to where this idea was previously rejected if it was already discussed).

> the data API was sufficient to manage the view API given flexibility afforded by the permissions system

Are we conflating something between quasar and quasar-advanced (as only the latter has a permissions system)? Or is this just a way of saying that this change here won't create something insurmountable in quasar-advanced?

core/src/main/scala/quasar/fs/mount/view.scala
mossprescott: Yes, that's a good point, and I think it's probably best to move them as is and let the errors happen when they're used.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
```diff
@@ -124,19 +126,18 @@ final class EvaluatorMounter[F[_], S[_]: Functor](
   private def updateComposite[T[_]: Functor]
```
wemrysi: This function was always a bit awkward and now that the view interpreter reads the current set of views as an effect, I think it is even less compelling. I propose dropping this compositing of interpreters here alltogether an instead move it to the "final" interpreter assembled in `main.CompFsEff`. Instead of trying to precompose an interpreter when the mounts are changed, we'll just compose the interpreters when a request is made, which I think is more straightforward. Thoughts?

mossprescott:
I'll just throw one more thing in here because it's been on my mind and I haven't had a place to capture it. 

What is the correct way, at the HTTP level, to handle the errors that happen when reading from a view that has become invalid?

Currently, you can get a 404 Not Found if you try to read from a view where the view exists but refers to a nonexistent collection. That isn't right, since the path you supplied is correct. The true error is either:
1. The view is incorrect, which is a configuration problem, so maybe some kind of 5xx?
2. The view is "right" but the referenced collection has gone missing for some reason. This is more of a "bad state" problem, but here the problem is not actually in quasar, but in the mounted DB.

A runtime error in a read on a view should probably be no different from a runtime error while reading a file, but I'm not sure what the right status is, other than 500.

jdegoes:
>  i.e. a view would be a file with a particular extension containing the metadata representing the view. So creating a view would literally be creating a file having a particular format (similar to how notebooks are handled), this file could be operated on like any other file and would just be interpreted differently when read.

This is a very interesting idea, just needs fleshing out. Beyond API unification, would allow us, possibly, to use one security domain for mounts and views, with some additional work there.

Let me think about that, and also about the related issues (one needs to be able to change the mount, and read from it as well).

mossprescott:
A little historical perspective that may be of interest: "Views as metadata" was actually what we had in mind for views a long time ago.

The idea was to have a place to store metadata about arbitrary collections, for example when creating a result collection has the data nested under `value` (due to map-reduce) we would record that in the metadata so we could strip it out when reading. And the thought was that a query like `select * from foo` would _only_ write metadata with a reference to the source to avoid actually copying it (this was before we had true streaming evaluation, or course.) But this idea morphed into views as a type of mount by the time we got around to implementing it.

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
drostron: Sounds reasonable. I'll give it a go.

wemrysi:
> Let me think about that, and also about the related issues (one needs to be able to change the mount, and read from it as well).

One idea that came to mind, especially if we're going to use an extension, would be to just copy the file to a non-view name, edit it and then replace the original, i.e.:
1. Assume some view exists, `foo.view`
2. Copy as `foo.view.edit`
3. Read the contents and modify as normal
4. MOVE `foo.view.edit` -> `foo.view`

core/src/main/scala/quasar/fs/mount/EvaluatorMounter.scala
drostron: I'm going to defer this change until later, due to current time constraints.

drostron:
@jedesah this should be ready for another look.

core/src/main/scala/quasar/effect/KeyValueStore.scala
```diff
@@ -86,6 +90,14 @@ object KeyValueStore {
     def get(k: K): OptionT[F, V] =
       OptionT(lift(Get[K, V](k)))
 
+    /** Returns current keys */
+    val keys: F[Vector[K]] =
+      lift(Keys[K, V]())
+
+    /** Moves/renames key */
+    def move(src: K, dst: K): F[Unit] =
+      get(src).flatMapF(delete(src) *> put(dst, _).map(Some(_))).run.as(())
```
jedesah: I don't think the `map` here is necessary.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -87,9 +97,9 @@ object view {
 
   /** Intercept and fail any write to a view path; all others are passed untouched. */
   def writeFile[S[_]: Functor]
-      (views: Views)
       (implicit
-        S0: WriteFileF :<: S
```
jedesah: Technically, does `writeFile` really need `MountConfigsF`?

core/src/main/scala/quasar/fs/mount/view.scala
jedesah: Is this comment no longer valid because we changed something, because at this point it does not seem like the comment needs to change.

core/src/main/scala/quasar/fs/mount/view.scala
jedesah: nvm. I see that it probably does.

core/src/main/scala/quasar/fs/mount/view.scala
drostron: Right, we decided to postpone handling of views via PUT and POST of the data API and therefore the `writeFile` behavior and comment didn't need to change.

core/src/test/scala/quasar/fs/mount/ViewMounterSpec.scala
```diff
@@ -61,18 +76,203 @@ class ViewMounterSpec extends mutable.Specification {
...
+    }
+  }
+
+  "lookup" >> {
+    "trivial read with relative path" >> {
+      val f = rootDir[Sandboxed] </> dir("foo") </> file("justZips")
+      val vc = viewConfig("select * from zips")
+      val vs = Map[APath, MountConfig](f -> MountConfig.viewConfig(vc))
+
+      val r = eval(vs)(ViewMounter.lookup[MountConfigsF](f).run)
```
jedesah: `r` doesn't strike me as very useful here 😛

drostron: Haha, true!

core/src/main/scala/quasar/effect/KeyValueStore.scala
drostron: Ah, yes. Looks like `.as(())` could be `.void` as well.

core/src/test/scala/quasar/fs/mount/ViewMounterSpec.scala
```diff
@@ -61,18 +76,203 @@ class ViewMounterSpec extends mutable.Specification {
...
+        Fix(Take(
+          Fix(Drop(
+            Read(p),
+            Constant(Data.Int(5)))),
+          Constant(Data.Int(10))))
+
+      val innerLP = (quasar.queryPlan _).tupled(inner).run.run._2.toOption.get
+
+      val vs = Map[APath, MountConfig](
+        (rootDir </> dir("view") </> file("simpleZips")) -> MountConfig.viewConfig(inner))
```
jedesah: Might as well use `p` here as well

core/src/test/scala/quasar/fs/mount/ViewMounterSpec.scala
```diff
@@ -61,18 +76,203 @@ class ViewMounterSpec extends mutable.Specification {
...
+      val exp = Fix(InnerJoin(
+        Fix(Squash(Read(zp))),
+        Fix(Squash(Read(zp))),
+        Constant(Data.Bool(true))))
+
+      eval(vs)(ViewMounter.rewrite[MountConfigsF](q).run)
+        ._2 must beRightDisjunction.like { case r => r must beTree(exp) }
+    }
+
+    "self reference" >> {
```
jedesah: It's kind of weird that a view can co-exist with a file. I can see how it sort of works but it seems surprising nonetheless. I think with this change it's even more bizarre because if you try to move or a delete a "file" through the Data API, are you referring to the view or the data file at the specified location if there is both a traditional file and a view.

core/src/test/scala/quasar/fs/mount/ViewMounterSpec.scala
```diff
@@ -61,18 +76,203 @@ class ViewMounterSpec extends mutable.Specification {
...
+        ._2 must beRightDisjunction.like { case r => r must beTree(qlp) }
+    }
+
+    "circular reference" >> {
+      // NB: this situation probably results from user error, but since this is
+      // now the _only_ way the view definitions can be ill-formed, it seems
+      // like a shame to introduce `\/` just to handle this case. Instead,
+      // the inner reference is treated the same way as self-references, and
+      // left un-expanded. That means the user will see an error when the query
+      // is evaluated and there turns out to be no actual file called "view2".
```
jedesah: To the point I made above, what if there is a file and the circular reference then because I two way decorator? Is that what the user would expect?

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -119,123 +180,174 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
 
-    "fail with view destination path" in {
-      val viewPath = rootDir </> dir("view") </> file("simpleZips")
-      val otherPath = rootDir </> dir("foo") </> file("bar")
-      val q = Read(rootDir </> file("zips"))
+    "succeed when destination view exists and semantic is Overwrite" in
+      moveShouldSucceed(Map(srcPath -> expr, dstPath -> expr), Nil, Overwrite)
+
+    "succeed when destination file exists and semantic is Overwrite" in
+      moveShouldSucceed(Map(srcPath -> expr), Nil, Overwrite)
```
jedesah: I think `Nil` is a mistake here.

drostron: Perhaps it reads in a misleading way? The params to `moveShouldSucceed` specify the starting state for a test (_`views` and `files`_). `moveShouldSucceed` verifies that a view is moved from `srcPath` to `dstPath`.

core/src/test/scala/quasar/fs/mount/ViewMounterSpec.scala
drostron: As for the co-existence that seems to be a contentious point of this change and was discussed a bit above. From my understanding we decided to move forward with this, but if there is strong reason not to we should revisit. With this change it seems that we shouldn't allow a view and file at the same location. I need to look and see if this is possible. I imagine it is since you bring it up.

core/src/test/scala/quasar/fs/mount/view.scala
jedesah: It describes the file as existing, yet `Nil` tells me there is no file. Another way of looking at it is how come the test below has the exact same implementation and yet the description is different?

drostron: Ah, haha, the description 😄 . Yeah, `Nil` needs to be replaced with a starting state that includes a file.

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -119,123 +180,174 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
+      moveShouldSucceed(Map(srcPath -> expr), List(dstPath), FailIfMissing)
 
-      viewInterp(views, Map(), f) must_==(
-        (Vector.empty,
-          -\/(FileSystemError.pathErr(PathError.invalidPath(viewPath, "cannot move file to view location")))))
+    "fail when destination doesn't exist and semantic is FailIfMissing" in
+      moveShouldFail(Map(srcPath -> expr), Nil, FailIfMissing, PathError.pathNotFound(dstPath))
+
+    "move view and file subpaths" in {
+      val sd = rootDir[Sandboxed] </> dir("view")
```
jedesah: These names make sense once you understand the test, but they don't really help you understand it! 😄

drostron: Righto, I'm open to suggestions if you have any.

jedesah: I would be okay with `srcDir`, `destDir`, `viewFile`, `dataFile` in the interest of keeping them short yet meaningful.

drostron: Sounds good

drostron:
@jr, when you're done with the first pass I'll push a rebased version along with commit(s) containing updates per review.

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -119,123 +180,174 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
+      val vp = rootDir[Sandboxed] </> dir("view")
+      val p = vp </> file("simpleZips")
+      val expr = parseExpr("select * from zips")
+
+      val views = Map(p -> expr)
+
+      val f = manage.delete(vp).run
+
+      viewInterpTrace(views, Map(), f) must_== ViewInterpResultTrace(
+        traceInterp(f, Map())._1,
```
jedesah: This might be me not fully getting my head around this stuff, but why do we use `Vector.empty` in the test above and `traceInterp(f, Map())._1`. If I understand correctly, I see why it ends up being the same (at least for now), but semantically it seems different and I don't see the rationale for having one way above and another way down here.

jedesah:
sounds good. I will let you know, getting close to the end I think.

core/src/test/scala/quasar/fs/mount/view.scala
```diff
@@ -347,18 +468,25 @@ class ViewFSSpec extends Specification with ScalaCheck with TreeMatchers {
...
       }
       val noFile = {
-        viewInterp(Views.empty, Map(), program) must_==((ops, false))
+        viewInterpTrace(Map(), Map(), program) must_== ViewInterpResultTrace(ops, VS.empty, false)
       }
       hasFile and noFile
     }
 
-    "return true if there is a view at that path" ! prop { (file: AFile, lp: Fix[LogicalPlan]) =>
+    "return true if there is a view at that path" ! prop { file: AFile =>
```
jedesah: Why not keep the `LogicalPlan` abstract?

drostron: A relic from the first pass at getting things working. I'll switch it back.

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -43,10 +44,17 @@ object mount {
...
         respond(requiredHeader(Destination, req).map(_.value).fold(
-          err => EitherT.leftU[String](err.point[Free[S, ?]]),
-          dst => refineType(src).fold(
-            srcDir  => move[S, Dir](srcDir,  dst, parseAbsDir,  "directory"),
-            srcFile => move[S, File](srcFile, dst, parseAbsFile, "file"))).run)
+            err => EitherT.leftU[String](err.point[Free[S, ?]]),
+            dst => EitherT[Free[S, ?], ApiError, String] {
+              OptionT[Free[S, ?], (AFile, AFile)](
+                (refineType(src).toOption |@| UriPathCodec.parseAbsFile(dst).map(sandboxAbs)).tupled.point[Free[S, ?]])
+                .flatMap { case sd @ (s, _) => ViewMounter.lookup[S](s).map(κ(sd)) }
```
jedesah: I think `.map(κ(sd))` can be replaced with `as(sd)` just to better signal intent.

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -43,10 +44,17 @@ object mount {
 
       case req @ MOVE -> AsPath(src) =>
         respond(requiredHeader(Destination, req).map(_.value).fold(
-          err => EitherT.leftU[String](err.point[Free[S, ?]]),
-          dst => refineType(src).fold(
-            srcDir  => move[S, Dir](srcDir,  dst, parseAbsDir,  "directory"),
-            srcFile => move[S, File](srcFile, dst, parseAbsFile, "file"))).run)
+            err => EitherT.leftU[String](err.point[Free[S, ?]]),
+            dst => EitherT[Free[S, ?], ApiError, String] {
```
jedesah: I think this whole block could be improved. I think it would be clearer if the view case was handled in the `file` case of the refinement fold over `src`.

core/src/test/scala/quasar/fs/mount/view.scala
drostron: I spoke too soon. Views are now only stored as a `ViewConfig` which refers to an `Expr` rather than a `LogicalPlan`. I'll look at making `expr` arbitrary here.

web/src/test/scala/quasar/server/ServiceSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+      val webConfig = WebConfig.mountings.set(
+        MountingsConfig(Map(srcPath -> viewConfig)))(
+        WebConfig.default)
+
+      val r = withServer(port, webConfig) { baseUri: Uri =>
+        client.fetch(
+          Request(
+            uri = baseUri / "mount" / "fs" / "view" / "a",
+            method = Method.MOVE,
+            headers = Headers(Header("Destination", posixCodec.printPath(dstPath))))
```
jedesah: This will have to change if #1166 gets merged first or else it will need to be changed #1166. Mentioning, because git will not flag conflict.

web/src/main/scala/quasar/api/services/mount.scala
drostron: Sure, I rushed through this part a bit. Maybe we could pair quickly on this to arrive at something clearer

jedesah: Sure.

jedesah:
I am a little concerned that this does not fix the intent of SD-1339. @cryogenian is probably referring to using the `MOVE` operation on the `DATA` API when he refers to moving/deleting folders in the ticket.

If this "fixes" (quotes are for controversy around unification of data and mount api for which I am happy to chime in with my 2 cents, sorry I didn't do so earlier when it might have been more valuable) views being moved/deleted by such an operation, I don't believe there is any test asserting that.

drostron:
The "fix" should be tested due to the updates to `view` and `ViewMounter` and associated tests along with those being utilized already by the data api to verify that views weren't being referenced. That being said I think it would be clearer to have an http client level test for this case especially since I saw a regression in the mount api due to these changes.

Capturing an IRL discussion:
@jr and I were discussing the fact that view move operations happening via both the data api and mount api could be confusing. It might be clearer to support only FileSystem mount moves/deletes via the mount api and only View mount moves/deletes via the data api.

jedesah:
And I would just add, with the ultimate goal of doing both kind of moves through the Data API eventually.

web/src/main/scala/quasar/api/services/mount.scala
drostron: @jedesah and I took a closer look at this and decided to keep it as is, at least for now.

drostron:
Capturing an in person discussion with @jedesah:
Due to time constraints, we're going to postpone constraining view move and delete operations to only the data API.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -147,7 +147,13 @@ object view {
               f => fileToFileMove(src </> file1(f), dst </> file1(f), semantics)))
           }.run
 
-      ViewMounter.ls[S](src).flatMap(move).flatMap(κ(manage.moveDir(src, dst, semantics).run))
+          ViewMounter.ls[S](src).flatMap { ps =>
```
jedesah: What does `ps` stand for?

drostron: PathSegments

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -147,7 +147,13 @@ object view {
...
+            if (ps.isEmpty)
```
jedesah: Why do we need to guard against an empty list? does `move` not handle an empty `Set`?

drostron: If there are no views we want to preserve the behavior of `moveDir`.

drostron:
@jedesah, I believe I've addressed the comments. Anything else before this is ready?

core/src/test/scala/quasar/fs/mount/view.scala
jedesah: Was this ever answered? I guess, considering the circumstances, not a big deal if the answer is "meh" as I don't think it changed, just got moved.

web/src/main/scala/quasar/api/services/mount.scala
jedesah: So I think the answer to this is that scalac is being flaky and not recognizing the `Functor` which is why `as` is not compiling.

jedesah:
So what was the issue with MongoDB 2.6 that was failing the Travis build?

drostron:
Unfortunately an opaque intermittent failure. I was unable to reproduce it locally.

drostron:
It does seem troubling and perhaps similar to what ControlServiceSpec was running into. That could be an argument for skipping with an issue reference.

web/src/main/scala/quasar/api/services/mount.scala
drostron: Correct.

jedesah:
I am okay with leaving as is and see if it's a problem.

core/src/test/scala/quasar/fs/mount/view.scala
drostron: Yeah, they were moved and/or written before the non-trace `viewInterpret` existed.

PR #1145, opened 2016-03-23
[SD-1323] Properly escape all paths

jedesah:
- Kill `quasar.fs.Path` (convert `sql.Expr` and `LogicalPlan`)
- Add a bunch of (shady) functions in `quasar.fs._` to replace missing `pathy` APIs until fixed in `pathy`
- Add documentation on `QueryFile` term. In particular documenting the fact that a filesystem implementation can safely assume that the plans it receives contain only absolute file references
- Add `AlphaAndSpecialCharacters` for debugging purposes


jedesah:
@sellout Any idea what could be causing the library functions not to be found? Having a hard time coming up with a hypothesis for things I changed could affect that?

core/src/main/scala/quasar/sql/parser.scala
```diff
@@ -430,10 +433,11 @@ class SQLParser extends StandardTokenParsers {
 }
 
 object SQLParser {
-  def parseInContext(sql: Query, basePath: Path):
+
+  def parse(sql: Query): ParsingError \/ Expr =
+    new SQLParser().parse(sql).map(_.transAna(repeatedly(normalizeƒ)))
```
mossprescott: NB: this code has moved to the package object on master, so this change will need to be manually merged.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -239,7 +238,11 @@ trait Compiler[F[_]] {
 
     def compileRelation(r: SqlRelation[CoExpr]): CompilerM[Fix[LogicalPlan]] =
       r match {
-        case TableRelationAST(name, _) => emit(LogicalPlan.Read(Path(name)))
+        case TableRelationAST(name, _) =>
+          // All paths should have been made absolute at this point by calling the `relativeTo` method
+          // on `SqlExpr`, but that is not reflected in the types so we add rootDir to any relative files
+          // we might come accross which seems like a pretty safe operation
+          emit(LogicalPlan.Read(name))
```
mossprescott: Does this comment still apply?

Also, `name` seems like a confusing choice of name now that these are Path.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -391,7 +394,7 @@ trait Compiler[F[_]] {
               rName   <- relationName(node).fold(fail, emit)
               table   <- CompilerState.subtableReq(rName)
             } yield
-              if (Path(rName).filename ≟ name) table
+              if ((rName:String) ≟ name) table
```
mossprescott: What's this about? It seems like `relationName` should maybe become `relationPath`, and then you might need to extract the filename as it did before. This seems more than a bit stringly-typd as is.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -66,6 +93,7 @@ object QueryFile {
   final case class ListContents(dir: ADir)
     extends QueryFile[FileSystemError \/ Set[PathName]]
 
+  /** This operation should return whethere a file exists in the filesystem.*/
```
mossprescott: Spelling: "whether" :)

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -44,6 +45,8 @@ package object fs {
   type APath = AbsPath[_]
   type RPath = RelPath[_]
 
+  type FilePath = pathy.Path[_,File,Sandboxed]
```
mossprescott: This is getting to be a lot of type aliases to keep track of:

- `AbsPath[T]` and `RelPath[T]` (defined in pathy, right?)
- APath and RPath
- PathName (just below), which should maybe be called PathSegment
- in repl.Command: XDir and XFile
- now FilePath

I would feel better if there was at least some kind of convention. For instance, it seems like there should be `FilePath[B]` and `DirPath[B]`, for symmetry, and then maybe `FPath = FilePath[_]` and `DPath = DirPath[_]`.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -18,6 +18,7 @@ package quasar
 
 import quasar.Predef._
 import quasar.effect.Failure
+import quasar.fp._
 import quasar.fp.free._
 
 import pathy.{Path => PPath}, PPath._
```
mossprescott: These aliased imports can be cleaned up now, right? Since fs.Path is gone.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -83,12 +86,36 @@ package object fs {
...
+  @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.OptionPartial"))
+  def sandboxCurrent[A,T](path: PPath[A,T,Unsandboxed]): PPath[A,T,Sandboxed] =
+    refineTypeAbs(path).fold(
+      abs => (rootDir[Sandboxed] </> abs.relativeTo(rootDir).get).asInstanceOf[PPath[A,T,Sandboxed]],
+      rel => (currentDir[Sandboxed] </> rel.relativeTo(currentDir).get).asInstanceOf[PPath[A,T,Sandboxed]])
+
+  // TODO[pathy]: Offer clean API in pathy to do this
+  def refineTypeAbs[T,S](path: PPath[_,T,S]): PPath[Abs,T,S] \/ PPath[Rel,T,S] = {
+    if (path.isAbsolute) path.asInstanceOf[PPath[Abs,T,S]].left
+    else path.asInstanceOf[PPath[Rel,T,S]].right
```
mossprescott: I think we should probably insist on these going into Pathy to avoid `asInstanceOf`.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -83,12 +86,36 @@ package object fs {
...
+  }
+
+  def mkAbsolute[T,S](path: PPath[_,T,S]): PPath[Abs,T,S] =
+    refineTypeAbs(path).fold(ι, rootDir[S] </> _)
```
mossprescott: Would this be better expressed as something like `path.from(rootDir)`? I.e. make the base dir a parameter.

core/src/main/scala/quasar/fs/path.scala
```diff
@@ -1,240 +0,0 @@
...
-      case _                       => None
-    }
-  }
-}
-
-final case class DirNode(value: String)
-object DirNode {
-  val Current = DirNode(".")
-}
-final case class FileNode(value: String)
```
mossprescott: :tada: :ice_cream: :gift: :beers:

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -124,11 +125,11 @@ object LogicalPlan {
         }
     }
 
-  final case class ReadF[A](path: Path) extends LogicalPlan[A] {
-    override def toString = s"""Read(Path("${path.simplePathname}"))"""
+  final case class ReadF[A](path: FilePath) extends LogicalPlan[A] {
+    override def toString = s"""Read("${posixCodec.printPath(path)}")"""
```
mossprescott: Although it's slightly less readable, I guess I'd prefer `path.shows` for this sort of thing. That way there's never any question about escaping, and it's also easier to copy and paste as source.

web/src/test/scala/quasar/api/UriPathCodecSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+        codec.parseRelFile(codec.printPath(path)) must_== Some(path)
+      }
+      "absolute dir" ! prop { path: ADir =>
+        codec.parseAbsDir(codec.printPath(path)) must_== Some(path)
+      }
+      "relative dir" ! prop { path: RDir =>
+        codec.parseRelDir(codec.printPath(path)) must_== Some(path)
+      }
+    }
+  }
```
mossprescott: There's one test hiding in AsPathSpec that should be moved here. And I'd also like to see a handful of tests documenting/verifying the behavior around `.` and `..`.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -143,28 +142,28 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s with Pa
 
       // FIXME: use APath (see PathGen)
       // TODO: escaped paths do not survive being embedded in error messges
```
mossprescott: These TODOs are all resolved, right?

core/src/main/scala/quasar/compiler.scala
jedesah: Yes, the comment still applies because `TableRelationAST` has a `File` which I called `FilePath` (`FilePath = Path[_,File,Sandboxed]`) because `File` would conflict with the phantom type (is that how we call it) from `pathy`. I thought about type indexing `Expr` (and `ExprF`) on what type of file it contains, relative or absolute, and I still think that's reasonable, but I am keeping it that work for another day because it will likely bring up all kinds of issues with `Fix` because `ExprF` would have to become a higher kinded type with 2 holes instead of just one.

Happy to change `name` to `file`. Unless `path` is better? My initial reasoning for leaving it intact was because even though it was represented by a `String` before, it still was a "Path" and so I figured it was named for what it represented in sql which is the name of a table. But I think path to a table is fine, I don't have a strong opinion which is why I am happy to rename it to `file`.

core/src/main/scala/quasar/compiler.scala
jedesah: It represents the name associated with a relation. i.e the alias in the case there is one. If not, unlike before, we handle converting a file path into a name somewhere else where it can be handled only once as opposed to in many different places as is the case currently. I think `String` is probably fine, but we could create a wrapper class `case class TableName(name: String)` for more type safety, probably not a bad idea.

As for why the type annotation is necessary for a successful compile, no idea.

core/src/main/scala/quasar/fs/package.scala
jedesah: Okay, I can get behind that.

core/src/main/scala/quasar/fs/package.scala
jedesah: They sure can!

core/src/main/scala/quasar/fs/package.scala
jedesah: You want to hold up this PR for a new version of `pathy`. I don't mind but I thought that that was a relatively lengthy process.

core/src/main/scala/quasar/fs/package.scala
jedesah: Sure.

core/src/main/scala/quasar/logicalplan.scala
jedesah: Don't disagree.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
jedesah: Are they? I am not sure. I don't fully understand them to be honest.

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -39,22 +39,49 @@ object QueryFile {
...
   final case class More(h: ResultHandle)
     extends QueryFile[FileSystemError \/ Vector[Data]]
 
+  /** Represents an "close" operation. Used to halt streaming of a result set initiated using
+    * the `EvaluatePlan` operation.
+    */
   final case class Close(h: ResultHandle)
     extends QueryFile[Unit]
 
+  /** Represents an "explain plan" operation. This operation should not actually have any side effect
```
sellout: I would be inclined to remove the first sentence from each of these comments, possibly renaming `Explain` to `ExplainPlan` since that’s the only one that differs from the description.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
```diff
@@ -258,15 +258,16 @@ private final class QueryFileInterpreter[C](
...
   private def checkPathsExist(lp: Fix[LogicalPlan]): MongoLogWFR[Unit] = {
-    def checkPathExists(p: QPath): MongoFsM[Unit] = for {
+    // Documentation on `QueryFile` guarantees absolute paths, so calling `mkAbsolute`
+    def checkPathExists(p: AFile): MongoFsM[Unit] = for {
       coll <- EitherT.fromDisjunction[MongoDbIO](Collection.fromPath(p))
-                .leftMap(e => pathError(invalidPath(p.asAPath, e.message)))
+                .leftMap(pathError(_))
       _    <- EitherT(MongoDbIO.collectionExists(coll)
-                .map(_ either (()) or pathError(pathNotFound(p.asAPath))))
+                .map(_ either (()) or pathError(PathError2.pathNotFound(p))))
```
sellout: I thought we got rid of all the `2` suffixed things. Did this just get left behind, or is there a reason we still have duplicated PathErrors?

sellout: Ah, I see – so `PathError` disappeared with `quasar.fs.Path`. This PR should rename `PathError2` to `PathError`, then.

core/src/main/scala/quasar/sql/ast.scala
```diff
@@ -153,20 +157,31 @@ sealed trait SqlRelation[A] {
 
     collect(this).groupBy(_._1).mapValues(_.map(_._2))
   }
+
+  def mapPathsM[F[_]: Monad](f: FPath => F[FPath]): F[SqlRelation[A]]
```
sellout: I’d rather see this defined functionally – all here, instead of distributed across the sum components.

core/src/test/scala/quasar/fs/PathArbitrary.scala
```diff
@@ -27,6 +30,52 @@ trait PathArbitrary {
   implicit val arbitraryRPath: Arbitrary[RPath] =
     Arbitrary(Gen.oneOf(Arbitrary.arbitrary[RFile], Arbitrary.arbitrary[RDir]))
 
+  implicit val arbitraryFPath: Arbitrary[FPath] =
```
sellout: Should this stuff all be in pathy-scalacheck?

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
sellout: It looks like it to me.

sellout:
Very happy with all of this. Some fairly minor comments.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
jedesah: It should probably be fixed because it's pathy all the way down now, but it's weird removing a `TODO` on a test without changing the test, I would expect some constraint to be removed or something. I mean, if the test was already passing unconstrained, what was the problem? If no one can explain its meaning, then it's just a dead comment and should be removed anyway.

Also, as for the `FIXME`, there now is an `Arbitrary[APath]` so I could change it to `APath`, but I don't fully understand why that's desired, I thought mounts could only be directories?

sellout: Views are currently modeled as file mounts, aren’t they? And in the future there will definitely be file mounts (like, pointing to a random JSON file on disk). But I read that as “generate an arbitrary directory path,” which I assumed was more flexible than what `AbsDirOf` was doing.

With the `TODO`, I may be wrong, but I wouldn’t expect the fix for it to be here. My assumption was that Pathy paths behave correctly when printing characters that require escaping, while the fs.Path ones didn’t.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
jedesah: I tried that, but the problem is that we have a `PathError` that is a subclass of `FileSystemError` which takes as a parameter a `PathError2`. Since they are almost always used together I think it would be a huge pain to have them be named the same. Maybe @wemrysi had something else in mind when he first wrote this stuff since he did put a comment saying to rename to `PathError` once `fs.Path` was killed.

wemrysi: If the naming is a problem, I'd rename the `FileSystemError` constructor to something that doesn't clash and then rename `PathError2` to `PathError`.

jedesah: @wemrysi Do you have a suggestion for the constructor name?

wemrysi: `PathErr`?

core/src/test/scala/quasar/fs/PathArbitrary.scala
jedesah: Some of it should probably be there. Not all of it, since `AFile` and such are defined in `quasar`.

We already have two or three files in the codebase that contain pathy related stuff. I think it makes sense to periodically go through these files and add what makes sense to `pathy`, do a new release and cleanup the newly duplicated code in `quasar`. I think it's a fair question whether we should try and avoid putting things in `quasar` that should really go in a third party library (well, `pathy` is borderline 3rd party, but from the perspective of `quasar`, I guess it's all the same whether it's `pathy` or `scalaz`) or if it's okay for them to "incubate" in `quasar` until they have proven useful and well tested and then migrate them to the external library where they can be re-used by other projects.

Not to mention that periodically migrating things from `quasar` to `pathy` seems more efficient than constantly juggling concurrent releases.

jedesah:
@sellout I think I addressed all of your comments. Mind doing a quick review of the additional commits?

sellout:
Ok, sorry that took so long! 👍🏾

PR #1146, opened 2016-03-23
Fix two memory leaks

mossprescott:
See [SD-950](https://slamdata.atlassian.net/browse/SD-950).

This fixes two leaks that showed up pretty quickly with some profiling and a very simple shell script to generate a steady request load.

The one that presumably was causing the crashes reported by the user was the `waitForInput` loop, which would consume 100s of MB over the course of a week even with no requests coming in.

### DIFF
--- a/core/src/main/scala/quasar/sql/package.scala
+++ b/core/src/main/scala/quasar/sql/package.scala
@@ -27,7 +27,7 @@ package object sql {
   type Expr = Fix[ExprF]
 
 
-  private def parser = new SQLParser()
+  private val parser = new SQLParser()
 
   val parse: Query => (ParsingError \/ Expr) = parser.parse
 
--- a/core/src/main/scala/quasar/sql/package.scala
+++ b/core/src/main/scala/quasar/sql/package.scala
@@ -26,6 +26,19 @@ import scalaz._, Scalaz._
 package object sql {
   type Expr = Fix[ExprF]
 
+
+  private def parser = new SQLParser()
+
+  val parse: Query => (ParsingError \/ Expr) = parser.parse
+
+  def parseInContext(sql: Query, basePath: Path):
+      ParsingError \/ Expr =
+    parser.parse(sql)
+      .flatMap(relativizePaths(_, basePath).bimap(
+        ParsingPathError,
+        _.transAna(repeatedly(normalizeƒ))))
+
+
   def CrossRelation(left: SqlRelation[Expr], right: SqlRelation[Expr]) =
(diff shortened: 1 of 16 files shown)


core/src/main/scala/quasar/sql/package.scala
```diff
@@ -26,6 +26,19 @@ import scalaz._, Scalaz._
 package object sql {
   type Expr = Fix[ExprF]
 
+
+  private val parser = new SQLParser()
+
+  val parse: Query => (ParsingError \/ Expr) = parser.parse
```
jedesah: This does not call `normalizeƒ` which it probably should, but I have fixed this in #1145.

web/src/main/scala/quasar/server/Http4sUtils.scala
```diff
@@ -45,13 +45,14 @@ object Http4sUtils {
...
-      done <- if (test) waitForInput else Task.now(())
-    } yield done
+
+    val sleep = Task.delay(java.lang.Thread.sleep(250))
+                    .handle { case _: java.lang.InterruptedException => () }
+
+    val inputPresent = Task.delay(Option(System.console).nonEmpty && System.in.available() > 0)
+                           .handle { case _ => false }
+
+    Process.eval(sleep *> inputPresent).repeat.takeWhile(!_).run
```
jedesah: Nice! :+1:

So the memory leak was being caused because of the non tail-recursive function?

core/src/main/scala/quasar/sql/package.scala
mossprescott: I'm just moving the code here. When I rebase, I'll make sure I didn't lose that change.

web/src/main/scala/quasar/server/Http4sUtils.scala
mossprescott: Pretty much. I'm not really sure exactly how this works, but it seems that Task is trampolined enough to not blow the stack, and instead piles up thunks in the heap. I tried a couple of different fixes, and this one seemed the cleanest (and looks good in the profiler).

mossprescott:
I've been running a build of this branch for 21 hours, with a script running a query, downloading an HTML page, and uploading a small JSON dataset every 5 seconds. There are no new retained objects visible in the memory profiler, and the used hep size is steady at just under 200 MB.

wemrysi:
@mossprescott Thanks for your efforts in fixing these! :+1:

mossprescott:
Merged as 5.0.2 via the script.

PR #1147, opened 2016-03-24
Close backend connections after running regression tests

mossprescott:
This solves the problem of long SBT sessions resulting in many open connections piling up and eventually making mongodb unresponsive.

Confirmed that _all_ connections are closed by watching the `mongod` log and by inspecting the heap (which is apparently my new thing).

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
+++ b/core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
@@ -21,19 +21,17 @@ import quasar.EnvironmentError
 import quasar.fs.{FileSystem, FileSystemType}
 
 import scala.StringContext
-
 import scalaz._
-import scalaz.syntax.bifunctor._
-import scalaz.std.tuple._
 
 import FileSystemDef._
 
 final case class FileSystemDef[F[_]](run: FsCfg => DefErrT[F, DefinitionResult[F]]) {
+
   def apply(typ: FileSystemType, uri: ConnectionUri): DefErrT[F, DefinitionResult[F]] =
     run((typ, uri))
 
   def translate[G[_]: Functor](f: F ~> G): FileSystemDef[G] =
-    FileSystemDef(c => EitherT(f(run(c).run)).map(_.bimap(f compose _, f(_))))
+    FileSystemDef(c => EitherT(f(run(c).run)).map(_.translate(f)))
 }
 
 object FileSystemDef {
@@ -41,7 +39,11 @@ object FileSystemDef {
   /** Reasons why the configuration is invalid or an environment error. */
   type DefinitionError        = NonEmptyList[String] \/ EnvironmentError
   type DefErrT[F[_], A]       = EitherT[F, DefinitionError, A]
-  type DefinitionResult[F[_]] = (FileSystem ~> F, F[Unit])
+
+  final case class DefinitionResult[F[_]](run: FileSystem ~> F, close: F[Unit]) {
+    def translate[G[_]: Functor](f: F ~> G): DefinitionResult[G] =
+      DefinitionResult(f compose run, f(close))
+  }
 
   def fromPF[F[_]: Monad](
     pf: PartialFunction[FsCfg, DefErrT[F, DefinitionResult[F]]]
--- a/core/src/test/scala/quasar/fs/WriteFileSpec.scala
+++ b/core/src/test/scala/quasar/fs/WriteFileSpec.scala
@@ -19,6 +19,7 @@ package quasar.fs
 import quasar.Predef._
 import quasar.{Data, DataArbitrary}
 import quasar.fp._
+import quasar.specs2.DisjunctionMatchers
 
 import org.specs2.mutable.Specification
 import org.specs2.ScalaCheck
@@ -28,7 +29,7 @@ import scalaz.std.vector._
 import scalaz.syntax.monad._
 import scalaz.stream._
 
-class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture {
+class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture with DisjunctionMatchers {
   import DataArbitrary._, FileSystemError._, PathError._
 
   type DataWriter = (AFile, Process0[Data]) => Process[write.M, FileSystemError]
@@ -74,6 +75,15 @@ class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture
       }
     }
 
+    "append should fail after writing some when source fails" ! prop {
+      (f: AFile, xs: Vector[Data]) =>
+
+      // TODO: handle the error, and check the contents of the file
+
+      val p = write.append(f, xs.toProcess ++ Process.fail(new RuntimeException("source failed")))
+      MemTask.runLogEmpty(p).attemptRun must beLeftDisjunction
+    }
+
     withDataWriters(("save", write.save), ("saveThese", write.saveThese)) { (n, wt) =>
       s"$n should replace existing file" ! prop {
         (f: AFile, xs: Vector[Data], ys: Vector[Data]) =>
@@ -102,6 +112,15 @@ class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture
         }
       }
     }
+
+    "save should fail and write nothing when source fails" ! prop {
+      (f: AFile, xs: Vector[Data]) =>
+
+      // TODO: handle the error, and check that the file is not written
+
+      val p = write.save(f, xs.toProcess ++ Process.fail(new RuntimeException("source failed")))
+      MemTask.runLogEmpty(p).attemptRun must beLeftDisjunction
+    }.pendingUntilFixed("SD-1544")
 
     withDataWriters(("create", write.create), ("createThese", write.createThese)) { (n, wt) =>
       s"$n should fail if file exists" ! prop {
--- a/it/src/test/scala/quasar/TaskResource.scala
+++ b/it/src/test/scala/quasar/TaskResource.scala
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar
+
+import quasar.Predef._
+
+import java.util.concurrent.atomic.AtomicLong
+import scalaz._, Scalaz._
+import scalaz.concurrent.{Strategy, Task}
+import scalaz.stream._
+
+sealed abstract class TaskResource[A] {
+  /** Get an instance of the resource, first acquiring it if necessary. */
+  def get: Task[A]
+
+  /** Release the cached instance, if it was ever acquired. */
+  def release: Task[Unit]
+}
+
+object TaskResource {
+  /** "Memo-ized" resource, given an effect to create it and an effect to
+    * dispose of it. Effects are processed asynchronously using the provided
+    * Strategy.
+    * @param acquire Executed exactly once for any number of `get` calls as
+    *   long as `release` is not called.
+    * @param release0 Executed exactly once for any number of `release`
+    *   calls as long as `get` is not called.
+    */
+  def apply[A](acquire: Task[A], execStrat: Strategy)(release0: A => Task[Unit]): Task[TaskResource[A]] = Task delay {
+    val counter = new AtomicLong(0)
+    def nextId: Task[Long] = Task.delay(counter.getAndIncrement)
+
+    sealed trait RsrcState
+    case object Start extends RsrcState
+    case class Acquiring(id: Long) extends RsrcState
+    case class Acquired(a: A) extends RsrcState
+
+    val signal = async.signalOf[RsrcState](Start)
+
+    def orFail[B]: Option[B] => Task[B] = _.cata(Task.now, Task.fail(Cause.Terminated(Cause.End)))
+
+    /** The current value of the signal, or fail if it hasn't been initialized
+      * (and ours is initialized when it's created.) This is the behavior you
+      * presumably want for `Signal.get`, but that method actually uses the
+      * `discrete` stream.
+      */
+    val getCurrent: Task[RsrcState] =
+      signal.continuous.take(1).runLast.flatMap(orFail)
+
+    /** Asynchronously wait for the signal to contain the resource, which
+      * will have been acquired by another thread.
+      */
+    val awaitAcquired: Task[A] =
+      // NB: something effectful is hiding in the `discrete` process
+      Task.delay(signal.discrete.collect { case Acquired(a) => a }.take(1).runLast).join.flatMap(orFail)
+
+    new TaskResource[A] {
+      def get = {
+        def attemptAcquire(id: Long): Task[A] =
+          signal compareAndSet {
+            case Some(Start) => Acquiring(id).some
+            case otherwise   => otherwise
+          } flatMap {
+            case Some(Acquiring(`id`)) => acquire >>= (a => signal.set(Acquired(a)).as(a))
+            case _                     => awaitAcquired
+          }
+
+        getCurrent flatMap {
+          case Acquired(a) => Task.now(a)
+          case _           => nextId >>= attemptAcquire
+        }
+      }
+
+      def release =
+        signal.getAndSet(Start).attempt flatMap {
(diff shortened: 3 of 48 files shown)


mossprescott:
Looks like I spoke too soon. There are still some threads active. I'm looking at it some more.

core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
```diff
@@ -41,7 +39,11 @@ object FileSystemDef {
   /** Reasons why the configuration is invalid or an environment error. */
   type DefinitionError        = NonEmptyList[String] \/ EnvironmentError
   type DefErrT[F[_], A]       = EitherT[F, DefinitionError, A]
-  type DefinitionResult[F[_]] = (FileSystem ~> F, F[Unit])
+
+  final case class DefinitionResult[F[_]](run: FileSystem ~> F, close: F[Unit]) {
```
wemrysi: :+1:

it/src/test/scala/quasar/physical/mongodb/filesystems.scala
```diff
@@ -20,37 +20,40 @@ import quasar.Predef._
...
-  ): Task[FileSystem ~> Task] = {
-    val prg = for {
-      client   <- createAsyncMongoClient[MongoEff](uri)
-      mongofs0 =  mongoDbFileSystem[MongoEff](client, DefaultDb fromPath prefix)
-      mongofs  <- envErr.unattempt(lift(mongofs0.run).into[MongoEff])
-    } yield mongofs
+  ): Task[(FileSystem ~> Task, Task[Unit])] = {
+    val fsDef = quasar.physical.mongodb.fs.mongoDbFileSystemDef[MongoEff].apply(MongoDBFsType, uri).run
+      .flatMap[FileSystemDef.DefinitionResult[MongoEffM]] {
+        case -\/(-\/(strs)) => injectFT[Task, MongoEff].apply(Task.fail(new RuntimeException(strs.list.mkString)))  // FIXME?
```
wemrysi: If we don't want to fail here, we could change the signature of `TestConfig#externalFileSystems` to allow for failure in the `PartialFunction` (maybe via `DefErrT`). This would also allow us to report initialization/config errors as such there.

mossprescott:
This is ready for review finally.

See the final set of commits, which breaks down into:
- The actual change to close/avoid connections, along with several fixes and improvements in the handling of paths and datasets in the regression tests.
- Changing the handling of paths in the SQL AST to allow unsandboxed paths (notably `../foo`)
- Increasing the MongoDB connection timeout, which was and is hard-coded. This is our own timeout on the async task that attempts to check the version.

The last change is required because with these changes Travis was running into the timeout fairly often, most commonly with the RO configuration (in which two connections are created at the same time). I infer that it has something to do with the way timing has changed. We now wait until the last possible moment to create connections, whereas previously connections were created for earlier and perhaps ended up being more spread out in time. Perhaps now when two test suites are running in parallel their multiple connections are created at the same time, compete with each other for the CPU/IO and take longer to be established. Something like that anyway.

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  /** Release the cached instance, if it was ever acquired. */
+  def release: Task[Unit]
+}
+
+object TaskResource {
+  def apply[A](acquire: Task[A])(release0: A => Task[Unit]): Task[TaskResource[A]] = Task.delay {
+    // Call this a thought experiment. I have few illusions about getting
+    // this past the FP thought police, but it works and demonstrates what
+    // needs to be accomplished here.
```
mossprescott: @wemrysi: Please do object to this approach and propose an alternative. I believe it can be implemented on top of TaskRef, but it will involve spinning. I would be happier with a better encapsulated lock-in-Task abstraction. Thoughts?

core/src/test/scala/quasar/fs/WriteFileSpec.scala
```diff
@@ -74,6 +75,15 @@ class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture
       }
     }
 
+    "append should fail after writing some when source fails" ! prop {
+      (f: AFile, xs: Vector[Data]) =>
+
+      // TODO: handle the error, and check the contents of the file
```
wemrysi: Can any of this be done now? If not, what's preventing it?

mossprescott: Mostly it just wasn't obvious how to thread the error through the Process types here and I figured we'd come back to it when we fix the bug in `save` (the pending test). If you've got a simple way to implement it, I could add it.

it/src/test/scala/quasar/TaskResource.scala
wemrysi: @mossprescott I think I may have come up with a (untested) [non-blocking implementation](https://gist.github.com/wemrysi/d0e77f317db8ad750fc457424032d6db), let me know what you think. I'd prefer avoid locks wherever possible, especially if we want to make this general-purpose.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -55,11 +55,12 @@ abstract class QueryRegressionTest[S[_]: Functor](
   val injectTask: Task ~> F =
     liftFT[S].compose(injectNT[Task, S])
 
-  val TestsRoot = new File("it/src/main/resources/tests")
+  // val TestsRoot = new JFile("it/src/main/resources/tests")
```
wemrysi: Debug comment?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -77,12 +78,14 @@ abstract class QueryRegressionTest[S[_]: Functor](
...
   fileSystemShould { fs =>
     suiteName should {
+      step(println("testDir: " + fs.testDir.shows))
       step(prepareTestData(tests, fs.setupInterpM).run)
 
       tests.toList foreach { case (f, t) =>
         regressionExample(f, t, fs.name, fs.testInterpM)
       }
 
+      // step(println("delete testDir: " + fs.testDir.shows))
```
wemrysi: Debug comment?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -91,19 +94,22 @@ abstract class QueryRegressionTest[S[_]: Functor](
...
     def runTest = (for {
       _    <- Task.delay(println(test.query))
-      _    <- run(test.data.cata(verifyDataExists, success.point[F]))
-      data =  testQuery(test.query, test.variables)
+      _    <- run(test.data.cata(
+                fn => verifyDataExists(dataFile(sandbox(currentDir, unsandbox(fileParent(loc)) </> fn).get /* FIXME */)),
+                success.point[F]))
+      data =  testQuery(DataDir </> fileParent(loc), test.query, test.variables)
       res  <- verifyResults(test.expected, data, run)
+      _    <- Task.delay(println("done: " + test.query))
```
wemrysi: Did you intend to keep this? I think it was added for debugging purposes.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -156,33 +165,44 @@ abstract class QueryRegressionTest[S[_]: Functor](
...
-  def prepareTestData(tests: Map[File, RegressionTest], run: Run): Task[Unit] = {
+  def prepareTestData(tests: Map[RFile, RegressionTest], run: Run): Task[Unit] = {
     val throwFsError = rethrow[Task, FileSystemError]
 
-    val dataLoc: ((File, RegressionTest)) => Set[File] = {
-      case (f, t) => t.data.map(new File(f.getParent, _)).toSet
+    val dataLoc: ((RFile, RegressionTest)) => Set[RFile] = {
+      case (f, t) =>
+        (for {
+          // f1 <- posixCodec.parseRelFile(f.getPath)
```
wemrysi: Debug comment

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -156,33 +165,44 @@ abstract class QueryRegressionTest[S[_]: Functor](
...
+          d  <- t.data
+          p  = fileParent(f)
+          f2 <- sandbox(currentDir, unsandbox(p) </> d)
+        } yield f2).toSet
     }
 
+    // println(tests.toList.foldMap(dataLoc).toVector.mkString("\n"))
```
wemrysi: Debug comment

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -156,33 +165,44 @@ abstract class QueryRegressionTest[S[_]: Functor](
...
+        Process.eval_(Task.delay { println(s"loading ${posixCodec.printPath(dataFile(file))}") }) ++
+        write.createChunked(
+          dataFile(file),
           testData(file).chunk(500).translate(injectTask)
-        ).translate[Task](throwFsError.compose[FsErr](runT(run)))
+        ).translate[Task](throwFsError.compose[FsErr](runT(run))) ++
+        Process.eval_(Task.delay { println(s"... ${fileName(file)}") })
       }
 
+    // throwFsError(EitherT(loads.join.map(e => { println("error: " + e); e }).take(1).runLast.map(_ <\/ (()))))
```
wemrysi: Debug comment

mossprescott:
Sorry, I had left some debug in for troubleshooting Travis and missed some of it. I'll do a pass through the diff and get rid of it.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -156,33 +165,44 @@ abstract class QueryRegressionTest[S[_]: Functor](
...
+        } yield f2).toSet
     }
 
+    // println(tests.toList.foldMap(dataLoc).toVector.mkString("\n"))
+
     val loads: Process[Task, Process[Task, FileSystemError]] =
       Process.emitAll(tests.toList.foldMap(dataLoc).toVector) map { file =>
-        write.saveChunked(
-          dataFile(file.getName),
+        Process.eval_(Task.delay { println(s"loading ${posixCodec.printPath(dataFile(file))}") }) ++
```
wemrysi: Did you intend to keep this and the one that follows as well?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -190,19 +210,22 @@ abstract class QueryRegressionTest[S[_]: Functor](
                )),
         Process.emit(_))
 
-    io.linesR(new FileInputStream(file)) flatMap parse
+    val jf = jFile(TestsRoot </> file)
+    Task.delay(jf.exists).map(x => { println("exists: " + x + "; " + jf); x}).liftM[Process].ifM(
```
wemrysi: Debug code (`println`)

it/src/test/scala/quasar/regression/RegressionTest.scala
```diff
@@ -19,21 +19,23 @@ package regression
 
 import quasar.Predef._
 import quasar.fp._
+// import quasar.fs.RFile
```
wemrysi: Delete this one too.

core/src/test/scala/quasar/fs/WriteFileSpec.scala
wemrysi: Ok, is this actually testing the issue then? If not then I think I'd prefer it to be pending as well.

wemrysi:
Thanks for your efforts in tackling another sticky issue! Let me know what you think about the memoized resource impl, other than that and the debug code, looks good to me.

core/src/test/scala/quasar/fs/WriteFileSpec.scala
mossprescott: Well, it's demonstrating `append` at least does fail, unlike `save`, which I think is interesting.

wemrysi: :+1:

mossprescott:
@wemrysi Note the last commit turing the value into a simple ADT, which makes it a lot easier to follow as values are pushed in and out of the signal. Also, I eliminated the recursive call to `get` which was never being reached, since the `runLast` in `awaitAcquired` should never actually produce None.

core/src/test/scala/quasar/fs/WriteFileSpec.scala
wemrysi: Up to you, but I have started working on SD-1544 in another branch and have fleshed out these tests to do what the `TODO` describes, I think you could safely delete them from this PR if you want.

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -0,0 +1,90 @@
...
+package quasar
+
+import quasar.Predef._
+
+import java.util.concurrent.atomic.AtomicLong
+import scalaz._, Scalaz._
+import scalaz.concurrent.{Strategy, Task}
+import scalaz.stream._
+
+sealed abstract class TaskResource[A] {
```
wemrysi: What do you think of just `Resource` or maybe `MemoResource` as a name as the fact that it is in `Task` is more of an implementation detail?

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -0,0 +1,90 @@
...
+  /** Get an instance of the resource, first acquiring it if necessary. */
+  def get: Task[A]
+
+  /** Release the cached instance, if it was ever acquired. */
+  def release: Task[Unit]
+}
+
+object TaskResource {
+  def apply[A](acquire: Task[A], execStrat: Strategy)(release0: A => Task[Unit]): Task[TaskResource[A]] = Task delay {
```
wemrysi: Mind adding some documentation here regarding the at-most-once guarantees of running `acquire` and `release0`?

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -0,0 +1,90 @@
...
+    val counter = new AtomicLong(0)
+    def nextId: Task[Long] = Task.delay(counter.getAndIncrement)
+
+    sealed trait RsrcState
+    case object Start extends RsrcState
+    case class InProgress(id: Long) extends RsrcState
```
wemrysi: Minor suggestion, what about `Acquiring`?

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -0,0 +1,90 @@
...
+      * (and ours is initialized when it's created.) This is the behavior you
+      * presumably want for `Signal.get`, but that method actually uses the
+      * `discrete` stream.
+      */
+    val getCurrent: Task[RsrcState] =
+      signal.continuous.take(1).runLast.flatMap(orFail)
+
+    /** Asynchronously wait for the signal to contain the resource, which
+      * will have been acquired by another thread.
+      * NB: this has to be a def, not a val, so that the discrete stream is
```
wemrysi: Interesting and somewhat surprising...does this imply that `signal.discrete` isn't referentially transparent?

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -0,0 +1,90 @@
...
+    def awaitAcquired: Task[A] =
+      signal.discrete.collect { case Acquired(a) => a }.take(1).runLast.flatMap(orFail)
+
+    new TaskResource[A] {
+      def get =
+        getCurrent flatMap {
+          case Acquired(a) => Task.now(a)
+          case _           => for {
+            id <- nextId
+            s  <- signal compareAndSet {
```
wemrysi: Any reason to prefer the for-comprehension to a separate function using combinators (similar to the one from the gist)? I only ask as I find having to name the intermediate values just to match on them more noisy than the alternative.

mossprescott: I just worked it back and forth and this is what I ended up liking best. The starting point was wanting to get rid of the inner fn called (meaninglessly) "getAndSet".

Having done that, there were three Tasks being chained together with flatMap and a for comprehension seemed simpler, and required less awkward line breaking.

Which name(s) is it that you find noisy? Just `s`?

it/src/test/scala/quasar/regression/RegressionTest.scala
wemrysi: This one's a survivor ;)

it/src/test/scala/quasar/TaskResource.scala
mossprescott: Yes, I think it does. I ran into it when I reflexively made this a `val` and everything hung again.

it/src/test/scala/quasar/TaskResource.scala
wemrysi: Yeah, `s` and `a`. If you're looking for a more evocative name for the inner fn, how about `attemptAcquire`?

That said, this looks to be just a difference in preferences, so feel free to keep this as-is if you disagree, it's fine to merge either way.

mossprescott: I took that suggestion and tried again and I like it better now that the fn and the state both have meaningful names.

wemrysi:
LGTM, :+1:

mossprescott:
6.1.0

PR #1150, opened 2016-03-27
SD-824 Standardize and enrich REST API errors

wemrysi:
A breaking change implementing the enrichment and standardization features for REST API errors as described in [SD-824](https://slamdata.atlassian.net/browse/SD-824).

* More meaningful HierarchicalFileSystemError
* Introduces ApiError as a single datatype to represent an error to be conveyed via HTTP
* Adds conversions from all errors to ApiError in lieu of converting directly to a response for consistency, including error-specific data where applicable

### DIFF
--- a/core/src/main/scala/quasar/types.scala
+++ b/core/src/main/scala/quasar/types.scala
@@ -22,6 +22,7 @@ import SemanticError.{TypeError, MissingField, MissingIndex}
 
 import scala.Any
 
+import argonaut._, Argonaut._
 import scalaz._, Scalaz._, NonEmptyList.nel, Validation.{success, failure}
 
 sealed trait Type { self =>
@@ -170,6 +171,8 @@ sealed trait Type { self =>
 }
 
 trait TypeInstances {
+  import Type._
+
   val TypeOrMonoid = new Monoid[Type] {
     def zero = Type.Bottom
 
@@ -202,6 +205,55 @@ trait TypeInstances {
 
   implicit val TypeRenderTree: RenderTree[Type] =
     RenderTree.fromToString[Type]("Type")
+
+  implicit val typeEncodeJson: EncodeJson[Type] =
+    EncodeJson {
+      case Top =>
+        jString("Top")
+      case Bottom =>
+        jString("Bottom")
+      case Const(d) =>
+        Json("Const" -> DataCodec.Precise.encode(d).getOrElse(jNull))
+      case Null =>
+        jString("Null")
+      case Str =>
+        jString("Str")
+      case Int =>
+        jString("Int")
+      case Dec =>
+        jString("Dec")
+      case Bool =>
+        jString("Bool")
+      case Binary =>
+        jString("Binary")
+      case Timestamp =>
+        jString("Timestamp")
+      case Date =>
+        jString("Date")
+      case Time =>
+        jString("Time")
+      case Interval =>
+        jString("Interval")
+      case Id =>
+        jString("Id")
+      case Arr(types) =>
+        Json("Array" := types)
+      case FlexArr(min, max, mbrs) =>
+        val flexarr =
+          ("minSize" :=  min) ->:
+          ("maxSize" :?= max) ->?:
+          ("members" :=  mbrs) ->:
+          jEmptyObject
+        Json("FlexArr" -> flexarr)
+      case Obj(assocs, unkns) =>
+        val obj =
+          ("associations" :=  assocs) ->:
+          ("unknownKeys"  :?= unkns)  ->?:
+          jEmptyObject
+        Json("Obj" -> obj)
+      case cp @ Coproduct(l, r) =>
+        Json("Coproduct" := cp.flatten)
+    }
 }
 
 final case object Type extends TypeInstances {
--- a/web/src/main/scala/quasar/api/services/mount.scala
+++ b/web/src/main/scala/quasar/api/services/mount.scala
@@ -17,12 +17,12 @@
 package quasar.api.services
 
 import quasar.Predef._
-import quasar.api._
-import quasar.fp._
+import quasar.api._, ToApiError.ops._
+import quasar.fp._, PathyCodecJson._
 import quasar.fs.{AbsPath, APath, sandboxAbs}
 import quasar.fs.mount._
 
-import argonaut._
+import argonaut._, Argonaut._
 import org.http4s._, Method.MOVE
 import org.http4s.dsl._
 import pathy.Path, Path._
@@ -30,30 +30,34 @@ import scalaz._, Scalaz._
 import scalaz.concurrent.Task
 
 object mount {
-  import ToQResponse.ops._
   import posixCodec._
 
   def service[S[_]: Functor](implicit M: Mounting.Ops[S], S0: Task :<: S): QHttpService[S] =
     QHttpService {
       case GET -> AsPath(path) =>
-        def err = s"There is no mount point at ${printPath(path)}"
-        respond(M.lookup(path).toRight(QResponse.error[S](NotFound, err)).run)
+        def err = ApiError.fromMsg(
+          NotFound withReason "Mount point not found.",
+          s"There is no mount point at ${printPath(path)}",
+          "path" := path)
+        respond(M.lookup(path).toRight(err).run)
 
       case req @ MOVE -> AsPath(src) =>
-        requiredHeader(Destination, req).map(_.value).fold(
-          (_: QResponse[S]).point[Free[S, ?]],
+        respond(requiredHeader(Destination, req).map(_.value).fold(
+          err => EitherT.leftU[String](err.point[Free[S, ?]]),
           dst => refineType(src).fold(
-            srcDir  => move(srcDir,  dst, parseAbsDir,  "directory"),
-            srcFile => move(srcFile, dst, parseAbsFile, "file")))
+            srcDir  => move[S, Dir](srcDir,  dst, parseAbsDir,  "directory"),
+            srcFile => move[S, File](srcFile, dst, parseAbsFile, "file"))).run)
 
       case req @ POST -> AsDirPath(parent) => respond((for {
-        hdr <- EitherT.fromDisjunction[M.F](requiredHeader[S](XFileName, req))
+        hdr <- EitherT.fromDisjunction[M.F](requiredHeader(XFileName, req))
         fn  =  hdr.value
         dst <- EitherT.fromDisjunction[M.F](
                 (parseRelDir(fn) orElse parseRelFile(fn))
                   .flatMap(sandbox(currentDir, _))
                   .map(parent </> _)
-                  .toRightDisjunction(QResponse.error[S](BadRequest, s"Not a relative path: $fn")))
+                  .toRightDisjunction(ApiError.apiError(
+                    BadRequest withReason "Filename must be a relative path.",
+                    "fileName" := fn)))
         _   <- mount[S](dst, req, replaceIfExists = false)
       } yield s"added ${printPath(dst)}").run)
 
@@ -69,17 +73,21 @@ object mount {
 
   ////
 
-  private def move[S[_]: Functor, F[_], T](
+  private def move[S[_]: Functor, T](
     src: AbsPath[T],
     dstStr: String,
     parse: String => Option[Path[Abs, T, Unsandboxed]],
     typeStr: String
   )(implicit
     M: Mounting.Ops[S]
-  ): Free[S, QResponse[F]] =
+  ): EitherT[Free[S, ?], ApiError, String] =
     parse(dstStr).map(sandboxAbs).cata(dst =>
-      respond(M.remount[T](src, dst).as(s"moved ${printPath(src)} to $dstStr").run),
-      QResponse.error(BadRequest, s"Not an absolute $typeStr path: $dstStr").point[M.F])
+      M.remount[T](src, dst)
+        .as(s"moved ${printPath(src)} to $dstStr")
+        .leftMap(_.toApiError),
+      EitherT.leftU[String](ApiError.apiError(
+        BadRequest withReason s"Expected an absolute $typeStr.",
+        "path" := dstStr).point[Free[S, ?]]))
 
   private def mount[S[_]: Functor](
     path: APath,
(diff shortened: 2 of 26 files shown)


core/src/main/scala/quasar/types.scala
```diff
@@ -202,6 +205,57 @@ trait TypeInstances {
 
   implicit val TypeRenderTree: RenderTree[Type] =
     RenderTree.fromToString[Type]("Type")
+
+  implicit val typeEncodeJson: EncodeJson[Type] = {
+    def singleton(key: String) = Json(key -> jNull)
```
mossprescott: Why not just `jString(key)`?

core/src/main/scala/quasar/types.scala
```diff
@@ -202,6 +205,57 @@ trait TypeInstances {
...
+    EncodeJson {
+      case Top =>
+        singleton("$top")
```
mossprescott: I don't think there's any need for the `$`s here.

core/src/main/scala/quasar/types.scala
```diff
@@ -202,6 +205,57 @@ trait TypeInstances {
...
+          jEmptyObject
+        Json("$flexarr" -> flexarr)
+      case Obj(assocs, unkns) =>
+        val obj =
+          ("associations" :=  assocs) ->:
+          ("unknownKeys"  :?= unkns)  ->?:
+          jEmptyObject
+        Json("$obj" -> obj)
+      case Coproduct(l, r) =>
+        Json("$coproduct" -> Json("left" := l, "right" := r))
```
mossprescott: Probably would be better to do use `flatten` and an array instead of left/right. The typical case is something like `Int | Dec | Timestamp | Interval | ...`.

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -30,30 +30,34 @@ import scalaz._, Scalaz._
...
 
   def service[S[_]: Functor](implicit M: Mounting.Ops[S], S0: Task :<: S): QHttpService[S] =
     QHttpService {
       case GET -> AsPath(path) =>
-        def err = s"There is no mount point at ${printPath(path)}"
-        respond(M.lookup(path).toRight(QResponse.error[S](NotFound, err)).run)
+        def err = ApiError.fromMsg(
+          NotFound withReason "Mount point not found.",
+          s"There is no mount point at ${printPath(path)}",
+          "path" := path)
```
mossprescott: @jedesah: does this seem right, in the light of your recent changes?

What does the `EncodeJson` instance for `Path` do? Is it consistent with `posixCodec.printPath`, which appears in `message` here? Is that what we want?

web/src/main/scala/quasar/api/services/query/execute.scala
```diff
@@ -38,56 +38,53 @@ object execute {
...
+    val Q = QueryFile.Ops[S]
 
     val removePhaseResults = new (FileSystemErrT[PhaseResultT[Free[S,?], ?], ?] ~> FileSystemErrT[Free[S,?], ?]) {
       def apply[A](t: FileSystemErrT[PhaseResultT[Free[S,?], ?], A]): FileSystemErrT[Free[S,?], A] =
         EitherT[Free[S,?],FileSystemError,A](t.run.value)
     }
 
-    val Q = QueryFile.Ops[S]
+    val destinationFile: String => ApiError \/ (Path[Abs,File,Unsandboxed] \/ Path[Rel,File,Unsandboxed]) = {
+      val err = -\/(ApiError.fromStatus(BadRequest withReason "Destination must be a file."))
```
mossprescott: It seems the the provided path be included here, for consistency.

web/src/test/scala/quasar/api/matchers.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  def beDetailedError(status: String, details: (JsonAssoc \/ JsonField)*): Matcher[Json] =
+    haveErrorStatus(status) and haveErrorDetail(details: _*)
+
+  def beMsgError(status: String): Matcher[Json] =
+    beDetailedError(status, \/.right("message"))
+
+  def bePathNotFoundError(path: APath): Matcher[Json] =
+    beDetailedError("Path not found.", \/.left("path" := path))
```
mossprescott: Not sure I like this style. You're duplicating the implementation's code in that you build the expected value with `"path" := path`. That means, for one thing, even if the `EncodeJson[Path]` is changed in a way that changes the result here, these tests continue to pass. It also obfuscates the actual expected result, making the tests less useful as documentation. On the other hand, it saves duplication in the tests, which is certainly an understandable goal.

I guess I would prefer the make the tests more explicit, even at the cost of some repetition. In this case, that might mean no `bePathNotFound` matcher, instead inlining it in each test, and probably using `posixCodec.printPath` instead of the implicit encoder.

Thoughts?

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -200,7 +203,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
                   uri = Uri(path = printPath(path)).+?("offset", offset.shows).+?("limit", limit.shows))
                 val response = service(InMemState.empty)(request).run
                 response.status must_== Status.BadRequest
-                response.as[Json].run must_== Json("error" := s"invalid limit: $limit (must be >= 1)")
+                response.as[Json].run must beMsgError("Invalid query parameter.")
```
mossprescott: Did we lose some of the detail in the error message here, or just not testing for it?

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -208,7 +211,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
                 uri = Uri(path = printPath(path)).+?("offset", offset.shows).+?("limit", limit.shows))
               val response = service(InMemState.empty)(request).run
               response.status must_== Status.BadRequest
-              response.as[Json].run must_== Json("error" := s"invalid offset: $offset (must be >= 0)")
+              response.as[Json].run must beMsgError("Invalid query parameter.")
```
mossprescott: Similar.

web/src/main/scala/quasar/api/services/mount.scala
jedesah: Looks like it's fine: https://github.com/quasar-analytics/quasar/blob/master/core/src/main/scala/quasar/fp/PathyCodecJson.scala#L28

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -234,7 +237,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
                 uri = Uri(path = printPath(path)).+?("limit", "a"))
               val response = service(InMemState.empty)(request).run
               response.status must_== Status.BadRequest
-              response.as[Json].run must_== Json("error" := s"""invalid limit: Query decoding Long failed (For input string: "a")""")
+              response.as[Json].run must beMsgError("Invalid query parameter.")
```
mossprescott: Similar.

web/src/main/scala/quasar/api/services/mount.scala
jedesah: Ya, I think the `posixCodec` is a reasonable codec to use when printing paths for message bodies. I did not change that in my PR.

mossprescott: And I guess the posixCodec encoding is what we settled on for headers and messages?

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -563,10 +583,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
           to = fs.dir,
           state = fs.state,
           status = Status.Conflict,
-          body = (json: Json) => json must beOneOf(
-            fs.filesInDir.list.map { case (p, _) =>
-              Json("error" := s"${printPath(fs.dir </> p.path)} already exists")
-            }: _*),
+          body = (_: Json) must haveErrorStatus("Path exists."),
```
mossprescott: This is not testing for the presence/absence of the `path` field, right? Why not?

web/src/main/scala/quasar/api/services/mount.scala
jedesah: Ya, seems like a reasonable default and can't think of a better alternative.

mossprescott:
Other than a couple of minor comments, the implementation seems fine.

A few general comment about the test approach:

1) I think it's preferable to have the tests stand alone as much as possible. That is, they shouldn't depend on the same code that the implementation does, because it tends to make them harder to read as a specification of what _should_ happen (and more of a mere restatement of what currently _does_ happen), and it makes them less of an independent verification. Of course, it makes sense to leverage code we are explicitly not testing (e.g. http4s) to keep the tests concise.

2) Something similar applies to factoring shared code out of individual tests. If assertions are pulled out into shared code, it can obscure the behavior being verified more than it benefits by removing duplication. I have a slightly higher tolerance for duplication in tests, if by making assertions explicit in the test you get better clarity about what the behavior actually is.

3) I think these tests should verify the presence of all expected field and the absence of any others. I believe they did previously (by simply comparing to an expected Json value in most cases), but now some of them seem to only verify the main message, while other look for a subset of fields, but don't fail if anything else is present. Or maybe I'm mistaken and that just needs clarification in the matchers.

jedesah:
@mossprescott @wemrysi Regarding remark number 3. I agree and that is why In #1145 I added a test dependency on `rapture-json` for exactly that reason. `rapture-json` allows to pattern match on json like [this](https://github.com/quasar-analytics/quasar/pull/1145/files#diff-381378efb2f35ba0d44d8a6bc766ed61R112) which makes it much more straighforward to check for the exact "structure" of the json without resorting to straight up equality comparison because that would be annoying in the linked example, not too mention that it would probably make the test unecessarily fragile.

As a totally selfish request, this PR is going to have a fair amount of conflicts with #1145 and I was wondering if we could wait to merge that one before merging this one, especially since that would allow using the rapture stuff to address the above remark. I feel like my PR consists of more mechanical changes and it would be easier to merge this on top of that one than the opposite. I could be wrong, I don't mean to be pushing work on @wemrysi. I am happy to pair on the merge whichever one gets in first if it turns out to be non-trivial.

@sellout Can you hurry up and review #1145 pretty pretty please? :smile:

jdegoes:
Needs a changelog. Really looking forward to this! :tada:

core/src/main/scala/quasar/types.scala
wemrysi: I was taking cues from how we encode `Data` as JSON. I think just encoding the these singleton types as strings fails to convey their singleton-ness not to mention their membership in a sum type.

core/src/main/scala/quasar/types.scala
wemrysi: Similar to above, I was basing on prior art from encoding `Data`, I sort of like the distinction between the cases in the sum type vs normal fields, but am fine with omitting the `$` prefix if preferred.

core/src/main/scala/quasar/types.scala
wemrysi: Yeah, that does seem better. Any reason (historical or otherwise) that `flatten` isn't defined for `Type` itself?

web/src/test/scala/quasar/api/matchers.scala
wemrysi: I see your point, though I read the assertion this matcher makes as "The error should have this particular status message and a `path` member containing the JSON representation of the given `APath`, whatever that may be."

One method presupposes how we convert to JSON the other presupposes what the actual encoding is. I'm not sure which one is better. If we change the JSON representation for `Path`, is it desirable to have to change every test that expects a `JSON` representation of a path, or just a few tests that assert what said JSON representation is expected to be?

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
wemrysi: This asserts that the error has the given status and that the `message` field is present in the detail object (which contains the previous detail tested against), but doesn't validate the content of the message itself. This seemed like a reasonable compromise vs. exact matching the detailed message, which isn't guaranteed to remain constant across non-breaking version changes like the structure of the error is.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
wemrysi: You're correct, oversight on my part.

web/src/main/scala/quasar/api/services/query/execute.scala
wemrysi: Aye, will add it.

wemrysi:
@jdegoes Note the addition of `CHANGELOG/major.md` ;)

@mossprescott Regarding your three points:

1) I agree with you, though I'm conflicted as to how far to extend the notion of "the same code the implementation does". Using some of the abstractions keeps the tests focused and reduces overlap between them, but maybe that isn't necessarily a goal, or even a good idea?

2) I think de-duplicating test code can be useful if it helps to reduce noise around the actual bit being tested, but I also agree that if taken too far it can make tests hard to understand locally and am also fine with some duplication if it increases clarity/comprehension/readability.

3) You're correct that the tests don't check for the absence of extraneous fields, I'll update the matcher to do so.

@jedesah I think the rapture matching could be an alternative to some of what I have, but I think I can easily check what @mossprescott suggested without it.

As for merge order, I'm fine with sequencing this after #1145 if you think the other direction will be more arduous.

jedesah:
@wemrysi #1145 just got merged. I don't know if the label is up to date on this one, but thx for waiting (or being willing to wait). Unsurprisingly, looks like there are some conflicts, let me know if you think it would be more efficient to tackle the merge conflicts by pairing on them.

wemrysi:
@jedesah No problem, I'll let you know if I run into trouble.

wemrysi:
@mossprescott I've incorporated all the changes we discussed, mind giving this another look?

core/src/main/scala/quasar/types.scala
```diff
@@ -202,6 +205,55 @@ trait TypeInstances {
...
+        Json("Flexarr" -> flexarr)
+      case Obj(assocs, unkns) =>
+        val obj =
+          ("associations" :=  assocs) ->:
+          ("unknownKeys"  :?= unkns)  ->?:
+          jEmptyObject
+        Json("Obj" -> obj)
+      case cp @ Coproduct(l, r) =>
+        Json("Coproduct" := cp.flatten)
+    }
```
mossprescott: This looks good to me, but I'd still like to see some kind of test coverage. I'd be happy with just tests of the encoding of the non-"singleton" types and one or two examples of the simple ones.

mossprescott:
No other comments. Looks good!

core/src/main/scala/quasar/types.scala
wemrysi: Will do, thanks for the reminder.

wemrysi:
@mossprescott I've added encoder tests and rebased onto master if you want to have a final look.

mossprescott:
👍

PR #1152, opened 2016-03-29
[SD-1503] Improved disambiguation of projections in output

jedesah:
- Regression tests now support declaring a dependancy on more than one data source

### DIFF
--- a/core/src/main/scala/quasar/sql/package.scala
+++ b/core/src/main/scala/quasar/sql/package.scala
@@ -43,15 +43,15 @@ package object sql {
   def projectionNames[T[_[_]]:Recursive](projections: List[Proj[T[ExprF]]], relName: Option[String]):
       SemanticError \/ List[(String, T[ExprF])] = {
     def extractName(expr: T[ExprF]): Option[String] = expr.project match {
-      case IdentF(name) if Some(name) != relName      => Some(name)
+      case IdentF(name) if Some(name) != relName => Some(name)
       // TODO[matryoshka]: Change to Embed(StringLiteralF(name)
-      case BinopF(_, stringLiteral, FieldDeref) => stringLiteral.project match {
+      case BinopF(_, stringLiteral, FieldDeref)  => stringLiteral.project match {
         case string: ExprF.StringLiteralF[_] => Some(string.v)
-        case _ => None
+        case _                               => None
       }
-      case UnopF(expr, FlattenMapValues)              => extractName(expr)
-      case UnopF(expr, FlattenArrayValues)            => extractName(expr)
-      case _                                         => None
+      case UnopF(expr, FlattenMapValues)         => extractName(expr)
+      case UnopF(expr, FlattenArrayValues)       => extractName(expr)
+      case _                                     => None
     }
 
     val aliases = projections.flatMap{ case Proj(expr, alias) => alias.toList}
--- a/core/src/main/scala/quasar/sql/package.scala
+++ b/core/src/main/scala/quasar/sql/package.scala
@@ -40,23 +40,34 @@ package object sql {
   def CrossRelation(left: SqlRelation[Expr], right: SqlRelation[Expr]) =
     JoinRelation(left, right, InnerJoin, BoolLiteral(true))
 
-  def namedProjections(e: Expr, relName: Option[String]):
-      List[(String, Expr)] = {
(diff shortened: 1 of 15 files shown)


sellout:
Our eventual goal is to get rid of the `data` field, and instead pull the relation names from the SQL. Accepting arrays is a step in the right direction, but just mentioning in case you want to take it any further 😉

core/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -3447,8 +3447,17 @@ class PlannerSpec extends Specification with ScalaCheck with CompilerHelpers wit
...
+    * @param q A selection query
+    * @return The list of expected names for the projections of the selection
+    * @throws AssertionError If the `Query` is not a selection
+    */
   def columnNames(q: Query): List[String] =
-    sql.parse(q).foldMap(sql.namedProjections(_, None).map(_._1))
+    sql.parse(q).toOption.get.project match {
+      case select: sql.ExprF.SelectF[sql.Expr] =>
+        sql.projectionNames(select.projections, None).toOption.get.map(_._1)
+      case _ => throw new java.lang.AssertionError("Query was expected to be a selection")
```
jedesah: You mean change the return type of the function?

The behavior hasn't changed here. It would have thrown before, it's still throwing now. I think that is reasonable since if that case gets hit, it's a the fault of the person writing the test as opposed to a normal test failure.

What I would really like to do is accept a `SelectF`, but I think that would not be convenient right now because we would need a macro string interpolator for our sql AST in order to not make that overly inconvenient for the caller.

drostron: Yes, the return type. I'd prefer avoiding methods that throw even in test code — especially when it's easy to return a `\/`. That way the tests consuming the method can make the choice of how to handle it without needing exception handling code.

jedesah: They don't need to handle the exceptions because it should never throw unless they did not properly read the documentation. On the other hand, if we make the return type `\/`, it needs to be handled everywhere.

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -42,23 +42,34 @@ package object sql {
...
-  def namedProjections(e: Expr, relName: Option[String]):
-      List[(String, Expr)] = {
-    def extractName(expr: Expr): Option[String] = expr match {
-      case Ident(name) if Some(name) != relName      => Some(name)
-      case Binop(_, StringLiteral(name), FieldDeref) => Some(name)
-      case Unop(expr, FlattenMapValues)              => extractName(expr)
-      case Unop(expr, FlattenArrayValues)            => extractName(expr)
+  def projectionNames[T[_[_]]:Recursive](projections: List[Proj[T[ExprF]]], relName: Option[String]):
+      SemanticError \/ List[(String, T[ExprF])] = {
+    def extractName(expr: T[ExprF]): Option[String] = expr.project match {
```
drostron: Format Nit (_I hate these — if only there was a reasonable auto-formatter._). We should either line all `=>`s up or none.

drostron: Meaning, I almost prefer to not mention them.

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -42,23 +42,34 @@ package object sql {
...
+    val aliases = projections.flatMap{ case Proj(expr, alias) => alias.toList}
+
+    (aliases diff aliases.distinct).headOption.cata(
+      duplicateAlias => SemanticError.DuplicateAlias(duplicateAlias).left,
+      projections.zipWithIndex.mapAccumLeft1(aliases.toSet) { case (used, (Proj(expr, alias), index)) =>
+        alias.cata(
+          a => (used, a -> expr),
+          {
+            val tentativeName = extractName(expr) getOrElse index.toString
+            val alternatives = Stream.from(0).map(suffix => tentativeName + suffix.toString)
```
drostron: `Stream` :smiley:

drostron:
👍🏾

jedesah:
@drostron Had to do some non-trivial (but minor) changes in order to rebase over some of @mossprescott's changes in case you want to have a quick look again before I merge once the build succeeds.

drostron:
LGTM.

PR #1155, opened 2016-03-30
Implement conversions to/from String for types with syntax.

sellout:
SD-840 #done

Also,
* Use strings for `NumberLong` and `NumberInt` in JS;
* extract ExprOp typechecks into self-contained functions;
* fix some pattern matches in Expr => JS typecheck conversion;
* remove useless type parameters from `jsExprƒ` and `selectorƒ`;
* remove broken `Reduction` impls from `jsExprƒ`;
* have workflowƒ fall back to jsExprƒ (which allows us to eliminate some
  duplicated JS impls);
* delete `SemanticError.ExpectedLiteral`, which can no longer happen; and
* delete EnvT.scala, which should have disappeared when Matryoshka moved.

However, why do I have to call `ToAlgebraOps` explicitly on `jsExprƒ` now?


jdegoes:
:+1: I assume the names of these functions are patterned after Postgres, like usual?

sellout:
So … not quite. There’s a couple reasons. I named them to match the other “constructors”, so `boolean("true")` and `integer("-2132")` work like `oid("something")` and `timestamp("1234-01-02T12:34:56")`, but none of them are restricted to literals anymore. The Postgres functions are also a bit more complicated in terms of what they do, e.g., `to_number('12,454.8-', '99G999D9S')` does … something. And `to_char` is named `to_string` in our code because char is a different type for us.

jdegoes:
:+1: Need all this in the changelog and ultimately propagated to quasar docs. /cc @damonLL

sellout:
Argh – also need to skip some tests on read-only instances.

core/src/main/scala/quasar/types.scala
```diff
@@ -204,7 +204,7 @@ trait TypeInstances {
     RenderTree.fromToString[Type]("Type")
 }
 
-final case object Type extends TypeInstances {
+final object Type extends TypeInstances {
```
mossprescott: `final object`? Surely just `object` is what you want.

sellout: Yeah, true. Should have deleted more keywords 😃

core/src/main/scala/quasar/types.scala
```diff
@@ -430,4 +430,30 @@ final case object Type extends TypeInstances {
   val Numeric = Int ⨿ Dec
   val Temporal = Timestamp ⨿ Date ⨿ Time
   val Comparable = Numeric ⨿ Interval ⨿ Str ⨿ Temporal ⨿ Bool
+  val Syntaxed = Type.Null ⨿ Type.Comparable
+
+  implicit val equal: Equal[Type] = Equal.equal((a, b) => (a, b) match {
+    case (Top,       Top)
+       | (Bottom,    Bottom) // maybe?
```
mossprescott: Seems like this needs to be true.

core/src/main/scala/quasar/types.scala
```diff
@@ -430,4 +430,30 @@ final case object Type extends TypeInstances {
   val Numeric = Int ⨿ Dec
   val Temporal = Timestamp ⨿ Date ⨿ Time
   val Comparable = Numeric ⨿ Interval ⨿ Str ⨿ Temporal ⨿ Bool
+  val Syntaxed = Type.Null ⨿ Type.Comparable
+
+  implicit val equal: Equal[Type] = Equal.equal((a, b) => (a, b) match {
+    case (Top,       Top)
```
mossprescott: Isn't this the same as simply `Equal.equal { case (Top, Top) ...`?

it/src/main/resources/tests/convertDates.test
```diff
@@ -0,0 +1,14 @@
...
+    "predicate": "containsExactly",
+    "expected": [
+        { "0": { "$timestamp": "2014-08-17T06:00:00Z" }, "1": "2014-08-17T06:00:00.000Z" },
+        { "0": { "$timestamp": "2014-08-18T07:00:00Z" }, "1": "2014-08-18T07:00:00.000Z" },
+        { "0": { "$timestamp": "2014-08-19T08:00:00Z" }, "1": "2014-08-19T08:00:00.000Z" },
+        { "0": { "$timestamp": "2014-08-20T09:00:00Z" }, "1": "2014-08-20T09:00:00.000Z" },
+        { "0": { "$timestamp": "2014-08-21T10:00:00Z" }, "1": "2014-08-21T10:00:00.000Z" },
+        { "0": { "$timestamp": "2014-08-22T11:00:00Z" }, "1": "2014-08-22T11:00:00.000Z" },
+        { "0": { "$timestamp": "2014-08-23T12:00:00Z" }, "1": "2014-08-23T12:00:00.000Z" }]
+}
```
mossprescott: I suggest creating a new sub-directory for these.

CHANGELOG/minor.md
```diff
@@ -0,0 +1,2 @@
+- add functions to parse strings into various
```
mossprescott: I think it's worth including the actual list of new functions here.

CHANGELOG/minor.md
```diff
@@ -0,0 +1,2 @@
...
+- add `toString` function to convert various values to strings
```
mossprescott: Should say `to_string`, because that's what users will see.

core/src/main/scala/quasar/std/string.scala
```diff
@@ -195,6 +195,99 @@ trait StringLib extends Library {
...
+        str.parseDouble.fold(
+          κ(failure(nel(InvalidStringCoercion(str, "a string containing an decimal number".left), Nil))),
+          i => success(Type.Const(Data.Dec(i))))
+      case Type.Str                    :: Nil => success(Type.Int)
+    },
+    untyper(x => ToString(List(x)).map(List(_))))
+
+  val Null = Mapping(
+    "null",
+    "Converts strings containing “null” into the null value.",
```
mossprescott: Huh? And what does it do with other values at runtime? I guess it's an error, because a non-"null" constant is a type error. No?

I don't think I understand what this could be useful for.

core/src/main/scala/quasar/std/string.scala
```diff
@@ -195,6 +195,99 @@ trait StringLib extends Library {
     },
     basicUntyper)
 
-  def functions = Concat :: Like :: Search :: Length :: Lower :: Upper :: Substring :: Nil
+  val Boolean = Mapping(
+    "boolean",
+    "Converts the strings “true” and “false” into boolean values.",
```
mossprescott: I think it would be helpful to mention the behavior when the argument isn't matched (can't be parsed as an int, etc.), for each of these conversions.

core/src/main/scala/quasar/std/string.scala
```diff
@@ -195,6 +195,99 @@ trait StringLib extends Library {
...
+    partialTyperV {
+      case Type.Const(Data.Str("null"))  :: Nil => success(Type.Const(Data.Null))
+      case Type.Const(Data.Str(str))     :: Nil =>
+        failure(nel(InvalidStringCoercion(str, List("null").right), Nil))
+      case Type.Str                      :: Nil => success(Type.Null)
+    },
+    untyper(x => ToString(List(x)).map(List(_))))
+
+  val ToString: Func = Mapping(
+    "to_string",
```
mossprescott: Is there some reason this can't be called `string`? I think that would be a lot more obvious when all the rest of the conversions are named that way.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -999,6 +984,13 @@ object MongoDbPlanner extends Planner[Crystallized] {
                   // TODO: in Mongo 3.0, we can have a tighter type check.
                   // $lt(expr, $literal(Bson.Timestamp(Instant.ofEpochMilli(0), 0)))))
                   $lt(expr, $literal(Bson.Regex("", "")))))
+            case Type.Syntaxed =>
+              ((expr: Expression) =>
+                $or(
+                  $lt(expr, $literal(Bson.Doc(ListMap()))),
+                  $and(
+                    $lte($literal(Bson.ObjectId(Array[Byte](0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0))), expr),
```
mossprescott: Can you somehow use `minOid` here?

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -194,6 +179,25 @@ object MongoDbPlanner extends Planner[Crystallized] {
                   If(insen, Literal(Js.Str("im")), Literal(Js.Str("m"))))),
                 "test"),
               List(field)))
+        case Null => Arity1(str => If(BinOp(jscore.Eq, str, Literal(Js.Str("null"))), Literal(Js.Null), ident("undefined")))
+        case Boolean => Arity1(str => If(BinOp(jscore.Eq, str, Literal(Js.Str("true"))), Literal(Js.Bool(true)), If(BinOp(jscore.Eq, str, Literal(Js.Str("false"))), Literal(Js.Bool(false)), ident("undefined"))))
+        case Integer => Arity1(str => Call(ident("NumberLong"), List(Call(ident("parseInt"), List(str, Literal(Js.Num(10, false)))))))
```
mossprescott: I don't think you want to use `parseInt` here.

Quick experiment:
```
> parseInt("123")
123
> parseInt("123a")
123
> parseInt("a123")
NaN
```

That doesn't really strike me as what we want, and anyway it works only up to 53 bits.

Why not filter with a regex and then pass the string to `NumberLong` directly?

sellout: That’s a good idea. I didn’t realize you _could_ hand a String to NumberLong until our conversation yesterday.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -194,6 +179,25 @@ object MongoDbPlanner extends Planner[Crystallized] {
...
+        case Decimal | Interval =>
+          Arity1(str => Call(ident("parseFloat"), List(str)))
```
mossprescott: Similar issue here with `parseFloat`, but probably no way around using it. Maybe still worth pre-filtering.

core/src/main/scala/quasar/physical/mongodb/planner.scala
sellout: Yep.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -194,6 +179,25 @@ object MongoDbPlanner extends Planner[Crystallized] {
...
+        case Timestamp => Arity1(str => New(Name("Date"), List(str)))
```
mossprescott:
```
> Date("foo")
Thu Mar 31 2016 10:32:50 GMT-0700 (PDT)
```

Hmm.

mossprescott: Even better:
```
> new Date("foo")
ISODate("0NaN-NaN-NaNTNaN:NaN:NaNZ")
```

core/src/main/scala/quasar/types.scala
sellout: Does that work? `Equal.equal` takes a function of two parameters, not a tuple, so I thought I had to tuple them explicitly to match like that.

mossprescott: Right you are. Silly language.

core/src/main/scala/quasar/std/string.scala
sellout: Mostly just for consistency – I wanted symmetry between these constructors and `to_string`. It behaves the same as the other constructors at runtime – if the argument isn’t a valid value, the result is Undefined.

core/src/main/scala/quasar/std/string.scala
sellout: No reason it can’t, but all these “typename” functions construct a value from a String, so I’d expect `string(…)` to basically be a `String ⇒ String` identity. Whereas `to_string` converts any^Mmost types to a string.

core/src/main/scala/quasar/types.scala
sellout: Yeah, the distinction between tuples and multple parameters was a real pain when generalizing some Matryoshka stuff from `(A, _)` (which I had just used multi-param functions for) to `W[_]: Comonad` (which meant they all had to be tupled).

core/src/main/scala/quasar/std/string.scala
mossprescott: Yes, maybe.

But that raises the question of whether these conversions should work on other types. For instance, `integer(pop/1000)` would be useful. Maybe also Int <-> Boolean, but I'm dubious.

I see we currently have `to_timestamp :: Int -> Timestamp`, so maybe `to_integer` is a better solution (or even `round` and/or `truncate`).

Ok, I've convinced myself. Nevermind.

sellout:
Ok, this thing finally passes everything. Sorry, I had to rebase along the way (but haven’t squashed). This is ready for final review.

mossprescott:
LGTM

PR #1156, opened 2016-03-31
Upgrade Matryoshka to v0.8.0.

sellout:
Significant changes are

- generalization of Elgot algebras;
- renaming `once` to `orOriginal`;
- `Embed.unapply`, which obsoletes `IsInvoke`; and
- remove `import Fix._`[†].

[†]: Removing this import requires that we add a bunch of type
annotations. But that’s helpful, because it makes it easier to
generalize over `T`, and it shows where we need to use kinda-curried
types in Matryoshka.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- upgrade to Matryoshka 0.8.0
--- a/core/src/main/scala/quasar/compiler.scala
+++ b/core/src/main/scala/quasar/compiler.scala
@@ -23,7 +23,7 @@ import quasar.sql._
 import quasar.std.StdLib._
 import quasar.SemanticAnalysis._, quasar.SemanticError._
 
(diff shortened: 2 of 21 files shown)


CHANGELOG/revision.md
```diff
@@ -0,0 +1 @@
+- refactoring
```
wemrysi: Mind being more specific?

wemrysi:
:+1: one minor comment regarding changelog message, but that's it.

CHANGELOG/revision.md
sellout: So “refactoring” may not be the best description, but the changelog is really for user-visible changes. So maybe “no visible changes” or something?

wemrysi: I don't really see the advantage of adding a content-free message vs one that actually describes what happened, even if an API user won't care. Someone depending on quasar as a library might care (especially if they're also depending on matryoshka and can now update their local dep. etc).

PR #1159, opened 2016-04-04
Add summary of MOVE for mounts to readme

garyb:


jedesah:
This looks good to me but assigning @mossprescott for a quick sanity check.

Technically, we need a changelog file in order to merge this. @sellout and I had a heated discussion about whether a change to the README warrants a changelog or not and I lost 😄 .

README.md
```diff
@@ -433,6 +433,10 @@ Creates a new mount point or replaces an existing mount point using the JSON con
 
 Deletes an existing mount point, if any exists at the given path. If no such mount exists, the request succeeds but the response has no content.
 
+### MOVE /mount/fs/[path]
+
+Moves a mount from one path to another. The new path must be provided in the `Destination` request header. This will return a 409 Conflict if a database mount is being moved under the path of an existing database mount.
```
mossprescott: It would be more precise to say "above or below an existing database mount."

mossprescott:
@garyb: Thanks for the contribution :)

And @jedesah's right, please add a file `CHANGELOG/revision.md` with a single line like:
```
- add missing documentation
```
That's what our merge script uses to figure out how to bump the version.

README.md
garyb: I saw that in the description of the other endpoints, but I wouldn't have thought you could move it above an existing mount, as it would have to already contain the mount that is being moved for that condition to happen?

garyb:
Ah ok! I'll add that once the 409 conflict description is confirmed.

README.md
mossprescott: I'm not sure we're speaking the same language.

Say you have db mounts at `/a/b/` and `/c/`. Attempting to move the second mount from `/c/` to `/a/` would put it _above_ `/a/b/`. Moving it to `/a/b/d/` would put it below. Either one is an error.

Does that make more sense?

garyb: Ok yep, that totally makes sense. I assumed it would fail some other way there actually (like the "already exists" error), but I see now that doesn't really apply to mounts, since the parent parts of the path don't really exist.

garyb:
How's this look? I assume we needed a new changelog file since 6.1.1 has been released?

mossprescott:
That wording looks good to me.

The changelog file actually has to be called, literally, `revision.md`. Our merge script figures out what the new version is based on that and whatever happens to be on master and bumps the actual version for the build. See https://github.com/quasar-analytics/quasar/wiki/Versioning if you're curious.

garyb:
I see, neat!

I think I've got it this time... at least I should know what I'm doing next time I attempt to contribute!

mossprescott:
@garyb: One more thing for the script... this branch needs to be rebased on latest master so the script can merge it with --ff-only. Sorry about all the fuss.

garyb:
No problem, it's rebased now.

mossprescott:
Merged as 7.0.3. Thanks again!

PR #1160, opened 2016-04-05
fix check for no backend configuration in regression test

mossprescott:
Probably has been broken for some time, but only recently encountered.


sellout:
👍🏾

mossprescott:
6.1.1

PR #1162, opened 2016-04-06
SD-1544 WriteFile#save should not succeed when the src process fails

wemrysi:
`WriteFile#save` would still end up writing the file when the source process halted due to an error, this fixes the function to correctly abort, cleanup and avoid writing the (possibly partial) file.

### DIFF
--- a/core/src/main/scala/quasar/fs/WriteFile.scala
+++ b/core/src/main/scala/quasar/fs/WriteFile.scala
@@ -65,12 +65,13 @@ object WriteFile {
       * exhausted.
       */
     def appendChannel(dst: AFile): Channel[M, Vector[Data], Vector[FileSystemError]] = {
+      def closeHandle(h: WriteHandle): Process[M, Nothing] =
+        Process.eval_[M, Unit](unsafe.close(h).liftM[FileSystemErrT])
+
       def writeChunk(h: WriteHandle): Vector[Data] => M[Vector[FileSystemError]] =
         xs => unsafe.write(h, xs).liftM[FileSystemErrT]
 
-      Process.bracket(unsafe.open(dst))(h => Process.eval_[M, Unit](unsafe.close(h).liftM[FileSystemErrT])) { h =>
-        channel.lift(writeChunk(h))
-      }
+      Process.bracket(unsafe.open(dst))(closeHandle)(h => channel.lift(writeChunk(h)))
     }
 
     /** Same as `append` but accepts chunked [[Data]]. */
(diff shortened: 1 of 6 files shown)


core/src/test/scala/quasar/fs/WriteFileSpec.scala
```diff
@@ -76,12 +76,18 @@ class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture
     }
 
     "append should fail after writing some when source fails" ! prop {
```
jedesah: Trying to understand the change in the implementation of the test made me realize that potentially the wording on this test name is not idea. Perhaps something along the lines of "append should produce a Stream that fails when the source stream fails, however, the data written so far should remain".

core/src/test/scala/quasar/fs/WriteFileSpec.scala
```diff
@@ -114,13 +120,19 @@ class WriteFileSpec extends Specification with ScalaCheck with FileSystemFixture
...
 
-      val p = write.save(f, xs.toProcess ++ Process.fail(new RuntimeException("source failed")))
-      MemTask.runLogEmpty(p).attemptRun must beLeftDisjunction
-    }.pendingUntilFixed("SD-1544")
+      val p = write.save(f, src).drain onFailure {
+        case err if err.getMessage == "SRCFAIL" => read.scanAll(f)
+        case err                                => Process.fail(err)
+      }
+
+      MemTask.runLogEmpty(p).run.toEither must beLeft(pathErr(pathNotFound(f)))
```
jedesah: I feel like this is not fully testing what is described.

If I understand correctly, this is testing that saving to a file is failing because the path is not found because we are interpreting it with an empty interpreter. Should we not be successfully writing to a destination up until the source fails and then make sure the original data is still there?

core/src/main/scala/quasar/fs/WriteFile.scala
```diff
@@ -215,12 +216,15 @@ object WriteFile {
       def cleanupTmp(tmp: AFile)(t: Throwable): Process[M, Nothing] =
         Process.eval_(MF.delete(tmp)).causedBy(Cause.Error(t))
 
-      MF.tempFile(dst).liftM[Process] flatMap (tmp =>
-        appendChunked(tmp, src).terminated.take(1)
+      MF.tempFile(dst).liftM[Process] flatMap { tmp =>
+        appendChunked(tmp, src)
+          .map(some).append(Process.emit(none))
```
jedesah: Not sure I understand this fix. So if the first process fails, then we will emit a `None` and ultimately proceed to move the file as if everything happened correctly?

jedesah: Nevermind, just reread the docs, and `append` and `complete` behave inversely to what I had initially understood. So I understand why this fixes the problem.

core/src/test/scala/quasar/fs/WriteFileSpec.scala
wemrysi: This demonstrates that when the source process fails after emitting some data, no file is actually written. This is witnessed by the "path not found" error which, before the bug was fixed, would instead be a result vector == `xs`.

wemrysi:
@jedesah Updated the description of the test you mentioned, any other comments?

core/src/test/scala/quasar/fs/WriteFileSpec.scala
jedesah: Ah, I understand better now. I think this test would be better if there was some arbitrary data already there and we made sure it was still there as from a black box perspective it seems believable that the operation would wipe existing data and then leave it empty if the source were to fail.

jedesah:
Just the response to your earlier comment.

wemrysi:
@jedesah Ok, I updated the `save` test as well.

jedesah:
LGTM!

PR #1163, opened 2016-04-06
SD-1239 Prevent changing to a non-existent directory in the repl

wemrysi:
Ensure the directory exists before changing to it the repl.


sellout:
👍🏾

PR #1164, opened 2016-04-06
Pull common test-related code out of `core`

mossprescott:
This allows `web` and `it` to depend on `core` and a couple of new, small subprojects, instead of `core`'s tests. This means slightly faster local builds and less hassle when things are in flux.

It will also allow other projects to make use of the Arbitrary instances and other utilities.

_This brings us in line with how scala-pathy is organized, for example, and makes building `web` and `it` a bit more sane._

_On the other hand, it involves moving all of the `core` test files, so will possibly cause some disruption._


core-scalacheck/src/main/scala/quasar/fs/PathArbitrary.scala
```diff
@@ -43,6 +44,7 @@ trait PathArbitrary {
     parentDir(dir).getOrElse(dir) </> rFile
   }
 
+  @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.OptionPartial"))
```
jedesah: Why did we need to add this?

jedesah:
What is the advantage of creating a `core-tests` project as opposed to leaving those tests in `core`?

jedesah:
Why is `ConfigSpec` in `core-specs2`?

core-scalacheck/src/main/scala/quasar/fs/PathArbitrary.scala
mossprescott: This code used to under `core/src/test`, which is apparently compiled without WartRemover. In the new structure, the shared code goes to `.../src/main` under each of the new projects.

mossprescott:
The tests moved to a separate subproject because they need to depend on `scalacheck` and `specs2`, which in turn depend on `core`. It may be possible to make only the `test` configuration in `core` depend on those two projects; I didn't try that because I noticed that pathy was broken down in this way and figured I'd be consistent with that.

`ConfigSpec` had to go to a shared location because it is extended in both `core-tests` and `web`'s tests (and perhaps also in quasar-advanced?) It's slightly awkward, but I think it's better than putting in one of the other projects or making yet another project for the one file.

One way to look at it is that `scalacheck` contains `Arbitrary`s and anything else that might be useful for somebody using some of the datatypes from `core`, whereas `specs2` contains stuff that's really about us writing tests for `core` and `web`.

jedesah:
I understand. I think we can probably come up with a better name for the `specs2` project. I might even argue that it could be folded into the `scalacheck` project and have that project be "test-utils" (find better name) as opposed to "test utils, but only for this one library scalacheck".

jedesah:
Especially considering there is a bunch of useful arbitraries in the `specs2` sub-project currently.

mossprescott:
You mean `AlphaCharacters` and `AlphaAndSpecialCharacters`? I would be fine with moving those to `scalacheck`. I do think there's probably _some_ value in keeping the Arbitrary's separate, but I agree that the `specs2` project is a bit of a catch-all and not all of it really relates to specs2. And I don't think `core-test-utils` is _that_ bad of a name.

How about I rename `core-specs2` to `core-test-utils`, move the string arbitraries to `core-scalacheck` somewhere, and call it a day?

jedesah:
Sorry for the late response, was gone to the gym. I'm okay with just renaming the `specs2` project. although I would be curious for you to elaborate on that "some" value that having a `scalacheck` only sub-project provides (not because I don't believe you, just not obvious to me!).

Also, I would be okay with yet another project for the actual specs2 related stuff in the current `specs2` sub-project because they seem to have nothing to do with `quasar`. Perhaps with a disclaimer suggesting we consider contributing back to `specs2` proper.

I appreciate that you want to get this merged as soon as possible, so feel free to just rename the `specs2` project, merge, and we can continue pointlessly splitting hairs after 😝

mossprescott:
Just noticed the coverage hit and will resolve that before merging.

mossprescott:
I didn't get anywhere with codecov, and I don't want to put more time into this now, so I'm closing the PR.

It won't be hard to repeat the exercise later on if we choose.

PR #1165, opened 2016-04-06
Upgrade to scalaz 7.2.x

jedesah:


jedesah:
@mossprescott Assigned you. Posting because I am not sure whether you get pinged or not because the PR was already opened in the "In Progress" state without any reviewer.

jedesah:
The build is failing because I need to publish new versions of both `matryoshka` and `pathy` that have been upgraded to the same version of scalaz.

Also, I want to make a change where I define an alternative `EntityDecoder[Json]` (as opposed to the default `http4s` one) which avoids the bug that seems to have been introduced in the latest version of `http4s`. This would be better than the few places currently where I changed `as[Json]` to `as[String]` and call the `argonaut` `Parse.parse` function "manually". But I figured I would put this in review since I can work on that while the rest gets reviewed (also because I thought about it as a better solution at the last minute).

jedesah:
Okay, just updated to recently released versions of `pathy` and `matryoshka`. Just waiting on release `0.13.2` of `http4s` which contains a [fix](https://github.com/http4s/http4s/issues/604) we need in order to merge this.

CHANGELOG/7.1.0.md
```diff
@@ -1,3 +0,0 @@
-- [SD-1524] Capture debug info for runtime errors during query exec
-  - Adds a new class of error describing failures during plan execution
-  - Ensures that runtime failures during MongoDB plan execution are trapped and returned as values.
```
mossprescott: Accidental deletion?

core/src/main/scala/quasar/fp/free/package.scala
```diff
@@ -122,6 +122,7 @@ package object free {
           case \/-(a) => a
           case -\/(t) => t.left[A].point[G]
         }).join
+      }
```
mossprescott: Leftover from debugging?

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -287,7 +287,7 @@ object LogicalPlan {
         args0 <- types.align(args).traverseU(_.onlyBoth match {
           case Some((t, arg)) => inferTypes(t, arg).disjunction
           case None =>
-            SemanticError.WrongArgumentCount(func, types.length, args.length)
+            (SemanticError.WrongArgumentCount(func, types.length, args.length): SemanticError)
```
mossprescott: I'd rather see a `wrongArgumentCount` constructor/prism than this, now that we're using that approach elsewhere.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -330,7 +330,7 @@ object LogicalPlan {
       emitName(freshName("check").map(name =>
         ConstrainedPlan(inf, List(NamedConstraint(name, inf, term)), Free(name))))
     }
-    else lift(SemanticError.GenericError(s"couldn’t unify inferred (${inf}) and possible (${poss}) types in $term").wrapNel.left)
+    else lift((SemanticError.GenericError(s"couldn’t unify inferred (${inf}) and possible (${poss}) types in $term"): SemanticError).wrapNel.left)
```
mossprescott: Similar.

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -97,7 +97,7 @@ package object sql {
 
   def pprint(sql: Expr) = sql.para(pprintƒ)
 
-  private val SimpleNamePattern = "[_a-zA-Z][_a-zA-Z0-9$]*".r
+  private val SimpleNamePattern = "[_a-zA-Z][_a-zA-Z0-9]*".r
```
mossprescott: How is this related to the rest?

core/src/main/scala/quasar/std/array.scala
```diff
@@ -29,7 +29,7 @@ trait ArrayLib extends Library {
     noSimplification,
     partialTyperV {
       case _ :: Type.Const(Data.Int(dim)) :: Nil if (dim < 1) =>
-        failure(nel(GenericError("array dimension out of range"), Nil))
+        failure(nel(GenericError("array dimension out of range"), IList.empty))
```
mossprescott: Why not `INil`, as was used elsewhere?

core/src/main/scala/quasar/std/library.scala
```diff
@@ -58,7 +58,7 @@ trait Library {
   private def partialUntyperOV(f: Type => Option[ValidationNel[SemanticError, List[Type]]]):
       Func.Untyper =
     (func, rez) => Type.typecheck(rez, func.codomain).fold(
-      e => f(rez).getOrElse(failure(e)),
+      e => f(rez).getOrElse(failure(e.map(ι[SemanticError]))),
```
mossprescott: This is pretty gross, and this isn't the only place it happened. I'd rather fix the underlying problem, which would probably look like having `TypeError` as a separate type (or just a standalone case class), and a wrapper extending SemanticError.

core/src/test/scala/quasar/std/math.scala
```diff
@@ -154,18 +154,18 @@ class MathSpec extends Specification with ScalaCheck with TypeArbitrary with Val
       expr must beFailing
     }
 
-    "typecheck number raised to 0th power" ! arbitraryNumeric { (t: Type) =>
+    "typecheck number raised to 0th power" ! prop { (t: Type) =>
       Power(t, TZero()) should beSuccessful(TOne())
-    }
+    }.setArbitrary(arbitraryNumeric)
```
mossprescott: I'm still not a fan of this switch, which makes it harder to tell what's being tested.

The old way reads to me as "for all numeric types...", whereas the new way says "for all types... just kidding only for numeric types."

core/src/test/scala/quasar/physical/mongodb/collection.scala
```diff
@@ -153,7 +154,7 @@ class CollectionSpec extends Specification with ScalaCheck with DisjunctionMatch
           err  => scala.sys.error(err.toString),
           coll => identicalPath(f.path, coll.asFile) must beTrue)
       }
-    }
+    }.set(maxDiscardRatio = 100.0f)
```
mossprescott: ?!?

it/src/test/scala/quasar/fs/ManageFilesSpec.scala
```diff
@@ -233,6 +233,6 @@ class ManageFilesSpec extends FileSystemTest[FileSystem](
       }
 
       step(deleteForManage(fs.setupInterpM).runVoid)
-    }; ()
+    }
```
mossprescott: :+1:

project/Dependencies.scala
```diff
@@ -5,59 +5,54 @@ import scala.collection.Seq
...
 
 object Dependencies {
-  private val scalazVersion  = "7.1.6"
+  private val scalazVersion  = "7.2.1"
   private val slcVersion     = "0.4"
   private val monocleVersion = "1.1.1"
-  private val pathyVersion   = "0.0.3"
-  private val http4sVersion  = "0.12.4"
+  private val pathyVersion   = "0.2.0"
+  private val http4sVersion  = "0.13.2a-SNAPSHOT"
```
mossprescott: Just leaving a marker here to remember to go to the released version.

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -235,10 +235,11 @@ object MessageFormat {
...
 
-  def fromAccept(accept: Option[Accept]): MessageFormat =
+  def fromAccept(accept: Option[Accept]): MessageFormat = {
     // TODO: MediaRange needs an Order instance – combining QValue ordering
     //       with specificity (EG, application/json sorts before
     //       application/* if they have the same q-value).
-    accept.flatMap(_.values.sortBy(_.qValue).list.map(fromMediaType).flatten(Option.option2Iterable).headOption)
+    accept.flatMap(_.values.sortBy(_.qValue).list.map(_.mediaRange).map(fromMediaType).flatten(Option.option2Iterable).lastOption)
       .getOrElse(MessageFormat.Default)
+  }
```
mossprescott: Another leftover {} pair.

web/src/main/scala/quasar/api/services/query/package.scala
```diff
@@ -38,6 +38,7 @@ package object query {
   object QueryParam extends QueryParamDecoderMatcher[Query]("q")
 
   def queryParam(params: Map[String, scala.collection.Seq[String]]): ApiError \/ Query =
+    // TODO[jr]: remove if
```
mossprescott: Leftover TODO.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -510,7 +510,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
             "srcPath" := fs.file,
             "dstPath" := dir),
           newState = Unchanged)
-      }
+      }.set(minTestsOk = 300)
```
mossprescott: Leftover from debugging?

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -86,7 +87,7 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s with Pa
       val service = mount.service[Eff].toHttpService(liftMT[Task, ResponseT] compose eff)
 
       f(service.run andThen (free.lift(_).into[Eff])).foldMap(eff)
-    }.run.run
+    }.unsafePerformSync.unsafePerformSync
```
mossprescott: That's surprising. Should/could this be `merge.unsafePerformSync`?

core/src/main/scala/quasar/physical/mongodb/execution.scala
```diff
@@ -63,7 +63,7 @@ private[mongodb] object execution {
             (by, field) match {
               case (\/-($var(DocField(origField @ BsonField.Name(_)))), (newField, \/-($var(DocField(IdName))))) =>
                 (origField, newField).some
-              case (-\/(Reshape(map)), (newField, \/-($var(DocField(BsonField.Path(NonEmptyList(IdName, x))))))) if map.size ≟ 1 =>
+              case (-\/(Reshape(map)), (newField, \/-($var(DocField(BsonField.Path(NonEmptyList(IdName, ICons(x, INil())))))))) if map.size ≟ 1 =>
```
mossprescott: Did they get rid of the varargs constructor? This is pretty ugly.

mossprescott:
Lots of little things and a chance to revisit the debate over how to specify non-implicit arbitraries.

Is there something I can look at to understand the library changes that happened? I see the following:
- MonadError and MonadState type parameters changed
- _something_ now requires `Monad` instead of maybe `Functor`.
- `IList` is being used in `NonEmptyList`?
- use of `IList` forcing us to be more precise about types in `Nel[SemanticError]`?
- Argonaut switched from `\/` to scala's `Either`?

Is that about it?

core/src/main/scala/quasar/effect/AtomicRef.scala
```diff
@@ -97,7 +97,7 @@ object AtomicRef {
       }
     }
 
-  def toState[F[_, _], S](implicit F: MonadState[F, S])
+  def toState[F[_, _], S](implicit F: MonadState[F[S,?], S])
```
wemrysi: I think this signature should change to `toState[F[_], S](implicit F: MonadState[F, S]): AtomicRef[S, ?] ~> F`, right?

core/src/main/scala/quasar/effect/Failure.scala
```diff
@@ -99,7 +99,7 @@ object Failure {
       }
     }
 
-  def toError[F[_, _], E](implicit F: MonadError[F, E]): Failure[E, ?] ~> F[E, ?] =
+  def toError[F[_, _], E](implicit F: MonadError[F[E,?], E]): Failure[E, ?] ~> F[E, ?] =
```
wemrysi: Change to `toError[F[_], E](implicit F: MonadError[F, E]): Failure[E, ?] ~> F`

core/src/main/scala/quasar/effect/KeyValueStore.scala
```diff
@@ -158,7 +158,7 @@ object KeyValueStore {
       new Aux[F]
 
     final class Aux[F[_, _]] {
-      def apply[K, V, S](l: Lens[S, Map[K, V]])(implicit F: MonadState[F, S])
+      def apply[K, V, S](l: Lens[S, Map[K, V]])(implicit F: MonadState[F[S,?], S])
```
wemrysi: Similarly, `Aux[F[_]]` etc.

core/src/main/scala/quasar/effect/MonotonicSeq.scala
```diff
@@ -75,7 +75,7 @@ object MonotonicSeq {
       new Aux[F]
 
     final class Aux[F[_, _]] {
-      def apply[S](l: Lens[S, Long])(implicit F: MonadState[F, S])
+      def apply[S](l: Lens[S, Long])(implicit F: MonadState[F[S,?], S])
```
wemrysi: Change to `Aux[F[_]]`

core/src/main/scala/quasar/fs/mount/Mounter.scala
```diff
@@ -44,9 +44,10 @@ object Mounter {
     type FreeS[A]  = Free[S, A]
     type MntE[A]   = MntErrT[FreeS, A]
     type Err[E, A] = EitherT[FreeS, E, A]
+    type ErrM[A] = Err[MountingError,A]
```
wemrysi: Looks like `Err` could just become `Err[A]`, as there is no need for the 2-arg version anymore.

wemrysi: Actually, it already exists as `MntE`, so you should be able to use that instead and get rid of `Err` and `ErrM`.

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -113,6 +113,7 @@ object Mounting {
     ////
 
     private type ErrF[A, B] = EitherT[F, A, B]
+    private type ErrFM[A] = ErrF[MountingError \/ PathTypeMismatch, A]
```
wemrysi: No need for `ErrF` and `ErrFM` either.

core/src/main/scala/quasar/fs/mount/hierarchical.scala
```diff
@@ -135,6 +135,7 @@ object hierarchical {
 
     type M[A] = Free[S, A]
     type ME[A, B] = EitherT[M, A, B]
+    type MES[A] = ME[FileSystemError, A]
```
wemrysi: Similarly, no need for both `ME` and `MES`.

core/src/main/scala/quasar/fs/mount/hierarchical.scala
```diff
@@ -352,7 +353,7 @@ object hierarchical {
         // Documentation on `QueryFile` guarantees absolute paths, so calling `mkAbsolute`
         case ReadF(p) => mountFor(mkAbsolute(rootDir, p))
         case _        => ().point[M]
-      }.run.run(initMnt) match {
+      }(implicitly,Monad.monadMTMAB).run.run(initMnt) match {
```
wemrysi: Ew.

core/src/main/scala/quasar/physical/mongodb/WorkflowExecutor.scala
```diff
@@ -180,7 +180,7 @@ private[mongodb] abstract class WorkflowExecutor[F[_]: Monad, C] {
     val tempDst: N[Collection] =
       tempDbName cata (
         tempColl,
-        noDatabase(()).raiseError[E, Collection].liftM[TempsT])
+        noDatabase(()).raiseError[M, Collection].liftM[TempsT])
```
wemrysi: I think we can get of the, now unused, `E` aliases.

core/src/main/scala/quasar/physical/mongodb/fs/fsops.scala
```diff
@@ -43,7 +41,7 @@ object fsops {
       cs     <- MongoDbIO.collectionsIn(dbName)
                   .filter(_.collectionName startsWith cName)
                   .runLog.map(_.toVector).liftM[FileSystemErrT]
-      _      <- if (cs.isEmpty) pathErr(pathNotFound(dir)).raiseError[MongoE, Unit]
+      _      <- if (cs.isEmpty) pathErr(pathNotFound(dir)).raiseError[MongoFsM, Unit]
```
wemrysi: Can delete `MongoE` now.

core/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
```diff
@@ -190,11 +190,11 @@ object managefile {
     collFromPathM(file) flatMap (c =>
       collectionExists(c).liftM[FileSystemErrT].ifM(
         dropCollection(c).liftM[FileSystemErrT],
-        pathErr(pathNotFound(file)).raiseError[MongoE, Unit]))
+        pathErr(pathNotFound(file)).raiseError[MongoFsM, Unit]))
 
   private def freshName: MongoManage[String] =
     for {
-      in <- MonadReader[R, ManageIn].ask
+      in <- MonadReader[R[ManageIn,?], ManageIn].ask
```
wemrysi: Can delete `R` and replace with `MongoManage`.

core/src/main/scala/quasar/physical/mongodb/fs/queryfile.scala
```diff
@@ -166,7 +166,7 @@ private final class QueryFileInterpreter[C](
     WorkflowExecErrT[ReaderT[SeqNameGeneratorT[JavaScriptLog,?],String,?], A]
 
   private val queryR =
-    MonadReader[R, (Option[DefaultDb], TaskRef[EvalState[C]])]
+    MonadReader[R[(Option[DefaultDb], TaskRef[EvalState[C]]), ?], (Option[DefaultDb], TaskRef[EvalState[C]])]
```
wemrysi: Should be able to get rid of `W` and `R` aliases and replace with `MongoLogWF` and `MQ`, respectively.

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
```diff
@@ -54,6 +54,7 @@ class FileSystemMounterSpec extends mutable.Specification {
     new (EffM ~> EffR) {
       type MT[F[_], A] = EitherT[F, String, A]
       type M[A]        = MT[ResMntsS, A]
+      import EitherT.eitherTMonad
```
wemrysi: This a leftover? Or actually necessary?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -102,7 +101,7 @@ abstract class QueryRegressionTest[S[_]: Functor](
     test: RegressionTest,
     backendName: BackendName,
     run: Run
-  ): Example = {
+  ) = {
```
wemrysi: Why delete the return type?

core/src/main/scala/quasar/effect/AtomicRef.scala
jedesah: Sure, I can try it, I mean I don't really care either way, I am not 100% clear on why this change was made in `scalaz`. It seems to me like the `MonadState` shape is not harder to work with, so I was thinking it might make sense the clients of this API from this change.

wemrysi: The new shape is actually nicer to work with when you consider transformer stacks, previously you'd have to quotient out the error/state/environement type from a transformer stack, now you don't have to.

jedesah: Fair enough, but in other circumstances, it's kind of awkward and those are the ones I seemed to hit upgrading in `quasar`

core/src/main/scala/quasar/fs/mount/Mounting.scala
jedesah: I can get rid of `ErrF` but not `ErrFM`.

core/src/main/scala/quasar/fs/mount/hierarchical.scala
jedesah: Same here. Not clear on how I can remove `MES`, but I can remove `ME`.

core/src/main/scala/quasar/fs/mount/hierarchical.scala
jedesah: Do you have a better solution?

I could probably create an implicit val within the local scope and that should make scala choose that one without complaining about ambiguous implicits.

jedesah: But not sure, that is necessarily so much better.

core/src/main/scala/quasar/logicalplan.scala
jedesah: Just want to say that I disagree with dumbing down the Scala type system to simulate ADTs, but this is not the place to open that can of worms so I will create the prisms.

sellout: I don’t think it’s dumbing down – just a tradeoff. Much like how Scala defaults to eager and you can be explicitly lazy; Scala defaults to, uh … GADTs-ish, but you can explicitly choose inference over that. Haskell just chooses the opposite default in both cases.

core/src/main/scala/quasar/fs/mount/hierarchical.scala
wemrysi: Yeah, that's what I meant, sorry for not being more clear.

core/src/main/scala/quasar/fs/mount/hierarchical.scala
wemrysi: Yeah, wasn't a comment on your solution, just that it is necessary at all =(

core/src/main/scala/quasar/logicalplan.scala
jedesah: To summarize the offline discussion at the office. It would be interesting to invest the time into enabling the "Infer Any" wart and then it might make sense to embrace Scala variance safe in the notion that `Any` will not be inferred and that way we can use GADTs, i.e. the normal Scala type system that requires less boilerplate.

wemrysi: It is still debatable whether using the constructors directly is always good idea, as they can be viewed as an implementation detail. Depending only on the prisms can make client code more resilient to changes in representation.

core/src/main/scala/quasar/physical/mongodb/execution.scala
jedesah: It's not a constructor as it's in a pattern match. It does appear as though `IList` no longer has an `unapply` (if it even ever had one, since `NonEmptyList` was previously backed by `List`)

core/src/main/scala/quasar/sql/package.scala
jedesah: The sql parser does not support `$` in an identifier if it's not quoted.

Why exactly upgrading to scalaz `7.2.1` triggers this error, I don't quite remember (or never figured out), but I think it might have something to do with slightly different path generators in the latests `pathy` or something.

core/src/main/scala/quasar/std/array.scala
jedesah: Type inference.

core/src/main/scala/quasar/physical/mongodb/execution.scala
mossprescott: Looks like they removed `NonEmptyList.unapplySeq`. Awesome.

core/src/main/scala/quasar/std/array.scala
mossprescott: Seems like `nels()` or `NonEmptyList()` with the single argument should work.

core/src/main/scala/quasar/sql/package.scala
sellout: So, this isn’t a breaking change, since we already couldn’t parse queries like this, right? We just never generated tests that hit this issue before.

I’m assuming the `SimpleNamePattern` was defined based on Mongo?

CHANGELOG/minor.md
```diff
@@ -0,0 +1 @@
+Upgrade to `scalaz` version `7.2.1`
```
sellout: Shouldn’t this be revision.md, since it doesn’t affect users of the Web API?

core/src/main/scala/quasar/sql/package.scala
mossprescott: I don't think it can be a breaking change, because pretty-printing the AST isn't part of the API, right.

It was meant to match the parser, but I'd be surprised if this was ever exactly right. I was satisfied that @jedesah at least had a reason why removing `$` is correct.

One more reason to switch to iso.

jedesah: @sellout I believe that is correct. I don't know even know whether parsing identifiers with `$` characters in them that are not quoted is something we want to do. Probably.

sellout: @jedesah Yeah, I don’t think we want to parse bare identifiers with `$`. It’s not permitted in SQL92, and it doesn’t give us any flexibility/compositionality to add it.

core/src/main/scala/quasar/std/library.scala
jedesah: So apparently, @sellout has an upcoming PR that changes the return type of `typecheck` to `SemanticError`. I might be inclined to wait for this because I remember trying to change that and backing away because It was getting relatively involved (keep in mind that at the time I had an endless see of compile errors still ahead of me! 😄 )

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
jedesah: Actually necessary. But fair question!

core/src/test/scala/quasar/physical/mongodb/collection.scala
jedesah: We seem to be generating longer paths with greater frequency now in the latest version of `pathy` because even with this I have seen the occasional failure. Perhaps I need to write a custom generator.

core/src/test/scala/quasar/std/math.scala
jedesah: I didn't do the change to open up that can of worms again, it would seem that `specs2` 3.x is more opinionated about how to accomplish this.

jedesah: At least now, it doesn't require a ton of `implicitly` (that makes me think, did I remove those huge lists of `implicitly`s in the web tests? I don't think so)

jedesah: Looks like I did.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
jedesah: Because I was too lazy to figure out what the new return type is now with `specs2` `3.x` and it kind of seemed not that important.

project/Dependencies.scala
jedesah: Done!

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -235,10 +235,11 @@ object MessageFormat {
     }
   }
 
-  def fromAccept(accept: Option[Accept]): MessageFormat =
+  def fromAccept(accept: Option[Accept]): MessageFormat = {
     // TODO: MediaRange needs an Order instance – combining QValue ordering
     //       with specificity (EG, application/json sorts before
     //       application/* if they have the same q-value).
-    accept.flatMap(_.values.sortBy(_.qValue).list.map(fromMediaType).flatten(Option.option2Iterable).headOption)
+    accept.flatMap(_.values.sortBy(_.qValue).list.map(_.mediaRange).map(fromMediaType).flatten(Option.option2Iterable).lastOption)
```
jedesah: Btw, I never figured out what changed which made me have to change from `headOption` to `lastOption`. Kind of disconcerting, but luckily we have tests covering the functionality.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
jedesah: You mean `join`? I don't see why not.

mossprescott: Right, `join`. Or maybe it just needs to be `flatMap` above. It just seems perverse to build a `Task[Task[_]]` and then perform twice.

By the way, I'm really happy they switched from `run` and now these things are easier to spot.

jedesah: Ya, quite happy with the `run` rename, if only because most Monad Transformers seemed to use the same name, making it super confusing!

CHANGELOG/minor.md
jedesah: I can't think of why I put "minor" so I will change it.

jedesah:
@mossprescott @wemrysi Thx for all the comments! I think I have addressed everything.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
wemrysi: I think it is good practice to be explicit about return types in the public API.

core/src/main/scala/quasar/effect/KeyValueStore.scala
```diff
@@ -109,7 +109,7 @@ object KeyValueStore {
     */
   def fromTaskRef[K, V](ref: TaskRef[Map[K, V]]): KeyValueStore[K, V, ?] ~> Task =
     new (KeyValueStore[K, V, ?] ~> Task) {
-      val toST = toState[State](Lens.id[Map[K,V]])
+        val toST = toState[State[Map[K,V],?]](Lens.id[Map[K,V]])
```
wemrysi: Extra indent here.

core/src/test/scala/quasar/std/math.scala
mossprescott: That's disappointing, but I don't think it's worth fighting with this, say by creating newtypes for each.

I guess I'll just have to live with it, at least until we start our own property-test project ;)

core/src/test/scala/quasar/physical/mongodb/collection.scala
```diff
@@ -155,7 +155,12 @@ class CollectionSpec extends Specification with ScalaCheck with DisjunctionMatch
           err  => scala.sys.error(err.toString),
           coll => identicalPath(f.path, coll.asFile) must beTrue)
       }
-    }.setGen(Gen.resize(120,PathOf.absFileOfArbitrary[SpecialStr].arbitrary))
+      // MongoDB doesn't like "paths" that are longer than 120 bytes of utf-8 encoded characters
+      // So there is absolutely no point generating paths longer than 120 characters. However, a 120 character path still
+      // has a good change of being refused by MongoDB because many of the characters generated by SpecialStr have large
+      // utf-encodings so we retryUntil we find an appropriate one even if that is somewhat dangerous (unbounded execution time)
+      // For this particular test, execution time does not seem to be a large concern (couple seconds).
+    }.setGen(Gen.resize(120,PathOf.absFileOfArbitrary[SpecialStr].arbitrary).retryUntil(f => Collection.fromPath(f.path).isRight))
```
mossprescott: I think `set(maxSize=...)` would be better here, as `Gen.resize` will force it to always be 120, won't it? Also, It would seem that the use of `==>` is redundant with `retryUntil`.

What if you just use `set(maxSize = 5)` as in the test above? I'm not exactly sure how that constrains the paths (probably, up to five segments of up to five characters each). If that seems too limited, use a bigger maxSize and set the maxDiscardRatio as needed.

jedesah: > I think set(maxSize=...) would be better here, as Gen.resize will force it to always be 120, won't it?

Ah, yes, was not familiar with `set(maxSize=...)`. That does seem superior to `Gen.resize` for the reason you mention.

> Also, It would seem that the use of ==> is redundant with retryUntil.

I disagree and this comes back to our previous discussion about where to stick custom `Gen`s (I swear I did not secretly ask Eric to get rid of your version! 😛 ). The way I see it, everything at the bottom should be considered "implementation details" of making a property that executes in a reasonable amount of time and without giving up whereas everything above should be part of the formal specification. So in this case, the `retryUntil` and `set(maxSize=...)` are details of how we are getting the property to not give up whereas the `==>` is part of the formal specification.

> What if you just use set(maxSize = 5) as in the test above? I'm not exactly sure how that constrains the paths (probably, up to five segments of up to five characters each). If that seems too limited, use a bigger maxSize and set the maxDiscardRatio as needed.

How do you not remember that, you are the one who implemented `size` for generated paths not that long ago! 😄 . For some reason, I believe that size for paths represents to total amount of characters in the path. I considered trying to set the size smaller, but that meant there was some part of the domain that wasn'tcoveredd, even though it did of course reduce the amount of discarded tests. Ultimately, I think that by setting the `maxSize` to `120` and using a pretty big value of `maxDiscardRatio` seems like the best approach to me.

jedesah: So turns out for some reason that I can only hypothesize explanations for that when I use `retryUntil`, the property executes almost instantly whereas if I set a really high `maxDiscardRatio` (which is required if I am going to set `maxSize` to `120`), the test just hangs.

mossprescott: That's hilarious—now that you mention it I do remember writing a fancy size-handling generator.

I just wanted to raise the question again; I'm ok with maxSize and retryUntil if that's what seems to work best.

jedesah:
@mossprescott @wemrysi Addressed the rest of your comments. I think this is ready to go if the build passes.

mossprescott:
Two things I just noticed:

There is a warning showing up in every build: `[info] Warning: Unknown ScalaCheck args provided: include exclusive`. I don't see it on master, so it might be worth looking into.

Specs2 now has a more FP-happy of doing what we do with `step` in a few places. I think the equivalent would be `.beforeAll` and `.afterAll` or something like that. I think it would be good to switch, if you feel inspired, but it's ok with me if you just leave it and we get to that later.

jedesah:
@mossprescott I looked into the warning for some part of this morning and I was not able to come up with an explanation for why it suddenly has appeared. I can only assume that the `testOptions` we are setting are being passed both to `specs2` and `scalacheck` and scalacheck (or sbt) is complaining because they don't make sense for `scalacheck`. Is that because `specs2` is no longer "consuming" them, maybe but I have to reason to believe that as the arguments are still documented in the latest version. What should we do?

Also, I think it would be best to merge this ASAP and move onto advanced work so I would be inclined to leave that additional change you mentioned that is now possible for another PR.

jedesah:
I am a little worried about whether the build will pass or not, it seems like last time it failed two of them because of the too many value generation attempts even though I set `retryUntil(κ(true))` which seems very odd to me.

jedesah:
So the build passed. My guess is luck, but I am okay with that in the present circumstances.

jedesah:
@mossprescott ready?

mossprescott:
LGTM

PR #1167, opened 2016-04-13
Enable rewriting joins without filters.

sellout:
Required inlining lets where the binding is simply a variable.

This fixes a chunk of SD-1190, but we still don’t entirely rewrite
many-way joins.


mossprescott:
LGTM

sellout:
Argh, forgot to add a changelog here.

PR #1169, opened 2016-04-14
Some parser changes while writing the SQL² grammar.

sellout:
* collapse `is`/`is not` syntax into `relational_suffix` and generalize
  it to be equivalent to `=`/`!=`;
* `negatable_suffix`es can now have an optional leading `is`, to
  parallel the other `is not` construction (e.g., `where foo is not between 3
  and 10`);
* remove the `IsNull` function (it’s equivalent to `Eq(_, Null)` now,
  and allowing it to be generated in arbitraries broke round-tripping
  tests); and
* extract function syntax into its own parse rule, and explicitly match
  `is_null` to avoid a breaking change.

### DIFF
--- a/core/src/main/scala/quasar/sql/parser.scala
+++ b/core/src/main/scala/quasar/sql/parser.scala
@@ -185,7 +185,8 @@ private[sql] class SQLParser extends StandardTokenParsers {
     op("<")  ^^^ Lt  |
     op("<=") ^^^ Le  |
     op(">")  ^^^ Gt  |
-    op(">=") ^^^ Ge
+    op(">=") ^^^ Ge  |
+    keyword("is") ~> opt(keyword("not")) ^^ (_.fold[BinaryOperator](Eq)(κ(Neq)))
 
   def relational_suffix: Parser[Expr => Expr] =
     relationalOp ~ default_expr ^^ {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/quasar/sql/parser.scala
```diff
@@ -312,6 +305,14 @@ private[sql] class SQLParser extends StandardTokenParsers {
 
   def set_list: Parser[List[Expr]] = op("(") ~> repsep(expr, op(",")) <~ op(")")
 
+  def function_expr: Parser[Expr] =
+    ident ~ set_list ^^ {
+      // TODO: `is_null` is deprecated, but leaving this here until we figure
+      //       out how to message deprecation to users.
```
drostron: I take it this is the first time there's been a SQL² deprecation?

drostron:
Do we typically add parser tests for this sort of change?

core/src/main/scala/quasar/sql/parser.scala
sellout: Yep, although there are definitely other times we’ve wanted to message warnings to users without failing queries. I think we want something like `\&/` as `Validation` is to `\/`. I.e., a `Validation` where we can still succeed without losing the messages. This is just one more use case for that.

sellout:
Right, tests …

sellout:
Well, the tests were already useful 😄 Back in your court, @drostron.

drostron:
👍🏽

core/src/main/scala/quasar/sql/parser.scala
drostron: 👌🏽

PR #1170, opened 2016-04-19
Wip/sd 1007

alissapajer:
@sellout so we are failing to compile nested `Let` like so:
```
"""foo := 5; bar := 7; bar + foo"""
```
This parses correctly, but the problem is, once we get to the compiler, we add each of `bar` and `foo` to two *distinct* `TableContext` `subtables`, which are then put into the first and second indices of `List[TableContext]` in `CompilerState`.

And when we want to determine which `LogicalPlan`s `foo` and `bar` each refer to, we look in the `TableContext` at index 0 for *both* of them, without dropping the head of the list when we want to look for the second one. Really we've stored `bar` at index 0 and `foo` at index 1 in the `List[TableContext]` but we never check the second index.

I'm not sure exactly the way forward, but I think the `List[TableContext]` needs to be rethought in light of `Let`. Or in any case, I don't fully understand it :) As it is now, it seems like we don't combine `subtables` properly to account for a case such as this, when we need to access `foo` and `bar` in the same `TableContext`.

We handle this case correctly in a join relation on two tables by using `++` on `TableContext`, which concats the contained `subtables`. So we essentially need to consider `"bar + foo"` as a join (or sorts) of `Ident`s so that we can correctly concat the `subtables`. Basically, when more than one `Let` is in scope, the `subtables` need to reflect this :)

### DIFF
--- a/core/src/main/scala/quasar/compiler.scala
+++ b/core/src/main/scala/quasar/compiler.scala
@@ -57,41 +57,86 @@ trait Compiler[F[_]] {
         this.subtables ++ that.subtables)
   }
 
+  private final case class BindingContext(
+    subbindings: Map[String, Fix[LogicalPlan]]) {
+    def ++(that: BindingContext): BindingContext =
+      BindingContext(this.subbindings ++ that.subbindings)
+  }
+
+  private final case class Context(
+    bindingContext: List[BindingContext],
+    tableContext: List[TableContext]) {
+
+    def add(bc: BindingContext, tc: TableContext): Context = {
+      val modBindingContext: List[BindingContext] =
+        this.bindingContext match {
+          case head :: tail => head ++ bc :: head :: tail
+          case Nil => bc :: Nil
+        }
+
+      val modTableContext: List[TableContext] =
+        tc :: this.tableContext
+
+      Context(modBindingContext, modTableContext)
+    }
+
+    def dropHead: Context =
+      Context(this.bindingContext.drop(1), this.tableContext.drop(1))
+  }
+
   private final case class CompilerState(
-    fields:       List[String],
-    tableContext: List[TableContext],
-    nameGen:      Int)
+    fields: List[String],
+    context: Context,
+    nameGen: Int)
 
   private object CompilerState {
-    /** Runs a computation inside a table context, which contains compilation
-      * data for the tables in scope.
+
+    /** Runs a computation inside a binding/table context, which contains
+      * compilation data for the bindings/tables in scope.
       */
-    def contextual[A](t: TableContext)(f: CompilerM[A])(implicit m: Monad[F]):
-        CompilerM[A] =
-      mod((s: CompilerState) => s.copy(tableContext = t :: s.tableContext)) *>
-        f <*
-        mod((s: CompilerState) => s.copy(tableContext = s.tableContext.drop(1)))
+    def contextual[A](bc: BindingContext, tc: TableContext)(
+      compM: CompilerM[A])(
+      implicit m: Monad[F]): CompilerM[A] = {
+
+      def preMod: CompilerState => CompilerState =
+        (state: CompilerState) => state.copy(context = state.context.add(bc, tc))
+
+      def postMod: CompilerState => CompilerState =
+        (state: CompilerState) => state.copy(context = state.context.dropHead)
+
+      mod(preMod) *> compM <* mod(postMod)
+    }
 
     def addFields[A](add: List[String])(f: CompilerM[A])(implicit m: Monad[F]):
         CompilerM[A] =
       for {
         curr <- read[CompilerState, List[String]](_.fields)
         _    <- mod((s: CompilerState) => s.copy(fields = curr ++ add))
-        a <- f
+        a    <- f
       } yield a
 
     def fields(implicit m: Monad[F]): CompilerM[List[String]] =
       read[CompilerState, List[String]](_.fields)
 
     def rootTable(implicit m: Monad[F]): CompilerM[Option[Fix[LogicalPlan]]] =
-      read[CompilerState, Option[Fix[LogicalPlan]]](_.tableContext.headOption.flatMap(_.root))
+      read[CompilerState, Option[Fix[LogicalPlan]]](_.context.tableContext.headOption.flatMap(_.root))
 
     def rootTableReq(implicit m: Monad[F]): CompilerM[Fix[LogicalPlan]] =
       rootTable.flatMap(_.map(emit).getOrElse(fail(CompiledTableMissing)))
 
+    // prioritize binding context - when we want to prioritize a table,
+    // we will have the table reference already in the binding context
     def subtable(name: String)(implicit m: Monad[F]):
         CompilerM[Option[Fix[LogicalPlan]]] =
-      read[CompilerState, Option[Fix[LogicalPlan]]](_.tableContext.headOption.flatMap(_.subtables.get(name)))
+      read[CompilerState, Option[Fix[LogicalPlan]]]{ state =>
+        state.context.bindingContext.headOption.flatMap { bc =>
+          bc.subbindings.get(name) match {
+            case None =>
+              state.context.tableContext.headOption.flatMap(_.subtables.get(name))
+            case s => s
+          }
+        }
+      }
 
     def subtableReq(name: String)(implicit m: Monad[F]):
         CompilerM[Fix[LogicalPlan]] =
@@ -99,7 +144,7 @@ trait Compiler[F[_]] {
         _.map(emit).getOrElse(fail(CompiledSubtableMissing(name))))
 
     def fullTable(implicit m: Monad[F]): CompilerM[Option[Fix[LogicalPlan]]] =
-      read[CompilerState, Option[Fix[LogicalPlan]]](_.tableContext.headOption.map(_.full()))
+      read[CompilerState, Option[Fix[LogicalPlan]]](_.context.tableContext.headOption.map(_.full()))
 
     /** Generates a fresh name for use as an identifier, e.g. tmp321. */
     def freshName(prefix: String)(implicit m: Monad[F]): CompilerM[Symbol] =
@@ -152,7 +197,7 @@ trait Compiler[F[_]] {
 
     def flattenJoins(term: Fix[LogicalPlan], relations: SqlRelation[CoExpr]):
         Fix[LogicalPlan] = relations match {
-      case _: NamedRelation[_] => term
+      case _: NamedRelation[_]             => term
       case JoinRelation(left, right, _, _) =>
         Fix(ObjectConcat(
           flattenJoins(Left.projectFrom(term), left),
@@ -197,7 +242,13 @@ trait Compiler[F[_]] {
         for {
           stepName <- CompilerState.freshName("tmp")
           current  <- current
-          next2    <- CompilerState.contextual(tableContext(LogicalPlan.Free(stepName), relations))(next)
+          bc        = relations match {
+            case ExprRelationAST(_, name)        => BindingContext(Map(name -> LogicalPlan.Free(stepName)))
+            case TableRelationAST(_, Some(name)) => BindingContext(Map(name -> LogicalPlan.Free(stepName)))
+            case id @ IdentRelationAST(_, _) => BindingContext(Map(id.aliasName -> LogicalPlan.Free(stepName)))
+            case r                               => BindingContext(Map())
+          }
+          next2    <- CompilerState.contextual(bc, tableContext(LogicalPlan.Free(stepName), relations))(next)
         } yield LogicalPlan.Let(stepName, current, next2)
       }.getOrElse(next)
     }
@@ -235,11 +286,16 @@ trait Compiler[F[_]] {
 
     def compileRelation(r: SqlRelation[CoExpr]): CompilerM[Fix[LogicalPlan]] =
       r match {
+        case IdentRelationAST(name, _) =>
+          CompilerState.subtableReq(name)
+
         case TableRelationAST(path, _) =>
           sandboxCurrent(canonicalize(path)).cata(
             p => emit(LogicalPlan.Read(p)),
             fail(InvalidPathError(path, None)))
+
         case ExprRelationAST(expr, _) => compile0(expr)
+
         case JoinRelation(left, right, tpe, clause) =>
           for {
             leftName <- CompilerState.freshName("left")
@@ -249,6 +305,7 @@ trait Compiler[F[_]] {
             left0 <- compileRelation(left)
             right0 <- compileRelation(right)
             join <- CompilerState.contextual(
+              BindingContext(Map()),
               tableContext(leftFree, left) ++ tableContext(rightFree, right))(
               compile0(clause).map(c =>
                 LogicalPlan.Invoke(
@@ -356,6 +413,11 @@ trait Compiler[F[_]] {
               })
         })
 
+      case LetF(name, form, body) => {
+        val rel = ExprRelationAST(form, name)
+        step(rel)(compile0(form).some)(compile0(body))
+      }
+
       case SetLiteralF(values0) =>
         values0.traverse(compile0).map(vs =>
           ShiftArray(MakeArrayN(vs: _*).embed).embed)
@@ -385,10 +447,10 @@ trait Compiler[F[_]] {
               Fix(ObjectProject(obj, LogicalPlan.Constant(Data.Str(name)))))
           else
             for {
-              rName   <- relationName(node).fold(fail, emit)
-              table   <- CompilerState.subtableReq(rName)
+              rName <- relationName(node).fold(fail, emit)
+              table <- CompilerState.subtableReq(rName)
             } yield
-              if ((rName:String) ≟ name) table
+              if ((rName: String) ≟ name) table
               else Fix(ObjectProject(table, LogicalPlan.Constant(Data.Str(name)))))
 
       case InvokeFunctionF(name, args) =>
@@ -422,7 +484,7 @@ trait Compiler[F[_]] {
   }
 
   def compile(tree: Cofree[ExprF, Annotations])(implicit F: Monad[F]): F[SemanticError \/ Fix[LogicalPlan]] = {
-    compile0(tree).eval(CompilerState(Nil, Nil, 0)).run.map(_.map(Compiler.reduceGroupKeys))
+    compile0(tree).eval(CompilerState(Nil, Context(Nil, Nil), 0)).run.map(_.map(Compiler.reduceGroupKeys))
   }
 }
 
--- a/core/src/main/scala/quasar/repl/Command.scala
+++ b/core/src/main/scala/quasar/repl/Command.scala
@@ -29,7 +29,7 @@ object Command {
   private val ExitPattern         = "(?i)(?:exit)|(?:quit)".r
   private val HelpPattern         = "(?i)(?:help)|(?:commands)|\\?".r
   private val CdPattern           = "(?i)cd(?: +(.+))?".r
-  private val NamedExprPattern    = "(?i)([^ :]+) *:= *(.+)".r
+  private val NamedExprPattern    = "(?i)([^ :]+) *<- *(.+)".r
   private val LsPattern           = "(?i)ls(?: +(.+))?".r
   private val SavePattern         = "(?i)save +([\\S]+) (.+)".r
   private val AppendPattern       = "(?i)append +([\\S]+) (.+)".r
--- a/core/src/main/scala/quasar/semantics.scala
+++ b/core/src/main/scala/quasar/semantics.scala
@@ -172,11 +172,19 @@ trait SemanticAnalysis {
     case _ => Nil
   }
 
+  case class BindingScope(scope: Map[String, SqlRelation[Expr]])
+
+  implicit val ShowBindingScope: Show[BindingScope] = new Show[BindingScope] {
+    override def show(v: BindingScope) = v.scope.toString
+  }
+
   case class TableScope(scope: Map[String, SqlRelation[Expr]])
 
   implicit val ShowTableScope: Show[TableScope] = new Show[TableScope] {
     override def show(v: TableScope) = v.scope.toString
   }
+
+  case class Scope(tableScope: TableScope, bindingScope: BindingScope)
 
   import Validation.FlatMap._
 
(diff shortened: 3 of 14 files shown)


alissapajer:
```
new ts is TableScope(Map(fooooo -> ExprRelationAST(Fix(IntLiteralF(5)),fooooo)))
new ts is TableScope(Map(fooooo -> ExprRelationAST(Fix(IntLiteralF(5)),fooooo), baaaar -> ExprRelationAST(Fix(IntLiteralF(7)),baaaar)))
get for baaaar was Some(ExprRelationAST(Fix(IntLiteralF(7)),baaaar))
get for fooooo was Some(ExprRelationAST(Fix(IntLiteralF(5)),fooooo))
stepping with fooooo and scalaz.Cofree$$anon$11@400cc694 and scalaz.Cofree$$anon$11@4232ce62
stepping with baaaar and scalaz.Cofree$$anon$11@83bb17b and scalaz.Cofree$$anon$11@24b254a
name for baaaar is baaaar
headOption is Some(TableContext(Some(Fix(FreeF('tmp1))),<function0>,Map(baaaar -> Fix(FreeF('tmp1))))) for name baaaar
name for fooooo is fooooo
headOption is Some(TableContext(Some(Fix(FreeF('tmp1))),<function0>,Map(baaaar -> Fix(FreeF('tmp1))))) for name fooooo
```

Some debugging. On the second to last line, you can see we're looking for name `fooooo` in a `TableContext` with only `baaaar`. I dropped the first thing from that list and confirmed that name `fooooo` is then at the head.

mossprescott:
Have you given any thought to how this affects the REPL's use of `:=`? I think it would be okay to change that syntax, mind you. I don't _think_ it's being used much.

sellout:
@mossprescott Yeah, that currently conflicts … but not sure what to do about it.

@alissapajer I think you’re right about the `++` thing. The way we currently do it, we ignore outer tables, which I think is to prevent `select foo from bar, (select bar from baz) as quux`. from capturing the outer bar when resolving the inner one … but such “correlated subqueries” are a part of SQL that we should support, so I think accumulating the table context is useful. I.e., in contextual, instead of `t :: s.tableContext`, we should perhaps `TableContext(s.tableContext.scope ++ t.scope) :: s.tableContext`, so we have all the outer tables, not just the ones in the current select.

sellout:
Actually, I'm rethinking that.

alissapajer:
```scala
    def contextualLet[A](t: TableContext)(f: CompilerM[A])(implicit m: Monad[F]):
        CompilerM[A] =
      mod((s: CompilerState) => s.copy(tableContext = s.tableContext match {
        case head :: tail => t ++ head :: tail
        case Nil => t :: Nil
      })) *> f
```
I think we need two different `CompilerState.contextual`, because in the case of a `Let` binding, once you are within the scope of a `Let`, as long as you continue to traverse the tree through the body of the `Let`, you are *forever* in its context. Meaning, if you consider `foo := bar; baz`, then *any* reference to `foo` inside `baz` must reference `bar`. Is this the semantic we want? To me this is the intuitive semantic, but if/when it's consistent with SQL, I'm not sure.

But in the case `select foo from bar, (select bar from baz) as quux`, we may (or may not! I'm not sure...) want the outer `bar` to shadow the inner one. But in any case, I think we need to handle shadowing of `Let` bindings *separately* from the shadowing of `Select` statements. And so this will allow us to choose how to handle each separately.

With this change we at least get the following. I'll report back about the status of the rest of the tests :) And then we pretty much need a whole slew of tests with shadowed names so we can make future changes knowing we have tests in this vein :)
```
💪 $ foo := 5; bar := 7; bar + foo
;

Query time: 0,2s
 value |
-------|
    12 |
```

core/src/main/scala/quasar/compiler.scala
```diff
@@ -63,21 +63,39 @@ trait Compiler[F[_]] {
...
-    /** Runs a computation inside a table context, which contains compilation
-      * data for the tables in scope.
-      */
-    def contextual[A](t: TableContext)(f: CompilerM[A])(implicit m: Monad[F]):
-        CompilerM[A] =
-      mod((s: CompilerState) => s.copy(tableContext = t :: s.tableContext)) *>
-        f <*
-        mod((s: CompilerState) => s.copy(tableContext = s.tableContext.drop(1)))
+
+    sealed trait StepContext {
```
alissapajer: Here is the proposed `StepContext`. We guarantee a `contextual` and then can choose the impl based on the situation.

This (or something else) has created an `AmbiguousReference` bug in the query
```
"select city, pop from (select city, pop from zips) as temp"
```
The test in `CompilerSpec` is testing that the inner selection names don't refer to the outer selection names. This the situation I was attempting to differentiate here, so either this was introduced here, or introduced earlier. Will verify either way.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -63,21 +63,39 @@ trait Compiler[F[_]] {
...
-    /** Runs a computation inside a table context, which contains compilation
-      * data for the tables in scope.
-      */
-    def contextual[A](t: TableContext)(f: CompilerM[A])(implicit m: Monad[F]):
-        CompilerM[A] =
-      mod((s: CompilerState) => s.copy(tableContext = t :: s.tableContext)) *>
-        f <*
-        mod((s: CompilerState) => s.copy(tableContext = s.tableContext.drop(1)))
+
+    sealed trait StepStrategy {
```
alissapajer: Here is the proposed `StepStrategy`. We guarantee a contextual and then can choose the impl based on the situation.

This (or something else) has created an `AmbiguousReference` bug in the query

"select city, pop from (select city, pop from zips) as temp"

The test in `CompilerSpec` is testing that the inner selection names don't refer to the outer selection names. This the situation I was attempting to differentiate here, so either this was introduced here, or introduced earlier. Will verify either way.

core/src/main/scala/quasar/semantics.scala
```diff
@@ -195,9 +197,20 @@ trait SemanticAnalysis {
         relations.fold[ValidSem[Map[String, SqlRelation[Expr]]]](
           success(Map[String, SqlRelation[Expr]]()))(
           findRelations)
-          .map(m => sel.map((TableScope(m), _)))
-      case x => success(x.map((p._1, _)))
+          .map(m => sel.map((TableScope(m ++ ts.scope), _)))
```
alissapajer: Gotcha. I think we *don't* want to do this anymore and want to revert to what we had before, and just have the scope be accumulated in the `LetF` case.

core/src/main/scala/quasar/semantics.scala
```diff
@@ -197,7 +197,7 @@ trait SemanticAnalysis {
         relations.fold[ValidSem[Map[String, SqlRelation[Expr]]]](
           success(Map[String, SqlRelation[Expr]]()))(
           findRelations)
-          .map(m => sel.map((TableScope(m ++ ts.scope), _)))
```
alissapajer: Gotcha. I think we don't want to do this anymore and want to revert to what we had before, and just have the scope be accumulated in the LetF case.

alissapajer:
Ok at this point all tests are passing locally for me at 952f84824965a19661bf44b47e1a3fc98f3d7409! Next step is writing more tests and discussing if these are the semantics we actualy desire.

core/src/main/scala/quasar/compiler.scala
alissapajer: Fix: don't accumulate `TableScope` in the `SelectF` case.

alissapajer:
It seems
```
[error]     x using disposition to download as zipped directory
[error]      Gave up after only 0 passed tests. 2 tests were discarded. (DataServiceSpec.scala:59)
```
is possibly flakey in master. Saw it fail in PR #1171 too. Has this been failing before?

core/src/test/scala/quasar/compiler.scala
```diff
@@ -135,6 +135,98 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
             "name" -> ObjectProject(read("city"), Constant(Data.Str("name"))))))
     }
 
+    "compile basic let" in {
```
alissapajer: @sellout let me know if you agree with the expectations of the tests here, especially the last couple new tests added at the bottom.

core/src/main/scala/quasar/semantics.scala
```diff
@@ -158,6 +158,12 @@ trait SemanticAnalysis {
     case _ => Nil
   }
 
+  case class BindingScope(scope: Map[String, SqlRelation[Expr]])
```
alissapajer: We have a `BindingScope` now. This way we can have `Let` bindings in scope inside a `Select`. The example query that requires this addition is `foo := 12; select foo from bar`.

core/src/main/scala/quasar/semantics.scala
```diff
@@ -323,21 +342,26 @@ trait SemanticAnalysis {
...
-          Provenance.anyOf[Map[String, ?]](tableScope.scope ∘ (Provenance.Relation(_))) match {
+      case SetLiteralF(_)           => success(Provenance.Value)
+      case ArrayLiteralF(_)         => success(Provenance.Value)
+      case MapLiteralF(_)           => success(Provenance.Value)
+      case SpliceF(expr)            => success(expr.getOrElse(Provenance.Empty))
+      case VariF(_)                 => success(Provenance.Value)
+      case BinopF(left, right, _)   => success(left & right)
+      case UnopF(expr, _)           => success(expr)
+      case IdentF(name)             =>
+        val scope = tableScope.scope.get(name) match {
```
alissapajer: Which scope do we actually want to prioritize here? We need test cases that fail when the wrong scope is prioritized. I'm not sure how to construct such an example yet...

core/src/main/scala/quasar/compiler.scala
```diff
@@ -57,18 +57,27 @@ trait Compiler[F[_]] {
         this.subtables ++ that.subtables)
   }
 
+  private final case class BindingContext(
```
alissapajer: We also now have a `BindingContext`. Before I had mad a change to treat tables and bindings differently in how we add them to the context (tables get prepended, bindings get concatted) but I *think* this separation is what we actually need. Because then we explicitly carry around the `BindingContext` so we know what variables are bound in the context, and then can decide how to handle those, given the tables that are also present in the context... again though this needs specific examples...

core/src/test/scala/quasar/compiler.scala
```diff
@@ -135,6 +135,98 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
...
+              ArrayConcat[FLP](
+                MakeArrayN[Fix](Constant(Data.Int(1))),
+                MakeArrayN[Fix](Constant(Data.Int(2)))),
+              MakeArrayN[Fix](Constant(Data.Int(3))))))
+
+      testLogicalPlanCompile(query, expectation)
+    }
+
+    "compile let with select in body selecting let binding ident" in {
+      val query = """foo := 12; select foo from bar"""
```
alissapajer: Getting this query and `select foo from (select foo from bar)` to *both* compile properly was the the challenging part.

core/src/main/scala/quasar/semantics.scala
alissapajer: So it turns out we need to have `Let` bindings in context here, but not tables from *other* `Select`s.

core/src/main/scala/quasar/semantics.scala
```diff
@@ -333,23 +354,28 @@ trait SemanticAnalysis {
...
-        tableScope.scope.get(name).fold(
-          Provenance.anyOf[Map[String, ?]](tableScope.scope ∘ (Provenance.Relation(_))) match {
+      case SetLiteralF(_)           => success(Provenance.Value)
+      case ArrayLiteralF(_)         => success(Provenance.Value)
+      case MapLiteralF(_)           => success(Provenance.Value)
+      case SpliceF(expr)            => success(expr.getOrElse(Provenance.Empty))
+      case VariF(_)                 => success(Provenance.Value)
+      case BinopF(left, right, _)   => success(left & right)
+      case UnopF(expr, _)           => success(expr)
+      case IdentF(name)             =>
```
alissapajer: `select (foo := select * from bar; select * from foo) from foo`

core/src/main/scala/quasar/compiler.scala
alissapajer: `select (foo := select * from bar; select * from foo), foo from bar`

core/src/main/scala/quasar/compiler.scala
```diff
@@ -197,7 +216,12 @@ trait Compiler[F[_]] {
         for {
           stepName <- CompilerState.freshName("tmp")
           current  <- current
-          next2    <- CompilerState.contextual(tableContext(LogicalPlan.Free(stepName), relations))(next)
+          bc        = relations match {
+            case ExprRelationAST(_, name) => { /*println(s"adding to binding context");*/ BindingContext(Map(name -> LogicalPlan.Free(stepName))) }
+            case IdentRelationAST(name, _) => {/* println(s"adding to binding context");*/ BindingContext(Map(name -> LogicalPlan.Free(stepName))) }
```
alissapajer: handle alias

alissapajer: if we had two nested lets with the same name - would we always refer to the outer one?

alissapajer:
```
foo := 4; (select * from (bar as foo))
foo := 4; (select * from (foo := bar; foo))
```

alissapajer: `case TableRelationAST(_, Some(alias))`

core/src/main/scala/quasar/compiler.scala
```diff
@@ -91,7 +100,17 @@ trait Compiler[F[_]] {
 
     def subtable(name: String)(implicit m: Monad[F]):
         CompilerM[Option[Fix[LogicalPlan]]] =
-      read[CompilerState, Option[Fix[LogicalPlan]]](_.tableContext.headOption.flatMap(_.subtables.get(name)))
+      read[CompilerState, Option[Fix[LogicalPlan]]]{ x =>
```
alissapajer: prioritize `BindingContext` here - when we want to prioritize a table, we will have the table reference already in the binding context

core/src/main/scala/quasar/compiler.scala
```diff
@@ -57,49 +57,94 @@ trait Compiler[F[_]] {
         this.subtables ++ that.subtables)
   }
 
+  private final case class BindingContext(
+    subbindings: Map[String, Fix[LogicalPlan]]) {
+    def ++(that: BindingContext): BindingContext =
+      BindingContext(this.subbindings ++ that.subbindings)
+  }
+
+  private final case class Context(
```
alissapajer: @sellout made a `Context` along with `add` and `dropHead` functions, so we can better see the transformations that we're making

alissapajer: And then in `contextual` we call these as we `mod` the `CompilerState`.

core/src/main/scala/quasar/repl/Command.scala
```diff
@@ -29,7 +29,7 @@ object Command {
   private val ExitPattern         = "(?i)(?:exit)|(?:quit)".r
   private val HelpPattern         = "(?i)(?:help)|(?:commands)|\\?".r
   private val CdPattern           = "(?i)cd(?: +(.+))?".r
-  private val NamedExprPattern    = "(?i)([^ :]+) *:= *(.+)".r
+  //private val NamedExprPattern    = "(?i)([^ :]+) *:= *(.+)".r
```
sellout: We can’t disable this in the REPL. I think we discussed adding an optional `let` keyword before a binding, which could be used to explicitly disambiguate this case in the REPL?

core/src/main/scala/quasar/sql/ast.scala
```diff
@@ -168,6 +169,11 @@ sealed trait NamedRelation[A] extends SqlRelation[A] {
   def aliasName: String
 }
 
+final case class IdentRelationAST[A](name: String, alias: Option[String])
```
sellout: Maybe add a doc comment here, explaining how this allows us to refer to a let binding in relation context, where `Ident` allows us to make that reference in expression context, and that there shouldn’t be two separate contexts in the first place 😉

core/src/main/scala/quasar/sql/ast.scala
```diff
@@ -219,9 +225,16 @@ object SelectF {
...
+      case _ => None
     }
 }
+object LetF {
+  def apply[A](name: String, form: A, body: A): ExprF[A] = ExprF.LetF(name, form, body)
+  def unapply[A](obj: ExprF[A]): Option[(String, A, A)] = obj match {
+    case ExprF.LetF(name, form, body) => Some((name, form, body))
+    case _                            => None
+  }
+}
```
sellout: Oh, I’m going to be so happy to get rid of these broken wrappers in my other PR.

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -79,6 +81,37 @@ package object sql {
   implicit class ExprOps(q: Expr) {
     def mkPathsAbsolute(basePath: ADir): Expr =
       q.cata(mapPathsMƒ[Id](refineTypeAbs(_).fold(ι, pathy.Path.unsandbox(basePath) </> _)))
+
+    def makeTables(bindings: List[String]): Expr = q.project match {
+      case sel @ SelectF(_, _, _, _, _, _) => {
+        // perform all the appropriate recursions
+        val sel2 = sel.map(_.makeTables(bindings)).asInstanceOf[ExprF.SelectF[Expr]]  // TODO
```
sellout: Add an explanation to this TODO

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -79,6 +81,37 @@ package object sql {
...
+
+        def mkRel[A](rel: SqlRelation[A]): SqlRelation[A] = rel match {
+          // TODO use rewriteRelationsM
```
sellout: Is it reasonable to change this now?

core/src/test/scala/quasar/compiler.scala
```diff
@@ -135,6 +135,136 @@ class CompilerSpec extends Specification with CompilerHelpers with PendingWithAc
...
+      val query = """foo := 4; select * from bar as foo"""
+      val expectation = Squash(read("bar"))
+
+      testLogicalPlanCompile(query, expectation)
+    }
+
+    "fail to compile let with an inner context of let that shares a binding name in expression context" in {
+      val query = """foo := 4; select * from (foo := bar; foo) as quag"""
+
+      compile(query) must beLeftDisjunction // ambiguous reference for `bar` - `4` or `foo`
```
sellout: Hrmm, is this one correct? I would think that this should expand to `select * from bar as quag`. Oh, but the problem is that bar is in expression context, and so you really need `foo := select * from bar` for that to behave properly … right?

core/src/main/scala/quasar/repl/Command.scala
alissapajer: I think a optional case `let` keyword is a good idea.

An alternative would be to change this syntax in the REPL. 

Though actually, how is naming an `Expr` like this in the REPL actually different than a regular let binding in normal SQL^2?

core/src/test/scala/quasar/compiler.scala
alissapajer: Yes, that's what I was thinking. The next query in the file tests that the case with `foo := select * from bar` compiles correctly.

The differentiation between table and expression context is confusing from a user point of view, though. But I'm not sure the best way to resolve this...

core/src/main/scala/quasar/sql/package.scala
alissapajer: I don't think it actually makes sense to rewrite that in terms of `rewriteRelationsM` anymore. `rewriteRelationsM` is rewriting the `JoinRelation` case, which is not what we're doing here.

On a different note, `rewriteRelationsM` actually isn't used at all in quasar. Do we want to keep it around?

alissapajer:
@sellout replied to your comments - one or two things still need more discussion - also pushed some changes in a new commit

core/src/test/scala/quasar/compiler.scala
sellout: Relation context has to die. It conflicts with MRA semantics anyway. This is just the first time it’s really been an issue (although, I know the front-end has run into similar issues/confusion when generating queries).

core/src/main/scala/quasar/repl/Command.scala
sellout: There are a few differences, one is that the REPL syntax binds variables (`:foo`) rather than identifiers (`foo`) … somewhat more akin to prepared queries. I.e., you can prepare a query with some variables in it, then run it with different variables each time. The front-end uses this feature (although not with the REPL’s syntax) to put editable fields in Markdown that modify the results of a subsequent query.

I’m fine with changing the REPL syntax as well. We don’t currently version the REPL, so it’s pretty much a free-for-all.

alissapajer: I think I'd prefer to change the REPL syntax so that let binding syntax is consistent everywhere. Some ideas for syntax. What do you think?

`:>`
`::`
`->`
`:=:`  // would this be a problem to parse?

core/src/main/scala/quasar/sql/package.scala
sellout: `rewriteRelationsM` rewrites _all_ relations, it only calls out JoinRelation to drill down into the subrelations. But, if it’s not used elsewhere, I don’t think it’s worth rewriting this code to change it, it should just get deleted. And it’ll go away anyway when SqlRelation goes away.

core/src/main/scala/quasar/repl/Command.scala
sellout: @mossprescott Thoughts?

I think `<-` would be better than `->`, but I don’t know if that’s better than other options.

alissapajer: I just pushed a change with `<-`. We can change this if need be.

core/src/main/scala/quasar/sql/package.scala
alissapajer: `rewriteRelationsM` is deleted.

alissapajer:
@sellout I just pushed changes addressing all comments.

sellout:
👍🏾

PR #1172, opened 2016-04-21
fix one randomly failing test

mossprescott:
See SD-1572.

- use `maxSize` to keep the generated data relatively small
- set `minTestsOk` to 10, since 1 seems to confuse specs2/scalacheck
- fix spelling of the disposition header
- rename the test, since the disposition header actually doesn't affect
the zipping behavior

### DIFF
--- a/web/src/test/scala/quasar/api/services/DataServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/DataServiceSpec.scala
@@ -35,7 +35,6 @@ import argonaut.Argonaut._
 import org.http4s._
 import org.http4s.headers._
 import org.http4s.server.middleware.GZip
-import org.scalacheck.Arbitrary
 import org.specs2.specification.core.Fragments
 import org.specs2.execute.AsResult
 import org.specs2.matcher.MatchResult
@@ -127,7 +126,9 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
           }
           "in any supported format if specified" >> {
             "in the content-type header" >> {
-              def testProp(format: MessageFormat) = prop { filesystem: SingleFileMemState => test(format, filesystem) }
+              def testProp(format: MessageFormat) = prop { filesystem: SingleFileMemState =>
+                test(format, filesystem)
+              }
               def test(format: MessageFormat, filesystem: SingleFileMemState) = {
                 val request = Request(
                   uri = pathUri(filesystem.file),
@@ -149,7 +150,9 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
                   testProp(jsonPreciseArray)
                 }
               }
-              "csv" ! prop { (filesystem: SingleFileMemState, format: MessageFormat.Csv) => test(format, filesystem) }
+              "csv" ! prop { (filesystem: SingleFileMemState, format: MessageFormat.Csv) =>
+                test(format, filesystem)
+              }
               "or a more complicated proposition" ! prop { filesystem: SingleFileMemState =>
                 val request = Request(
                   uri = pathUri(filesystem.file),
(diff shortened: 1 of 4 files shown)


web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -262,8 +261,8 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
           }
         }
       }
-      "using disposition to download as zipped directory" ! prop { filesystem: NonEmptyDir =>
-        val disposition = `Content-Disposition`("attachement", Map("filename" -> "foo.zip"))
+      "download as zipped directory" ! prop { filesystem: NonEmptyDir =>
+        val disposition = `Content-Disposition`("attachment", Map("filename" -> "foo.zip"))
```
jedesah: Eh..., does this mean the test was incorrect and still passing?

mossprescott: Not exactly. It's testing that the header is copied to the response, which is all we really can test. The server doesn't do anything else with it; it will zip anytime you give it a directory path.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -272,7 +271,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
         response.status must_== Status.Ok
         response.contentType must_== Some(`Content-Type`(MediaType.`application/zip`))
         response.headers.get(`Content-Disposition`) must_== Some(disposition)
-      }.set(minTestsOk = 1).setGen(Arbitrary.arbitrary[NonEmptyDir].retryUntil(κ(true))) // This test is relatively slow TODO[scalacheck]: Remove retryUnit once upgrade to 0.13.0 as it should no longer be necessary
+      }.set(minTestsOk = 10, maxSize = 20)
```
jedesah: So I am not too surprised that the build passes now because the issue was with the discard ratio being screwed up by the "unexpectedly" (really?) low value of `minTestsOk`. However, I am curious how much compute power is required to "verify" this property now as I am not 100% convinced that `NonEmptyDir` actually "supports" `maxSize`.

jedesah: Have you enable time reporting?

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
jedesah: Ah, yes.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
mossprescott: It runs in less than 2 seconds locally, and I confirmed that the number of files is affected by maxSize. I did see random failures locally with `minTestsOk = 1`, and never with a larger value.

Unfortunately it just failed with "No output" on the coverage run in Travis, which makes me think it must be leaking memory. Let me take a look at that.

jedesah: You mean that you think the code that zips a directory is leaking memory? Ya, does seem rather plausible.

mossprescott:
Looking good, with all four builds successful the first time on both my branch and the PR.

I'm going to kick both builds one more time just to be sure, but this looks promising.

Also, it looks like it shaved a couple of minutes off the build time.

mossprescott:
OK, all 16 builds succeeded. I think this is fixed.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -262,8 +268,8 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
           }
         }
       }
-      "using disposition to download as zipped directory" ! prop { filesystem: NonEmptyDir =>
```
jedesah: As a side note (should not hold up this PR). Should we separate this test into two? It seems, on the one hand we are making sure that directories are returned in zip format and on the other making sure the `Content-Disposition` is being returned intact.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -498,7 +514,7 @@ class DataServiceSpec extends Specification with ScalaCheck with FileSystemFixtu
             "srcPath" := fs.dir,
             "dstPath" := file),
           newState = Unchanged)
-      }
+      }.set(minTestsOk = 10)
```
jedesah: Are these tests particularly slow? If so I wonder if it might be worth adding a comment with that explanation.

mossprescott: Yes; if I restrict the `NonEmptyDir` generator enough to make them run quickly then I still get random failures, and since there were only a few tests it seemed expedient to go this way. I just added comments.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
mossprescott: Sure. That would make it more clear. Maybe three tests, to show that `Content-Disposition` works on files as well as directories.

But I'd really like to get this merged as is if you're otherwise happy.

jedesah:
👍

mossprescott:
Well, this went 11 for 12 in Travis in the process of rebasing, with one unresponsive build getting killed. I'm gonna call that a win and I'm merging.

mossprescott:
Merged as 8.0.2

PR #1173, opened 2016-04-21
Update various dependencies and add explanations

jedesah:


jedesah:
@sellout Paging you in case you have an objection to specifying dependencies using ranges.

alissapajer:
For reference (because I had to just go read about this :) ), here is the [link to the docs](https://ant.apache.org/ivy/history/2.3.0/ivyfile/dependency.html#revision) about dynamic revisions in ivy.

wemrysi:
@alissapajer Thanks for the link!

@jedesah What is the motivation for switching to dynamic revisions? I have my reservations about them as they could lead to reduced build stability amongst us developers (as we may be individually using different versions of a dependency) as well as some uncertainty between our local builds and what travis builds based on what was available when the build started.

I raise the issue as this exact situation just occurred last week for me with a transitive java dependency in the nimbusds oidc library we're using in quasar-advanced. I had a version that was using a slightly older point release of a library that the oidc library depended on (via a range) and had no problems. Travis ran and grabbed a newly released version of the same dependency which was a bugfix release but appeared to not be binary compatible with the old one or something as it resulted in class not found runtime errors during the build. It took me awhile to figure out what was going on and I had to clear out my ivy cache to reproduce the issue locally in order to fix things.

sellout:
@wemrysi This is a good point. You basically want to use write the ranges in a build file, but then a successful build should generate a non-ranged version so you can then get repeatable builds.

mossprescott:
@sellout: I like the concept, but do you have an idea for how that might work?

I think it's important to identify specific versions for all dependencies. Otherwise the commit/tag doesn't even identify a unique configuration, and test results can no longer be considered reliable, etc.

But it would be good to have something to counter the tendency to fall behind. https://github.com/rtimush/sbt-updates seems helpful, and can even fail the build if updates are available, based on configuration.

sellout:
Oh, I don’t think it’d work for SBT, just thinking about idealized dependency management :D

jedesah:
Not going to lie. I switched over to dynamic dependencies because I didn't know we could do that and I thought it was KIND OF COOL.

But I have heard some pretty solid arguments against it so I will change the PR to just update to latest of things.

@mossprescott I have sbt-updates installed locally which is super useful for doing upgrades like this. I think it sounds like a reasonable idea to have it fail the build. We would need to make a few fixes to some false positives it generates for `quasar` and we would probably need to filter out something like `scalaz` `7.3.x` whenever that comes out until when we are willing to make the upgrade, but I am pretty sure there is a `SettingKey` for "ignored" updates.

mossprescott:
Indeed, e.g. `dependencyUpdatesExclusions := moduleFilter(organization = "org.scala-lang")`

jedesah:
Okay, reverted back to specifying dependency versions exactly and added an extra comment.

jedesah:
@wemrysi I think this is a quick review now.

wemrysi:
:+1:

PR #1174, opened 2016-04-22
Modernize sql expr

sellout:
- rename `ExprF` to `Sql` and remove the `F` suffix from all the case classes;
- add lenses and prisms to `Sql` and get rid of the apply/unapply objects;
- added a `Composition` trait and `∘` alias for it to simplify some things that required kind-projector before;
- remove the `Expr` alias for `Fix[Sql]` (née `Fix[ExprF]`);
- restrict the fixpoint wrappers to tests[†]; 

[†]: I wanted to do `trait fixpoint[T[_[_]]]` instead of `package object fixpoint`, so you could choose which fixpoint to use, but I couldn’t get that to compile.


sellout:
@alissapajer You might want to look at this. General cleanup of SQL stuff. I won’t merge it until after your `Let` stuff is in, though ;) Just working with you on that made me want this to be better. So we have exhaustiveness checking now, and lenses (although I didn’t get rid of the uses of `copy` yet, we should at least be able to avoid new ones).

core/src/main/scala/quasar/semantics.scala
```diff
@@ -158,11 +162,12 @@ trait SemanticAnalysis {
     case _ => Nil
   }
 
-  case class TableScope(scope: Map[String, SqlRelation[Expr]])
+  case class TableScope(scope: Map[String, SqlRelation[Unit]])
```
alissapajer: I'm surprised we have enough information to drop the `Expr` here.

So in `TableScope` we don't need to preserve the `Expr` information in `ExprRelationAST`. All that matters is that there is *some* `Expr` that is associated with a given `aliasName: String`. Basically, we're annotating the tree with information about the *existence* of a relation, without providing the relation itself (because the relation itself still exists in the tree inside a `Select`).

@sellout is that what you were thinking?

sellout: Exactly – we only need the name of any relation. I started by parameterizing TableScope on `T`, then realized we don't need it, and it was simpler to forget it.

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -20,38 +20,55 @@ import quasar.Predef._
...
+  def boolLiteral[A] = pPrism[Sql[A], Boolean] { case BoolLiteral(v) => v } (BoolLiteral(_))
 
   // NB: we need to support relative paths, including `../foo`
   type FUPath = pathy.Path[_, pathy.Path.File, pathy.Path.Unsandboxed]
 
-  private val parser = new SQLParser()
+  private def parser[T[_[_]]: Recursive: Corecursive] = new SQLParser[T]()
 
-  val parse: Query => (ParsingError \/ Expr) = parser.parse(_).map(_.transAna(repeatedly(normalizeƒ)))
+  def parse[T[_[_]]: Recursive: Corecursive](q: Query): ParsingError \/ T[Sql] =
```
wemrysi: I think this is nice, but it does have performance implications as @mossprescott made this a `val` to avoid allocating a new parser on each request (or wherever else it is being used).

What do you think about adding an "implementation" object containing these functions specialized to `Fix` (or whatever you're planning on using in practice)? Then we still enjoy the benefits of generality here while avoiding all the extra allocations in practice?

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -20,38 +20,55 @@ import quasar.Predef._
...
+  def floatLiteral[A] = pPrism[Sql[A], Double] { case FloatLiteral(v) => v } (FloatLiteral(_))
+  def stringLiteral[A] = pPrism[Sql[A], String] { case StringLiteral(v) => v } (StringLiteral(_))
+  def nullLiteral[A] = pPrism[Sql[A], Unit] { case NullLiteral() => () } (_ => NullLiteral())
+  def boolLiteral[A] = pPrism[Sql[A], Boolean] { case BoolLiteral(v) => v } (BoolLiteral(_))
 
   // NB: we need to support relative paths, including `../foo`
   type FUPath = pathy.Path[_, pathy.Path.File, pathy.Path.Unsandboxed]
 
-  private val parser = new SQLParser()
+  private def parser[T[_[_]]: Recursive: Corecursive] = new SQLParser[T]()
```
mossprescott: (Responding to @wemrysi's comment just below, but the pertinent change is actually here).

Just to clarify: this is a fairly significant leak, not just wasteful allocation (30k per query, according to https://github.com/quasar-analytics/quasar/commit/20c1aa70c018dd534376ba41e54162404dc51473). The most practical fix is just to avoid constructing a new parser each time, so I like his suggestion.

core/src/main/scala/quasar/sql/package.scala
sellout: Yeah, seems totally reasonable to choose the type exactly once and hold onto the value.

wemrysi:
Other than the parser comment, LGTM

alissapajer:
@sellout `Let` PR just landed - so we can land this now whenever it's ready. Yay!

core/src/main/scala/quasar/sql/package.scala
sellout: Just realized we currently use both Mu and Fix (because of this PR). Might switch back to just using Fix everywhere, or might make two parser `val`s.

sellout:
@wemrysi Ok, made the parser change. I know you signed off, but just give it a quick look to see if it seems reasonable.

core/src/main/scala/quasar/sql/package.scala
```diff
@@ -50,12 +50,10 @@ package object sql {
 
   private def parser[T[_[_]]: Recursive: Corecursive] = new SQLParser[T]()
```
wemrysi: Constraints on `T[_[_]]` should be `Corecursive : FunctorT` to match `SQLParser` no?

core/src/main/scala/quasar/package.scala
```diff
@@ -45,7 +45,8 @@ package object quasar {
   type SaltedSeqNameGeneratorT[F[_], A] = ReaderT[SeqNameGeneratorT[F, ?], String, A]
 
   /** Returns the `LogicalPlan` for the given SQL^2 query. */
-  def queryPlan(query: Expr, vars: Variables): CompileM[Fix[LogicalPlan]] = {
+  def queryPlan(query: Fix[Sql], vars: Variables)(implicit RT: RenderTree[Fix[Sql]]):
```
wemrysi: I think this could still be parametric in `T[_[_]]`, can't it?

sellout: No, because `substVars` no longer is.

core/src/main/scala/quasar/sql/package.scala
sellout: Yep, I’ll fix that.

core/src/main/scala/quasar/planner.scala
```diff
@@ -35,18 +35,17 @@ trait Planner[PhysicalPlan] {
     EitherT[WriterResult, E, A]((Vector(PhaseResult.Detail(name, plan.toString)), \/-(a)))
   }
 
-  def compileToLP(implicit RA: RenderTree[PhysicalPlan]):
-      QueryRequest => EitherT[(Vector[quasar.PhaseResult], ?), CompilationError, Fix[LogicalPlan]] = { req =>
+  def compileToLP(query: Fix[Sql], variables: Variables)(implicit RA: RenderTree[PhysicalPlan]):
```
wemrysi: Does this actually make use of the `RenderTree[PhysicalPlan]`? Also looks suspiciously similar to `queryPlan`, possible to replace with it? (If this is outside the scope of this PR and you want to hold off, that's fine, I just happened to notice now).

core/src/main/scala/quasar/package.scala
wemrysi: Ah, ok.

core/src/main/scala/quasar/planner.scala
sellout: You’re right, it doesn’t use `RA`. This is that bit of code that got lost in the move to Free, where we no longer short-circuit when we have a constant plan. This code is actually still used in tests. I have a branch that restores this behavior and eliminates this function, but it got mired in trying to make the types stricter (`Data \/ LogicalPlan` rather than just noticing that `LogicalPlan` is a `Constant`). I should back out that bit of it and just make it work for now.

sellout:
Ok, updated.

wemrysi:
LGTM, :+1:

PR #1175, opened 2016-04-24
Convert the Func case classes into an enumeration.

sellout:
They all had the same shape, so distinguishing them as subtypes wasn’t
useful to us.

This seemed like it’d be more of a win when it popped into my head earlier. As it is, I think it’s still better (it caught a few unnecessary cases of relying on subtyping and an `isInstanceOf`), but overall ¯\\\_(ツ)_/¯.


core/src/main/scala/quasar/functions.scala
```diff
@@ -21,22 +21,37 @@ import quasar.Predef._
...
-  def domain: List[Type]
-
-  /** This handles rewrites that constant-folding (handled by the typers) can’t.
-    * I.e., any rewrite where either the result or one of the relevant arguments
-    * is a non-Constant expression. It _could_ cover all the rewrites, but
-    * there’s no need to duplicate the cases that must also be handled by the
-    * typer.
-    */
-  def simplify: Func.Simplifier
+sealed trait Style
```
mossprescott: Not crazy about `Style` for a name, just because it implies formatting or at least something superficial. We used to have `MappingType` for a similar concept (see https://github.com/quasar-analytics/quasar/commit/61197ae620c4da336988ceca32e15912b4e15919). Other possibilities: `FuncType`, or any of the usual synonyms for "type"; `Form`, `Nature`.

core/src/main/scala/quasar/functions.scala
```diff
@@ -21,22 +21,37 @@ import quasar.Predef._
...
+/** A function that reduces a set of values to a single value. */
```
mossprescott: These are no longer functions, but the types/styles of functions, so maybe edit the comments to reflect that? "Style for functions that ..."

core/src/main/scala/quasar/functions.scala
```diff
@@ -21,22 +21,37 @@ import quasar.Predef._
...
+final case object Reduction extends Style
```
mossprescott: I thought we weren't using `final` for case objects, but I could be wrong.

core/src/main/scala/quasar/functions.scala
```diff
@@ -69,37 +80,16 @@ trait FuncInstances {
...
-
-/** A function that operates on the set containing values, not modifying
-  * individual values. (EG, filter, sort, take)
-  */
-final case class Sifting(name: String, help: String, codomain: Type, domain: List[Type], simplify: Func.Simplifier, apply: Func.Typer, untype0: Func.Untyper) extends Func
-
-/** A function that operates on the set containing values, potentially modifying
-  * individual values. (EG, joins).
-  */
-final case class Transformation(name: String, help: String, codomain: Type, domain: List[Type], simplify: Func.Simplifier, apply: Func.Typer, untype0: Func.Untyper) extends Func
```
mossprescott: Yeah, this was gross.

core/src/main/scala/quasar/mra.scala
```diff
@@ -243,14 +243,14 @@ object MRA {
...
-        case Transformation(_, _, _, _, _, _, _) => d
-        case Squashing(_, _, _, _, _, _, _)      => d.squash
+      func.style match {
+        case Mapping        => d
+        case Expansion      => d.expand
+        case ExpansionFlat  => d.flatten
+        case Reduction      => d.aggregate
+        case Sifting        => d
+        case Transformation => d
+        case Squashing      => d.squash
```
mossprescott: We're still "maintaining" this code, huh? Maybe just delete it instead, unless we think it can be refactored into the thing we actually need.

core/src/main/scala/quasar/compiler.scala
```diff
@@ -467,7 +467,7 @@ object Compiler {
         t match {
           case InvokeF(set.GroupBy, List(src, structural.MakeArrayN(keys))) =>
             Some(keys.map(_.transCataT(t => if (t ≟ src) newSrc else t)))
-          case InvokeF(Sifting(_, _, _, _, _, _, _), src :: _) =>
+          case InvokeF(Func(Sifting, _, _, _, _, _, _, _), src :: _) =>
```
mossprescott: Easier to write with `if func.style === Reduction` now?

core/src/main/scala/quasar/compiler.scala
```diff
@@ -487,7 +487,7 @@ object Compiler {
       def strip(v: Cofree[LogicalPlan, Boolean]) = Cofree(false, v.tail)
 
       t => t.tail match {
-        case InvokeF(func @ Reduction(_, _, _, _, _, _, _), arg :: Nil) =>
+        case InvokeF(func @ Func(Reduction, _, _, _, _, _, _, _), arg :: Nil) =>
```
mossprescott: Similar.

core/src/main/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -480,7 +480,7 @@ object MongoDbPlanner extends Planner[Crystallized] {
         }
       }
 
-      def reversibleRelop(f: Mapping): Output =
+      def reversibleRelop(f: Func): Output =
```
mossprescott: Does this make sense for non-Mapping funcs? If so, why is it calling `relMapping`?

mossprescott:
Lots of nit-picks, but really other than not liking `Style` as a name, this seems sensible and a nice little improvement.

core/src/main/scala/quasar/compiler.scala
sellout: Ah, yes. I just did a project-wide find/replace and fixed what didn’t compile, but yeah, that’s a better option.

core/src/main/scala/quasar/mra.scala
sellout: Yeah, I’ll delete it.

core/src/main/scala/quasar/functions.scala
sellout: What about something like `DimensionalEffect`? I don’t really like that particular name, but these categories really describe how the identity/dimensionality is affected by the function.

core/src/main/scala/quasar/functions.scala
sellout: I think we’ve been somewhat consistent about using it. I just never type `case` without a leading `final` 😉

sellout:
Ok, back in review!

core/src/main/scala/quasar/functions.scala
```diff
@@ -21,22 +21,41 @@ import quasar.Predef._
...
-  def help: String
+sealed trait DimensionalEffect
+/** Describes a function that reduces a set of values to a single value. */
+final case object Reduction extends DimensionalEffect
+/** Describes a function that expands a compound value into a set of values for
+  * an operation.
+  */
+final case object Expansion extends DimensionalEffect
+/** Describes a function that expands a compound value into a set of values. */
+final case object ExpansionFlat extends DimensionalEffect
```
mossprescott: I realize you're just moving these comments, but now that I'm reading them I notice they're not very clear. Feel free to punt for now, but I'll comment on what I noticed:

The difference between `Expansion` and `ExpansionFlat` isn't very clear.

core/src/main/scala/quasar/functions.scala
```diff
@@ -21,22 +21,41 @@ import quasar.Predef._
...
+/** Describes a function that each individual value. */
+final case object Mapping extends DimensionalEffect
```
mossprescott: This comment actually makes no sense.

core/src/main/scala/quasar/functions.scala
```diff
@@ -21,22 +21,41 @@ import quasar.Predef._
...
+/** Describes a function that compresses the identity information. */
+final case object Squashing extends DimensionalEffect
```
mossprescott: I hope there's a better way of describing this one, because I don't think "identity information" has any clear meaning in this context.

core/src/main/scala/quasar/functions.scala
sellout: Yeah, I think it’s not even accurate. I’ll make a note to clarify that if they survive the qScript work.

core/src/main/scala/quasar/functions.scala
sellout: It’s ok, though. That’s the operation that everyone understands 😉

core/src/main/scala/quasar/functions.scala
sellout: Yeah, we’ll deal with all of this in the qScript work. They’ll probably be changing to some extent anyway.

Anything else in this PR?

mossprescott:
No other comments. LGTM with or without improved javadoc.

PR #1176, opened 2016-04-26
Add sbtopts

jedesah:


mossprescott:
Why not? I notice this file only takes effect if I use `./sbt` (and I usually use `sbt`), but if it helps you, I have no objection.

sellout:
I assume it takes effect if you use the sbt-extras script, which is what my system-wide `sbt` is. We only still have a copy in the repo at this point because Travis’ version of sbt-extras is too old to work with our `-no-imports` usage.

jedesah:
@mossprescott Weird, I use the system wide `sbt` and it looks like it takes effect for me.

PR #1177, opened 2016-04-28
expose mappings from config to a key and connectionUri

mossprescott:
...as used in the JSON codec.

Required by forthcoming PR for SD-1569.


wemrysi:
:+1:

PR #1178, opened 2016-05-02
avoid blocking in the default executor's thread pool

mossprescott:
Use `Task.after` rather than `Thread.sleep`.

When running on a single-core machine, scalaz's default executor
service uses a pool of exactly one thread. `waitForInput` was blocking
that thread continuously, so requests would just hang.

This fix was verified (applied to 8.0.1) on @damonLL's single-core Ubuntu image.

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+- [SD-1574] when running on a single-core, web server would hang and not respond to any request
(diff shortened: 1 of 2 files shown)


sellout:
LGTM. I’m going to declare it ready, just to get it out the door faster (hope I’m not stepping on your toes @drostron).

mossprescott:
Merged as 9.2.2

PR #1179, opened 2016-05-03
Backport the fix from 9.2.2.

sellout:
Fixes a case where we would hang on single-core machines.

### DIFF
--- a/scripts/publishJarOnUpstream
+++ b/scripts/publishJarOnUpstream
@@ -14,8 +14,8 @@ if [[ ! -v GITHUB_TOKEN ]] ; then
     exit 0
 fi
 
-# only publish on quasar-analytics/quasar#master
-if [[ "$TRAVIS" == "true" && "$TRAVIS_BRANCH" == "master" && "$TRAVIS_REPO_SLUG" == "quasar-analytics/quasar" ]] ; then
+# only publish on quasar-analytics/quasar
+if [[ "$TRAVIS" == "true" && "$TRAVIS_BRANCH" =~ '^(master$|backport/)' && "$TRAVIS_REPO_SLUG" == "quasar-analytics/quasar" ]] ; then
   "$SBT" 'project web'    githubRelease
 else
   echo "GITHUB_TOKEN defined, but Travis not running in quasar-analytics/quasar#master, so skipping publish"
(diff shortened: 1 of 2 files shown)


web/src/main/scala/quasar/server/Http4sUtils.scala
```diff
@@ -45,13 +45,11 @@ object Http4sUtils {
...
-        .handle { case _: java.lang.InterruptedException => () }
-      test <- Task.delay(Option(System.console).isEmpty || System.in.available() <= 0)
-        .handle { case _ => true }
-      done <- if (test) waitForInput else Task.now(())
-    } yield done
+
+    val inputPresent = Task.delay(Option(System.console).nonEmpty && System.in.available() > 0)
+                           .handle { case _ => false }
+
+    Process.eval(inputPresent.after(250.milliseconds)).repeat.takeWhile(!_).run
```
mossprescott: Wow, this is an old version. This fixes a memory leak, so that's not a bad thing.

mossprescott:
LGTM

scripts/publishJarOnUpstream
```diff
@@ -14,8 +14,8 @@ if [[ ! -v GITHUB_TOKEN ]] ; then
     exit 0
 fi
 
-# only publish on quasar-analytics/quasar#master
-if [[ "$TRAVIS" == "true" && "$TRAVIS_BRANCH" == "master" && "$TRAVIS_REPO_SLUG" == "quasar-analytics/quasar" ]] ; then
+# only publish on quasar-analytics/quasar
+if [[ "$TRAVIS" == "true" && "$TRAVIS_REPO_SLUG" == "quasar-analytics/quasar" ]] ; then
```
mossprescott: It's not obvious to me that this won't publish PRs, but I suppose we'll find out soon enough.

sellout: Yeah … I tried to find the values for these variables in PR builds, but they don’t seem to be reported in the build log.

sellout: Maybe we should name these branches like `backport/v4.5.6+`?

mossprescott: Indeed, it looks like Travis tried to run the script and the chmod bits weren't right.

I think `backport/` is a good idea. Makes it clear that it's not being actively developed.

PR #1180, opened 2016-05-03
Backport the fix from 9.2.2.

sellout:
Fixes a case where we would hang on single-core machines.

Take 2 – this time with a better branch name, and a (hopefully) correct Travis script.


sellout:
How does this decrease coverage?!

Also, @mossprescott … good?

mossprescott:
:+1:

PR #1181, opened 2016-05-05
SD-1601 Fix PUT/POST mount API regression

drostron:


web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -130,7 +131,16 @@ object mount {
                   (msg, _) => ApiError.fromMsg_(
                     BadRequest, msg).left))
       exists <- EitherT.right(M.lookup(path).isDefined)
-      mnt    =  if (replaceIfExists && exists) M.replace(path, bConf) else M.mount(path, bConf)
+      mnt    =
```
mossprescott: This doesn't feel like the right way to go. Fixing it here means that our `Mounting` interpreter is still broken. Requiring `MountConfigsF` and using `ViewMounter` directly couples this service to a particular `Mounting` interpreter. And I can't say for sure that `ViewMounter` even does the right thing when used this way.

I'd rather see it fixed in the interpreter (which is EvaluatorMounter?). If there's a reason it can't be fixed there, let's straighten that out. I looked, but this stuff is hard to follow.

Happy to discuss and/or help.

it/src/test/scala/quasar/server/ServiceSpec.scala
```diff
@@ -64,6 +64,47 @@ class ServiceSpec extends mutable.Specification {
...
+        client.fetch(
+          Request(
+            uri = baseUri / "mount" / "fs" / "a",
+            method = Method.GET)
+          )(Task.now)
+      }
+
+      r.map(_.status) must beRightDisjunction(Ok)
+    }
+
```
mossprescott: If these are meant to test nesting views under other mounts, then there should be some clue in the example name or at least a comment.

If the fix can be done in the `Mounting` interpreter instead (see my other comment) we can have tests at that level, which would be better IMO.

drostron: It is to test that POST works at all for a view. The regression caused it to always fail as far as I could tell. I'm open to improved naming.

Agreed. It would be great to be able to test this sort of thing at the `Mounting` interpreter level. Unfortunately the current way we are wiring interpreters together for the server vs tests causes some false test passes. This was the reason for creating `ServiceSpec` in the first place. Ideally we can rework things so that it isn't necessary. I'd like to take that on now, but am not sure it trumps our other priorities. _The rework would likely take a trivial amount of time._

web/src/main/scala/quasar/api/services/mount.scala
drostron: Agreed and share your concerns. I quickly rat holed when attempting to fix it more fundamentally.

It would be good to pair with you for a bit on this to see if we can find a relatively contained and improved fix.

drostron:
@mossprescott Pushed a commit adding a comment explaining the `MountConfigsF` workaround in `mount` as we discussed.

mossprescott:
I was hoping for comments on the code in the mount service as well; do you think this TODO will lead us to that code?

drostron:
I'm thinking it is, since relevant code will cease to compile due to the missing `MountConfigsF` effect evidence. I'm happy to add a few additional comments if you think it's helpful.

mossprescott:
OK I'm over it. LGTM

PR #1182, opened 2016-05-06
Break MongoDB into its own subproject.

sellout:
This creates three new subprojects – the intended MongoDB one; “main”,
containing the bits that tie core and the backends together; and “repl”,
containing … the repl.

This means that `core/run` is now `repl/run`.

There are some minor changes beyond moving those packages to their own subprojects:
- `repl.Prettify` is now `main.Prettify`, as it’s also used by `web`
- `CoreConfig` is tested in `main` because the tests require `mongodb`


sellout:
BTW, this reduces the number of files in core by 35%, so hopefully a corresponding reduction in compile times when working in core. 🍀

build.sbt
```diff
@@ -127,24 +127,63 @@ lazy val oneJarSettings =
...
+//        /   \
+// mongodb    «other backends»
+//        \   /
+//        main
+//        /  \
+//    repl    web
+//        \  /
+//         it
+//          |
+//        root
```
alissapajer: very nice diagram :)

alissapajer:
@sellout LGTM +1 to merge

PR #1187, opened 2016-05-17
Invoke LP nodes have Sized domain and arguments

alissapajer:


alissapajer:
@sellout the PR is ready for review

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -248,11 +252,24 @@ object LogicalPlan {
   val normalizeLetsƒ: LogicalPlan[Fix[LogicalPlan]] => Option[LogicalPlan[Fix[LogicalPlan]]] = {
       case LetF(b, Fix(LetF(a, x1, x2)), x3) => LetF(a, x1, Let(b, x2, x3)).some
 
-      case InvokeF(func, xs) =>
-        splitAtFirstMatch(xs) { case Fix(LetF(_, _, _)) => true; case _ => false }.map {
-          case (prefix, Fix(LetF(a, x1, x2)), suffix) =>
-            LetF(a, x1, Invoke(func, prefix ++ (x2 :: suffix)))
-        }
+      // TODO generalize the following three `GenericFunc` cases
```
alissapajer: outstanding TODO

core/src/main/scala/quasar/optimizer.scala
```diff
@@ -212,13 +213,31 @@ object Optimizer {
...
     }
 
-    implicit val ComponentFunctor = new Functor[Component] {
-      def map[A, B](fa: Component[A])(f: A => B) = fa match {
-        case EquiCond(run)    => EquiCond((l, r) => f(run(l, r)))
-        case LeftCond(run)    => LeftCond((l) => f(run(l)))
-        case RightCond(run)   => RightCond((r) => f(run(r)))
-        case OtherCond(run)   => OtherCond((l, r) => f(run(l, r)))
-        case NeitherCond(run) => NeitherCond(f(run))
+    // TODO add scalaz propery test
```
alissapajer: outstanding TODO

core/src/main/scala/quasar/package.scala
```diff
@@ -65,4 +68,41 @@ package object quasar {
       typechecked <- phase("Typechecked", LogicalPlan.ensureCorrectTypes(optimized).disjunction)
     } yield typechecked
   }
+
+  // TODO write generic impls (not using IndexedSeq) and put in fp package
```
alissapajer: outstanding TODO

core/src/main/scala/quasar/std/library.scala
```diff
@@ -14,89 +14,94 @@
...
-  protected implicit class TyperW(self: Func.Typer) {
-    def ||| (that: Func.Typer): Func.Typer = { args =>
+  protected implicit class TyperW[N <: Nat](self: Typer[N]) {
+    def ||| (that: Typer[N]): Typer[N] = { args =>
       self(args) ||| that(args)
     }
   }
 
-  def functions: List[Func]
+  // TODO exhaustiveness checking for each arity
```
alissapajer: outstanding TODO

alissapajer:
I marked the `TODO`s that I added - some may be out of the scope of this PR.

core/src/main/scala/quasar/std/library.scala
alissapajer: I think this is out of scope of this PR.

When we add a new library function, we are required to add it to the list of their appropriate arity, and if we forget to add the function to the list, it won't be recognized in the quasar compiler. So of course compiler tests would catch this bug, but I just want to make it as easy as possible (read: touch as few code places as possible) to support new functions. So this is still something I'm thinking about how to address.

core/src/main/scala/quasar/package.scala
alissapajer: @sellout I would be fine keeping this as is for this PR, unless you have reason to want to generalize these before landing them in master?

Ultimately I'd also like to PR them to [shapeless-contrib](https://github.com/typelevel/shapeless-contrib).

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -283,14 +301,11 @@ object LogicalPlan {
         success(ConstantF[Typed[LogicalPlan]](d))
 
       case InvokeF(func, args) => for {
-        types <- func.untype(typ)
-        args0 <- types.align(args).traverseU(_.onlyBoth match {
-          case Some((t, arg)) => inferTypes(t, arg).disjunction
-          case None =>
-            (SemanticError.wrongArgumentCount(func, types.length, args.length))
```
alissapajer: wrong argument count gets pushed earlier!

core/src/test/scala/quasar/optimizer.scala
```diff
@@ -158,24 +166,41 @@ class OptimizerSpec extends Specification with CompilerHelpers with TreeMatchers
...
+
+    implicit def ArbComponentInt2Int: Arbitrary[Optimizer.Component[Int => Int]] =
+      Arbitrary(Optimizer.NeitherCond((_: Int) => 3))
+
+    implicit def EqualComponent: Equal[Optimizer.Component[Int]] = new Equal[Optimizer.Component[Int]] {
+      def equal(a1: Optimizer.Component[Int], a2: Optimizer.Component[Int]): Boolean = true
+    }
+
+    "obey applicative laws" in {
+      applicative.laws[Optimizer.Component]
```
alissapajer: @sellout here are the implicits required to give us a scalaz property test for `Applicative[Component]`. 

The problem we hit is that the constructors for `Component` require a function as an argument, for example `run0: Fix[LogicalPlan] => A`. And we cannot create an `Equal` for a function1... so we're in an awkward place, and I'm not sure yet how we could rework `Component` to get around this issue...

core/src/main/scala/quasar/Predef.scala
```diff
@@ -57,6 +57,7 @@ object Predef extends LowPriorityImplicits {
 
   type Iterable[+A] = C.Iterable[A]
   type IndexedSeq[+A] = C.IndexedSeq[A]
+  type IS[+A] = C.IndexedSeq[A]
```
sellout: I would at least like to scope this a bit tighter. In general Predef should be getting smaller, and the name is opaque. I would maybe suggest a `def ArgList(args: _*) = Sized[IndexedSeq](args)` (in whatever the tighter scope is), then we don’t need the type alias at all (I think). And at that point, it could be `Sized[List]`, which would allow us to remove the `.toList` everywhere (since if we ever do change the type, it would likely be to some other concrete type that also has a functor/traverse/whatever).

core/src/main/scala/quasar/functions.scala
```diff
@@ -94,6 +125,15 @@ object Func extends FuncInstances {
     def apply[T[_[_]]: Recursive: Corecursive](orig: LogicalPlan[T[LogicalPlan]]):
         Option[LogicalPlan[T[LogicalPlan]]]
   }
-  type Typer      = List[Type] => ValidationNel[SemanticError, Type]
-  type Untyper    = (Func, Type) => ValidationNel[SemanticError, List[Type]]
+
+  type Input[A, N <: Nat] = Sized[IndexedSeq[A], N]
```
sellout: Ah, yeah, this looks like maybe a good place to define `ArgList`? Or maybe just `Input` to parallel this?

core/src/main/scala/quasar/compiler.scala
```diff
@@ -179,29 +181,52 @@ trait Compiler[F[_]] {
...
+        }
+
+    def findTernaryFunction(name: String): CompilerM[GenericFunc[nat._3]] =
+      library.functions.find(f => f.name.toLowerCase === name.toLowerCase).fold[CompilerM[GenericFunc[nat._3]]](
+        fail(FunctionNotFound(name))) {
+          case func @ TernaryFunc(_, _, _, _, _, _, _, _) => emit(func)
+          case func => fail(WrongArgumentCount(name, func.arity, 3))
+        }
+
+    def findNaryFunction(name: String, length: Int): SemanticError =
```
sellout: What if you moved the `fail` here and made it return `CompilerM[Unit]`, just for consistency?

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -139,15 +140,25 @@ object LogicalPlan {
...
       funcName + "(" + values.mkString(", ") + ")"
     }
+    override def equals(that: scala.Any): Boolean = that match {
+      case that @ InvokeF(_, _) =>
+        (this.func == that.func) && (this.values.unsized == that.values.unsized)
+      case _ => false
+    }
+  }
+  // we create a custom `unapply` to bypass a scalac pattern matching bug
+  // https://issues.scala-lang.org/browse/SI-5900
```
sellout: I would prefix this with a TODO or FIXME – easier to audit in the code later.

core/src/main/scala/quasar/package.scala
sellout: Yeah, I’m fine with this. I would just go from this right to type class instances in shapeless-contrib.

core/src/main/scala/quasar/std/agg.scala
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-package quasar.std
+package quasar
+package std
```
sellout: We don’t use this package style – prefer to explicitly import what we need.

core/src/main/scala/quasar/std/library.scala
sellout: I agree that this is out of scope.

This approach should be changed in general. The names of the functions to look up are really specific to SQL. The mapping of those strings to these vals should be handled in the `sql` package, not tied to the Funcs. I.e., these lists should go away.

core/src/main/scala/quasar/functions.scala
```diff
@@ -47,39 +48,69 @@ object DimensionalEffect {
...
+final case class UnaryFunc(
+    val effect: DimensionalEffect,
+    val name: String,
+    val help: String,
+    val codomain: Func.Codomain,
+    val domain: Func.Domain[nat._1],
+    val simplify: Func.Simplifier,
+    val typer0: Func.Typer[nat._1],
+    val untyper0: Func.Untyper[nat._1]) extends GenericFunc[nat._1] {
+  def apply0[A](a1: A): LogicalPlan[A] = apply(Sized[IS](a1))
```
sellout: I feel like these should be `apply`, and the `apply` in `GenericFunc` should be something else. But I’m not sure how often the latter is used. The former are used a ton in the tests, at least..

sellout:
Ok, a few comments. Nothing too significant. Back to you.

core/src/main/scala/quasar/package.scala
alissapajer: Yeah that progession makes sense - I'll change the TODO comment.

core/src/main/scala/quasar/std/agg.scala
alissapajer: ah! I did this in a handful of files - I'll revert them back!

alissapajer: @sellout I noticed in some places we do `import quasar._` and in other places `import quasar.{Foo, Bar, Baz}`. Does this project prefer one over the other?

core/src/main/scala/quasar/functions.scala
alissapajer: makes sense - I removed the `IS` type alias in predef, and added "constructor functions" for `Input`, and additionally changed the type `IndexedSeq` to `List` since we were often requiring a list anyways.

core/src/main/scala/quasar/Predef.scala
alissapajer: +1 `.toList` is removed from everywhere now

core/src/main/scala/quasar/compiler.scala
alissapajer: @sellout I made it return `CompilerM[Fix[LogicalPlan]]` since that's what the use case calls for - but I'm not sure that's actually what we want. 

As I think about it, when we return a `CompilerM[_]` we lose the information at the type level that `findNaryFunction` always returns an error. Even if it returns `CompilerM[Unit]`, we still don't know anymore based on the type that we always get a `SemanticError`, and we could legally get a success with value `Unit`. So that is why I had it the way it was originally. What do you think?

core/src/main/scala/quasar/functions.scala
alissapajer: Ok I reworked this as so. In `GenericFunc` we have the following, which has a call site in the compiler.
```scala
def applyGeneric[A](args: Func.Input[A, N]): LogicalPlan[A] =
    LogicalPlan.InvokeF[A, N](this, args)
```
And then in each of the unary, binary, ternary case classes we have this or the analogous:
```scala
  def apply[A](a1: A, a2: A): LogicalPlan[A] =
    applyGeneric(Func.Input2[A](a1, a2))
```
I like this because it exposes a nice `apply` constructor which creates the `Sized` for you, so you can just pass 1, 2, or 3 arguments cleanly, and it puts them into the sized list for you.

There were some call sites in tests for `apply0` that were easy to change by reverting the file :)

alissapajer:
@sellout I pushed a few commits addressing your comments. Back to you.

core/src/test/scala/quasar/optimizer.scala
```diff
@@ -158,24 +165,41 @@ class OptimizerSpec extends Specification with CompilerHelpers with TreeMatchers
...
+      ko // TODO here so we don't forget that this test is basically testing nothing
```
alissapajer: @sellout this test isn't being run in the build, or locally when we do `sbt test`. I need to investigate what's up.

alissapajer: never mind it's because I accidentally deleted it

core/src/test/scala/quasar/optimizer.scala
sellout: Yeah, that’s a problem. Definitely outside the scope of this PR, I think.

Also, you can replace your Arbitrary instances with one that looks like
```scala
implicit def componentArbitrary[A: Arbitrary]: Arbitrary[Optimizer.Component[A]] =
  Arbitrary.arbitrary[A] ∘ (Optimizer.NeitherCond(_))
```

sellout: Oh, I’d also add a FIXME here, mentioning that this test doesn’t test much yet because we don’t have a strongly normalizing language 😉

alissapajer:
@sellout ok the `Arbitrary` change is pushed. let me know how other other stuff looks!

sellout:
Argh, sorry – I think this looks good. Just need to resolve conflicts.

alissapajer:
@sellout I have resolved all the conflicts against master.

sellout:
Cool. Squash (if you plan to) and merge!

PR #1188, opened 2016-05-18
Add dependency on si2712fix-plugin.

sellout:
This has allowed a few things so far:

- removing `implicitly` from our Predef,
- replacing `traverseU`, etc. with `traverse`, and
- removing various implicit parameter lists from calls.

There are additional improvements we can make with this, but we should
address them as we come across them.

Unfortunately, also ran into an ambiguous `monadMTMAB` implicit, so
added TODOs around getting rid of that.

Also,

- added a `RenderTree` instance to resolve natural transformation
  instances,
- replaced some `.map(...).sequence` with `.traverse(...)`,
- updated Matryoshka to 0.10.2 (which also has the SI-2712 fix), and
- replaced `CoproductN` with `:+:` syntax (which seems to infer better
  for some reason).

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+- Added a dependency on the SI-2712 fix.
--- a/build.sbt
+++ b/build.sbt
@@ -45,8 +45,9 @@ lazy val commonSettings = Seq(
     "JBoss repository" at "https://repository.jboss.org/nexus/content/repositories/",
     "Scalaz Bintray Repo" at "http://dl.bintray.com/scalaz/releases",
     "bintray/non" at "http://dl.bintray.com/non/maven"),
-  addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.7.1"),
-  addCompilerPlugin("org.scalamacros" % "paradise"       % "2.1.0" cross CrossVersion.full),
+  addCompilerPlugin("org.spire-math" %% "kind-projector"   % "0.7.1"),
+  addCompilerPlugin("org.scalamacros" % "paradise"         % "2.1.0" cross CrossVersion.full),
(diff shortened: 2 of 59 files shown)


drostron:
Nice to see si2712fix-plugin in action.

wemrysi:
@sellout Taking inspiration from [mandubian/freek](https://github.com/mandubian/freek/blob/master/src/main/scala/FX.scala), what do you think of the following syntax? Does it still allow inference to work like your version of `:+:` does?
```scala
sealed abstract class CP {
 type l[_]
}

sealed abstract class :\:[L[_], R <: CP] extends CP {
  type l[A] = Coproduct[L, CP#l, A]
}

sealed abstract class :\/:[L[_], R[_]] extends CP {
  type l[A] = Coproduct[L, R, A]
}

abstract class Foo {
  type F[_]
  type G[_]
  type H[_]

  type M[A] = (F :\: F :\: G :\/: H)#l[A]
}
```

sellout:
I think that looks good. I’ll give it a try. Seems like it should work the same. And not conflict with `scalaz.:+:` to boot.

What about just using freek?

wemrysi:
freek is a possibility, some of it uses cats types, but the interesting stuff (`CoproductK`/`FX`) doesn't. I have tried it at all yet, but it looks nice. Probably makes sense to do that in a separate PR, though.

sellout:
I tried a few variants,
```scala
sealed abstract class CP {
  type F[_]
  type G[_]
  type λ[A] = Coproduct[F, G, A]
}
```
seemed to be the most promising, but I still kept getting type mismatches everywhere. I think we need all of `CoproductK` for this. So, maybe just try to use freek directly later?

wemrysi:
Sadface. LGTM otherwise.

PR #1189, opened 2016-05-18
Add `onFail` to `quasar.effect.Failure`

jedesah:


core/src/main/scala/quasar/effect/Failure.scala
```diff
@@ -47,6 +48,9 @@ object Failure {
         e => f(Some(e)) *> fail(e),
         a => f(None)    as a))
 
+    def onFail[A](fa: F[A], f: E => F[Unit]): F[A] =
+      onFinish(fa, _.cata(f,().pure[F]))
```
mossprescott: Seems like there would be a handier way to do this, along the lines of `whenM`, but I can't find it.

mossprescott:
LGTM

PR #1190, opened 2016-05-19
Get rid of our subproject build files.

sellout:
It‘s too easy to overlook stuff that is implicitly referenced. Anything
outside of the build.sbt should be in Scala files that are imported.


drostron:
Appreciate seeing those subproject build files go away, 👍🏼.

PR #1191, opened 2016-05-20
SD-1643 Publish to sonatype

drostron:


build.sbt
```diff
@@ -145,27 +186,33 @@ lazy val oneJarSettings =
 
 lazy val root = project.in(file("."))
   .settings(commonSettings: _*)
-  .aggregate(repl, web, it)
+  .settings(noPublishSettings)
+  .aggregate(core, main, mongodb, repl, web, it)
```
sellout: 😵 Is _this_ why coverage broke?

build.sbt
```diff
@@ -145,27 +186,33 @@ lazy val oneJarSettings =
...
   .enablePlugins(AutomateHeaderPlugin)
 
 lazy val core = project
+  .settings(name := "quasar-core")
   .settings(oneJarSettings: _*)
+  .settings(publishSettings: _*)
   .settings(
     libraryDependencies ++= Dependencies.core,
+    publishArtifact in (Test, packageBin) := true,
```
sellout: We shouldn’t publish `core` or `mongodb`. These are things internal to Quasar. E.g., we don’t want a backend to be able to depend on `core`.

I know that it’s common for a project to have `-core` as its primary archive. Which makes me think we should rename `main` to `core` and come up with a new name for `core`. I wanted to do this anyway, because the name `main` is confusing as hell, since it’s also used in paths as `src/main/scala`.

build.sbt
drostron: Maybe? Seems like codecov has been a bit flakey in general though.

sellout: Well, the problem hasn’t been with codecov – our drop from 80% to 6% is because scoverage is claiming there’s no coverage and not generating reports for most subprojects.

drostron: Ah ok, that aspect seems related.

sellout:
This also means we have to start thinking about the newly-exposed APIs when we version. Just something for us all to be aware of. Yet another reason to not expose core (which should be renamed) and mongodb. Then almost any change becomes at least a minor bump.

build.sbt
```diff
@@ -26,6 +26,7 @@ lazy val checkHeaders =
   taskKey[Unit]("Fail the build if createHeaders is not up-to-date")
 
 lazy val commonSettings = Seq(
+  organization := "quasar",
```
mossprescott: Should be `org.quasar-analytics`, no?

build.sbt
```diff
@@ -80,6 +81,46 @@ lazy val commonSettings = Seq(
       sys.error("headers not all present")
   })
 
+lazy val publishSettings = Seq(
+  organizationName := "SlamData Inc.",
```
mossprescott: Is there a name for the Quasar project, as distinct from SlamData?

build.sbt
drostron: Haha, folks are starting to comment before I placed some questions I had on the PR. Nice!

I was thinking `quasar-analytics` for simplicity but suppose I'm ok with `org.quasar-analytics`.

build.sbt
drostron: OK. I needed to publish them for downstream consumption likely due to depending on `core` instead of `main`. I'll check and see if `core` aka `main` is sufficient.

Renaming `main` to `core` sounds good.

build.sbt
drostron: Opinions?

build.sbt
mossprescott: I don't follow. We have `mongodb -> core` and `main -> mongodb` at the moment. Wouldn't a new backend depend on `core` also? Other derived projects would obviously depend on what's now called `main`.

If we want to rename so the "main" aritfact is called `core`, that's fine by me, and I agree we should publish only that one and `web` for now.

build.sbt
drostron: Also, something I'm interested in knowing.

build.sbt
```diff
@@ -80,6 +81,46 @@ lazy val commonSettings = Seq(
...
+  releaseCrossBuild := true,
+  autoAPIMappings := true,
+  scmInfo := Some(
+    ScmInfo(
+      url("https://github.com/quasar-analytics/quasar"),
+      "scm:git@github.com:quasar-analytics/quasar.git"
+    )
+  ),
+  developers := List(
+    Developer(
```
drostron: Is there a Quasar project equivalent here or is this sufficient?

build.sbt
sellout: Yeah, if it’s not sufficient, that’s stuff that we should move to `main` or elsewhere. E.g., we’re going to pull out a `backend` submodule that contains fs, EJson, QScript, etc. (that is, anything that a backend implementer is going to need access to). Any interpreter is going to need access to probably all of that, too. So, that’s something else to publish that QA will need to depend on.

build.sbt
mossprescott: The quasar-analytics homepage has a SlamData copyright on it, so perhaps not.

build.sbt
mossprescott: I thought the rule was you had to control the corresponding domain name to use it, but compliance seems to be uneven in the Scala world.

sellout:
I’m very happy to see this PR, but we may need some more modularization before it’s done. Enumerating the bits of Quasar that QA requires would be helpful.

build.sbt
wemrysi: I'm down with all of what is described here, but we may need to defer until we can further decouple things as quasar-advanced depends on all sorts of things inside what is now `core` and we'll need to publish it if we want to depend on it.

sellout:
We should probably get subtasks into Jira covering these steps.

build.sbt
mossprescott: Aren't we just going to depend on `quasar-web` and get the core stuff indirectly? Or are there other pieces, such as maybe the arbitraries?

sellout: If what @mossprescott suggests works, than that seems reasonable. Otherwise, I would suggest publishing core under some other name like `quasar-qa-dependency`, and not publish anything as `core` until everything is sorted out.

And does qa need to depend on `mongodb`? That seems more problematic to me.

wemrysi: Agreed, that sounds like a plan. It looks like QA only depends on `mongodb` indirectly through `main`.

drostron: After switching QAdv's dependency to `quasar-main` it would still depend on `quasar-mongodb` due to it being a dependency of `quasar-main`.

sellout: @drostron But we don’t need to _publish_ `quasar-mongodb` for that, right? I _think_ I only care about direct dependencies. Is there still a problem that you could potentially `import quasar.physical.mongodb._`? But we’ll handwave around that and just say you _shouldn’t_ if it’s not in a subproject you explicitly depend on.

drostron: SBT resolves transitive dependencies and `quasar-mongodb` is a transitive dependency of `quasar-main` currently. Also `quasar-web` currently depends on `quasar-main`.

sellout: Are you saying that if someone makes a new project that uses Quasar, they could just add an SBT dependency on Quasar, then write some code that did `import scalaz._` without adding a dependency on scalaz? That sounds terrible. I know that’s the case within an SBT project, but I had hoped library dependencies were a bit more encapsulated. Although, I guess if it’s not, then we just have the same handwavy problem anywhere, so I guess that’s fine ¯\\\_(ツ)_/¯

Also, let’s distinguish between the subprojects, like `main` and the published versions like `quasar-main`. So, we do not publish `mongodb` so there is no `quasar-mongodb`.

drostron: Yes, for better or worse, you can `import scalaz._` in the case you describe for library dependencies. I think you're hunting for OSGi 😉 . There might be hope in Project Jigsaw, but I haven't dug into it enough yet to say for sure.

That distinguishing will be useful in discussion.

I'll post a quick description of what QAdv depends on and we can go from there.

sellout: Ok, cool. Well, since that problem exists even across libraries, I’m happy to only publish `main` (as `quasar-core`, of course) and let QA “cheat” by depending on bits of `core` (unpublished) until we get things properly modularized.

mossprescott: Yay, Java!

Actually, you can improve this situation for compiles if you structure things in the right way; for example we could include only the algebras in the library that downstream projects compile against, and not the interpreters. And you can isolate things at runtime by doing Classloader hacks, but yes, this is all supposed to get better in Java 9 (or is it 10?)

Anyway all these issues are not new and don't ned to be addressed right now, I don't think.

drostron: I wasn't clear. The `scalaz` import isn't available due to being included in the Quasar published artifact it's available due to the transitive dependency pulling it into the classpath. To clarify, if we publish `main` as `quasar-core` we'll also have to publish `mongodb` as `quasar-mongodb` _(or some other name)_ so that transitive dependency resolution succeeds. `main` depends on `mongodb` therefore `quasar-core` depends on `quasar-mongodb` and must be published for said resolution to succeed. Maybe this is acceptable if we have the above suggested naming convention?

sellout: I thought that since `main` depends on `mongodb`, then all of the `mongodb` classes end up included in the `quasar-main` JAR (using the old name for consistency), so we don’t _need_ library-level transitive dependencies. But I’m going to guess that’s not true.

Oh! Can we use `.aggregate` to do that?

drostron: Correct, not true.

`.aggregate` aggregates tasks and not dependencies. There might be some other sbt technique to achieve it that I'm unaware of.

drostron:
Summary of QAdv's ProjectRef _(current)_ and Ivy style Quasar dependencies _(proposed)_:

##### ProjectRef

QAdv core → Quasar core

QAdv web → Quasar web → Quasar main → Quasar mongodb


##### Ivy

Naming as of aef9e6bafe5b7c90b314ab136a3eeb04f1b21ce2.

QAdv core → quasar-core

QAdv web → quasar-web → quasar-main → quasar-mongodb

sellout:
Sorry that I’m still so confused about this, but since `mongodb` (and `main`) depends on `core`, don’t we also have a transitive dependency on `quasar-core` from `quasar-mongodb`? Or was it just not important in the diagram since we’re already depending on `quasar-core` from QA `core`?

drostron:
Not sure I follow your question accurately. `quasar-core` is depended by `quasar-mongodb` but does not depend on it. In that sense it's a transitive dependency _(direction agnostic)_. In the dependency resolving sense, `quasar-core`'s pom doesn't refer to `quasar-mongodb` but `quasar-mongodb`'s pom refers to `quasar-core`.

sellout:
You answered my question with your diagram edit – i.e., that’s what I thought.

build.sbt
mossprescott: I think this is OK. The project is independent and SD is the primary developer.

drostron:
So, current project organization of quasar says that we need to publish the artifacts currently named  `quasar-web`, `quasar-main`, `quasar-mongodb`, `quasar-core` to satisfy QAdv. Do we need to modify Quasar's project structure in some way before this can move forward or would a pass at renaming the artifacts, in line with suggestions above, suffice until we have time for some restructuring?

sellout:
Ideally, I’d like to rename the subprojects and packages along with the artifacts. I don’t think it’s be much noise – Git[Hub] should identify those as renames pretty easily.

drostron:
Rename proposal:

project
artifact
package

core → feni
quasar-core → quasar-internal-feni
quasar → quasar _(NB: package doesn't currently reflect project name)_

main → core
quasar-main → quasar-core
quasar.main →  quasar.core

mongodb → mongodb
quasar-mongodb → quasar-internal-mongodb
quasar.physical.mongodb → quasar.physical.mongodb

feni is for Fe and Ni the elements that make up Earth's inner core. Please offer up a more descriptive name.

mossprescott:
After mulling this over a bit, I disagree about the rename of `core`.

I think the spirit of `core` for any project is that it should define the essential types and interfaces of the project, but not the pieces needed to interface with specific outside systems. Every typical "downstream" project will always depend on `core`. For quasar, those dependent projects might define new backends _or_ front-ends.

I don't really see the problem with publishing `quasar-mongodb`. If down the road someone wants to assemble a quasar-based project combining our mongodb backend with some other backend(s), they would want to depend on `quasar-mongodb`, not `quasar-main`.

If we want to pull things that are needed only by backend implementers as `quasar-backend`, I have no objection, but we can do that later. The only likely cost is the potential of breaking dependent projects, which we can live with for now.

I agree `main` isn't a great name, and I propose we simply eliminate that project in the following way:
- move `Prettify` into `core`, or if preferred, into a separate `util` project.
- refactor the package object's code so that the few occurrences of `mongodb` types are pulled out as parameters, and find a place for this code in `core`. If that leads to any duplication of code between `web` and `repl`, then we could put the common bits in `mongodb`. The goal would be to have the code to compose an interpreter from a arbitrary set of backends be so trivial that it doesn't merit a (package) object of its own.

mossprescott:
Following up on my own comment:

Although I think this project organization discussion is important, I'd like to see it happen separately from the pragmatic goal of improving build times and build stability for our downstream projects. My hope is that there's a way to move forward that doesn't commit us too much to a setup that we don't like.

I toyed with a couple of possible approaches to project names and versions, but it's not really clear to me what problem needs to be solved so I think we need to discuss this some more.

sellout:
So, the problem for me is that when we start publishing this stuff, we now have a very broad interface that is versioned together, when really, there are three distinct APIs that should be versioned distinctly (the frontend API used by SQL², the backend API used by MongoDB, and the client API, used by the REPL and Web … and the interpreter API(?) used by QAdv(?) – is this just the same as the Repl/Web API?). Basically – we don’t want breaking version increments on the backend API just because we changed a /server/info endpoint, or vice versa.

I feel like we could give quasar-mongodb, etc. names that indicate that we’re not actually versioning them yet or something (like quasar-mongodb-internal), so our current versions continue to reflect only the “main” API. This way we can get the pragmatic benefits immediately, without changing how we handle versioning.

sellout:
Re: core rename.

Neither backends nor frontends will depend on core. They’ll depend on subprojects (like the quasar-backend you mentioned) that provide only the relevant interfaces (frontends will see LP & EJSON; backends will see QSCRIPT, EJSON, fs). So that was the motivation to rename main to core – backends will depend on quasar-backend, frontends on quasar-frontend, and then REPL, Web, etc. on … well, quasar-core. If we can merge/eliminate `main`, then yeah, sounds good. That serves the same goal of clients depending on quasar-core.

But it sounds like you plan to push the code that brings frontend/core/backend together down into REPL and Web? (Since you mentioned having backends depend on core, so then core can’t depend on backends – but I probably misunderstood something, because I’m not very familiar with what’s in main in the first place.) Making it trivial to do something like

```scala
quasar.core.init(
  Map(
    "SQL" -> quasar.sql.Frontend,
    "LP" -> quasar.logicalplan.Frontend,
    "Datalog" -> quasar.datalog.Frontend),
  Map(
    "MongoDB" -> quasar.physical.mongodb.Fs,
    "Spark" -> quasar.physical.spark.Fs))
```
would be awesome.

There isn’t a particular problem with publishing quasar-mongodb. I expect it to be a separate repo, like any other backend, eventually. The problem is just in broadening the public API to include a lot more stuff, and therefore affect versioning as I mentioned in the previous comment. If it’s a separate repo, or somehow separately versioned, then cool.

mossprescott:
Thanks, that's clarified things for me.

I think we're in agreement that the end state looks like multiple projects in multiple repos, each with a version tracking its own API's history. That's definitely not something we can take on right now.

In the interim, I like the idea of calling out the components that aren't yet meant to be taken seriously as stand-alone libraries, and I can't think of a better way than adding `-internal` to the name. In the README where we give the sbt dependency name we can make a point of warning people about those components not being tracked by semver.

That being the case, are you still uncomfortable with publishing `quasar-core-internal` and `quasar-main-internal` in their current state? Or does the "internal" disclaimer give us some leeway to put that issue off until later?

build.sbt
jedesah: @sellout @drostron Isn't that's what `oneJar` does. But `oneJar` unless otherwise configured is going to include EVERYTHING in that jar including `scalaz`, `monocle`, etc. I don't think it's really intended for this situation, it's more about making it easier to deploy which is how we use it already.

sellout:
Yeah, I’m fine with publishing `-internal` and not changing the scope of our semver.

build.sbt
jedesah: @sellout I am not sure I am sold on making `core` purely internal. Once we remove backend related stuff into a `backend` subproject, `fp` related stuff into some kind of `fp` library and the `sql` parser into it's own sub-project or project. What's left, among other things, is config and filesystem stuff. Quasar Advanced and future similar projects would still need to depend on that stuff and I think it's perfectly reasonable to do so. I guess I am just not convinced that there will be enough stuff that truly should remain entirely internal to warrant using a subproject for encapsulation as opposed to just standard Scala encapsulation primitives where appropriate.

Also, I would advise against creating too many sub-projects unless there is a specific reason to do so as it will add some overhead that may not be necessary.

mossprescott: Yes, but, oneJar also creates a kind of frankenjar with all the dependency jars embedded inside it, then does ClassLoader magic at runtime to unpack them. That's good for deployment, as you say, but not good for building a library other projects can use in a simple way.

What we _might_ want is more like what `sbt-assembly` does (just adds the contents of the dependency jars to the project's jar). But it's pretty clearly not intended to do what we're proposing here, and I suspect it would be a drag to try to force it to do what we want.

jedesah:
@mossprescott Why multiple repos? I mean for external projects of course, but are you suggesting that each of our current sub-project in `quasar` should eventually have it's own git repo? I am not sure I see the benefit and I can certainly remind myself the pains of such a setup. Just the fact that quasar-advanced is a separate repo is painful enough, of course that has a very specific benefit, although if we were using a vcs that supported making a certain part of the repo private, I would certainly entertain the idea of merging them together.

drostron:
Does this capture the result of the renaming discussion? Do we care about adding `internal` to package naming?

project: core → coreInternal
artifact: quasar-core → quasar-core-internal
package: quasar → quasar.core.internal (NB: package doesn't currently reflect project name)

project: main → mainInternal
artifact: quasar-main → quasar-main-internal
package: quasar.main → quasar.main.internal

project: mongodb → mongodbInternal
artifact: quasar-mongodb → quasar-mongodb-internal
package: quasar.physical.mongodb → quasar.physical.mongodb.internal

sellout:
Nah, `-internal` should only affect the artifacts, not the project or package name, IMO.

drostron:
@sellout Sounds good. I was taking one of your earlier comments on renaming packages and projects too far 😄 . Edited the above rename comment, will make that change, and place this _in review_.

mossprescott:
@jedesah My assumption was that tracking different versions for sub-projects in the same repo would be unworkable, either for SBT reasons or git reasons. But yes, there's an obvious cost so I'd love to be proved wrong on that.

scripts/constants
```diff
@@ -16,12 +16,12 @@ mkdir -p "$TEMP_DIR"
 # quasar version, from build.sbt:
 QUASAR_VERSION=$(cat $WS_DIR/version.sbt | sed -n -e 's/^ *version.*:= "\(.*\)"[^\n]*/\1/p')
 
-QUASAR_WEB_JAR="web_2.11-$QUASAR_VERSION-one-jar.jar"
+QUASAR_WEB_JAR="quasar-web_2.11-$QUASAR_VERSION-one-jar.jar"
```
mossprescott: This name also appears in the README.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "9.3.1"
+version in ThisBuild := "9.3.2"
```
mossprescott: Why not `CHANGELOG/bugfix.md`?

scripts/constants
drostron: Will update, thanks.

version.sbt
drostron: Had my QAdv hat on momentarily, switching back.

jedesah:
@mossprescott That's a good point. I am also not sure how easily we could automate our current merge and release process if everything were in the same repo. But glad we agree that that would be nice.

sellout:
@jedesah We are at least forced to have separate repos for “things that 3rd-party front/backends depend on” and “things that depend on 3rd-party front/backends” … I think, right?

jedesah:
@sellout What do you mean "things that depend on 3rd-party front/backends". We plan on building things that depend on 3rd party backends?

sellout:
Definitely – if someone writes a DynamoDB backend or something, we’d use it.

jedesah:
If that happens, and we deem it to be high enough quality. I think that we should try and convince whoever developed it, to continue development inside the quasar repo. I think that would be better for everyone involved. Ya if that wasn't possible for some reason, then we might consider some exotic setup if we want  to actually develop and publish a version of quasar with that backend included, but I don't think that would be ideal.

wemrysi:
Or we make backends and frontends pluggable at runtime so there is no need to include them in quasar at all.

jedesah:
@wemrysi I dig that. Perhaps with a plugin manager similar to Atom which uses GitHub stars in lieu of a standalone review system.😄

drostron:
@sellout and others, does this look reasonable for the initial phase of publishing?

sellout:
@wemrysi I shy away from imposing dependency hell on our users, but I’m ok with it if at least our distribution (and maybe the quasar repo itself) is pre-configured with a known-compatible backend list.

sellout:
@drostron 👍🏾, but I guess wait for some other responses, since this has been such an active thread 😄

wemrysi:
@sellout Agreed, if we ever did plugins we'd try and make them nice to use. Including a set of high-quality backends/frontends with the standard distribution is a good idea.

wemrysi:
@drostron LGTM

mossprescott:
LGTM

PR #1192, opened 2016-05-21
Add IfUndefined (??) and Range (..) binops to SQL².

sellout:
- More squeezing `Set` out of `Data`. It’s now an error if Data.Set gets
to Mongo.
- remove duplicated planning in `compileToLP`
- Restore short-circuiting of constant plans.
- Decouple `Sql` from `QueryFile` algebra
- unify handling of offset/limit, and make it frontend-agnostic
- move planner tests on constants to compiler tests

### DIFF
--- a/core/src/main/scala/quasar/package.scala
+++ b/core/src/main/scala/quasar/package.scala
@@ -17,9 +17,13 @@
 import quasar.Predef.{List, Long, String, Vector}
 import quasar.effect.Failure
 import quasar.fp._
+import quasar.fp.numeric._
 import quasar.sql._
+import quasar.std.StdLib.set._
 
-import matryoshka._
+import scala.Option
+
+import matryoshka._, Recursive.ops._
 import scalaz._
 import scalaz.Leibniz._
 import scalaz.std.vector._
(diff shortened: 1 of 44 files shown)


core/src/main/scala/quasar/std/set.scala
```diff
@@ -55,18 +49,30 @@ trait SetLib extends Library {
...
+    },
     untyper(t => success(t :: Type.Int :: Nil)))
 
+  val Range = Func(Expansion, "(..)", "Creates a set of values in the range from `a` to `b`, inclusive.",
+    Type.Int, Type.Int :: Type.Int :: Nil,
+    noSimplification,
+    partialTyper {
+      case Type.Const(Data.Int(a)) :: Type.Const(Data.Int(b)) :: Nil =>
+        Type.Const(
+          if (a ≟ b) Data.Int(a)
```
wemrysi: Because any single value can also be viewed as a singleton set?

sellout: Yep.

core/src/main/scala/quasar/types.scala
```diff
@@ -257,14 +258,16 @@ trait TypeInstances {
 }
 
 object Type extends TypeInstances {
-  private def fail[A](expected: Type, actual: Type, message: Option[String]): ValidationNel[TypeError, A] =
+  type ValidType[A] = ValidationNel[SemanticError, A]
```
wemrysi: What about something like `SemanticResult[A]` as `ValidType[Type]` is a bit confusing?

core/src/main/scala/quasar/package.scala
```diff
@@ -44,25 +48,59 @@ package object quasar {
...
+  }
+
+  /** Returns the `LogicalPlan` for the given SQL^2 query, or a list of
+    * results, if the query was foldable to a constant. This also takes a
+    * function to apply any extra-query operations to the LP prior to
+    * optimization.
+    */
+  def queryPlan(
+    query: Fix[Sql], vars: Variables)(
+    preprocess: Fix[LogicalPlan] => Fix[LogicalPlan])(
```
wemrysi: While I see the purpose of this function, it bothers me a bit that the usual argument to it is `identity`. I wonder if creating something like `finalizePlan(lp: Fix[LogicalPlan]): CompileM[List[Data] \/ Fix[LogicalPlan]]` that does optimization, typechecking and the constant check would be useful? This then becomes `precompile(query, vars) >>= finalizePlan` and the places that need to get in-between `precompile` and `finalizePlan` can do so explicitly. We could even create a version of `queryPlan` that accepts the offset/limit (since that is the only other function ever passed as `preprocess`).

sellout: Yeah, I think exposing the offset and limit via `queryPlan` sounds fine. I didn’t want to expose the “seam” between `precompile` and `finalizePlan` because a big part of this change was to have the types ensure that we don’t bypass short-circuiting (although, going forward, bypassing short-circuiting should break _lots_ of stuff, not just the types here).

sellout:
@wemrysi Updated. Actually in review now ;)

core/src/main/scala/quasar/package.scala
wemrysi: Ok, that works for me. Maybe in a future version we can have a newtyped `LogicalPlan` that is only obtainable from `finalizePlan` and then `QueryFile` only accepts this newtyped thing. I also started pondering if we could leverage subtyping to represent a `LogicalPlan` that doesn't simplify to a constant, but even non-constant plans can still have constant terms, so that may be a dead-end.

sellout: Yeah – we’ve managed to improve that in QScript – there is no `Literal` node. Literals can only exist as part of an expression in a `Map` node. It’s still _possible_ to represent a constant plan, but it’s at least much more awkward. And it’s done in a way that even if it _does_ get to a backend (which we promise it won’t), there’s no special handling required (unlike the current situation with `$pure` in the Mongo backend).

wemrysi:
@sellout So I know I mentioned "in a future" commit in my other comment, but what _do_ you think about a newtyped `LogicalPlan` that is produced by `finalizePlan` (i.e. there is no way to create one other than running that function). Would this allow us to break things apart a bit?

My main gripe with the current impl. is the loss of compositionality, i.e. that we have to force either some function or a particular behavior (i.e. `addOffsetLimit`) vs just composing the transformations naturally.

sellout:
Well, in a few weeks, `finalizePlan` will be returning QScript rather than LP. So, that’s one better than newtype, and not that far away.

wemrysi:
@sellout Ok, that answers a question I forgot to ask, in that case I'm ok with the awkwardness for the time being given there is an end in sight :smile:

it/src/test/scala/quasar/StreamingQueryRegressionSpec.scala
```diff
@@ -27,5 +28,5 @@ class StreamingQueryRegressionSpec
   val suiteName = "Streaming Queries"
 
   def queryResults(expr: Fix[Sql], vars: Variables) =
-    query.evaluateQuery(expr, vars)
+    fsQ.evaluateQuery(expr, vars)(ι)
```
wemrysi: I think most of the call-sites to `evaluateQuery` missed the signature change.

sellout: Yeah, I caught that locally. And I just accidentally pushed it, squashed into the previous commit. Sorry.

core/src/main/scala/quasar/package.scala
```diff
@@ -76,23 +76,25 @@ package object quasar {
...
 
-  def addOffsetLimit[T[_[_]]: Corecursive](off: Natural, lim: Option[Positive]):
-      T[LogicalPlan] => T[LogicalPlan] = lp => {
-    val skipped = if (off.get != 0L) Drop(lp, LogicalPlan.ConstantF[T[LogicalPlan]](Data.Int(off.get)).embed).embed else lp
-    val limited = lim.fold(skipped)(l => Take(skipped, LogicalPlan.ConstantF[T[LogicalPlan]](Data.Int(l.get)).embed).embed)
-    limited
+  def addOffsetLimit[T[_[_]]: Corecursive](
+    lp: T[LogicalPlan], off: Natural, lim: Option[Positive]):
+      T[LogicalPlan] = {
+    val skipped =
```
wemrysi: Nice, I forgot to mention the premature optimization here, but you found it anyhow.

wemrysi:
Ok, I think this looks good modulo rebasing/travis. :+1:

PR #1193, opened 2016-05-23
A little cleanup

mossprescott:
- delete `parseInt` and `parseDouble`; they're provided by scalaz
- recast `parseBigInt` and `parseBigDecimal` as ops on String, to parallel scalaz

### DIFF
--- a/core/src/main/scala/quasar/fp/package.scala
+++ b/core/src/main/scala/quasar/fp/package.scala
@@ -19,6 +19,8 @@ package quasar
 import quasar.Predef._
 import quasar.RenderTree.ops._
 
+import java.lang.NumberFormatException
+
 import monocle.{Lens, Prism}
 import scalaz.{Lens => _, _}, Liskov._, Scalaz._
 import scalaz.stream._
@@ -339,6 +341,19 @@ trait SKI {
 }
 object SKI extends SKI
 
+trait StringOps {
+  final implicit class StringOps(val s: String) {
+    // NB: see scalaz's `parseInt`, et al.
+    // These will appear in scalaz 7.3.
+
+    def parseBigInt: Validation[NumberFormatException, BigInt] =
+      Validation.fromTryCatchThrowable[BigInt, NumberFormatException](BigInt(s))
+
+    def parseBigDecimal: Validation[NumberFormatException, BigDecimal] =
+      Validation.fromTryCatchThrowable[BigDecimal, NumberFormatException](BigDecimal(s))
+  }
+}
+
 package object fp
     extends TreeInstances
     with ListMapInstances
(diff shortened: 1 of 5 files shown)


core/src/main/scala/quasar/fp/package.scala
```diff
@@ -339,6 +341,26 @@ trait SKI {
 }
 object SKI extends SKI
 
+trait StringOps {
+  final implicit class StringOps(val s: String) {
+    // NB: see scalaz's `parseInt`, et al.
+
+    def parseBigInt: Validation[NumberFormatException, BigInt] =
+      try {
```
jedesah: How about `\/.fromTryCatchNonFatal(BigInt(s)).validation`?

main/src/main/scala/quasar/main/prettify.scala
```diff
@@ -151,11 +151,9 @@ object Prettify {
     else if (str == "true") Some(Data.Bool(true))
     else if (str == "false") Some(Data.Bool(false))
     else
-      parseBigInt(str).fold(
-        parseBigDecimal(str).fold(
-          DataCodec.Readable.decode(Json.jString(str)).toOption)(
-          x => Some(Data.Dec(x))))(
-        n => Some(Data.Int(n)))
+      str.parseBigInt.toOption.map(Data.Int(_)) orElse
```
jedesah: 👍 Much clearer this way!

core/src/test/scala/test/Trampoline.scala.sav
```diff
@@ -0,0 +1,60 @@
+/*
```
jedesah: What does `.sav` mean?

mossprescott: It means I checked that file in accidentally. Oops.

core/src/main/scala/quasar/fp/package.scala
mossprescott: That would produce `Validation[Throwable, BigInt]`, which isn't wrong, I guess. I was just following the pattern established by scalaz, and it does seem that `NumberFormatException` is the only exception that can be thrown there.

jedesah: `\/.fromTryCatchThrowable[BigInt, NumberFormatException](BigInt(s)).validation` 😄 ?

mossprescott: Yeah, that's nicer and should be equivalent. I don't know, make it nicer or mirror the implementation in scalaz (https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/std/String.scala#L47-L87)?

Maybe make it nicer and submit a PR that adds our two functions _and_ uses the nicer syntax? I'll try that and see if I get any interesting feedback over there.

jedesah: That sounds like a decent plan. In the meantime you bring up a good point about consistency though so I could go either way on this PR.

mossprescott:
Got positive feedback from scalaz folks on adding these and switching to `fromTryCatchThrowable`, so updated this PR to match.

jedesah:
LGTM.

PR #1194, opened 2016-05-24
Aggregate all subprojects in `root`.

sellout:
This is stolen from @drostron’s #1191, in the hopes that it fixes coverage.

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+- restore coverage and full testing to the build
--- a/build.sbt
+++ b/build.sbt
@@ -145,7 +145,7 @@ lazy val oneJarSettings =
 
 lazy val root = project.in(file("."))
   .settings(commonSettings: _*)
-  .aggregate(repl, web, it)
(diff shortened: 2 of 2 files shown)


mossprescott:
After a bunch of tedious local testing, I'm building a branch with the same change at this moment. Arg. Obviously time to quit for the day.

sellout:
Well, I think we didn’t get a codecov comment because no code changed, but [our coverage is up to 78%](https://codecov.io/gh/quasar-analytics/quasar/commit/1caf0e339f61f1aabada83f7c7c011d036d3c9b3), lower than it should be, but better than 6%.

mossprescott:
Yes, and actually running the tests in `core` and `mongodb` is a plus, too 😄 

LGTM

sellout:
Holy crap – it wasn’t running the tests!

I thought `.dependsOn(core % "test->test;compile->compile")` caused the tests to be run when you run the tests on the dependent project. And since we aggregated the projects at the bottom of the tree, that should have cascaded up.

I _really_ don’t get SBT.

PR #1198, opened 2016-05-29
Add coursier plugin to project

rabbitonweb:
Coursier works as a plugin for a SBT project. Coursier is a super fast
dependency resolver that allows fetching the dependencies
in parallel.

More about the plugin here: https://github.com/alexarchambault/coursier

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,6 +8,7 @@ addSbtPlugin("com.jsuereth"          % "sbt-pgp"         % "1.0.0")
 addSbtPlugin("org.brianmckenna"      % "sbt-wartremover" % "0.14")
 addSbtPlugin("de.heikoseeberger"     % "sbt-header"      % "1.5.0")
 addSbtPlugin("com.eed3si9n"          % "sbt-buildinfo"   % "0.6.1")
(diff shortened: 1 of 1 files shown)


rabbitonweb:
It fails on a single build MONOGO_RELEASE_3_0_RO - I truly don't know why.

jedesah:
So is the build passing now?

jedesah:
I don't have any objections to this if it works on Travis. The question arises as to whether this should be installed by each developer since in theory it doesn't really affect the semantics of the build. However, it probably makes sense to add it to the project if we want to commit to making sure the build works with this plugin even if bugs/issues pop up in the future.

Also, if this makes our builds faster (especially on Travis) then 💃  💃  💃 . And that would be a pretty strong argument for adding it the project's build.

jedesah:
The travis build times though are pointing towards this not being faster. It was actually a little slower than the previous PR. I wonder why that is?

jedesah:
As @sellout pointed out at lunch, slower build times are probably due to Travis instances only having 1.5 cores. My vote would be to wait until we have a beefier build machine before making use of this plugin.

jedesah:
Unless there are any objections, I am going to go ahead and close this PR as it's a really neat idea but it's not really practical for us right now.

PR #1200, opened 2016-06-02
Add `toCatchable` interpreter to `quasar.effect.Failure`

jedesah:


wemrysi:
Good idea. How about deleting `toTaskFailure` and reimplementing `toRuntimeException` using `toCatchable` as it is more general?

Also, why not `bugfix` as this doesn't modify the public (HTTP) api?

jedesah:
I figured maybe `toTaskFailure` was there for inference purposes.

I didn't realize we had changed our naming scheme for version bumps.

jedesah:
Why did we change the naming scheme? It's not a bug fix, I prefered `revision`.

wemrysi:
I don't think `toTaskFailure` exists for inference purposes, I just didn't think to implement `toCatchable`. 

As for the naming scheme, see https://github.com/slamdata/devtools/pull/9.

jedesah:
@wemrysi Implemented suggestions.

wemrysi:
:+1:

PR #1201, opened 2016-06-03
Remove mountingLens from ConfigOps

drostron:
- Motivated by downstream needs for SD-1667
- Also brings `ConfigOps` closer to its typeclass disposition


mossprescott:
:+1:

PR #1202, opened 2016-06-08
Add interpret8

drostron:
Required downstream for SD-1659


jedesah:
LGTM, assuming build passes.

PR #1204, opened 2016-06-09
Remove interpretN in favor of :+:

drostron:
Per conversation with @wemrysi.

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+Remove interpretN in favor of :+:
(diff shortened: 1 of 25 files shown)


jedesah:
This is pretty cool. Especially getting rid of the duplication of `interpretN`. I wonder if we might still want to annotate the types in certain situations to help readability, but it's certainly nice to have the option on whether or not to use type inference.

I guess, time will tell is this makes it easier to compose interpreters, but I am optimistic.

wemrysi:
:+1:

PR #1205, opened 2016-06-10
Upgrade Matryoshka and Monocle.

sellout:
- remove `pPrism` (now `Prism.partial`) and `PrismOps.apply` (now
  `Prism.apply`)
- remove our local version of `trait ∘[F[_], G[_]]`, now available via
  Matryoshka
- replace usage of `FunctorT.translate` with `FunctorT.transAna`
- convert some `apply`/`unapply` objects to Prisms, since those
  operations (and many others) are exposed via the Prism.


drostron:
👍🏽

project/Dependencies.scala
```diff
@@ -7,7 +7,7 @@ import sbt._, Keys._
 object Dependencies {
   private val scalazVersion  = "7.2.2"
   private val slcVersion     = "0.4"
-  private val monocleVersion = "1.2.1"
+  private val monocleVersion = "1.2.2"
```
jedesah: Btw. `1.3.0-SNAPSHOT` is what is actually being brought into `quasar`, this doesn't really change anything (except for being semantically more correct).

`argonaut` `6.2-M2` was just [released](https://github.com/argonaut-io/argonaut/issues/235). Might make sense to upgrade to that in this PR as well so that we actually bring in `monocle` `1.2.2` from now on.

If not, I will open a PR after this one to upgrade `argonaut` as well.

CHANGELOG/bugfix.md
```diff
@@ -0,0 +1 @@
+- upgrade dependencies
```
jedesah: I think it would be nice to be a little more precise.

core/src/main/scala/quasar/logicalplan.scala
```diff
@@ -155,7 +157,7 @@ object LogicalPlan {
   // TODO we create a custom `unapply` to bypass a scalac pattern matching bug
   // https://issues.scala-lang.org/browse/SI-5900
   object InvokeFUnapply {
-    def unapply[A, N <: Nat](in: InvokeF[A, N]): Option[(GenericFunc[N], Func.Input[A, N])] = Some((in.func, in.values))
+    def unapply[A, N <: Nat](in: InvokeF[A, N]): Some[(GenericFunc[N], Func.Input[A, N])] = Some((in.func, in.values))
```
jedesah: Hmm. Curious why this is useful. I mean I am all for more precise types, but this seems to go against our general pattern of hiding more precise types.

project/Dependencies.scala
sellout: Ah yeah, I’ll bump the argonaut number here, too.

core/src/main/scala/quasar/logicalplan.scala
sellout: It does go against our general pattern, but it preserves exhaustiveness checking, whereas using `Option` does not.

sellout:
Updated this. I’ve checked the rebase locally, but didn’t want to disrupt the PR.

sellout:
@jedesah You may want to look, since @drostron already signed off.

jedesah:
👍

PR #1206, opened 2016-06-11
Cleanup resulting from new Free implementation

wemrysi:
Cleans up the following, now possible due to the improved implementation of `Free` in scalaz-7.2.x:

* Remove all `Functor` constraints previously imposed by `Free`
* Add `QueryFile#enumerate` as an alternative to `evaluate` to provide a streaming construct which can be "run" without `Catchable`
* Delete the dubious `Catchable[Free[S, ?]]` instance (no longer necessary thanks to `enumerate`)
* Remove all unnecessary usage of Coyoneda
* Move all `Free`-related natural transformations to the `free` package


core/src/main/scala/quasar/fp/free/package.scala
```diff
@@ -34,6 +31,27 @@ package object free {
     }
   }
 
+  def flatMapSNT[S[_], T[_]](f: S ~> Free[T, ?]): Free[S, ?] ~> Free[T, ?] =
+    new (Free[S, ?] ~> Free[T, ?]) {
+      def apply[A](fa: Free[S, A]) = fa.flatMapSuspension(f)
+    }
```
sellout: I figure I’ll mention this, since you’re touching the code here – I’m adding a bunch of higher-order stuff to Matryoshka (which should maybe eventually be pulled into yet another library).

`liftFT` and `flatMapSNT` are `Hmonad` `hpoint` and `hbind`, respectively; `mapSNT` is `HFunctor.hmap`; etc.

sellout:
This looks great – no idea how it could have reduced coverage, though.

PR #1207, opened 2016-06-15
Revert back to hand-nested effect type aliases

wemrysi:
The :+: syntax causes significant performance degredation in implicit
search for Inject instances on large effect types.


mossprescott:
LGTM

PR #1212, opened 2016-06-25
SD-1757 Remove MountConfigs interpretation from CoreEff interpreter

drostron:
- Provide non-IO variants for a few effect types


wemrysi:
:+1:

PR #1214, opened 2016-06-28
Update wart remover

sellout:


sellout:
There’s a new comment on puffnfresh/wartremover#226 that may allow me to enable the ExplicitImplicitTypes wart, but I’m not sure if the cure is worse than the disease in this case. Thoughts?

core/src/main/scala/quasar/sql/parser.scala
```diff
@@ -41,6 +41,11 @@ object ParsingError {
 
 final case class Query(value: String)
 
+sealed trait DerefType[T[_[_]]]
```
drostron: Did this sneak in or was it just something you noticed needing updating while passing though?

drostron:
Great to see WartRemover 1.0.0 in the mix. Other than the one question, 👍🏽.

drostron:
Oh, missed your earlier comment — I'd say let's leave the wart disabled and keep an eye on the issue.

core/src/main/scala/quasar/sql/parser.scala
sellout: Needed to extract it from the class – WartRemover doesn't like path-dependent types now.

drostron: Ah, interesting. OK.

PR #1218, opened 2016-07-02
SD-1774 Warn and skip failed mounts on startup

wemrysi:


main/src/main/scala/quasar/main/package.scala
```diff
@@ -251,18 +251,30 @@ package object main {
...
-
-    type MainF[A] = EitherT[mnt.F, String, A]
+  /** Mount all the mounts defined in the given configuration, returning
+    * the paths that failed to mount along with the reasons why.
+    */
+  def attemptMountAll[S[_]](
+    config: MountingsConfig
+  )(implicit
+    mounting: Mounting.Ops[S]
+  ): Free[S, IList[(APath, String)]] = {
```
sellout: Why a List rather than a Map?

main/src/main/scala/quasar/main/package.scala
```diff
@@ -251,18 +251,30 @@ package object main {
...
+    }
 
-    def toMainF(v: mnt.M[PathTypeMismatch \/ Unit]): MainF[Unit] =
-      EitherT[mnt.F, String, Unit](
-        v.fold(_.shows.left, _.fold(_.shows.left, _.right)))
+    config.toMap.toList foldMapM attemptMount map (_.toIList)
+  }
 
-    mc.toMap.toList.traverse_ { case (p, cfg) => toMainF(mnt.mount(p, cfg)) }.run
+  /** Prints a warning about the mount failure to the console. */
```
sellout: This needs a TODO to serve the warning to clients.

sellout:
A couple minor things.

main/src/main/scala/quasar/main/package.scala
wemrysi: No reason, I can switch to a map.

main/src/main/scala/quasar/main/package.scala
wemrysi: Not sure I understand what you mean.

sellout: I thought this was in an issue somewhere, but I can’t find it. We should message stuff like this to the user. SlamData is often started with a double-click or with an automated process. In those cases, the user still has no indication of why their mount isn’t there.

We could inject a warning message into static content served by Quasar, at least on that initial page that’s opened when the server starts, but potentially more generally than that.

wemrysi: Ah ok, yeah you're looking for SD-1131. This PR is an interim fix for that issue that just deals with the startup part as the metastore has made things worse by requiring interacting with a database to find the offending mounts.

I'll go ahead and add a TODO as a reminder.

wemrysi:
@sellout Updated, let me know if there is anything else.

sellout:
👍🏾

PR #1220, opened 2016-07-04
Add Ensime to .gitignore

rabbitonweb:
Ignore Ensime specific files


wemrysi:
:+1:

As an aside, it is also possible to exclude files like these from all your projects by adding it to the global excludes file configured in your global git config under `core.excludesFile`.

PR #1221, opened 2016-07-05
Add script that runs tests on dockerized mongodb

rabbitonweb:
This script will
1. clean any existing docker container named some-mongo
2. run new instance of mongodb
3. run tests on that instance
4. clear afterwards

### DIFF
--- a/full-it-tests.sh
+++ b/full-it-tests.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+echo "clear some-mongo docker container if exists"
+docker stop some-mongo
+docker rm some-mongo
+
+echo "start clear mongo instance, expose 27017 port"
+docker run --name some-mongo -p 27017:27017 -d mongo
+
+echo "set QUASAR_MONGODB_3_2 variable"
+export QUASAR_MONGODB_3_2="{\"mongodb\":{\"connectionUri\":\"mongodb://localhost:27017\"}}"
+
+echo "run tests"
+./sbt test
+
+echo "clear some-mongo docker container after tests"
+docker stop some-mongo
+docker rm some-mongo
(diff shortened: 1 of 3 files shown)


full-it-tests.sh
```diff
@@ -0,0 +1,17 @@
+!/bin/bash
```
paulp: Your shebang is missing the leading hash.

alissapajer:
@rabbitonweb LGTM

jedesah:
@rabbitonweb You should squash this down to a single commit and run the merge script (in SlamData/devtools, if you haven't already checked out that repo).

alissapajer:
@rabbitonweb @jedesah This may be a discussion we should have with all of engineering, but sometimes it's nice to have multiple commits in the git history. Each PR being squashed to a single commit doesn't always make sense. If a few distinct things were implemented/fixed in a single PR, then having each of those as a single commit is nice. It makes it easy for someone else to look at a specific change without it being entangled with another change in the diff. And if we ever have to revert a commit, having smaller diffs in each commit is nicer for that reason too.

(And the merge script as it stands does support multiple commits merging into master.)

jedesah:
@alissapajer I agree and I think most of the team would agree as well, but having a formal discussion might prove fruitful nonetheless.

A cursory look at these commits led me to believe that there was not much benefit in having them be separate, which is why I suggested squashing them before merging.

full-it-tests.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+echo "clear some-mongo docker container if exists"
+docker stop some-mongo
+docker rm some-mongo
```
paulp: I'm a little late here, but it's bad practice to be deleting peoples' docker containers without knowing where they came from. This depends on someone not picking the name "some-mongo" for their valuable mongo things, and that's not exactly a random ten character string. I suggest at least putting "quasar" into the container name, like "quasar-some-mongo".

PR #1222, opened 2016-07-05
Update README to reflect use of backticks for identifiers

jedesah:


README.md
```diff
@@ -183,7 +183,7 @@ the root, and it contains a database called `test`:
 
 The "tables" in SQL queries refer to collections in the database by name:
 ```
-💪 $ select * from zips where state='CO' limit 3
+💪 $ select * from zips where state=`CO` limit 3
```
mossprescott: This isn't right. `CO` here is meant to be a String literal, so it should be `"CO"` (not single-quotes, as it was before).

README.md
```diff
@@ -213,14 +213,14 @@ Query time: 0.1s
 You may also store the result of a SQL query:
 
 ```sql
-💪 $ out1 := select * from zips where state='CO' limit 3
+💪 $ out1 := select * from zips where state=`CO` limit 3
```
mossprescott: Similar here: `"CO"`

README.md
```diff
@@ -213,14 +213,14 @@ Query time: 0.1s
...
 ```
 
 The location of a collection may be specified as an absolute path by
 surrounding the path with double quotes:
 
 ```sql
-select * from "/test/zips"
+select * from `/test/zips`
```
mossprescott: This looks good.

mossprescott:
I see one related problem where String literals should be using `"` in place of `'`. I believe this will actually work at the moment, but according to the spec, `'` is for single-char literals only and eventually we'll insist on `"` for these. 

Some string constants [here](https://github.com/quasar-analytics/quasar/pull/1222/files#diff-04c6e90faac2675aa89e2176d2eec7d8R171) should use `"`.

The same examples also appear [here](https://github.com/quasar-analytics/quasar/pull/1222/files#diff-04c6e90faac2675aa89e2176d2eec7d8R239).

And again [here](https://github.com/quasar-analytics/quasar/pull/1222/files#diff-04c6e90faac2675aa89e2176d2eec7d8R254).

And maybe one more time.

sellout:
I _think_ the parser already insists that `'` is used only for a single char.

mossprescott:
I suppose it's too much to ask to generate this doc with `tut`?

jedesah:
@mossprescott You mean 2 more times 😄 

Also, what `DATE '2015-07-06'`, should it be `DATE "2015-07-06"`? Is `DATE` a function that accepts a String or is it some kind of literal?

sellout:
@jedesah That should actually look like `DATE("2015-07-06")`.

jedesah:
@sellout Ah, yes, I see. Tried different combinations in the REPL, but did not try that one.

jedesah:
@mossprescott Ya, `tut` or using `specs2` to generate the README would be an improvement I think.

README.md
```diff
@@ -363,15 +363,15 @@ Compiles, but does not execute, a SQL query, contained in the single, required
...
 The resulting plan is returned in the response body.
 
-SQL<sup>2</sup> supports variables inside queries (`SELECT * WHERE pop < :cutoff`). Values for these variables, which can be any expression, should be specified as query parameters in this API using the variable name prefixed by `var.` (e.g. `var.cutoff=1000`). Failure to specify valid values for all variables used inside a query will result in an error. These values use the same syntax as the query itself; notably, strings should be surrounded by single quotes. Some acceptable values are `123`, `'CO'`, and `DATE '2015-07-06'`.
+SQL<sup>2</sup> supports variables inside queries (`SELECT * WHERE pop < :cutoff`). Values for these variables, which can be any expression, should be specified as query parameters in this API using the variable name prefixed by `var.` (e.g. `var.cutoff=1000`). Failure to specify valid values for all variables used inside a query will result in an error. These values use the same syntax as the query itself; notably, strings should be surrounded by single quotes. Some acceptable values are `123`, `"CO"`, and `DATE("2015-07-06")`.
```
sellout: Why is this block repeated? Can’t we just add a new section and a link or reference to it?

jedesah:
This isn't fair. This PR should have more lines removed than lines added. The only reason it doesn't have more red is because the lines removed are huge lines of about 600 characters! 😞

jedesah:
@mossprescott @sellout Let me know if you think this is ready. I know there is a `TODO` in there, but I don't think I should be the one to write a high-level overview of SQL<sup>2<sup>. I mean, maybe I should since I don't know as much about it as perhaps I should, but I would rather do that one day where @sellout is at the office! 😄

sellout:
👍🏾

jedesah:
Just realized, I need to do `<sup>2</sup>` instead of `<sup>2<sup>` because of how my comment initially rendered. So will fix that before rebasing 😄

sellout:
Or just use ² 😉

jedesah:
@sellout Ok, will try that!

jedesah:
@sellout You didn't catch that I used `revision` instead of `bugfix`! 😄

sellout:
Oh, I _did_ notice – was going to ask if you made the change to `merge` already. But forgot!

jedesah:
I wish I had...

README.md
```diff
@@ -8,6 +8,16 @@
...
 
+## SQL²
+
+SQL² is the dialect of SQL that Quasar understands.
+
+SQL² is a superset of standard SQL. Therefore, in the following documentation SQL² will be used interchangeably with SQL.
+
+TODO: Provide more details on SQL²
+
+SQL² supports variables inside queries (`SELECT * WHERE pop < :cutoff`). Values for these variables, which can be any expression, should be specified as additional parameters in the connectionUri using the variable name prefixed by `var.` (e.g. `var.cutoff=1000`). Failure to specify valid values for all variables used inside a query will result in an error when the mount is created or used. These values use the same syntax as the query itself; notably, strings should be surrounded by double quotes. Some acceptable values are `123`, `"CO"`, and `DATE("2015-07-06")`.
```
mossprescott: I guess the language wasn't simply copied, because this text seems to be specific to views ("will result in an error _when the mount is created or used_"). Probably just make that generic—i.e. remove the last bit.

README.md
```diff
@@ -8,6 +8,16 @@
 
 Quasar is an open source NoSQL analytics engine that can be used as a library or through a REST API to power advanced analytics across a growing range of data sources and databases, including MongoDB.
 
+## SQL²
+
+SQL² is the dialect of SQL that Quasar understands.
+
+SQL² is a superset of standard SQL. Therefore, in the following documentation SQL² will be used interchangeably with SQL.
+
+TODO: Provide more details on SQL²
```
mossprescott: How about just linking to http://quasar-analytics.org/docs/sqltutorial/?

mossprescott:
One minor edit and a suggestion for the TODO.

But seriously, :+1:.

PR #1224, opened 2016-07-06
Add `inputs` to `/compile` endpoint

jedesah:


sellout:
I think this looks good, but I have two questions:

1. are there any other responses we should be wrapping in a JSON doc (even if there’s only one key) to make it easier to have non-breaking extensions to them in the future and
2. what comes back if (i.e., can we have a test for) someone compiles a query that is statically evaluated (e.g. `4 + 3`) – there are no inputs, and I’m at least curious if the physicalPlan is also missing, or if the plan is the results themselves, etc.

jedesah:
@sellout Ya the `/compile` route is under-specified and I am happy to correct that but I did not want to get overly distracted since @cryogenian is waiting on this change. I am happy to add the tests. @cryogenian How urgent is this change (keeping in mind the time required to trickle this down to `quasar-advanced`? I can either add the additional tests to this PR or open a new one.

Btw @sellout who's fault is it that this route is underspecified huh!? 😄  Git blame tells me someone _cough, cough_ added some functionality to this route and did not add any tests...

You are lucky I wasn't the reviewer! 😛

cryogenian:
The faster is the better of course :)

jedesah:
Ok, then I would suggest merging this PR and opening a new PR to add the missing tests for this route.

sellout:
I grudgingly acquiesce to that. I’d like to see a PR with the new test ASAP, though.

PR #1225, opened 2016-07-06
Polish `/compile` API

jedesah:


sellout:
Thanks!

web/src/main/scala/quasar/api/services/query/compile.scala
```diff
@@ -60,7 +61,9 @@ object compile {
     ): Free[S, QResponse[S]] =
       respond(queryPlan(expr, vars, offset, limit)
         .run.value.traverse[Free[S, ?], SemanticErrors, QResponse[S]](_.fold(
-          dataResponse(_).toResponse[S].point[Free[S, ?]],
+          κ(Json(
+            "physicalPlan" := "none",
```
sellout: What about using `null`? Isn’t a physical plan generally represented by a string here, or is there more structure to it?

jedesah: Sure, I forget `null` is a valid json value.

jedesah:
@sellout Fixed `null`.

README.md
```diff
@@ -363,16 +363,16 @@ error at the root of the response):
...
-Compiles, but does not execute, a SQL² query, contained in the single, required
-query parameter, on the backend responsible for the request path. The resulting
-plan is returned in the response body.
+Compiles (but does not execute) a SQL² query, contained in the single, required query parameter.
+Returns a Json object with the following shape:
 
+```json
+{
+  "inputs": [<filePath>, ...],
+  "logicalPlan": "Description of the logical plan that would be executed by this query"
```
sellout: This should say `"physicalPlan"`, and mention that `null` indicates that no backend is used in the query … or something like that.

Ok, think that’s the last comment.

jedesah: Oh, oops. Nice catch.

sellout:
👍🏾

PR #1226, opened 2016-07-08
Skeleton FS

sellout:
A minimal filesystem implementation that simply fails any request.

Also eliminated references to Mongo-specific stuff outside of the mongo (and IT) subprojects.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -152,30 +152,27 @@ lazy val oneJarSettings =
         commitReleaseVersion,
         pushChanges))
 
-//      foundation
-//     / / | | \ \
-//
-//        ejson
-//          |
-//        core
-//        /   \
-// mongodb    «other backends»
-//        \   /
-//        main
-//        /  \
-//    repl    web
-//        \  /
-//         it
-//          |
-//        root
-
-// common components
-
 lazy val root = project.in(file("."))
   .settings(commonSettings: _*)
   .settings(noPublishSettings)
-  .aggregate(foundation, ejson, core, mongodb, main, repl, web, it)
+  .aggregate(
+        foundation,
+//     / / | | \ \
+//
+          ejson,
+//          |
+          core,
+//        /   \
+   mongodb,    skeleton,
+//        \   /
+          main,
+//        /  \
+      repl,   web,
+//        \  /
+           it)
   .enablePlugins(AutomateHeaderPlugin)
+
+// common components
 
 lazy val foundation = project
   .settings(name := "quasar-foundation-internal")
@@ -208,7 +205,9 @@ lazy val core = project
 
 lazy val main = project
   .settings(name := "quasar-main-internal")
-  .dependsOn(mongodb % "test->test;compile->compile")
+  .dependsOn(
+    mongodb  % "test->test;compile->compile",
+    skeleton % "test->test;compile->compile")
   .settings(oneJarSettings: _*)
   .settings(publishSettings: _*)
   .enablePlugins(AutomateHeaderPlugin)
@@ -220,6 +219,15 @@ lazy val mongodb = project
   .dependsOn(core % "test->test;compile->compile")
   .settings(oneJarSettings: _*)
   .settings(publishSettings: _*)
+  .settings(libraryDependencies +=
+    "org.mongodb" % "mongodb-driver-async" % "3.2.2")
+  .enablePlugins(AutomateHeaderPlugin)
+
+lazy val skeleton = project
+  .settings(name := "quasar-skeleton-internal")
+  .dependsOn(core % "test->test;compile->compile")
+  .settings(oneJarSettings: _*)
+  // .settings(publishSettings: _*) // NB: uncomment this line when you copy it
   .enablePlugins(AutomateHeaderPlugin)
 
 // frontends
--- a/core/src/main/scala/quasar/fs/package.scala
+++ b/core/src/main/scala/quasar/fs/package.scala
@@ -25,7 +25,7 @@ import argonaut._
 import pathy.Path, Path._
 import scalaz.{Failure => _, _}, Scalaz._
 
-package object fs {
+package object fs extends PhysicalErrorPrisms {
   type FileSystem0[A] = Coproduct[WriteFile, ManageFile, A]
   type FileSystem1[A] = Coproduct[ReadFile, FileSystem0, A]
   type FileSystem[A]  = Coproduct[QueryFile, FileSystem1, A]
@@ -59,6 +59,8 @@ package object fs {
 
   type FileSystemFailure[A] = Failure[FileSystemError, A]
   type FileSystemErrT[F[_], A] = EitherT[F, FileSystemError, A]
+
+  type PhysErr[A] = Failure[PhysicalError, A]
 
   def interpretFileSystem[M[_]](
     q: QueryFile ~> M,
--- a/main/src/main/scala/quasar/main/package.scala
+++ b/main/src/main/scala/quasar/main/package.scala
@@ -17,16 +17,13 @@
 package quasar
 
 import quasar.Predef._
-
 import quasar.effect._
 import quasar.fp._
 import quasar.fp.free._
 import quasar.fs._
 import quasar.fs.mount._
 import quasar.fs.mount.hierarchical._
-import quasar.physical.mongodb._
 
-import com.mongodb.MongoException
 import monocle.Lens
 import pathy.Path.posixCodec
 import scalaz.{Failure => _, Lens => _, _}, Scalaz._
@@ -46,13 +43,13 @@ package object main {
 
   /** Effects that physical filesystems require.
     */
-  type PhysFsEff[A]  = Coproduct[MongoErr, Task, A]
+  type PhysFsEff[A]  = Coproduct[PhysErr, Task, A]
   type PhysFsEffM[A] = Free[PhysFsEff, A]
 
   object PhysFsEff {
     // Lift into FsEvalIOM
     val toFsEvalIOM: PhysFsEff ~> FsEvalIOM =
-      injectFT[MongoErr, FsEvalIO] :+:
+      injectFT[PhysErr, FsEvalIO] :+:
       injectFT[Task, FsEvalIO]
   }
 
@@ -62,6 +59,7 @@ package object main {
     * to a fixed set of effects that filesystems must interpret into.
     */
   val physicalFileSystems: FileSystemDef[PhysFsEffM] =
+    quasar.physical.skeleton.fs.definition[PhysFsEff] |+|
     quasar.physical.mongodb.fs.mongoDbFileSystemDef[PhysFsEff]
 
   /** The intermediate effect FileSystem operations are interpreted into.
@@ -112,7 +110,7 @@ package object main {
     * configuration based on the final context of interpretation
     * (i.e. web service vs cmd line).
     */
-  type FsEval[A]    = Coproduct[MongoErr, MountConfigs, A]
+  type FsEval[A]    = Coproduct[PhysErr, MountConfigs, A]
   type FsEvalIO[A]  = Coproduct[Task, FsEval, A]
   type FsEvalIOM[A] = Free[FsEvalIO, A]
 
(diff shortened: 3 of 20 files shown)


build.sbt
```diff
@@ -224,6 +223,13 @@ lazy val mongodb = project
     "org.mongodb" % "mongodb-driver-async" % "3.2.2")
   .enablePlugins(AutomateHeaderPlugin)
 
+lazy val skeleton = project
```
mossprescott: How about a comment here on what this project is for?

skeleton/src/main/scala/quasar/physical/skeleton/fs/package.scala
```diff
@@ -0,0 +1,41 @@
...
+package quasar.physical
+package skeleton
+
+import quasar.fs._
+import quasar.fs.mount.FileSystemDef, FileSystemDef.DefErrT
+
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+
+package object fs {
```
mossprescott: Definitely could use a couple of comments here on what this is meant to be an example of.

skeleton/src/main/scala/quasar/physical/skeleton/fs/package.scala
```diff
@@ -0,0 +1,41 @@
...
+    FileSystemDef.fromPF {
+      case (FsType, uri) =>
+        FileSystemDef.DefinitionResult[Free[S, ?]](
+          interpretFileSystem(
+            Empty.queryFile[Free[S, ?]],
+            Empty.readFile[Free[S, ?]],
+            Empty.writeFile[Free[S, ?]],
+            Empty.manageFile[Free[S, ?]]),
+          Free.point(())).point[DefErrT[Free[S, ?], ?]]
+    }
```
mossprescott: How about defining each of these as, e.g. `def query[S[_]]: QueryFile ~> Free[S, ?] = Empty.queryFile[Free[S, ?]]`? I would even suggest `???` as the implementation but I guess we've got a policy against that.

mossprescott:
I'm a little dubious about the value of an example filesystem that we don't even deploy or test.

I'd rather see a trivial filesystem that actually does something useful, such as reading JSON from a directory on disk, even if it had no query ability at the moment. But I realize even that would be more work than this, so I'm not saying it has to happen in this PR.

main/src/main/scala/quasar/main/package.scala
```diff
@@ -60,6 +59,7 @@ package object main {
     * to a fixed set of effects that filesystems must interpret into.
     */
   val physicalFileSystems: FileSystemDef[PhysFsEffM] =
+    // quasar.physical.skeleton.fs.definition[PhysFsEff] |+|
```
mossprescott: What would be the harm in un-commenting this line? I suppose that would require publishing the sub-project, right? I could see adding a `skeleton` mount being handy for troubleshooting installs, as long as the errors it returns look different than having no mount at all.

sellout: Yeah, I’m cool with that.

sellout:
This isn’t meant to be an example (so, yes, code-level docs would help explain that). It’s a skeleton to be copied to a new backend and incrementally improved. There are docs at https://github.com/quasar-analytics/quasar/wiki/Adding-a-Filesystem

skeleton/src/main/scala/quasar/physical/skeleton/fs/package.scala
sellout: Yeah, the former is a good idea – anything to get it closer to a fill-in-the-blanks structure. I’m not sure about `???`, pointing to `Empty` gives more idea of the structure to implement.

jedesah:
I feel like if this backend does not have any intrinsic use. We should probably put in some sort of `example` project.

sellout:
@jedesah But it’s not an example, it’s a skeleton, and it _is_ in a skeleton project.

jedesah:
I guess I just don't see the distinction between a "skeleton" and an example.

If it's genuinely useful for debugging mounts, then okay, I guess it can be added as an "official" backend. But IMHO, we should give it a better name in that case. If it's main is use is to demonstrate how to put together a minimal backend then I guess my initial statement still holds (as in I am happy to be convinced of the distinction).

sellout:
So, “skeleton” is a fairly common term for something that’s intended to be copied and built upon. It should work as much as is possible without having extra noise or anything does the wrong thing (e.g., `???` is better than an implementation that reads files from local disk). We should definitely do more here – have a skeleton of the QScript planner type class, etc. (I expect a QScript skeleton just punts everything it can to the fallback interpreter).

Whereas an example is a fully-implemented thing that does something different but comparable. It’s possible we might have an explicit example at some point, but I’m _hoping_ that implementations like JSON files, etc. are clean enough to serve as examples.

skeleton/src/main/scala/quasar/physical/skeleton/fs/package.scala
wemrysi: I also like `Empty` in that as you start running tests you should get failures instead of errors. With `???` you might not even be able to run some of the tests as the can't even construct the interpreters. If we don't want to use `Empty`, then maybe something like

```scala
new (ReadFile ~> Free[S, ?]) {
  def apply[A](rf; ReadFile[A]) = ???
}
```

as then at least you'll only get the error when you try and use one of the unimplemented interpreters.

Anyhow, I'm fine with `Empty`.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -49,6 +49,20 @@ package object fs {
   type FileSystemFailure[A] = Failure[FileSystemError, A]
   type FileSystemErrT[F[_], A] = EitherT[F, FileSystemError, A]
 
+  sealed trait PhysicalError extends Throwable {
```
wemrysi: Mind moving this to its own file, adding making a `sealed abstract class` and adding `Prism`s? Maybe a `Lens` to the `cause`?

What is the reason for this to extend Throwable?

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -49,6 +49,20 @@ package object fs {
...
+    val err: Throwable
+    override def getMessage = err.getMessage
+  }
+  /** While this has the same parameter type as `UnknownError` the distinction
+    * is that this is an exception that came from the filesystem itself, rather
+    * than from the JVM, etc. (e.g., in MongoDB, this would hold
+    * `MongoException`s, whereas NPE would end up in `UnknownError`).
+    */
+  final case class UnhandledFSError(err: Throwable) extends PhysicalError
+  final case class UnknownError(err: Throwable) extends PhysicalError
```
wemrysi: I'm not sure this is necessary. The filesystem should only be catching it's own exceptions and whatever actually runs the resulting `Task` (or whatever) should be responsible for handling any uncaught exceptions at that point.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -49,6 +49,20 @@ package object fs {
...
+  sealed trait PhysicalError extends Throwable {
+    val err: Throwable
+    override def getMessage = err.getMessage
+  }
+  /** While this has the same parameter type as `UnknownError` the distinction
+    * is that this is an exception that came from the filesystem itself, rather
+    * than from the JVM, etc. (e.g., in MongoDB, this would hold
+    * `MongoException`s, whereas NPE would end up in `UnknownError`).
+    */
+  final case class UnhandledFSError(err: Throwable) extends PhysicalError
```
wemrysi: How about `cause: Exception` rather than `Throwable`? A bit more precise as we don't want any `Error`s in here.

Also, do we want any identifiying information like `FSType` or a high-level message field? I'm guessing if filesystems create their own (or use an existing one like `MongoException`) exception hierarchies, then this might not be necessary.

web/src/main/scala/quasar/api/ToApiError.scala
```diff
@@ -141,10 +140,10 @@ sealed abstract class ToApiErrorInstances {
     }
   }
 
-  implicit def mongoExceptionToApiError: ToApiError[MongoException] =
-    error(merr => fromMsg_(
-      InternalServerError withReason "MongoDB error.",
-      merr.getMessage))
+  implicit def mongoExceptionToApiError: ToApiError[PhysicalError] =
+    error(err => fromMsg_(
+      InternalServerError withReason "physical filesystem error.",
```
wemrysi: Capitalize 'Physical'.

core/src/main/scala/quasar/fs/package.scala
wemrysi: Another thought might be allowing for the distinction between recoverable and unrecoverable errors, mongo's db + collection name length limit vs a network outage, for example.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIO.scala
```diff
@@ -46,10 +47,10 @@ final class MongoDbIO[A] private (protected val r: ReaderT[Task, MongoClient, A]
     new MongoDbIO(r mapK (_.attempt))
 
   def attemptMongo: MongoErrT[MongoDbIO, A] =
-    EitherT(attempt flatMap {
```
jedesah: What's the advantage of `∘` over `flatMap` here?

sellout: Well, nothing in the new version returns a value in `MongoDbIO`, so mapping is better than flatmapping with point.

But this may change back, as @wemrysi suggested not handling errors outside the filesystem this way.

jedesah: Ah, I didn't look too closely and I thought `∘` was a synonym for `flatMap`. I guess my point still stands, `map` is a rather short, well recognized identifier, what's the advantage of using `∘` instead of `map`?

Perhaps if I understood the explanation for it. That symbol is usually used for function composition right, is that why it's appropriate for mapping over values with a function? I don't entirely see the link, but I am sure there is one.

sellout: Right, it kind of goes the opposite direction – `∘` is just the operator for `map`, and composition is just `Functor[Function1[A, ?]].map`.

For the specific case of function composition, I tend to use `⋘` (from `Category[Function1]`) instead since it makes the direction explicit and (at least in Scala) you often have to reverse it and use `⋙` to get things to infer (so making the direction explicit is even more helpful).

rabbitonweb:
Will skeleton be in master any day soon? I sort of... rebased to that branch with my current PR and just wonder now how much big mistake that was :)

sellout:
@rabbitonweb I’m about to push an update to this PR, then we’ll see what @wemrysi thinks.

This got left behind when the QScript crit bug popped up, but figured it’s important to get it in.

core/src/main/scala/quasar/fs/package.scala
sellout: Yeah, I think we need a richer set of options here – but I’m not going to get it in this PR. This was really just to sever the tie to Mongo.

core/src/main/scala/quasar/fs/package.scala
sellout: So, it extends Exception (née Throwable), because `CfgsErrs` needed a `toCatchable` on it, and I got tired of digging.

I’ll at least add a comment that we should get rid of that … or maybe you know an easy way to fix the problem?

sellout:
@wemrysi Updated this. Think I addressed everything?

jdegoes:
@sellout Well, merge conflicts...

sellout:
@jdegoes Yes, but I can’t address those until after squashing.

core/src/main/scala/quasar/fs/package.scala
wemrysi: If you add a `Show` instance, you can use `toRuntimeError` instead.

wemrysi:
:shipit:

sellout:
Will merge once CI passes.

rabbitonweb:
https://www.youtube.com/watch?v=6b9ci_z4v7M

PR #1227, opened 2016-07-08
SD-1839 MongoDB: generate temp file names in same directory as hint path

wemrysi:
Previously, we were generating temp files under the first directory of the hint path, i.e.

```
hint = /foo/bar/baz
tmp = /foo/2348ujsdlf.tmp
```

this changes it so we always create the temp file in the directory itself, if the hint path is a directory, or in the hint's parent if it is a file.

This makes temp file location more predictable which makes the permissions required for operations that involve temp files in quasar-advanced more predictable (which is the motivating problem here).

### DIFF
--- a/it/src/test/scala/quasar/fs/ManageFilesSpec.scala
+++ b/it/src/test/scala/quasar/fs/ManageFilesSpec.scala
@@ -21,12 +21,16 @@ import quasar.Predef._
 import quasar.{Data, TestConfig}
 import quasar.fp._
 
+import org.specs2.ScalaCheck
 import pathy.Path._
+import pathy.scalacheck.PathyArbitrary._
 import scalaz._, Scalaz._
 import scalaz.stream._
 
 class ManageFilesSpec extends FileSystemTest[FileSystem](
-    FileSystemTest.allFsUT.map(_.filterNot(fs => TestConfig.isMongoReadOnly(fs.name)))) {
+    FileSystemTest.allFsUT.map(_.filterNot(fs => TestConfig.isMongoReadOnly(fs.name))))
+    with ScalaCheck {
+
   import FileSystemTest._, FileSystemError._, PathError._
   import ManageFile._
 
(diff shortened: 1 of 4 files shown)


mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
```diff
@@ -55,11 +55,19 @@ object managefile {
...
+            .void
+
+        val mkTemp =
+          freshName.liftM[FileSystemErrT] map { n =>
+            refineType(path).fold(
+              _ </> file(n),
+              f => fileParent(f) </> file(n))
+          }
+
+        (checkPath *> mkTemp).run
```
mossprescott: The old implementation would look up the argument's DB path and generate a file at its root, which meant that it would fail if the path didn't include a DB (e.g. `/` or `/foo`). This new version still has that behavior, but it's not quite so obvious why, now that the result is derived more directly from the argument. However, I still think it's the right thing since a temp file that's _not_ in a DB isn't going to be useful.

it/src/test/scala/quasar/fs/ManageFilesSpec.scala
```diff
@@ -232,6 +232,23 @@ class ManageFilesSpec extends FileSystemTest[FileSystem](
         runLogT(run, p).runEither must beRight(anotherDoc)
       }
 
+      "temp file should be generated in hint directory" >> {
```
mossprescott: Seems like these two tests should use properties for the paths, to look for edge cases and to express the domain of the operation.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
wemrysi: The subtlety comes from how we used to treat the hint path. We used to try to extract a db name, and then use it to form the temp path if successful. This meant that the paths `/foo` and `/foo/` would both result in a tempfile of `/foo/some.tmp` as we just treat the first segment of a path as the db name. Now, we treat files and directories differently, so `/foo` results in `/some.tmp` and `/foo/` results in `/foo/some.tmp`.

it/src/test/scala/quasar/fs/ManageFilesSpec.scala
wemrysi: Sounds good, I'll update them.

wemrysi:
@mossprescott Updated with property tests.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
mossprescott: Good point. That means that `TempFile(/foo)` used to work, but now it will produce `/something.tmp`, which is a path that mongodb can't write to. I don't think anyone was likely to be using a path like that (since you couldn't have actual data at that path), so it may not matter to us that it no longer works.

Essentially we're talking about a limitation of the backend which may be hard to observe except by direct use of the FS algebras. I don't think we have an obvious place to document that sort of thing, nor is it clear who would be interested in it (except us). So I guess it's OK to let this go.
If there was a clear place to document this kind of limitation, I would

mossprescott:
LGTM

PR #1228, opened 2016-07-09
Use `pathy-argonaut` and remove duplicate code in `quasar`

jedesah:


core/src/main/scala/quasar/fs/PathyCodecJson.scala
```diff
@@ -21,19 +21,12 @@ import quasar.Predef._
 import argonaut._, Argonaut._
 import pathy.Path, Path._
 
-// TODO: These belong in argonaut sub-module in scala-pathy
 object PathyCodecJson {
```
wemrysi: What do you think about rename this object to `APath` now?

wemrysi:
One minor comment, otherwise LGTM :+1:

core/src/main/scala/quasar/fs/PathyCodecJson.scala
jedesah: Actually, how about just sticking it in `fs/package.scala` right next to the definition of `APath`?

wemrysi: I'd still prefer the `APath` object, regardless, for organizational reasons and for making it easier to see where the instances come from when looking at imports and if you want to move the object def to the package object, I won't object.

jedesah:
Didn't really mean to merge this as two commits, but I guess it's not too bad.

PR #1229, opened 2016-07-10
[SD-1568] handle var as relation

mossprescott:
Two separate changes:

1. parse `:foo` when a "relation" (table/source/collection) is expected, to a new `VariRelationAST` node, and rewrite these in `substVarsƒ` to `TableRelationAST`
2. move the "absolutization" of paths so that it happens _after_ variable substitution, so that a relative path that is supplied by a variable will get rewritten

### DIFF
--- a/core/src/main/scala/quasar/sql/ast.scala
+++ b/core/src/main/scala/quasar/sql/ast.scala
@@ -207,10 +207,20 @@ sealed trait SqlRelation[A] {
 
   def mapPathsM[F[_]: Monad](f: FUPath => F[FUPath]): F[SqlRelation[A]] = this match {
     case IdentRelationAST(_, _) => this.point[F]
+    case VariRelationAST(_, _) => this.point[F]
     case TableRelationAST(path, alias) => f(path).map(TableRelationAST(_, alias))
     case rel @ JoinRelation(left, right, _, _) =>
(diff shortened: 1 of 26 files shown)


mossprescott:
@sellout: please have a look, too, if you get a chance.

core/src/main/scala/quasar/sql/ast.scala
```diff
@@ -229,6 +239,12 @@ sealed trait NamedRelation[A] extends SqlRelation[A] {
   def aliasName = alias.getOrElse(name)
 }
 
+@Lenses final case class VariRelationAST[A](vari: Vari[A], alias: Option[String])
+    extends NamedRelation[A] {
+  def aliasName = alias.getOrElse(vari.symbol)  // FIXME: not really a useful name
```
sellout: Why isn’t that useful? It seems like `select foo.baz, bar.quux from :foo join bar on foo.id = bar.id` is useful.

sellout:
This looks great to me.

wemrysi:
Ditto, :+1:

core/src/main/scala/quasar/sql/ast.scala
mossprescott: Hmm. That _is_ useful, but feels wrong to me somehow. I think this might be better: `select :foo.baz from :foo`. That currently works, but only if the value is a simple name (not a path).

In any case, I'll remove the comment and leave it as is for now.

PR #1230, opened 2016-07-10
Initial version of normalization.

alissapajer:
Implement EqualT and ShowT.

This allows us to toss a bunch of explicit constraints on `T[EJson]`. We
still have a couple extra `Show` constraints for debugging
normalization.

Coalesce any Map into a preceding ThetaJoin.

Make a Normalizable type class.

MapFunc normalization is now applied across all components in the tree.

Would be better to apply all the QScript transformations across all the
FreeQS structures, but Free gets in the way currently.

Make QScript conversion available to backends.

Adding NameGen to Mergeable.

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+-Work on QScript normalization.
--- a/core/src/main/scala/quasar/fs/QueryFile.scala
+++ b/core/src/main/scala/quasar/fs/QueryFile.scala
@@ -17,11 +17,12 @@
 package quasar.fs
 
 import quasar.Predef._
-import quasar._, RenderTree.ops._
+import quasar._, Planner._, RenderTree.ops._
 import quasar.effect.LiftedOps
 import quasar.fp._
+import quasar.qscript._
 
-import matryoshka._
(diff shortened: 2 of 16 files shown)


alissapajer:
@sellout this is the normalization stuff we paired on - I'm merging it is soon as the build is green

alissapajer:
3/4 passed - `MetadataServiceSpec` failed one

jdegoes:
> Would be better to apply all the QScript transformations across all the FreeQS structures, but Free gets in the way currently.

This is a bit concerning — what's this about?

sellout:
It's getting fixed. Just need to do a fairly mechanical change to some of the algebras so that they work on Free[F, A] in addition to the T[F]. They're still valid QScript, just not optimized. But this is already after the point of stuff that ensures we don't do multiple reads from the same collection, etc., so that's fine.

CHANGELOG/bugfix.md
```diff
@@ -0,0 +1 @@
+-Work on QScript normalization.
```
sellout: If this is the only thing, I wouldn’t bother fixing it – but there should be a space after the hyphen.

sellout:
Looks good!

PR #1231, opened 2016-07-10
SD-1846 Correct generation of mongo collection prefix from dir path

wemrysi:
Previously, when generating a collection prefix from a directory path, we failed to terminate the prefix with a `.`, thus it would match too many collection names, namely anything it was a string prefix of, for example the directory `/foo/bar/Untitled Folder/` would translate to the collection prefix `bar.Untitled Folder` and thus would match all of `bar.Untitled Folder.somefile`, `bar.Untitled Folder 1.somefile`, `bar.Untitled Folder and anything else.anotherfile`.

This corrects the problem by formalizing that collections can only be created from file paths and dir paths represent prefixes. So now the collection prefix from `/foo/bar/Untitled Folder/` is now `bar.Untitled Folder.`.

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+- [SD-1846] Correct generation of mongo collection prefix from dir path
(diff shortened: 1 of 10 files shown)


mossprescott:
LGTM

PR #1233, opened 2016-07-10
[SD-1836] fix reading from a view that refers to another view

mossprescott:
This affected only `ReadFile.open` (so `GET /data`) and not the `QueryFile` operations (`/query`).

Add two missing tests:
- one for reading from a view referring to a view (as opposed to querying from it)
- one for reading from a view that resolves to a constant plan (which was working but untested)

Along the way:
- hide the dangerous `ViewMounter.lookup` function that was mostly being used for existence testing except in the one case where it was doing the wrong thing
- pull out part of `quasar.queryPlan` so it can be applied to LogicalPlan
- fix queries in `ViewFSSpec` to explicitly refer to the right paths instead of patching them up after


core/src/main/scala/quasar/fs/mount/ViewMounter.scala
```diff
@@ -37,25 +37,25 @@ object ViewMounter {
     : Free[S, Vector[AFile]] =
     mntCfgs[S].keys.flatMap(_.foldMap(refineType(_).fold(
       κ(Vector.empty[AFile].point[Free[S, ?]]),
-      f => lookup[S](f).fold(κ(Vector(f)), Vector.empty))))
+      f => exists[S](f).map(if (_) Vector(f) else Vector.empty))))
```
mossprescott: Somehow this became the PR of `if (_)`. Seems like there are probably slicker ways to do the couple of different things that I do with the `Free[S, Boolean]` that `exists` produces. Suggestions?

wemrysi: In this case, since the result is a monoid, you can do `exists[S](f).map(_ ?? Vector(f))`.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -56,25 +55,28 @@ object view {
...
+                  lp => for {
+                    qh <- EitherT(queryUnsafe.eval(lp).run.value)
+                    h  <- seq.next.map(ReadHandle(path, _)).liftM[FileSystemErrT]
+                    _  <- viewState.put(h, ResultSet.Results(qh)).liftM[FileSystemErrT]
+                  } yield h))
+            } yield h).run
+          }
 
-          ViewMounter.lookup[S](path).run.flatMap(_.cata((vOpen _).tupled, fOpen))
+          ViewMounter.exists[S](path).flatMap(if (_) vOpen else fOpen)
```
wemrysi: `ViewMounter.exists[S](path).ifM(vOpen, fOpen)`

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -116,9 +118,9 @@ object view {
     new (WriteFile ~> Free[S, ?]) {
       def apply[A](wf: WriteFile[A]): Free[S, A] = wf match {
         case Open(p) =>
-          ViewMounter.lookup[S](p).run.flatMap(_.cata(
-            κ(emit[S, A](-\/(pathErr(invalidPath(p, "cannot write to view"))))),
-            writeUnsafe.open(p).run))
+          ViewMounter.exists[S](p).flatMap(
```
wemrysi: Can leverage `ifM` here as well.

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -259,8 +261,8 @@ object view {
           }}
 
         case FileExists(file) =>
-          ViewMounter.lookup[S](file).run.flatMap(mc =>
-            if(mc.isDefined) true.point[Free[S, ?]] else query.fileExists(file))
+          ViewMounter.exists[S](file).flatMap(
```
wemrysi: `ifM`

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -62,7 +62,7 @@ object mount {
             dst => EitherT[Free[S, ?], ApiError, String] {
               OptionT[Free[S, ?], (AFile, AFile)](
                 (refineType(src).toOption |@| UriPathCodec.parseAbsFile(dst).map(sandboxAbs)).tupled.point[Free[S, ?]])
-                .flatMap { case sd @ (s, _) => ViewMounter.lookup[S](s).map(κ(sd)) }
+                .flatMap { case sd @ (s, _) => OptionT(ViewMounter.exists[S](s).map(if (_) sd.some else None)) }
```
wemrysi: Can replace `if (_) sd.some else None` with `_ option sd`

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -56,25 +55,28 @@ object view {
...
-                } yield h))
-          }.run
+          def vOpen: Free[S, FileSystemError \/ ReadHandle] = {
+            val readLp = LogicalPlan.Read(path)
+            (for {
+              lp <- ViewMounter.rewrite[S](readLp).leftMap(se => planningFailed(readLp, Planner.InternalError(se.shows)))
+              h  <- preparePlan(lp, off, lim).run.value.fold[EitherT[queryUnsafe.F, FileSystemError, ReadHandle]](
+                e => EitherT[Free[S, ?], FileSystemError, ReadHandle](
+                  // TODO: more sensible error?
+                  Free.point(pathErr(invalidPath(path, e.shows)).left[ReadHandle])),
```
wemrysi: Would a `planningFailed` be more appropriate here?

wemrysi:
Some mechanical comments addressing your questions on `if (_)` alternatives, otherwise LGTM, glad to have this fixed! :+1:

core/src/main/scala/quasar/fs/mount/view.scala
mossprescott: Perhaps but for the moment I didn't fell like sorting it out, since I was just pushing that code around.

PR #1235, opened 2016-07-12
[SD-1844] Fix error where moving fs mounts only updated configuration

wemrysi:
Moving a filesystem mount was being done directly via the `MountConfigs` effect instead of using the `Mounting` effect. The result was that the mount configuration would be updated, but there would be no change in the actual mounted filesystems.

The fix was to revert back to using `Mounting#remount` to move the mount.

In order to expose the problem, the mount service specs were refactored to verify the state of the mounted filesystems as well as their configuration.

### DIFF
--- a/web/src/main/scala/quasar/api/services/mount.scala
+++ b/web/src/main/scala/quasar/api/services/mount.scala
@@ -18,7 +18,6 @@ package quasar.api.services
 
 import quasar.Predef._
 import quasar.api._, ToApiError.ops._
-import quasar.effect.KeyValueStore
 import quasar.fp._
 import quasar.fs.{AbsPath, AFile, APath, sandboxAbs}
 import quasar.fs.mount._
@@ -102,26 +101,16 @@ object mount {
     parse: String => Option[Path[Abs, T, Unsandboxed]],
     typeStr: String
   )(implicit
-    M: Mounting.Ops[S],
-    S0: MountConfigs :<: S
-  ): EitherT[Free[S, ?], ApiError, String] = {
-    val mntCfgs = KeyValueStore.Ops[APath, MountConfig, S]
-
-    parse(dstStr).map(sandboxAbs).cata(dst => {
-      val msg = s"moved ${printPath(src)} to ${printPath(dst)}"
-
-      val op = M.remount[T](src, dst).as(msg).leftMap(_.toApiError).run
-
-      EitherT[Free[S, ?], ApiError, String](mntCfgs.get(src).run.flatMap(_.cata(
-        MountConfig.fileSystemConfig.getOption(_).cata(
-          κ(mntCfgs.move(src, dst).as(msg.right[ApiError])),
-          op),
-        op)))
-      },
-      EitherT.leftU[String](ApiError.apiError(
+    M: Mounting.Ops[S]
+  ): EitherT[Free[S, ?], ApiError, String] =
+    parse(dstStr).map(sandboxAbs).cata(dst =>
+      M.remount[T](src, dst)
+        .as(s"moved ${printPath(src)} to ${printPath(dst)}")
+        .leftMap(_.toApiError),
+      ApiError.apiError(
         BadRequest withReason s"Expected an absolute $typeStr.",
-        "path" := transcode(UriPathCodec, posixCodec)(dstStr)).point[Free[S, ?]]))
-  }
+        "path" := transcode(UriPathCodec, posixCodec)(dstStr)
+      ).raiseError[EitherT[M.F, ApiError, ?], String])
 
   private def mount[S[_]](
     path: APath,
--- a/web/src/test/scala/quasar/api/services/MountServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/MountServiceSpec.scala
@@ -24,10 +24,10 @@ import quasar.effect.KeyValueStore
 import quasar.fp._
 import quasar.fp.free._
 import quasar.fs._, PathArbitrary._
-import quasar.fs.mount._
+import quasar.fs.mount.{MountRequest => MR, _}
 
 import argonaut._, Argonaut._
-import org.http4s._
+import org.http4s._, Status._
 import org.http4s.argonaut._
 import org.specs2.specification.core.Fragments
 import org.specs2.ScalaCheck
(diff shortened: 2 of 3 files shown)


web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -24,10 +24,10 @@ import quasar.effect.KeyValueStore
 import quasar.fp._
 import quasar.fp.free._
 import quasar.fs._, PathArbitrary._
-import quasar.fs.mount._
+import quasar.fs.mount.{MountRequest => MR, _}
```
jedesah: What is this conflicting with?

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -102,26 +101,16 @@ object mount {
...
-        op)))
-      },
-      EitherT.leftU[String](ApiError.apiError(
+    M: Mounting.Ops[S]
+  ): EitherT[Free[S, ?], ApiError, String] =
+    parse(dstStr).map(sandboxAbs).cata(dst =>
+      M.remount[T](src, dst)
+        .as(s"moved ${printPath(src)} to ${printPath(dst)}")
+        .leftMap(_.toApiError),
+      ApiError.apiError(
```
jedesah: Not sure I understand how this fixes the problem from a quick glance. But hey! A shorter more correct solution, I will take that! Does this remove a feature or is this just a simplification?

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -141,10 +157,11 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s with Pa
...
               _    <- M.mountView(f, cfg._1, cfg._2).run.flatMap(orFailF)
 
-              resp <- service(Request(uri = pathUri(f)))
+              r    <- service(Request(uri = pathUri(f)))
+              (res, _) = r
+              body <- lift(res.as[Json]).into[Eff]
             } yield {
-              resp.as[Json].unsafePerformSync must_== cfgStr
-              resp.status must_== Status.Ok
+              (body must_== cfgStr) and (res.status must_== Ok)
```
mossprescott: As a general comment, it would be better if this sort of refactoring was done in at least a separate commit, so we could more easily understand the changes that actually fixed the bug.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -330,18 +377,22 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s with Pa
...
+                dst   =  parent </> f
+                after <- M.lookup(dst).run
               } yield {
-                resp.as[String].unsafePerformSync must_== s"added ${printPath(parent </> f)}"
-                resp.status must_== Status.Ok
-
-                after must beSome(MountConfig.viewConfig(cfg))
+                (body must_== s"added ${printPath(dst)}")         and
+                (res.status must_== Ok)                           and
+// TODO: This fails due to how we're bypassing Mounting for views, enable and fix.
```
mossprescott: Is there a task to come back to this after 3.0(.1)?

web/src/main/scala/quasar/api/services/mount.scala
mossprescott: I'm equally mystified. What was this code trying to accomplish by special-casing fileSystemConfigs? You've essentially reverted part of https://github.com/quasar-analytics/quasar/commit/90eef735f9c3fea2e89b9881361e7b9f35d2b89a; did you understand the intention of that change? Why was it not actually necessary?

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
wemrysi: Nothing, just a shorter name.

jedesah: I think we should avoid renaming imports unless we have a pretty good reason to do so. I also think that in this particular case the new name is on the short (almost impossible to guess) side anyway.

web/src/main/scala/quasar/api/services/mount.scala
wemrysi: Yes, I forgot to mention this pretty much just reverted that commit.

As for why the original change was made, I do not know and I don't think it was necessary. I also don't think `MountConfigs` should appear anywhere in the api as it is an implementation detail of a particular implementation of `Mounting`, i.e. using it assumes how `Mounting` will be interpreted.

As for exactly what was going on, the actual interpreter of `Mounting` does a couple things when mounting: 1) validates the new mount doesn't violate structure (like mounting a filesystem inside another one), 2) does the mount (i.e. connect to mongo, compile sql, etc) and then, if (2) succeeded, updates the configuration with the new mount. The previous code essentially skipped (1) and (2) for filesystem mounts.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
wemrysi: Agreed, I was a bit cavalier here.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
wemrysi: https://slamdata.atlassian.net/browse/SD-1859

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
wemrysi: In this case I needed access to `MountRequest` constructors as well as prisms and, while I could have just imported all of `MountRequest._` into scope, I thought this way made it easier to tell where those identifiers were coming from.

web/src/main/scala/quasar/api/services/mount.scala
mossprescott: It seems like at least some of that might call for special treatment in the case of a move (no need to drop the connector and re-test the connection, for example), but I suppose `remount` could be worrying about it at that level. And just to clarify your description, "doing the mount" also involves updating the filesystem interpreter with the new hierarchy, right? Is that what was really causing the crazy behavior seen in 1844?

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
mossprescott: OK, that's good. Now can you explain what it is that we're not able to test right now? Is it just something the test can't verify, or something that's actually not working right?

web/src/main/scala/quasar/api/services/mount.scala
wemrysi: Aye, but if we want to talk about "moving" first-class, then `Mounting` should get a `Move` primitive.

Regarding "doing the mount", yes, in our current implementation it does and that was also being skipped and yes that is what was causing trouble.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
wemrysi: This asserts that the effect associated with "mounting a view" happens (and also that the view is "mounted" using the `Mounting` api) and it's not so much that we're not able to test this but that it fails and making it pass is a non-trivial effort.

The only reason the service behaves is that our assumption that `Mounting` will be interpreted into `MountConfigs` (among other things) happens to be correct at the moment. If we changed the implementation of `qusar.fs.mount.Mounter` to use some other mechanism, this service wouldn't work.

wemrysi: @mossprescott Turns out the effort wasn't as bad as I thought as I was able to get all these tests passing as part of the fix for SD-1833.

wemrysi:
Any other comments? What needs to happen for this to be ready?

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
mossprescott: That's good news and makes me much more comfortable merging this.

mossprescott:
I'd love for @drostron to have a chance to look at this, but under the circumstances I say let's go ahead.

LGTM.

PR #1236, opened 2016-07-12
Misc QScript bug fixes related to source merging.

alissapajer:


alissapajer:
@sellout here are the bug fixes I made separated from all the other work regarding general handling of invoke.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -74,11 +74,22 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
   type Inner = T[F]
   type InnerPure = T[QScriptPure[T, ?]]
 
-  type Pures[A] = List[F[A]]
-  type DoublePures = (Pures[Unit], Pures[Unit])
-  type TriplePures = (Pures[Unit], Pures[Unit], Pures[Unit])
+  type Fs[A] = List[F[A]]
 
-  type DoubleFreeMap = (FreeMap[T], FreeMap[T])
+  case class ZipperAcc(
```
alissapajer: These tuples have real names now.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -97,57 +108,70 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
...
-      ma.mergeSrcs(lm, rm, l, r).fold[TriplePures \/ ListF[F[Unit], (DoubleFreeMap, DoublePures)]]({
-          case AbsMerge(inn, lmf, rmf) => ConsF(inn, ((lmf, rmf), (ls, rs))).right[TriplePures]
-        }, (Nil, l :: ls, r :: rs).left)
+      ma.mergeSrcs(lm, rm, l, r).fold[ZipperAcc \/ ListF[F[Unit], (FreeMapLR, ZipperLR)]]({
+          case SrcMerge(inn, lmf, rmf) =>
+            ConsF(inn, (FreeMapLR(lmf, rmf), ZipperLR(ls, rs))).right[ZipperAcc]
+        }, ZipperAcc(Nil, l :: ls, r :: rs, lm, rm).left)
     }
-    case ((_, _), (l, r)) => (Nil: Pures[Unit], l, r).left.point[State[NameGen, ?]]
+    case (FreeMapLR(lm, rm), ZipperLR(l, r)) =>
```
alissapajer: We were dropping the `FreeMap[T]` information on ground - so this bug fix instead carries it through in the `ZipperAcc` so we can `rebase` with it later.

core/src/main/scala/quasar/qscript/ThetaJoin.scala
```diff
@@ -86,7 +86,7 @@ object ThetaJoin {
         p2: ThetaJoin[IT, Unit]) =
         OptionT(state(
           if (p1 == p2)
-            AbsMerge[ThetaJoin[IT, Unit], FreeMap[IT]](p1, UnitF, UnitF).some
+            SrcMerge[ThetaJoin[IT, Unit], FreeMap[IT]](p1, left, right).some
```
alissapajer: Use the real `FreeMap[T]`s.

core/src/main/scala/quasar/qscript/QScriptBucket.scala
```diff
@@ -130,7 +130,13 @@ object QScriptBucket {
         left: FreeMap[IT],
         right: FreeMap[IT],
         p1: QScriptBucket[IT, Unit],
-        p2: QScriptBucket[IT, Unit]) = OptionT(state(None))
+        p2: QScriptBucket[IT, Unit]) =
+        OptionT(state {
+          if (p1 == p2)
```
alissapajer: If they're equal they merge.

core/src/main/scala/quasar/qscript/package.scala
```diff
@@ -83,19 +82,7 @@ package object qscript {
 
   def UnitF[T[_[_]]] = ().point[Free[MapFunc[T, ?], ?]]
 
-  final case class AbsMerge[A, B](src: A, left: B, right: B)
+  final case class SrcMerge[A, B](src: A, left: B, right: B)
 
-  type Merge[T[_[_]], A] = AbsMerge[A, FreeMap[T]]
-  type MergeJoin[F[_], A] = AbsMerge[A, Free[F, Unit]]
```
alissapajer: These aliases were barely used and not really gaining us anything, so they're gone.

core/src/main/scala/quasar/qscript/Bucketable.scala
```diff
@@ -42,4 +42,12 @@ object Bucketable {
         fg.run.bitraverse(FB.digForBucket[H](_), GB.digForBucket[H](_)) ∘
           (Coproduct(_))
     }
+
+  implicit def const[T[_[_]]]:
+      Bucketable.Aux[T, Const[DeadEnd, ?]] =
```
sellout: The reason I hadn’t already moved this here is that it mentions `DeadEnd` explicitly. I think it’d be fine to really make this `implicit def const[T[_[_]], A]: Bucketable.Aux[T, Const[A, ?]]` and have it here, though.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -74,11 +74,22 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
...
-  type TriplePures = (Pures[Unit], Pures[Unit], Pures[Unit])
+  type Fs[A] = List[F[A]]
 
-  type DoubleFreeMap = (FreeMap[T], FreeMap[T])
+  case class ZipperAcc(
+    acc: Fs[Unit],
+    lSide: Fs[Unit],
+    rSide: Fs[Unit],
+    lMap: FreeMap[T],
+    rMap: FreeMap[T])
```
sellout: Now that we have `ZipperLR` and `FreeMapLR`, does it make sense to use those here rather than holding unwrapped versions?

alissapajer: Yeah, the semantics of both those tuples are exactly what's being accumulated here. And making that change would actually simplify the constructions of `ZipperAcc` in a couple cases. Also I think it will make the type clearer about what's being accumulated. So will do.

core/src/main/scala/quasar/qscript/Bucketable.scala
alissapajer: Right, I'll param it over `A`. It's not using `DeadEnd` in any special way.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -97,57 +108,70 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
...
-            .get(lTail.reverse.ana[T, CoEnv[Unit, F, ?]](delinearizeFreeQS[F, Unit] ⋙ (CoEnv(_)))),
-          foldIso(CoEnv.freeIso[Unit, F])
-            .get(rTail.reverse.ana[T, CoEnv[Unit, F, ?]](delinearizeFreeQS[F, Unit] ⋙ (CoEnv(_)))))
+  def merge(left: Inner, right: Inner): State[NameGen, SrcMerge[Inner, Free[F, Unit]]] = {
+    val lLin: Fs[Unit] = left.cata(linearize).reverse
+    val rLin: Fs[Unit] = right.cata(linearize).reverse
+
+    elgotM(
+      FreeMapLR(UnitF[T], UnitF[T]),
+      ZipperLR(lLin, rLin))(
```
sellout: So, this looks wrong to me (it was wrong when I wrote it in the last PR, just didn’t notice). The two args here should be in a tuple, and we shouldn’t rely on auto-tupling.

It looks like `-Xlint` (which would disable “features” like this) is disabled outside of the build project. Should revisit enabling that, see how problematic it is.

core/src/main/scala/quasar/qscript/QScriptBucket.scala
sellout: Should use `≟` (or `===`) – at some point we’ll get around to filling in the gaps so we can disable `==` via a wart.

sellout:
Cool, just a few minor things.

core/src/main/scala/quasar/qscript/QScript.scala
alissapajer: haha so I just read this comment, then jumped back to my compile errors, and the changes I just made to the accumlator case classes in the zipper caused this to give a compile error instead of auto-tupling like it used to happily do... not sure exactly why, but in any case, consider it done. :)

core/src/main/scala/quasar/qscript/QScriptBucket.scala
alissapajer: oh yeah, I just type `==` out of habit :) will change.

alissapajer: doing a quick audit of all of qscript package to remove `==` everywhere

alissapajer: @sellout ok all `==` is removed from qscript package!

alissapajer:
@sellout ok right back at you :) all discussed changes have been addressed.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -76,21 +76,19 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
...
-  case class FreeMapLR(
+  case class ZipperSides(
     lSide: FreeMap[T],
     rSide: FreeMap[T])
 
+  case class ZipperTails(
+    lTail: Fs[Unit],
+    rTail: Fs[Unit])
+
+  case class ZipperAcc(
```
alissapajer: These should all have better names now.

core/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -125,7 +125,7 @@ object QScriptCore {
         }
     }
 
-  implicit def mergeable[T[_[_]]]:
+  implicit def mergeable[T[_[_]]: EqualT]:
```
sellout: Yay – adding `EqualT` makes figuring out equality constraints much less guessy. Might go back to the `EqualF` approach (vs `Delay`), too.

sellout:
Looks great. 👍🏾 (modulo the build passing).

alissapajer:
2/4 passed `DataServicesSpec` fail in one, as well as maybe a hang in the other.

PR #1237, opened 2016-07-12
[SD-1833] Fix an issue where attempting to replace a view mount would fail

wemrysi:
The fix reverts all of the mount service changes made in

8c1eff6614cfe055ee1039ace88369773308e355
f6e846d0c299d7eaaae0c17a4614db33e3c3583b

the problem stemmed from `ViewMounter.mount` being used in
`EvaluatorMounter`. `EvaluatorMounter.mount/unmount` exist to
handle `MountRequests` issued by `Mounter`. If requests are
handled successfully, then `Mounter` updates the mount configuration
appropriately.

When `ViewMounter.mount` began modifying mount configuration itself
this began to cause conflicts as `Mounter` would end up calling
`EvaluatorMounter.mount`, which handed off to `ViewMounter.mount`
which ended up modifying configuration. When control returned to
`Mounter`, it would find the configuration had changed and abort
the mount.

The fix was to only perform validation on the candidate view mount
in `EvaluatorMounter` rather than modifying the configuration directly.

### DIFF
--- a/it/src/test/scala/quasar/server/ServiceSpec.scala
+++ b/it/src/test/scala/quasar/server/ServiceSpec.scala
@@ -17,17 +17,20 @@
 package quasar.server
 
 import quasar.Predef._
+import quasar.{TestConfig, Variables}
 import quasar.config.{ConfigOps, FsPath, WebConfig}
 import quasar.main.MainErrT
 import quasar.api.UriPathCodec
 import quasar.fs._, mount._
 import quasar.server.Server.QuasarConfig
-import quasar.TestConfig
+import quasar.sql.{fixParser, Query}
 
 import java.io.File
 
+import argonaut._, Argonaut._
 import org.http4s.Uri.Authority
 import org.http4s._, Status._
+import org.http4s.argonaut._
 import org.specs2.mutable
 import org.specs2.scalaz.DisjunctionMatchers._
 import pathy.Path._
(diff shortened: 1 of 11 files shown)


web/src/test/scala/quasar/api/services/MountServiceSpec.scala
```diff
@@ -525,25 +517,6 @@ class MountServiceSpec extends Specification with ScalaCheck with Http4s with Pa
           }
         }
 
-        "be 400 with unbound variable in view" ! prop { (parent: ADir, f: RFile) =>
```
jedesah: Not obvious to me why this was removed.

it/src/test/scala/quasar/server/ServiceSpec.scala
```diff
@@ -104,6 +107,34 @@ class ServiceSpec extends mutable.Specification {
       r.map(_.status) must beRightDisjunction(Ok)
     }
 
+    "[SD-1833] replace view" in {
+      val port = Http4sUtils.anyAvailablePort.unsafePerformSync
+      val sel1 = "sql2:///?q=%28select%201%29"
```
jedesah: What does `%28` and `%29`  do here? Forgive my url encoding ignorance.

mossprescott: `(` and `)`

jedesah: Ah, I see.

web/src/test/scala/quasar/api/services/MountServiceSpec.scala
wemrysi: It supposed a particular implementation of what it means to "mount" a view (namely the one in `ViewMounter.mount`). Now that the service only deals with the `Mounting` abstraction it only needs to verify that any `MountingError`s emitted are faithfully communicated in the response, which is already happening in the example on line 536.

it/src/test/scala/quasar/server/ServiceSpec.scala
wemrysi: I don't know, I copied it from the test above it, my guess would be a backtick or quote of some sort.

jedesah:
LGTM.

drostron:
Great to see some reservations I had with this code addressed. Thanks @wemrysi.

PR #1239, opened 2016-07-13
Separating product bucket

sellout:
This is the setup for the Read rewrite pass. It extracts projections from QScriptBucket and moves Map from SourcedPathable to QScriptCore (since Map was only Pathable in the case where it consisted of only projections, which is now handled by explicit projection nodes). It also adds a `simplifyProjections` pass which converts any projection nodes into Map nodes (this is what `elideBuckets` did with projection nodes, but now we keep them around after eliding buckets for the static path handling).

### DIFF
--- a/core/src/main/scala/quasar/qscript/QScript.scala
+++ b/core/src/main/scala/quasar/qscript/QScript.scala
@@ -63,16 +63,16 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
            SP: SourcedPathable[T, ?] :<: F,
            QC: QScriptCore[T, ?] :<: F,
            TJ: ThetaJoin[T, ?] :<: F,
+           PB: ProjectBucket[T, ?] :<: F,
            QB: QScriptBucket[T, ?] :<: F,
            // TODO: Remove this one once we have multi-sorted AST
            FI: F :<: QScriptInternal[T, ?],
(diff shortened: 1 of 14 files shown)


sellout:
@wemrysi You might also want to look at some of this, see if any of the Coproduct stuff can be handled differently. I’d like to be able to “shrink” coproducts more, but have run into various problems doing so. Once the current QScript kerfuffle is sorted out, we can talk about it.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -332,68 +330,43 @@ class Transform[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]: Traverse](
...
-          src,
-          inner,
-          { case SrcMerge(src, mfl, mfr) =>
-            (TJ.inj(src).embed, rebase(bucket, mfl), mfr)
-          })
-      case -\/(BucketField(src, _, name)) =>
-        mergeTheta[(Inner, FreeMap[T], FreeMap[T])](
-          src,
+    bucketDisj.fold(
+      bucketable.applyBucket[F](_, inner)(mergeTheta), {
```
alissapajer: +1 to `applyBucket` as a typeclass

alissapajer:
@sellout I'd like to base my current work off of this, so +1 to getting it merged ASAP

PR #1241, opened 2016-07-13
Migrate some `quasar.effect` related code from Quasar Advanced

jedesah:


wemrysi:
@jedesah So I don't have to compare, is the new `Read` just a verbatim copy from advanced?

wemrysi:
Also, thanks for doing this, something I can cross of my list!

jedesah:
Yup! Verbatim copy!

wemrysi:
:

PR #1242, opened 2016-07-14
SparkCore Init & ReadFile implementation

rabbitonweb:
This PR implements `ReadFile` algebra for SparkCore connector. Details in https://github.com/quasar-analytics/quasar/issues/1253

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -165,7 +165,7 @@ lazy val root = project.in(file("."))
 //          |
           core,
 //      / / | \ \
-  mongodb, skeleton, postgresql,
+  mongodb, skeleton, postgresql, sparkcore,
 //      \ \ | / /
           main,
 //        /  \
@@ -212,6 +212,7 @@ lazy val main = project
   .dependsOn(
     mongodb    % BothScopes,
     skeleton   % BothScopes,
+    sparkcore   % BothScopes,
     postgresql % BothScopes)
   .settings(oneJarSettings: _*)
   .settings(publishSettings: _*)
@@ -241,6 +242,16 @@ lazy val postgresql = project
   .settings(oneJarSettings: _*)
   .settings(publishSettings: _*)
   .enablePlugins(AutomateHeaderPlugin)
+
+lazy val sparkcore = project
+  .settings(name := "quasar-sparkcore-internal")
+  .dependsOn(core % BothScopes)
+  .settings(oneJarSettings: _*)
+  .settings(publishSettings: _*)
+  .settings(libraryDependencies +=
+    "org.apache.spark"  %  "spark-core_2.11"           % "1.6.2")
+  .enablePlugins(AutomateHeaderPlugin)
+
 
 // frontends
 
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_readfile.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_readfile.scala
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.physical.sparkcore.fs
+
+import quasar.Predef._
+import quasar.fs._
+import quasar.fp.free._
+import quasar.effect.Read
+
+import java.io.File
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd._
+import pathy.Path.posixCodec
+import scalaz._
+import scalaz.concurrent.Task
+
+object local_readfile {
+
+  def rddFrom[S[_]](f: AFile)(implicit read: Read.Ops[SparkContext, S]): Free[S, RDD[String]] =
+    read.asks(sc => sc.textFile(posixCodec.unsafePrintPath(f)))
+
+  def fileExists[S[_]](f: AFile)(implicit s0: Task :<: S): Free[S, Boolean] =
+    injectFT[Task, S].apply(Task.delay(new File(posixCodec.unsafePrintPath(f)).exists()))
+
+  def input[S[_]](implicit read: Read.Ops[SparkContext, S], s0: Task :<: S) =
+    Input(f => rddFrom(f), f => fileExists(f))
+
+}
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.physical.sparkcore.fs
+
+import quasar.Predef._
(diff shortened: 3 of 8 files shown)


rabbitonweb:
PR ready to review!

There are still few things to be improved (below). Those will be added asap, however the "main" functionality is ready and will not suffer from those changes. 


Missing bits are:
1. write a test ` pass test open file -> read chunk until empty vector -> close == file content`, could anyone willing to PP on it?
2. tests should call `sc.stop()` regardless of whether they've passed or fail (there can be only one SC in the single JVM)
3. `SparkContextIO` should be just type alias to `ReaderT`
4. `ReadFile.sparkCursorL` should get  better name (still thinking on the name :) )

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
```diff
@@ -0,0 +1,134 @@
...
+      Lens.lensu[SparkCursor, Option[RDD[Data]]] (
+        (spCur, v) => spCur.copy(rdd = v),
+        _.rdd
+      )
+
+    def maybeRddL(h: ReadHandle): ReadState @?> Option[RDD[Data]] =
+      ~sparkCursorL compose PLens.somePLens compose ~cursorL(h)
+
+  }
+
```
rabbitonweb: @jedesah mentioned changing signature to `ReadFile ~> Free` (correct me if I'm wrong) but I don't really want to do it on my own, would really benefit from PP here

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_readfile.scala
```diff
@@ -0,0 +1,37 @@
...
+import quasar.Predef._
+import quasar.fs._
+import org.apache.spark.rdd._
+
+import java.io.File
+import pathy.Path.posixCodec
+
+import scalaz._
+import Scalaz._
+
```
rabbitonweb: `spark-core` is a foundation (a base) to build future Spark connectors. First production ready connector will be `spark-hdfs` that will work on Hadoop distributed file system. However for now, to "prove" correctness of `spark-core` I implement also sort of "dummy" connectr  called spark-local.

the idea of spark-local is to run on a single-node spark cluster where all files are on the same file system (not a distributed file system)

wemrysi: Does spark have different capabilities or limitations based on the backend you use? Just trying to understand what will be different about `spark-hdfs` vs, say, `spark-cassandra`.

project/Dependencies.scala
```diff
@@ -15,7 +15,8 @@ object Dependencies {
   private val scalazVersion   = "7.2.3"
   private val scodecVersion   = "1.0.10"
   private val slcVersion      = "0.4"
-
+  private val sparkVersion    = "1.6.2"
```
jedesah: I don't think we need this anymore.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_readfile.scala
rabbitonweb: @wemrysi let me answer with an example:

for ReadFile majority of code will be common for both `spark-hdfs` and `spark-cassandra`, the only difference is how we going to construct the `RDD[String` and those will be implemented by `spark-hdfs` and `spark-cassandra` separately

for WriteFile there is really no common code so each of the subprojects will have to provide its own interpretation of what that is

rabbitonweb:
@jr: rdy right?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_readfile.scala
```diff
@@ -0,0 +1,43 @@
...
+import java.io.File
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd._
+import pathy.Path.posixCodec
+import scalaz._
+import scalaz.concurrent.Task
+
+object local_readfile {
```
wemrysi: We've tended to avoid snake case thus far, maybe just name the object `local`?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
```diff
@@ -0,0 +1,101 @@
...
+  rddFrom: AFile => Free[S, RDD[String]],
+  fileExists: AFile => Free[S, Boolean]
+)
+
+object readfile {
+
+  import ReadFile.ReadHandle
+
+  def interpret[S[_]](input: Input[S])(implicit
+    s0: KeyValueStore[ReadHandle, SparkCursor, ?] :<: S,
```
wemrysi: Something we've started doing elsewhere is making aliases for these partially applied effects, i.e.
```scala
type SparkReadCursors[A] = KeyValueStore[ReadHandle, SparkCursor, A]

object SparkReadCursors {
  def Ops[S[_]](implicit S: SparkReadCursors :<: S) =
    KeyValueStore.Ops[ReadHandle, SparkCursor, S]
}
```
which makes them a bit more meaningful, I think.

wemrysi:
A couple of minor comments, but otherwise LGTM :shipit:

build.sbt
```diff
@@ -202,14 +202,16 @@ lazy val core = project
     ScoverageKeys.coverageMinimum := 79,
     ScoverageKeys.coverageFailOnMinimum := true,
     buildInfoKeys := Seq[BuildInfoKey](version, ScoverageKeys.coverageEnabled),
-    buildInfoPackage := "quasar.build")
+    buildInfoPackage := "quasar.build"
+  )
```
jedesah: This is an unnecessary change and not the style adopted in the rest of the codebase.

build.sbt
```diff
@@ -240,6 +242,16 @@ lazy val postgresql = project
   .settings(publishSettings: _*)
   .enablePlugins(AutomateHeaderPlugin)
 
+lazy val sparkcore = project
+  .settings(name := "quasar-sparkcore-internal")
+  .dependsOn(core % "test->test;compile->compile")
+  .settings(oneJarSettings: _*)
+  .settings(publishSettings: _*)
+  .settings(libraryDependencies +=
+    "org.apache.spark"  %  "spark-core_2.11"           % "1.6.2")
```
jedesah: Perhaps we can use `%%` operator?

build.sbt
```diff
@@ -240,6 +242,16 @@ lazy val postgresql = project
   .settings(publishSettings: _*)
   .enablePlugins(AutomateHeaderPlugin)
 
+lazy val sparkcore = project
+  .settings(name := "quasar-sparkcore-internal")
+  .dependsOn(core % "test->test;compile->compile")
```
jedesah: We can use `BothScopes` now.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_readfile.scala
rabbitonweb: this will probably end up in it's own subproject (will emerge spark-local subproject), can add it to single package for now

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
rabbitonweb: This is what I;ve seen in Noel Welsh blog post when he was introducing monad trnasformes - he used the same trick here

So should I go with type aliasing like that and remove all `?`  ??

build.sbt
rabbitonweb: Will revert

build.sbt
rabbitonweb: I will leave it for now as it is as there is one issue I have to talk with John regarding scala version & spark

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
wemrysi: I think it's a good idea, helps to separate semantics from implementation.

build.sbt
```diff
@@ -240,6 +241,16 @@ lazy val postgresql = project
   .settings(publishSettings: _*)
   .enablePlugins(AutomateHeaderPlugin)
 
+lazy val sparkcore = project
+  .settings(name := "quasar-sparkcore-internal")
+  .dependsOn(core % BothScopes)
+  .settings(oneJarSettings: _*)
+  .settings(publishSettings: _*)
+  .settings(libraryDependencies +=
+    "org.apache.spark"  %  "spark-core_2.11"           % "1.6.2")
```
jedesah: Still no conclusion on this from your discussion with John? I am not sure I see why we would ever want to use a version that is not compiled with the same of scala as our own. Anyway, I trust you have a good reason for wanting to do it this way.

jedesah:
I think there are a few of @wemrysi's concerns that you did not fully address.

I think it's OK to merge, I expect it to change some more soon so I won't be too picky since this PR has been open for a while now and it would be good to have in `master` for everyone to look at and base work on.

I noticed that some commits have compile errors in them, so I won't merge like we talked about in the retrospective meeting today. Instead, I will mark it as ready and let you rebase and merge in order to remove the compile errors.

PR #1265, opened 2016-07-18
Consistently apply type-checks around view interpretation

mossprescott:
See SD-1869

Run all regression test queries as if the query was mounted as a view, by interpreting `ReadFile.scanAll`. This turned up a number of queries that were affected by the rewriting:
- some simply had their results nested under `value`
- some joins returned no results (as observed in SD-1869)
- some joins led to endless map-reduce

The fix is to be more consistent about how and when type-checks are applied the LogicalPlan:
- queries submitted via `QueryFile.EvaluatePlan` and `.ExecutePlan` will already be "prepared" with type-checking (as before)
- that type-checking is now stripped by `ViewMounter` before view queries are expanded, and then the expanded LP is re-prepared
- to that end, more of the LP transformation steps are now exposed in quasar package obj
- this also allows for more clarity about when it's necessary to "refine" an LP to a constant, which also cleaned up a TODO

### DIFF
--- a/CHANGELOG/bugfix.md
+++ b/CHANGELOG/bugfix.md
@@ -1,0 +1,1 @@
+- [SD-1869] fix for reading from views containing certain queries including most joins
--- a/core/src/main/scala/quasar/fs/mount/ViewMounter.scala
+++ b/core/src/main/scala/quasar/fs/mount/ViewMounter.scala
@@ -24,7 +24,7 @@ import quasar.fs._
 import quasar.sql.Sql
 
 import eu.timepit.refined.auto._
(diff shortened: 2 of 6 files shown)


jdegoes:
@sellout @drostron Please review this ASAP, would like to get into SlamData 3.0.2 today.

sellout:
👍🏾

drostron:
👍🏽

jdegoes:
@mossprescott OK, looks like this one's all set. After you merge it in, please update Quasar Advanced with this version of Quasar so we have the fix there, too.

PR #1268, opened 2016-07-19
[SD-1871] Add a write-through cache interpreter to KeyValueStore

wemrysi:
Adds a transformation from `KeyValueStore` into `Free[Coproduct[Cache[KeyValueStore, ?], KeyValueStore, ?], ?]`. This allows any `KeyValueStore` implementation to be used as a cache for another.

This particular cache is write-through and only caches reads, however it caches _all_ reads (a read will never be sent to the backing store, even on miss) and thus must be pre-warmed in order to return existing values, otherwise it will only return values that have changed.

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/package.scala
+++ b/core/src/main/scala/quasar/fs/mount/package.scala
@@ -18,7 +18,7 @@ package quasar.fs
 
 import quasar.Predef.Map
 import quasar.effect._
-import quasar.fp.{TaskRef}
+import quasar.fp.{evalNT, TaskRef}
 import quasar.fp.free, free._
 
 import scala.collection.immutable.Vector
@@ -30,8 +30,6 @@ import scalaz.concurrent.Task
 package object mount {
   type MntErrT[F[_], A] = EitherT[F, MountingError, A]
 
-  type MountConfigs[A] = KeyValueStore[APath, MountConfig, A]
-
   type MountingFileSystem[A] = Coproduct[Mounting, FileSystem, A]
 
   def interpretMountingFileSystem[M[_]](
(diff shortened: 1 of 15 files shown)


jdegoes:
@wemrysi Does this improve the performance of Cloud?

wemrysi:
@jdegoes It will be used to in quasar-advanced, yes.

core/src/main/scala/quasar/fs/mount/package.scala
```diff
@@ -60,7 +71,9 @@ package object mount {
       KeyValueStore.Ops[ReadFile.ReadHandle, ResultSet, S]
 
     def toTask(initial: ViewHandles): Task[ViewState ~> Task] =
-      TaskRef(initial) map KeyValueStore.fromTaskRef
+      TaskRef(initial) map { ref =>
+        foldMapNT(AtomicRef.fromTaskRef(ref)) compose KeyValueStore.toAtomicRef
```
mossprescott: This fragment seems to appear every time. Why not keep the old `KeyValueStore.fromTaskRef`, with this new implementation?

core/src/main/scala/quasar/fs/mount/package.scala
```diff
@@ -40,6 +38,19 @@ package object mount {
...
+  object MountConfigs {
+    type MCfgs = Map[APath, MountConfig]
+
+    def ephemeral[F[_]: Monad]: MountConfigs ~> F =
+      evalNT[F, MCfgs](Map()) compose
+      KeyValueStore.toState[StateT[F, MCfgs, ?]](Lens.id[MCfgs])
+
+    val atomicRef: MountConfigs ~> Free[AtomicRef[MCfgs, ?], ?] =
+      KeyValueStore.toAtomicRef[APath, MountConfig]
+  }
```
mossprescott: Once again, I would have preferred a separate commit with this refactoring, if practical.

wemrysi:
Not having much success with this approach, going to try a different direction.

PR #1269, opened 2016-07-19
Upgrade various dependencies

jedesah:
Cannot merge this just yet. Need to wait for this to get released: https://github.com/http4s/http4s-argonaut61/pull/2


mossprescott:
LGTM when it builds. I just changed it to `in progress` so I wouldn't keep checking it.

PR #1272, opened 2016-07-19
Toward typed equality in all things

paulp:
Primarily looking for feedback though I wouldn't be averse to merging it as-is either.

### DIFF
--- a/core/src/main/scala/quasar/fs/package.scala
+++ b/core/src/main/scala/quasar/fs/package.scala
@@ -37,13 +37,13 @@ package object fs extends PhysicalErrorPrisms {
   type RDir  = RelDir[Sandboxed]
   type AFile = AbsFile[Sandboxed]
   type RFile = RelFile[Sandboxed]
-  type APath = AbsPath[_]
-  type RPath = RelPath[_]
+  type APath = AbsPath[scala.Any]
+  type RPath = RelPath[scala.Any]
 
   type FilePath[B] = pathy.Path[B,File,Sandboxed]
   type DirPath[B]  = pathy.Path[B,Dir, Sandboxed]
-  type FPath = FilePath[_]
-  type DPath = DirPath[_]
+  type FPath = FilePath[scala.Any]
+  type DPath = DirPath[scala.Any]
 
   type PathSegment = DirName \/ FileName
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/quasar/fs/package.scala
```diff
@@ -37,13 +37,13 @@ package object fs extends PhysicalErrorPrisms {
   type RDir  = RelDir[Sandboxed]
   type AFile = AbsFile[Sandboxed]
   type RFile = RelFile[Sandboxed]
-  type APath = AbsPath[_]
-  type RPath = RelPath[_]
+  type APath = AbsPath[scala.Any]
```
wemrysi: Mind explaining why this is an improvement? Is it just that it is no longer existential?

wemrysi: Ah, nm, just read the commit msg.

wemrysi:
I think this looks good, appreciate the tidying.

wemrysi:
Unless anyone else has an opinion, feel free to merge (using the `merge` script in `slamdata/devtools`, if you haven't found it already).

paulp:
I can do that, but I'm not used to merging my own pull requests, and I'm not completely clear on what the script accomplishes. Can you perhaps merge this one if that's long enough to allow others to comment, and I'll look to acquire a clearer idea of prevailing policy?

wemrysi:
@paulp Sure I can merge, would you mind adding a file named `revision.md` in the `CHANGELOG` directory containing a line(s) describing the change(s) (see the other files in there for examples).

sellout:
You’ll need to add a “CHANGELOG/revision.md” (literally called “revision” – the merge script renames it to the correct version number) containing a bullet point or more about what’s changed.

The script also ensures that you’ve rebased on master (but mostly it handles incrementing versions properly, since we publish a release on every merge). And we tend to squash commits – doesn’t have to be a single commit, but logical pieces.

wemrysi:
As for squashing, I think these commits are fine as-is. Or at least I'd like for the existential one to survive for the nice commit message.

paulp:
Indeed the commit separation here was an intentional un-squash.

I'll push another one as described but my initial reaction is that it is largely busywork and better suited to tooling. I'll follow that up independently.

paulp:
I rebased on master as well - it's unwise that we have so thoroughly normalized the practice of force pushing, but the way github has structured things there are no good alternatives.

PR #1275, opened 2016-07-21
SD-1871 Mounting HavingPrefix

wemrysi:
Adds two new terms to `Mounting`: `Move` and `HavingPrefix`. The former is an optimization to allow us not to have to unmount/remount a mount just to change location. `HavingPrefix` exists as a primitive to allow discovering many mounts at once. This allows us to implement the `/metadata/fs` endpoint much more efficiently.

Both also are needed for the next phase, which is to rewrite the view mounter in terms of the `Mounting` algebra so it can use different implementations.

While I was in there, I also got rid of `EitherT` in `Mounting.Ops` in favor of `Failure` effects, which makes them a bit nicer to use in many cases (@mossprescott you may find this interesting).

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/MountType.scala
+++ b/core/src/main/scala/quasar/fs/mount/MountType.scala
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.fs.mount
+
+import quasar.Predef._
+import quasar.SKI._
+import quasar.fs.FileSystemType
+
+import monocle.Prism
+import scalaz._, Scalaz._
+
+sealed abstract class MountType {
+  import MountType._
+
+  def fold[X](fs: FileSystemType => X, v: => X): X =
+    this match {
+      case FileSystemMount(t) => fs(t)
+      case ViewMount          => v
+    }
+
+  override def toString: String =
+    this.shows
+}
+
+object MountType {
+  final case class FileSystemMount(fsType: FileSystemType) extends MountType
+  case object ViewMount extends MountType
+
+  val fileSystemMount: Prism[MountType, FileSystemType] =
+    Prism.partial[MountType, FileSystemType] {
+      case FileSystemMount(fs) => fs
+    } (FileSystemMount)
+
+  val viewMount: Prism[MountType, Unit] =
+    Prism.partial[MountType, Unit] {
+      case ViewMount => ()
+    } (κ(ViewMount))
+
+  implicit val order: Order[MountType] =
+    Order.orderBy(mt => (viewMount.isMatching(mt), fileSystemMount.getOption(mt)))
+
+  implicit val show: Show[MountType] =
+    Show.shows(_.fold(t => s"FileSystemMount(${t.shows})", "ViewMount"))
+}
--- a/core/src/main/scala/quasar/fs/mount/Mounting.scala
+++ b/core/src/main/scala/quasar/fs/mount/Mounting.scala
@@ -14,6 +14,22 @@
  * limitations under the License.
  */
 
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package quasar.fs.mount
 
 import quasar.Predef._
@@ -66,8 +82,6 @@ object Mounting {
 
     import MountConfig._
 
-    type M[A] = EitherT[F, MountingError, A]
-
     def lookupType(path: APath): OptionT[F, MountType] =
       OptionT(lift(LookupType(path)))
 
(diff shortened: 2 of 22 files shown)


core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -33,6 +33,9 @@ sealed trait Mounting[A]
 
 object Mounting {
 
+  final case class GetDescendants(dir: ADir)
```
mossprescott: I think it's worth commenting somewhere that this is intended to support more efficient lookups, especially since the implementation in this repo is _not_ more efficient.

Also, it's interesting to me that this now makes it possible to enumerate the mounts directly using this algebra, which seems like a good thing to me.

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -78,80 +76,107 @@ object Mounting {
...
+      S0: MountingErr :<: S,
+      S1: PathMismatchErr :<: S
+    ): F[Unit] =
       config match {
         case ViewConfig(query, vars) =>
           D.right.getOption(refineType(loc)) cata (
-            file => mountView(file, query, vars).map(_.right),
-            PathTypeMismatch(loc).left.point[M])
+            file => mountView(file, query, vars),
+            mismatchErr[S].fail(PathTypeMismatch(loc)))
```
mossprescott: I would call this an improvement. In the "some" case, there's no noise, and in the "none" case, it's much more obvious what's being produced.

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -78,80 +76,107 @@ object Mounting {
...
     /** Remount `src` at `dst`, results in an error if there is no mount at
       * `src`.
       */
-    def remount[T](src: Path[Abs,T,Sandboxed], dst: Path[Abs,T,Sandboxed]): M[Unit] =
+    def remount[T](
+      src: Path[Abs,T,Sandboxed],
+      dst: Path[Abs,T,Sandboxed]
+    )(implicit
+      S0: MountingErr :<: S,
+      // TODO: This isn't necessary as we know it can't happen.
```
mossprescott: So, why is it here?

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -78,80 +76,107 @@ object Mounting {
...
-
-    private def bifold[G[_]: Functor, E, A, EE, AA](v: EitherT[G,E,A])(e: E => EE \/ AA, a: A => EE \/ AA): EitherT[G, EE, AA] =
-      EitherT[G,EE,AA](v.run.map(_.fold(e, a)))
-
-    private def toLeft[G[_]: Functor, E1, E2, A](v: EitherT[G, E1, E2 \/ A]): EitherT[G, E1 \/ E2, A] =
-      bifold(v)(_.left.left, _.fold(_.right.left, _.right))
-
-    private def toRight[G[_]: Functor, E1, E2, A](v: EitherT[G, E1 \/ E2, A]): EitherT[G, E1, E2 \/ A] =
-      bifold(v)(_.fold(_.left, _.left.right), _.right.right)
-
```
mossprescott: :tada:

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -78,80 +76,107 @@ object Mounting {
...
+        _       <- unmount(src)
+        mod     =  mount(dst, f(cfg))
+        restore =  mount(src, cfg)
+        res1    =  mntErr.onFail(mod, err => restore *> mntErr.fail(err))
+        _       <- mmErr.onFail(res1, err => restore *> mmErr.fail(err))
+      } yield ()
     }
+
+    private def mountErr[S[_]](implicit S: MountingErr :<: S) =
+      Failure.Ops[MountingError, S]
```
mossprescott: Should these be defined along with the Error and Failure type alias? I.e. `MountingErr.Ops`.

core/src/main/scala/quasar/fs/mount/package.scala
```diff
@@ -16,21 +16,21 @@
...
 import monocle.Lens
-import scalaz.{Lens => _, _}
+import scalaz.{Lens => _, Failure => _, _}
 import scalaz.concurrent.Task
 
 package object mount {
   type MntErrT[F[_], A] = EitherT[F, MountingError, A]
 
-  type MountConfigs[A] = KeyValueStore[APath, MountConfig, A]
+  type MountingErr[A]     = Failure[MountingError, A]
```
mossprescott: Not a fan of this naming convention. `ErrT` conveyed "transformer", but just dropping two letters is just going to be confusing. How about `MountingFailure`?

main/src/main/scala/quasar/main/package.scala
```diff
@@ -253,14 +258,29 @@ package object main {
...
     val attemptMount: ((APath, MountConfig)) => Free[S, Map[APath, String]] = {
       case (path, cfg) =>
-        mounting.mount(path, cfg).run map {
-          case \/-(\/-(_)) => Map.empty
-          case \/-(-\/(e)) => Map(path -> e.shows)
-          case -\/(e)      => Map(path -> e.shows)
+        mounting.mount(path, cfg).foldMap(runErrs).run map {
+          case \/-(_)      => Map.empty
+          case -\/(-\/(e)) => Map(path -> e.shows)
+          case -\/(\/-(e)) => Map(path -> e.shows)
```
mossprescott: Best diff ever.

mossprescott:
So, I would say on the whole the switch to `Failure` seems to complicate the interpreters somewhat, since they seem to end up interpreting into `EitherT`. But that's a kind of complexity that we're already dealing with, not some new thing.

And the signatures of the constructors are _much_ easier to follow now, since they are expressed in terms of separate effects rather than obliquely named `EitherT` type aliases. So I'm favor of making the switch based on what I see here. I think the effect will be even more pronounced when it comes to the more complex effects (`CompExecM`, I'm looking at you.)

core/src/main/scala/quasar/fs/mount/MountType.scala
```diff
@@ -0,0 +1,63 @@
...
+    Order.order((x, y) => (x, y) match {
+      case (FileSystemMount(t1), FileSystemMount(t2)) => t1 ?|? t2
+      case (ViewMount          , ViewMount          ) => Ordering.EQ
+      case (FileSystemMount(_) , ViewMount          ) => Ordering.GT
+      case (ViewMount          , FileSystemMount(_) ) => Ordering.LT
+    })
+
+  implicit val show: Show[MountType] =
+    Show.shows(_.fold(t => s"FileSystemMount(${t.shows})", "ViewMount"))
+}
```
mossprescott: Wow. This is like a tutorial on how much boilerplate is involved in this style. I count exactly two lines of substantive code, and probably 3 or 4 opportunities for easy to miss bugs.

paulp: One should never have to write code like this! Orders can be composed mechanically. This should be something like (pseudocode) `Order.orderBy[MountType](m => (!m.isView, m.opGettMount))`

core/src/main/scala/quasar/fs/mount/Mounting.scala
wemrysi: Sure, after discussing some of this with @drostron I'm also going to rename it to something like `LookupPrefixed` to convey its intent better, i.e. it is a way of retrieving a (possible) subset of mounts, not a "directory listing" mechanism.

core/src/main/scala/quasar/fs/mount/MountType.scala
wemrysi: Good point, especially since the prisms above provide exactly those operations.

core/src/main/scala/quasar/fs/mount/Mounting.scala
wemrysi: Was a necessary evil at this phase of the PR, it removed once `Move` is added.

wemrysi:
Thanks for the feedback, I think I've addressed all the issues raised thus far.

core/src/test/scala/quasar/fs/mount/MounterSpec.scala
```diff
@@ -45,23 +47,34 @@ class MounterSpec extends MountingSpec[Mounting] {
...
+    val interpEff =
+      mm :+: injectFT[MountingFailure, MEff] :+: injectFT[PathMismatchFailure, MEff]
+
     val interpMnts: MountConfigs ~> Task =
       KeyValueStore.fromTaskRef(cfgRef)
 
-    val interpEff: MEff ~> Task =
-      NaturalTransformation.refl[Task] :+: interpMnts
+    val interpMEff: MEff ~> Task =
+      NaturalTransformation.refl[Task]               :+:
```
drostron: This one missed out on the `reflNT` party.

drostron:
👍🏽

PR #1278, opened 2016-07-22
Work over RenderedTreeSpec.

paulp:
I tend to glaze over looking at large nested case
class constructions so I took a flier at making this
test more visual. Here's a sample from the diff:

```diff
  -    "find simple difference in child" in {
  -      val t1 = NonTerminal(Nil, Some("A"), Terminal(Nil, Some("B")) :: Nil)
  -      val t2 = NonTerminal(Nil, Some("A"), Terminal(Nil, Some("C")) :: Nil)
  -      t1.diff(t2) must_==
  -        NonTerminal(Nil, Some("A"),
  -          Terminal(">>>" :: Nil, Some("B")) ::
  -            Terminal("<<<" :: Nil, Some("C")) ::
  -            Nil)
  -    }
  +    "find simple difference in child" in expect(
  +      'A /\ 'B,
  +      'A /\ 'C,
  +      'A /\ (-'B, +'C)
  +    )
```


foundation/src/test/scala/quasar/RenderTree.scala
```diff
@@ -19,145 +19,135 @@ package quasar
...
-      val t2 = NonTerminal(Nil, Some("C"), Terminal(Nil, Some("B")) :: Nil)
-      t1.diff(t2) must_==
-        NonTerminal("[Root differs]" :: Nil, None,
-          NonTerminal(List(">>>"), Some("A"), Terminal(Nil, Some("B")) :: Nil) ::
-            NonTerminal(List("<<<"), Some("C"), Terminal(Nil, Some("B")) :: Nil) ::
-            Nil)
-    }
+class RenderedTreeSpec extends quasar.QuasarSpecification with ScalaCheck {
+  private def expectJson(t: RenderedTree, expect: Json)                        = t.asJson must_scalaz_== expect
+  private def expect(t1: RenderedTree, t2: RenderedTree, result: RenderedTree) = (t1 diff t2).asJson must_scalaz_== result.asJson
```
jedesah: Is there a good reason not to define an instance of `Equal` for `RenderedTree` here?

foundation/src/test/scala/quasar/RenderTree.scala
```diff
@@ -19,145 +19,135 @@ package quasar
...
 
-    "find added child" in {
-      val t1 = NonTerminal(Nil, Some("A"), Terminal(Nil, Some("B")) :: Nil)
-      val t2 = NonTerminal(Nil, Some("A"), Terminal(Nil, Some("B")) :: Terminal(Nil, Some("C")) :: Nil)
-      t1.diff(t2) must_== NonTerminal(Nil, Some("A"), Terminal(Nil, Some("B")) :: Terminal("<<<" :: Nil, Some("C")) :: Nil)
-    }
+  private implicit class TreeSymbolOps(sym: scala.Symbol) {
+    def unary_- = RenderedTree(List(">>>"), Some(sym.name), Nil)
```
jedesah: This syntax is new to me, but I like it.

foundation/src/test/scala/quasar/RenderTree.scala
```diff
@@ -19,145 +19,135 @@ package quasar
...
-      val t1 = NonTerminal(Nil, Some("A"), Terminal(Nil, Some("B")) :: Terminal(Nil, Some("C")) :: Terminal(Nil, Some("D")) :: Nil)
-      val t2 = NonTerminal(Nil, Some("A"), Terminal(Nil, Some("C")) :: Terminal(Nil, Some("C1")) :: Terminal(Nil, Some("D")) :: Nil)
-      t1.diff(t2) must_==
-        NonTerminal(Nil, Some("A"),
-          Terminal(">>>" :: Nil, Some("B")) ::
-            Terminal(Nil, Some("C")) ::
-            Terminal("<<<" :: Nil, Some("C1")) ::
-            Terminal(Nil, Some("D")) :: Nil)
-    }
+    def :+(child: RenderedTree): RenderedTree  = t.copy(children = t.children :+ child)
```
jedesah: I think these two methods seem useful enough that they could be defined outside of this test file, but I am okay with them being here until usage arises elsewhere.

foundation/src/test/scala/quasar/RenderTree.scala
paulp: Only that I don't know exactly what's intended for it and adding an Equal instance is a stronger statement.

jedesah: I think it's probably safe to add the instance (globally). That being said if we feel like it should not be generally available, it would still make sense to define it locally here.

jedesah:
This makes the tests much more concise and easier to understand at a glance!

jedesah:
@paulp I marked this as ready, so feel free to make the minor tweak I suggested or not before merging.

PR #1279, opened 2016-07-22
Route all test specifications through QuasarSpecification.

paulp:
Then ugrade the musts for the specs in project core.

### DIFF
--- a/core/src/main/scala/quasar/fs/FileSystemError.scala
+++ b/core/src/main/scala/quasar/fs/FileSystemError.scala
@@ -93,6 +93,8 @@ object FileSystemError {
     case WriteFailed(d, r) => (d, r)
   } (WriteFailed.tupled)
 
+  implicit val fileSystemErrorEqual: Equal[FileSystemError] = Equal.equalA
+
   implicit def fileSystemErrorShow: Show[FileSystemError] =
     Show.shows {
       case ExecutionFailed(_, rsn, _, c) =>
--- a/core/src/test/scala/quasar/codec.scala
+++ b/core/src/test/scala/quasar/codec.scala
@@ -20,14 +20,13 @@ import argonaut._, Argonaut._
 import quasar.DataEncodingError.{UnrepresentableDataError, UnescapedKeyError}
 import quasar.Predef._
 
-import org.specs2.mutable._
 import org.specs2.scalaz._
 import org.specs2.ScalaCheck
 
 import org.threeten.bp._
-import scalaz._
+import scalaz._, Scalaz._
 
-class DataCodecSpecs extends Specification with ScalaCheck with DisjunctionMatchers {
+class DataCodecSpecs extends quasar.QuasarSpecification with ScalaCheck with DisjunctionMatchers {
   import DataArbitrary._
 
   implicit val DataShow = new Show[Data] { override def show(v: Data) = v.toString }
@@ -211,11 +210,11 @@ class DataCodecSpecs extends Specification with ScalaCheck with DisjunctionMatch
 
   "Error messages" should {
     "UnrepresentableDataError" ! prop { any: Data =>
-      UnrepresentableDataError(any).message must_== ("not representable: " + any)
+      UnrepresentableDataError(any).message must_= ("not representable: " + any)
     }
     "UnescapedKeyError" in {
       val sample:Json = jString("foo")
-      UnescapedKeyError(sample).message must_== s"un-escaped key: $sample"
+      UnescapedKeyError(sample).message must_= s"un-escaped key: $sample"
     }
   }
 }
--- a/core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
+++ b/core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
@@ -25,11 +25,9 @@ import quasar.effect._
 import monocle.function.Field1
 import monocle.std.tuple2._
 import pathy.Path._
-import org.specs2.mutable
-import scalaz.{Failure => _, _}
-import scalaz.syntax.applicative._
+import scalaz.{Failure => _, _}, Scalaz._
 
-class FileSystemMounterSpec extends mutable.Specification {
+class FileSystemMounterSpec extends quasar.QuasarSpecification {
   import FileSystemDef._
 
   type Abort[A]  = Failure[String, A]
@@ -137,7 +135,7 @@ class FileSystemMounterSpec extends mutable.Specification {
         val cln = "CLEAN"
         val (rmnts, signal) = eval(Mounts.singleton(d, fsResult(cln)))(mount(d))
 
-        (rmnts.toMap.isEmpty must beFalse) and (signal must_== \/.left(cln))
+        (rmnts.toMap.isEmpty must beFalse) and (signal must_=== \/.left(cln))
       }
     }
 
@@ -146,7 +144,7 @@ class FileSystemMounterSpec extends mutable.Specification {
         val d = rootDir </> dir("unmount") </> dir("cleanup")
         val undo = "UNDO"
 
-        eval(Mounts.singleton(d, fsResult(undo)))(unmount(d))._2 must_== \/.left(undo)
+        eval(Mounts.singleton(d, fsResult(undo)))(unmount(d))._2 must_= \/.left(undo)
       }
 
       "deletes the entry from mounts" >> {
--- a/foundation/src/test/scala/quasar/RenderTree.scala
+++ b/foundation/src/test/scala/quasar/RenderTree.scala
@@ -19,9 +19,7 @@ package quasar
 import quasar.Predef._
 
 import argonaut._, Argonaut._
-import org.specs2.mutable._
-
-class RenderedTreeSpec extends Specification {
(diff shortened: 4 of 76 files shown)


core/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -93,6 +93,8 @@ object FileSystemError {
     case WriteFailed(d, r) => (d, r)
   } (WriteFailed.tupled)
 
+  implicit val fileSystemErrorEqual: Equal[FileSystemError] = Equal.equal(_ eq _)
```
jedesah: `Equal.equalA` would be more idiomatic "deferral to universal equality".

core/src/test/scala/quasar/codec.scala
```diff
@@ -20,14 +20,13 @@ import argonaut._, Argonaut._
 import quasar.DataEncodingError.{UnrepresentableDataError, UnescapedKeyError}
 import quasar.Predef._
 
-import org.specs2.mutable._
 import org.specs2.scalaz._
 import org.specs2.ScalaCheck
 
 import org.threeten.bp._
-import scalaz._
+import scalaz._, Scalaz._
```
jedesah: Why is `Scalaz._` suddenly needed?

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
```diff
@@ -25,11 +25,9 @@ import quasar.effect._
 import monocle.function.Field1
 import monocle.std.tuple2._
 import pathy.Path._
-import org.specs2.mutable
-import scalaz.{Failure => _, _}
-import scalaz.syntax.applicative._
+import scalaz.{Failure => _, _}, Scalaz._
```
jedesah: I am curious why you prefer `Scalaz._` over the explicit import when there is just one.

Don't get me wrong, I am lazy and I probably would have authored `Scalaz._` myself, but since someone already went through the trouble of importing only that specific import...

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
```diff
@@ -146,7 +144,7 @@ class FileSystemMounterSpec extends mutable.Specification {
         val d = rootDir </> dir("unmount") </> dir("cleanup")
         val undo = "UNDO"
 
-        eval(Mounts.singleton(d, fsResult(undo)))(unmount(d))._2 must_== \/.left(undo)
+        eval(Mounts.singleton(d, fsResult(undo)))(unmount(d))._2 must_scalaz_== \/.left(undo)
```
jedesah: Why `must_scalaz_==` here but `must_===` above?

foundation/src/test/scala/quasar/RenderTree.scala
```diff
@@ -19,9 +19,7 @@ package quasar
 import quasar.Predef._
 
 import argonaut._, Argonaut._
-import org.specs2.mutable._
-
-class RenderedTreeSpec extends Specification {
+class RenderedTreeSpec extends quasar.QuasarSpecification {
```
jedesah: I think you probably did not mean to erase the newline here.

jedesah:
@paulp Thx for doing this. Once again, a few minor comments which you can change or we can discuss or wtv, but nothing major so marking as "ready".

I assume the build failure is probably some failed `Gen` as usual... 😞 

One side comment, not sure how I feel about `must_scalaz_==`. I think it's a little long and ugly for something we want to start using as much as possible. It might make sense to setup `QuasarSpecification` in such a way that `must_==` is the `Equal` based matcher within that trait.

core/src/main/scala/quasar/fs/FileSystemError.scala
paulp: It's deferring to reference equality to be conservative. There was (surprisingly) no method in scalaz for it.

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
paulp: I don't see any upside to import babysitting as the contents of a file evolve. Maybe if I had tooling which automatically adjusted it to the minimum, but even then I wouldn't be super excited about it. I think if `import Scalaz._` creates a problem (e.g. over a naming or implicit conflict) that should be caught right away so as to keep that bundle working smoothly.

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
paulp: It's (intended to be) `must_scalaz_==` everywhere which it works, and `must_===` everywhere which it doesn't. For the scalaz comparison to work there have to be equals instances for the turtles all the way down, which in many cases there aren't and it'll be a while before they're there. In the meantime it's still much better to require comparisons to be between values of the same type.

paulp:
`must_scalaz_==` is horrible, I actually mean that as a placeholder hoping for a better suggestion. But I'm not sure what's better without being unclear as to what it is. The `_==...` namespace is pretty heavily trafficked.

core/src/test/scala/quasar/codec.scala
paulp: `must_scalaz_==` has to be able to find all the Equal instances.

wemrysi:
`must_really_==`? :wink:

core/src/main/scala/quasar/fs/FileSystemError.scala
jedesah: Hmm... Not sure about reference equality. I don't think we use anywhere else. How is reference equality conservative?

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
jedesah: I think @wemrysi mentioned once that he had reason to believe that avoiding `Scalaz._` improved compilation speed which is somewhat of an issue for us. Once again, I almost never "babysit" imports, but if someone already did that for me, I would be inclined to leave it alone unless I need something else from `scalaz`.

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
jedesah: Right. I got that part. But aren't lines 138 and 147 operating over the same types? I figured it's available for line 147, it should be possible to use `must_scalaz_==` on line 138 as well.

core/src/main/scala/quasar/fs/FileSystemError.scala
paulp: In that reference equality is the lower bound for universal equality, so it can only be too tight, not too generous. Which means if a test is comparing two instances, it might fail when it should pass, but it won't pass when it should fail. (This assumes it's far more likely a test will hinge on an equivalence than an inequivalence.)

core/src/test/scala/quasar/codec.scala
jedesah: Ah, that makes sense.

jedesah:
I would just use `must_==`. It has the added benefit of making the original `specs2` `must_==` really hard to use. Which is probably how it should be.

paulp:
I am leery of repurposing `must_==` to mean something different than specs does when using specs. It would mislead every person who looks at the code until they discovered what's going on.

paulp:
I'd like to make the specs2 `must_==` hard (actually impossible) to use though, yes.

core/src/main/scala/quasar/fs/FileSystemError.scala
jedesah: Okay, sure, but I think it would be okay to rely on universal equality here. If we feel like that's not good enough, then I would suggest we write out the `Equal` instance explicitly.

I am worried that reference equality will be somewhat of a surprise in a codebase that largely ignores the distinction. I, for one, did not really remember off the top of my head that `eq` was reference equality as opposed to `==` and I could easily see myself bang my head for a long time trying to diagnose a test failure caused by it.

paulp:
I rebased it and renamed the operator `must_=` at @jedesah's suggestion.

core/src/test/scala/quasar/fs/mount/FileSystemMounterSpec.scala
paulp: It's quite possible there's a compilation speed impact, but measurement is required. With scala it's easy to fall into superstition when it comes to what slows things down and what doesn't. If the Scalaz bundle brings in too many implicits, I'd still much prefer to create a single bundle which is used as a standard import - I'd hand roll the utilized subset of the Scalaz object.

paulp:
I'll document the various must_ operators and the idea that all specs extend QuasarSpecification, and look into compilation speed as well - specs uses lots of implicits so actually inheriting from a specs trait might contribute as much to compilation time as importing from the Scalaz object. Any thoughts where such documentation should land in order to be likely to be seen by test writers?

jedesah:
@paulp The GitHub Wiki is the best place I can think of. It has a few articles on "development practices" so I think it would be somewhat appropriate for a new article to talk about our "testing style".

PR #1281, opened 2016-07-25
Some optimizations related to LeftShift.

sellout:
Also some better test expectations.

### DIFF
--- a/core/src/main/scala/quasar/qscript/QScript.scala
+++ b/core/src/main/scala/quasar/qscript/QScript.scala
@@ -798,6 +798,33 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
     case _ => None
   }
 
+  def coalesceMapShift[F[_]: Functor](implicit SP: SourcedPathable[T, ?] :<: F, QC: QScriptCore[T, ?] :<: F):
+      QScriptCore[T, T[F]] => F[T[F]] = {
+    case x @ Map(Embed(src), mf) => (SP.prj(src) >>= {
+      case LeftShift(srcInner, struct, repair) =>
+        SP.inj(LeftShift(srcInner, struct, mf >> repair)).some
+      case _ => None
+    }).getOrElse(QC.inj(x))
+    case x => QC.inj(x)
+  }
+
+  def simplifySP[F[_]: Functor](
+    implicit SP: SourcedPathable[T, ?] :<: F, QC: QScriptCore[T, ?] :<: F):
+      SourcedPathable[T, T[F]] => F[T[F]] = {
+    case x @ LeftShift(src, struct, repair) =>
+      if (!repair.element(RightSide))
+        QC.inj(Map(src, repair ∘ κ(SrcHole)))
+      else if (!repair.element(LeftSide))
+        (QC.prj(src.project) >>= {
+          case Map(innerSrc, mf) =>
+            SP.inj(LeftShift(innerSrc, struct >> mf, repair)).some
+          case _ => None
+        }).getOrElse(SP.inj(x))
+      else
+        SP.inj(x)
+    case x => SP.inj(x)
+  }
+
   def coalesceQCCo[F[_]: Functor, A](implicit QC: QScriptCore[T, ?] :<: F, FI: F :<: QScriptProject[T, ?]):
       QScriptCore[T, T[CoEnv[A, F, ?]]] => Option[QScriptCore[T, T[CoEnv[A, F, ?]]]] = {
     case Map(Embed(src), mf) =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -798,6 +798,33 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
     case _ => None
   }
 
+  def coalesceMapShift[F[_]: Functor](implicit SP: SourcedPathable[T, ?] :<: F, QC: QScriptCore[T, ?] :<: F):
+      QScriptCore[T, T[F]] => F[T[F]] = {
+    case x @ Map(Embed(src), mf) => (SP.prj(src) >>= {
```
alissapajer: @sellout I think it would make sense to rename `LeftSide` and `RightSide` in this case to something like `ProvenanceAccess` and `ValueAccess`. `LeftSide` and `RightSide` make sense in the case of a `ThetaJoin` because there really are two sides. But here the notion of two "sides" isn't really accurate here since we don't have a join.

Also, as I think about it, why are we even giving this provenance information to the backend? Shouldn't a backend never care about `ProvenanceAccess` at all?

I think in the case of `LeftShift` (and analogously for `Reduce`) we can have a `LeftShiftInternal` that carries the provenance, and then that gets optimized to a `LeftShift` that has no provenance information, and this `LeftShift` is given to a backend. Basically, is there any value in giving a backend a `repair` function with two holes, when ultimately a backend should only need access one of the holes?

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -798,6 +798,33 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
...
+      case _ => None
+    }).getOrElse(QC.inj(x))
+    case x => QC.inj(x)
+  }
+
+  def simplifySP[F[_]: Functor](
+    implicit SP: SourcedPathable[T, ?] :<: F, QC: QScriptCore[T, ?] :<: F):
+      SourcedPathable[T, T[F]] => F[T[F]] = {
+    case x @ LeftShift(src, struct, repair) =>
+      if (!repair.element(RightSide))
```
alissapajer: @sellout Do we expect to ever hit the case when there are no `LeftSide` (`ProvenanceAccess`) at any of the leaves?

core/src/main/scala/quasar/qscript/QScript.scala
sellout: So, the distinction isn’t quite so clear – both the provenance of the src and the un-flattened value is stored in `LeftSide` and both the value and the provenance from the `LeftShift` itself are accessed via `RightSide`.

Having different names for the two leaves makes sense. However, I don’t know if it makes sense to have different types for the join case and the leftshift case (and the union case, etc.). Depends what is easier for the backends. In the MongoDB backend, we handle all `JoinSide` funcs the same, so it’s just more work if there are different types. For other backends, they might have to behave differently depending on the context of the `JoinFunc`, so different types wouldn’t be a problem.

Provenance always needs to be carried through everything, because later steps may need to reify it for bucketing. But again, in the specific case of `LeftShift`, `LeftSide` also has the unflattened values that may be used. (e.g., if `src` evaluated to a set with values that look like `{"foo": 3, "bar": [0, 1 ,2]}` and a `struct` that looks like `ProjectField(SrcHole, StrLit("bar"))`, the repair function may use `LeftSide` in order to build a result that looks like `{"foo": 3, "bar": 0}, {"foo": 3, "bar": 1}, …`).

If provenance isn’t used by a later step (e.g., if LeftShift is the last operation), then it’ll get cleaned up in optimization.

core/src/main/scala/quasar/qscript/QScript.scala
sellout: Yeah – two of the test cases below have no reference to `LeftSide`.

But, maybe you mean `RightSide`, since that’s the line this comment is on. I don’t know if that’s something that’s likely to be hit. It would be an oddly written (but not impossible) query. It’s the kind of thing that should maybe trigger a warning (if we had such things).

alissapajer: Yes I meant `RightSide` :)

core/src/main/scala/quasar/qscript/QScript.scala
alissapajer: Ok, I think a solid approach is removing unused elements during optimization. Then we don't have to reason (like I was trying to do) about all the myriad cases when we'll be using different elements in different situations. If it's unused, we'll just remove it, and that allows us to handle a case like this more generally, not *assuming* anything will ever be unused.

alissapajer: Also, we can hold off on any renaming until we have a few backend impls to compare. And if this is really confusing for backends, then we'll address that when it comes up.

alissapajer:
@sellout ok merge away!

alissapajer:
See #1292

Issue #1254, opened 2016-07-15
SparkCore Connector: WriteFile interpreter


rabbitonweb:
- [x] implementation for local-spark 
- [x] test with combination of local-read following: 
``` 
       wr.open -> 
       wr.write chunk ->  
       wr.close ->  
       rd.open ->  
      rd.read ->  
      rd.close ->  
      assert that all written was then read
```
- [x] test altogether with `ReadFile` and `ManageFile`

PR #1282, opened 2016-07-25
WriteFile for spark-local

rabbitonweb:
Implements WriteFile for spark-local. 
Connected with https://github.com/quasar-analytics/quasar/issues/1254

### DIFF
--- a/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
+++ b/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
@@ -90,7 +90,7 @@ class ReadFileSpec extends Specification with ScalaCheck  {
     
 
   private def inter(implicit sc: SparkContext): ReadFile ~> Task =
-    readfile.interpret[Eff](local_readfile.input[Eff]) andThen foldMapNT[Eff, Task](run)
+    readfile.interpret[Eff](local.readfile.input[Eff]) andThen foldMapNT[Eff, Task](run)
 
   private def newSc(): SparkContext = {
     val config = new SparkConf().setMaster("local[*]").setAppName(this.getClass().getName())
--- a/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
+++ b/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.physical.sparkcore.fs
+
+import quasar.Predef._
+import quasar.QuasarSpecification
+import quasar.Data
+import quasar.fp.TaskRef
+import quasar.fp.numeric._
+import quasar.fp.free._
+import quasar.fs._
+import quasar.fs.WriteFile.WriteHandle
+import quasar.fs.FileSystemError._
+import quasar.effect._
+
+import java.lang.System
+import java.io._
+import java.nio.file._
+
+import org.specs2.ScalaCheck
+import pathy.Path.posixCodec
+import pathy.Path._
+import scalaz._, Scalaz._, concurrent.Task
+
+class WriteFileSpec extends QuasarSpecification with ScalaCheck  {
+
+  type Eff0[A] = Coproduct[KeyValueStore[WriteHandle, PrintWriter, ?], Task, A]
+  type Eff[A] = Coproduct[MonotonicSeq, Eff0, A]
+
+  "writefile" should {
+
+
+    "open -> write chunks -> close -> assert file with results" in {
+      // given
+      val chunks = (user("john", 31) :: user("anna", 21) :: Nil).toVector
+      val program = (path: AFile) => define { writeUnsafe =>
+        for {
+          wh <- writeUnsafe.open(path)
+          errors <- writeUnsafe.write(wh, chunks).liftM[FileSystemErrT]
+          _ <- writeUnsafe.close(wh).liftM[FileSystemErrT]
+        } yield (errors)
+      }
+      // when
+      withTempFile { path =>
+        for {
+          result <- execute[Vector[FileSystemError]](program(path))
+          content <- getContent(path)
+        } yield{
+          // then
+          result must_= \/-(Vector.empty[FileSystemError])
+          content  must_= List(
+            """{ "login": "john", "age": 31 }""",
+            """{ "login": "anna", "age": 21 }"""
+          )
+        }
+      }
+    }
+
+    "(open -> write chunks -> close) -> (open -> write chunks -> close) -> assert file with results" in {
+      // given
+      val chunks1 = (user("john", 31) :: user("anna", 21) :: Nil).toVector
+      val chunks2 = (user("kate", 22) :: Nil).toVector
+
+      val program = (path: AFile) => define { writeUnsafe =>
+        for {
+          wh1 <- writeUnsafe.open(path)
+          errors1 <- writeUnsafe.write(wh1, chunks1).liftM[FileSystemErrT]
+          _ <- writeUnsafe.close(wh1).liftM[FileSystemErrT]
+          wh2 <- writeUnsafe.open(path)
+          errors2 <- writeUnsafe.write(wh2, chunks2).liftM[FileSystemErrT]
+          _ <- writeUnsafe.close(wh2).liftM[FileSystemErrT]
+
+        } yield (errors1 ++ errors2)
+      }
+      // when
+      withTempFile { path =>
+        for {
+          result <- execute[Vector[FileSystemError]](program(path))
+          content <- getContent(path)
+        } yield {
+          // then
+          result must_= \/-(Vector.empty[FileSystemError])
+          content must_= List(
+            """{ "login": "john", "age": 31 }""",
+            """{ "login": "anna", "age": 21 }""",
+            """{ "login": "kate", "age": 22 }"""
+          )
+        }
+      }
+    }
+  }
+
+  private def execute[C](program: FileSystemErrT[Free[WriteFile, ?], C]):
+      Task[FileSystemError \/ C] = interpreter.flatMap(program.run.foldMap(_))
+ 
+  private def interpreter: Task[WriteFile ~> Task] = {
+
+    def innerInterpreter: Task[Eff ~> Task] =  {
+      (TaskRef(0L) |@| TaskRef(Map.empty[WriteHandle, PrintWriter])) { (genState, kvsState) =>
+        MonotonicSeq.fromTaskRef(genState) :+:
+        KeyValueStore.fromTaskRef[WriteHandle, PrintWriter](kvsState) :+:
+        NaturalTransformation.refl[Task]
+      }
+    }
+
+    innerInterpreter.map { inner =>
+      local.writefile.interpret[Eff] andThen foldMapNT[Eff, Task](inner)
+    }
+  }
+
+  private def withTempFile[C](run: AFile => Task[C]): C = {
+
+    def genTempFilePath: Task[AFile] = Task.delay {
+      val path = System.getProperty("java.io.tmpdir") +
+      "/" + scala.util.Random.nextInt().toString + ".tmp"
+      sandboxAbs(posixCodec.parseAbsFile(path).get)
+    }
+
+    def deleteFile(file: AFile): Task[Unit] = Task.delay {
+      Files.delete(Paths.get(posixCodec.unsafePrintPath(file)))
+    }
+
+    val execution: Task[C] = for {
+      filePath <- genTempFilePath
+      result <- run(filePath).onFinish {
+        _ => deleteFile(filePath)
+      }
+    } yield result
+
+    execution.unsafePerformSync
+  }
+
+  private def define[C]
+    (defined: WriteFile.Unsafe[WriteFile] => FileSystemErrT[Free[WriteFile, ?], C])
+    (implicit writeUnsafe: WriteFile.Unsafe[WriteFile])
+      : FileSystemErrT[Free[WriteFile, ?], C] =
+    defined(writeUnsafe)
+  
+  private def user(login: String, age: Int) =
+    Data.Obj(ListMap("login" -> Data.Str(login), "age" -> Data.Int(age)))
+
+  private def getContent(f: AFile): Task[List[String]] =
+    Task.delay(scala.io.Source.fromFile(posixCodec.unsafePrintPath(f)).getLines.toList)
+
+}
--- a/CHANGELOG/feature.md
+++ b/CHANGELOG/feature.md
@@ -1,0 +1,1 @@
+- Add WriteFile implementation for spark-local
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/readfile.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/readfile.scala
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
-package quasar.physical.sparkcore.fs
+package quasar.physical.sparkcore.fs.local
 
 import quasar.Predef._
 import quasar.fs._
 import quasar.fp.free._
 import quasar.effect.Read
+import quasar.physical.sparkcore.fs.Input
 
 import java.io.File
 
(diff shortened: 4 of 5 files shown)


sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_writefile.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import java.io.{File, PrintWriter, FileOutputStream}
+// import scala.util.Try
+
+import pathy.Path.posixCodec
+import scalaz._
+import Scalaz._
+import scalaz.concurrent.Task
+
+object local_writefile {
```
drostron: Same as https://github.com/quasar-analytics/quasar/pull/1242/files#r72072145.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_writefile.scala
```diff
@@ -0,0 +1,116 @@
...
+      val tasks: Vector[Task[Vector[FileSystemError]]] = (maybeLines.map {
+        case (-\/(error), data) => Task.now(Vector(writeFailed(data, error.message)))
+        case (\/-(line), data) => Task.now {
+          \/.fromTryCatchNonFatal(pw.append(s"$line\n")).fold(
+            ex => Vector(writeFailed(data, ex.getMessage)),
+            u => Vector.empty[FileSystemError]
+          )
+        }
+      })
+      tasks.sequence.map(_.flatten)
```
drostron: Could eliminate the need for `sequence.map` here by wrapping the whole thing in `Task.delay` since it's getting transformed to that anyway.
```scala
Task.delay(maybeLines.flatMap {
  case (-\/(error), data) => Vector(writeFailed(data, error.message))
  case (\/-(line), data) =>
    \/.fromTryCatchNonFatal(pw.append(s"$line\n")).fold(
      ex => Vector(writeFailed(data, ex.getMessage)),
      u => Vector.empty[FileSystemError]
    )
})
```

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_writefile.scala
```diff
@@ -0,0 +1,116 @@
...
+import quasar.{Data, DataCodec, DataEncodingError}
+import quasar.fs._
+import quasar.fs.PathError._
+import quasar.fs.FileSystemError._
+import quasar.fs.WriteFile._
+import quasar.fp.free._
+import quasar.effect._
+
+import java.io.{File, PrintWriter, FileOutputStream}
+// import scala.util.Try
```
drostron: Stray import

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local_writefile.scala
rabbitonweb: @drostron @wemrysi 

So you guys are saying instead of having:
```
quasar/physical/sparkcore/fs/local_read.scala
quasar/physical/sparkcore/fs/local_write.scala
quasar/physical/sparkcore/fs/local_manage.scala
```

I should do

```
quasar/physical/sparkcore/fs/write/local.scala
quasar/physical/sparkcore/fs/read/local.scala
quasar/physical/sparkcore/fs/manage/local.scala
```

drostron: That could work. Or if you wanted to group them by local:
```
quasar/physical/sparkcore/fs/local/readfile.scala
quasar/physical/sparkcore/fs/local/writefile.scala
quasar/physical/sparkcore/fs/local/managefile.scala
```

jedesah: I have a preference for what @drostron suggested.

rabbitonweb: Done, PR back in "review"

drostron:
Changelog needed.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
```diff
@@ -25,6 +25,7 @@ import quasar.fp.free._
 import quasar.fs._
 import quasar.fs.ReadFile.ReadHandle
 import quasar.effect._
+// import quasar.physical.sparkcore.fs.local.readfile
```
wemrysi: Spurious comment?

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+      )
+      ok
+    }
+  }
+
+  private def interpreter: WriteFile ~> Task = {
+
+    def innerInterpreter: Eff ~> Task =  {
+
+      val genState = TaskRef(0L).unsafePerformSync
```
wemrysi: Rather than running these tasks here, you should return the interpreter in `Task`, i.e.
```scala
(TaskRef(0L) |@| TaskRef(Map.empty[WriteHandle, PrintWriter]))((genState, kvsState) => ...)
```

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+
+class WriteFileSpec extends Specification with ScalaCheck  {
+
+  type Eff0[A] = Coproduct[KeyValueStore[WriteHandle, PrintWriter, ?], Task, A]
+  type Eff[A] = Coproduct[MonotonicSeq, Eff0, A]
+
+  "writefile" should {
+    "open -> write chunks -> close -> assert file with results" in {
+      // given
+      val path: AFile = genTempPath
```
wemrysi: If a file at path is created by the test I think it should also be cleaned up (deleted) by the test.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+      }
+
+      val result = program.run.foldMap(interpreter).unsafePerformSync
+      // when
+      result must_== \/-(Vector.empty[FileSystemError])
+      getContent(path) must_== List(
+        """{ "login": "john", "age": 31 }""",
+        """{ "login": "anna", "age": 21 }"""
+      )
+      ok
```
wemrysi: I don't think the trailing `ok` is necessary, is it?

paulp: Trailing ok should not be necessary. Also, though I haven't made the repo-wide push for it yet, we should really lose the habit of using `must_==`. It should always be `must_=` if possible, or `must_===` where not. The former requires an implicit `Equal` instance, while the latter at least ensures the operands have the same type.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+    local.writefile.interpret[Eff] andThen foldMapNT[Eff, Task](innerInterpreter)
+  }
+
+  private def define
+    (defined: WriteFile.Unsafe[WriteFile] => FileSystemErrT[Free[WriteFile, ?], Vector[FileSystemError]])
+    (implicit writeUnsafe: WriteFile.Unsafe[WriteFile])
+      : FileSystemErrT[Free[WriteFile, ?], Vector[FileSystemError]] =
+    defined(writeUnsafe)
+
+  private def genTempPath: AFile = {
```
wemrysi: Should return `Task[AFile]` as it reads from the environment and uses `Random`.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+    val path = System.getProperty("java.io.tmpdir") + "/" + scala.util.Random.nextInt().toString + ".tmp"
+    sandboxAbs(posixCodec.parseAbsFile(path).get)
+  }
+
+  private def user(login: String, age: Int) =
+    Data.Obj(ListMap("login" -> Data.Str(login), "age" -> Data.Int(age)))
+
+  private def getContent(f: AFile): List[String] =
```
wemrysi: Should return `Task[List[String]]` due to IO.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
jedesah: I claim responsibility for suggesting to do it this way as it made the code simpler and it's in a test file.

I have been meaning to create a convenience function that gives you a `KeyValueStore` interpreter inside of a Task with the empty state so that we don't constantly have to go through this little dance of first creating a `TaskRef` with an empty `Map` and then passing that to the function that creates a KVS interpreter. It's come up in a few places now already.

I kind of got hooked up though because I tried to make that interpreter use `scala.concurrent.collection.TrieMap` to implement it and bypass the `TaskRef` entirely, but was unsuccessful in making that new interpreter pass the tests. Maybe I will just submit the `TaskRef` convenience function first and then eventually make it use the concurrent Map instead.

wemrysi: I disagree that it makes anything simpler, there is no reduction in complexity it is only less explicit as possible thrown exceptions and side-effects still exist, they just aren't tracked in the type.

And while we allow for some more "lenient" practices in tests, I don't think it is license to ignore doing prudent things like keeping as many side-effects in `Task` as possible, for all the normal reasons to use `Task` in the first place. In this case, it is also useful to ensure the files created by the test are cleaned up in the presence of spurious failure as we will have `Task#onFinish` available.

jedesah: To be clear, I did not suggest not to use `Task`, I simply said it was probably okay to call `unsafePerformSync` inside of the function that creates the interpreter.

Like I said, the ideal solution is to provide a new, more convenient, way to create a KVS interpreter (inside of `Task`).

rabbitonweb:
@wemrysi I've changed both methods `interepreter` & `getContent` to return `Task` - can you check if that's what you were hoping to see

wemrysi:
@rabbitonweb Aye, that's better, but `genTempPath` should also be in `Task` and the files that are created need to be deleted. Maybe `execute` should supply the file path to the test and then delete it if it exists afterwards, i.e. `def execute(f: AFile => Free[Eff, A]): A = ...`.

rabbitonweb:
@wemrysi why should `genTempPath` return a Task? This is not really side-effecting function  - it returns only path, never creates a file directly

wemrysi:
@rabbitonweb Reading system properties and random values is not referentially transparent, so it should either be replaced with something that is or be made so by wrapping in `Task` (in this case, the latter is fine).

rabbitonweb:
Back in review. 
There is one feature missing (as it turned out on Friday last week) - creation of non-existing dirs. However I will provide that as a separate PR.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -42,51 +43,54 @@ class WriteFileSpec extends QuasarSpecification with ScalaCheck  {
...
         } yield (errors)
       }
-
-      val result = execute(program).unsafePerformSync
       // when
+      val (result, content) = (for {
+        path <- genTempPath
+        result <- execute[Vector[FileSystemError]](program(path))
+        content <- getContent(path)
+        _ <- deleteFile(path)
```
drostron: I like seeing temporary files being cleaned up. What happens if `execute` or `getContent` fails?

rabbitonweb: Should I put `deleteFile` as a separate call than? That's the only valid solution I can think of, though it makes a bit of noise in the code. Something like:

```
val (result, content) = (for {
        path <- genTempPath
        result <- execute[Vector[FileSystemError]](program(path))
        content <- getContent(path)
        _ <- deleteFile(path)
} yield(result, content)).unsafePerformSync
       // then
       result must_= \/-(Vector.empty[FileSystemError])
       content  must_= List(
         """{ "login": "john", "age": 31 }""",
         """{ "login": "anna", "age": 21 }"""
      )

deleteFiles(path)
```

rabbitonweb: Well, it does not make sense. According to  http://timperrett.com/2014/07/20/scalaz-task-the-missing-documentation/ if Task fails, when I call `unsafePerformSync` it will explode with exception. I have to check Spec2 some clever way to act on when test is done (regardless of the outcome)

paulp: When you need something to exist for a fixed time frame it's usually best to encapsulate that time frame, like `def withTempFile[A](f: AFile => A): A = ...` which produces the temporary file, lets you operate on it in function `f`, and then deletes it after the function returns.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  private def define[C]
+    (defined: WriteFile.Unsafe[WriteFile] => FileSystemErrT[Free[WriteFile, ?], C])
+    (implicit writeUnsafe: WriteFile.Unsafe[WriteFile])
+      : FileSystemErrT[Free[WriteFile, ?], C] =
+    defined(writeUnsafe)
+
+  private def genTempPath: Task[AFile] = Task.delay {
+    val path = System.getProperty("java.io.tmpdir") + "/" + scala.util.Random.nextInt().toString + ".tmp"
```
paulp: This should be something more like
```
java.nio.file.Files.createTempFile("quasar.", ".tmp")
```

rabbitonweb: Not really, since I don't want to "create" a file here, gust generate path to it. Maybe function name should be different.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/WriteFileSpec.scala
rabbitonweb: That was my original thinking (same way as I've created `define` function). 
However I fear that `withTempFile[A](f: AFile => A): A = ...` is not referentially transparent and next reviewer will point this out... 
I can not have both of two worlds here. Either one or the other approach.

rabbitonweb: Ok, I'm seriously stuck. What I got is:
```
  private def withTempFile[C](run: AFile => Task[C]): C = {
    val filePath = genTempPath
    val execution: Task[C] = run((filePath)).onFinish {
      _ => deleteFile(filePath)
    }
    execution.unsafePerformSync
  }

private def genTempPath: AFile = {
    val path = System.getProperty("java.io.tmpdir") + "/" + scala.util.Random.nextInt().toString + ".tmp"
    sandboxAbs(posixCodec.parseAbsFile(path).get)
  }
```

But that sucks because `genTempPath` is not referentially transparent. If I change `genTempPath` to return a Task like:

```
private def genTempPath: Task[AFile] = Task.delay {
    val path = System.getProperty("java.io.tmpdir") + "/" + scala.util.Random.nextInt().toString + ".tmp"
    sandboxAbs(posixCodec.parseAbsFile(path).get)
  }
```

Then I'm back to square one. Below will not compile because I can not pass the generated `filePath` to `onFinish` call.

```
private def withTempFile[C](run: AFile => Task[C]): C = {
   val execution = (for {
      filePath <- genTempPAth
      result <- run(filePath)
    } yield(filePath, result)).onFinish {
      _ => deleteFile(filePath)
    }
    execution.unsafePerformSync
  }
```


This code above is blocking my further work (I need to add new functionality and tests to `WriteFile`) - and I've already spent a lot time on it. For now I'm going with option where `genTempFile` will not return a `Task[AFile]` but just `AFile`. I know this is not 100% the right way to do it (referential transparency broken) and I will be really happy to talk through other solutions, but for now I have stuff that need to be implemented by the end of today.


Would really appreciate if someone would like to do PP to solve my dilemma described above.

rabbitonweb: As described above, I'm treating 'withTempFile` as place where all side effects are taking place. Thus genTempFilePath was move there as inner function. Hope to clear this out today in PP session. Any comments welcomed.

jdegoes: > However I fear that withTempFile[A](f: AFile => A): A = ... is not referentially transparent and next reviewer will point this out... 

Just make it so:

```
def withTempFile[A](f: AFile => Task[A]): Task[A]
```

rabbitonweb:
@drostron @wemrysi how about now :) ?

wemrysi:
LGTM

drostron:
LGTM. As discussed in chat and a related PR, let's get `sparkcore` re-enabled via #1333 and included before merging.

mossprescott:
@rabbitonweb Probably worth merging master into this branch which will get you a new build with a higher probability of success.

rabbitonweb:
@mossprescott thx, @drostron mentioned it a sec ago, just did, waiting for CI to do its magic :)

PR #1283, opened 2016-07-25
Add `ReadFile` interpreter convenience function

jedesah:


core/src/main/scala/quasar/fs/ReadFile.scala
```diff
@@ -136,6 +136,36 @@ object ReadFile {
       lift(Close(rh))
   }
 
+  final case class CursorOffset(offset: Natural, limit: Option[Positive])
+
+  def interpret[S[_], C](open: CursorOffset => Free[S,FileSystemError \/ C], read: C => Free[S,FileSystemError \/ Vector[Data]], close: C => Free[S,Unit])(implicit
+    state: KeyValueStore.Ops[ReadHandle, C, S],
```
rabbitonweb: Maybe it's just me, but when I see variable `state` I immediately think `State` monad. Maybe naming it into `kvs` would be better? This is very subjective, I know :)

jedesah: What I don't like about `kvs` is that it's not conveying any semantic on how this `KeyValueStore` effect is intended to be used.

How about `cursorState` or `cursors`?

jedesah: I like `cursors.get` and `cursors.put`. I think it's better than `state`. Let me know if you can get behind the name `cursors`?

core/src/main/scala/quasar/fs/ReadFile.scala
```diff
@@ -136,6 +136,36 @@ object ReadFile {
...
+      case ReadFile.Read(handle) =>
+        (for {
+          cursor <- state.get(handle).toRight(FileSystemError.unknownReadHandle(handle))
+          data <- EitherT(read(cursor))
+        } yield data).run
+
+      case ReadFile.Close(handle) =>
+        (for {
+          cursor <- state.get(handle)
+          _ <- close(cursor).liftM[OptionT]
```
rabbitonweb: where is close defined?

rabbitonweb:
LGTM
When on master I can recreate my SparkCore to use the template

core/src/main/scala/quasar/fs/ReadFile.scala
```diff
@@ -136,6 +136,36 @@ object ReadFile {
       lift(Close(rh))
   }
 
+  final case class CursorOffset(offset: Natural, limit: Option[Positive])
+
+  def interpret[S[_], C](open: CursorOffset => Free[S,FileSystemError \/ C], read: C => Free[S,FileSystemError \/ Vector[Data]], close: C => Free[S,Unit])(implicit
```
wemrysi: What do you think about moving this to somewhere like `qusar.fs.impl`? Just to keep implementations separate from the definitions.

core/src/main/scala/quasar/fs/ReadFile.scala
```diff
@@ -136,6 +136,36 @@ object ReadFile {
       lift(Close(rh))
   }
 
+  final case class CursorOffset(offset: Natural, limit: Option[Positive])
```
wemrysi: How does the physical filesystem know which file to open? If we want to keep this datatype, how about renaming it to `ReadCursorOpts`? or keep it as `CursorOpts` and have `open: (AFile, CursorOpts) => Free[S, ...]`?

jedesah: Nice catch! That was meant to be a tuple. Unfortunately, wasn't able to make use of the thing yet.

`ReadCursorOpts` is a better name.

core/src/main/scala/quasar/fs/ReadFile.scala
jedesah: If you think that would be better, I can do that.

jedesah: Is the intent that something equivalent for `WriteFile` would also go in the `impl` object?

I would be somewhat nice to have something like `quasar.fs.ReadFile.impl`

core/src/main/scala/quasar/fs/ReadFile.scala
jedesah: Or actually, how about `ReadOpts`.

core/src/main/scala/quasar/fs/ReadFile.scala
wemrysi: Yes, the idea is that any implementation code should go in there (we can restructure/rename when necessary) as it is useful to separate the interfaces/abstractions from their implementation, ideally we'll eventually also do so at the artifact level.

jedesah:
@rabbitonweb @wemrysi Let me know if you think this is ready once the build passes.

I rebased already because it's a simple change and the diff would have been just as bad since I created a new file per @wemrysi's suggestion.

wemrysi:
:+1:

PR #1285, opened 2016-07-25
Qscript reduce optimize

alissapajer:


core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -829,13 +830,32 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
...
+        reducers0.zipWithIndex.filter {
+          case (_, idx) => used.contains(idx)
+        }.unzip
+      }
+
+      // reset the indices in `repair0`
+      val repair: Free[MapFunc[T, ?], ReduceIndex] = repair0.map {
+        case ReduceIndex(idx) =>
+          // TODO returns -1 if not found
+          // case should never be hit, but we don't guarantee that at the type level
```
alissapajer: @sellout I'd still like to handle this before merge - just haven't decided the best thing to do now that our lists are any size and our ints are any int.

I think I'll try using `refined` for `ReduceIndex` to ensure it's a positive integer, and see what pain points arise. I also haven't ruled out assertions which then fail out to `PlannerError`. Though that means we have to thread error handling through all optimizations.

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -127,15 +126,15 @@ class QScriptSpec extends CompilerHelpers with ScalazMatchers {
     }
 
     // TODO: Needs list compaction
-    "convert a basic reduction" in skipped {
+    "convert a basic reduction" in {
```
alissapajer: This test is now in action.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -696,7 +696,7 @@ class Transform[T[_[_]]: Recursive: Corecursive: FunctorT: EqualT: ShowT, F[_]:
   }
 }
 
-class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
+class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] extends Helpers[T] {
```
alissapajer: `ShowT` is threaded through because we need to be able to print things to debug. And it's enough work to change all the call sites where we create a `new Optimize[T]` that I'd like to keep this in here, even though exposing debugging-related necessities isn't ideal.

core/src/main/scala/quasar/qscript/QScript.scala
sellout: Yeah, I think we need to use PlannerError. Most of the optimizations shouldn’t need to change, but we do currently apply them to the `FreeQS` via the `Normalizable` type class, so that would need to have the disjunction added.

alissapajer: We need `compactReduction` to return type 
```scala
QScriptCore[T, T[F]] => PlannerError \/ QScriptCore[T, T[F]]
```
Why are you suggesting that `normalize` is special, compared to all the other optimizations we compose? We still need `applyAll` to return `F[T[F]] => F[T[F]]`, for example.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -856,13 +883,32 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
...
+          case (_, idx) => used.contains(idx)
+        }.unzip
+      }
+
+      // reset the indices in `repair0`
+      val repair: Free[MapFunc[T, ?], Int] = repair0.map {
+        case ReduceIndex(idx) => indices.indexOf(idx)
+      }
+
+      if (repair.element(-1))
```
alissapajer: @sellout this is the hacky fast solution. Whenever I have to deal with something like `indexOf` which returns something like `-1` to proclaim "failure!" it feels hacky. I thought about making the `-1` into `None` immediately and handling `Option`, but that wasn't much cleaner...

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -766,6 +766,32 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
       }
     }
 
+  def elideConstantJoin[F[_]: Traverse](
+    implicit TJ: ThetaJoin[T, ?] :<: F, QC: QScriptCore[T, ?] :<: F, FI: F :<: QScriptProject[T, ?]):
+      ThetaJoin[T, T[F]] => F[T[F]] = {
+    case x @ ThetaJoin(src0, l, r, on, Inner, combine) if on ≟ BoolLit(true) =>
+      // TODO other side
```
alissapajer: remaining TODO

alissapajer:
@sellout ok all the stuff I had in mind to address I've addressed - let me know what other things you see!

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -766,6 +766,51 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT] extends Helpers[T] {
...
+                  case RightSide => HoleF
+                }))).getOrElse(TJ.inj(x))
+              }
+              case _ => TJ.inj(x)
+            }
+          case _ => TJ.inj(x)
+        }
+        case _ => TJ.inj(x)
+      }
+    case x @ ThetaJoin(src0, l, r, on, Inner, combine) if on ≟ BoolLit(true) =>
```
sellout: This pattern can’t ever match, can it? It looks identical to the previous one.

core/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -34,6 +33,19 @@ sealed abstract class QScriptCore[T[_[_]], A] {
 @Lenses final case class Map[T[_[_]], A](src: A, f: FreeMap[T])
     extends QScriptCore[T, A]
 
+@Lenses final case class ReduceIndex(idx: Int)
+
+object ReduceIndex {
+  implicit def equal: Equal[ReduceIndex] =
+    Equal.equal((a, b) => a.idx ≟ b.idx)
```
sellout: Very minor, but this could be replaced by `Equal.equalBy(_.idx)`.

core/src/main/scala/quasar/qscript/QScript.scala
alissapajer: That is a very good call, fixing, thanks!

PR #1286, opened 2016-07-25
SD-1871 Refactor view filesystem in terms of the Mounting algebra

wemrysi:
The goal of this PR was to allow for multiple implementations of `Mounting`. To that end the view filesystem was refactored in terms of the `Mounting` algebra instead of `MountConfigs`, the latter being a particular implementation of the former.

`EvaluatorMounter` (now known as `MountRequestHandler`) was also refactored to only effect the hierarchical filesystem mounts. Composing the view filesystem was pushed all the way out to the edge of the system (see `CompositeFileSystem` in the `quasar-main` package object).

I tried to keep cleanup/deletes/renames in their own commits, as a result it may be easier to see the evolution by looking at the individual commits rather than the aggregate changes.


main/src/main/scala/quasar/main/package.scala
```diff
@@ -38,219 +38,261 @@ package object main {
...
+      *       currently need to due to how we initialize the system using
+      *       one mounting interpreter that updates these refs, but doesn't
+      *       persist configs, and then switch to another that does persist
+      *       configs post-initialization.
+      *
+      *       This should be unnecessary once we switch to lazy, on-demand
+      *       mounting.
+      *
+      * @param cfgsImpl  a `KeyValueStore` interpreter to `Task`
+      * @param hrchFsRef the current hierarchicl FileSystem interpreter, updated whenever mounts change
```
drostron: s/hierarchicl/hierarchical/

drostron:
I appreciate this view of things... Also, great to see some helpful comments, general reorganization, and clean up. Once mergeable, LGTM!

PR #1273, opened 2016-07-20
Add skeleton PostgreSQL connector

drostron:
resolves #1252

PR #1287, opened 2016-07-26
Switch to single package declaration per file

drostron:
Follow up for: https://github.com/quasar-analytics/quasar/pull/1273#discussion_r72055770


sellout:
Wow, this had snuck in more places than I realized. Thanks for cleaning them all up!

👍🏾

PR #1292, opened 2016-07-26
Shift optimizations

alissapajer:


alissapajer:
@sellout I needed to base my work off this right now, so I had to merge it. Otherwise if I based my work off of it before you merged it, I'd have to cherry-pick all my new commits on top of it after it merged. And cherry-picking isn't safe because it's really easy to lose work that way. Cherry-picking bypasses the git features of merging that guarantee work doesn't get lost.

What I didn't isn't much better, (opened a second duplicate PR, rebased against master) because your original PR here didn't close automatically because I had to rebase the duplicate against master to be able to merge it with our PR merge script. We are purposefully bypassing helpful git features (ability to recognize when a merged (not rebased) commit has landed in master) in order to have a "clean" commit history in master. And we're guaranteeing this "clean" history at the cost of not being able to work off each other's branches, which forces us to cherry-pick, which isn't safe. 

If we had the protocol that once a PR is submitted, it cannot be rebased (not against itself, not against master) then we could safely work off each others commits. As it stands now, there is no proof (at the git level) that I didn't mess up and lose some of the work that was in Greg's PR by merging it the way I did. If I didn't merge it in this way, then I'd be risking losing my current work later via cherry-pick. There is no win.

Additionally, we can still have "wip" PRs that are designated as such, the protocol being "this PR can be rebased at any point so don't base your work on it!" With the additional text "Talk to the PR author if you need to base yoru work on it to see if the commits can be finalized."

Clearly this is a conversation for the whole backend team. But since I was experiencing frustration I decided to type this all here now. I don't like that I bypassed the PR process like this, but I didn't see a better option. If anyone else has a better option let me know.

PR #1293, opened 2016-07-26
Fix up travis config.

paulp:
Add cache of ~/.ivy2 and ~/.sbt, this speeds up builds quite a bit.
Set scala version to 2.11.8 and note that it's not doing anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: scala
-scala:
-- 2.11.7
+scala: 2.11.8 # but this setting isn't used when there's an explicit script
 jdk: oraclejdk8
 sudo: false
 script:
@@ -17,6 +16,16 @@ notifications:
       %{build_url}"
   slack:
     secure: KPCXtCAjtlpRQcH7TQtprrf67yIqPVpht9L6/+wbbNd14Tj3x53fiZ3OTabHVx3SL2ZWxbWq+FJGxdIFMyxC/fC0bV9flB7PlRtqHJV1Q7g49odAMAZL7syWr7vFP5bJt+TXuxHvMBihDHFk1Cehy4a1cnrOMReDgqGIFCwQtbI=
+
+cache:
+  directories:
+  - $HOME/.ivy2/cache
+  - $HOME/.sbt
+
+before_cache:
+- find "$HOME/.sbt/" -name '*.lock' -print0 | xargs -0 rm
+- find "$HOME/.ivy2/" -name 'ivydata-*.properties' -print0 | xargs -0 rm
+
 env:
   matrix:
   - MONGO_RELEASE=2_6
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Add caching to travis config.


.travis.yml
```diff
@@ -17,6 +16,16 @@ notifications:
...
   slack:
     secure: KPCXtCAjtlpRQcH7TQtprrf67yIqPVpht9L6/+wbbNd14Tj3x53fiZ3OTabHVx3SL2ZWxbWq+FJGxdIFMyxC/fC0bV9flB7PlRtqHJV1Q7g49odAMAZL7syWr7vFP5bJt+TXuxHvMBihDHFk1Cehy4a1cnrOMReDgqGIFCwQtbI=
+
+cache:
+  directories:
+  - $HOME/.ivy2/cache
+  - $HOME/.sbt
+
+before_cache:
+- find "$HOME/.sbt/" -name '*.lock' -print0 | xargs -0 rm
```
jedesah: Is this just printing out the contents of the two cached directories?

My bash scripting expertise is limited...

paulp: It's removing all the files called *.lock found under there, as they defeat caching. (Travis things the directory contents have changed even when the change is meaningless.)

paulp: xargs just means "take everything on the incoming pipeline and use it as arguments to the command I'm about to give"

paulp: ...and `-0` is so as not to be defeated by files with spaces in them, it uses null as a token separator.

paulp: You may be aware that some versions of find have a `-delete` option, which we'd use if we could. But the one on travis doesn't.

paulp:
Any objections to me merging this?

jedesah:
LGTM.

mossprescott:
Sorry I was too slow, but I would have objected if I had caught this in time.

I experimented with Travis's caching earlier and found that it made builds slower. Based on the two build attempts on this PR, I'd say it's not clear if it's helping now. Also in general we try not to merge if Travis builds fail; even if we have to restart them once or twice occasionally due to random test failures.

paulp:
We're not getting a clear picture of what's happening at travis, but we can certainly revert this if it doesn't speed things up. In my experience, it definitely does.

[The tests did pass](https://travis-ci.org/quasar-analytics/quasar/builds/147560036) - this is the same commit, rebased. Or maybe you are referring to a different PR.

paulp:
In any case there's no way to make any reliable determination based on a couple of builds. Look at the variation in the time to completion among the four builds. The existing variation in build times dwarfs the impact of the caching, whatever that impact is.

screen shot 2016-07-26 at 3 34 42 pm

paulp:
Noting for the record that I have since observed that the lion's share of the timing disparity is because the first one executes a bunch of tasks the others don't. But even the five minute spread is a lot.

mossprescott:
FYI, I just restarted various recent failed builds in Travis, including the build of this commit on master so we can get some kind of measure of the effect.

The time also seems to depend on the time of day (presumably because of unrelated load on the build farm).

paulp:
Ok - to be clear there will be no effect whatsoever from the caching commit until travis actually utilizes the cache in builds. I think this requires all the jobs on the main branch succeeding so the cache is marked as valid, after which pull requests against that branch will be built with the cache populated.

paulp:
To contrast the behavior with and without the cache, you can examine the logs from these builds:

[No cache](https://travis-ci.org/paulp/quasar/jobs/147605716)
[With cache](https://travis-ci.org/paulp/quasar/jobs/147629746)

Here is the salient difference between them:
```
% egrep 'downloading|SUCCESSFUL' log1.txt |wc -l
     334
% egrep 'downloading|SUCCESSFUL' log2.txt |wc -l
       0
```
Here is the sum of the measurements from the run without caching:
```
egrep 'SUCCESSFUL' log1.txt | perl -pe 's#.*[(](\d+)ms[)]$#\1#' | sum-numbers
0.0 + 416 + 1987 + 1188 + 503 + 585 + 612 + 437 + 510 + 995 + 471 + 405 + 389
+ 423 + 507 + 51 + 458 + 556 + 535 + 492 + 373 + 348 + 396 + 487 + 388 + 406 +
5861 + 1345 + 782 + 560 + 146 + 82 + 439 + 684 + 539 + 490 + 415 + 366 + 131 +
282 + 249 + 480 + 371 + 414 + 103 + 729 + 146 + 70 + 495 + 541 + 519 + 53 +
515 + 511 + 112 + 122 + 273 + 65 + 34 + 468 + 43 + 87 + 39 + 102 + 525 + 43 +
104 + 816 + 70 + 37 + 132 + 83 + 49 + 167 + 2434 + 179 + 285 + 396 + 189 + 294
+ 213 + 4566 + 1486 + 355 + 303 + 234 + 4731 + 313 + 709 + 671 + 563 + 200 +
228 + 227 + 329 + 253 + 297 + 294 + 833 + 1150 + 573 + 294 + 385 + 214 + 178 +
226 + 653 + 199 + 289 + 307 + 307 + 414 + 214 + 359 + 185 + 199 + 1123 + 224 +
191 + 205 + 536 + 190 + 178 + 1564 + 198 + 347 + 418 + 1087 + 223 + 190 + 196
+ 1262 + 1147 + 582 + 384 + 360 + 342 + 1071 + 571 + 516 + 179 + 217 + 262 +
373 + 174 + 187 + 215 + 199 + 441 + 480 + 461 + 189 + 178 + 261 + 1288 + 891 +
203 + 277 + 181 + 210 + 275 + 316 + 174 + 199 + 738 + 205 + 392 + 183
-----
83788.0
```
84 seconds. I *hope* it doesn't take travis 84 seconds to populate the cache.

mossprescott:
I once had hope, too.

paulp:
Observing logs unfold in real time, it appears the cache is populated plus-or-minus instantaneously. Did you witness that specifically being slow? Or was it just that end-to-end times were worse after enabling caching?

mossprescott:
When I looked at it, it was taking several minutes to populate the cache, _and_ the builds were slower overall as a result. But this was some time ago (right after we switched to containerized builds, when they were a new thing). It's certainly possible it's gotten better since then. I just don't want to make assumption and not verify it.

Issue #1277, opened 2016-07-22
"Warning: Unknown ScalaCheck args provided: exclude exclusive"


paulp:
That message is produced a bunch of times if I run `./sbt test`. I see there's something going on with sbt configurations, but the message suggests that it is unintentionally being passed as an argument to scalacheck. Maybe it's harmless (but, still noise.)

paulp:
#1296

PR #1296, opened 2016-07-27
Pass specs arguments only to specs.

paulp:


alissapajer:
@paulp does this fix #1277? If so, is this ready for merge? If not, since we use a couple different test frameworks, it still seems good practice to specificy which one we'd like to specify the arguments for.

paulp:
Yep, fixes #1277 and I'm not sure why sbt even offers an option to pass options to "all test frameworks" because if there are two or more they won't work somewhere.

paulp:
Oh, the "merge conflicts" are revision.md. That means any pull request which includes that file is immediately in conflict after something else is merged. Not so good.

paulp:
I see, the script is also supposed to prevent `CHANGELOG/revision.md` from turning up in master, right? But there it is.

alissapajer:
The revision.md file should get deleted after merge, so the merge script likely isn't working properly.

alissapajer:
Ok, looking at [the script](https://github.com/alissapajer/devtools/blob/master/bin/merge#L35), it looks like the four valid file names are:
```
breaking="CHANGELOG/breaking.md"
feature="CHANGELOG/feature.md"
bugfix="CHANGELOG/bugfix.md"
release="CHANGELOG/release.md"
```
So the `revision.md` file that exists in master should just be deleted.

alissapajer:
Take back what I wrote above - the link is to my own out-of-date fork of `devtools`. The actual valid file names are
```
breaking="CHANGELOG/breaking.md"
feature="CHANGELOG/feature.md"
revision="CHANGELOG/revision.md"
release="CHANGELOG/release.md"
```
as seen in the [merge script on master](https://github.com/slamdata/devtools/blob/master/bin/merge).

alissapajer:
This has been resolved in slack in the #backend room. It would be awesome if there was a way to link here to a specific chat in a slack channel!

jdegoes:
@alissapajer http://gitter.im/quasar-analytics/quasar

PR #1297, opened 2016-07-28
Publish foundation's test artifact

wemrysi:
Otherwise downstream dependencies that make use of quasar tests cannot
access QuasarSpecification.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Publish the foundation project's test artifacts for access to QuasarSpecification
--- a/build.sbt
+++ b/build.sbt
@@ -180,7 +180,9 @@ lazy val foundation = project
   .settings(name := "quasar-foundation-internal")
   .settings(oneJarSettings: _*)
   .settings(publishSettings: _*)
-  .settings(libraryDependencies ++= Dependencies.core)
+  .settings(
+    libraryDependencies ++= Dependencies.core,
+    publishArtifact in (Test, packageBin) := true)
   .enablePlugins(AutomateHeaderPlugin)
 
 lazy val ejson = project


paulp:
LGTM! Bonus points for avoiding the doc/src artifacts. If only the tests would cooperate.

paulp:
(I restarted the job which failed but as things currently stand I really think 3/4 passing jobs ought to be a pass.)

mossprescott:
I strongly disagree and if I have to drop everything to fix those random test failures I'm willing to do it.

paulp:
You are more than welcome to join me in my current CI investigations. I don't think it'll be a matter of fixing a thing but a longer process of hardening/improving a number of things.

wemrysi:
@paulp Yeah, I think you're right about the CI issues. Though even if this PR passes, I'd like to wait and see if #1286 gets a clean build, I've been trying to merge that one since yesterday and gave up last night.

alissapajer:
This is where we're at. We've been merging 3/4 PRs (and even 2/4 PRs) for a while now :/

bildschirmfoto 2016-07-28 um 20 36 45

mossprescott:
Yes, I'm aware of how bad it's gotten. With the larger team, it's much more difficult now to wait for clean builds. That doesn't make it right.

paulp:
It would be useful to know exactly what is being accomplished with the four different test runs. I understand they are using different versions of mongo, but how many different code paths are genuinely exercised? The first run also does code coverage and generates the secondary artifacts; in my branch-in-progress I broke those tasks up across multiple jobs to get a little job parallelism out.

wemrysi:
This passed 3/4 builds and doesn't actually change any non-build code, so I'm going to merge unless anyone objects.

PR #1299, opened 2016-07-28
Various fixes to help filesystem authors.

sellout:
- Eliminate the DeadEnd Empty case from QScript.
- Make Read not a functor
- Generalize FreeQS to include any QScript component.
- Track LeftShift provenance properly.

### DIFF
--- a/core/src/main/scala/quasar/qscript/DeadEnd.scala
+++ b/core/src/main/scala/quasar/qscript/DeadEnd.scala
@@ -24,7 +24,6 @@ sealed abstract class DeadEnd
   * in the structure a backend sees, it represents the mount point.
   */
 final case object Root extends DeadEnd
-final case object Empty extends DeadEnd
 
 object DeadEnd {
   implicit def equal: Equal[DeadEnd] = Equal.equalRef
--- a/core/src/main/scala/quasar/qscript/EquiJoin.scala
+++ b/core/src/main/scala/quasar/qscript/EquiJoin.scala
@@ -19,6 +19,7 @@ package quasar.qscript
 import quasar.fp._
 
 import matryoshka._
+import matryoshka.patterns._
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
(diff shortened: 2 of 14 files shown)


core/src/main/scala/quasar/qscript/DeadEnd.scala
```diff
@@ -24,7 +24,6 @@ sealed abstract class DeadEnd
   * in the structure a backend sees, it represents the mount point.
   */
 final case object Root extends DeadEnd
-final case object Empty extends DeadEnd
```
alissapajer: Ah right, this was left over from when we thought `QScript` would handle `Let`.

core/src/main/scala/quasar/qscript/package.scala
```diff
@@ -34,35 +34,27 @@ import scalaz._, Scalaz._
...
+  private type QScriptTotal1[T[_[_]], A] =
+    Coproduct[ThetaJoin[T, ?], QScriptTotal0[T, ?], A]
+  private type QScriptTotal2[T[_[_]], A] =
+    Coproduct[EquiJoin[T, ?], QScriptTotal1[T, ?], A]
+  /** This type is used for join branch-like structures. It’s an unfortunate
+    * consequence of not having mutually-recursive data structures. Once we do,
+    * this can go away.
     */
-  type EquiQScript[T[_[_]], A] = Coproduct[EquiJoin[T, ?], QScriptCommon[T, ?], A]
+  type QScriptTotal[T[_[_]], A] =
```
alissapajer: @sellout so a backend author is now working with a `QScriptTotal`, correct? Since `QScriptTotal2` is `private`, they don't have the option of a qscript without `Read`, for example.

core/src/main/scala/quasar/qscript/EquiJoin.scala
```diff
@@ -55,6 +56,21 @@ object EquiJoin {
         }
     }
 
+  implicit def show[T[_[_]]: ShowT]: Delay[Show, EquiJoin[T, ?]] =
+    new Delay[Show, EquiJoin[T, ?]] {
+      def apply[A](showA: Show[A]): Show[EquiJoin[T, A]] = Show.show {
+        case EquiJoin(src, lBr, rBr, lkey, rkey, f, combine) =>
```
alissapajer: @sellout do you know of any libraries that create `Show` instances like this for us? For something like `Show` on a case class, we generally just want it to print out the standard thing.

alissapajer:
@sellout LGTM

PR #1302, opened 2016-07-28
More declarative sbt publishing settings

wemrysi:
Makes sbt publishing settings more declarative by publishing all of a project's non-test artifacts "by default" (i.e. included in `commonSettings`). Opting out of publishing can be done by including the `noPublishSettings` and enabling publishing of a project's tests can be done by including `publishTestsSettings`.

Also enables publishing of the `skeleton` project as it is depended on by `main` and breaks downstream projects depending on quasar otherwise.


sellout:
Sorry for the delay. This looks good, but what’s the motivation for publishing tests?

wemrysi:
quasar-advanced

sellout:
Sorry, could you elaborate on that? Why are the tests needed in that case? Is this a case where it would be nice to have something like a `scalacheck` project, where arbitrary instances, laws, etc. are exposed, or is there something else here?

Also, I’m marking this ready, just want to understand this better myself.

wemrysi:
@sellout Yes, all those would be good to have, but we also have specifications for all the api algebras that anyone writing other implementations of those algebras would want to test their implementations against. It might make sense to figure out how to make them into their own project at some point, though I think that might be a bit premature as we're still figuring out how to organize the rest of the project.

The immediate need is for `quasar-advanced`, specifically for the `MountingSpec` (as `quasar-advanced` uses a different implementation of this algebra than `quasar`) which now inherits from `QuasarSpecification`, the latter of which was not being published and thus one could not compile quasar-advanced.

wemrysi:
@sellout I did some minor tidying of the onejar settings and also merged in master. It is building again, so feel free to merge (or reject) at your convenience.

sellout:
Looks good. Just waiting on Travis 😆

wemrysi:
@sellout Attempting the RO job once again, I've managed 3/4 greens on this one, but even the spark "fix" doesn't seem to have helped here. Maybe we'll get lucky this time.

sellout:
I think my spark fix just fixes an issue that prevented the 2.6 build from ever(?) passing.

wemrysi:
@sellout Still no luck here. I guess we can wait and see if @paulp's CI improvements permit progress.

wemrysi:
@sellout Ok, up to date on master with a clean build!

PR #1303, opened 2016-07-29
[WIP] support for new pipeline ops in MongoDB 3.2 with type safety

mossprescott:
For an introduction, see the first commit's comment.

This is not ready to merge yet:
- needs to be rebased/merged with master
- names and other details could use another look
- some drift has occurred that I intend to clean up to reduce noise

But I would very much appreciate any feedback on whether there are easier ways to do some of what I'm doing here.

If any large-scale changes result from feedback, I'll close this and re-submit a cleaned-up PR.

@sellout: feel free to give as much or as little time as you can afford. I'm afraid you're the only one with much hope of understanding the ramifications of much of this.

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.physical.mongodb
+
+import scalaz.Semigroup
+
+sealed trait MongoQueryModel
+
+object MongoQueryModel {
+  /** The oldest supported version. */
+  case object `2.6` extends MongoQueryModel
+
+  /** Adds a few operators. */
+  case object `3.0` extends MongoQueryModel
+
+  /** Adds `$lookup` and `$distinct`, several new operators, and makes
+    * accumulation operators available in `$project`. */
+  case object `3.2` extends MongoQueryModel
+
+  implicit val semigroup: Semigroup[MongoQueryModel] =
+    new Semigroup[MongoQueryModel] {
+      def append(f1: MongoQueryModel, f2: => MongoQueryModel) = (f1, f2) match {
+        case (`3.2`, _) => `3.2`
+        case (_, `3.2`) => `3.2`
+        case (`3.0`, _) => `3.0`
+        case (_, `3.0`) => `3.0`
+        case (`2.6`, _) => `2.6`
+      }
+    }
+}
--- a/mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
@@ -27,8 +27,8 @@ object MongoQueryModel {
   /** Adds a few operators. */
   case object `3.0` extends MongoQueryModel
 
-  /** Adds `$lookup` and `$distinct`, several new operators, and makes
-    * accumulation operators available in `$project`. */
+  /** Adds \$lookup and \$distinct, several new operators, and makes
+    * accumulation operators available in \$project. */
   case object `3.2` extends MongoQueryModel
 
   implicit val semigroup: Semigroup[MongoQueryModel] =
--- a/mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
@@ -29,7 +29,7 @@ import quasar.qscript._
 import matryoshka._, Recursive.ops._, FunctorT.ops._
 import scalaz._, Scalaz._
 
-sealed trait WorkflowBuilderF[+A]
+sealed trait WorkflowBuilderF[F[_], +A]
 
 object WorkflowBuilder {
   import quasar.physical.mongodb.accumulator._
@@ -38,7 +38,7 @@ object WorkflowBuilder {
   import IdHandling._
 
   /** A partial description of a query that can be run on an instance of MongoDB */
-  type WorkflowBuilder = Fix[WorkflowBuilderF]
+  type WorkflowBuilder[F[_]] = Fix[WorkflowBuilderF[F, ?]]
   /** If we know what the shape is, represents the list of Fields. */
   type Schema = Option[NonEmptyList[BsonField.Name]]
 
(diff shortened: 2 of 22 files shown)


mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  /** Adds a few operators. */
+  case object `3.0` extends MongoQueryModel
+
+  /** Adds `$lookup` and `$distinct`, several new operators, and makes
+    * accumulation operators available in `$project`. */
+  case object `3.2` extends MongoQueryModel
+
+  implicit val semigroup: Semigroup[MongoQueryModel] =
+    new Semigroup[MongoQueryModel] {
```
wemrysi: Might want to consider making this `Semigroup[MongoQueryModel @@ MaxVal]` (from `scalaz.Tags`)

mossprescott: That's interesting, but actually I no longer have any need to combine these values, so I'm tempted to just remove it.

mossprescott:
I'm happier with the way the types read now after coming up with a new scheme.

@sellout, @wemrysi: If you haven't seriously looked yet, you're once again invited ;)

wemrysi:
@mossprescott Thanks, am working my way through it.

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -62,29 +62,29 @@ object WorkflowBuilder {
...
+      extends WorkflowBuilderF[F, A]
   {
     lazy val dummyOp =
       op(
         inputs.zipWithIndex.map {
           case (_, index) => BsonField.Name("_" + index)
         })(
         // Nb. This read is an arbitrary value that allows us to compare the partial function
-        $read(Collection("", "")))
+        $read[F](Collection("", "")))
```
wemrysi: I know this wasn't added in this PR (and not suggesting that it needs to change in this PR) and that I also don't fully understand its ramifications, but nonetheless this seems dubious to me.

Do we use universal equality on this type? If we didn't we might be able to localize the `dummyOp` to the `Equal` instance (if it isn't possible to remove it alltogether), which would also mean we wouldn't need the typeclass constraints on the constructor.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -114,185 +163,211 @@ object Workflow {
...
+        case _                       => None
+      }
+      case $out(src, _) => src.unFix match {
+        case $read(_) => src.unFix.some
+        case _        => None
+      }
+      case _ => None
     }
+  }
+  implicit lazy val coalesce2_6: Coalesce[WorkflowOpCoreF] = coalesceAll[WorkflowOpCoreF]
```
wemrysi: Are these necessary? Couldn't you just make `coalesceAll` implicit?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -348,55 +423,143 @@ object Workflow {
...
+    }
   }
 
+  def crushInjected[F[_]: Functor, G[_]: Functor](implicit I: F :<: G, CG: Crush[G]): Crush[F] =
+    new Crush[F] {
+      def crush(op: F[(Fix[F], (DocVar, WorkflowTask))]): (DocVar, WorkflowTask) =
+        CG.crush(I.inj(op.map { case (f, t) => (f.transCata(I.inj(_)), t) }))
+    }
+
+  implicit val crush2_6: Crush[WorkflowOpCoreF] = crushInjected[WorkflowOpCoreF, WorkflowF]
```
wemrysi: You should be able to get rid of this by moving `crushInjected` to the companion object of `Crush` and making it implicit.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -412,82 +575,261 @@ object Workflow {
...
+    def shapePreserving[A](v: Coproduct[F, G, A]) =
+      v.run.fold(
+        CF.shapePreserving(_).map(_.fmap(ι, Inject[F, Coproduct[F, G, ?]])),
+        CG.shapePreserving(_).map(_.fmap(ι, Inject[G, Coproduct[F, G, ?]])))
+  }
+
+
+  /** Newtype for source ops (that is, ops that are themselves sources). */
+  // TODO: prevent construction of invalid instances
+  final case class SourceF[F[_]: Functor, A](wf: F[A]) {
```
wemrysi: I'd move the functor constraint onto `op` in order to remove it from `fmap` and make better use of parametricity.

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
mossprescott: That's a good point and this seems like the time to address it.

Removing `equals` does cause a single test failure, but I should be able to track it down and switch to using `Equal`.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -412,82 +575,261 @@ object Workflow {
...
+
     /** Reparenting that handles coalescing (but is more restrictive as a
       * result).
       */
-    def reparentW(newSrc: Workflow): Workflow = Fix(reparent(newSrc))
+    // TODO: this doesn't seem to actually handle coalescing, so what was the
+    // comment referring to?
+    def reparentW(newSrc: Fix[F]): Fix[F] = Fix(reparent(newSrc).wf)
+
+    def fmap[G[_]: Functor, B](f: A => B, g: F ~> G): SingleSourceF[G, B] =
```
wemrysi: It doesn't look like this `Functor` constraint is necessary, am i missing something?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -503,145 +845,172 @@ object Workflow {
...
+          // TODO: this can coalesce more cases, but hasn’t been done thus far and
+          //       requires rewriting many tests in a much less readable way.
+          // .cata[Workflow](x => coalesce(uncleanƒ(x).unFix))
+      )
+    }
   }
 
-  final case class $Pure(value: Bson) extends SourceOp
-  def $pure(value: Bson) = Fix(coalesce($Pure(value)))
+  implicit lazy val crystallize2_6: Crystallize[WorkflowOpCoreF] = crystallizeAll[WorkflowOpCoreF]
```
wemrysi: So I'm seeing a pattern, what's the reason for preferring the `lazy val` to making `crystallizeAll` implicit?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -1231,58 +1835,79 @@ object Workflow {
...
+    def render(v: WorkflowOp3_2F[Unit]) = v match {
+      case $LookupF(_, from, localField, foreignField, as) =>
+        Terminal("$LookupF" :: wfType, Some(s"$from with $foreignField = $localField as $as"))
+      case $SampleF(_, size) =>
+        Terminal("$SampleF" :: wfType, Some(size.toString))
+    }
+  }
+
+  // TODO: where does this belong?
+  implicit def coproductRenderTree[F[_], G[_], A]
```
wemrysi: `RenderTree` companion?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
mossprescott: It was necessitated by the constraint on SourceF, so it goes away too.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
mossprescott: It had to do with mutual recursion which seems to have resolved itself since they were written. I was able to get rid of these.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
mossprescott: For some reason, this does not work. I still think moving `crushInjected` to the companion makes sense, though.

mossprescott:
@wemrysi: there's more that could be done here, but I think it's in a good place so I'm planning to squash, rebase, and submit a new PR on a fresh branch. Hopefully that goes smoothly and you can review it quickly ;)

PR #1304, opened 2016-07-29
Qscript leftshift optimize

alissapajer:


core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -108,8 +108,7 @@ object MapFunc {
 
     def apply[T[_[_]]: Recursive: Corecursive, T2[_[_]]: Corecursive, A](args: List[Free[MapFunc[T2, ?], A]]):
         Free[MapFunc[T2, ?], A] =
-      apply(args.map(_.ana[T, CoEnv[A, MapFunc[T2, ?], ?]](CoEnv.freeIso[A, MapFunc[T2, ?]].reverseGet)))
-        .embed.cata(CoEnv.freeIso[A, MapFunc[T2, ?]].get)
+      apply(args.map(_.toCoEnv[T])).embed.fromCoEnv
```
alissapajer: It's considerably easier to go between `Free` and `CoEnv` now!

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -168,7 +168,7 @@ class QScriptSpec extends CompilerHelpers with ScalazMatchers {
           Free.roll(MakeMap(StrLit("loc"), Free.point(RightSide))))).embed.some)
     }
 
-    "convert a constant shift array" in pending {
```
alissapajer: These two more tests are passing.

foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -624,7 +624,7 @@ package object fp
   // apomorphism - short circuit by returning left
   def substitute[T[_[_]], F[_]](original: T[F], replacement: T[F])(implicit T: Equal[T[F]]):
       T[F] => T[F] \/ T[F] =
-   tf => if (tf ≟ original) replacement.left else original.right
+   tf => if (tf ≟ original) replacement.left else tf.right
```
alissapajer: super important bug fix - no more SOE!

alissapajer: and there's a test (indirectly) covering the regression as well

core/src/main/scala/quasar/qscript/package.scala
```diff
@@ -126,21 +126,6 @@ package object qscript {
       Free.roll(ProjectIndex(HoleF[T], IntLit[T, Hole](1))),
       Free.roll(ProjectIndex(HoleF[T], IntLit[T, Hole](2))))
 
-  /** Applies a transformation over `Free`, treating it like `T[CoEnv]`.
-    */
-  def freeTransCata[T[_[_]]: Recursive: Corecursive, F[_]: Functor, A](
```
alissapajer: These just moved to regular `fp` package.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -899,6 +898,47 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] extends Helpers[T
...
+          val zeroRef: T[CoEnv[JoinSide, MapFunc[T, ?], ?]] =
+            Free.roll[MapFunc[T, ?], JoinSide](ProjectIndex(rightSide, IntLit(0))).toCoEnv[T]
+          val oneRef: T[CoEnv[JoinSide, MapFunc[T, ?], ?]] =
+            Free.roll[MapFunc[T, ?], JoinSide](ProjectIndex(rightSide, IntLit(1))).toCoEnv[T]
+
+          val zerosCount: Int = repair.para(count(zeroRef))
+          val onesCount: Int = repair.para(count(oneRef))
+
+          if (zerosCount < 1 && onesCount < 1) {   // access neither element
+            SP.inj(x) // TODO `struct` is never used - should we remove it?
```
alissapajer: @sellout here is the case when we don't directly deref either side.

Option A: We return `SP.inj(x)`. If there is a situation where the specific check done here to see if index `0` or `1` exists fails, but one or both sides are still accessed, we would be breaking the query. In general, it's possible for an index or field to be referenced dynamically rather than statically, but in this situation, I _think_ we are always explicitly creating a `ProjectIndex` of the integer literal. So _maybe_ it's safe to remove `struct`, but I'm not confident about that yet.

Option B: We remove `struct`. The only reason to do this is so that backends don't compute the contents of `struct` only to not use them.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -899,6 +898,47 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] extends Helpers[T
...
+            val replacement: T[CoEnv[JoinSide, MapFunc[T, ?], ?]] =
+              transApoT(repair)(substitute(zeroRef, rightSideCoEnv))
+            SP.inj(LeftShift(src, Free.roll[MapFunc[T, ?], Hole](DupArrayIndices(elem)), replacement.fromCoEnv))
+          } else if (zerosCount < 1) {   // only access element 1
+            val replacement: T[CoEnv[JoinSide, MapFunc[T, ?], ?]] =
+              transApoT(repair)(substitute(oneRef, rightSideCoEnv))
+            SP.inj(LeftShift(src, elem, replacement.fromCoEnv))
+          } else {   // access both elements
+            SP.inj(x)
+          }
```
alissapajer: Also, I'm not thrilled with the algorithm here. It reminds me of
```scala
if (option.isDefined) option.get else ...
```
where you can generally use `getOrElse` to avoid that pattern. Ideally here we could tie the existence check and the substitution more closely together, but I don't have a solution that does that yet.

core/src/main/scala/quasar/qscript/QScript.scala
sellout: If this optimization gets applied again after removing the explicit `ProjectIndex(RightSide, 1)`, then this would break the query.

We’re already replacing LeftShifts that don’t access RightSide into Maps, so the safe cases of this are covered.

I think what we really need to check here is that there are no dynamic accesses of either side. E.g., rather than `if (onesCount < 1)` we need `if (zerosCount ≟ totalRightSideCount)` (i.e., all references of RightSide point to the provenance).

sellout:
Ok, I think that handling of 0/1 counts is my only comment.

core/src/main/scala/quasar/qscript/QScript.scala
alissapajer: Right, that's a good point. A query could reference `RightSide`, which would be the entire `ZipArrayIndices`, and also project out only one of the indices. In that case, we'd still need the entire `ZipArrayIndices` structure.

alissapajer:
@sellout I pushed a fix addressing your comment. Let me know if that addresses it in full.

core/src/main/scala/quasar/qscript/QScript.scala
sellout: Oh, and just noticed (since I’m basing on this work now) that there’s a parallel optimization not being applied – `ZipMapKeys` should be optimized the same way as `ZipArrayIndices`.

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -918,20 +918,17 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] extends Helpers[T
           val zeroRef: T[CoEnv[JoinSide, MapFunc[T, ?], ?]] = makeRef(0)
           val oneRef: T[CoEnv[JoinSide, MapFunc[T, ?], ?]] = makeRef(1)
 
-          val zerosCount: Int = repair.para(count(zeroRef))
-          val onesCount: Int = repair.para(count(oneRef))
+          val rightCount: Int = repair.para(count(rightSideCoEnv))
```
sellout: You can theoretically do something like
```
val (rightCount, (zerosCount, onesCount)) =
  repair.para(count(rightSideCoEnv) zip (count(zeroRef) zip count(oneRef)))
```
to do all the counts in one pass. But I have a hunch that the type annotations currently required for this make it a frustrating proposition.

sellout:
And this is going to require a `CHANGELOG/revision.md` file.

core/src/main/scala/quasar/qscript/QScript.scala
alissapajer: I think just to keep moving on things I'll keep it the way it is now, if you're okay with that too.

alissapajer:
@sellout changelog is in

alissapajer:
@sellout ok we now have support for optimization of map keys and array indices

core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -901,9 +901,15 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] extends Helpers[T
...
-        case -\/(ZipArrayIndices(elem)) => {
+    case x @ LeftShift(src, struct, repair0) => {
+      val pattern: Option[(FreeMap[T], FreeMap[T] => Unary[T, FreeMap[T]])] = struct.resume match {
+        case -\/(ZipArrayIndices(elem)) => Some((elem, fm => DupArrayIndices(fm)))
+        case -\/(ZipMapKeys(elem)) => Some((elem, fm => DupMapKeys(fm)))
+        case _ => None
+      }
+
+      pattern match {
+        case Some((elem, dup)) => {
```
sellout: Why not make this a function, then the `val pattern …` can be replaced with

```scala
struct.resume match {
  case -\/(ZipArrayIndices(elem)) => newFunc(elem, DupArrayIndices(_))
  case -\/(ZipMapKeys(elem)) => newFunc(elem, DupMapKeys(_))
  case _ => SP.inj(x)
}
```

alissapajer: right that would get rid of the `Option`y thing I had going there

alissapajer:
@sellout just pushed a commit rewriting the optimization to not use `Option`

alissapajer:
also @sellout if you come across something minor and want to merge this today, feel from to make a commit and then still merge, just ping me on it so I can check it later - I'm off for the night now!

core/src/main/scala/quasar/qscript/QScript.scala
alissapajer: Arg, I just did this, but I didn't see this comment until now because github hid it :( Need to investigate if there is a github setting where we can turn off the collapsing of comments on outdated diffs, because those comments are very often still important!

alissapajer: Well, the best answer I got was to run this javascript:
```javascript
javascript:(function(){$(".outdated-diff-comment-container").addClass("open");}());
```

alissapajer:
@sellout the best way to handle this might be for you to merge this PR first, then merge master into your PR #1306, and then I'll merge #1306. Let me know what you think.

alissapajer:
@sellout or you just merge this into #1306 (some of these commits are merged already, but there are new commits now) and then I merge #1306 closing this PR too! :)

alissapajer:
@sellout ok I decided the best approach was actually for me to merge our work together and then just fix the test expectations. PR to you here https://github.com/sellout/quasar/pull/3.

PR #1305, opened 2016-07-29
Project pillars

alissapajer:
Moves various files (basically the ones not dependent on `fs` and `logicalplan` and `data`) into a new project `pillars`, which is a dependency of `core`.


alissapajer:
@wemrysi these are the files that I could move out without tackling the big mess of the filesystem stuff entangled with logicalplan and now qscript.

I don't know yet if this will make `core` compile times faster by much, but it's a start. And in any case, it makes our dependencies a bit clearer.

wemrysi:
Thanks for breaking the seal on this, after looking at the usages, what do you think about

1. Move `quasar/console` into the `effect` package (probably deserves to be made algebraic at some point)
2. Put all of the effect stuff into an `effect` project
3. Move `jscore` and `javascript` into their own project, `js`?
4. Leave `csv`, `namegen` and `stacktrace` in `core` for now (or can put them into `pillars` but it would be less compelling)

Also, if we moved the `Data#toJs` functionality, which is only used in `quasar.api.services.query.compile` and the `mongodb` connector, into `mongodb` then the only project that would need `js` would be `mongodb`. I think we could replace the usage in `api.services` with a conversion to `EJson`, though it might require a pretty-printer or somesuch, so we can punt on this if so (but sounds like we'll need it eventually).

alissapajer:
@wemrysi I think it makes sense to have a `js` project containing `javascript` and `jscore`, so I went ahead and did that. (Side note - typed "package" instead of "project" in a commit message :/ )

Regarding the creation of an `effects` project, and what should go in it:

1. I think `quasar/console` makes sense into `effects`, since it's interacting with the environment.
2. I would consider putting `csv` into `effects`. We explicitly create a `java.io.StringReader` in `csv`, which is performing IO.
3. I would also consider putting `stacktrace` into `effects`, as stacktraces only come into play when we are running the code and thus performing IO.
4. I think `namegen` would be fine in either `pillars` or `core`. I'm fine putting it in `core` for now.
5. I don't see a reason for `LiftedOps.scala` to be in `effects`, so I'd move that into the `fp` package object in `foundation`. It's useful for effect-related operations, but really it's just a general functional operation involving `Free`.

And then with `effects` and `js` both their own projects, there would be nothing left in `pillars`, and so `pillars` would be deleted.

Let me know what you think.

alissapajer:
Also, I think it would be great to remove `core`'s dependency on `js`. We could then have:
```
     core,
// / / | \ \
      js,       \         \          \
//     |         \         \          \
     mongodb, skeleton, postgresql, sparkcore,
//      \ \ | / /
          main,
```
Once the old mongo backend is gone and replaced with the new qscript-based mongo connector, we can delete `Data` and use `EJson` exclusively. So we'll definitely need to replace the `api.services` `Data#toJs` call at some point.

Also, looking at that diagram, if the functionality in `js` is 100% mongodb-specific, and no other backends could benefit from it, we could move it into the `mongodb` project. Though since small projects are the goal (both for understanding/isolating dependencies and for sbt compile times), keeping `js` separate very well may be the best way to go.

sellout:
Does js need to depend on core? If so, it’s probably just some bits that need to be detangled, right? Like conversion to/from `Data`?

wemrysi:
@alissapajer 

1. I'd prefer to keep `csv` out of `effect` (do you mind the singular naming?) as I see the latter as being more focused on core, generic algebraic effects (adding `console` is fine as we hope to make that algebraic when time permits). `csv` could move to `main` as it is needed by both `repl` and and `web` if you want to get it out of `core`.
2. I just checked and `stacktrace` isn't used by anything, so let's delete it!
3. I'm fine with `LiftedOps` in `effect` or `foundation` with a small preference of just leaving it where it is as I'm not sure if we use it anywhere but for effects at the moment as it is tied to the sort of `Ops` pattern that we've adopted for those.

alissapajer:
@sellout right, we could put `js` parallel with `core` then, or even above `core` where it is now, parallel with `ejson`. I suppose then, in general, it's best to push each project "up" in the dependency graph as high as possible.

alissapajer:
@wemrysi ok cool, here's the plan then:

1. `effect` will be a singularly named new project, with `console` moved into it
2. `pillars` will be deleted, as it's been replaced with `effect` and `js`
3. `stacktrace` will be deleted (yay!)
4. `csv` will be moved into `main`, assuming all the dependencies work out, else it will stay in `core`
5. `LiftedOps` will stay in `effect` for now

alissapajer:
@wemrysi passing this back to you - I believe I have made all the changes we discussed!

wemrysi:
:shipit:

alissapajer:
@wemrysi cool, it's yours to merge!

wemrysi:
Ok, will do once I hear from Travis.

wemrysi:
@alissapajer Any idea if the rearrangement caused any of these [classloader errors](https://travis-ci.org/quasar-analytics/quasar/jobs/149007857#L5437)? Given the packages involved, I suspect they may be related to #1242 (or whatever introduced the spark dependencies), but I'm not sure and we have so few successful builds that it is hard to tell from the travis history.

I can see that as of [this build](https://travis-ci.org/quasar-analytics/quasar/jobs/148184255) things were ok, so must have happened sometime after that.

alissapajer:
@wemrysi so, one interesting thing is that I only see that `JarClassLoader: Warning` in one of the four builds (mongo 2.6). Two others passed (3.0 RO and 3.2), and the third hung (3.0) in the same way we've been seeing in master for a while. So either the errors are specific to that build, or it's a flukey as-yet unexplained thing that may or may not be related to this PR. I'm going to restart the build to see if we consistently get those errors or not, and additionally see if I can replicate the problem locally.

alissapajer:
Here's [a gist](https://gist.github.com/alissapajer/e21cbc60d224478ecf8c171ecdb0629f) with the classloader issue since travis overwrites previously build files when you restart a build :/

alissapajer:
Updates:

1. Reran the travis build. Builds #s 2,3,4 all passed. Build 1 hung at the end. No builds showed the classloader errors.
2. Locally running `sbt test`, `sbt oneJar`, and `sbt doc` all complete sucecssfully with no errors/warnings.

alissapajer:
@wemrysi I ran the travis build a third time. Builds 1-3 all failed with the message
```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```
which is how all the "hung" builds fail these days, with each build hanging at a different stage. Build 4 passed.

I have not been able to replicate the classloader issue locally or remotely, and I can't think of how this PR would cause that issue.

wemrysi:
@alissapajer You only see it in the `2.6` one because they are runtime errors/warnings when starting the server and the `2.6` build is the only build we attempt this in (it doesn't have anything to do with `2.6`, that just happens to be the first build which is the one we pick, arbitrarily). To reproduce it locally, you'd have to run the server.

Anyhow, it doesn't look like it has anything to do with this PR, so I'll merge and investigate.

alissapajer:
@wemrysi ah I see where we run `scripts/testJar` in `scripts/build` now.

PR #1306, opened 2016-07-29
Temorarily insert QS.LeftShift after LP.Read.

sellout:


core/src/main/scala/quasar/qscript/QScript.scala
```diff
@@ -493,7 +493,11 @@ class Transform[T[_[_]]: Recursive: Corecursive: FunctorT: EqualT: ShowT, F[_]:
   // TODO: Replace disjunction with validation.
   def lpToQScript: LogicalPlan[T[Target]] => PlannerError \/ TargetT = {
     case LogicalPlan.ReadF(path) =>
-      pathToProj(path).right
+      // TODO: Compilation of SQL² should insert a ShiftMap at each FROM,
+      //       however doing that would break the old Mongo backend, and we can
+      //       handle it here for now. But it should be moved to the SQL²
+      //       compiler when the old Mongo backend is replaced. (#1298)
```
sellout: This is the justification.

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -72,41 +71,69 @@ class QScriptSpec extends CompilerHelpers with ScalazMatchers {
...
     "convert a basic invoke" in {
       QueryFile.convertToQScript(
         math.Add(lpRead("/foo"), lpRead("/bar")).embed).toOption must
       equal(
-        QC.inj(
-          Map(RootR,
-            Free.roll(Add(
-              ProjectFieldR(HoleF, StrLit("foo")),
-              ProjectFieldR(HoleF, StrLit("bar")))))).embed.some)
+        TJ.inj(ThetaJoin(
```
alissapajer: @sellout so we have a `ThetaJoin` here now where we didn't used to. Shouldn't this query still be able to be written as a `Map`? I think this could be rewritten to remove index `0` from each of the `ZipMapKeys`.

Given that we can better optimize it, I'm not sure we should check in a test expectation that isn't what we actually want...

alissapajer: We would need a generalization of the work that's on my `LeftShift` optimization PR #1304

sellout: You’re right about the `ZipMapKeys`, but I didn’t want to redo your work. When your PR gets in, these tests will fail until the rewrite is applied to the expectations.

This does need a ThetaJoin, though – at least given the current assumption about LP.Read. We need to join the collection “/foo” with the collection “/bar”.

Now, once we can actually make the ShiftMap implied by each LP.Read explicit, we have to decide whether this test should represent the query ```select root.foo + root.bar from `/` as root ``` or the query ```select foo + bar from `/foo` cross join `/bar` ```, but currently this LP can only represent the latter.

sellout: I guess I could base this on your work now, though 😄

sellout: Oh right – it’s actually got the Join in between. That does complicate things.

But I do think that having tests that create correct (if not optimal) plans is important. Optimizations like this can be issues referenced in a TODO on the test. But we shouldn’t have to have perfect optimization before we can test anything. The main problem with the earlier tests is that they were _so_ complicated, we needed optimization to even comprehend them.

alissapajer: Yeah that's true, thinking about tests like that is thinking about them as documentation of the current state of the world.

PR #1307, opened 2016-07-29
spark-local manage file

rabbitonweb:


sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,158 @@
...
+    new (ManageFile ~> Free[S, ?]) {
+      def apply[A](mf: ManageFile[A]): Free[S, A] = mf match {
+        case Move(FileToFile(sf, df), semantics) => moveIt(sf, df, semantics)(moveFile _)
+        case Move(DirToDir(sd, dd), semantics) => moveIt(sd, dd,semantics)(moveDir _)
+        case Delete(path) => delete(path)
+        case TempFile(near) => tempFile(near)
+      }
+    }
+
+  def moveFile(src: APath, dst: APath):FileSystemError \/ Unit =
```
jedesah: This should probably be `AFile` since it's `moveFile`

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,158 @@
...
+    \/.fromTryCatchNonFatal(
+      Files.move(toNioPath(src), toNioPath(dst), StandardCopyOption.REPLACE_EXISTING)
+    ) .leftMap {
+      case e => pathErr(invalidPath(dst, e.getMessage()))
+    }.void
+
+  def moveDir(src: APath, dst: APath) =
```
jedesah: Same thing here, should probably be `ADir`.

jedesah: Maybe just a comment saying that this implementation does not yet pay attention to `MoveSemantics`

jedesah: Ah, nevermind, I think it's handled below.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,158 @@
...
+    }}
+
+
+  private def delete[S[_]](path: APath)(implicit
+    s0: Task :<: S
+  ): Free[S, FileSystemError \/ Unit] = {
+    val task: Task[FileSystemError \/ Unit] = Task.delay {
+      \/.fromTryCatchNonFatal(Files.delete(toNioPath(path)))
+        .leftMap {
+        case e: NoSuchFileException => pathErr(invalidPath(path, "File does not exist"))
```
jedesah: I think `pathNotFound` would be more appropriate here.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,158 @@
...
+        case e: DirectoryNotEmptyException => pathErr(invalidPath(path, "Directory is not empty"))
+        case e => pathErr(invalidPath(path, e.getMessage()))
```
jedesah: And probably `pathExists` here

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,158 @@
...
+    }
+    injectFT[Task, S].apply(task)
+  }
+
+  private def tempFile[S[_]](near: APath)(implicit
+    s0: Task :<: S
+  ): Free[S, FileSystemError \/ AFile] = {
+
+    def handleCreationError: Throwable => FileSystemError = {
+      case e: NoSuchFileException => pathErr(invalidPath(
```
jedesah: idem

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ManageFileSpec.scala
```diff
@@ -0,0 +1,568 @@
...
+         }
+        // when
+        withTempFile(createIt = withNoContent) { path =>
+          for {
+            existed <- exists(path)
+            result <- execute(program(path))
+            existsAfter <- exists(path)
+          } yield {
+            // then
+            result  must_== \/-(())
```
jedesah: We should use `must_=`

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ManageFileSpec.scala
```diff
@@ -0,0 +1,568 @@
...
+            result <- execute(program(dirPath))
+          } yield {
+            result must_== -\/((PathErr(InvalidPath(dirPath, "Directory is not empty"))))
+          }
+        }
+      }
+    }
+
+    "tempFile" should {
+      "create temp file neear existing path" in {
```
jedesah: typo

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,158 @@
...
+  def moveDir(src: APath, dst: APath) =
+    \/.fromTryCatchNonFatal{
+      val deleted = FileUtils.deleteDirectory(toNioPath(dst).toFile())
+      FileUtils.moveDirectory(toNioPath(src).toFile(), toNioPath(dst).toFile())
+    }
+      .leftMap {
+      case e => pathErr(invalidPath(dst, e.getMessage()))
+    }.void
+
+  private def moveIt[S[_]](src: APath, dst: APath, semantics: MoveSemantics)
```
jedesah: Would you be willing to put this function somewhere where I can make use of it for MarkLogic as well? If you extract the function to check for existence into a parameter to this function, then others could reuse this logic in order to support proper `MoveSemantics` in their own connectors.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
rabbitonweb: Not in this case because generic function `moveIt` requires `APath`.
`moveIt` is a function that deals with the boilerplate

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
rabbitonweb: Isn't `pathexists` an error when we were NOT expecting a thing but it actually exists? Here we actually are expecting a thing, and if we get some other exception then `NoSuchFileEx` or `DirectoryNotEmptyEx` then it means that something went wrong really bad. Is `Pathexists` the right way to communicate the problem?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
rabbitonweb: Happy to do it, can you guide me on when would be the best place to put it (in terms of project & package)

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
jedesah: Sorry, I think I was referring to line 119, but I accidentally clicked on 120.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
jedesah: Ya, I noticed that after, should have revised my comment.

What do you think of changing `moveIt` to `def ensureMoveSemantics(pathExists: APath => Task[Boolean], moveScenario: MoveScenario): OptionT[Task, FileSystemError]`, and putting it somewhere where other connectors can use it as well?

Then, we could `fold` on that `OptionT[Task, FileSystemError]` and call either `moveFile` or `moveDir` if the option is `None`. That way, we can also make the signatures of `moveFile` and `moveDir` more precise without causing any issues.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
jedesah: See my previous comment about changing the signature of this function a little to make it more composable. Probably `quasar.fs.impl` would be an adequate place to put this new function.

rabbitonweb:
@jedesah I've crafted `ensureMoveSemantics` - check if it's the implementation you've been thinking about.

Only thing that worries me is the fact that both `moveFile` and `moveDir` are side-effecting but they do not return a `Task`. They are being called WITHIN a `Task` but they do not return `Task` by themselves

rabbitonweb:
@jedesah I'm taking all back what I've written yesterday:
`moveFile` and `moveDir` return Task now
`ensureMoveSemantics` is moved to `core`

core/src/main/scala/quasar/fs/impl.scala
```diff
@@ -60,6 +65,29 @@ object impl {
     }
   }
 
+  def ensureMoveSemantics[S[_]](
+    dst: APath,
+    dstExists: APath => Task[Boolean],
+    semantics: MoveSemantics): OptionT[Task, FileSystemError] = {
+
+    def failBecauseExists = PathErr(InvalidPath(dst,
```
wemrysi: We already have `PathExists` and `PathNotFound` cases for `PathError`, I think they would be more appropriate here.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,146 @@
+/*
```
wemrysi: Double copyright.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  private def doesPathExist: APath => Task[Boolean] = path => Task.delay {
+    Files.exists(toNioPath(path))
+  }
+
+  private def moveFile(src: AFile, dst: AFile): Task[FileSystemError \/ Unit] = Task.delay {
+    \/.fromTryCatchNonFatal(
+      Files.move(toNioPath(src), toNioPath(dst), StandardCopyOption.REPLACE_EXISTING)
+    ) .leftMap {
+      case e => pathErr(invalidPath(dst, e.getMessage()))
```
wemrysi: I don't think this is an appropriate use of `InvalidPath`, it is intended to be for issues like unsupported characters, length, etc.

You might consider creating an error type for this implementation that can capture errors outside the scope of `FileSystemError`, or, if you don't need the granularity, you can use `PhysicalError`.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    def fileToMaybeAFile: Path => FileSystemError \/ AFile = f =>
+        toAFile(f) \/> (pathErr(invalidPath(near, s"Could not create temp file in dir $near")))
+    
+    val task: Task[FileSystemError \/ AFile] = Task.delay {
+      val posix = posixCodec.unsafePrintPath(near) + "/"
+      val prefix = "quasar"
+      val suffix = ".tmp"
+      \/.fromTryCatchNonFatal(Files.createTempFile(Paths.get(posix), prefix, suffix))
```
wemrysi: The `TempFile` term is somewhat poorly named as it only needs to return the _path_ to a temp file, it doesn't actually need to create it, though I'm not sure if there is any problem in doing so.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,146 @@
...
+  private def tempFile[S[_]](near: APath)(implicit
+    s0: Task :<: S
+  ): Free[S, FileSystemError \/ AFile] = {
+
+    def handleCreationError: Throwable => FileSystemError = {
+      case e: NoSuchFileException => pathErr(pathNotFound(near))
+      case e: FileAlreadyExistsException => pathErr(pathExists(near))
+      case e: FileSystemException if e.getMessage.contains("Not a directory") => pathErr(invalidPath(
+        near,
+        s"Provided $near is not a directory"))
```
wemrysi: You may find this is rather limiting as in current usage, `near` will usually be a file, not a directory.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
rabbitonweb: you serious? :) just path?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
rabbitonweb: Yes, I was informed about it yesterday. I'm fixing it already in a different branch where I tweak all implementations to make all `FileSystemTest`pass

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
rabbitonweb: Does it mean that not only temp file does not have to exist but also the entire path? 

For example let's say that given on file system exists only folder `/var/` and after running `TempFile` you get `/var/foo/bar/baz/0000.tmp` where obviously `foo` (and it's children) does not exist -> and it's still is a valid implementation?

wemrysi: Yes, as long as the path would be valid to use for a new file. See the tests in `ManageFilesSpec` that deal with temp files for the current expectations on their behavior.

rabbitonweb:
Closing. All addressed int https://github.com/quasar-analytics/quasar/pull/1369

PR #1319, opened 2016-08-01
Support for Mongo 3.2 pipeline ops with type safety

### LINKED ISSUE #1270: MongoDB 3.2: low-level support for new query capabilities
- [x] add `$lookup` (and `$sample`) to the `Workflow` ADT with type-safety to prevent using them when running against 2.6/3.0
- [x] detect MongoDB version and map specific versions to "classes" (core/2.6, 3.0, 3.2)
- [x] run the planner in the correct type

mossprescott:
Supercedes #1303 and closes #1270.

`WorkflowF` is replaced by a group of related types:
- `WorkflowOpCoreF`, containing all the old ops (aliased as `Workflow2_6F`)
- `WorkflowOp3_2F`, containing `$lookup` and `$sample`
- `WorkflowF`, which is a coproduct of the two types (aliased as `Workflow3_2F`)

WorkflowBuilder, `optimize` and the MongoDB planner are all parameterized so that either the 2.6 or 3.2 type may be used. When a query is to be evaluated, the backend version is interrogated to determine which type to use.

When the workflow is "crushed" to `WorkflowTask`, this information is lost and the executors always see the largest coproduct type.

In this commit, the new type is used when running against 3.2 or later, but the new ops are never actually generated and the only difference is whether ops are injected into the coproduct or not.

Some significant re-arranging was required, but it was mostly superficial:
- many functions over `Workflow` have become typeclasses. This results in several single-function typeclasses; some consolidation might be in order.
- ops no longer inherit from the traits `SingleSourceF`, `PipelineF`, and `ShapePreservingF`. Instead, instances of those traits are available as wrappers around the actual ops (which may be embeddind in a coproduct).
- there are now "convenience" extractors for each op type, which project from any suitable coproduct.
- most of the public API of WorkflowBuilder is now packaged in an `Ops` class which captures the workflow type `F`.
- the op case classes all got a `F` suffix to avoid a compiler warning about name collisions on case-insensitive file systems. This is our convention for fixpoint types anyway.


mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
```diff
@@ -0,0 +1,44 @@
...
+  case object `2.6` extends MongoQueryModel
+
+  /** Adds a few operators. */
+  case object `3.0` extends MongoQueryModel
+
+  /** Adds \$lookup and \$distinct, several new operators, and makes
+    * accumulation operators available in \$project. */
+  case object `3.2` extends MongoQueryModel
+
+  implicit val semigroup: Semigroup[MongoQueryModel] =
```
wemrysi: I seem to remember some mention of getting rid of this? That still a thing?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -114,185 +163,222 @@ object Workflow {
...
+  }
+  object Crush {
+    implicit def crushInjected[F[_]: Functor, G[_]: Functor](implicit I: F :<: G, CG: Crush[G]): Crush[F] =
+      new Crush[F] {
+        def crush(op: F[(Fix[F], (DocVar, WorkflowTask))]): (DocVar, WorkflowTask) =
+          CG.crush(I.inj(op.map { case (f, t) => (f.transCata(I.inj(_)), t) }))
+      }
+  }
+
+  implicit def crushWorkflowF(implicit I: WorkflowOpCoreF :<: WorkflowF): Crush[WorkflowF] = new Crush[WorkflowF] {
```
wemrysi: Any reason not to move this and other crush-related things to the `Crush` companion?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -348,55 +434,139 @@ object Workflow {
...
+    }
+
+    implicit def coproductRefs[F[_], G[_]](implicit RF: Refs[F], RG: Refs[G]): Refs[Coproduct[F, G, ?]] =
+      new Refs[Coproduct[F, G, ?]] {
+        def refs[A](op: Coproduct[F, G, A]): List[DocVar] =
+          op.run.fold(RF.refs, RG.refs)
+      }
+  }
+
+  implicit val refs2_6: Refs[WorkflowOpCoreF] = Refs.fromRewrite[WorkflowOpCoreF](rewriteRefs2_6)
```
wemrysi: Similarly here, can these be moved to the `Refs` companion?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -412,82 +582,261 @@ object Workflow {
...
+          CG.pipeline(_).map(_.fmap(ι, Inject[G, Coproduct[F, G, ?]])))
+
+      def shapePreserving[A](v: Coproduct[F, G, A]) =
+        v.run.fold(
+          CF.shapePreserving(_).map(_.fmap(ι, Inject[F, Coproduct[F, G, ?]])),
+          CG.shapePreserving(_).map(_.fmap(ι, Inject[G, Coproduct[F, G, ?]])))
+    }
+  }
+
+  implicit val classify2_6: Classify[WorkflowOpCoreF] = new Classify[WorkflowOpCoreF] {
```
wemrysi: And here.

wemrysi:
This looks good to merge. I would like to see `workflowop.scala` broken apart at some point, especially now that the new typeclasses exist, they should probably be in their own files. Maybe a `workflow` package is in order to hold all the things?

mossprescott:
Yes, I considered breaking up workflowop.scala into a package and pulling out the various types, which is all long overdue. But I'd like to make progress on the actual feature so I think I'll leave it alone for now.

mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
mossprescott: I forgot about that and I will go ahead and remove it.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
mossprescott: My general thought is that as these become "open" types (where we can add new types to the coproducts later on) we will want to implement the typeclasses with each type as we go. For now, with them all in the same file, it's somewhat academic.

wemrysi: Agreed, and it's fine to leave them where they are for the time being, though hopefully we can tidy up as more work is done here. These things have a tendency to stick around even though we know we're going to replace them Real Soon Now , so might as well keep them neat ;)

wemrysi:
@mossprescott Mind merging master into this, once we get a successful build I'll merge.

mossprescott:
@wemrysi: merged and looks good locally. Who knows what Travis will make of it, of course.

wemrysi:
Well, one clean build and the others appear to have stalled, not fail plus your local success, so I'll go ahead and merge.

PR #1323, opened 2016-08-02
clean up http4s thread pools after tests run

mossprescott:
- run the "servers" process which actually shuts down the server after
the "shutdown" effect is run
- also shutting down the client instance
- this reduces the number of leaked threads to a single pool (from one
pool per example)

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- clean up http4s thread pools in regression tests
--- a/it/src/test/scala/quasar/server/ServiceSpec.scala
+++ b/it/src/test/scala/quasar/server/ServiceSpec.scala
@@ -58,8 +58,11 @@ class ServiceSpec extends quasar.QuasarSpecification {
 
     (for {
       svc           <- service
-      (_, shutdown) <- Http4sUtils.startServers(port, svc).liftM[MainErrT]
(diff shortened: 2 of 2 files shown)


mossprescott:
This looks a bit better locally, but it looks like there is another source of orphaned threads. I'm looking into that, but in the meantime wanted to get a data point from Travis.

mossprescott:
This second fix seems to help the MongoDB driver to release its thread pools sooner when we create many driver instances in the regression tests, so I wanted to see it run in Travis as well. However:

It seems like the MongoDB driver just doesn't want to be used in this way. We probably need to rework the tests so that we create just one "client" instance for each database and reuse them across tests. Otherwise we're always going to have multiple thread pools in various stages of delayed shutdown hanging around.

There is another source of abandoned threads that seems to be more significant, which I haven't tracked down yet.

wemrysi:
LGTM, if we can get a clean build.

mossprescott:
Wanted to see how this would look after other recent fixes, so here goes another attempt at building this one…

mossprescott:
Hmm. At least on build failed in the test I'm monkeying with here, so I want to give this another look.

mossprescott:
Self-merging this, based on @wemrysi's earlier comment and the latest build.

PR #1325, opened 2016-08-03
Update specs to compatible version.

paulp:
Binary compatibility issues: specs 3.8.4 is built against
scalacheck 1.13.1 but we must use 1.12.5. Eric was kind enough
to publish a 3.8.4-against-1.12.5 artifact.


paulp:
I need this to get quasar and precog onto the same dependency set (but it also fixes a time bomb in the test dependencies.)

PR #1327, opened 2016-08-03
[WIP] CI improvements.

paulp:
- travis only runs main job on PRs and non-master pushes
- Added flakyTest test operator
- Added timestamps to build output


foundation/src/test/scala/quasar/QuasarSpecification.scala
```diff
@@ -20,10 +20,26 @@ import quasar.Predef._
...
+   *  fail regardless of whether it seems to pass or fail.
+   */
+  implicit class FlakyTest[T : AsResult](t: => T) {
+    import org.specs2.execute._
+    def flakyTest: Result = flakyTest("")
+    def flakyTest(m: String): Result = ResultExecution.execute(AsResult(t)) match {
+      case s: Success => s
+      case r          =>
+        val explain = if (m == "") "" else s" ($m)"
+        Skipped(s"${r.message}, but test is marked as flaky$explain", r.expected)
```
jedesah: perhaps `"... as flaky because $explain"`?

paulp: I was trying to avoid trailing off like `is marked as flaky because ` on the no-message case.

jedesah: ah, nvm, I didn't notice the line above.

paulp:
How do we feel about this as it stands? It'd be nice to have these changes, the timestamps especially.

sellout:
Other than the comment, I’m excited to get this in as well.

wemrysi:
If we're only going to run one of the jobs on a PR, maybe we should switch things around so the mongo 3.2 job does coverage, docs, etc and run it every time as it is likely a more important mongo to continually verify and also has the benefit of being faster at running the tests than 2.6.

paulp:
@wemrysi You're right.

jdegoes:
Ultimately, any solution that relies on running just one backend is not scalable. Nor do you have any guarantee the code works on other backends. We're using matrix builds, so it seems like the "problem" that testing one backend fixes is that our probability of a spurious failure is X, so because we run on all 4 backends in parallel, our probability of a spurious failure shoots up to 4 * X.

I'd suggest the root cause is the spurious failure, not the running of multiple backend tests in parallel via a matrix build.

paulp:
@sellout Hmm, do you mean that the folded sbt task succeeded, but something else caused the build to fail?

paulp:
@jdegoes I don't disagree - it's exactly the 4x chance of spurious failures that is being combated here. I intend this as a palliative, not a permanent measure. But it's a huge drag - including on the task of rooting out the spurious failures - to have the test suite regularly failing for reasons unrelated to the change being proposed.

sellout:
@paulp Yeah: https://travis-ci.org/quasar-analytics/quasar/jobs/149607501 – you can see it looks like the sbt task succeeded (which it did), but that fold (accidentally?) subsumes tasks that failed, which you can see when you unfold it.

paulp:
@jdegoes Also, maybe you didn't note that the tests ARE run everywhere, just not on every pull request. They're run everywhere when it's merged into master.

paulp:
@sellout That should be an earlier incarnation of this pull request. There are more separately folded steps now, and those definitely shouldn't show up folded together - if they do I'll fix it.

sellout:
@paulp That build is the latest on this PR (the last one that finished, just before merging from master). But glad we’re in agreement that it should be (or even is already) fixed.

paulp:
@sellout I see what I did, I improved the folding in the precog build and hadn't propagated that change back to here yet. One moment.

jdegoes:
@paulp Ah, I missed that. That's better, although it does create chaos and havoc when a broken pull request (or worse, multiple broken pull requests) are merged into master, the tests are run on all backends, and then everyone scrambles to figure out what went wrong and who is responsible.

The scenario we should all be planning for is the tests all being run in parallel by 20, 50, or 100 connectors. I'm not sure how you figure out which ones to run on the pull requests, but likely any decision other than "run all of them all the time" will lead to some pain and suffering.

Any process that makes it easier to ignore spurious failures will increase the incidence of spurious failures. i.e. the strategy of running only one connector on pull requests will increase the number and severity of spurious failures, because they'll only have to be dealt with once code is in master and people will just restart the build a bunch of times until it succeeds.

The strategy of running all connectors on all pull requests and not allowing restarts, for example, will eliminate spurious failures quickly and keep them out of the code base permanently.

That said, I ultimately don't care what decision is made here. We have stuff to build and that means making short/long-term tradeoffs, so if that's what we have to do, let's do it!

paulp:
@jdegoes Another benefit of testing 1/4 as much on PRs is that travis resource-constrains each account, so when you start a test run with four parallel jobs you will consume 4x as much of your available travis time, stalling all the PRs beyond your capacity. This becomes very significant on an active repository. You wind up having to go around manually cancelling flawed in-progress and pending jobs just to see the relevant tests run. We could leave travis and maybe we should, but it would include a significant time investment. In the meantime, it's desirable that people not have to babysit travis to see a green check mark in reasonable time.

jdegoes:
> Another benefit of testing 1/4 as much on PRs is that travis resource-constrains each account, so when you start a test run with four parallel jobs you will consume 4x as much of your available travis time, stalling all the PRs beyond your capacity. 

Sounds like a good argument for leaving Travis. Again, this is fine by me. It will increase incidence of spurious failures but make life easier in the short-term, which seems like a prudent tradeoff.

paulp:
It occurs to me we should also consider failing fast: aborting the job on travis as soon as anything fails. At least on PRs.

paulp:
Ugh, I don't know what this is all about. It almost looks like it was cancelled.
```
16:27:47.544573 Server started listening on port 43755
16:27:47.553079 Server started listening on port 58519
16:27:47.594531 Server started listening on port 33452
./scripts/build: line 77:  6443 Killed                  ./sbt ++$TRAVIS_SCALA_VERSION -batch "$@"
      6444 Done                    | ts '%H:%M:%.S'
```

paulp:
Oh, it was cancelled, by me. I thought I was looking at the latest, but no.

paulp:
First line of folded sections now gives an indication what's happening inside.

screen shot 2016-08-04 at 11 37 50 am

dwijnand:
@paulp 

> It occurs to me we should also consider failing fast: aborting the job on travis as soon as anything fails. At least on PRs.

Not exactly what you (rightly) desire, but you might find this useful: https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/

paulp:
I divided the main task into two jobs, the one responsible for code coverage and the one which runs oneJar and builds the docs. Hoping to reproduce [these green check marks](https://travis-ci.org/paulp/quasar/builds/150074146) here now.

paulp:
I'm going to open a new PR in the near future.

PR #1328, opened 2016-08-03
Refactor FileSystemDef monoid to not forget errors

wemrysi:
Now the monoid will only choose an alternate def if the first could not handle the arguments whereas previously we would try another if the first _did_ handle the arguments and failed.

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
+++ b/core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
@@ -16,53 +16,51 @@
 
 package quasar.fs.mount
 
-import quasar.Predef.{PartialFunction, String, Unit}
+import quasar.Predef.{None, Option, PartialFunction, String, Unit}
+import quasar.SKI.κ
 import quasar.EnvironmentError
 import quasar.fs.{FileSystem, FileSystemType}
 
 import scala.StringContext
+
 import scalaz._
+import scalaz.syntax.either._
+import scalaz.syntax.monadError._
 
 import FileSystemDef._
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
```diff
@@ -16,53 +16,51 @@
...
 
-  def fromPF[F[_]: Monad](
+  def fromPF[F[_]](
     pf: PartialFunction[FsCfg, DefErrT[F, DefinitionResult[F]]]
   ): FileSystemDef[F] =
-    FileSystemDef(cfg => pf.lift(cfg).getOrElse(Monoid[FileSystemDef[F]].zero.run(cfg)))
+    FileSystemDef(pf.lift)
 
-  implicit def fileSystemDefMonoid[F[_]](implicit F: Monad[F]): Monoid[FileSystemDef[F]] =
+  implicit def fileSystemDefMonoid[F[_]]: Monoid[FileSystemDef[F]] =
```
sellout: Is there a law check for this Monoid somewhere?

sellout:
This looks good, other than my one comment.

(Took this over, because @alissapajer is in the process of moving, and I want to see it get in ASAP)

core/src/main/scala/quasar/fs/mount/FileSystemDef.scala
wemrysi: There is not, as the automated one requires an `Equal[FsCfg => Option[DefinitionResult[F]]]`, which also implies an `Equal[FileSystem ~> F]`. I tried to get coverage on the functionality by moving the interesting code into `orElse` and testing that.

wemrysi:
@sellout Bah, added, thanks.

PR #1329, opened 2016-08-04
Topic/marklogic

jedesah:


wemrysi:
@jedesah Given our discussion yesterday, should probably close this and we can submit another with the other read/write approach.

Issue #1284, opened 2016-07-25
MongoDB 3.2: compile JOIN to $lookup 


mossprescott:
- [x] make the strategy for planning JOIN a function that's provided to the `workflow` stage of the planner
- [x] refactor the current implementation as the 2.6 strategy
- [x] implement a second strategy with `$lookup` for 3.2 and later

In the initial PR, I intend to handle only "trivial" equi-joins, where one of the sources is a simple `$read` and the condition on that side is a simple reference.

- [x] check for sources in same db
- [x] check for "right" source collection not sharded

mossprescott:
Here's a working, hand-written plan for the query `select zips2.city from zips join zips2 on zips._id = zips2._id` (aka the `plan simple join` test):

```javascript
db.zips.aggregate([
  { $project: { left: "$$ROOT" } }, 
  { $lookup: { from: "zips", localField: "_id", foreignField: "_id", as: "right" } }, 
  { $unwind: "$right" },
  { $project: { city: "$right.city" } }
])
```

PR #1330, opened 2016-08-04
[#1284] Support for simple joins with $lookup on MongoDB 3.2 and later

mossprescott:
Pull `join` out of WorkflowBuilder, and implement a pair of `JoinHandler`s with different requirements on the workflow type. If the plan's structure does not conform to the very simple pattern, then we fall back to map-reduce as before.

@jdegoes: This commit does not include any checking of the collections `$lookup` is applied to, so it will fail if the sources are in different databases, or if the collection being merged _in_ is sharded. These checks to prevent runtime errors are the next thing on my list. If you would rather not merge this code in a state where certain queries can fail, I will add that to this PR before it is merged. Otherwise, this is certainly useful as a demonstration of the benefit of `$lookup`.

A simple example joining the 29k-document `zips` dataset to itself runs at least 20x faster with `$lookup`:

```
💪 $ cd local3.0/quasar-test
💪 $ select a.city, b.state from zips as a join zips as b on a._id = b._id
MongoDB
db.zips.aggregate(...);
db.zips.mapReduce(...);
db.tmp.gen_ebcd9878_0.aggregate(...);

Query time: 4.2s
 city          | state |
---------------|-------|
 AGAWAM        | MA    |
...

💪 $ cd /local3.2/quasar-test
💪 $ select a.city, b.state from zips as a join zips as b on a._id = b._id
MongoDB
db.zips.aggregate(
  [
    { "$match": { "_id": { "$ne": null } } },
    { "$project": { "left": "$$ROOT" } },
    {
      "$lookup": {
        "from": "zips",
        "localField": "left._id",
        "foreignField": "_id",
        "as": "right"
      }
    },
    { "$unwind": "$right" },
    {
      "$project": {...}
    },
    { "$project": { "city": true, "state": true, "_id": false } }],
  { "allowDiskUse": true });


Query time: 0.1s
 city          | state |
---------------|-------|
 AGAWAM        | MA    |
...

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
@@ -1,0 +1,276 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TOOD: move to .planner package
+package quasar.physical.mongodb
+
+import quasar.Predef._
+import quasar.{TernaryFunc}
+import quasar.fp._
+import quasar.javascript._
+import quasar.jscore, jscore.{JsCore, JsFn}
+import quasar.std.StdLib._
+import quasar.physical.mongodb.accumulator._
+import quasar.physical.mongodb.expression._
+import Workflow._
+import WorkflowBuilder._
+
+import matryoshka._
+import scalaz._, Scalaz._
+
+final case class JoinHandler[F[_], G[_]](run: (TernaryFunc, JoinSource[F], JoinSource[F]) => G[WorkflowBuilder[F]]) {
+
+  def apply(tpe: TernaryFunc, left: JoinSource[F], right: JoinSource[F]): G[WorkflowBuilder[F]] =
+    run(tpe, left, right)
+}
+
+final case class JoinSource[F[_]](src: WorkflowBuilder[F], keys: List[WorkflowBuilder[F]], js: Option[List[JsFn]])
+
+object JoinHandler {
+
+  // FIXME: these have to match the names used in the logical plan. Should
+  //        change this to ensure left0/right0 are `Free` and pull the names
+  //        from those.
+  val LeftName: BsonField.Name = BsonField.Name("left")
+  val RightName: BsonField.Name = BsonField.Name("right")
+
+  /** When possible, plan a join using the more efficient \$lookup operator in
+    * the aggregation pipeline.
+    */
+  def pipeline[F[_]: Functor: Coalesce: Crush: Crystallize]
+    (implicit ev0: WorkflowOpCoreF :<: F, ev1: WorkflowOp3_2F :<: F, ev2: Show[WorkflowBuilder[F]])
+    : JoinHandler[F, OptionT[WorkflowBuilder.M, ?]] = JoinHandler({ (tpe, left, right) =>
+
+    val WB = WorkflowBuilder.Ops[F]
+
+    def lookup(
+      lSrc: WorkflowBuilder[F], lField: BsonField, lName: BsonField.Name,
+      rColl: Collection, rField: BsonField, rName: BsonField.Name) =
+    {
+      val left = WB.makeObject(lSrc, lName.asText)
+      val filtered = WB.filter(left,
+        List(ExprBuilder(left, \/-($var(DocField(lName \ lField))))),
+        { case List(f) => Selector.Doc(f -> Selector.Neq(Bson.Null)) })
+      workflow(filtered).map { case (left, _) =>
+        CollectionBuilder(
+          chain[Fix[F]](
+            left,
+            $lookup(rColl, lName \ lField, rField, rName),
+            $unwind(DocField(rName))),
+          Root(),
+          None)
+        }
+    }
+
+    (tpe, left, right) match {
+      case (set.InnerJoin, IsLookupInput(src, expr), IsLookupFrom(coll, field)) =>
+        lookup(src, expr, LeftName, coll, field, RightName).liftM[OptionT]
+
+      case (set.InnerJoin, IsLookupFrom(coll, field), IsLookupInput(src, expr)) =>
+        lookup(src, expr, RightName, coll, field, LeftName).liftM[OptionT]
+
+      case _ => OptionT.none
+    }
+  })
+
+  object IsLookupInput {
+    /** Matches a source which is suitable to be the "left" side of a
+      * \$lookup-based join; essentially, it needs to have a single key.
+      */
+    def unapply[F[_]](arg: JoinSource[F])(implicit ev: Equal[Fix[WorkflowBuilderF[F, ?]]])
+      : Option[(WorkflowBuilder[F], BsonField)] =
+      arg.keys match {
+        // TODO: generalize to handle more sources (notably ShapePreservingBuilders)
+        case List(Fix(ExprBuilderF(src, \/-($var(DocVar(_, Some(field))))))) if src ≟ arg.src =>
+          (src, field).some
+        case _ =>
+          None
+      }
+  }
+  object IsLookupFrom {
+    /** Matches a source which is suitable to be the "right" side of a
+      * \$lookup-based join; this has to be a \$project of a single field from
+      * a raw \$read.
+      */
+    def unapply[F[_]](arg: JoinSource[F])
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by jdegoes** [comment]
@mossprescott Let's get it merged now and patch up with a followup PR.

**Comment by mossprescott** [comment]
Re-assigned in the hope of getting it merged today. @wemrysi, if you're not comfortable giving this a thumbs up, @sellout said he'd be able to look at after his talk today.

File: mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala [hunk 1]
```diff
@@ -0,0 +1,276 @@
...
+import quasar.std.StdLib._
+import quasar.physical.mongodb.accumulator._
+import quasar.physical.mongodb.expression._
+import Workflow._
+import WorkflowBuilder._
+
+import matryoshka._
+import scalaz._, Scalaz._
+
+final case class JoinHandler[F[_], G[_]](run: (TernaryFunc, JoinSource[F], JoinSource[F]) => G[WorkflowBuilder[F]]) {
```
wemrysi: How about an alias for the first argument like `type JoinType = TernaryFunc`?

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
```diff
@@ -601,19 +592,23 @@ object Workflow {
...
+    def applyFieldName(name: BsonField): BsonField = {
+      applyVar(DocField(name)).deref.getOrElse(name) // TODO: Delete field if it's transformed away to nothing???
+    }
+
+    (op match {
+      case $LookupF(src, from, lf, ff, as) =>
+        // NB: rewrite only the source reference; the foreignField is not part of
+        // the workflow at this point
+        $LookupF(src, from, applyFieldName(lf), ff, as)
+    }).asInstanceOf[A]
```
wemrysi: This necessary because of pattern matching?

wemrysi:
A couple minor comments, otherwise LGTM.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
mossprescott: Not sure why scalac doesn't get this one right; I was just following the pattern from the parallel 2.6 implementation. But I will update this to remove the duplication and implement the other 3.2 case.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
mossprescott: Hmm. I feel like that's just hiding the problem, which is ironic considering that just today I added a type alias elsewhere for a similar reason.

It _would_ make this declaration clearer of course, … ah screw it, I just talked myself into it.

mossprescott:
Updated for those two comments, and merged from master in the hope of getting better builds.

@sellout: it'd still be great to get your eyes on this if you have a chance. Not as much of a rush, now that it's the weekend already.

mongodb/src/main/scala/quasar/physical/mongodb/workflowop.scala
paulp: You guys have never run into problems using `$` in identifiers? It's asking for trouble. For reasons never made clear to me the compiler doesn't warn you away, but it's a reserved character - the only metacharacter in an incredibly complicated mangling scheme which is supposed to be bidirectional but isn't, and that's before you start using dollar signs in the upper layer.

mongodb/src/main/scala/quasar/physical/mongodb/optimize/optimize.scala
```diff
@@ -279,7 +279,13 @@ package object optimize {
     def inlineGroupProjects[F[_]: Functor](g: $GroupF[Fix[F]])
       (implicit I: WorkflowOpCoreF :<: F)
       : Option[(Fix[F], Grouped, Reshape.Shape)] = {
-      val (rs, src) = g.src.para(collectShapes[F])
+      def collectShapesƒ: F[(Fix[F], (List[Reshape], Fix[F]))] => (List[Reshape], Fix[F]) =
```
sellout: Can you rewrite this type as `GAlgebra[(Fix[F], ?), F, (List[Reshape], Fix[F])]`? It makes it easier for someone unfamiliar to jump to an explanation of algebras/folds. And we don’t use the `ƒ` suffix anymore (that was basically an obscure way to communicate what the type does explicitly) – unless there’s a name conflict, but I _think_ you removed that(?).

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
```diff
@@ -0,0 +1,275 @@
...
+    /** Matches a source which is suitable to be the "right" side of a
+      * \$lookup-based join; this has to be a \$project of a single field from
+      * a raw \$read.
+      */
+    def unapply[F[_]](arg: JoinSource[F])
+      (implicit ev0: WorkflowOpCoreF :<: F, ev1: Equal[Fix[WorkflowBuilderF[F, ?]]])
+      : Option[(Collection, BsonField)] =
+      arg match {
+        case JoinSource(
+              Fix(CollectionBuilderF(Fix($read(coll)), Root(), None)),
```
sellout: This is temporary, right? Can you add a TODO indicating that? We should be able to create a temp collection when necessary.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
```diff
@@ -0,0 +1,275 @@
...
+              List(Fix(ExprBuilderF(src, \/-($var(DocVar(_, Some(field))))))),
+              _) if src ≟ arg.src =>
+          (coll, field).some
+        case _ =>
+          None
+      }
+  }
+
+  /** Plan an arbitrary join using only "core" operators, which always means a map-reduce. */
+  def mapReduce[F[_]: Functor: Coalesce: Crush: Crystallize]
```
sellout: I wish I could easily diff this against the old impl, but it at least looks very similiar, if not identical.

sellout:
I think this looks good, modulo terrible CI.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
mossprescott: Yes and no. I think something like this is always going to be the pattern for this simple case. We can add another case at some point which deals with writing a temporary collection when needed. That's going to be a little more involved because I don't think we currently have a way to represent that in `WorkflowF`.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
mossprescott: It's exactly the same code, just with the arguments packaged a bit differently.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
sellout: Well, we currently have `$foldLeft`, which is like this, but MapReduce-specific. Maybe that could be generalized? ¯\_(ツ)_/¯

mossprescott: Exactly. I think we need an op that says run _a_, write to a temp collection, then run _b_ which refers to that collection. But I think temp collections aren't materialized until the executor, so it affects WorkflowTask and the associated machinery.

In any case, there are lots of limitations here that can be addressed in one way or another and will be separately. Anyway I didn't feel like a TODO was needed because the desire to support more joins will drive that work.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
```diff
@@ -0,0 +1,275 @@
...
+  def pipeline[F[_]: Functor: Coalesce: Crush: Crystallize]
+    (implicit ev0: WorkflowOpCoreF :<: F, ev1: WorkflowOp3_2F :<: F, ev2: Show[WorkflowBuilder[F]])
+    : JoinHandler[F, OptionT[WorkflowBuilder.M, ?]] = JoinHandler({ (tpe, left, right) =>
+
+    val WB = WorkflowBuilder.Ops[F]
+
+    def lookup(
+      lSrc: WorkflowBuilder[F], lField: BsonField, lName: BsonField.Name,
+      rColl: Collection, rField: BsonField, rName: BsonField.Name) =
+    {
```
mossprescott: @sellout: this is quite restricted by requiring `BsonField` on both sides. I actually had an implementation which accepted any `Expr` on the left and it was nearly as simple but wrote uglier plans because it introduced a temp to hold the left value for the join, and we do not yet optimize that away.

Straightening that out is second-next on my list (after I deal with checking preconditions).

Furthermore, I'm not really happy with the way this projects `$$ROOT` under `left` when in many cases we could project just the fields that will be referenced later. This seems quite a bit trickier; to do it right I would need to capture the $lookup/$unwind in a WorkflowBuilder of some kind so that the shape would be known by the time it is crushed. As it is, calling `toCollectionBuilder` early like this gives away the game. Perhaps that sort of cleverness is best put off until the QScript connector.

mossprescott:
@estewei: with a successful build (after a couple of restarts) and the build hopefully getting more stable, I don't see any reason to keep holding this back. You ok with me merging it as soon as it builds one more time?

wemrysi:
Yep, fine by me.

mossprescott:
Build #1 failed in `testJar`. Restarted.

PR #1331, opened 2016-08-04
Add PostgreSQL ReadFile, WriteFile, and ManageFile interpreters

drostron:
Addresses: #1243, #1244, #1245 

NB: Path to table name mapping is preliminary and only intended to be capable enough to enable the next phases of PostgreSQL connector development. It should be revisited.

### DIFF
--- a/core/src/main/scala/quasar/fs/impl.scala
+++ b/core/src/main/scala/quasar/fs/impl.scala
@@ -31,7 +31,7 @@ object impl {
 
   def read[S[_], C](
     open: (AFile, ReadOpts) => Free[S,FileSystemError \/ C],
-    read: C => Free[S,FileSystemError \/ (C, Vector[Data])],
+    read: C => Free[S,FileSystemError \/ Vector[Data]],
     close: C => Free[S,Unit])(implicit
     cursors: KeyValueStore.Ops[ReadFile.ReadHandle, C, S],
     idGen: MonotonicSeq.Ops[S]
@@ -48,9 +48,7 @@ object impl {
       case ReadFile.Read(handle) =>
         (for {
           cursor <- cursors.get(handle).toRight(FileSystemError.unknownReadHandle(handle))
-          result <- EitherT(read(cursor))
-          (newCursor, data) = result
-          _ <- cursors.put(handle, newCursor).liftM[FileSystemErrT]
+          data <- EitherT(read(cursor))
         } yield data).run
 
       case ReadFile.Close(handle) =>
--- a/core/src/main/scala/quasar/fs/impl.scala
+++ b/core/src/main/scala/quasar/fs/impl.scala
@@ -17,13 +17,12 @@
 package quasar.fs
 
 import quasar.Predef._
-import quasar.fp.numeric._
-import quasar.fp.free._
 import quasar.Data
 import quasar.effect.{KeyValueStore, MonotonicSeq}
+import quasar.fp.free._
+import quasar.fp.numeric._
 
 import scalaz._, Scalaz._
-import scalaz.concurrent.Task
 import scalaz.stream.Process
 
 object impl {
@@ -63,19 +62,21 @@ object impl {
     }
   }
 
-  type ReadStream = Process[Task, FileSystemError \/ Vector[Data]]
+  type ReadStream[F[_]] = Process[F, FileSystemError \/ Vector[Data]]
 
-  def readFromProcess[S[_]](f: (AFile, ReadOpts) => FileSystemError \/ ReadStream)(
+  def readFromProcess[S[_], F[_]: Monad: Catchable](
+    f: (AFile, ReadOpts) => Free[S, FileSystemError \/ ReadStream[F]]
+  )(
     implicit
-      state: KeyValueStore.Ops[ReadFile.ReadHandle, ReadStream, S],
+      state: KeyValueStore.Ops[ReadFile.ReadHandle, ReadStream[F], S],
       idGen: MonotonicSeq.Ops[S],
-      S0: Task :<: S
+      S0: F :<: S
   ): ReadFile ~> Free[S, ?] =
     new (ReadFile ~> Free[S, ?]) {
       def apply[A](fa: ReadFile[A]): Free[S, A] = fa match {
         case ReadFile.Open(file, offset, limit) =>
           (for {
-            readStream <- EitherT(f(file, ReadOpts(offset, limit)).pure[Free[S,?]])
+            readStream <- EitherT(f(file, ReadOpts(offset, limit)))
             id <- idGen.next.liftM[FileSystemErrT]
             handle = ReadFile.ReadHandle(file, id)
             _ <- state.put(handle, readStream).liftM[FileSystemErrT]
@@ -95,7 +96,7 @@ object impl {
         case ReadFile.Close(handle) =>
           (for {
             stream <- state.get(handle)
-            _      <- injectFT[Task, S].apply(stream.kill.run).liftM[OptionT]
+            _      <- lift(stream.kill.run).into.liftM[OptionT]
             _      <- state.delete(handle).liftM[OptionT]
           } yield ()).run.void
       }
--- a/core/src/main/scala/quasar/fs/impl.scala
+++ b/core/src/main/scala/quasar/fs/impl.scala
@@ -18,10 +18,13 @@ package quasar.fs
 
 import quasar.Predef._
 import quasar.fp.numeric._
+import quasar.fp.free._
 import quasar.Data
 import quasar.effect.{KeyValueStore, MonotonicSeq}
 
 import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+import scalaz.stream.Process
 
 object impl {
 
@@ -29,7 +32,7 @@ object impl {
 
   def read[S[_], C](
     open: (AFile, ReadOpts) => Free[S,FileSystemError \/ C],
-    read: C => Free[S,FileSystemError \/ Vector[Data]],
+    read: C => Free[S,FileSystemError \/ (C, Vector[Data])],
     close: C => Free[S,Unit])(implicit
     cursors: KeyValueStore.Ops[ReadFile.ReadHandle, C, S],
     idGen: MonotonicSeq.Ops[S]
@@ -46,7 +49,9 @@ object impl {
       case ReadFile.Read(handle) =>
         (for {
           cursor <- cursors.get(handle).toRight(FileSystemError.unknownReadHandle(handle))
-          data <- EitherT(read(cursor))
+          result <- EitherT(read(cursor))
+          (newCursor, data) = result
+          _ <- cursors.put(handle, newCursor).liftM[FileSystemErrT]
         } yield data).run
 
       case ReadFile.Close(handle) =>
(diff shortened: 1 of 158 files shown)


postgresql/src/main/scala/quasar/physical/postgresql/fs/package.scala
```diff
@@ -16,25 +16,69 @@
...
+import quasar.fs._, ReadFile.ReadHandle, WriteFile.WriteHandle
+import quasar.fs.mount.{ConnectionUri, FileSystemDef}, FileSystemDef.DefErrT
 
+import doobie.imports._
 import scalaz._, Scalaz._
 import scalaz.concurrent.Task
 
+// NB [quasar.physical.postgresql]:
+// - The PostgreSQL Connector is incomplete and under development.
+// - Queries are constructed manually instead of using doobie's sql interpolator
```
mossprescott: Actually an overstatement, isn't it? :-P

drostron: Heh, true 😆. I suppose I'm staying bullish.

mossprescott:
I'll give this a closer look tomorrow or Monday, but other than looking for coding/SQL problems, I'm not exactly sure how to review this.

What's the standard of "good enough" for this at this point? It seems like this PR gives us a usable FileSystem for PostgreSQL. But I suppose without query capability we don't want to expose that as an available mount type. And if that's the case, is there much value in having this in the code base? Just so we keep it compiling? I wonder if it makes more sense to just put this whole connector on a branch and worry about merging it back in later on. Apologies if this was already discussed (especially if I was present for the discussion).

drostron:
Having this hang out in a branch vs master hasn't been discussed much yet, and I've wondered which is preferable since we're hitting the pause button on this connector for a bit. I'm ok with either. Opinions/votes?

wemrysi:
:+1: on the branch.

jdegoes:
It'll be dead in weeks if it's on the branch, and it has enough value, I want to work on this part-time if I'm able.

wedens:
Won't there be problems with effect interleaving and transactional semanctics due to using `ConnectionIO` in coproduct? You can achieve proper semantics by interpreting everything to `ConnectionIO` and then executing transaction.
Please tell me if I'm wrong.

drostron:
We've already pulled `ConnectionIO` to the edge in a downstream project, but since this is a WIP I chose postpone that work for a bit. I created an issue for it #1340.

mossprescott:
What's stopping us from running the `it` tests with this interpreter in CI? Obviously you would have to make all the query tests `pending`, but the rest should work, right? I assume you've run them locally, since you're updating them in this PR.

See q-a for Travis PostgreSQL setup.

postgresql/src/main/scala/quasar/physical/postgresql/fs/readfile.scala
```diff
@@ -0,0 +1,59 @@
...
+    impl.readFromProcess[S, ConnectionIO] { (file: AFile, readOpts: impl.ReadOpts) =>
+      (for {
+        dt <- dbTableFromPath(file)
+        _  <- EitherT(lift(tableExists(dt.table)).into.map(_
+                .either(())
+                .or(FileSystemError.pathErr(PathError.pathNotFound(file)))))
+      } yield {
+        val lim = readOpts.limit.map(lim => s"limit ${lim.unwrap}").orZero
+
+        val qStr = s"""select v from "${dt.table}" $lim offset ${readOpts.offset.unwrap}"""
```
mossprescott: `v`?

Ok so we're generating SQL strings directly. I'd still like to see something like 
```scala
def select(exprs: List[Ident], table: Ident, offset: Int, limit: Option[Int])

case class Ident(name: String)
object Ident {
  def sanitize(name: String): InvalidNameError \/ Ident = ???
}
```

Even if `sanitize` is pretty lame for now.

core/src/main/scala/quasar/fs/impl.scala
```diff
@@ -17,19 +17,21 @@
...
+import scalaz.stream.Process
 
 object impl {
 
   final case class ReadOpts(offset: Natural, limit: Option[Positive])
 
   def read[S[_], C](
     open: (AFile, ReadOpts) => Free[S,FileSystemError \/ C],
-    read: C => Free[S,FileSystemError \/ Vector[Data]],
+    read: C => Free[S,FileSystemError \/ (C, Vector[Data])],
```
mossprescott: I'm confused.

`C => Free[S, E \/ Data]` means "read a chunk statefully from this cursor".

`C => Process[S, Free[S, ?]. E \/ Data]` would mean "here's a cursor, give me a process that reads from it (statefully)", and maybe that would have been a better way to express what we had before, but maybe it conflicts with how we're using it.

Adding `C` to the return type lets you thread the state through the reading of chunks, so it's no longer (required to be) handled as a side effect. I guess that's more flexible. But I don't understand why there doesn't seem to be any code affected by this change or taking advantage of it. Is anyone actually calling this?

postgresql/src/main/scala/quasar/physical/postgresql/common.scala
```diff
@@ -0,0 +1,56 @@
...
+      Path.flatten(None, None, None, Some(_), Some(_), f)
+        .toIList.unite.uncons(
+          FileSystemError.pathErr(PathError.invalidPath(f, "no database specified")).left,
+          (h, t) => DbTable(h, t.intercalate("☠")).right)
+        .point[Free[S, ?]])
+
+  def tableExists[S[_]](tableName: String): ConnectionIO[Boolean] = {
+    val qStr =
+      s"""select table_name from information_schema.tables
+          where table_name = '$tableName'"""
```
mossprescott: Similar thoughts here. Actually, why not use doobie's parameterized query mechanism here?

postgresql/src/main/scala/quasar/physical/postgresql/common.scala
```diff
@@ -0,0 +1,56 @@
...
+
+     Query[HNil, String](qStr, none).toQuery0(HNil).list.map(_.nonEmpty)
+  }
+
+  def tablesWithPrefix(tableNamePrefix: String): ConnectionIO[List[String]] = {
+    val qStr =
+      s"""select table_name from information_schema.tables
+          where table_name like '$tableNamePrefix%'"""
```
mossprescott: Need to escape `LIKE`s special characters here (at least `%`, `_`, and `\`). See recent discussion here: https://github.com/slamdata/quasar-advanced/pull/120

postgresql/src/main/scala/quasar/physical/postgresql/fs/package.scala
```diff
@@ -16,25 +16,69 @@
...
+  type Eff[A]  = Coproduct[ConnectionIO, Eff0, A]
+
+  def interp[S[_]](
+      uri: ConnectionUri
+    )(implicit
+      S0: Task :<: S,
+      S1: PhysErr :<: S
+    ): Free[S, Free[Eff, ?] ~> Free[S, ?]] = {
+
+    val transactor = DriverManagerTransactor[Task]("org.postgresql.Driver", uri.value)
```
mossprescott: Probably worth at least a TODO here saying we need to think about connection pooling.

postgresql/src/main/scala/quasar/physical/postgresql/fs/queryfile.scala
```diff
@@ -0,0 +1,77 @@
...
+  ): Free[S, FileSystemError \/ Set[PathSegment]] =
+    (for {
+      dt  <- dbTableFromPath(dir)
+      r   <- lift(tablesWithPrefix(dt.table)).into.liftM[FileSystemErrT]
+      _   <- EitherT((
+               if (r.isEmpty) FileSystemError.pathErr(PathError.pathNotFound(dir)).left
+               else ().right
+             ).point[Free[S, ?]])
+    } yield r
+      .map(_.stripPrefix(dt.table).stripPrefix("☠").split("☠").toList)
```
mossprescott: Wouldn't it be as easy to map these table names back to paths and then do this on path segments? That would at least put the `☠` hack all in one place.

drostron:
If we think our CI build is in a place where running `it` tests for an additional connector makes sense them I'm all for it. I was running `quasar.fs.{ReadFilesSpec, WriteFilesSpec, ManageFilesSpec}` locally.

postgresql/src/main/scala/quasar/physical/postgresql/fs/readfile.scala
```diff
@@ -0,0 +1,59 @@
...
+        Query[HNil, String](qStr, none)
+          .toQuery0(HNil)
+          .process
+          .chunk(1024) // arbitrary size for the moment
+          .map(_.traverse(s => DataCodec.parse(s).leftMap(
```
mossprescott: So this assumes a field `v` containing JSON? OK if that's what we've got for the moment, but how about doing `select *` and then writing a "codec" `ResultSet => Data` (even if that codec just does `DataCodec.parse(rs.getString("v"))` for now)? If that's not simple (say, because doobie) then I think this deserves a comment saying what the impediment is.

postgresql/src/main/scala/quasar/physical/postgresql/fs/writefile.scala
```diff
@@ -0,0 +1,115 @@
...
+                  val qStr = s"""CREATE TABLE "${dt.table}" (v json)"""
+                  lift(Update[HNil](qStr, none).toUpdate0(HNil).run.void).into
+                }
+              ).liftM[FileSystemErrT]
+      i    <- MonotonicSeq.Ops[S].next.liftM[FileSystemErrT]
+      h    =  WriteHandle(file, i)
+      _    <- writeHandles.put(h, TableName(dt.table)).liftM[FileSystemErrT]
+    } yield h).run
+
+  def escape(json: String) = json.replace("'", "''")
```
mossprescott: That's the idea! 😆

postgresql/src/main/scala/quasar/physical/postgresql/fs/readfile.scala
drostron: Mostly just following along from the original doobie sql interpolation that was abandoned due to needing dynamic table names. I don't mind the more direct SQL strings but am open to methodizing them.

Sanitizing is important and I looked into it a bit, but unfortunately is not trivial to get right. We will need to have a form of it for sure. I'll create an issue in addition to the existing NB.

postgresql/src/main/scala/quasar/physical/postgresql/common.scala
```diff
@@ -0,0 +1,56 @@
...
+import pathy.Path
+import scalaz._, Scalaz._
+import shapeless.HNil
+
+object common {
+
+  final case class DbTable(db: String, table: String)
+
+  // TODO: Going with ☠ style path table names for the moment
+  def dbTableFromPath[S[_]](f: APath): FileSystemErrT[Free[S, ?], DbTable] =
```
mossprescott: I'd rather see this be just `PathErr \/ DbTable` and then have a lifted version if needed. `S[_]` with no constraints is a code smell to me.

postgresql/src/main/scala/quasar/physical/postgresql/common.scala
drostron: True, I'll update this one to the interpolator.

postgresql/src/main/scala/quasar/physical/postgresql/fs/package.scala
drostron: Agreed, been on my TODO list but deserves a transfer. I'll add it as an issue instead.

postgresql/src/main/scala/quasar/physical/postgresql/common.scala
drostron: Just convenience of use — the bare `\/` will be more reusable as well. @jedesah offered a similar comment awhile back when pairing.

core/src/main/scala/quasar/fs/impl.scala
drostron: This snuck in from a previous version before the switch to `readFromProcess`.

drostron: Oh right, this arrived with a merge from @jedesah's branch offering `readFromProcess`. It's not being used here and I don't believe this form is being leveraged by any underway connector development, so will revert it.

postgresql/src/main/scala/quasar/physical/postgresql/common.scala
drostron: I gave it a go, but the table name quoting causes trouble.

mossprescott: How so? Generally you don't need quotes when you use prepared statements, i.e. `where table_name = :t`.

drostron: Momentary mis-assumption. So yes, works fine.

mossprescott:
[Just noticed I never hit the submit button on this comment I wrote hours ago]

If the integration tests _can_ run, I think we should be running them. I don't think an additional build is going to kill us at this point, especially if it's not trying to run Mongo ;) What could go wrong?

drostron:
Sounds good. I'll look into enabling `it`.

postgresql/src/main/scala/quasar/physical/postgresql/fs/readfile.scala
drostron: Yes, this mapping is only meant to be temporary and was the minimal required to get things stood up and working. We'll want to revisit both reading and writing with data types and table schema in mind. I'll created an issue for it, #1363, and will add TODO comments.

drostron:
@mossprescott pushed a few commits addressing comments.

scripts/build
```diff
@@ -49,12 +49,6 @@ if [[ ${LOCAL_MONGODB:-} == "true" ]] ; then
     export QUASAR_MONGODB_READ_ONLY_INSERT="{\"mongodb\": {\"connectionUri\": \"mongodb://quasar-dbOwner:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\"}}"
   elif [[ "$QUASAR_MONGODB_RELEASE" == "3_2" ]]; then
     export QUASAR_MONGODB_3_2="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_2}\"}}"
-  else
```
mossprescott: Probably should change line 36 to default to "NONE" or just not set the var now that it's no longer being used that way. I don't think anyone uses the "all" option anymore.

mossprescott:
LGTM

scripts/build
drostron: Strict mode will insist on it being bound, so will switch to the more apt "NONE".

drostron:
@mossprescott marked an additional test as flaky. This should now be ready for a merge.

PR #1332, opened 2016-08-05
Skip pending regression tests

mossprescott:
This also makes skipping pending tests automatic for QuasarSpecification. We were previously using it inconsistently.

### DIFF
--- a/foundation/src/test/scala/quasar/QuasarSpecification.scala
+++ b/foundation/src/test/scala/quasar/QuasarSpecification.scala
@@ -22,7 +22,7 @@ import org.specs2.mutable._
 import org.specs2.scalaz.ScalazMatchers
 import scalaz._
 
-trait QuasarSpecification extends SpecificationLike with ScalazMatchers {
+trait QuasarSpecification extends SpecificationLike with ScalazMatchers with PendingWithAccurateCoverage {
   implicit class Specs2ScalazOps[A : Equal : Show](lhs: A) {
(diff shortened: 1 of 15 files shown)


mossprescott:
Hmm. One failure in DataServiceSpec and one "no output for 10 minutes" so far.

foundation/src/test/scala/quasar/QuasarSpecification.scala
```diff
@@ -22,7 +22,7 @@ import org.specs2.mutable._
 import org.specs2.scalaz.ScalazMatchers
 import scalaz._
 
-trait QuasarSpecification extends SpecificationLike with ScalazMatchers {
+trait QuasarSpecification extends SpecificationLike with ScalazMatchers with PendingWithAccurateCoverage {
```
paulp: I'd say if QuasarSpecification is going to extend PendingWithAccurateCoverage and everything is going to extend QuasarSpecification, then there's no benefit to PendingWithAccurateCoverage existing at all, and you could just move its contents into QuasarSpecification.

paulp:
Other than the one quibble, looks great to me.

foundation/src/test/scala/quasar/QuasarSpecification.scala
mossprescott: I wouldn't make the assumption that _everything_ will _always_ extend QuasarSpec., but on the other hand it's not that hard to change from one to the other, so :shrug:.

PR #1333, opened 2016-08-05
Switch to sbt-assembly for artifact packaging

wemrysi:
Also reenables the sparkcore project, but disables the offending readfile spec until we can get a handle on the resource leaks.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -130,8 +130,24 @@ lazy val publishSettings = Seq(
   )
 )
 
+lazy val assemblySettings = Seq(
+  test in assembly := {},
+
+  assemblyMergeStrategy in assembly := {
+    case PathList("META-INF", "io.netty.versions.properties") =>
+      MergeStrategy.last
+    case PathList("org", "apache", "hadoop", "yarn", xs @ _*) =>
+      MergeStrategy.last
+    case PathList("com", "google", "common", "base", xs @ _*) =>
+      MergeStrategy.last
+    case other =>
+      val defaultStrat = (assemblyMergeStrategy in assembly).value
+      defaultStrat(other)
+  }
+)
+
 // Build and publish a project, excluding its tests.
-lazy val commonSettings = buildSettings ++ publishSettings
+lazy val commonSettings = buildSettings ++ publishSettings ++ assemblySettings
 
 // Include to also publish a project's tests
 lazy val publishTestsSettings = Seq(
@@ -146,11 +162,10 @@ lazy val noPublishSettings = Seq(
   publishArtifact := false
 )
 
-lazy val oneJarSettings =
-  com.github.retronym.SbtOneJar.oneJarSettings ++
+lazy val githubReleaseSettings =
   githubSettings ++
   Seq(
-    GithubKeys.assets := { Seq(oneJar.value) },
+    GithubKeys.assets := { Seq(assembly.value) },
     GithubKeys.repoSlug := "quasar-analytics/quasar",
     releaseVersionFile := file("version.sbt"),
     releaseUseGlobalVersion := true,
@@ -274,7 +289,7 @@ lazy val repl = project
   .dependsOn(main % BothScopes)
   .settings(commonSettings)
   .settings(noPublishSettings)
-  .settings(oneJarSettings)
+  .settings(githubReleaseSettings)
   .settings(
     fork in run := true,
     connectInput in run := true,
@@ -286,7 +301,7 @@ lazy val web = project
   .dependsOn(main % BothScopes)
   .settings(commonSettings)
   .settings(publishTestsSettings)
-  .settings(oneJarSettings)
+  .settings(githubReleaseSettings)
   .settings(
     mainClass in Compile := Some("quasar.server.Server"),
     libraryDependencies ++= Dependencies.web)
--- a/build.sbt
+++ b/build.sbt
@@ -152,7 +152,6 @@ lazy val oneJarSettings =
   Seq(
     GithubKeys.assets := { Seq(oneJar.value) },
     GithubKeys.repoSlug := "quasar-analytics/quasar",
-
     releaseVersionFile := file("version.sbt"),
     releaseUseGlobalVersion := true,
     releaseProcess := Seq[ReleaseStep](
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -130,8 +130,24 @@ lazy val publishSettings = Seq(
...
+    case PathList("org", "apache", "hadoop", "yarn", xs @ _*) =>
+      MergeStrategy.last
+    case PathList("com", "google", "common", "base", xs @ _*) =>
+      MergeStrategy.last
+    case other =>
+      val defaultStrat = (assemblyMergeStrategy in assembly).value
+      defaultStrat(other)
+  }
+)
+
```
paulp: I don't mean to be a style pedant and I don't know if anyone agrees with my ideas on the matter anyway, but just for the record I think this sort of thing should look more like 
```scala
lazy val assemblySettings = Seq(
                   test in assembly := {},
  assemblyMergeStrategy in assembly := {
    case PathList("META-INF", "io.netty.versions.properties") => MergeStrategy.last
    case PathList("org", "apache", "hadoop", "yarn", xs @ _*) => MergeStrategy.last
    case PathList("com", "google", "common", "base", xs @ _*) => MergeStrategy.last
    case other                                                => (assemblyMergeStrategy in assembly).value apply other
  }
)
```
...for me when match logic is laid out like that it jumps off the page, whereas I find the downward zigzag style relatively laborious to internalize.

build.sbt
```diff
@@ -146,13 +162,10 @@ lazy val noPublishSettings = Seq(
...
 )
 
-lazy val oneJarSettings =
-  com.github.retronym.SbtOneJar.oneJarSettings ++
-  githubSettings ++
-  Seq(
-    GithubKeys.assets := { Seq(oneJar.value) },
+lazy val githubReleaseSettings =
+  githubSettings ++ Seq(
+    GithubKeys.assets := { Seq(assembly.value) },
```
paulp: These braces aren't doing anything. (Please understand such style notes as purely advisory.)

project/Dependencies.scala
```diff
@@ -57,6 +57,15 @@ object Dependencies {
     "io.netty"    % "netty-buffer"         % nettyVersion,
     "io.netty"    % "netty-handler"        % nettyVersion
   )
+  def sparkcore = Seq(
+    "io.netty"          %  "netty-all"                 % nettyVersion            % "compile",
+    ("org.apache.spark" %% "spark-core"                % "1.6.2"                 % "compile")
```
paulp: Since my overhaul of Dependencies.scala was merged into master I should point in that direction, particularly:
- "compile" is superfluous as it's the default configuration
- If you did need to specify it, the less stringly typed, the better - `Compile` and `Test` are named configurations
- All dependencies enumerated in the same file seems better to me
- That would also be a good place to encapsulate the exclude logic being applied here

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
```diff
@@ -46,7 +46,8 @@ class ReadFileSpec extends Specification with ScalaCheck  {
 
 
   "readfile" should {
-    "open - read chunk - close" in {
+    // FIXME: Skipping until we can avoid the local spark emulator which appears to leak resources, even after stop()
+    "open - read chunk - close" in skipped {
```
paulp: There's some way to pass a reason string to skipped, right? The FIXME comment might get more attention that way.

build.sbt
mossprescott: I would agree except when the lines get ridiculously long, which this one is flirting with.

In a case like this, how about collapsing those lines with `|`?

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
mossprescott: Yes. We've got code that looks like `Skipped(...)` (capital-S), and you might have to use it effectfully; both for reasons I don't entirely understand.

mossprescott:
Looks like `build` needs to be updated to run the right task.

Other than that, I don't see any problems with these changes, but I'm still not entirely sold on making this switch:

- Are there any possible ramifications of packaging other project's code in this way as opposed to oneJar. @jdegoes?
- Won't this affect the way downstream dependencies use the quasar jar? Or does it simply replace `-one-jar.jar` with `-assembly.jar` in a way that is mostly transparent?
- At a minimum, there are various references to the name `-one-jar.jar` that we'll have to track down and update.

jdegoes:
> Are there any possible ramifications of packaging other project's code in this way as opposed to oneJar. @jdegoes?

Not sure, but I don't think so.

> At a minimum, there are various references to the name -one-jar.jar that we'll have to track down and update.

Including in the installers project.

mossprescott:
Answering my own question (2) from above: I don't think this affects anyone who gets quasar via Ivy. And I think we can live with updating all of our own screwy downstream projects. So if John says it's probably OK, I'm not going to stand in the way.

project/Dependencies.scala
wemrysi: Ah yes, I meant to remove the "compile" config.

I'm not sure I understand what you mean by your last two points, mind clarifying? Or maybe you didn't notice this _is_ in `Dependencies.scala`?

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
wemrysi: Aye, I tried adding a message to the version I used, but didn't seem to be recognized, I'll try harder.

wemrysi:
We could name the assembled jar with the `-one-jar.jar` suffix if that helps. Otherwise we should probably name it something independent of the mechanism used to create it.

project/Dependencies.scala
paulp: Boy, github transitions so smoothly between files I only hear the "whoosh". Yeah, though I was looking at build.sbt.

paulp:
If my own experience is any guide, you want `aggregate in assembly := false` in the root project so it doesn't assemble giant jars in each subproject if you call `sbt assembly`. Then if we want `sbt assembly` to produce something sensible it can be forwarded, but I just referenced `web/assembly` and/or `repl/assembly` in the couple places it came up.

sellout:
I restarted the failing job because it had just hit #1335.

PR #1336, opened 2016-08-08
Run Spark-local tests only when env QUASAR_SPARK_LOCAL is set

rabbitonweb:
Ignore tests otherwise

note: This tests 'ReadFileSpec' is going to be deleted once all algebras are implemented and they will be tested using generic `FileSystemSpec` test. But for now we want to disable tests for Spark if somebody is not willing to run them

### DIFF
--- a/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
+++ b/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
@@ -101,7 +101,6 @@ class ReadFileSpec extends Specification with ScalaCheck  {
     slData <- OptionT(Task.now(uriData.asInstanceOf[Obj].value.get("sparklocal")))
     uri <- OptionT(Task.now(slData.asInstanceOf[Obj].value.get("connectionUri")))
   } yield {
-    println("************ ")
     val master = uri.asInstanceOf[Str].value
     val config = new SparkConf().setMaster(master).setAppName(this.getClass().getName())
     new SparkContext(config)
--- a/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
+++ b/sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
@@ -19,21 +19,23 @@ package quasar.physical.sparkcore.fs
 
 import quasar.Predef._
 import quasar.Data
+import quasar.Data._
+import quasar.console
 import quasar.fp.TaskRef
 import quasar.fp.numeric._
 import quasar.fp.free._
 import quasar.fs._
 import quasar.fs.ReadFile.ReadHandle
 import quasar.effect._
+import quasar.Data
+import quasar.DataCodec
+
+import java.io._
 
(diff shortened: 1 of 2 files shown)


alissapajer:
@rabbitonweb Since the `sparkcore` project is disabled in `build.sbt` (because it breaks the travis build), none of the tests in `sparkcore` are being run anymore. This puts PRs like this in an awkward position because we cannot verify (without locally enabling `sparkcore` and running tests) that they are ok.

What's the status on reenabling `sparkcore` in `build.sbt`?

alissapajer:
@rabbitonweb Actually, I just realized it's even worse. Files in the the path `quasar/sparkcore/` are not being compiled at all (`test` or `main`), verified by typing some gibberish :) So, I think we need to get project `sparkcore` back in action before we can merge any PRs to that project.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
```diff
@@ -92,8 +95,14 @@ class ReadFileSpec extends Specification with ScalaCheck  {
   private def inter(implicit sc: SparkContext): ReadFile ~> Task =
     readfile.interpret[Eff](local_readfile.input[Eff]) andThen foldMapNT[Eff, Task](run)
 
-  private def newSc(): SparkContext = {
-    val config = new SparkConf().setMaster("local[*]").setAppName(this.getClass().getName())
+  private def newSc(): OptionT[Task, SparkContext] = for {
+    uriStr <- console.readEnv("QUASAR_SPARK_LOCAL")
+    uriData <- OptionT(Task.now(DataCodec.parse(uriStr)(DataCodec.Precise).toOption))
+    slData <- OptionT(Task.now(uriData.asInstanceOf[Obj].value.get("sparklocal")))
```
alissapajer: As a general rule, I think we should avoid `asInstanceOf` at all costs, even in tests. Since you're already inside `Option` here, you could match on the `Data` result of `DataCodec.parse(uriStr)(DataCodec.Precise)`, and when the match hits `Obj(value)` return `Some(value)`, otherwise `None`. Or something along that vein.

alissapajer:
Also, just for the sake of documenting all of our build failures, build 3 failed with a "double free or corruption" error, which is something I haven't seen before. https://gist.github.com/alissapajer/80349a13a1cac233a5cea0a677c244c6#file-gistfile1-txt-L6998

wemrysi:
@alissapajer #1333 reenables the `sparkcore` project, dealing with the dependency issues but is blocked on CI (specifically the read-only build).

alissapajer:
@wemrysi ah ok great!

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ReadFileSpec.scala
alissapajer: @rabbitonweb so once this is unblocked by #1333, this is the only issue that will be preventing merge.

alissapajer:
@rabbitonweb let me know when you'd like me to look at this again

rabbitonweb:
This issue will be addressed in different PR

PR #1339, opened 2016-08-09
fix hang in regression tests when connection failure occurs

mossprescott:
This should turn "no output for 10 minutes" into an immediate failure with a record of the cause. We had a simple bug in the handling of those connection failures during the construction of interpreters for regression tests.

I'll move this to "in review" as soon as I get confirmation from Travis.

### DIFF
--- a/it/src/test/scala/quasar/TaskResource.scala
+++ b/it/src/test/scala/quasar/TaskResource.scala
@@ -48,6 +48,7 @@ object TaskResource {
     case object Start extends RsrcState
     case class Acquiring(id: Long) extends RsrcState
     case class Acquired(a: A) extends RsrcState
+    case class Failed(t: Throwable) extends RsrcState
 
     val signal = async.signalOf[RsrcState](Start)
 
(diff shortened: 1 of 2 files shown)


alissapajer:
@mossprescott so, we'd be expecting this branch to fail on jenkins as frequently as we've seen hangs (assuming there is only one cause to the hangs). I agree with merging this before we actually fix the failing tests because a build failed in this way on jenkins will tell us that it's hitting a known problem and will expose any other hangs that could be caused by something else.

mossprescott:
Looking good so far. #3 failed after 21 minutes with "Connection failed" errors; #1 got restarted somehow (I assume I hit the button accidentally), but seemed to be heading for success.

mossprescott:
#1 failed with an unrelated error ("jar failed to start"). #3 showed the fix is effective, so I think this is good to go.

paulp:
Can you please add a test which would avoid this scenario in the future? As in, a test for which success is immediate failure? (I don't care if it's in this PR, I too would like to see the tests passing again.)

it/src/test/scala/quasar/TaskResource.scala
```diff
@@ -62,11 +63,11 @@ object TaskResource {
       signal.continuous.take(1).runLast.flatMap(orFail)
 
     /** Asynchronously wait for the signal to contain the resource, which
-      * will have been acquired by another thread.
+      * will have been acquired by another thread, or fail if that thread failed.
       */
     val awaitAcquired: Task[A] =
       // NB: something effectful is hiding in the `discrete` process
-      Task.delay(signal.discrete.collect { case Acquired(a) => a }.take(1).runLast).join.flatMap(orFail)
+      Task.delay(signal.discrete.collect { case Acquired(a) => Task.now(a); case Failed(t) => Task.fail(t) }.take(1).runLast).join.flatMap(orFail).join
```
wemrysi: Very minor, but I think this would read better with the cases on their own lines.

mossprescott: Undoubtedly. I'll address it in another PR with some tests.

alissapajer:
I see this is merged :) but 👍 from me too!

PR #1343, opened 2016-08-10
Add plugin sbt-dependency-graph.

paulp:
It's super useful.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Add plugin sbt-dependency-graph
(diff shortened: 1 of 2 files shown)


sellout:
I use this, but I keep it in `~/.sbt`. Any particular reason for having it in the build?

paulp:
I'm generally super averse to adding anything to ~/.sbt because it makes the build non-reproducible, so I didn't really think of it. But I suppose this is the place to do it if there is one, an informational tool which isn't used directly in the non-interactive build. Fair enough.

PR #1346, opened 2016-08-10
Change some test options.

paulp:
Stop running a specification on failure. Right now we don't
have the sort of test suite where there is useful information
to be found beyond the first failure.

Show all test timings.

### DIFF
--- a/it/src/test/scala/quasar/fs/FileSystemTest.scala
+++ b/it/src/test/scala/quasar/fs/FileSystemTest.scala
@@ -50,7 +50,7 @@ abstract class FileSystemTest[S[_]](
   val fileSystems: Task[IList[FileSystemUT[S]]]
 ) extends quasar.QuasarSpecification {
 
-  args.report(showtimes = true)
+  sequential
 
   type F[A]      = Free[S, A]
   type FsTask[A] = FileSystemErrT[Task, A]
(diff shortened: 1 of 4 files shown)


sellout:
I think this would be useful for fail-fast in CI, but when making changes locally, the set of things that fails is often a hint as to what you broke. I would hate to have to fix tests one at a time without this insight.

paulp:
That's a good point, and easy to fix. One moment.

paulp:
That sucks, [this failure](https://travis-ci.org/paulp/quasar/jobs/151313645) is full of exactly the spew I had hoped this would limit. I fear that `stopOnFail=true` does nothing unless the test is also marked as sequential.

paulp:
I think the change in 6d56280047a will further stabilize CI.

it/src/test/scala/quasar/fs/FileSystemTest.scala
```diff
@@ -50,7 +50,7 @@ abstract class FileSystemTest[S[_]](
   val fileSystems: Task[IList[FileSystemUT[S]]]
 ) extends quasar.QuasarSpecification {
 
-  args.report(showtimes = true)
+  sequential
```
sellout: Do we want this sequential even when not in Travis?

paulp: I'd say so. Differences between local test environment and CI should be extremely well motivated. If it's important to test this code concurrently, then we should do that in a more focused way.

PR #1349, opened 2016-08-10
disable parallel execution for it tests in Travis

mossprescott:
This simple change seems to eliminate "connection timeout" failures (which until recently were hangs).

However, the cost is steep: roughly 20 minutes longer to run build #1 in Travis (just over an hour vs about 40 minutes, in my most recent comparison). The times for the other 3 builds seem to be affected very little.

I'm submitting this to get a build everyone can look at, and also to raise the question of whether this is yet bad enough to consider this fix.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- disable parallel execution of integration tests, which seems to contribute to build instability and doesn't help performance much anyway
--- a/build.sbt
+++ b/build.sbt
@@ -309,4 +309,5 @@ lazy val it = project
   // Configure various test tasks to run exclusively in the `ExclusiveTests` config.
   .settings(inConfig(ExclusiveTests)(Defaults.testTasks): _*)
   .settings(inConfig(ExclusiveTests)(exclusiveTasks(test, testOnly, testQuick)): _*)
+  .settings(parallelExecution in Test := !(sys.env contains "TRAVIS"))
   .enablePlugins(AutomateHeaderPlugin)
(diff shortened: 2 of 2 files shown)


jdegoes:
👍 20 minutes is a small price to pay, also we can and should just move to faster hardware.

paulp:
I think this should just be `parallelExecution in Test := false` and here is why. These are timings collected running `sbt test` on my macbook pro.
```
master:
  200.855 real, 871.625 user, 25.211 sys
  212.024 real, 909.128 user, 24.361 sys

parallelExecution := false
  210.263 real, 661.738 user, 20.491 sys
  211.927 real, 619.954 user, 19.660 sys
  209.187 real, 627.565 user, 19.453 sys
  215.456 real, 644.066 user, 19.250 sys
```
If it's any faster with parallel execution, it's *barely*. Yet it uses 50% more CPU. It would appear to be almost pure contention with negligible benefit.

paulp:
BTW partly it's the build number one is bonkers.
```
clean checkHeaders coverage test exclusive:test coverageReport "set coverageEnabled := false" oneJar doc
```
See, it builds everything with coverage, runs the test, publishes the reports... and now REBUILDS everything with no coverage, builds the jar, and builds the docs.

Meanwhile builds two through four also built everything with no coverage, and that work is just thrown away.

mossprescott:
On the first attempt, #1 ran in just 38 minutes, and #4 four failed with a random scalacheck failure. The other two seemed fine.

Restarting #4.

mossprescott:
@paulp: OK by me to abandon parallel builds for now. I had not tried it locally, and I'm not sure why it's sometimes _much_ slower in Travis and other times not (albeit this last build was done at the end of the work day, when Travis seems to be more responsive I've noticed.

And yes, #1 is building everything twice, which is dumb. When it wasn't slow, it didn't seem like a big deal. I think the answer to that is gitlab or some other tool that can fork sub-tasks.

In any case, I'll push the change to make it unconditional (plus a changelog entry), and if that build looks good then maybe we have something here.

paulp:
Those timings were with `parallelExecution in Test := false` in all projects, just for the record.

drostron:
This looks promising — 👍🏽 and merge upon successful build.

drostron:
@mossprescott, merge script says this needs a merge from master.

PR #1354, opened 2016-08-12
QS Optimization and Reorg

alissapajer:


core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -179,21 +179,35 @@ object MapFunc {
             Nullary[T2, T[CoEnv[A, MapFunc[T2, ?], ?]]](CommonEJson.inj(
               ejson.Bool[T2[EJson]](v1 ≟ v2)).embed).right).some
 
-        case ProjectIndex(Embed(StaticArrayPrefix(as)), Embed(CoEnv(\/-(Nullary(Embed(Inj(ejson.Int(index)))))))) =>
+        case ProjectIndex(Embed(StaticArrayPrefix(as)), Embed(CoEnv(\/-(Nullary(Embed(InjE(ejson.Int(index)))))))) =>
```
sellout: Oh, I think @mossprescott has a good solution for the `Inj` unapply. He implemented `unapply` for each trait, where `SomeTrait :<: G` (look in mongodb/workflowop.scala) for examples (WorkflowOpCoreF / WorkflowOp3_2F). Since we actually always need to know what we’re extracting anyway, it makes sense.

core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -179,21 +179,35 @@ object MapFunc {
...
               value
-            case Embed(CoEnv(\/-(Nullary(Embed(Inj(ejson.Map(m))))))) =>
+            case Embed(CoEnv(\/-(Nullary(Embed(InjE(ejson.Map(m))))))) =>
               m.find {
                 case (k, v) => k ≟ field
               }.map(p => CoEnv[A, MapFunc[T2, ?], T[CoEnv[A, MapFunc[T2, ?], ?]]](Nullary[T2, T[CoEnv[A, MapFunc[T2, ?], ?]]](p._2).right)).get
           }
+
+        // elide Nil on the left
+        case ConcatArrays(
```
sellout: We should have a parallel construct for `ConcatMaps`.

sellout:
Looks good other than those two comments.

core/src/main/scala/quasar/qscript/MapFunc.scala
alissapajer: Yes thanks for reminding me! Will add to the PR.

alissapajer:
@sellout okie I addressed the PR comments. It's back in your hands!

ejson/src/main/scala/quasar/ejson/package.scala
```diff
@@ -16,6 +16,8 @@
 
 package quasar
 
+import quasar.Predef._
```
alissapajer: huh this didn't show up as an unused import.... I'll delete it

ejson/src/main/scala/quasar/ejson/EJson.scala
```diff
@@ -30,6 +30,9 @@ final case class Str[A](value: String)     extends Common[A]
 final case class Dec[A](value: BigDecimal) extends Common[A]
 
 object Common {
+  def unapply[A](ej: EJson[A])(implicit C: Common :<: EJson): Option[Common[A]] =
+    C.prj(ej)
```
sellout: These should look like
```scala
def unapply[F[_], A](fa: F[A])(implicit C: Common :<: F): Option[Common[A]] =
  C.prj(fa)
```
particularly in this case, because we also want to be able to project `Common` from `Json`. But in general, we don’t want to make assumptions about the contexts a functor is used in.

alissapajer: @sellout good point, changed pushed.

alissapajer:
@sellout ok this should be g2g now!

sellout:
Tried to merge already – needs a CHANGELOG.

alissapajer:
@sellout dangit! :) ok changelog added :)

PR #1359, opened 2016-08-13
Split MongoDB Workflow into its own package.

sellout:
- rearranged some type classes to help with resolution

Otherwise, mostly updating imports.


mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
```diff
@@ -63,7 +63,7 @@ object JoinHandler {
       val filtered = WB.filter(left,
         List(ExprBuilder(left, \/-($var(DocField(lName \ lField))))),
         { case List(f) => Selector.Doc(f -> Selector.Neq(Bson.Null)) })
-      workflow(filtered).map { case (left, _) =>
+      WorkflowBuilder.workflow(filtered).map { case (left, _) =>
```
mossprescott: Maybe we should rename this to avoid the conflict? Never been a big fan anyway.

mongodb/src/main/scala/quasar/physical/mongodb/WorkflowExecutor.scala
```diff
@@ -328,7 +327,7 @@ private[mongodb] abstract class WorkflowExecutor[F[_]: Monad, C] {
 }
 
 object WorkflowExecutor {
-  import Workflow.{Crystallized, WorkflowF}
+  import quasar.physical.mongodb.workflow.{Crystallized, WorkflowF}
```
mossprescott: I think this is redundant now.

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
```diff
@@ -0,0 +1,691 @@
...
+  /** The type for workflows supporting the most advanced capabilities. */
+  type WorkflowF[A] = Workflow3_2F[A]
+  type Workflow = Fix[WorkflowF]
+
+  type FixOp[F[_]] = Fix[F] => Fix[F]
+
+  /** A "newtype" for ops that appear in pipelines, for use mostly after a
+    * workflow is constructed, with fixed type that can represent any workflow.
+    */
+  final case class PipelineOp(op: WorkflowF[Unit], bson: Bson.Doc) {
```
mossprescott: I'd like to see this come out, maybe even out of the `workflow` package, since it's using the types in an unusual way that could be confusing. It's used in the `Crush` implementation, which is arguably more about `WorkflowTask` than `Workflow` anyway. Maybe it could move to the `workflowtask` package?

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
```diff
@@ -0,0 +1,691 @@
...
+
+        def op = self.op
+        def rhs = self.rhs
+      }
+
+    // NB: needed because making A covariant breaks pattern-matching ("GADT skolem" errors)
+    override def widen[B >: A]: PipelineF[F, B] = reparent(src)
+  }
+
+  object PipelineF {
```
mossprescott: This is obviously dead code you might as well remove.

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
```diff
@@ -0,0 +1,691 @@
...
+
+      def render(v: T[F]) = v.project match {
+        case IsSource(s)       => s.op.render
+        case IsSingleSource(_) =>
+          NonTerminal("Chain" :: wfType, None, chain(v))
+        case $foldLeft(_, _) =>
+          NonTerminal("$FoldLeftF" :: wfType, None, v.children.map(render(_)))
+      }
+    }
+}
```
mossprescott: Down from 1900 lines in workflowop.scala :+1:

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
mossprescott: Actually, what is the proper place for `Crush`? It's not part of WorkflowOp, and it's not really part of WorkflowTask. But we already are using the PipelineOp/PipelineF ops in PipelineTask, so I guess I would vote for moving Crush to `workflowtask` as well.

mossprescott:
The type classes here are very _ad hoc_, basically they're just a translation of the signatures of a handful of functions that need to be generalized. There might be a way to collapse them into a smaller number of more meaningful types, but I had a go at it once already and there was no obvious way to do it. Even so, I think it's an improvement to break this out into separate files so it's easier to see what's there.

Pretty minor comments. LGTM unless you want to respond.

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
sellout: Detangling the circular dependency between workflow and workflowtask would be great. Moving `Crush` and this stuff over to workflowtask is _probably_ the right direction.

sellout: Fixed the rest, but punted on this for the time being. We really just need a way to avoid circular dependencies. That might be lihaoyi/acyclic, but it’d be nice to have something we can just enable globally.

PR #1361, opened 2016-08-14
[#1284] fall back to map-reduce when source collections are sharded or cross-DB

mossprescott:


mongodb/src/main/scala/quasar/physical/mongodb/collection.scala
```diff
@@ -21,22 +21,47 @@ import quasar.{RenderTree, Terminal}
...
+import scala.AnyVal
 import scala.util.parsing.combinator._
 
+import com.mongodb.MongoNamespace
 import scalaz._, Scalaz._
 import pathy.Path.{dir => pDir, file => pFile, _}
 
-final case class Collection(databaseName: String, collectionName: String) {
+
+final case class DatabaseName(value: String) extends AnyVal {
```
wemrysi: Nice, though I think some of the validation and escaping currently in `Collection` functions should be moved to the `DatabaseName` and `CollectionName` constructors, respectively, so that one can only create valid instances.

I realize we weren't doing this before (one could create a `Collection` with invalid database and collection names) so if it turns out to be too burdensome at the moment to have these new constructors be partial, then we can leave it as-is. They're still an improvement over `String`, maybe just not as much as they could be.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -2898,12 +2900,22 @@ class PlannerSpec extends quasar.QuasarSpecification with ScalaCheck with Compil
...
+    "plan simple join with sharded inputs" in {
+      val query = "select zips2.city from zips join zips2 on zips._id = zips2._id"
+      plan0(query,
+        MongoQueryModel.`3.2`,
+        c => Map(
+          collection("db", "zips") -> CollectionStatistics(10, 100, true),
+          collection("db", "zips2") -> CollectionStatistics(15, 150, true)).get(c)) must_==
+        plan2_6(query)
+    }
+
```
wemrysi: Do we need a test for joins in different dbs, or is that alreaady handled somewhere?

wemrysi:
@mossprescott A couple of comments, but LGTM in general.

mongodb/src/main/scala/quasar/physical/mongodb/collection.scala
mossprescott: That's a good point, but I'm reluctant to go down that path right now. I'll add a TODO.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
mossprescott: I'll add a test. I don't think this approach is super great, but it's better than nothing and I do think we need _some_ test of that code path.

mossprescott:
@wemrysi: Updated for your comments.

wemrysi:
@mossprescott Looks good, though now there are some conflicts with `master`. Once you've resolved those and we get a build, I'll merge.

PR #1367, opened 2016-08-17
Rework tests and project structures for performance.

paulp:
The major elements of this commit are:

1) Arrange the inter-project sbt dependencies in a less
linear way. Previously it was a line of projects each of
which depended on the previous project in both the Compile
and Test scopes. This meant that, for example, the tests in
the integration project could not even begin compiling until
the tests in the web project were completely compiled - even
though there is no meaningful dependency between them.

2) Reduce the specs2 implicit footprint. The common test trait
now inherits from the much more svelte specs SpecLike and not
the enormous SpecificationLike. Specs which need matchers not
included can easily import them from specs objects. Examples:

  import org.specs2.matcher.NumericMatchers._
  import org.specs2.matcher.MapMatchers._
  import org.specs2.matcher.AnyMatchers._
  import org.specs2.matcher.TraversableMatchers._

3) Reduce the specs2 bytecode footprint. By mixing the common
trait QuasarSpecification into an abstract class Qspec and then
inheriting from that class where possible, the total amount of
bytecode generated for all tests drops from 27MB to under 7MB.
That is not a typo.

4) Use inheritance much less. Inheritance should be a last
resort for code reuse. It's being used very casually and the
current performance issues are partly a consequence of that.


paulp:
There's one test in one spec commented out - if anyone can get that compiling that would be greatly appreciated. I spent quite a while picking through specs source code before giving up.

core/src/test/scala/MountServiceConfig.scala
```diff
@@ -0,0 +1,35 @@
...
+import quasar.sql.{ Sql, nullLiteral }
+
+/** This is used in both web and it tests.
+ *  If it's placed in the web test code, then it has to
+ *  depend on web's test configuration - which makes the
+ *  dependency chain linear. Placing it here allows
+ *  it to be independent of web.
+ */
+object MountServiceConfig {
+  def unsafeViewCfg(q: String, vars: (String, String)*): (Fix[Sql], Variables) =
```
mossprescott: Moving this to core makes sense, but it's now oddly named and placed, and silently turning errors into `null` doesn't seem appropriate for non-test code.

Suggest changing the signature to `WhateverError \/ (Fix[Sql], Variables)` and moving it to `MountConfig`, where there is already `viewCfgFromUri` which this code partially duplicates. Actually, we probably want `def viewConfigFromQuery(...): SomeError \/ MountConfig`, on top of which `viewConfigFromUri` can be implemented.

foundation/src/test/scala/quasar/decorators.scala
```diff
@@ -17,49 +17,27 @@
...
 trait PendingWithAccurateCoverage extends PendingUntilFixed {
-  def isCoverageRun: Boolean = quasar.build.BuildInfo.coverageEnabled
-  def isCIBuild: Boolean = quasar.build.BuildInfo.isCIBuild
-
-  /** Overrides the standard specs2 implicit. */
-  implicit def toPendingWithAccurateCoverage[T: AsResult](t: => T) = new PendingWithAccurateCoverage(t)
+  outer =>
 
-  class PendingWithAccurateCoverage[T: AsResult](t: => T) {
+  implicit class QuasarOpsForAsResultable[T: AsResult](t: => T) {
```
mossprescott: Why is this defined in two places?

it/src/test/scala/quasar/fs/FileSystemTest.scala
```diff
@@ -48,22 +48,23 @@ import scalaz.stream.Process
...
-  def fileSystemShould(examples: FileSystemUT[S] => Fragments): Unit =
+  def fileSystemShould(examples: FileSystemUT[S] => Fragment): Unit = (
     fileSystems.map(_ traverse_[Id] { fs =>
       s"${fs.name.name} FileSystem" should examples(fs)
 
       step(fs.close.unsafePerformSync)
 
       ()
     }).unsafePerformSync
+  )
```
mossprescott: Unnecessary parens, surely?

paulp: I don't believe in writing methods like 
```
if (foo)
  bar 
     baz
```
...because it's far too easy for a change to lose part of the expression. So anything which doesn't fit on one line I enclose in parentheses so that if it doesn't parse as a single expression it's a compile-time error.

foundation/src/test/scala/quasar/decorators.scala
paulp: Failed to remove that file.

mossprescott:
I pulled the branch and will have a look at that compile error.

I'm seeing what looks like more parallel compilation happening (as evidenced by my fans spinning up and by SBT's memory consumption going up.) It's not clear to me that will be a win in Travis, and the PR build certainly doesn't look like it's obviously improved. Any ideas about how to quantify the improvement?

core/src/test/scala/MountServiceConfig.scala
paulp: I can take a stab at this but it might be better off with someone else. I didn't want to change it at all, but for whatever reason there was a stronger wart enforcement at the new location (it wouldn't let me call .get on an option, as it was doing before.)

paulp:
I would say the PR build is very clearly improved. What times are you looking at? 

You can measure the time to finish `test:compile` locally - it's dropped by at least 30% for me.

core/src/test/scala/MountServiceConfig.scala
paulp: Also, MountConfig is in `core/src/main` but this is in `core/src/test`. That is, it's not non-test code. It's just in a location which doesn't cause `it/test` to depend on `web/test` since all `it/test` needed from `web/test` was this.

it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
```diff
@@ -339,7 +335,11 @@ class MongoDbFileSystemSpec
         }
       }
 
-      "Temp files" should {
+      /***
+
+      I can't figure out how to get this test compiling.
```
mossprescott: I was able to compile and run this by making two changes:

1. explicitly wrap the example in `Fragments(…)`
2. add a "dummy" example below `forEach`
3. also, I corrected the type of the expected value to match, but that didn't seem to be the problem.

This doesn't really make sense to me, but maybe it's a clue.

```scala
      "Temp files" >> {
        Fragments.foreach(Collection.DatabaseNameEscapes) { case (esc, _) => Fragments(

          s"be in the same database when db name contains '$esc'" >> {
            val pdir = rootDir </> dir(s"db${esc}name")

            runT(run)(for {
              tfile  <- manage.tempFile(pdir)
              dbName <- EitherT.fromDisjunction[manage.F](
                          Collection.dbNameFromPath(tfile).leftMap(pathErr(_)))
            } yield dbName).runEither must_== Collection.dbNameFromPath(pdir).leftMap(pathErr(_))
          }

        )}

        "foo" in { 0 must_== 0 }
      }
```

paulp: Thanks! I pushed an update.

it/src/test/scala/quasar/fs/FileSystemTest.scala
mossprescott: At the risk of getting into a religious discussion, these lines are enclosed in two pairs of delimiters, `... { ... ( ... )} ...`, so it's at least debatable.

More to the point, our mostly undocumented "standard" for formatting says not to put closing parens on their own lines like this, or at least that has overwhelmingly been our practice. (I notice the [guide](https://github.com/quasar-analytics/quasar/wiki/Scala%5BQuasar%5D) doesn't mention it, except by implication).

I suppose you could accomplish something similar with `{...}`, but I don't think that's great either.

Unless we want to adopt this as a rule, I'd rather not see it just in code *you* happened to touch ;)

In any case, anything in our code base returning `Unit` is obviously sketchy test code, and that's certainly true here. That's partially why it looks weird, with embedded blank lines, because it's doing a sequence of effectful steps. So I don't think it's all that important how we format this oddball code.

core/src/test/scala/MountServiceConfig.scala
mossprescott: In that case, how about adding a TODO that says it should be moved to MountConfig?

it/src/test/scala/quasar/fs/FileSystemTest.scala
paulp: If it were written like
```
def fileSystemShould(examples: FileSystemUT[S] => Fragment): Unit = fileSystems.map(_ traverse_[Id] { fs =>
...
}```
it would be a lot more tolerable. I realize that braces materialize, but it's that first step from the declaration to the next line which I have learned to fear. `{...}` doesn't accomplish the same thing because of semicolon inference. The parentheses draw attention to, and enforce, that the body is a single expression.

Indeed returning Unit makes it especially hard to avoiding touching it, although I didn't do it on that basis. I can probably hold myself back from this sort of change.

mossprescott:
Comparing with the most recent two builds on master:

   |  master | previous | PR
---|--------|---|---
.1 |  50:32 | 46:11 | 46:04
.2 |  20:40 | 20:21 | 18:52
.3 |  18:35 | 17:07 | 14:30
.4 |  20:47 | 20:43 | 22:50
.5 |  18:43 | 18:03 | 13:14
total | 2:09:07 | 2:02:25 | 1:55:30

So, yeah, I'd agree it's a bit faster. And making the build more parallel is an good thing regardless.

paulp:
There's a fair bit of variation in the travis times. If you take either the best time or the average time from a series of runs, and adjust for the shared constant factor (that is, everything except compiling the tests) I think you'll find it's way faster.

mossprescott:
That was my point exactly. Any conclusions you might draw from one or two Travis builds are going to be pretty shaky.

Locally I see it taking about a minute off a full compile, and possibly using more memory, but that's well worth it as long as Travis can handle it. So a nice improvement.

mossprescott:
@jedesah Unless you see anything else, I'll merge once the builds are done.

PR #1369, opened 2016-08-19
Enable file-integration tests

rabbitonweb:
Purpose of this PR is to finalize all `spark-local` FS implementations:

- [ ]  implement `ManageFile`
- [ ]  provide `spark-local`'s `FileSystemDef`
- [ ]  add ability to define a "root" folder when configuring connection
- [ ]  pass 100% for `ReadFilesSpec`, `WriteFilesSpec` and `ManageFilesSpec`


core/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -42,7 +42,9 @@ object FileSystemError {
     cause: Option[PhysicalError]
   ) extends FileSystemError
   final case class PathErr private (e: PathError)
-    extends FileSystemError
+      extends FileSystemError
+  final case class IOError private (reason: String)
```
wemrysi: I don't think this case is necessary as an implementation should use its own error type, or `PhysicalError` to communicate these sort of failures.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,121 @@
...
+  )
+  
+  private def tempFile[S[_]](near: APath)(implicit
+    s0: Task :<: S
+  ): Free[S, FileSystemError \/ AFile] = {
+
+    injectFT[Task, S].apply{
+      Task.delay {
+        val random = scala.util.Random.nextInt().toString
+        val parent = maybeFile(near)
```
wemrysi: See `pathy.Path.refineType` in order to avoid the `asInstanceOf`

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/package.scala
```diff
@@ -0,0 +1,95 @@
...
+
+      SparkFSDef(mapSNT[Eff, S](sInter), lift(Task.delay(sc.stop())).into[S])
+    }).into[S]
+  }
+
+  def definition[S[_]](implicit S0: Task :<: S, S1: PhysErr :<: S):
+      FileSystemDef[Free[S, ?]] =
+    FileSystemDef.fromPF {
+      case (FsType, uri) =>
+        val fsConf = parseUri(uri).getOrElse(SparkFSConf(new SparkConf().setMaster("local[*]"), rootDir </> dir("rabbit")))
```
wemrysi: If `parseUri` fails it should be an error rather than supplying this default config, which will confuse users.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/ManageFileSpec.scala
```diff
@@ -0,0 +1,512 @@
...
+import quasar.fs.FileSystemError._
+import quasar.fs._
+import quasar.fs.ManageFile.MoveSemantics
+
+import org.specs2.ScalaCheck
+import org.specs2.scalaz._
+import pathy.Path._
+import scalaz._, Scalaz._, concurrent.Task
+
+class ManageFileSpec extends QuasarSpecification with ScalaCheck with DisjunctionMatchers
```
wemrysi: There appears to be some duplication between this and `quasar.fs.ManageFilesSpec`, are all these tests necessary? Are there properties being tested that aren't covered by the general spec?

wemrysi:
This is looking good, though I wonder if we need all of the test code as much of it appears to duplicate the general specs in `quasar.fs` (correct me if I'm wrong). If there are spark-specific things being tested, that can't be covered by the general tests, then that's fine but it'd be good to avoid any duplication.

core/src/main/scala/quasar/fs/FileSystemError.scala
wemrysi: @rabbitonweb For an example of where we've done something similar, see `MongoDbIO#runF`.

rabbitonweb:
- [x] I've removed the default value (uri errors are now propagated to the caller)
- [x] I've used refineType to get rid of asInstanceOf

As for `moveFile` and `moveDir` I don't know 100% how that should be done. Currently those methods return `Task`. They would have to return `Free[S,]` where there would be `Task :<: S` and `PhysErr :<: S` - I can lift `Task` to that Free but I don't know how to handle situation when error happens within that Task. Would be nice to get help with that.

When it come to tests - well I guess we can think about deleting them. The problem with FS tests is that you need to implement everything upfront to have them running which does not help with incremental development. I'm ok with deleting them since all FS tests now pass.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/package.scala
```diff
@@ -0,0 +1,52 @@
...
+      posixCodec.parseAbsDir(rootPath)
+        .map { prefix =>
+        SparkFSConf(new SparkConf().setMaster(master), sandboxAbs(prefix))
+      }.fold(error(s"Could not extrat a path from $rootPath"))(_.right[DefinitionError])
+
+    uri.value.split('|').toList match {
+      case master :: prefixPath :: Nil => forge(master, prefixPath)
+      case _ =>
+        error("Missing master and prefixPath (seperated by |)" +
+          " e.g spark://host:port|/var/hadoop/")
```
wemrysi: Would it be possible to use the `path` component of the URI to represent the prefix, rather than the custom syntax? i.e. `spark://host:port/var/hadoop`

wemrysi:
I agree the fs specs aren't amenable to incremental development of the various components, something worth looking into eventually. My concern is mainly one of maintenance as all these tests have to be maintained and if we ever add or change the specs in `quasar.fs` in such a way that they disagree with these tests, we'd have to update these as well.

If there are any valuable cases here that aren't being tested in `quasar.fs` (i.e. something is underspecified, etc.), we can add them there. Any implementation-specific tests are still worth keeping, though.

wemrysi:
Also, looks like more conflicts in the assembly, this time between `slf4j-log4j12` and `logback`, maybe we can just keep `logback` and exclude the other?

rabbitonweb:
I did not add any dependency..

rabbitonweb:
@jedesah I've removed (https://github.com/quasar-analytics/quasar/pull/1369/commits/84e1e138d95b869f81d579778b6f9299cbfc8e99) logback dependency that you've added - because it had conflicts with slf4j on bytecode level. This was failing travis on this PR. Any particular reason why that dependency was explicitly added? Are you ok with me removing it?

jedesah:
@rabbitonweb I don't clearly remember why I added that explicit dependency. Might have added to resolve a conflict between two transitive dependencies at the time. For some reason I am under the impression `http4s` might used that lib. Or maybe it was as simple as using the latest (supposedly) binary compatible version in order to get rid of a result from running `dependencyUpdates`.

In any case, if it looks like everything works fine without it and it solves some problem, I can't think of a good reason to stop you from removing it. 😄

rabbitonweb:
@wemrysi all good

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  private def moveDir[S[_]](src: ADir, dst: ADir)(implicit
+    s0: Task :<: S,
+    s1: PhysErr :<: S
+  ): Free[S, Unit] = {
+
+    val move: Task[PhysicalError \/ Unit] = Task.delay {
+      val deleted = FileUtils.deleteDirectory(toNioPath(dst).toFile())
+      FileUtils.moveDirectory(toNioPath(src).toFile(), toNioPath(dst).toFile())
+    }.as(().right[PhysicalError])
```
wemrysi: Did you mean to `handle` any exceptions here like in `moveFile`?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/managefile.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private def delete[S[_]](path: APath)(implicit
+    s0: Task :<: S
+  ): Free[S, FileSystemError \/ Unit] = injectFT[Task, S].apply(
+    Task.delay {
+      \/.fromTryCatchNonFatal(FileUtils.forceDelete(toNioPath(path).toFile()))
+        .leftMap {
+        case e: FileNotFoundException => pathErr(pathNotFound(path))
+        case e => pathErr(invalidPath(path, e.getMessage()))
```
wemrysi: Looks like this needs the `PhysErr` treatment, similar to `moveFile`.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/package.scala
wemrysi: I still think it would be useful to avoid custom syntax, but if you want to do that in a later PR, maybe add a `TODO` here?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -0,0 +1,78 @@
...
+        if(directory.exists()) {
+          \/.fromTryCatchNonFatal{
+            directory.listFiles.toSet[File].map {
+              case file if file.isFile() => FileName(file.getName()).right[DirName]
+              case directory => DirName(directory.getName()).left[FileName]
+            }
+          }
+            .leftMap {
+            case e =>
+              pathErr(invalidPath(d, e.getMessage()))
```
wemrysi: This shouldn't be a `PathError`.

wemrysi:
@rabbitonweb What about the tests? Are there examples that aren't duplicating existing tests? Would they make sense to add to the general spec(s)? Can we delete the ones that duplicate existing specs?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/package.scala
rabbitonweb: I will add TODO here

rabbitonweb: The problem is it is not always in form of `spark://host:port` - it depends

rabbitonweb:
@wemrysi I've made changes u've noticed.

Few issues however

1. I have yet another class conflict for Spark (have to investigate that) - I love Spark for the dependency hell (every single time)
2. I will have to (maybe not in this PR) think how to finally run my tests on travis
3. I've changed `delete` to take advantage of `PhysicalError` but know I have two types of errors: one being handled by `FileSystemError`, other by `PhyicalError`. The implementation changes I've made compile and work BUT they look ... well I have a feeling there are look far from perfect and there is some FP sugars that could be done to it (or maybe I'm wrong) - could u verify if it is correct?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/package.scala
wemrysi: Ah ok, wasn't aware of that. If what I suggested won't work, then don't worry about it.

wemrysi:
@rabbitonweb I think the `delete` impl is ok, I agree it is a bit verbose, but don't see a quick way around it given that we have to catch both kinds of errors in `Task`.

Also, looks like there is one other comment to address in `queryfile.listContents`.

Looks like everything compiled in the most recent commit, does that mean the class conflict was solved?

rabbitonweb:
@wemrysi So this is ready. Spark tests are broken but it was not introduced in this PR. I will figure it out on a separate PR (this weekend, maybe even today). Can we merge?

wemrysi:
@rabbitonweb There is still an unaddressed comment in `listContents`, see my previous comment about that comment ;)

rabbitonweb:
I was 100% sure I've done that one... :) On it.

rabbitonweb:
@wemrysi Just realized, that listContest was largely refactored in other branch (succcessor of this PR). Could we merg this with current implementation and I will have that fixed in next PR ?

wemrysi:
@rabbitonweb Sure, that's fine. Looks like there are some conflicts with master, but I'll merge once those are resolved.

wemrysi:
@rabbitonweb Needs a `CHANGELOG` file. Include the text `[ci skip]` in the commit message to avoid kicking off another build.

rabbitonweb:
Done

PR #1370, opened 2016-08-19
Add a knownFiles transformation.

sellout:
Also,

- convert some uses of `Hole` back to `Unit` and
- add an optimization that collapses two Filters into one.


core/src/main/scala/quasar/qscript/StaticPath.scala
```diff
@@ -0,0 +1,141 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript
+
+import quasar.Planner.PlannerError
+import quasar.fp._
+
+import scala.inline
```
drostron: Ah simulacrum — wasn't clear to me initially. Should we import from `quasar.Predef` since it's available there?

project/Dependencies.scala
```diff
@@ -21,26 +21,26 @@ object Dependencies {
...
+    "com.slamdata"               %% "pathy-core"                %   pathyVersion,
+    "eu.timepit"                 %% "refined"                   %  refinedVersion,
     "com.chuusai"                %% "shapeless"                 % shapelessVersion,
-    "org.scalacheck"             %% "scalacheck"                % scalacheckVersion      % Test force(),
-    "com.github.mpilquist"       %% "simulacrum"                % "0.7.0"                % Test,
-    "org.typelevel"              %% "discipline"                % "0.5"                  % Test,
-    "org.specs2"                 %% "specs2-core"               % specsVersion           % Test,
-    "org.scalaz"                 %% "scalaz-scalacheck-binding" % scalazVersion          % Test,
-    "org.typelevel"              %% "shapeless-scalacheck"      % slcVersion             % Test,
-    "org.typelevel"              %% "scalaz-specs2"             % "0.4.0"                % Test
```
drostron: Why the centering of the 3rd column? Saw it appear a little while ago and have been curious.

drostron:
What's the motivation for switching some uses of `Hole` back to `Unit`?

alissapajer:
I second @drostron's question about `Hole` vs `Unit` (just wanted to put myself on the msg thread)

core/src/main/scala/quasar/qscript/StaticPath.scala
sellout: Yeah – I mix it up sometimes because Matryoshka doesn't have our Predef.

project/Dependencies.scala
sellout: ¯\_(ツ)_/¯ I think @paulp added it.

sellout:
So, we _should_ replace Unit in general, but with specific types with specific meanings. `qscript.Hole` has a specific meaning, of referring to the source. But there is another use of `Unit` in QScript – when we linearize it. But this is distinct from the source reference.

I can think of at least two other places in our code where we use Unit in the same was as QScript linearization

- we linearize `Workflow` in an aggregation pipeline, and
- Matryoshka converts arbtirary `T[F]` into `scala.Tree[F[Unit]]`.

The meaning here is something like “wish we could just delete this field in this case” or “we’re managing the recursion outside this structure” and we should have a name for it, but it’s not `qscript.Hole`.

drostron:
Explanation appreciated. Wished I could offer a reasonable name so we could introduce that instead of `Unit`, 🤔.

alissapajer:
@sellout @drostron Then could we give the `Unit` that are not `qscript.Hole` another name in this PR? How deep would that change have to go?

project/Dependencies.scala
paulp: It's just what my alignment macro does. Doesn't represent a strong preference.

drostron:
Thoughts on testing this transformation?

sellout:
Definitely needs testing. I’m adding that, but while I do, here are some general questions and discussion for connector authors:

Initially, I saw `knownFiles` as very controlled by the connector itself. At the very least, a connector needs to supply a way for QScript to determine if there is a file or directory at a particular location on a mount. So that opens it up to “what is the most helpful way to define this from a connector’s point of view?”

One option was letting connectors specify their own path representation. So rather than `final case class Read(path: pathy.AbsFile)`, we’d have `final case class Read[A](path: A)`. But that’s not feasible until we have mutual recursion in QScript.

Another option is that what exactly constitutes a `Read` node. The way most (all?) connectors work, the “natural” structure of a collection at “/foo/bar” is similar to `LeftShift(ProjectField(ProjectField(Root, "foo"), "bar"))`. So, a `Read` could represent that full structure. But then `ProjectField(ProjectField(Root, "foo"), "bar")` (no `LeftShift`[†]) would need to be represented as `Reduce(Read("/foo/bar"), List(UnshiftMap(…)))`, but a different connector might prefer to _not_ include the `LeftShift` in the `Read`, and then the latter case would be simply `Read("/foo/bar")`.

Finally, a connector could choose to not have `Read` at all, and just deal with doing “the right thing” as it processes the “bare” QScript, containing LeftShifts, projections, etc. that can refer to directories, files, or data depending on their location.

In any case, I feel like it’s hard to suss this out until people have really used QScript. So, I expect knownPaths to change in the next month. But if people have thoughts now, I’m more than happy to hear them.

[†]: In MRA, projections affect the provenance, but in a way that the dataset stays the same size. So, with there being exactly one `Root` object, each `ProjectField` drills down one level, maintaining the singleton set. So, when you get to a collection, you don’t have a set of records, but a single object, with an entry for each record, where the keys are the ids of the record, so you need a `LeftShift` after the final projection to get the traditional notion of of a collection as a set of records.

sellout:
@alissapajer I don’t think the non-Unit change would be deep. I’m happy to change it in this PR (I’d probably simultaneously change it in WorkflowTask, then later propagate it up to Matryoshka). @drostron suggested something along the lines of `ExternallyManaged` (with the single object maybe being called `Extern`?). That seems reasonable enough.

drostron:
Expecting it to change over the next month is reasonable. Happy there will be a test or two providing a hint on usage.

wemrysi:
@sellout Something that would be useful for incremental implementation, if possible, would be a transformation from `LogicalPlan` that just directly maps `LogicalPlan.ReadF` to `qscript.Read`. That'd at least allow us to sidestep this issue initially and then add support for arbitrarily pathing into objects within files.

This probably would cease to be possible once we start compiling `SQL^2` directly to `QScript` (as connectors won't be able to choose how to interpret `LogicalPlan`, but might be helpful until then. One downside might be that it will take longer to start compiling directly to `QScript` because of this.

sellout:
Updated this. Probably better to look at the whole thing, and not just the changes.

I’ll merge in master as soon as I get a chance.

core/src/main/scala/quasar/qscript/ConvertPath.scala
```diff
@@ -0,0 +1,93 @@
...
+      (AbsDir[Sandboxed], String) => OptionT[M, AbsDir[Sandboxed] \/ AbsFile[Sandboxed]] =
+    (dir, name) => listContents(dir).toOption >>=
+      (cont => OptionT((cont.find(_.fold(_.value ≟ name, _.value ≟ name)) ∘
+        (_.bimap(dir </> dir1(_), dir </> file1(_)))).point[M]))
+}
+
+object ConvertPath {
+  type Aux[T[_[_]], F[_], G[_]] = ConvertPath[T, F] { type H[A] = G[A] }
+
+  type ListContents[M[_]] = AbsDir[Sandboxed] => EitherT[M, FileSystemError, Set[PathSegment]]
```
wemrysi: You could use the slightly more terse `ADir` and `AFile` aliases since you're importing `quasar.fs` already (I only mention as I seem to remember you asking about something like this in slack).

core/src/main/scala/quasar/qscript/ConvertPath.scala
```diff
@@ -0,0 +1,93 @@
...
+  implicit def projectBucket[T[_[_]]: Recursive, G[_]: Functor](
+    implicit R: Const[Read, ?] :<: G, PB: ProjectBucket[T, ?] :<: G):
+      ConvertPath.Aux[T, ProjectBucket[T, ?], G] =
+    new ConvertPath[T, ProjectBucket[T, ?]] {
+      type H[A] = G[A]
+
+      def convertPath[M[_]: Monad](f: ListContents[M]): StaticPathTransformation[T, M, ProjectBucket[T, ?], G] = {
+        case x @ BucketField(src, _, StrLit(str)) =>
+          Recursive[T].project[Pathed[G, ?]](src.copoint).traverse(e => e.run.fold(
+            dir => fileType(f).apply(dir, str).map(df => CoEnv(df.map(file => R.inj(Const[Read, T[Pathed[G, ?]]](Read(file)))))).toRight(FileSystemError.pathErr(PathError.invalidPath(dir, "has no entry named “$str”."))),
```
wemrysi: Maybe "segment" instead of "entry"?

core/src/main/scala/quasar/qscript/StaticPath.scala
```diff
@@ -0,0 +1,174 @@
...
+          t.foldRight(
+            h)(
+            (elem, acc) => SP.inj(Union(DE.inj(Const[DeadEnd, IT[G]](Root)).embed, elem.embed.cata((g: G[Free[QScriptTotal[IT, ?], Hole]]) => Free.roll[QScriptTotal[IT, ?], Hole](FI.inj(g))), acc.embed.cata((g: G[Free[QScriptTotal[IT, ?], Hole]]) => Free.roll[QScriptTotal[IT, ?], Hole](FI.inj(g)))))).right
+        case Nil => FileSystemError.readFailed("Nil", "found no files").left
+      })
+    }
+
+  def toRead[M[_]: Monad, F[_]: Traverse, G[_]: Functor](
+    g: ConvertPath.ListContents[M])(
+    implicit TC: Corecursive[IT],
```
wemrysi: You may have won the prize for most implicit arguments :wink:

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -329,49 +348,94 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
+    * `f` takes QScript representing a _potential_ path to a file, converts
+    * `Root` and its children to path, with the operations post-file remaining.
+    */
+  def pathify[M[_]: Monad, F[_]: Traverse](
+    g: ConvertPath.ListContents[M])(
+    implicit FS: StaticPath.Aux[T, F],
+             DE: Const[DeadEnd, ?] :<: F,
+             F: Pathable[T, ?] :<: F,
+             QC: QScriptCore[T, ?] :<: F,
+             FI: F :<: QScriptTotal[T, ?],
```
wemrysi: Out of curiousity, will we be able to elide this constraint once we have mutual-recursion? Took me awhile to trace its origin, but appears to be due to `FreeQS`, no?

sellout: Yep. That is exactly correct.

core/src/main/scala/quasar/qscript/StaticPath.scala
sellout: As an aside, the implicit `~>` (which is not a type class, so shouldn’t be implicit) should actually be `:<:`, but because `:<:` doesn’t compose, we needed to fall back to `~>` where composition is necessary.

core/src/main/scala/quasar/qscript/package.scala
```diff
@@ -142,3 +186,20 @@ package object qscript {
     def apply[A](fa: EnvT[E, F, A]): F[A] = fa.lower
   }
 }
+
+sealed trait QScriptInstances extends LowPriorityQScriptInstances {
+  implicit def composedTraverse[F[_]: Traverse, G[_]: Traverse]:
```
wemrysi: Have you considered using a local instance of `Traverse[F].compose[G]` (similarly for `Functor`) where needed as opposed to these? I just wonder if these instances will be selected when they aren't desired or cause other issues (I haven't thought through all the implications, it may be fine)?

If nothing else, you could replace the implementation with the above.

wemrysi:
This version looks like it should be easy to use by backends, nice work. Nothing major from me other than a few minor comments.

core/src/main/scala/quasar/qscript/ConvertPath.scala
sellout: This is actually a premature error (whoops). In the case of a path like `/foo/*/baz`, we don’t want to error just because `/foo/q/` doesn’t have a `baz` file in it. As long as there is a `/foo/a/baz` or `/foo/b/baz` or whatever, we’re fine. And the empty set case (when no directory has a `baz`) is handled in StaticPath.

core/src/main/scala/quasar/qscript/package.scala
sellout: Yeah, I think you’re right. I made these because I had seen them in Cats, but I think they’re actually a bad idea.

sellout:
Oh yeah, that last commit also adds the version of `convertToQScript` that takes advantage of `QueryFile.Ops.ls` (i.e., knownFiles for free).

drostron:
🎊  👍🏽

Issue #1290, opened 2016-07-26
Basic implementation of ManageFile for MarkLogic


jedesah:
Only what is necessary to get the Regression tests to run.

Issue #1289, opened 2016-07-26
Basic implementation of ReadFile for MarkLogic


jedesah:
Since we are writing everything in the form of JSON (see #1288) for now. Simply expect the data to be JSON.

PR #1373, opened 2016-08-22
Non-queryable MarkLogic FileSystem

### LINKED ISSUE #1288: Basic implementation of WriteFile for MarkLogic
Write `quasar.Data` to MarkLogic in the form of JSON

wemrysi:
Initial `FileSystem` implementation for MarkLogic 8. All non-query related filesystem functionality is implemented and all non-`QueryRegression` filesystem tests are passing.

As this represents more of an internal milestone for MarkLogic connector development, there are still various `TODO`s and lots of `???` in the code in the places that aren't done or may change as the `QScript` translation proceeds as well as our experience with MarkLogic improves our understanding. That said, it seemed like a logical place to merge to master for the purposes of integration.

Fixes #1288 #1289 #1290


wemrysi:
@jedesah You'll probably want to take a look as well as this contains all the changes to our shared topic branch since we last merged.

marklogic/src/main/scala/quasar/physical/marklogic/fs/managefile.scala
```diff
@@ -0,0 +1,108 @@
...
+        case TempFile(path)            => tempFile(path)
+      }
+
+      def move(scenario: MoveScenario, semantics: MoveSemantics): Free[S, FileSystemError \/ Unit] =
+        lift(scenario match {
+          case MoveScenario.FileToFile(src, dst) => moveFile(src, dst, semantics)
+          case MoveScenario.DirToDir(src, dst)   => moveDir(src, dst, semantics)
+        }).into[S]
+
+      def checkMoveSemantics(
```
jedesah: Did @rabbitonweb end up contributing a general purpose function for this given a function to verify existence of an `APath` in one of his Spark PRs?

marklogic/src/main/scala/quasar/physical/marklogic/fs/managefile.scala
```diff
@@ -0,0 +1,108 @@
...
+        src: APath, dst: APath, sem: MoveSemantics
```
jedesah: Is `src` really needed here? Doesn't look like it's been used.

marklogic/src/main/scala/quasar/physical/marklogic/fs/managefile.scala
```diff
@@ -0,0 +1,108 @@
...
+          ops.ls(src0).flatMap(_.traverse_(_.fold(
+            d => moveContents(src0 </> dir1(d), dst0 </> dir1(d)),
+            f => ops.moveFile(src0 </> file1(f), dst0 </> file1(f)))))
+
+        def doMove =
+          checkMoveSemantics(src, dst, sem)        *>
+          moveContents(src, dst).liftM[FileSystemErrT] *>
+          ops.deleteDir(src).liftM[FileSystemErrT]
+
+        ops.exists(src).ifM(
```
jedesah: Random observation, it looks like there are a lot of these `ifM` with the same error message. Could introduce a `doIfExists` that takes care of the error part. Perhaps not worth it though or too much indirection.

marklogic/src/main/scala/quasar/physical/marklogic/fs/ops.scala
```diff
@@ -0,0 +1,240 @@
...
+  ): Free[S, Vector[FileSystemError]] = {
+    val dstDir = asDir(dstFile)
+
+    val createOptions = {
+      val copts = new ContentCreateOptions()
+      copts.setFormatJson()
+      copts
+    }
+
+    val seqFmt = {
```
jedesah: The purpose of this is not obvious to me, perhaps it warrants a comment.

marklogic/src/main/scala/quasar/physical/marklogic/fs/queryfile.scala
```diff
@@ -0,0 +1,94 @@
...
+        qs  <- hoist_(convertToQScript(lp))
+        xqy <- hoist_(qs.cataM(Planner[QScriptTotal[Fix, ?], XQuery].plan))
+        a   <- WriterT.put(f(xqy))(phase(xqy)).liftM[PlannerErrT]
+      } yield a
+
+      planning.leftMap(planningFailed(lp, _)).run.run
+    }
+
+    def exec(lp: Fix[LogicalPlan], out: AFile) =
+      plannedLP(lp)(κ(out.point[Free[S, ?]]))
```
jedesah: If I understand correctly, `execute` is not currently implemented to do anything right? Might be worth adding a `TODO` just to make that clear.

marklogic/src/main/scala/quasar/physical/marklogic/uuid.scala
```diff
@@ -0,0 +1,68 @@
...
+    * b: UUID and `b` was generated after `a` then
+    *
+    *   `toSequentialString(a) < toSequentialString(b) == true`
+    *
+    * returns None if the given UUID is not Type-1.
+    */
+  def toSequentialString(uuid: UUID): Option[String] =
+    if (uuid.version === 1) {
+      val parts = uuid.toString.split("-")
+      Some(s"${parts(2)}${parts(1)}${parts(0)}${parts(3)}${parts(4)}")
```
jedesah: I think this might warrant a comment to explain the rationale behind this particular ordering.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.physical.marklogic.xquery
+
+import quasar.Predef._
+
+final case class XQuery(override val toString: String) extends scala.AnyVal {
+  def apply(predicate: XQuery): XQuery = XQuery(s"${this}[$predicate]")
```
jedesah: Is `[` a valid character in an identifier?

jedesah:
Okay, just finished looking through this PR. Nothing too major to report, just a few comments, but nothing that stops this from getting merged. Let me if/when you make any small changes and I can merge it!

marklogic/src/main/scala/quasar/physical/marklogic/fs/package.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  implicit val resultCursorDataCursor: DataCursor[Task, ResultCursor] =
+    new DataCursor[Task, ResultCursor] {
+      def close(rc: ResultCursor) =
+        rc.close.void
+
+      def nextChunk(rc: ResultCursor) =
+        rc.nextChunk.map(_.foldLeft(Vector[Data]())((ds, x) =>
```
jedesah: Why can't we just call `map` on `Vector`?

marklogic/src/main/scala/quasar/physical/marklogic/xcc/ContentSourceIO.scala
```diff
@@ -0,0 +1,118 @@
...
+      liftT(qr.run(s)) map (r => new ResultCursor(chunkSize, s, r.resultSequence))
+    }
+
+  def resultStream(qr: SessionIO[QueryResults]): Process[ContentSourceIO, XdmItem] = {
+    def closeSession(s: Session) =
+      Process.eval_(liftT(Task.delay(s.close())))
+
+    def itemStream(s: Session) =
+      io.iterator(qr.run(s) map (_.resultSequence.iterator.asScala))
+        .flatMap(ritem => Process.eval(resultitem.loadItem(ritem)))
```
jedesah: Looks like `loadItem` calls `.cache`. Not clear to me why that is useful/necessary.

jedesah:
Oops, sorry, a few more questions I had written down on the side for further investigation once I was more familiar with the whole thing that I had forgotten to add to the PR.

marklogic/src/main/scala/quasar/physical/marklogic/fs/managefile.scala
wemrysi: Looks like there is something like this in #1369

marklogic/src/main/scala/quasar/physical/marklogic/fs/managefile.scala
wemrysi: Indeed it isn't.

marklogic/src/main/scala/quasar/physical/marklogic/fs/managefile.scala
wemrysi: Good idea.

marklogic/src/main/scala/quasar/physical/marklogic/fs/ops.scala
wemrysi: Sure, it has to do with how we're using lexical ordering on strings to preserve insertion order, this pads the sequence string we append to the "chunk id" so that all the strings are the same length. Without this, we could end up with strings like "abc2" and "abc10" where "abc" is the "chunk id" and "2" and "10" represent the second and tenth insertion, respectively. However `"abc2" < "abc10" == false`.

marklogic/src/main/scala/quasar/physical/marklogic/fs/package.scala
wemrysi: Looks like we can, I think this is a vestige from an earlier implementation, thanks!

marklogic/src/main/scala/quasar/physical/marklogic/fs/queryfile.scala
wemrysi: Sure, will do.

marklogic/src/main/scala/quasar/physical/marklogic/uuid.scala
wemrysi: That's what the above comment does...

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
wemrysi: It is the syntax for a [predicate XPath expression](https://www.w3.org/TR/2010/REC-xquery-20101214/#id-predicates). `apply` seemed like a close analog so that you could write `someXquery("foo".xqy == "bar".xs)` which translates to `someXquery[foo = "bar"]`

marklogic/src/main/scala/quasar/physical/marklogic/xcc/ContentSourceIO.scala
wemrysi: If the `ResultSequence` is in streaming mode, then the `ResultItem`s aren't all read into memory until you try and access their value. It wasn't clear that accesing `resultItem.getItem` read anything into memory, but `cache()` does just that. This seemed like the best way to ensure something accessing the values down the line didn't have them explode with an exception.

marklogic/src/main/scala/quasar/physical/marklogic/uuid.scala
jedesah: I am not sure I understand. So the above scaladoc comment is really useful and it explains how this function behaves, however, it doesn't seem to explain why that translates to an implementation where we separate the parts and then concatenates them back in that particular order. That's an implementation detail, so it shouldn't be part of the scaladoc, just a comment to help understand the implementation.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
jedesah: Oh, I meant it as in why does `this` need to be wrapped in `{}`?

marklogic/src/main/scala/quasar/physical/marklogic/xcc/ContentSourceIO.scala
jedesah: Ah, okay, but since we use this in a `scalaz.stream.Process`, is garbage collection going to happen properly and we won't end up overflowing our memory with a giant `Process` that is holding on to memory?

marklogic/src/main/scala/quasar/physical/marklogic/uuid.scala
wemrysi: Oh, ok. Though if you look at what a UUID type-1 is, I think this is the only? order that achieves the property described in the comment.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
wemrysi: Oh, not sure, doesn't hurt anything though, does it?

marklogic/src/main/scala/quasar/physical/marklogic/xcc/ContentSourceIO.scala
wemrysi: Why wouldn't it happen properly?

marklogic/src/main/scala/quasar/physical/marklogic/uuid.scala
jedesah: Maybe a link to some kind of specification of UUID type-1 would be sufficient.

I looked at the doc on the Java `UUID` class and it only describes the layout of a type 2 UUID. I found other non-java resources, but also nothing allowing me to understand why that ordering is the proper one (or even to know there will be hyphens in the string representation of the UUID)

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
jedesah: No it doesn't. It does make a curious programmer wonder why it's there in the first place though 😄

marklogic/src/main/scala/quasar/physical/marklogic/xcc/ContentSourceIO.scala
jedesah: Just general concern because of explicit caching, but ya, I think the `ResultItem`s should get garbage collected which in turn should collect the cached `XdmItem` it's holding on to.

wemrysi: Ah, I see, yeah as they go out of scope immediately as we only return the underlying `XdmItem`, so should get collected once the item is no longer referenced.

wemrysi: As an aside, the goal here was to just expose a reified `XdmItem` and avoid `ResultItem` at the first opportunity because it is so unpredictable (i.e. accessing a property on an uncached one could result in network operations).

wemrysi:
@jedesah Pushed a few commits that should address all your comments, thanks again.

marklogic/src/main/scala/quasar/physical/marklogic/fs/package.scala
```diff
@@ -94,7 +93,8 @@ package object fs {
         rc.close.void
 
       def nextChunk(rc: ResultCursor) =
-        rc.nextChunk.map(_.foldLeft(Vector[Data]())((ds, x) =>
-          ds :+ xdmitem.toData(x)))
+        TV.map(rc.nextChunk)(xdmitem.toData)
```
jedesah: syntax doesn't work here?

wemrysi: Not for `Functor` composition, that I know of.

jedesah: Ah, okay, I didn't initially grasp how this avoided nested calls to `.map`. I don't think `TV` is a very good identifier though. Maybe we can just inline the `Functor` composition at the call site.

I also wouldn't mind defining a `map2` syntax method for this. Now that I think of it, it's actually somewhat of a relatively common pattern.

PR #1375, opened 2016-08-23
read collection size from any BsonNumber

### LINKED ISSUE #1374: Bug: could not read collection statistics: Value expected to be of type INT32 is of unexpected type DOUBLE
We're getting this error when the collection size statistic is a Double instead of an Integer when running any query against a large collection.

mossprescott:
Fixes #1374.

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIO.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIO.scala
@@ -288,8 +288,8 @@ object MongoDbIO {
   def collectionStatistics(coll: Collection): MongoDbIO[CollectionStatistics] = {
     val cmd = Bson.Doc(ListMap("collStats" -> coll.collection.bson))
 
-    def intValue(doc: BsonDocument, field: String): String \/ Long =
-      \/.fromTryCatchNonFatal(Option(doc.getInt32(field)).map(_.longValue) \/>
+    def longValue(doc: BsonDocument, field: String): String \/ Long =
+      \/.fromTryCatchNonFatal(Option(doc.getNumber(field)).map(_.longValue) \/>
         s"expected field: $field").fold(_.getMessage.left, ι)
 
     def booleanValue(doc: BsonDocument, field: String): Boolean =
@@ -297,8 +297,8 @@ object MongoDbIO {
 
     runCommand(coll.database, cmd).map(doc =>
       (for {
-        count    <- intValue(doc, "count")
-        dataSize <- intValue(doc, "size")
+        count    <- longValue(doc, "count")
+        dataSize <- longValue(doc, "size")
         sharded  =  booleanValue(doc, "sharded")
       } yield CollectionStatistics(count, dataSize, sharded)))
         .flatMap(_.fold(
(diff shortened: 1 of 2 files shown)


mossprescott:
FYI: http://api.mongodb.com/java/current/org/bson/BsonDocument.html

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIO.scala
```diff
@@ -288,17 +288,17 @@ object MongoDbIO {
   def collectionStatistics(coll: Collection): MongoDbIO[CollectionStatistics] = {
     val cmd = Bson.Doc(ListMap("collStats" -> coll.collection.bson))
 
-    def intValue(doc: BsonDocument, field: String): String \/ Long =
-      \/.fromTryCatchNonFatal(Option(doc.getInt32(field)).map(_.longValue) \/>
+    def longValue(doc: BsonDocument, field: String): String \/ Long =
+      \/.fromTryCatchNonFatal(Option(doc.getNumber(field)).map(_.longValue) \/>
```
sellout: So, it’s still an integer, just stored in a Double? Or what is happening here?!

mossprescott:
Note: I have not actually tested this with a collection containing more than 2 GB of data, and don't intend to. Just going on the javadoc, along with the reported error.

sellout:
This raises a question – do we have any way to generate an installer (or even a jar) that we can give to a customer to test this before we merge a publish a new version?

mossprescott:
Not a legit way that I know of. We could certainly build a JAR manually if @boie0025 is willing to test it.

sellout:
Yeah, I think we should at least document somewhere how to build a JAR from a PR and where to tell a customer to copy it, then we can push that closer to test installers.

sellout:
I’ll merge once the build succeeds.

jdguzman:
@sellout @mossprescott we are willing to test if you can provide a repl jar at least. Or I can see if I can build the pr branch.

jdguzman:
FYI I work with Nic

boie0025:
FTR, the collection "size" attribute was 1461660637 at failure.

mossprescott:
@sellout: I think you should go ahead and merge. I think the risk of breakage is low and the easiest way to get jars to test is just to merge it. Make sense?

jdguzman:
@mossprescott @sellout, @boie0025 and myself are going to try building this PR branch and I'll report back what we find.

mossprescott:
@jdguzman: that's great! Let us know if you have any trouble. And welcome to the team ;)

jdguzman:
@mossprescott this is the one that can be confirmed fixed.  We are looking at your other PR now.

mossprescott:
@jdguzman: ok, that makes more sense and thanks again for helping confirm these fixes.

jdguzman:
@mossprescott ok for our particular use case this had a small affect but I'm thinking that the slowness may just be a byproduct of the dataset size and the query.

sellout:
@jdguzman Commenting on the wrong PR again? 😉

jdguzman:
trying to do to many things at once ...

sellout:
Just wanted to make sure before I merged it 😃

PR #1377, opened 2016-08-24
Efficient execution of simple counts with appended limit/skip ops

### LINKED ISSUE #1376: Simple count(*) performs very poorly
Two customers have stated that executing a `SELECT COUNT(*)` or similar causes Quasar to run an aggregate command rather than an equivalent .count() in MongoDB query language.  Testing confirms.

```
💪 $ select count(*) from patients
MongoDB
db.patients.aggregate(
  [
    {
      "$group": {
        "0": { "$sum": { "$literal": NumberInt("1") } },
        "_id": { "$literal": null }
      }
    },
    { "$limit": NumberLong("11") }],
  { "allowDiskUse": true });
```

One customer mentioned this took over 30 seconds in a large collection where the Mongo shell returned instantly.

Can this be optimized and prioritized?

mossprescott:
Fixes #1376.

When a simple `select count(*) from foo` is executed in the REPL, or saved as a view and then read with offset/limit parameters, `limit` and/or `offset` clauses are effectively added before the query is executed. Because the query always produces a single result, those clauses have no effect, but they were not being recognized when the query was executed, and so quasar was doing `aggregate` rather than `count`, which is potentially much slower. This commit just recognizes that pattern and disregards any skip/limit after a count operation, as long as they wouldn't affect the single result.

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/WorkflowExecutor.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/WorkflowExecutor.scala
@@ -294,11 +294,13 @@ private[mongodb] abstract class WorkflowExecutor[F[_]: Monad, C] {
         find(src, Find(None, bson.some, keys.some, skip, limit)) map (_.right)
       case (List(PipelineOpCore($MatchF((), sel)), Projectable(bson), PipelineOpCore($SortF((), keys))), Nil) =>
         find(src, Find(sel.some, bson.some, keys.some, skip, limit)) map (_.right)
-      case (List(Countable(field)), Nil) if skip ≟ None && limit ≟ None =>
+      case (List(Countable(field)), Nil)
+          if skip.getOrElse(0) == 0 && limit.cata(_ >= 1, true) =>
         labeledCount(src, Count(None, None, None), field) map (_.left)
       case (Nil, List(Countable(field))) =>
         labeledCount(src, Count(None, skip, limit), field) map (_.left)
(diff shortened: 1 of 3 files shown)


boie0025:
👍

mongodb/src/main/scala/quasar/physical/mongodb/WorkflowExecutor.scala
```diff
@@ -294,11 +294,13 @@ private[mongodb] abstract class WorkflowExecutor[F[_]: Monad, C] {
         find(src, Find(None, bson.some, keys.some, skip, limit)) map (_.right)
       case (List(PipelineOpCore($MatchF((), sel)), Projectable(bson), PipelineOpCore($SortF((), keys))), Nil) =>
         find(src, Find(sel.some, bson.some, keys.some, skip, limit)) map (_.right)
-      case (List(Countable(field)), Nil) if skip ≟ None && limit ≟ None =>
+      case (List(Countable(field)), Nil)
+          if skip.getOrElse(0) == 0 && limit.cata(_ >= 1, true) =>
```
sellout: Can you use `≟` (or at least `===`) here and in the other change? Don’t want to get farther for enabling that wart.

sellout:
Just the one thing, then I’ll merge.

jdguzman:
@sellout @mossprescott I can confirm that this PR fixes the issue.

jdguzman:
Sorry I meant to comment on the other PR.  We are testing this now actually.

boie0025:
@jdguzman and I have confirmed that this is resolved in the relevant circumstances.

Issue #1364, opened 2016-08-15
MongoDB 3.2: use $lookup for joins where only one source is a trivial read


mossprescott:
This will allow many more joins to run in the aggregation pipeline, including 3-way joins, as long as one of the sources in each merge step is an unfiltered read where the condition refers directly to a field.

- accept _any_ source on one side
- put the join value under a temp field before `$lookup`
- project the temp field away after `$lookup`
- optimize the temp away in the common case that it can be addressed directly (i.e. it's just a field, not an expression)

Issue #1311, opened 2016-07-29
MongoDB 3.2: support new expression and accumulation operators


mossprescott:
Depends on #1308.

New for `AccumF`:
- [ ] [`$stdDevSamp`](https://docs.mongodb.com/manual/reference/operator/aggregation/stdDevSamp/#grp._S_stdDevSamp) and [`$stdDevPop`](https://docs.mongodb.com/manual/reference/operator/aggregation/stdDevPop/#grp._S_stdDevPop)

New for `Expr`:
- [ ] `$sqrt`, `$abs`, `$log`, `$log10`, `$ln`, `$pow`, `$exp`, `$trunc`, `$ceil`, `$floor` will each come with a new standard library function
- NB: support for new array operators `$slice`, `$arrayElementAt`, `$concatArrays`, `$isArray`, `$filter` will be tracked a separate issue
- NB: a subset of `Accum` ops can now be used directly in `$project`. Using them to avoid `$group`/`$unwind` will be tracked in a separate issue.

zeeiyerWork:
This ticket can be closed since
- Arithmetic Operators are covered in #2272
- Array Operators in #2379
- Accum operators in

PR #1378, opened 2016-08-24
Handle join with arbitrary expression on one side of the condition

mossprescott:
This is a first step for #1364, handling joins where one source is a trivial read and the other source is a pipeline without filtering. Notably this allows a simple 3-way join to run in `aggregate`, which is pretty impressive:

```
 $ select a.foo, b.bar, c.baz from abc as a join abc as b on b.a_id = a._id join abc as c on c.b_id = b._id
MongoDB
db.abc.aggregate(
  [
    { "$match": { "_id": { "$ne": null } } },
    { "$project": { "left": "$$ROOT" } },
    {
      "$lookup": {
        "from": "abc",
        "localField": "left._id",
        "foreignField": "a_id",
        "as": "right"
      }
    },
    { "$unwind": "$right" },
    { "$match": { "right._id": { "$ne": null } } },
    { "$project": { "left": "$$ROOT" } },
    {
      "$lookup": {
        "from": "abc",
        "localField": "left.right._id",
        "foreignField": "b_id",
        "as": "right"
      }
    },
    { "$unwind": "$right" },
    {
      "$project": {
        "foo": {
```

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
@@ -21,6 +21,7 @@ import quasar.Predef._
 import quasar.fp._
 import quasar.javascript._
 import quasar.jscore, jscore.{JsCore, JsFn}
+import quasar.namegen._
 import quasar.std.StdLib._
 import quasar.physical.mongodb.accumulator._
 import quasar.physical.mongodb.expression._
@@ -56,7 +57,11 @@ object JoinHandler {
     */
   def pipeline[WF[_]: Functor: Coalesce: Crush: Crystallize]
     (stats: Collection => Option[CollectionStatistics])
-    (implicit ev0: WorkflowOpCoreF :<: WF, ev1: WorkflowOp3_2F :<: WF, ev2: Show[WorkflowBuilder[WF]])
+    (implicit
+      C: Classify[WF],
+      ev0: WorkflowOpCoreF :<: WF,
+      ev1: WorkflowOp3_2F :<: WF,
+      ev2: Show[WorkflowBuilder[WF]])
     : JoinHandler[WF, OptionT[WorkflowBuilder.M, ?]] = JoinHandler({ (tpe, left, right) =>
 
     val WB = WorkflowBuilder.Ops[WF]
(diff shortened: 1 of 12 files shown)


mossprescott:
@sellout: let me know when this flood of review gets to be too much and I'll spread it around.

jdegoes:
@sellout @mossprescott If this is solid, would be good to get this merged in ASAP for our release of 3.0.10.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
```diff
@@ -75,33 +91,85 @@ object JoinHandler {
...
+                  Reshape(ListMap(
+                    lName -> \/-($var(DocField(lName))),
+                    rName -> \/-($var(DocField(rName))))),
+                  IgnoreId),
+                $unwind(DocField(rName))),
+              Root(),
+              None)
+      }
+
+    def filterNonNull(wb: WorkflowBuilder[WF], field: BsonField): WorkflowBuilder[WF] =
```
sellout: So, I’m a bit nervous about this. In MRA, Null is a value (that would be better named Unit). We want Null to join with Null. Filtering it prevents that from happening, right?

If one side of the join is missing the field, that’s Undefined, not Null. If somehow a missing field is being converted to Null, we should fix that issue.

sellout:
Ok, so apologies for making you deal with `Delay`. And my only concern is `filterNonNull`. Back to you.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
paulp: This has been on my mind from the precog codebase too. Definitely a bad place to let the semantics blur.

sellout:
Oh, and you need a CHANGELOG.

mongodb/src/main/scala/quasar/physical/mongodb/JoinHandler.scala
mossprescott: Yes, this will prevent joining on the actual value `null`. What I really want here is a test that the field exists, and I was under the impression that there was no way to do that in ExprOp/Selector, but I now realize that `$exists` is there and I believe that will do exactly what I need. Make sense?

By the way, this is not new in this PR; I just refactored it a bit in anticipation of dealing with outer joins. That turned out to be problematic for other reasons, hence the TODOs lower down.

sellout: Oh, I wanted to ask about that, too – we can deal with outer joins in the simple both-sides-are-$read case, but not the more complex case? Or can we not deal with it in any case.

Because my impression was that `$lookup` is a left outer equi-join, and so should handle left and right outer joins easily, and then inner joins by post-filtering the un-paired results.

mossprescott: That's what my TODOs were referring to. `$lookup` is like a left outer join, which results in documents with `"right": []`. When we `$unwind` that, they get dropped. But MongoDb 3.2 also adds an option to `$unwind` that includes them, so I think it makes sense to use that rather than working around the limitation. Of course, that means we need support for new the operators in 3.2 (see #1311). See https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/#includearrayindex-and-preservenullandemptyarrays

sellout: Ah, right – in the non-$lookup case we replace `[]` with `[undefined]` (or, possibly, `[null]`, which would be wrong). But yeah, since $lookup only exists with the preserve-[] op, then we should just wait for that. Although it would be nice to also use preserve-[] in MR when possible, once it’s implemented.

Although, it preserves those as `null`, but since unwinding at that point should leave us with Bson.Doc at all the defined places, dropping `right: null` from the projections should be valid … right?

mossprescott:
Somehow that last commit was bogus, so I `--amend`ed it (sorry, world).

sellout:
Cool. I’ll merge upon success.

PR #1379, opened 2016-08-25
update to current MongoDB versions for integration tests

mossprescott:
Seems like the tarball for 3.0.9 went away and broke one of the builds. Maybe.


mossprescott:
Or maybe it was 3.2.3. Either way it's… bad.

mossprescott:
So, I guess that wasn't it. Can anyone explain this persistent error, only in the MongoDB 3.2 build?

https://travis-ci.org/quasar-analytics/quasar/jobs/154912156#L241

Maybe it's secondary to the real failure?

mossprescott:
Of course this turned out not to have anything to do with the broken builds, but it's probably a good idea to keep current on these MongoDB releases anyway.

PR #1385, opened 2016-08-26
Fix `Data.NA` breaking QScript conversion.

sellout:
The `Nullary` MapFunc has been renamed to `Constant`, as there are now
three nullaries – `Constant`, `Undefined`, and `Now`.

`Data.NA` is mapped to `Undefined` (note this is a MapFunc, _not_ an
EJson value).


wemrysi:
@sellout Mind adding some documentation for `Undefined` and `Now`? What do they mean? In the case of `Undefined` will it go away when we stop converting from `LogicalPlan`, i.e. why do we need to reify the absence of definition rather than just aborting the planner with an error?

sellout:
Yeah, I’ll add docs. `Now` just returns the current timestamp, unrelated to this change other than I know it’s needed and I was changing all the nullary stuff already.

We may still need `Undefined` regardless. We already fail when we can statically identify that a value will always be undefined. The places that I can think that it crops up currently are in non-exhaustive conditionals (`Cond` and `Guard`).

We could perhaps alternatively have single-arm versions of `Cond` and `Guard` (which is what I attempted with you the other day), but I don’t know if that’s simpler, and I’m not _certain_ this is the only place `Undefined` crops up.

sellout:
Updated docs.

wemrysi:
@sellout Thanks, LGTM. Needs a CHANGELOG and possibly a merge from master?

sellout:
Done and done.

wemrysi:
FYI, I cancelled the build for the CHANGELOG commit just to get it out of the queue.

PR #1386, opened 2016-08-26
fix latest merge

mossprescott:
_This_ one should get merged with the button ;)


jedesah:
Two wrongs make it right right? I am going to go ahead a merge this one before build passes.

PR #1387, opened 2016-08-26
More join via lookup

mossprescott:


mossprescott:
I just pushed another commit that fixes the one broken regression test, but I'm now looking at the other one where the plan seems correct but takes 10 minutes locally.

jdegoes:
@mossprescott When you get a chance, send me the two plans (map/reduce versus aggregation pipeline). The fact that `$lookup` sometimes takes so much longer is worth pointing out. 😆

mossprescott:
Here's the same query planned for map-reduce on 2.6 and for $lookup on 3.2:

```
💪 $ cd /local/quasar-test
💪 $ select z1.city as city1, z1.loc, z2.city as city2, z2.pop from zips as z1 join zips as z2 on z1.loc[*] = z2.pop
MongoDB
db.zips.aggregate(
  [
    { "$project": { "__tmp4": "$loc", "__tmp5": "$$ROOT" } },
    { "$unwind": "$__tmp4" },
    {
      "$group": { "__tmp6": { "$push": "$__tmp5" }, "_id": { "0": "$__tmp4" } }
    },
    {
      "$project": {
        "_id": true,
        "value": { "left": "$__tmp6", "right": { "$literal": [] }, "_id": "$_id" }
      }
    },
    { "$out": "tmp.gen_b307a398_0" }],
  { "allowDiskUse": true });
db.zips.mapReduce(
  function () {
    emit.apply(
      null,
      (function (key, value) { return [{ "0": value.pop }, { "left": [], "right": [value] }] })(
        this._id,
        this))
  },
  function (key, values) {
    var result = { "left": [], "right": [] };
    values.forEach(
      function (value) {
        result.left = result.left.concat(value.left);
        result.right = result.right.concat(value.right)
      });
    return result
  },
  {
    "out": { "reduce": "tmp.gen_b307a398_0", "db": "quasar-test", "nonAtomic": true }
  });
db.tmp.gen_b307a398_0.aggregate(
  [
    {
      "$match": {
        "value.left": { "$not": { "$size": NumberInt("0") } },
        "value.right": { "$not": { "$size": NumberInt("0") } }
      }
    },
    { "$unwind": "$value.left" },
    { "$unwind": "$value.right" },
    {
      "$project": {
        "city1": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$value.left"] },
                { "$lt": ["$value.left", { "$literal": [] }] }]
            },
            "$value.left.city",
            { "$literal": undefined }]
        },
        "loc": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$value.left"] },
                { "$lt": ["$value.left", { "$literal": [] }] }]
            },
            "$value.left.loc",
            { "$literal": undefined }]
        },
        "city2": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$value.right"] },
                { "$lt": ["$value.right", { "$literal": [] }] }]
            },
            "$value.right.city",
            { "$literal": undefined }]
        },
        "pop": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$value.right"] },
                { "$lt": ["$value.right", { "$literal": [] }] }]
            },
            "$value.right.pop",
            { "$literal": undefined }]
        }
      }
    },
    {
      "$project": { "city1": true, "loc": true, "city2": true, "pop": true, "_id": false }
    }],
  { "allowDiskUse": true });

Query time: 3.0s
 city1      | loc[0]      | loc[1] | city2            | pop |
------------|-------------|--------|------------------|-----|
 WYNNEWOOD  |  -75.275984 |     40 | VAN HORNESVILLE  |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | STAR PRAIRIE     |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | MERIDEN          |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | FELT             |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | NEWKIRK          |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | VIDAL            |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | SELBYVILLE       |  40 |

💪 $ cd /local3.2/quasar-test
💪 $ select z1.city as city1, z1.loc, z2.city as city2, z2.pop from zips as z1 join zips as z2 on z1.loc[*] = z2.pop
MongoDB
db.zips.aggregate(
  [
    { "$project": { "__tmp5": "$loc", "__tmp6": "$$ROOT" } },
    { "$unwind": "$__tmp5" },
    { "$project": { "left": "$__tmp6", "__tmp4": "$__tmp5" } },
    {
      "$lookup": {
        "from": "zips",
        "localField": "__tmp4",
        "foreignField": "pop",
        "as": "right"
      }
    },
    { "$project": { "left": true, "right": true } },
    { "$unwind": "$right" },
    {
      "$project": {
        "city1": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$left"] },
                { "$lt": ["$left", { "$literal": [] }] }]
            },
            "$left.city",
            { "$literal": undefined }]
        },
        "loc": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$left"] },
                { "$lt": ["$left", { "$literal": [] }] }]
            },
            "$left.loc",
            { "$literal": undefined }]
        },
        "city2": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$right"] },
                { "$lt": ["$right", { "$literal": [] }] }]
            },
            "$right.city",
            { "$literal": undefined }]
        },
        "pop": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": {  } }, "$right"] },
                { "$lt": ["$right", { "$literal": [] }] }]
            },
            "$right.pop",
            { "$literal": undefined }]
        }
      }
    },
    {
      "$project": { "city1": true, "loc": true, "city2": true, "pop": true, "_id": false }
    }],
  { "allowDiskUse": true });


Query time: 612.0s
 city1      | loc[0]      | loc[1] | city2            | pop |
------------|-------------|--------|------------------|-----|
 WYNNEWOOD  |  -75.275984 |     40 | VAN HORNESVILLE  |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | SELBYVILLE       |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | STAR PRAIRIE     |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | MERIDEN          |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | FELT             |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | NEWKIRK          |  40 |
 WYNNEWOOD  |  -75.275984 |     40 | VIDAL            |  40 |
```


When the pipeline with $lookup is running, I see a flood of these in the log: 
```
2016-08-29T14:26:10.828-0400 I QUERY    [conn171] query quasar-test.zips query: { pop: { $eq: 42.990043 } } planSummary: COLLSCAN ntoreturn:0 ntoskip:0 keysExamined:0 docsExamined:29353 cursorExhausted:1 keyUpdates:0 writeConflicts:0 numYields:229 nreturned:0 reslen:20 locks:{ Global: { acquireCount: { r: 15816052 } }, MMAPV1Journal: { acquireCount: { r: 7908026 } }, Database: { acquireCount: { r: 7908026 } }, Collection: { acquireCount: { R: 7908026 } } } 9ms
2016-08-29T14:26:10.839-0400 I QUERY    [conn171] query quasar-test.zips query: { pop: { $eq: -97.455095 } } planSummary: COLLSCAN ntoreturn:0 ntoskip:0 keysExamined:0 docsExamined:29353 cursorExhausted:1 keyUpdates:0 writeConflicts:0 numYields:229 nreturned:0 reslen:20 locks:{ Global: { acquireCount: { r: 15816512 } }, MMAPV1Journal: { acquireCount: { r: 7908256 } }, Database: { acquireCount: { r: 7908256 } }, Collection: { acquireCount: { R: 7908256 } } } 11ms
2016-08-29T14:26:10.850-0400 I QUERY    [conn171] query quasar-test.zips query: { pop: { $eq: 42.936629 } } planSummary: COLLSCAN ntoreturn:0 ntoskip:0 keysExamined:0 docsExamined:29353 cursorExhausted:1 keyUpdates:0 writeConflicts:0 numYields:229 nreturned:0 reslen:20 locks:{ Global: { acquireCount: { r: 15816972 } }, MMAPV1Journal: { acquireCount: { r: 7908486 } }, Database: { acquireCount: { r: 7908486 } }, Collection: { acquireCount: { R: 7908486 } } } 10ms
```

Which makes some kind sense if the computed key means there's no index, but strangely it seems that the ops the *follow* the `$lookup` have an impact on this behavior. I'm working on minimizing it now.

jdegoes:
@mossprescott There should be an index on `loc` and `pop`, right?

mossprescott:
We don't create any indexes when we upload these test data sets. The odd thing is that `$lookup` seems to sometimes behave a if it's constructing an "index" on the fly and other time does a COLLSCAN for each value. I'm still trying to nail down what makes the difference.

mossprescott:
After much experimenting, I think my earlier impression was wrong and the problem is actually quite simple.

`$lookup` seems to always do a query for each key it finds in the "left" source. If the field referenced by the join on the right side is indexed (or is `_id`, which is always indexed), then those queries are fast and everything's fine.

On the other hand, if the field on the right is not indexed, then each query is a COLLSCAN and therefore linear in the size of the "right" collection, which makes the `$lookup` O(m*n).

NB: my confusion was because we typically run queries with a `limit` in effect, and the `$lookup` is done lazily, so that if you limit the result to, say, 10 rows, then you only do enough queries on the "right" collection to produce the first 10 results (so, O(10*n)). But as soon as you add something to the query that forces it to produce the entire joined result, then you incur the entire O(m*n) scan.

The equivalent map-reduce, on the other hand, by definition builds a data structure indexed by the join keys and then inserts each value from the right into it, making it O(m + n). But of course map-reduce is doing it's work in a JS interpreter, so it's slower by a large constant factor.

That means that $lookup will be faster when the "right" collection is indexed or "small". I think the best approach here will be to generate `$lookup` only when we know that there's an index on the key for the right side, or when the number of documents (_n_) is smaller than some magic number that we choose (say, 100?)

I've already made the row count for each collection available in the planner, and I've looked at adding information about the available indexes. With that information it would be easy to apply the above heuristic. What might be tricky is getting the magic number right; ideally it would be tuned based on actual empirical data about how slow these queries are as compared to map-reduce. Alternatively, we might eventually make it configurable.

@jdegoes: unless you think I'm getting some of this wrong, that's the approach I'll pursue.

By the way, this problem already existed prior to this PR, and we just weren't running into it because we didn't have a simple test query that joined the full `zips` dataset. It wasn't until I started compiling the more complex flattening to `$lookup` that we happened to do that join. But I can easily write a query that works in master that runs into it, e.g. 
```sql
select z1.city, z1.pop, z2.city, z2.pop 
from zips as z1 join zips as z2 on z1.pop = z2.pop 
order by z1.pop desc
```

jdegoes:
I'm not a big fan of magic numbers. In theory, (n + m) is always less than (n * m) for n, m >= 1. So it's really just the constant factor, but if we're talking constant factor, the execution of the query will be "relatively" fast and can be made faster by indexing. So I'd just use the indexing criteria.

mossprescott:
Yes, that's a good point. And some more testing suggests that when the "right" collection is small, it doesn't make much difference which way we run the query, which makes sense because the data handled by map-reduce is small. In that case, the only advantage would be that the $lookup version avoids generating temporary collections.

The only problem I see with that is that it will make it harder to demo `$lookup` since our test data is only ever indexed on `_id`. I suppose 

I'd say these are all problems we can deal with down the road. For now I'll just do the index check.

mossprescott:
Here are a couple of representative timings for joining `zips` with `zips` (*m* = 29k, *n* = 29k):

   | not indexed | indexed
---|---|---|
map-reduce | 3.5s | 3.5s | 
$lookup | 600s | 1.8s |

And another case with *m* = 4k and *n* = 29k. This case is a bigger win for $lookup because it just never has to look at most of the larger data set:

   | not indexed | indexed
---|---|---|
map-reduce | 2.2s | 2.2s |
$lookup | 45s | 0.3s |

jdegoes:
> In that case, the only advantage would be that the $lookup version avoids generating temporary collections.

That's a good point. If you already have the information on "read-only" at this point, then it would be nice to stick with `$lookup`, no matter how much slower.

> The only problem I see with that is that it will make it harder to demo $lookup since our test data is only ever indexed on _id. 

We need some way to specify / create indexes in the tests. Which will be coming (probably) in 3.2 as part of an "optimize query family" feature that, given a query with variables, creates appropriate indexes to accelerate the query.

mossprescott:
@drostron: Ready for review, finally.

This is really three changes, with a somewhat twisted commit history:

1. handling arbitrary "left" sources in JoinHandler (that's #1364)
2. checking for the presence of an index on the key field in the "right" source
3. timing out test query execution after 5 minutes so that we'll get informative failures in CI

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIO.scala
```diff
@@ -306,6 +306,49 @@ object MongoDbIO {
...
+      case x: java.lang.Integer if x.intValue ≟ 1  => IndexType.Ascending.some
+      case x: java.lang.Integer if x.intValue ≟ -1 => IndexType.Descending.some
+      case "hashed"                                => IndexType.Hashed.some
+      case _                                       => None
+    }
+
+    collection(coll)
+      .flatMap(c => collect[Document](c.listIndexes))
+      .map(_.map(doc =>
+        (for {
```
drostron: Would it be clearer to pull this into a method like `indexFromDoc`?

drostron:
LGTM.

@mossprescott Only a minor comment. Will merge once you update or decide to leave as is.

drostron:
Looks like another build stall and needs a merge from master

PR #1388, opened 2016-08-29
Traverse branches when converting to Read.

sellout:
There are still problems with over-unioning due to other QScript
optimizations that need to be applied.

This fixes a few other issues as well:
- minor improvement to QScript rendering,
- add an optimization that pushes Maps ahead of Drop/Take to help with
normalization,
- add LP simplifications for `Take(Take())` and `Drop(Drop())`,
- fix the generation of the `count` in LP => QS of Drop/Take, and
- add a `chain` helper to write QScript flows more naturally.


core/src/main/scala/quasar/qscript/ConvertPath.scala
```diff
@@ -99,14 +95,194 @@ object ConvertPath {
...
+      DE: Const[DeadEnd, ?] :<: G,
+      EJ: EquiJoin[T, ?] :<: G,
+      FI: G :<: QScriptTotal[T, ?]):
+      ConvertPath.Aux[T, EquiJoin[T, ?], G] =
+    new ConvertPath[T, EquiJoin[T, ?]] {
+      type H[A] = G[A]
+
+      def convertPath[M[_]: Monad](f: ListContents[M]): StaticPathTransformation[T, M, EquiJoin[T, ?], G] = {
+        case EquiJoin(src, lb, rb, lk, rk, jType, combine)
+            // FIXME: Replace this inequality with a boolean
```
wemrysi: Not sure if this is what you meant, bu you could replace the inequality with `Recursive[T].project[Pathed[G, ?]](src.copoint).exists(_.run.isLeft)`.

sellout: Argh, yes. That’s all I wanted. I just couldn’t remember the name of the function, and my Internet was down when I was typing that. Meant to substitute it before I PRed 😆

core/src/main/scala/quasar/qscript/ConvertPath.scala
```diff
@@ -99,14 +95,194 @@ object ConvertPath {
...
+        f => List(wrapDir[T, F](f.value, makeRead(dir, f))).point[M]))))
+
+  def readFile[T[_[_]]: Corecursive, M[_]: Monad, F[_], G[_]: Functor](
+    f: ListContents[M])(
+    implicit R: Const[Read, ?] :<: G, QC: QScriptCore[T, ?] :<: G, F: F ~> G):
+      CoEnv[ADir, F, T[G]] => M[List[G[T[G]]]] =
+    _.run.fold(allDescendents[T, M, G](f).apply(_), fa => List(F(fa)).point[M])
+
+  def union[T[_[_]]: Recursive: Corecursive, F[_]: Functor]
+    (elems: List[F[T[F]]])
```
wemrysi: How about changing this type to `NonEmptyList[F[T[F]]]`, pushing error handling closer to the source of the problem?

core/src/main/scala/quasar/qscript/ConvertPath.scala
```diff
@@ -99,14 +95,194 @@ object ConvertPath {
...
-      def convertPath[M[_]: Monad](f: ListContents[M]): StaticPathTransformation[T, M, SourcedPathable[T, ?], G] =
-        sp => EitherT(List(CoEnv(SP.inj(sp).right[ADir])).right.point[M])
+      def convertPath[M[_]: Monad](f: ListContents[M]):
+          StaticPathTransformation[T, M, SourcedPathable[T, ?], G] = {
+        case Union(src, lb, rb)
+            // FIXME: Replace this inequality with a boolean
+            if None != Recursive[T].project[Pathed[G, ?]](src.copoint).find(_.run.isLeft) =>
+          convertBranchingOp(src, lb, rb, f)((s, l, r) =>
+            SP.inj(Union(s, l, r)))
+        // case LeftShift(src, HoleF, repair)
```
wemrysi: Should this be deleted? Completed?

sellout: Ah, I’ll remove it for now. It’s something I want to do, but isn’t part of this change.

core/src/main/scala/quasar/qscript/ConvertPath.scala
sellout: I waffled a bit on this. `union` gets the list from `readFile, which could error on the left side of its fold. Beyond that, `allDescendents` can legitimately return an empty list without it being an error. If it moves to `readFile` then it makes `postPathify` more complicated, since I can’t just `traverseM` from `List` to `EitherT[M, …, NonEmptyList]`.

wemrysi: You could just do the traversal in `postPathify` and then check that the result is non-empty before passing it to union. 

Also, `readFailed` doesn't seem like the appropriate error case to use here (as evidenced by the bogus `Nil` value supplied), maybe we need to add a `PlannerError` case to handle this scenario?

sellout: Yeah, `readFailed` isn’t right. There’s a problem with `PlannerError` in that it’s currently doing double duty and we have to separate that. The original use was as a way to indicate some failure in the connector’s processing. Now it’s also used by QScript, since QScript now does a lot of the work of (and prevents a lot of the PlannerErrors from happening in) the connector.

Should the old usage of PlannerError just be merged with FileSystemError, or is there some benefit in keeping it distinct? If we merge it, we’d keep the `planningFailed` case around until QScript conversion happens outside the algebra.

Anyway, that was a bit of noise to convince myself that this error should exist as a QScript-PlannerError, and not a FileSystem-PlannerError.

wemrysi: Ah, yeah I see. As for coalescing old `PlannerError` and `FileSystemError`, I guess the benefit of keeping them separate would make it easier/possible to separate the purely filesystem code from the querying code (i.e. an `fs` library with all of the read/write/manage/meta operations). The current `PlannerError` would become possibly its own type or renamed to something and used in the yet-to-be-defined `Querying` (or whatever we call it) operations.

core/src/main/scala/quasar/planner.scala
```diff
@@ -66,6 +67,15 @@ object Planner {
     def message = s"The variable “$name” is unbound at a use site."
   }
 
+  final case class NoFilesFound(dirs: List[ADir]) extends PlannerError {
+    def message = dirs.map(posixCodec.printPath) match {
+      case Nil    => "No paths provided to read from."
+      case h :: t =>
+        "None of these directories contain any files to read from: " ++
+          t.foldRight(h)(_ ++ ", " ++ _)
```
wemrysi: Alternatively, `mkString(", ")`

sellout: I knew that …

core/src/main/scala/quasar/qscript/ConvertPath.scala
```diff
@@ -99,14 +96,181 @@ object ConvertPath {
...
+    _.run.fold(allDescendents[T, M, G](f).apply(_), fa => List(F(fa)).point[M])
+
+  def union[T[_[_]]: Recursive: Corecursive, F[_]: Functor]
+    (elems: NonEmptyList[F[T[F]]])
+    (implicit
+      DE: Const[DeadEnd, ?] :<: F,
+      SP: SourcedPathable[T, ?] :<: F,
+      FI: F :<: QScriptTotal[T, ?])
+      : F[T[F]] =
+    elems.tail.foldRight(
```
wemrysi: Alternatively, `elems.foldRight1((elem, acc) => ...)`

wemrysi:
@sellout Changes look good, a couple minor comments. If you want to address, then I'll wait until you do, otherwise I can merge.

sellout:
Yep, I’ll fix those.

wemrysi:
@sellout Also a CHANGELOG while you're in there.

sellout:
Ok, this is updated.

wemrysi:
@sellout Have a test failure that, at first glance, looks to just be due to ordering in the results? Possibly because the previous `union` impl had `head` as the rightmost argument and now `foldRight1` has it as the leftmost.

sellout:
But odd that it only failed on one version of Mongo, right? In any case, I used `ISet` now and it should be deterministic?

Also, I think I realized what has been confusing me lately – `mongodb/test-quick` _only_ tests mongodb now, rather than all the projects it depends on. So I have to remember to run from the root project to get everything to run.

wemrysi:
Aye, that is odd. `ISet` should be deterministic, so we'll see how it fares.

PR #1389, opened 2016-08-29
Concat empty buckets

alissapajer:
In `concatBuckets` when there are no buckets, we no longer concat an empty array.


core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -212,26 +212,63 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
           Free.roll(MakeMap(StrLit("loc"), Free.point(RightSide))))).embed.some)
     }
 
-    "convert a constant shift array" in pending {
+    "convert a constant shift array of size one" in pending {
```
alissapajer: @sellout I have a separate branch that enables the shift array tests of size one, two, and three. But the changes in this PR are distinct from that bug fix (though required for it).

sellout:
Needs a CHANGELOG.

core/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -170,12 +170,20 @@ object QScriptCore {
...
+              case Some((buck, newBuckets)) => {
+                val (full, buckAccess, valAccess) = concat(buck, mf)
+                SrcMerge(
+                  EnvT((Ann(newBuckets.list.toList.map(_ >> buckAccess), valAccess), Map(Extern, full): QScriptCore[T, ExternallyManaged])),
+                  lv,
+                  rv).some
+              }
+              case None =>
+                SrcMerge(
+                  EnvT((Ann(Nil, HoleF[T]), Map(Extern, mf): QScriptCore[T, ExternallyManaged])),
```
sellout: There are a few places here that could use `EmptyAnn` instead, no?

alissapajer: Yes thanks, I meant to audit for those! I'll search for the pattern and change them all.

core/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -226,12 +244,18 @@ class Transform[T[_[_]]: Recursive: Corecursive: FunctorT: EqualT: ShowT, F[_]:
...
+        val (merged, b, v) = concat(bucks, func(lval, rval).embed)
+        EnvT((
+          Ann[T](newBucks.list.toList.map(_ >> b), v),
+          QC.inj(Map(EnvT((EmptyAnn[T], src)).embed, merged))))
+      }
+      case None =>
+        EnvT((
+          EmptyAnn[T],
+          QC.inj(Map(EnvT((EmptyAnn[T], src)).embed, func(lval, rval).embed))))
+    }
```
sellout: My own inclination here would be to pull the `EnvT` constructor outside of the match, rather than duplicating it, but that’s just my own style.

sellout:
I’m not a fan of the complexity spread to all of the cases, but I can’t come up with a better way to handle this.

I’ll merge once you make any of the minor changes I suggested and resolve conflicts.

alissapajer:
@sellout re the added complexity, yeah, this was the best solution I came up with. The thing is, each case _does_ handle the `Some` and `None` of `concatBuckets` differently. There may be similarities in how the cases handle the `None` case though - I'll glance at it from this angle to see if anything can be done.

core/src/main/scala/quasar/qscript/QScriptCore.scala
alissapajer: actually, this is the only place! I guess I changed the others already :)

sellout: There are other instances in this PR – maybe they were changed in the other PR?

alissapajer: I just found them - I needed to generalize my regex :)

alissapajer:
Pushed a change to use `EmptyAnn` but still working on merging in master... the `sbt compile` hangs...

alissapajer:
@sellout ok, your PR is merged in

PR #1390, opened 2016-08-29
Add `ExternallyManaged` to replace some usages of `Unit`.

sellout:
This was discussed in #1370, but I never made the change. So, here’s the change.


alissapajer:
LGTM - I'll merge when jenkins agrees

PR #1391, opened 2016-08-29
fix `mergeSrcs` bug

alissapajer:
Fixes a bug in `mergeSrcs` where we falsely assumed that we had the full source information, and as such were performing an incorrect optimization.


alissapajer:
@sellout this includes PR #1389

alissapajer:
@sellout so this should fix any issues you were seeing where, during the merging process, zipped pairs of sources reconverged. It might be difficult to know if a query would land in this state, so all buggy queries should be retried against this branch.

alissapajer:
@sellout this is up to date with master now

CHANGELOG/11.4.25.md
```diff
@@ -1 +1 @@
-- Do not concat an empty array when there are no buckets.
+- Fix bug in mergeSrcs where we assumed that if two sides are equal, we could optimize that case. This was based on the false assumption that the sources included full source information.
```
sellout: Gotta fix this – one of the dangers of our current merge with these changelog files is that bringing in another PR means your new changelog is seen as an edit of the old.

core/src/main/scala/quasar/qscript/Mergeable.scala
```diff
@@ -27,6 +27,17 @@ import scalaz._, Scalaz._
...
+  // `a1` and `a2` are the two sources to be merged.
+  //
+  // This function merges `a1` and `a2` into a common source, providing access
+  // through that common source to X and Y.
+  //
+  // In general, if `a1` and `a2` are equal, we cannot simply return
+  // `SrcMerge(a1, fm1, fm2)` because the `Hole` in `fm1` and `fm2` reference
+  // some unknown source and do not reference `a1`. In this case, our source
+  // merging has converged (at least for this iteration), but `fm1` and `fm2` still
+  // reference the old common source (which is not known to us here).
```
sellout: I would make at least the first four lines of this into Scaladoc.

CHANGELOG/11.4.25.md
alissapajer: Why does it matter that this is seen as an edit of the old? I merged in a branch which had a changelog file already, so that makes sense. Won't the merge script just look for a file of the correct name?

sellout: Yeah, and this doesn’t have the correct name. This is saying that 11.4.25 fixes this bug in mergeSrcs, when that version was actually the empty concat change.

I.e., the merge renamed your revision.md to 11.4.25.md. At least the merge script would catch this for us if we overlooked it, though.

sellout:
So, just the CHANGELOG thing, and I’ll merge (after Travis passes) – might want to `[skip ci]` on the CHANGELOG commit.

CHANGELOG/11.4.25.md
alissapajer: Oh! It doesn't have the correct name, you're right!

CHANGELOG/revision.md
```diff
@@ -0,0 +1 @@
+- Fix bug in mergeSrcs where we assumed that if two sides are equal, we could optimize that case. This was based on the false assumption that the sources included full source information.
```
alissapajer: @sellout this is now showing up as the correctly named `revision.md` file, and I reverted the other version-named file to be what's on master.

alissapajer:
@sellout okie, addressed the changelog thing and the scaladoc thing!

Also, how does one `[skip ci]`. Do we actually have a way to tell jenkins not to build a pushed commit?

sellout:
Oh yeah – you can just literally insert `[skip ci]` anywhere in the commit message, and it won’t try to build it.

PR #1393, opened 2016-08-30
Improvements for the MongoDB-QScript connector

sellout:
- ported Selector algebra
- simplified JS algebra
- always attempt to generate ExprOp before JS in FreeMap cases.


mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -40,8 +40,6 @@ object MongoDbQScriptPlanner {
 
   type OutputM[A] = PlannerError \/ A
 
-  type PartialJs = Partial[JsFn, JsFn]
```
mossprescott: 😢 🍺 

also, 👍

mossprescott:
This looks like some good stuff but I'm beginning to wonder when we'll start running the tests against it. As it is, this is a lot of partially duplicated code with no test coverage, isn't it? How are you keeping track of progress?

sellout:
The types, Moss, the types.

But seriously … I should commit the IT test env var. I have it done locally. However it’s not particularly useful until I connect ListContents. Which hopefully I can do today.

core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -493,6 +499,11 @@ object MapFunc {
 
 // TODO we should statically verify that these have a `DimensionalEffect` of `Mapping`
 object MapFuncs {
+  // nullary
+  @Lenses final case class Constant[T[_[_]], A](ejson: T[EJson]) extends Nullary[T, A]
+  @Lenses final case class Undefined[T[_[_]], A]() extends Nullary[T, A]
+  @Lenses final case class Now[T[_[_]], A]() extends Nullary[T, A]
```
mossprescott: Okay, I give up. What the heck does `Now` mean?

sellout: Ignore that – note that it doesn’t show up in the full diff. It was added (and documented) in a PR that was already merged.

But `Now` returns the current timestamp.

wemrysi: @mossprescott https://github.com/quasar-analytics/quasar/blob/master/core/src/main/scala/quasar/qscript/MapFunc.scala#L524

mossprescott:
Sounds good to me. And will we be able to leverage the existing PlannerSpec? Because a lot of work went into it and it covers a lot of things that aren't covered by regression tests, but of course it's likely to require manual updating for cases where the new plan doesn't match but is equally good.

sellout:
So, this needs a CHANGELOG, but anything else?

sellout:
(it has a CHANGELOG now)

mossprescott:
Was hoping to see `it` tests turned on, but not all that optimistically. Just tell me that's not reasonable and I'll merge.

sellout:
Yeah, that’s not going to happen on this PR, I think. I had hoped to get the Read conversion turned on today, but didn’t get a chance to. So, that’s next. Then I can enable IT.

PR #1394, opened 2016-08-31
Add a ThetaJoin => EquiJoin transformation.

sellout:
This is an optional transformation applied by connectors. Also normalize
QS before analyzing Join condition, which prevents some cases of the
join condition not being handled properly.

Oh, and this also adds some docs to QScript stuff.


core/src/main/scala/quasar/qscript/SourcedPathable.scala
```diff
@@ -43,11 +56,7 @@ sealed abstract class SourcedPathable[T[_[_]], A] {
     extends SourcedPathable[T, A]
 
 /** Creates a new dataset, |a|+|b|, containing all of the entries from each of
-  * the input sets, without any indication of which set they came from
-  *
-  * This could be handled as another join type, the anti-join
-  * (T[EJson] \/ T[EJson] => T[EJson], specifically as `_.merge`), with the
-  * condition being `κ(true)`,
+  * the input sets, without any indication of which set they came from.
```
wemrysi: Does union preserve duplicates? i.e. Does `(1, 2) union (1, 2)` result in `(1, 2, 1, 2)` or `(1, 2)`?

sellout: Good question. I _think_ duplicates should _not_ be preserved. Although, these semantics are open to debate …

From the user’s perspective, there may be duplicates, because if the provenance between the two sets differs (e.g., if you had gotten the previous example by doing ```
select * from `/foo` union all select * from `/bar`
``` where /foo and /bar both happen to have the elements `1` and `2`) then the values you’re _actually_ unioning at the QScript level would contain some representation of that difference and the final result would be `(1, 2, 1, 2)`.

But if the query were ```
select * from `/foo` union select * from `/bar`
``` (no `all`), then that provenance wouldn’t end up in the value, so you truly eliminate duplicates and the final result would be `(1, 2)`.

However … our provenance tracking version isn’t quite the same as SQL’s `union all`. In SQL, you get duplicates even if you join a collection against itself, which our provenance tracking wouldn’t do. We _could_ deal with that by having a version that inserted an extra bit of provenance on each side of the union.

@jdegoes?

Comment (jdegoes) [inline core/src/main/scala/quasar/qscript/SourcedPathable.scala]: The fundamental operation in MRA is an identity-aware union, similar to but not quite the same as `union all`. However, `union all` can be implemented by `union` followed by a `distinct` on values (i.e. `group by` value, then `arbitrary`).

That said, if you're talking about set literals, then all bets are off, as this is not (yet) specified in MRA. Does (1, 2) generate a map with unique keys, and distinct provenance from another occurrence of (1, 2)? Or is (1, 2) simply a set without identity or provenance? Or, possibly, is (1, 2) exactly equivalent to {1:1, 2:2}{*}?

All of these have different ramifications and the issue hasn't been explored in MRA yet (beyond whatever has been done in QScript).

Comment (sellout) [inline core/src/main/scala/quasar/qscript/SourcedPathable.scala]: “`union` followed by a `distinct`” would be `union` not `union all`. Yeah, we’re not talking about set literals here (I don’t think).

Comment (sellout) [inline core/src/main/scala/quasar/qscript/SourcedPathable.scala]: In any case, we don’t currently have a way to say “union these two things without doing any work that involves looking at the values at all”, which is really the point of `union all`, I think – just an optimization. We’re always going to compare the values.

**Comment by alissapajer** [comment]
`mongodb_read_only FileSystem` test failed in 3.0 RO build with a few connection timeout failures - rebuilding

File: core/src/main/scala/quasar/qscript/package.scala [hunk 2]
```diff
@@ -60,11 +73,8 @@ package object qscript {
   val ExtEJson = implicitly[ejson.Extension :<: ejson.EJson]
   val CommonEJson = implicitly[ejson.Common :<: ejson.EJson]
 
-  type FreeHole[F[_]] = Free[F, Hole]
```
alissapajer: oh good! I had wanted to remove this alias before, but had bizarre compilation errors...

core/src/main/scala/quasar/qscript/SourcedPathable.scala
jdegoes: Oops, sorry, yeah!

jdegoes: @sellout Why are the values compared in our union operation?

core/src/main/scala/quasar/qscript/package.scala
sellout: Ah yeah … I’ve been explaining the Free structures to a lot of people, and this alias kept making it more complicated.

The problem was that there are (illegitimate) `Corecursive` and `TraverseT` instances for `Free` and they are only found if there’s a type alias like the one above (because of the SI-2712-fix). So I had to change the places we were taking advantage of those bad instances.

core/src/main/scala/quasar/qscript/SourcedPathable.scala
sellout: Because our union distinguishes values with different provenances, and in QScript that is managed by reifying the provenance into the value. So SQL’s `union all` reifies the provenance and compares the values, whereas SQL’s `union` compares the values _without_ provenance.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -179,6 +179,60 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
+          (l.as[Hole](SrcHole), r.as[Hole](SrcHole)).some
+        else if (l.element(RightSide) && r.element(LeftSide) &&
+          !l.element(LeftSide) && !r.element(RightSide))
+          (r.as[Hole](SrcHole), l.as[Hole](SrcHole)).some
+        else None
+
+      def separateConditions(fm: JoinFunc[T]):
+          (List[(FreeMap[T], FreeMap[T])], Option[JoinFunc[T]]) = fm.resume match {
+        case -\/(And(a, b)) =>
+          val (fir, sec) = (separateConditions(a), separateConditions(b))
```
alissapajer: Instead of using these nested tuples, do you want to make a case class for the return type of `separateConditions`? Then we could use names instead of `._1` and `._2` below.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -179,6 +179,60 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
     }
 
+  def simplifyJoin[F[_]: Functor]
+    (implicit EJ: EquiJoin[T, ?] :<: F, QC: QScriptCore[T, ?] :<: F):
+      ThetaJoin[T, T[F]] => F[T[F]] =
+    tj => {
+      // TODO: This can potentially rewrite conditions to try to get left and right
+      //       references on distinct sides.
+      def alignCondition(l: JoinFunc[T], r: JoinFunc[T]):
+          Option[(FreeMap[T], FreeMap[T])] =
```
alissapajer: And maybe a case class for the tuple here too? Then we can see more clearly that this type is the same type as in the return type of `separateConditions`.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -179,6 +179,60 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
       }
     }
 
+  def simplifyJoin[F[_]: Functor]
```
alissapajer: @sellout can you add some scaladoc to this? Something about how we attempt to optimize away the `ThetaJoin` and just use `Filter`, but create `EquiJoin` when that's not possible.

sellout: Yes, but it’s more the opposite – we try to optimize the ThetaJoin into an EquiJoin, but need a Filter when there are inequalities.

PR #1395, opened 2016-08-31
Apply Read conversion to Mongo-QScript connector.

sellout:


wemrysi:
LGTM, looks like the `MONGO_RELEASE=3_0` build stalled so I restarted and will merge when it passes

PR #1396, opened 2016-08-31
A specification for the standard library functions (StringLib, so far)

mossprescott:
This partial test suite already serves to document the behavior of the functions defined in `StringLib` (excepting `Search`), and it verifies that both the typers/simplifiers and the compilation to MongoDB's pipeline are correct.

The pipeline runner identifies which functions are not implemented for the pipeline and those tests show up as "skipped" tests in the log.

It's expected that we'll follow this up with:
- tests (examples) for the rest of the standard library (well, the MapFuncs at least).
- test runners for MongoDB JS and `Selector` compilation
- test runners for additional backends, or for sub-components of them


sellout:
Are you telling me our StringLib implementations were already correct? Unpossible!

mossprescott:
Yes and no. There are some big caveats.

I restricted the domain of strings for MongoDB to just `alphaChar` because `$toLower` and `$toUpper` are defined only for ASCII (and I'm fixing it to be ASCII now). It's a bummer that there's no easy way to override the domain for a particular test.

Half of these String-related functions aren't implemented in aggregation ops anyway.

I declawed the `Substring` test because the simplifier blows up if it gets any invalid index. That one's an easy fix, so I'll do that now, too.

`Integer` and `Decimal` handle only `Int` and `Double` in the simplifier/typer, so I'll fix that, too.

Other than that, yeah, pretty much correct.

it/src/test/scala/quasar/physical/mongodb/ExprStdLibSpec.scala
```diff
@@ -0,0 +1,166 @@
...
+        prg: (Fix[LogicalPlan], Fix[LogicalPlan], Fix[LogicalPlan]) => Fix[LogicalPlan],
+        arg1: Data, arg2: Data, arg3: Data,
+        expected: Data): Result =
+        run(List(arg1, arg2, arg3), { case List(arg1, arg2, arg3) => prg(arg1, arg2, arg3) }, expected)
+    }
+
+    backend.name should tests(runner)
+  }
+
+  // TODO: where should this live?
```
wemrysi: `quasar.fp.catchable`?

core/src/test/scala/quasar/std/SimplifyStdLibSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package quasar.std
+
+import quasar.Predef._
+import quasar.{Data, LogicalPlan}, LogicalPlan._
+
+import matryoshka._
+import org.specs2.execute._
+// import org.specs2.matcher._
```
wemrysi: Thoughts on cleaning up these (and other) commented sections? Including `TODO`s is one thing, but I'd prefer to avoid committing commented partial implementations, if possible.

wemrysi:
@mossprescott LGTM, couple of minor comments I'll wait to hear from you about.

core/src/test/scala/quasar/std/SimplifyStdLibSpec.scala
mossprescott: I removed most of this stuff in the latest commit. I did leave behind a handful of tests that are just missing Arbitrary instances, which I will implement if you'd rather see that cleaned up as well.

wemrysi:
Restarted build job as it failed due to a reported race-cond in services spec regarding port usage.

wemrysi:
@mossprescott LGTM, CHANGELOG and I'll merge.

PR #1397, opened 2016-08-31
Use java's builtin facilities for URI encoding.

paulp:
There is a bug in http4s where %xx sequences are rendered
upper case if the xx characters might be hex digits. But it's
not clear why we're using NIH URI facilities in any case.


mossprescott:
👍

PR #1398, opened 2016-09-01
Initial MarkLogic MapFuncs and handful of other QScript ops

wemrysi:
Adds a lot of infrastructure for the MarkLogic planner, including variable name generation. Aside from some issues with `Take` (and probably `Drop` as well), generates valid xquery, though still isn't quite correct due to XML <-> JSON conversion issues which I hope to address shortly.


marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
```diff
@@ -0,0 +1,118 @@
...
+    case ConcatArrays(x, y) =>
+      ejson.arrayConcat(x, y)
+
+    case ProjectField(src, field) =>
+      for {
+        m <- freshVar[F]
+        k <- freshVar[F]
+        v <- ejson.mapLookup(m.xqy, k.xqy)
+      } yield {
+        let_(m -> src, k -> field) return_ {
```
jedesah: After discussing with @sellout, I don't understand why we are using `let_` here. And granted, I am fairly confused, but currently my understand leads me to believe we should only have the `if_` expression here.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
```diff
@@ -0,0 +1,118 @@
...
+          if_ (ejson.isMap(m.xqy)) then_ v else_ (m.xqy `/` k.xqy)
+        }
+      }
+
+    // TODO: What other types should this work with?
+    case ProjectIndex(arr, idx) =>
+      freshVar[F] map { i =>
+        let_(i -> idx) return_ {
+          (arr `/` s"child::${ejson.arrayEltName}[$i + 1]".xs `/` "child::node()".xs)
```
jedesah: I think this warrants some explanation in a comment or something. I have a hunch it is because of our encoding of array. To a naive observer, it seems surprising that these are `.xs` instead of `.xqy`.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
wemrysi: The `let_` is only there for now to help make the generated plan more readable until we have other things like pretty printing and function declarations working (the latter I hope to get to soon), otherwise it is pretty hard to follow any of the generated output as `src` and `k` can be large expressions.

There are other superfluous `let`s for this same reason.

jedesah: OH! My bad, I was confusing let with `for` for some reason. Ya, that makes sense.

wemrysi: Actually, I take that back, in this case the `let` is required to avoid evaluating `src` twice, (we need to test it and then project out of it depending on what kind of value it is).

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
wemrysi: I think this is due to lack of representation of [axes](https://www.w3.org/TR/xquery/#axes) and so we use `xs` so they are inserted literally into the XPath expression rather than forced into a call to `xdmp:value(...)`.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
```diff
@@ -17,34 +17,68 @@
...
+        fm <- rebaseXQuery(from, s.xqy)
+        ct <- rebaseXQuery(count, s.xqy)
+      } yield {
+        let_(s -> src, f -> fm, c -> ct) return_ {
+          if_ (f.xqy eq 0.xqy) then_ {
+            fn.subsequence(s.xqy, c.xqy)
+          } else_ {
+            mkSeq_(
+              fn.subsequence(s.xqy, 1.xqy, some(f.xqy)),
+              fn.subsequence(s.xqy, c.xqy))
```
jedesah: This is surprisingly confusing, but shouldn't the second argument here be `from + count`?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ReadPlanner.scala
```diff
@@ -26,10 +27,16 @@ import pathy.Path._
...
+
+  def plan[F[_]: NameGenerator: Monad]: AlgebraM[PlanningT[F, ?], Const[Read, ?], XQuery] = {
     case Const(Read(absFile)) =>
       val asDir = fileParent(absFile) </> dir(fileName(absFile).value)
       val dirRepr = posixCodec.printPath(asDir)
-      cts.directoryQuery(dirRepr.xs, "1".xs).point[Planning]
+
+      liftP((freshVar[F] |@| freshVar[F])((d, c) =>
+        for_(d -> cts.search(fn.doc(), cts.directoryQuery(dirRepr.xs, "1".xs)))
+        .let_(c -> d.xqy `/` "child::node()".xs)
```
jedesah: I just tried this query in the MarkLogic Query Console, but I am still confused as to what `"child::node()"` accomplishes. I can see that it works, I just don't understand what it does.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
```diff
@@ -49,4 +52,29 @@ package object qscript {
     implicit def equiJoin[T[_[_]]]: MarkLogicPlanner[EquiJoin[T, ?]] =
       new EquiJoinPlanner[T]
   }
+
+  def liftP[F[_]: Monad, A](fa: F[A]): PlanningT[F, A] =
+    fa.liftM[PhaseResultT].liftM[PlannerErrT]
+
+  def mapFuncXQuery[T[_[_]]: Recursive: ShowT, F[_]: NameGenerator: Monad](fm: FreeMap[T], src: XQuery): F[XQuery] =
```
jedesah: This function, `mergeXQuery` and `planMapFunc` seem like they could be made more generic. Maybe they could be defined once somewhere else for all connectors to use.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
wemrysi: I think it should be like `fn.subsequence(s.xqy, f.xqy, some(c.xqy))` with some off-by-one adjustments thrown in.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ReadPlanner.scala
wemrysi: The results of the `cts:search` are typed as document nodes (I think because we're using `fn:doc()` as the search expression), the `child::node()` projects out the underlying XML `element` or JSON `object` from the document node.

It may turn out this is unnecessary once the new `Read` transformation that allows a `Read` to imply a `LeftShift` (i.e. returns a dataset) is applied, but I'm not sure yet.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
wemrysi: That is likely true, but not sure if now is the time to try and generalize as this is likely to continue to change for the time-being.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
```diff
@@ -18,13 +18,88 @@ package quasar.physical.marklogic.xquery
...
+  def idiv(other: XQuery): XQuery = XQuery(s"$this idiv $other")
+  def mod(other: XQuery): XQuery = XQuery(s"$this mod $other")
   def and(other: XQuery): XQuery = XQuery(s"$this and $other")
   def or(other: XQuery): XQuery = XQuery(s"$this or $other")
   def seq: XQuery = mkSeq_(this)
   def to(upper: XQuery): XQuery = XQuery(s"$this to $upper")
-  def xp(xpath: XPath): XQuery = XQuery(toString + xpath)
+
+  def `/`(xqy: XQuery): XQuery = xqy match {
+    case StringLit(s)   => XQuery(s"$this/$s")
```
wemrysi: I'm not sure it's _wrong_ as there aren't really many facilities for programmatically constructing the segments of an XPath expression that I've found (maybe they exist?).

However I agree that `StringLit` is being misused in some places in the context of this operator.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
jedesah: I don't think so, we want to drop a certain number of elements from some index, so we concatenate a subsequence from the beginning  to "from", then we add to that a subsequence for which we have dropped the first "from" elements (in order to avoid duplicating them) + the number we should be dropping. Am I missing something?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ReadPlanner.scala
jedesah: Ok, that makes sense. I guess we'll see.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
jedesah: Ok, I can go with that.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
wemrysi: Yep, you're right.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ReadPlanner.scala
wemrysi: It is worth adding that we have to do this currently because sometimes the individual documents are pushed into a map and `document` nodes cannot be nested inside anything.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/XQuery.scala
jedesah: So if the `xqy` parameter is "truly" a `StringLit`, we should produce `XQuery(s"""$this/"$s"""")` which doesn't seem like the most useful thing but compiles and essentially maps all values of the sequence into that constant string.

Otherwise, it goes into the `Expression` case and at that point, we don't know if it's some static piece of XQuery with which we can just do `XQuery(s"$this/$s")` or if it's something dynamic that we need to use `xdmp:value` for.

What I am suggesting is that in order to stop abusing of `StringLit`, we add a `StaticExpression` that represents something that does not need to be "evaluated".

wemrysi: Not necessarily, i.e. this `/` operator that accepts any expression doesn't exist in xquery, so we get to determine what happens when it receives a string lit (or numeric lit, etc), I think we're saying the same thing though, I just wanted to mention that there isn't a "right" behavior yet, from what I can tell (unless we discover some way of constructing these programatically, then there would be).

wemrysi: I think adding another case for these sort of keyword-y expressions would help, but I think string lits should behave how they currently do.

jedesah: I think I am a little confused. How would one use this operator to express the representation of this XQuery:

```
for $d in cts:search(fn:doc(), cts:directory-query("/foo/", "1"))
let $c := $d/"child::node()"
return $c
```

wemrysi: Hm, good point, you couldn't, I didn't realize that was valid XPath. I guess when handling `ProjectField` if we find a `StringLit` we can convert it to a `StaticExpression` or whatever as a query containing `select foo.bar from ...` likely means `$foo/bar` and not `$foo/"bar"`.

jedesah: Oh, I didn't know `select foo.bar from ....` would return us a `StringLit`. Why is that?

wemrysi: I mean it will look something like `ProjectField((), Constant(Str("bar")))` no? And we convert that `Constant(Str("bar")) => XQuery.StringLit("bar")`.

jedesah: Ok, I understand. Ya, that seems like probably the best solution to me.

marklogic/src/main/scala/quasar/physical/marklogic/ejson/AsXQuery.scala
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+  implicit val extensionAsXQuery: AsXQuery[Extension] =
+    new AsXQuery[Extension] {
+      val asXQuery: Algebra[Extension, XQuery] = {
+        case Meta(value, meta) => ???
+        case Map(entries)      => ???
+        case Byte(b)           => xs.byte(b.toInt.xqy)
+        case Char(c)           => c.toString.xqy
+        case Int(i)            => xs.integer(i.xqy)
```
jedesah: Could we use actual XQuery integer literals here instead of the `integer` functions? What is the advantage of using one over the other?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
```diff
@@ -69,16 +65,6 @@ private[qscript] final class QScriptCorePlanner[T[_[_]]: Recursive: ShowT] exten
...
-          if_ (f.xqy eq 0.xqy) then_ {
-            fn.subsequence(s.xqy, c.xqy)
-          } else_ {
-            mkSeq_(
-              fn.subsequence(s.xqy, 1.xqy, some(f.xqy)),
-              fn.subsequence(s.xqy, c.xqy))
-          }
-        }
-      }
+      } yield let_(s -> src, f -> fm, c -> ct) return_ fn.subsequence(f.xqy, c.xqy + 1.xqy)
```
jedesah: More correct and simpler, that's the kind of change I like 😄

marklogic/src/main/scala/quasar/physical/marklogic/ejson/AsXQuery.scala
wemrysi: Possibly, I don't know about the advantages of one or the other, and was erring on the side of being explicit where the types are concerned given `xs:int` and `xs:integer`.

wemrysi:
@jedesah Ready for another look.

jedesah:
LGTM. Will merge when build passes.

PR #1399, opened 2016-09-01
Use typelevel compiler.

paulp:
The immediate consequence is the elimination of si2712fix-plugin.
A subsequent commit will exploit type literals and singleton inference
to eliminate the currently extensive reliance on shapeless's Nat.

The apparently unrelated change to move the convertError object
outside of the Errors object is actually necessary for things to
work with the typelevel options. This is life with the scala
compiler: any change at all is likely to alter the profile of
when types get forced, and you never know how it will materialize.
In fact it's very encouraging that only the one change was needed.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Switch to using typelevel scala compiler.
--- a/build.sbt
+++ b/build.sbt
@@ -34,6 +34,7 @@ lazy val buildSettings = Seq(
     ("scala", Apache2_0("2014–2016", "SlamData Inc.")),
     ("java",  Apache2_0("2014–2016", "SlamData Inc."))),
   scalaVersion := "2.11.8",
+  scalaOrganization := "org.typelevel",
   outputStrategy := Some(StdoutOutput),
   initialize := {
     val version = sys.props("java.specification.version")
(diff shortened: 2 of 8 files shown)


jdegoes:
👍 Looks good to me, assuming the build passes.

sellout:
Yeah, excited to get this in!

paulp:
Not jumping to any conclusions based on a single travis run, but it's not promising that the test jobs are all at 20 minutes and none of them has finished compiling. I'm really surprised if they snuck major performance regressions in with features which we aren't using yet (except inasmuch as we shifted from the plugin to the compiler) but the possibility exists.

sellout:
We could try without the new compiler options, if this is an issue. I don’t know if we’re at Travis rush hour or what.

paulp:
I'm doing some timings locally.

paulp:
master: 341.060 real, 1010.991 user, 21.135 sys
typelevel: 355.533 real, 1014.646 user, 23.126 sys

It's juuuuust big enough to make me keep looking.

paulp:
I forgot that the typelevel jars won't yet be cached on travis - this probably contributed to the constant factor increase in the test jobs.

paulp:
I'd say the performance merits further investigation (which I will do) but it's not so concerning that we shouldn't merge this.

sellout:
Just for fun, add a CHANGELOG/revision.md for me ;)

paulp:
@sellout You have a messed up idea of fun

PR #1401, opened 2016-09-01
Build fixes

mossprescott:
Various minor build/test cleanup, including actually skipping some tests in Travis.

Most importantly, there is one test that was meant to be skipped even in local builds because it takes ~2 minutes locally, and was not being skipped because specs2.

Despite that test alone being timed at more than 10 minutes in Travis, there's no clear payoff in overall build time.


paulp:
👍

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -386,9 +386,9 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
               structural.ObjectProject(LP.Free('__tmp0), LP.Constant(Data.Str("left"))),
               structural.ObjectProject(LP.Free('__tmp0), LP.Constant(Data.Str("right"))))))) must
       equal(chain(RootR).some) // TODO incorrect expectation
-    }
+    }.pendingUntilFixed
 
-    "convert basic join with explicit join condition" in pending {
```
jedesah: Hmm. What's the difference between the two?

web/src/test/scala/quasar/api/zip.scala
```diff
@@ -118,10 +118,9 @@ class ZipSpecs extends quasar.Qspec {
       z.map(_.size).sum.runLog.unsafePerformSync(0) must beBetween(MinExpectedSize, MaxExpectedSize)
     }
 
-    "zip many large files of random bytes (10 GB)" in {
-      // NB: comment this to verify the heap is not consumed
-      skipped("too slow to run every time (~2 minutes)")
```
jedesah: So this was being run anyway?

core/src/test/scala/quasar/qscript/QScript.scala
paulp: pendingUntilFixed runs the test and fails it unless it fails. Not sure what pending does, but will guess it doesn't run the test at all.

jedesah: Ya, I think pending is meant to replace the body of an example much like `todo` but conveying a slightly different semantic.

So I guess `specs2` was interpreting the block of code following the `pending` as something that needed to be executed before creating the next "examples"

mossprescott: They're supposed to be two syntaxes for the same thing, as far as I know, but we override `.pendingUntilFixed` so that pending tests are actually skipped in Travis, but they still turn into failures locally if they start passing.

I couldn't figure out how to make `pending { ... }` work the same way, so I settled for just switching these over.

web/src/test/scala/quasar/api/zip.scala
mossprescott: Yes. It seems like `skipped` used to throw an exception and this was once the approved way to do it, but at some point it started being ignored. Weirdly, although this test is timed at more than 10 minutes in Travis, it doesn't seem to add noticeably to the overall build time.

paulp: That might be related to inheriting from `ThrownMatchers`, which I think I read is how you influence what happens if an exception is thrown inside a test. It's madness that we have so little grasp of what really happens during test execution.

core/src/test/scala/quasar/qscript/QScript.scala
sellout: @alissapajer We need to remember this for the QScript tests – we should be using `.pendingUntilFixed` so that failing tests don’t count toward our coverage stats.

PR #1402, opened 2016-09-01
Adds specs2 showtimes to show test times for each test.

alissapajer:


alissapajer:
@mossprescott assigned this to you since you were doing other test suite and build work :) 

example:
```
[info]   + convert a simple read with path projects (998 ms)
[info]   + convert a basic invoke (1 second, 149 ms)
[info]   + convert project object and make object (1 second, 119 ms)
[info]   + convert a basic reduction (322 ms)
```

paulp:
It's already enabled on travis for everything which extends QuasarSpecification (which at the time was all of them) and personally I'm not such a fan of them on a continuous basis, especially because they are unparseable. If it would simply emit a single number, that'd be a lot better. Anyway, any who want them can enable them on the command line with
```
./sbt -Dshowtimes test
```
So, I vote against this.

alissapajer:
I have no problem enabling them locally and closing this PR. I pretty much always want them, which is why I suggested the PR. Any other opinions?

mossprescott:
I do like to see them locally for regression tests, and only decided not to re-enable that in my current PR at the last minute.

But I guess I agree with @paulp that it's more noisy than useful as a general thing. Good to know about that syntax for when you do want them, though.

alissapajer:
@mossprescott @psp thanks for the discussion. Closing PR.

PR #1406, opened 2016-09-02
Unreferenced

alissapajer:


core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -352,7 +353,7 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
             Free.point(ReduceIndex(0))))))).some)
     }.pendingUntilFixed
 
-    "convert a filter" in {
+    "convert a filter" in skipped { // takes 1 min 17 sec to run
```
alissapajer: This test became slow after I merged in master. So the slowing change is likely on master. But I haven't confirmed because it takes 10 minutes to check out master and then 10 minutes to rebuild the branch I need to be on.

In any case we'll skip it for now. And jenkins skips pending tests so I cannot see there how long it took to run.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -774,6 +774,7 @@ object MongoDbQScriptPlanner {
           (rebaseWB(joinHandler, from, src) ⊛
             (rebaseWB(joinHandler, count, src) >>= (HasInt(_).liftM[GenT])))(
             WB.skip)
+        case Unreferenced() => ???
```
alissapajer: @sellout this is a mongo thing

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -483,8 +509,7 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
       liftFF(repeatedly(coalesceQC[F, CoEnv[Hole, F, ?]](extractCoEnv[F, Hole], wrapCoEnv[F, Hole]))) ⋙
       liftFG(coalesceMapShift[F, CoEnv[Hole, F, ?]](extractCoEnv[F, Hole])) ⋙
       liftFG(coalesceMapJoin[F, CoEnv[Hole, F, ?]](extractCoEnv[F, Hole])) ⋙
-      // FIXME: currently interferes with elideConstantJoin
```
alissapajer: fixes the `FIXME`

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -36,7 +36,7 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
        // "select true"
        convert(listContents.some, LP.Constant(Data.Bool(true))) must
```
alissapajer: @sellout importantly, this test now takes < 1 sec instead of 20 sec

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
alissapajer: If you want, you can PR me the substitution for `???` and then we can merge that before merging this PR.

sellout: For the backends, the “safe” implementation of `Unreferenced` is whatever was being done for `Root` until now, and the `Root` impl should be changed to `_ => ???` (can’t use `κ`, because it’s eager, and will give a compiler error on `???`).

The reason being that `???` will error if it’s ever hit, and so it indicates a node that can’t exist in the plan.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
```diff
@@ -66,5 +66,6 @@ private[qscript] final class QScriptCorePlanner[T[_[_]]: Recursive: ShowT] exten
         fm <- rebaseXQuery(from, s.xqy)
         ct <- rebaseXQuery(count, s.xqy)
       } yield let_(s -> src, f -> fm, c -> ct) return_ fn.subsequence(f.xqy, c.xqy + 1.xqy)
+    case Unreferenced() => ???
```
alissapajer: @wemrysi are you the right person to ping for this marklogic thing? 

We added a new case `Unreferenced` in `QScriptCore` which replaces what used to be any unreferenced `Root`. This way, we won't unnecessarily generate the qscript for the entire file system `Root` when it's not even used and would later be elided.

core/src/test/scala/quasar/qscript/QScript.scala
alissapajer: Also, I couldn't find any alternative syntax for `skipped`, but if that's prefered by anyone, let me know why, and what the syntax is :)

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
alissapajer: Yes, `???` wasn't meant to be merged into master, I just hadn't addressed these cases yet and wanted to ping backend authors about them.

alissapajer: If we want something to actually error, we shouldn't use `???` and should do real error handling.

sellout: Yes, agreed. But we should be able to get rid of the Planner instances for `Root`, `ProjectBucket`, etc., so they don’t even exist.

sellout:
Heh, it seems that now that I’m using `pathify` for all the MongoDB planner tests, this is a PR I need merged 😆

core/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -72,7 +72,7 @@ object QueryFile {
     * LogicalPlan no longer needs to be exposed.
     */
   def convertToQScript[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, M[_]: Monad](
-    f: Option[ConvertPath.ListContents[M]])(
+    contents: Option[ConvertPath.ListContents[M]])(
```
sellout: I would like this to be called “listContents” or “getContents” – something verby that conveys that it’s a function.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -126,52 +126,78 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
              QC: QScriptCore[T, ?] :<: F,
              FI: F :<: QScriptTotal[T, ?]):
       ThetaJoin[T, T[G]] => F[T[G]] = {
-    case x @ ThetaJoin(src, l, r, on, Inner, combine) if on ≟ BoolLit(true) =>
-      (l.resume.leftMap(_.map(_.resume)), r.resume.leftMap(_.map(_.resume))) match {
+    case tj @ ThetaJoin(src, left, right, on, Inner, _) if on ≟ BoolLit(true) =>
+      (left.resume.leftMap(_.map(_.resume)), right.resume.leftMap(_.map(_.resume))) match {
         case (-\/(m1), -\/(m2)) => (FI.prj(m1) >>= QC.prj, FI.prj(m2) >>= QC.prj) match {
-          case (Some(Map(\/-(SrcHole), mf1)), Some(Map(\/-(SrcHole), mf2))) =>  // both sides are a Map
+          case (Some(Map(\/-(SrcHole), mf1)), Some(Map(\/-(SrcHole), mf2))) =>
```
sellout: Do we need to handle the “cross” cases – e.g., `case (\/-(SrcHole), -\/(src2)) if src2.length ≟ 0`?

core/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -90,6 +88,9 @@ object ReduceIndex {
 @Lenses final case class Drop[T[_[_]], A](src: A, from: FreeQS[T], count: FreeQS[T])
     extends QScriptCore[T, A]
 
+@Lenses final case class Unreferenced[T[_[_]], A]()
```
sellout: Can you add Scaladoc here? This is what I wrote on the wiki page: “a placeholder value that can appear in plans, but will never be referenced in the result. We consider this a wart. It should be implemented as an arbitrary value with minimal cost to generate (since it will simply be discarded).”

sellout:
Ok, back in your court.

core/src/main/scala/quasar/fs/QueryFile.scala
alissapajer: sure, I'll go with `listContents`

alissapajer: I just don't like single letter variable names because they are impossible to search for in my editor.

core/src/main/scala/quasar/qscript/Optimize.scala
alissapajer: hm, that is a good question indeed. I'm pretty sure we do need those cases, or rather, it would be correct to optimize those cases and so we should include them.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -126,52 +126,78 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
-              case -\/(Constant(_)) =>
-                rebase(l)(src).map(tf => QC.inj(Map(tf, combine >>= {
-                  case LeftSide  => HoleF
-                  case RightSide => mf2
-                }))).getOrElse(TJ.inj(x))
-              case _ => TJ.inj(x)
+              case -\/(Constant(_)) => rebaseLeft[F, G](rebase)(tj, mf2)
+              case _ => TJ.inj(tj)
+            }
+          case (Some(Map(-\/(src1), mf1)), Some(Map(-\/(src2), mf2))) if (src1.length ≟ 0) && (src1.length ≟ 0) =>
```
alissapajer: Just caught this double ref to `src1` - fixing.

core/src/main/scala/quasar/qscript/Optimize.scala
alissapajer: Ok, had to convinced myself that the "cross" cases are actually necessary and are not caught by the cases that match only one side of the pair, for example,
```scala
          case (Some(Map(\/-(SrcHole), mf1)), _) =>
```
So, I think the cross cases are necessary because, imagine we hit the following case:
```scala
          case (Some(Map(-\/(SrcHole), mf1)), Some(Map(-\/(src2), mf2))) if src2.length ≟ 0 =>
            (mf1.resume, mf2.resume) match { // if both sides are Constant, we hit the first case
              case (-\/(Constant(_)), _) => rebaseRight[F, G](rebase)(tj, mf1)
              case (_, -\/(Constant(_))) => rebaseLeft[F, G](rebase)(tj, mf2)
              case (_, _) => TJ.inj(tj)
            }
```
and then we hit the nested case `case (_, -\/(Constant(_))) =>`. If we don't actually match on the cross, then this would be caught by
```scala
          case (Some(Map(\/-(SrcHole), mf1)), _) =>
            mf1.resume match {
              case -\/(Constant(_)) => rebaseRight[F, G](rebase)(tj, mf1)
              case _ => TJ.inj(tj)
            }
```
but would not be optimized because the LHS isn't `Constant` even though RHS is.

alissapajer:
@sellout okie pushed a few changes, addressed your comments, back to you!

alissapajer:
@jedesah FYI this is the PR with the `Unreferenced` fix that you should run against to see if it fixes the issues you're seeing.

alissapajer:
^^ build failed with a compile error which is strange because I compiled before pushing, so maybe sbt was tricking me. I'll report back.

alissapajer:
yeah, it was an issue of `sbt clean` :/ didn't want to push something not compiling but oh well

alissapajer:
ok @sellout fix the compile errors, back to you for real now! :)

core/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -119,6 +126,7 @@ object QScriptCore {
           case Filter(a, func)            => f(a) ∘ (Filter(_, func))
           case Take(a, from, c)           => f(a) ∘ (Take(_, from, c))
           case Drop(a, from, c)           => f(a) ∘ (Drop(_, from, c))
+          case Unreferenced()             => (Unreferenced[T, B](): QScriptCore[T, B]).point[G]
```
sellout: Not something to change in this PR, but in this case I find the `G.point(Unreferenced[T, B]())` syntax nicer. Doesn’t need the explicit annotation.

sellout:
Ok, looks like this just needs the MarkLogic impl and a clean build, then 👍🏾

core/src/main/scala/quasar/qscript/QScriptCore.scala
alissapajer: But in order to do that, you need to have to write the implicit in the argument list like
```
def traverseImpl[G[_]](...)(implicit G: Applicative[G])
```
rather than
```
def traverseImpl[G[_]: Applicative](...)
```
And I find the later easier to read :) But I also don't like the explicit annotation, so... yeah :)

alissapajer:
@sellout
marklogic is done
master merge has been resolved
awaiting green build

PR #1407, opened 2016-09-02
New optional transformation: convert Reads to “shifted” Reads.

sellout:
- Use a Prism for CoEnv wrapping.
- Defined PrismNT
- Add an `ExtractId` MapFunc


jdegoes:
This seems a bit ad hoc. Why not imagine `LeftShift` as the composition of some new `LeftShift'` followed by an operation to extract out the identity and value components? In this case, your transformation goes from `Read` to `Read'`, where `Read'` does a `LeftShift'`, leaving the extraction of identity and value parts out?

sellout:
I’m not sure what the “operation to extract out the identity and value components” is.

I’m planning to introduce a new QScript component, something like `ShiftedRead`, where this transformation converts `Read` to `ShiftedRead`, but wanted to introduce that in a separate PR. So, this basically does two things:
```
Read; LeftShift => ShiftedRead
Read            => ShiftedRead; Reduce
```
Where a `Read` represents a single value, and a `ShiftedRead` represents a set of values. The `ExtractId` isn’t used with `LeftShift` but with the `Reduce` we have to insert if we ended up in the (generally uncommon) case of a `Read` not followed by a `LeftShift`. In SQL² this is the distinction between the two queries:
```sql
from `/foo/bar` -- Read; LeftShift
`/foo/bar`      -- Read
```
There are two things I think aren’t ideal about this transformation:
1. adding a special-purpose `ExtractId` and
2. ShiftedRead can only represent data that _can_ be shifted – i.e., maps and arrays (but this is negligible, as stores that don’t already have this restriction won’t apply this transformation).

Any way, that was mostly background for anyone else who might be looking at this. I’m happy to take a different approach, but I don’t know what that would be.

jdegoes:
> I’m not sure what the “operation to extract out the identity and value components” is.

I misspoke, but was thinking of a more generalized `ExtractId` (something like `ExtractIdValue`) which would assume that each input is a key/value pair, and it would extract out the key to use as the identity for the read/left-shifted combined operation, and extract out the value accordingly.

Any case, looks good to me. 👍

foundation/src/main/scala/quasar/fp/PrismNT.scala
```diff
@@ -0,0 +1,33 @@
...
+package quasar.fp
+
+import quasar.Predef._
+
+import matryoshka._
+import scalaz._, Scalaz._
+
+/** Just like Prism, but operates over Functors.
+  */
+final case class PrismNT[F[_], G[_]]
```
alissapajer: @ sellout do you want this in matryoshka?

foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -654,4 +654,17 @@ package object fp
       CoEnv[A, F, T[CoEnv[A, F, ?]]] => CoEnv[A, F, T[CoEnv[A, F, ?]]] =
     co => co.run.fold(κ(co), f)
 
+  def idPrism[F[_]] = PrismNT[F, F](
```
alissapajer: Same, should this be in matryoshka?

core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -595,6 +598,11 @@ object MapFuncs {
   @Lenses final case class Guard[T[_[_]], A](a1: A, pattern: Type, a2: A, a3: A)
       extends Ternary[T, A]
 
+  /** Extracts _some_ unique id from a value. This should only ever be applied
+    * at the top-level of a collection.
```
alissapajer: @sellout why the (unenforced) constraint about which level of a collection this can be applied to? 

At the top level, we have a unique id, but once we drill down, then we effectively have a list of provenance, so at that point that "id" would be the entire provenance list. Just talking through the various situations...

So, do you think it would it make sense to do something like create a `ReadId` node that reads out a collection and then necessarily extracts the id at the top level, because a `Read` by definition is at the top level?

alissapajer: In my suggestion above then, `ReadId` would not be a `MapFunc` like `ExtractId` is now, but instead would be an actual qscript node. `ExtractId` seems to be representing a very specific map operation that is not legal to apply to all collections, so because of that, it's not a true `MapFunc` since we cannot always map over a collection with it.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -418,21 +419,6 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
-    new (F ~> λ[α => Option[F[α]]]) {
-      def apply[A](fa: F[A]): Option[F[A]] = Some(fa)
-    }
-
-  def extractCoEnv[F[_], A]: CoEnv[A, F, ?] ~> λ[α => Option[F[α]]] =
-    new (CoEnv[A, F, ?] ~> λ[α => Option[F[α]]]) {
-      def apply[B](coenv: CoEnv[A, F, B]): Option[F[B]] = coenv.run.toOption
-    }
-
-  def wrapCoEnv[F[_], A]: F ~> CoEnv[A, F, ?] =
```
alissapajer: ha, cool, I hadn't thought about this pair as a prism before! nice change.

core/src/main/scala/quasar/qscript/MapFunc.scala
sellout: Ah, reading this I feel like I understand @jdegoes suggestion, and that it’s similar to this?

This sounds like a good idea. Basically, the `ShiftedRead` trait would either have two case classes (`SRead` and `SReadWithId`) or one case with a 2-valued `includeId` parameter (connector authors: any preference?). Because _usually_ we won’t want to propagate the id, so we should have a way to indicate it shouldn’t be generated.

foundation/src/main/scala/quasar/fp/PrismNT.scala
sellout: Hopefully it could get into Monocle, but I don’t know their stance on Scalaz/Cats dependency.

foundation/src/main/scala/quasar/fp/package.scala
sellout: And this should _really_ be an `Iso` (which gives us a `Prism` for free), but I didn’t want to define any more types than I had to (although I should probably at least add a comment to this effect).

Maybe it would also be taken by Monocle? ¯\\\_(ツ)_/¯

sellout:
Ok, different approach, using `ShiftedRead` rather than `ExtractId`. I think this is better, so thanks.

PR #1408, opened 2016-09-03
Fold SourcedPathable into QScriptCore.

sellout:
The purpose it was made for (the original version of `pathify`) no
longer exists, so there’s no reason to keep these nodes separate from
the other ones that occur in every form of QScript.


sellout:
@wemrysi This is yours because I edited the MarkLogic connector 😉

PR #1412, opened 2016-09-07
Fix a bug that caused infinite recursion.

sellout:


core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -524,7 +523,6 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
       liftFF(simplifyQC[F, CoEnv[Hole, F, ?]](coenvPrism)) ⋙
       liftFF(swapMapCount[F, CoEnv[Hole, F, ?]](coenvPrism.get)) ⋙
       liftFG(compactLeftShift[F, CoEnv[Hole, F, ?]]) ⋙
-      Normalizable[F].normalize ⋙
```
alissapajer: @sellout presumably all tests still pass without this second call here to `normalize`? Why is the second call a problem and not the first?

sellout: That is something I should have checked – Travis will tell us whether the tests pass (I merely ran stuff that was hanging and no longer is).

I think either could have been removed, but it is confusing since they’re sequential. I think the trick is that `normalize` recursively calls `applyToFreeQS` (which then calls `normalize` …) on any branches on the node, and so we end up with an extra fork on each recursive call. But I didn’t work it all the way through 😕

jedesah: hmm... not sure I fully understand, but it seems to work...

@alissapajer What do you think?

alissapajer: @jedesah so I think what was happening was the following: 

`normalize` would be called on the original qscript, which would then recurse into `applyToFreeQS`, which would call `normalize`, etc. until that recursion broke out. Then the subsequent optimizations in `applyToFreeQS` would be called, one of which is a second call to `normalize`. By this point, the qscript had changed shape such that `normalize` again calls `applyToFreeQS`, even though the previous call to `normalize` had broken out of the loop. Then we hit the first call to `normalize` again (likely with the same qscript as before), and the cycle continues.

So: we `normalize` form _X_. That breaks out of the loop and hits the second `normalize` as form _Y_. That calls `applyToFreeQS` which calls the first `normalize` with form _X_. Etc.

So removing the second call to `normalize` lets us break out of the loop.

Additionally, given that no tests were requiring the second call to `normalize`, I'm fine with this fix. If you can remove code and no tests break, then the code should not have been there! :)

Though, given that the second call to `normalize` decided that something isn't actually normalized, I suspect that in the future we may come across a test case that actually requires multiple calls to `normalize`. When this happens, we'll address the problem with test in hand, rather than just guessing where we might need to `normalize`.

jedesah: Okay, that's a fairly satisfying answer! 😄

PR #1413, opened 2016-09-07
Mark all regressions tests as pending for MarkLogic

jedesah:


alissapajer:
@jedesah is this build going to hang because it doesn't include fix #1412 ?

jedesah:
@alissapajer No, because we don't yet have a Travis build that actually runs the integrations tests for MarkLogic.

alissapajer:
@jedesah ah good point! :)

PR #1414, opened 2016-09-07
Improve names of some function arguments in core

jedesah:


alissapajer:
@jedesah LGTM - I'll merge once build is green

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -537,25 +537,25 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
              CP: ConvertPath.Aux[T, Pathable[T, ?], F]):
       T[F] => EitherT[M,  FileSystemError, T[QScriptTotal[T, ?]]] =
-    _.cataM[EitherT[M, FileSystemError, ?], T[QScriptTotal[T, ?]] \/ T[Pathable[T, ?]]](FS.pathifyƒ[M, F](g)) >>=
-      (_.fold(qt => EitherT(qt.right.point[M]), FS.toRead[M, F, QScriptTotal[T, ?]](g)))
+    _.cataM[EitherT[M, FileSystemError, ?], T[QScriptTotal[T, ?]] \/ T[Pathable[T, ?]]](FS.pathifyƒ[M, F](ls)) >>=
+      (_.fold(qt => EitherT(qt.right.point[M]), FS.toRead[M, F, QScriptTotal[T, ?]](ls)))
 
   def eliminateProjections[M[_]: Monad, F[_]: Traverse](
-    fs: Option[ConvertPath.ListContents[M]])(
+    ls: Option[ConvertPath.ListContents[M]])(
```
alissapajer: Since this one is an `Option` it could have a different name, but I never know what to name something with an optional argument. Maybe `lsOpt`? But I'm also fine with it as is.

jedesah: I also sometimes hesitate to add `Opt` to the end of an identifier to signify that it's an option. In part, that's because the type already tells you that, but I think in this case it's nice since the identifier is pretty short. I might change my mind if we had perfect semantic analysis tools that could instantly tell you the type of any identifier you hover over, but that's not likely to happen in `quasar` for a little while so a little help from the identifier name is somewhat welcome 😄

alissapajer:
@jedesah can you merge in master? The build script failed because it wasn't possible to do a fast-foward merge.

alissapajer:
Connection failure timeouts in the 2.6 build. I started the build again.

This is a situation when I wish jenkins would let you restart just one of the builds, since the others all passed. And also I wish it would keep the build history so we could see the history of all builds for this PR.

alissapajer:
Oh, we have travis not jenkins :) Jenkins does include my requested features, but it comes with its own cans of worms.

jedesah:
@alissapajer Actually, it is possible to restart just one build job in Travis. You need to click on that build and then press the restart build button from that screen.

alissapajer:
@jedesah hm, I'm 90% sure I did just that and it restarted all 5 builds, but next time I need to restart a build I will definitely try that to confirm!

alissapajer:
@jedesah so the 2.6 build failed again with the same connection error failures. I won't restart it this time so you can see it. I see no way this PR could have cause those failures, though.

PR #1416, opened 2016-09-08
Coproduct syntax.

paulp:
It's pretty modest machinery, but does wonders for readability. I understand there was once something along these lines and that there were performance issues. I think it likely it was the same bug being reported in [SI-8801](https://issues.scala-lang.org/browse/SI-8801) which is fixed in scala 2.11.6+.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Added coproduct syntax.
--- a/core/src/main/scala/quasar/fs/mount/package.scala
+++ b/core/src/main/scala/quasar/fs/mount/package.scala
@@ -19,6 +19,7 @@ package quasar.fs
 import quasar.Predef.{Map, Vector}
 import quasar.effect._
 import quasar.fp.TaskRef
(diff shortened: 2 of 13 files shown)


sellout:
This would be nice to have. Makes it a lot easier to have tight types through the QScript transformations.

@wemrysi had a syntax I really liked before –
```scala
sealed trait :\/:[F[_], G[_]] extends CoM
sealed trait :\:[F[_], T <: CoM] extends CoM
```
So we would end up with coproducts like
```scala
CfgErr :\: EnvErr :\: PhysErr :\/: Task
```

paulp:
Yeah, I was pretty unhappy with `CoId`, I'll change it to that syntax.

paulp:
@sellout After a little dance with travis it's green. Is my slightly variant syntax ok?

sellout:
This looks good – just wondering if you’ve tried benchmarking compilation against the old version at all.

paulp:
Only coarsely, but there was no observable difference.

PR #1425, opened 2016-09-09
Qscript misc

alissapajer:
@sellout some misc qscript changes that I didn't want to lose

### DIFF
--- a/core/src/main/scala/quasar/qscript/QScriptCore.scala
+++ b/core/src/main/scala/quasar/qscript/QScriptCore.scala
@@ -32,7 +32,9 @@ sealed abstract class QScriptCore[T[_[_]], A]
 @Lenses final case class Map[T[_[_]], A](src: A, f: FreeMap[T])
     extends QScriptCore[T, A]
 
-@Lenses final case class ReduceIndex(idx: Int)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -232,38 +234,37 @@ object QScriptCore {
             concatBuckets(b1.map(_ >> m1)) match {
```
alissapajer: @sellout in this case we just use the buckets `b1` and ignore the buckets from the other side. I am unable to convince myself this is correct. 

We still need bucket access into the new source, and I don't think we can just use the buckets from one side and not the other. Is this just a buggy remnant left over from an old version of source merging, maybe when we had a different provenance representation?

It seems like what we actually need to return here is a new common source (without provenance information) along with access into both the values and buckets for each side. Instead, what we are actually returning is a common source which includes bucketing information (from only one side), and then access to the left and right values.

Can you see through what was going on here?

alissapajer: (Also this is out of the scope of this PR, but I the discussion needed to be had somewhere...)

PR #1423, opened 2016-09-08
Replace transformers with mtl classes in `convertToQScript` and `ConvertPath`

wemrysi:
Replaces the explicit choices of `WriterT[?[_], PhaseResults, ?]` and `EitherT[?[_], FileSystemError, ?]` in `QueryFile.convertToQScript` with their equivalent "mtl" class. Also does the same in `ConvertPath`.

This expresses the required capabilities more directly, trading data types for constraints, and makes them easier to use with other transformer stacks (or algebraic effects) that satisfy the constraints.

PR #1426, opened 2016-09-09
Add support for XQuery module imports and namespace/function declarations

### LINKED ISSUE #1418: Module import and namespace/function declaration tracking
Add support for namespace and function declarations as well as module imports and track their usage so they can be added to the prolog of the xquery main module sent to the server.

This will also allow us to define actual local functions which should reduce the size and improve debugability of the generated plans. Currently these are huge as we're inlining all of our local "functions" at every call site.

wemrysi:
Implements XQuery main modules, namespace declarations, function declarations and module imports. These constructs can be defined via the scala dsl and used in planning QScript. Dependency tracking is used such that only the declarations and imports actually required for a plan are included in the final module sent to marklogic.

As a smoke test, the left shift functions have been converted into function declarations with the intention that the rest of the custom "functions" will be similarly converted in a separate PR.

This PR depends on #1423 so it contains some of the same commits, which can be ignored (i.e. all the changes in `quasar.qscript`).

The dsl for declaring functions is ok, but a little awkward, hopefully we can improve on this as we use it more, suggestions for improvement welcome!

Here is an example of a plan for `select * from /explain/data` using the new features

```xquery
xquery version "1.0-ml";

declare namespace ejson = "http://quasar-analytics.org/ejson";
declare namespace quasar = "http://quasar-analytics.org/quasar";

declare function ejson:array-left-shift($arr as element(ejson:array)) as item()* {
  $arr/child::ejson:array-element/child::node()
};

declare function ejson:map-left-shift($map as element(ejson:map)) as item()* {
  $map/child::ejson:map-entry/child::ejson:map-value/child::node()
};

declare function quasar:left-shift($node as node()) as item()* {
  if (fn:node-name($node) = fn:QName("http://quasar-analytics.org/ejson", "ejson:array")) then ejson:array-left-shift($node) else if (fn:node-name($node) = fn:QName("http://quasar-analytics.org/ejson", "ejson:map")) then ejson:map-left-shift($node) else quasar:node-left-shift($node)
};

declare function quasar:node-left-shift($node as node()*) as item()* {
  $node/child::node()/child::node()
};

for $v1 in ((: Read :)()) let $v2 := $v1, $v3 := quasar:left-shift($v2) return $v3
```

Fixes #1418


marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
```diff
@@ -68,48 +73,57 @@ object MapFuncPlanner {
...
+            isMap  <- ejson.isMap(m.xqy)
+          } yield {
+            let_(m -> src) return_ {
+              if_ (isMap)
+              .then_ { lookup }
+              .else_ { m.xqy `/` child(qn) }
+            }
           }
+        } getOrElse {
+          MonadError_[F, PlannerError].raiseError(InternalError(
```
jedesah: Is this really an internal error?

wemrysi: Nope, that was pretty much a placeholder, thanks for the reminder.

We need to decide if there are appropriate cases (or a catchall?) to add to `PlannerError` for backends to use to emit errors or to have backends introduce their own specific planner errors and handle them separately.

Maybe even something like an `UnsupportedPlan` case that allows supplying QScript instead of `LogicalPlan` since we don't have that in the planner? Or we could emit a placeholder error type and then turn it into `UnsupportedPlan` in the `QueryFile` implementation.

wemrysi: I guess I could just do the last option now and we can switch to something else later.

jedesah: That's fine by me.

marklogic-validation/src/main/java/org/apache/xerces/util/XMLChar.java
```diff
@@ -0,0 +1,1065 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
```
jedesah: I wonder if it's a good idea to have this here.

wemrysi: Oh, weird, I had included the original license header from the source file, looks like we have something that replaced it =/

wemrysi:
@jedesah Ready for another look.

PR #1424, opened 2016-09-09
Fix ineffectual `ShiftRead` transformation.

sellout:
To resolve the proper implicit, needed to add a “dummy” type parameter
to the `Aux` alias.

PR #1427, opened 2016-09-11
Tighten QScript coproduct types

sellout:
This is a bunch of changes that finally allow us to use smaller coproducts of QScript. This is the first stage of guarantees that invalid nodes won’t slip through (cf. #1424).

Unfortunately, we can’t yet apply this to branches, but we _promise_ (rather than guarantee) that types not in the current QScript coproduct won’t show up in branches either.

This is “in progress” because it does break some QScript tests, but I should have those fixed shortly.


sellout:
Ok, just moved this to in review.

@jedesah – I assigned you for a few reasons: 1. you’ve shown interest in working on QScript transformations, and this should give you a glimpse of some that are already there; 2. we’ve discussed the “shrinking coproducts” thing, which this accomplishes, 3. it’s a necessary step to the removal of Read explosion that you’re motivated to see added, and … there are probably more.

This is, unfortunately, bigger than I’d like. But both @alissapajer and I can help with anything that’s too unfamiliar.

alissapajer:
Build 5 has hung twice now in the middle of compiling (no output for 10 minutes causes build to fail). I'm running it a third time now.

alissapajer:
Build 5 (QUASAR POSTGRES) gets as far as
```
[info] Compiling 2 Scala sources and 1 Java source to /home/travis/build/quasar-analytics/quasar/marklogic-validation/target/scala-2.11/classes...
[info] Compiling 22 Scala sources to /home/travis/build/quasar-analytics/quasar/foundation/target/scala-2.11/classes...
[info] Compiling 3 Scala sources to /home/travis/build/quasar-analytics/quasar/ejson/target/scala-2.11/classes...
[info] Compiling 7 Scala sources to /home/travis/build/quasar-analytics/quasar/foundation/target/scala-2.11/test-classes...
[info] Compiling 9 Scala sources to /home/travis/build/quasar-analytics/quasar/effect/target/scala-2.11/classes...
[info] Compiling 5 Scala sources to /home/travis/build/quasar-analytics/quasar/js/target/scala-2.11/classes...
[info] Compiling 2 Scala sources to /home/travis/build/quasar-analytics/quasar/ejson/target/scala-2.11/test-classes...
[info] Compiling 1 Scala source to /home/travis/build/quasar-analytics/quasar/js/target/scala-2.11/test-classes...
[info] Compiling 27 Scala sources to /home/travis/build/quasar-analytics/quasar/frontend/target/scala-2.11/classes...
[info] Compiling 22 Scala sources to /home/travis/build/quasar-analytics/quasar/frontend/target/scala-2.11/test-classes...
[info] Compiling 2 Scala sources to /home/travis/build/quasar-analytics/quasar/sql/target/scala-2.11/classes...
[info] Compiling 50 Scala sources to /home/travis/build/quasar-analytics/quasar/connector/target/scala-2.11/classes...
[info] Compiling 3 Scala sources to /home/travis/build/quasar-analytics/quasar/sql/target/scala-2.11/test-classes...

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```
and then stops. I've seen this behavior 3 times now.

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -44,63 +44,117 @@ object QueryFile {
...
-  def convertToQScript[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, M[_]](
-    listContents: Option[ConvertPath.ListContents[M]])(
-    lp: T[LogicalPlan])(
-    implicit
-    merr: MonadError[M, FileSystemError],
-    mtell: MonadTell[M, PhaseResults]
-  ): M[T[QS[T, ?]]] = {
-    val transform = new Transform[T, QS[T, ?]]
+  def convertToQScript
+    [T[_[_]]: Recursive: Corecursive: EqualT: ShowT, QS[_]: Traverse: Normalizable]
```
jedesah: How is this used? If it's not common for clients to specify a larger `Coproduct`, then I would say just "hard code" it.

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -44,63 +44,117 @@ object QueryFile {
...
+      : EitherT[Writer[PhaseResults, ?], FileSystemError, T[QS]] = {
+    val transform = new Transform[T, QScriptInternal[T, ?]]
     val optimize = new Optimize[T]
 
-    val rawQS =
-      optimizeEval(lp)(optimize.applyAll).fold(
-        perr => merr.raiseError(FileSystemError.planningFailed(lp.convertTo[Fix], perr)),
-        merr.point(_))
+    val qs =
+      optimizeEval[T, QScriptInternal[T, ?]](lp)(optimize.applyAll).leftMap(FileSystemError.planningFailed(lp.convertTo[Fix], _)) ∘
```
jedesah: Why is this called `optimizeEval`? Maybe there is no better name, but if so then it should probably have a short sentence above the method to explain it's purpose because a quick look does make it obvious to me.

jedesah: It looks like it actually does the conversion from `LogicalPlan` to `QScript` and then optimizes `QScript`. Perhaps `convertToQScriptAndOptimize`?

connector/src/main/scala/quasar/fs/QueryFile.scala
jedesah: Oops, this was meant as a temporary comment that I was perhaps going to delete :smile:

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: Yeah, this is a good point. We do have to do some renaming – right now our normalization and optimization are a bit conflated. Really, all of our operations (`lpToQScript` `SimplifyJoins`, etc.) should be composed with `Optimize.applyAll` (which should be renamed to `Normalize`), so that really we have definitions like `def lpToQScript = Normalize.applyAll ⋘ lpToQScript0`.

I’ll at least come up with a way to restructure this for now.

alissapajer:
@sellout would it be possible/easy to extract out the part of this PR that fixes the giant `Union` of `Read` nodes? I'm still working on getting the build to build (current issue OOM as seen in #1443). And I'm wondering if we could get a fraction of this work past jenkins?

This is obviously silly, but since we're in a time crunch, I want backend authors (and qscipt unit test writers) to be able to use the `Read` fix ASAP.

We can do more work on jenkins with the PR as it is, but at some point we might need to do the above.

sellout:
@alissapajer That change isn’t in this PR. This is just the PR that needs to happen _before_ that PR.

alissapajer:
@sellout oh! So, I don't see a `Read` explosion when I run `select foo from bar` against my `autojoin3` bug fix including this PR, but I do see the explosion when I run `select foo from bar` against master. So I assumed this PR was directly fixing that.

sellout:
Ah, well there are multiple causes for Read explosion. One is just otherwise broken queries – e.g., the autojoin3 fix itself could be avoiding explosions, because it (perhaps) now preserves projections that were getting thrown away before. When the projections got thrown away (or put in the wrong place), Read conversion would be like “oh, I have this directory, I now have to turn it into the union of all the files under this directory” whereas when the projections are in the right place, Read conversion would be like “cool, I have a file, nothing more to do”

My PR that fixes Read explosion is more for the case where a _proper_ QS would correctly trigger the “I have to turn this directory into the union of all of its files” case. The PR turns that unioning into a separate operation, so MarkLogic in particular can skip doing it, since it’s happy to perform queries on directories directly.

So … any fixing of Read explosion that this PR fixes is purely incidental via fixing other problems with queries.

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: So, it depends on the caller. In the upcoming Read-explosion fixing PR, the Mongo backend does use a larger coproduct here, and filters it down (through transformations like `ShiftRead`) to just the components it wants.

It’s not _necessary_ to do it that way, and if it turns out we have to abandon abstraction for performance, this might be one of the things that has to change.

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: BTW, is this the only comment, or have you not finished reviewing? (Just wondering if I should push this change once we see that merging master made this successful, or if I should wait for more stuff to update.)

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -44,63 +44,117 @@ object QueryFile {
...
+
+    EitherT(Writer(
+      qs.fold(
+        κ(Vector()),
+        a => Vector(PhaseResult.Tree("QScript", a.cata(transform.linearize).reverse.render))),
+      qs))
+  }
+
+  def convertToQScriptRead
+    [T[_[_]]: Recursive: Corecursive: EqualT: ShowT, M[_], QS[_]: Traverse: Normalizable]
```
sellout: Because the one above it doesn’t use `listContents`, so it doesn’t need the connector’s monad at all.

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -44,63 +44,117 @@ object QueryFile {
...
+      (QScriptCore[T, ?] :\: ProjectBucket[T, ?] :\: ThetaJoin[T, ?] :/: Const[Read, ?])#M[A]
 
-    // TODO: Rather than explicitly applying multiple times, we should apply
-    //       repeatedly until unchanged.
-    val optimized =
-      merr.map(withoutProj)(_.transCata(optimize.applyAll).transCata(optimize.applyAll))
+    val qs =
+      merr.map(
+        merr.bind(
+          optimizeEval[T, QScriptInternal[T, ?]](lp)(optimize.applyAll).fold(
```
sellout: Yeah, that’s a fair cop. I’ll clean it up.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+import quasar.qscript.MapFunc._
+import quasar.qscript.MapFuncs._
+
+import matryoshka._
+import scalaz._, Scalaz._
+
+/** This handles coalescing anything preceded by a QScriptCore node.
+  */
+trait Coalesce[IN[_]] {
+  type IT[F[_]]
```
sellout: No, it’s just the parameter name in the not-yet-defined type function.

.travis.yml
```diff
@@ -12,7 +12,8 @@ script:
   - export MONGO_AUTH="MONGO_${MONGO_RELEASE}_AUTH"
   - ./scripts/installMongo mongodb-linux-x86_64-${!MONGO_LINUX} mongo ${!MONGO_PORT} ${!MONGO_AUTH}
   - psql -c 'CREATE DATABASE "quasar-test";' -U postgres
-  - ./scripts/build $MONGO_RELEASE
+  # FIXME: Need to improve compile times so we can remove `travis_wait`
+  - travis_wait 40 ./scripts/build $MONGO_RELEASE
```
alissapajer: @sellout You could add a comment here clarifying that we think this gives an _additional_ 40 minutes, which makes it 50 in total.

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -44,63 +44,117 @@ object QueryFile {
...
+  def optimizeEval
+    [T[_[_]]: Recursive: Corecursive: EqualT: ShowT, QS[_]: Traverse: Normalizable]
+    (lp: T[LogicalPlan])
+    (eval: QS[T[QS]] => QS[T[QS]])
+    (implicit
+      CQ: Coalesce.Aux[T, QS, QS],
+      DE:    Const[DeadEnd, ?] :<: QS,
+      QC:    QScriptCore[T, ?] :<: QS,
+      TJ:      ThetaJoin[T, ?] :<: QS,
+      PB:  ProjectBucket[T, ?] :<: QS,
```
alissapajer: +1 to these implicits - they make it easier to inject and whatnot

alissapajer: I'm still not satisfied with the ease of use of coproducts, but that's an aside out of scope of this PR. I think providing these implicits is an improvement, but in general for injecting, extracting, pattern matching, etc I still find it clunky. But that's a project for when we have more time :)

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+  type OUT[A]
+
+  def coalesce[F[_]: Functor]
+    (FToOut: PrismNT[F, OUT])
+    (implicit QC: QScriptCore [IT, ?] :<: OUT)
+      : IN[IT[F]] => Option[IN[IT[F]]]
+}
+
+object Coalesce {
```
alissapajer: @sellout I'm thinking (not in this PR) we could reorganize the `Optimizer.scala` file so that all the "within" component scope for a given component are in the same file. Then we'd have `QScriptCoreOptimize` etc.

Nothing to do with `Coalesce` just this file made me think of that...

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+      type IT[F[_]] = T[F]
+      type OUT[A] = G[A]
+
+      val FI = scala.Predef.implicitly[Injectable.Aux[QScriptCore[T, ?], QScriptTotal[T, ?]]]
+
+      def coalesce[F[_]: Functor]
+        (FToOut: PrismNT[F, OUT])
+        (implicit QC: QScriptCore [IT, ?] :<: OUT) = {
+        case Map(Embed(src), mf) => FToOut.get(src) >>= QC.prj >>= (s =>
+          if (mf.length ≟ 0 && (s match { case Unreferenced() => false; case _ => true }))
```
alissapajer: a one line pattern match within the conditional of another pattern match! blowing my mind... :)

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+
+import quasar.Predef._
+import quasar.fp._
+import quasar.qscript.MapFunc._
+import quasar.qscript.MapFuncs._
+
+import matryoshka._
+import scalaz._, Scalaz._
+
+/** This handles coalescing anything preceded by a QScriptCore node.
```
alissapajer: @sellout could you include in the comment here that this is where we include `Unreferenced` when we detect a zero length map func?

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+          from.resume.swap.toOption >>= FI.project >>= {
+            case Map(fromInner, mf) => Map(FToOut.reverseGet(QC.inj(Take(src, fromInner, count))).embed, mf).some
+            case _ => None
+          }
+        case Drop(src, from, count) => // Pull more work to _after_ limiting the dataset
+          from.resume.swap.toOption >>= FI.project >>= {
+            case Map(fromInner, mf) => Map(FToOut.reverseGet(QC.inj(Drop(src, fromInner, count))).embed, mf).some
+            case _ => None
+          }
+        case Filter(Embed(src), cond) => FToOut.get(src) >>= QC.prj >>= {
```
alissapajer: @sellout if the `src` of a `Filter` is a `Map` - couldn't we then rebase the `MapFunc` from the `Filer` on the `MapFunc` of the `Map`? Basically, add another case right below matching on `Map`.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+import matryoshka._
+import scalaz._, Scalaz._
+
+/** This handles coalescing anything preceded by a QScriptCore node.
+  */
+trait Coalesce[IN[_]] {
+  type IT[F[_]]
+  type OUT[A]
+
+  def coalesce[F[_]: Functor]
```
alissapajer: @sellout just questions on your naming of these types here:

(1) does `IT` refer to the thing we are coalescing, as in "coalesce it please"?
(2) I assumed at first that `OUT` would be the output (i.e. return) type - maybe change that to `EXT` for external? Or something else? Unless you have good reason to call it `OUT`?

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: If we had dotty with union types, we could use that instead. Otherwise, using subtyping, you end up with 2ⁿ - n (I think) superclasses in order to handle the various combinations – although, I think we could have ~12 superclasses to handle the useful combinations in our current model.

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: And we should `s/optimize/normalize/` at the same time 😉

sellout: (I think we only have one non-normalizing optimization at this point)

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: Yeah, this is one of those cases where we have to be careful not to unconditionally set the src to `Unreferenced`, because it would lead to infinite `repeatedly` as it keeps rewriting this node to its input.

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: You mean `Filter(Map(src, mf), cond)` => `Filter(src, cond >> mf)`?

We can’t do that because then the result set wouldn’t have the right data in it. If we only use the `mf` in the filter, we’ll exclude the right records, but then we would need to re-apply `mf` after the `Filter`, like `Map(Filter(src, cond >> mf), mf)`. Something like that could actually be more performant, if `cond >> mf` normalizes to a simpler operation, and the original `mf` could then be applied to a smaller data set … but we’re not making those sorts of optimizations yet.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+              } ∘ (Map(srcInner, _))
+            case _ =>
+              if (!shiftRepair.element(RightSide))
+                Map(src.embed, shiftRepair.as(SrcHole)).some
+              else None
+          }
+        // TODO: For Take and Drop, we should be able to pull _most_ of a Reduce repair function to after T/D
+        case Take(src, from, count) => // Pull more work to _after_ limiting the dataset
+          from.resume.swap.toOption >>= FI.project >>= {
+            case Map(fromInner, mf) => Map(FToOut.reverseGet(QC.inj(Take(src, fromInner, count))).embed, mf).some
```
sellout: Yeah, our datasets are always ordered (we just don’t know what that order is if an explicit sort hasn’t been applied somewhere).

So, the “simple” form of `Take` could be `case class Take(src: A, count: Long)`, but we allow the count to be generated by an arbitrary query, and its dataset is independent of the dataset we’re taking _from_ (which could be represented as `Take(src: A, count: A)`). So, like the other operations that operate on two data sets, it has a common source and then the branches for the two parts of the query that differ. I.e. `Take(src: A, from: FreeQS[T], count: FreeQS[T])`.

I wonder what `FoldFunc[T]` would look like. I do waffle on combining `Take` and `Drop` into a single operation – with maybe a simple enumeration to distinguish them. Then we could also add something like `Sample` (which would randomly select `count` from the dataset). Thoughts?

connector/src/main/scala/quasar/qscript/Coalesce.scala
alissapajer: You could also maybe leave an example of coalescing, something like if we have `Map` as the source of a `LeftShift` then in some cases we can coalesce the `Map` into the `LeftShift`

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: 1. `IT` is the `Internal T[_[_]]` type alias that we came up with at some point a few months ago (when we had to deal with this issue of needing a type member that was assigned from a type parameter)
2. Yeah, `OUT` is the output coproduct. I called it that because there is also an `IN` parameter for the input coproduct. I’m happy to use another pair of names – but I know people were getting confused with F/G/H/I/J everywhere.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -0,0 +1,176 @@
...
+                case RightSide => elem.some
+              } ∘ (Map(srcInner, _))
+            case Reduce(srcInner, _, List(ReduceFuncs.UnshiftMap(_, elem)), redRepair)
+                if struct ≟ HoleF && redRepair ≟ Free.point(ReduceIndex(0)) =>
+              shiftRepair.traverseM {
+                case LeftSide => None
+                case RightSide => elem.some
+              } ∘ (Map(srcInner, _))
+            case _ =>
+              if (!shiftRepair.element(RightSide))
```
alissapajer: @sellout in this case, even if we never reference the `RightSide`, don't we still need to keep around the `LeftShift`? 

This is claiming that `Map(src, f(SrcHole))` is semantically equivalent to `LeftShift(src, blah, f(LeftSide))`. Is that your intention? I thought `LeftShift` carried with it extra information to a backend?

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
```diff
@@ -0,0 +1,310 @@
...
+import quasar.contrib.pathy._
+import quasar.fp._
+import quasar.fs._
+import quasar.qscript.MapFuncs._
+
+import matryoshka._, Recursive.ops._, FunctorT.ops._
+import pathy.Path.{dir1, file1, FileName, rootDir}
+import scalaz._, Scalaz._, \&/._
+
+trait DiscoverPath[IN[_]] {
```
alissapajer: @sellout can you add some doc about this?

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: Yeah, you’re absolutely right – we need to have the right number of duplicates in the dataset, as introduced by the shift, even if the result of the shift is never used.

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
```diff
@@ -0,0 +1,310 @@
...
+
+  def discoverPath[M[_]: MonadFsErr](g: DiscoverPath.ListContents[M])
+      : AlgebraM[M, IN, List[ADir] \&/ IT[OUT]]
+}
+
+object DiscoverPath extends DiscoverPathInstances {
+  type Aux[T[_[_]], IN[_], F[_]] = DiscoverPath[IN] {
+    type IT[F[_]] = T[F]
+    type OUT[A] = F[A]
+  }
```
alissapajer: @sellout the types are so abstract here, I can't read this and say what it does. Are there any types that could be concrete?

`List[ADir]` is the only type that tells one anything...

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
sellout: I’ll add doc, but just for immediate clarification – this is the new name for what was previously split into `StaticPath` and `ConvertPath`. As 1. we no longer need them separate, and 2. keeping them separate tied our hands with regard to tightening coproducts.

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
```diff
@@ -0,0 +1,310 @@
...
+    implicit R: Const[Read, ?] :<: F,
+             QC: QScriptCore[T, ?] :<: F):
+      ADir => M[List[F[T[F]]]] =
+    dir => (listContents(dir) >>=
+      (ps => ISet.fromList(ps.toList).toList.traverseM(_.fold(
+        d => allDescendents[T, M, F](listContents).apply(dir </> dir1(d)) ∘ (_ ∘ (wrapDir(d.value, _))),
+        f => List(wrapDir[T, F](f.value, makeRead(dir, f))).point[M]))))
+      .handleError(κ(List.empty[F[T[F]]].point[M]))
+
+  def unionDirs[T[_[_]]: Corecursive, M[_]: MonadFsErr, OUT[_]: Functor]
```
alissapajer: @sellout should this be `private` or do you expect it to be called by someone else?

alissapajer: Same for `allDescendents` above - maybe just audit this file for what can be `private`.

connector/src/main/scala/quasar/qscript/ShiftRead.scala
```diff
@@ -127,12 +127,25 @@ object ShiftRead extends ShiftReadInstances {
...
-abstract class ShiftReadInstances {
-  implicit def inject[T[_[_]], F[_]: Functor, I[_]](implicit F: F :<: I): ShiftRead.Aux[T, F, I] =
+  def default[T[_[_]], F[_]: Functor, I[_]](implicit F: F :<: I):
+      ShiftRead.Aux[T, F, I] =
     new ShiftRead[F] {
       type G[A] = I[A]
       def shiftRead[H[_]](GtoH: G ~> H) = λ[F ~> FixFreeH[H, ?]](fa => GtoH(F inj (fa map Free.point)))
     }
+
+  implicit def deadEnd[T[_[_]], F[_]](implicit DE: Const[DeadEnd, ?] :<: F)
```
alissapajer: @sellout why weren't these implicits required before?

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
sellout: Well, `IN` and `OUT` are the input and output coproducts. We could maybe add package doc to `quasar.qscript` that mentions some of the commonly used type parameter names. But this is an algebra from some coproduct to another coproduct, where we collect statically-known directories along the way, and it’s in some connector-specific monad used by ListContents.

foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -299,6 +299,19 @@ trait QFoldableOps {
   }
 }
 
+trait DebugOps {
+  final implicit class ToDebugOps[A](val self: A) {
+    /** Applies some operation to a value and returns the original value. Useful
+      * for things like adding debugging printlns in the middle of an
+      * expression.
+      */
+    final def <|(f: A => Unit): A = {
```
alissapajer: 💯

alissapajer:
@sellout ok I'm done with my initial review.

connector/src/main/scala/quasar/qscript/Coalesce.scala
alissapajer: Yeah, after this lands, we can rename stuff!

connector/src/main/scala/quasar/qscript/ShiftRead.scala
sellout: Because `default` (née `inject`) was implicit itself before. But that caused a ton of subtle bugs that were very confusing. E.g., if some constraint was missing somewhere, the correct implicit would be missed, but the low-prio `inject` implicit would be matched, so it would look like it worked, but the transformation wouldn’t actually be applied, since `inject` just injects the component into the output coproduct without doing any work. So now `default` is not implicit, and we have specific implicits for the cases that are properly handled that way.

connector/src/main/scala/quasar/qscript/Coalesce.scala
alissapajer: @sellout Right, I see that now. The source of the `Filter` needs to (effectively) stay the same so that we end up with the correct results.

We could do this though: if we have `Filter(Map(src, Add(_, 5)), _ > 7)` then that becomes `Filter(src, _ > 7-5)` == `Filter(src, _ > 2)`. I'm not sure we ever would want to do that because that involves evalution and determining that the inverse of add is subtract. But maybe!

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: Hrmm, I feel like the duplication is pretty minimal, with the common parts pulled into `optimizeEval` and `normalizeQScript`, but I’m very open to suggestions as to how to extract more.

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: Yeah, and `Coalesce` might not be the best name for this – that’s just the name that had been given to the pieces of this when it was multiple operations in Optimize.scala. But I’ll try to describe it as best as I can :D

connector/src/main/scala/quasar/qscript/Coalesce.scala
alissapajer: Regarding a `FoldFunc[T]`, I wonder if some `ReduceFunc[T]` would actually be a fold? `Count`, `Sum`, `Min`, `Max`, and `Avg` are all folding operations, so maybe there's a generaliztion including those, as well as `Take` and `Drop`? Maybe `Take` and `Drop` could be funcs themselves instead of standalone qscript nodes.

`Sample` strikes me as different. If you're actually going to take random set elements, then we'd probably want to provide a seed value so it would be reproducible. Can `Sample` really be a fold? My first instinct is that it can't be a single pass fold unless you know the size of the entire dataset apriori. Because if you don't take enough elements during the beginning of your pass, you might run out!


I don't think we could implement `Sample` with a fold, unless we know ahead of time the total size of the dataset.

connector/src/main/scala/quasar/qscript/Coalesce.scala
alissapajer: OH it's _that_ `IT`!

Let's stick with `IN` and `OUT` for now. If I think of something better I'll run it through the #qscript channel later.

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
alissapajer: Right. I just think a sentence or two in each file is useful for someone jumping into the file to have a general sense of what they're about to get into :)

connector/src/main/scala/quasar/qscript/DiscoverPath.scala
alissapajer: Yeah, we should document `IT` somewhere, not in this PR is fine. But readability of code is important to me. So I think documenting commonly used aliases is important. I'd like a way to enforce that we use them consistently. For example we have this `Aux` pattern in at least a couple places, so I'd like to enforce that we use `IT` and `OUT` in all those places.

I think sometimes it's hard to read through the coproduct boiler plate until one is used to it.

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: The distinction between reductions and Take/Drop is that the former operate on a single dataset, whereas the latter operate over two data sets. I’m not hopeful that they can be unified – but I think it would be good to do if there is a way.

We could have sample with a seed – e.g. `case class Extract(src: A, from: FreeQS[T], count: FreeQS[T], ex: Extraction)` and `data Extraction = Take | Drop | Sample Long` (sorry, mixing Scala and Haskell for brevity).

You can randomly sample in a fold – there are a bunch of approaches, but if you’re sampling a single element, basically, pick the first one, then 50% of the time replace it with the second, then 33% of the time replace that with the third, etc.

However, even if we have a FoldFunc, I don’t think it could reasonably impose that sort of constraint on the connector. E.g., in Mongo, it would compile down to `$sample` – and who knows what that would return. I doubt Mongo has a way to provide a seed, even (perhaps it could be `Sample (Option Long)`), so if it’s not explicitly seeded, we can use Mongo’s `$sample`, otherwise, we fall back to in-mem sampling. Anyway, getting a bit far afield here.

connector/src/main/scala/quasar/qscript/Coalesce.scala
sellout: In the Mongo connector, we do have some places where we `flip` operations like that, `+ => -`, etc. There it’s because otherwise a simple operation would become JS, but I think it could be useful to move that to the MapFunc realm and make it more generally available, and it could be used for an optimization like that.

connector/src/main/scala/quasar/qscript/Coalesce.scala
alissapajer: Oh yeah I've totally thought about that sampling algorithm before!! :)

alissapajer:
@sellout reviewed the latest commit, and it LGTM. I'll merge once jenkins gives it the green.

alissapajer:
@sellout the first build mongo 2.6 was killed during scaladoc stuff (error `137`). :( I haven't seen problems in any other PRs since #1459 landed - any ideas why this PR would be behaving differently, or is it just coincidence?

alissapajer:
The scaladoc stuff shouldn't be timing out. Timing out in `connector` makes sense because we know how long that takes to compile. But I think the scaladoc issue is something totally different, which is probably why `travis_wait` didn't fix it. 

I think it's possible that the build ran out of memory in the scaladoc stuff, and that message somehow got swallowed by `travis_wait`... we saw OOM errors in scaladoc before - did those get resolved?

sellout:
I’m pretty sure this is just the same process hitting the `travis_wait` timeout farther along. It’s a hard timeout. If your process runs past it, it gets killed, regardless of whether it’s outputting anything.

alissapajer:
@sellout oh yeah, we did talk about that - we'll see what happens this time!

alissapajer:
I'd like to see it print out "travis_wait timeout hit. Killing build."

sellout:
I’d like it to just behave as documented ;)

alissapajer:
I can't argue with that! :)

PR #1428, opened 2016-09-12
Put the new kind-projector polymorphic lambdas to work!

paulp:


core/src/main/scala/quasar/fs/package.scala
```diff
@@ -74,18 +74,14 @@ package object fs extends PhysicalErrorPrisms {
   /** Rebases absolute paths onto the provided absolute directory, so
     * `rebaseA(/baz)(/foo/bar)` becomes `/baz/foo/bar`.
     */
-  def rebaseA(onto: ADir): AbsPath ~> AbsPath =
-    new (AbsPath ~> AbsPath) {
-      def apply[T](apath: AbsPath[T]) =
-        apath.relativeTo(rootDir[Sandboxed]).fold(apath)(onto </> _)
-    }
+  def rebaseA(onto: ADir) = λ[EndoM[AbsPath]](apath =>
```
jedesah: Where does `EndoM` come from? I text searched `quasar` and `scalaz` and nothing came up.

sellout: It’s added in this PR. Actually, I was wondering if `EndoK` might be a better name – in the vein of Cats. Or even `EndoF` in the vein of Haskelly things that are `F ~> G` rather than `A ~> B`.

paulp: I have no attachment whatsoever to the fifth letter. You pick.

paulp: Actually it occurs to me that `EndoF` looks like a weirdly capitalized `end-of`. So I like `EndoK` the best. (At some point I suggested to cats they should rename NaturalTransformation to FunctionK.)

core/src/main/scala/quasar/qscript/ShiftRead.scala
```diff
@@ -55,95 +59,80 @@ object ShiftRead extends ShiftReadInstances {
...
-              Free.liftF(GtoH(SR.inj(Const[ShiftedRead, A](ShiftedRead(
-                read.getConst.path,
-                IncludeId))))),
-              NullLit(),
-              List(ReduceFuncs.UnshiftMap(
-                Free.roll(ProjectIndex(HoleF, IntLit(0))),
-                Free.roll(ProjectIndex(HoleF, IntLit(1))))),
-              Free.point(ReduceIndex(0)))))
-        }
+      def shiftRead[H[_]](GtoH: G ~> H) = λ[Const[Read, ?] ~> FixFreeH[H, ?]](read =>
```
jedesah: Is it not possible to use `>>` here? Since we went through the trouble of defining it.

paulp: Extremely regrettably, it is not! It no longer compiles.

paulp: (It is almost certainly a dealiasing bug in scalac.)

effect/src/main/scala/quasar/effect/Read.scala
```diff
@@ -79,39 +76,9 @@ object Read {
...
-      }
-    }
-
-  def toState[F[_], R](implicit F: MonadState[F, R]): Read[R, ?] ~> F =
-    new (Read[R, ?] ~> F) {
-      def apply[A](ra: Read[R, A]) = ra match {
-        case Ask(f) => F.gets(f)
-      }
-    }
+  def contramapR[Q, R](f: Q => R)                      = λ[Read[R, ?] ~> Read[Q, ?]] { case Ask(g) => Ask(g compose f) }
```
jedesah: This is one of the most impressive example of boilerplate slaying! 😄

jedesah:
LGTM. Will merge unless @sellout has any objections once the build passes.

effect/src/main/scala/quasar/effect/Read.scala
paulp: It's nice that you can in completely clear conscience omit the return type on the method - because the lambda desugaring always translates to `new <type argument to lambda>`, that will always be exactly the return type.

core/src/main/scala/quasar/qscript/ShiftRead.scala
paulp: Oh and by the by, that's a very good observation. You've earned the "conscientious code reviewer" badge!

paulp:
For the record the build already passed, then I mechanically renamed EndoM to EndoK. It's identical to the passed build "up to renaming".

paulp:
Not certain, but I believe those test failures are the same ones which will be fixed upon merging #1429. (Feel free to review that one too!)

paulp:
🎉

paulp:
Thanks!

PR #1432, opened 2016-09-12
Constant fold

alissapajer:


alissapajer:
@sellout impl of constant folding, with a few cases implemented. This is essentially an evaluator for constant values. As we see fit, we can easily add more cases.

core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -164,20 +164,66 @@ object MapFunc {
...
+    _.run.fold(
+      κ(None),
+      {
+        // relations
+        case And(
+            Embed(CoEnv(\/-(Constant(Embed(ejson.Common(ejson.Bool(v1))))))),
+            Embed(CoEnv(\/-(Constant(Embed(ejson.Common(ejson.Bool(v2)))))))) =>
+          CoEnv[A, MapFunc[T, ?], T[CoEnv[A, MapFunc[T, ?], ?]]](
+            Constant[T, T[CoEnv[A, MapFunc[T, ?], ?]]](
+              CommonEJson.inj(ejson.Bool[T[EJson]](v1 && v2)).embed).right).some
```
sellout: It’s not important for this PR, but constant folding should be able to fail … or at least return `Undefined`. So, in this case, something like
```scala
case And(
  Embed(CoEnv(\/-(Constant(Embed(const1))))),
  Embed(CoEnv(\/-(Constant(Embed(const2)))))) =>
  (ejson.Common.prj(const1) ⊛ ejson.Common.prj(const2))(…).getOrElse(CoEnv(Undefined().right))
```
We don’t want to fail _immediately_, because it’s possible that something like `IfUndefined(x, y)` catches the statically-known `Undefined()` and gets rewritten to `y`.

core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -164,20 +164,59 @@ object MapFunc {
         })
   }
```
paulp: I think we must fight harder against duplication. The fact that scala requires so much handholding makes the penalty for unnecessarily copying the same parameters around vastly higher. The time comes when this must be changed, and it's simply too difficult. 

It took me rather a long time to extract the following from the function I found, mostly puzzling out the types. But it is faster to write it this way up front, especially if you're familiar with the types. And even if it weren't, the maintenance difference is enormous.

This is not tested - I only confirmed that it compiles.
```scala
  // Transform effectively constant `MapFunc` into a `Constant` value.
  // This is a mini-evaluator for constant qscript values.
  def foldConstant[T[_[_]]: Recursive: Corecursive, A]: CoMFR[T, A] => Option[CoMFR[T, A]] = {
    def resultExpr(result: ejson.Common[T[EJson]]): Option[CoMFR[T, A]] =
      scala.Some(CoEnv(Constant((CommonEJson inj result).embed).right))

    object EjConst {
      def unapply[B](x: T[CoMF[T, B, ?]]) = x match {
        case Embed(CoEnv(\/-(Constant(Embed(ejson.Common(x)))))) => Some(x)
        case _                                                   => None
      }
    }

    _.run.fold(
      κ(None),
      {
        // relations
        case And(EjConst(ejson.Bool(v1)), EjConst(ejson.Bool(v2))) => resultExpr(ejson.Bool(v1 && v2))
        case Or(EjConst(ejson.Bool(v1)), EjConst(ejson.Bool(v2)))  => resultExpr(ejson.Bool(v1 || v2))
        case Not(EjConst(ejson.Bool(v1)))                          => resultExpr(ejson.Bool(!v1))
        case Lower(EjConst(ejson.Str(v1)))                         => resultExpr(ejson.Str(v1.toLowerCase))
        case Upper(EjConst(ejson.Str(v1)))                         => resultExpr(ejson.Str(v1.toUpperCase))
        case MakeArray(Embed(CoEnv(\/-(Constant(v1)))))            => resultExpr(ejson.Arr(List(v1)))
        case _                                                     => None
      })
  }
```

sellout: Yeah, we actually already have some things like this, so
```scala
case And(BoolLit(v1), BoolLit(v2)) => BoolLit(v1 && v2)
```
should work.

sellout: Actually, those are all defined on `Free`, not `CoEnv` … which distinction would disappear with slamdata/matryoshka#21 – but they could have multiple unapplies in the mean time.

paulp: Overloading unapply is bad news. If you mean multiple unapplies by way of distinct objects, that's good. They can presumably inherit the same implementations in slightly variant forms.

alissapajer: @paulp thanks for the figuring, I'll update the PR with this or something similar.

paulp: @alissapajer no problem.

alissapajer:
Also, tests need to be updated since this work affects their results! So I'll do that too.

mossprescott:
It seems like this is re-implementing the constant-folding that's done at the LP level (in the typers and simplifiers). @sellout and I had some discussion about this recently, and I believe the idea was that the MapFuncs and the LP Funcs that they correspond to would be unified sometime soon and at that point we would be able to use the typers and/or the simplifiers on both the LP and QS representations. I'd rather see us working in that direction as opposed to adding another implementation.

One way or the other, this another implementation of the standard library, so I think we should attempt to implement StdLibSpec for this. I think that could be done with the MapFuncs as is since the test LPs can be converted to QS.

sellout:
@mossprescott I agree with this in principle, but practically there are a few things that make this necessary.

1. we can’t change LP without changing the existing Mongo connector that is about to be decommissioned
2. there are operations on QScript that can’t wait until after #1 happens.

Implementing StdLibSpec is probably a good idea. How does it deal with folding that we currently punt on (i.e., where we just return the original folded form)? Actually, I guess we return `None` in that case, so it should be easy to handle.

core/src/main/scala/quasar/qscript/Normalizable.scala
```diff
@@ -29,7 +29,8 @@ import simulacrum.typeclass
 
   def normalizeMapFunc[T[_[_]]: Recursive: Corecursive: EqualT, A](fm: Free[MapFunc[T, ?], A]):
       Free[MapFunc[T, ?], A] =
-    freeTransCata[T, MapFunc[T, ?], MapFunc[T, ?], A, A](fm)(repeatedly(MapFunc.normalize))
+    freeTransCata[T, MapFunc[T, ?], MapFunc[T, ?], A, A](fm)(
+      repeatedly(MapFunc.normalize[T, A]) compose repeatedly(MapFunc.foldConstant[T, A]))
```
sellout: We probably want `orOriginal(foldConstant)` here instead of `repeatedly`. Once we fold a constant, it can’t be folded again.

mossprescott:
@sellout: I think we can change LP and update the old connector as long as the fallout is fairly superficial. You may have thought about that enough to realize how naive that sounds. And in any case, if now's not the right time, then I understand that. Maybe the right call is to test what we can and remember to come back to this.

I'd be happy to help get the test working.

core/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -164,20 +164,62 @@ object MapFunc {
...
+      Some(CoEnv(Constant(EJson.fromExt.apply(result)).right))
+
+    object EjConstCommon {
+      def unapply[B](tco: T[CoMF[T, B, ?]]): Option[ejson.Common[T[EJson]]] = tco match {
+        case Embed(CoEnv(\/-(Constant(Embed(ejson.Common(v)))))) => Some(v)
+        case _                                                   => None
+      }
+    }
+
+    object EjConstExtension {
```
alissapajer: @paulp so the two extractors here turned out to not force an exponentially growing patern match as I thought they might, so this solution will be okay, at least for the interim.

alissapajer:
@sellout I've pushed all the changes I have planned, fixing the broken tests and addressing the code duplication.

@mossprescott what matters from the qscript point of view (and what matters for backend authors) is that this change can be accessible as soon as possible, and that when I need to add other constant-evaluable cases, I can add them easily. So, that's why I'd like this merged in the short term.

In the long term, perhaps it could make sense for the mini-constant-evaluator to be applied to a LP instead of a QS. But the problem I see with that, is that there are qscript transformations that could create a pattern that should be mini-evaluated, even after all such constructs were mini-evaluated in the LP. And so in that case, we'd still need a qscript transformation to evaluated that to a constant value.

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -295,21 +279,12 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
-              Free.roll(MakeArray(Free.roll(Constant(ExtEJson.inj(ejson.Int[Fix[ejson.EJson]](7)).embed)))),
-              Free.roll(MakeArray(Free.roll(Constant(ExtEJson.inj(ejson.Int[Fix[ejson.EJson]](8)).embed)))))),
-            Free.roll(MakeArray(Free.roll(Constant(ExtEJson.inj(ejson.Int[Fix[ejson.EJson]](9)).embed)))))),
+          Free.roll(Constant(
+            EJson.fromCommon[Fix].apply(ejson.Arr[Fix[ejson.EJson]](List(
+              EJson.fromExt[Fix].apply(ejson.Int[Fix[ejson.EJson]](7)),
+              EJson.fromExt[Fix].apply(ejson.Int[Fix[ejson.EJson]](8)),
+              EJson.fromExt[Fix].apply(ejson.Int[Fix[ejson.EJson]](9))))))),
           Free.point(RightSide)))).some)
-        // TODO optimize to eliminate `MakeArray`
```
alissapajer: These `TODO`s are fixed which is cool.

sellout:
@alissapajer We already apply constant folding to LP, but yes, the problem is that QScript then creates new things that need to be folded. The future plan is that MapFunc is used by both QScript and LP, and so whatever constant folding exists can be used both in LP and QScript contexts.

alissapajer:
@sellout ah perfect, then we will only have one mini-evalutor (just for `MapFunc`) instead one for LP and one for QS. I would still like this merged so we can have the benefits immediately.

Issue #1308, opened 2016-07-29
MongoDB: support for version-specific aggregation and group operators


mossprescott:
- [x] refactor `ExprF`to use coproducts as in `WorkflowF`
- [x] track those types in the planner
- [x] switch the types based on the runtime version of the MongoDB backend

mossprescott:
See partial implementation in https://github.com/mossprescott/quasar/commit/0ca228db0f5a7d6e4ec80dfd0f2f689760384d5f

mossprescott:
Calling this done with #1434 and will track accum. ops in another issue (#1311).

PR #1434, opened 2016-09-13
Implement ExprOps for new operators in MongoDB 3.0 and 3.2

mossprescott:
This introduces new types for the new operators, assembled into Coproducts mostly along the lines seen in `WorkflowF`.

See #1308.

Workflows and everything that uses them do not yet know about the new types; they just hard-code the new "core" (2.6) type. I'm working on that now, but it seems worthwhile to get this merged as is.

Once again, I'm not very happy with the names here, so I'm very open to suggestions. In particular, `EX` is a bit awkward. So are the expression traits: `ExprOpCoreF`, `ExprOp3_0F`, `ExprOp3_2F`, and coproducts: `Expr2_6`, `Expr3_0`, `Expr3_2`, and `ExprOp`.

Of general interest are new types `Inj` and `Prj` which each handle one half of `Inject`, are implicitly available whenever `Inject` is, and can be composed. They may be useful in QScript where I understand raw `~>`s are being used now. 

I notice that the new kind-projector syntax has been merged, so I will be updating some `~>`s to use it.


foundation/src/main/scala/quasar/fp/Inj.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package quasar.fp
+
+import quasar.Predef.Option
+
+import scalaz._
+
+/** Half of `Inject`, but composable. An instance is available whenever the
+  * corresponding `Inject` is. */
```
paulp: Can you elaborate on the problem this is attacking? Why isn't Inject composable?

mossprescott: You can see it in use here: https://github.com/quasar-analytics/quasar/pull/1434/commits/a4dec490c87b7a374a00bd04af8df2a46fe933a6#diff-a0172a6b99e1f93f910b5a63a1011028R52.

Essentially, I need (the equivalent of) `Inject[F, H]` where I have evidence for `Inject[Coproduct[F, G], H]`. It's obvious such an injection is possible, but the `Inject` type is closed and there was no way I could see to construct it. @sellout ran into a similar problem with the QScript types and solved it "manually" which seemed too cumbersome for these highly repetitive declarations.

paulp:
I think you can eliminate a lot of duplication among the extractors if you do something like this.
```scala
// R is the extractor return type
trait CompanionF[R] {
  def unapply[EX[_], A](expr: EX[A])(implicit prj: Prj[ExprOpCoreF, EX]): R
}
trait Companion[R] {
  protected def companion: CompanionF[R]
  def unapply[T[_[_]]: Recursive, EX[_]: Functor](expr: T[EX])(implicit prj: Prj[ExprOpCoreF, EX]): R =
    companion.unapply(Recursive[T] project expr)
}
```
Now `$addF extends CompanionF`, `$add extends $Companion`, and you just point `$add` at `$addF`. You can even do it like 
```scala
abstract class Companion[R](companion: CompanionF[R])  { ... same unapply as above }
// In which case $add becomes
object $add extends Companion($addF)
```

foundation/src/main/scala/quasar/fp/Inj.scala
sellout: My “manual” solution was to just use `implicit F: F ~> G` instead of `implicit F: F :<: G` (which is bad form, since natural transformations aren’t type classes). But since then I’ve actually managed to eliminate it completely. I’ll take a look at your case, see if I can figure out how to make it go away.

paulp: That would be awesome. There's a real cost to the proliferation of very similar classes and the bigger the code footprint gets, the harder it is to control yet further proliferation of such classes.

sellout: Yeah, so I’m thinking that any need for `Coproduct :<: X` is a code smell. We should be able to work with arbitrary coproducts that may contain the same components in different orders.

mossprescott:
@psp: thanks, that's a neat trick for those declarations. It's probably worth it, and I should probably add extractors for the rest of the case classes since they'll be more manageable.

I'd feel even better about this sort of thing if we could standardize it across the codebase, but it looks like it won't work for the somewhat-similar WorkflowF type since I'm now adding another type parameter. Not sure about the QScript types. Hmm.

foundation/src/main/scala/quasar/fp/Inj.scala
paulp: Probably we should take a close look at [https://github.com/ProjectSeptemberInc/freek](freek) then.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
```diff
@@ -0,0 +1,73 @@
...
+}
+object ExprOpOps {
+  def apply[EX[_]](implicit ops: ExprOpOps[EX]) = ops
+
+  implicit def coproduct[F[_], G[_]](implicit
+      F: ExprOpOps[F],
+      G: ExprOpOps[G])
+      : ExprOpOps[Coproduct[F, G, ?]] =
+    new ExprOpOps[Coproduct[F, G, ?]] {
+      def injF[H[_]](implicit I: Inj[Coproduct[F, G, ?], H]): Inj[F, H] = I compose Inj[F, Coproduct[F, G, ?]]
```
sellout: So, the way I’ve eliminated this is to move constraints to the _instances_ rather than the methods. This is really helpful for this case, because the `coproduct` instance doesn’t need an `inj` constraint, only the “leaf” instances do. This also means that `F` needs to become a type member, rather than a type parameter on the method.

It ends up looking like:
```scala
trait ExprOpOps[IN[_]] {
  type OUT[A]
  def simplify: AlgebraM[Option, IN, Fix[OUT]]
  ...
}

implicit def coproduct[F[_], G[_], H[_]]
  (implicit F: ExprOpOps[F], G: ExprOpOps[G])
    : ExprOpOps.Aux[Coproduct[F, G, ?], H] =
  new ExprOpOps[Coproduct[F, G, ?]] {
    type OUT[A] = H[A]
    def simplify = _.run.fold(F.simplify, G.simplify)
    ...
}
```
In this case, you have some operations that will need the constraint and others that don’t. You may want to pull them into separate type classes at some point, but seems fine to keep them together currently.

Note that the big win here is that you can now `simplify` from `Coproduct[F, G, ?]` to `Coproduct[G, F, ?]`.

In QScript, I’ve started only naming coproducts in very small scopes only at the edges (I wouldn’t name them at all, but this is Scala, so we need annotations throughout). Each backend defines its own QScript coproduct, and we don’t want to force them into a particular ordering.

paulp:
There are many forms of boilerplate which you can't beat this way, for reasons like adding a type parameter or an implicit or whatever. In those cases I would strongly consider macro annotations, especially at this volume.

mossprescott:
Less excited about that Companion trick after trying to apply it to the various shapes we have:
- no args
- one non-recursive argument
- 1, 2, or 3 recursive arguments
- 2+ var-arg-style recursive args
- 2 recursive with 1 non-recursive arg in the middle
- one non-rec and 1 rec.
- complex, recursive args: `ListMap[..., A], A`

So, that's 9 shapes in all, requiring 9 separate Companion/CompanionF trait pairs. So I'm probably going to skip it. Unless I'm missing something.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
mossprescott: Thanks, I wil have to take a closer look at this.

paulp:
Requires metaprogramming or code generation then. Still important to do, but I understand it's not going to happen right at this moment.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
```diff
@@ -25,50 +25,55 @@ import quasar.jscore, jscore.JsFn
...
   }
 }
 object ExprOpOps {
-  def apply[EX[_]](implicit ops: ExprOpOps[EX]) = ops
+  /** Useful in implementations, when you need to require an instance with a
+    * certain "output" type. */
+  type Aux[IN[_], F[_]] = ExprOpOps[IN] { type OUT[A] = F[A] }
+
+  /** For the typical use case where you want the in/out parameters to be the same. */
+  type Uni[F[_]] = Aux[F, F]
```
mossprescott: @sellout: Not completely sure I'm following the rationale here, but this is what I came up with that works, isn't too cumbersome, and allows me to eliminate `Inj`/`Prj` (in the next commit).

I felt that this simpler type was needed to hide the second parameter from the users, who will always have it the same. As usual, it's not a great name.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
```diff
@@ -25,50 +25,55 @@ import quasar.jscore, jscore.JsFn
...
 
-  implicit def coproduct[F[_], G[_]](implicit
-      F: ExprOpOps[F],
-      G: ExprOpOps[G])
-      : ExprOpOps[Coproduct[F, G, ?]] =
+  implicit def apply[F[_]](implicit ops: ExprOpOps.Aux[F, F]): ExprOpOps.Uni[F] = ops
```
mossprescott: So this seems to work, but leaves me with a vague feeling of dread somehow. Without it, implicit `Uni`s won't be resolved, and it also handles the "summoning" use case.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
wemrysi: Re the name, how about `Endo` as in endomorphism?

mongodb/src/main/scala/quasar/physical/mongodb/expression/package.scala
```diff
@@ -17,444 +17,95 @@
...
+    {
+      // matches the pattern the planner generates for converting epoch time
+      // values to timestamps. Adding numbers to dates works in ExprOp, but not
+      // in Javacript.
+      case $add($literal(Bson.Date(inst)), r) if inst.toEpochMilli ≟ 0 =>
+        Recursive[T].para(r)(toJs[T, EX])
+          .map(r => JsFn(JsFn.defaultName,
+            jscore.New(jscore.Name("Date"), List(r(jscore.Ident(JsFn.defaultName))))))
+
+      case Check(expr, typ) => typ match {
```
wemrysi: :tada:

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -884,14 +895,16 @@ object WorkflowBuilder {
...
-      case CollectionBuilderF(_, _, s2) => s2.map(s => Subset(s.toSet))
-      case DocBuilderF(_, shape) => Subset(shape.keySet).some
-      case FlatteningBuilderF(src, _) => findKeys(src)
-      case GroupBuilderF(src, _, Expr(\/-($$ROOT))) => findKeys(src)
-      case GroupBuilderF(_, _, Doc(obj)) => Subset(obj.keySet).some
+      case CollectionBuilderF(_, _, s2)       => s2.map(s => Subset(s.toSet))
+      case DocBuilderF(_, shape)              => Subset(shape.keySet).some
+      case FlatteningBuilderF(src, _)         => findKeys(src)
+      case GroupBuilderF(src, _,
+            Expr(\/-($$ROOT)))     => findKeys(src)
```
wemrysi: I only ask as you went to the trouble to align everything, any reason this and line 906 are on their own lines?

mossprescott: Seems like this one was spoiling the alignment, but that doesn't explain the extra spaces here, or the other one being split. I'll take another look.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
wemrysi: Hm, interesting, so are implicit `Aux[F, G]`s resolved?

wemrysi:
@mossprescott Some minor comments, but LGTM in general given functionality is preserved and the rest is a largely mechanical transformation into fixpoints of coproducts.

I don't have any better suggestions for the names you mentioned in the description.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOpOps.scala
mossprescott: Yes, see a couple of lines below.

wemrysi:
@mossprescott Also just noticed this needs a changelog =/ Let me know if you planning on making any changes, otherwise once the changelog hits I'll merge.

mossprescott:
@wemrysi: this should be ready to go, pending a successful build. I just restarted #1 which had "no output for 10 minutes."

wemrysi:
Ok, I'll watch for it...not unlike a hawk.

wemrysi:
@mossprescott Grr, looks like it failed again for the same reason :disappointed:

mossprescott:
Finally looked at my fork and I see the builds are failing in a similar way on the branches, so I'm working on the theory that that my changes made scaladoc a lot slower.

PR #1435, opened 2016-09-14
Extract “frontend” project from “core”.

sellout:
This project is conceptually something we want (so we can eventually
plug in multiple languages besides SQL²), but right now it definitely
has some extra bits in it. Including code that should become a separate
“sql” project.

Additionally, this introduces new packages with the names
`quasar.contrib.pathy` and `quasar.contrib.shapeless`. This is a
convention for where we put code that we would like to push upstream
– rather than hiding it in `quasar.fp` or just whereever we happen to be
coding at the moment we need it.


sellout:
This is also a step toward extracting a `connector` subproject. There are some tangly dependencies, and this way `connector` can (temporarily) depend on `frontend`, rather than being tangled in `core`. And it means the actual connector subprojects will no longer depend on `core`.

core/src/main/scala/quasar/fs/package.scala
```diff
@@ -21,43 +21,11 @@ import quasar.effect.Failure
...
-  object APath {
-
-    implicit val aPathDecodeJson: DecodeJson[APath] =
-      DecodeJson.of[String] flatMap (s => DecodeJson(hc =>
-        posixCodec.parseAbsFile(s).orElse(posixCodec.parseAbsDir(s))
-          .map(sandboxAbs)
-          .fold(DecodeResult.fail[APath]("[T]AbsPath[T]", hc.history))(DecodeResult.ok)))
-
-  }
-
```
sellout: The deleted stuff here got moved to foundation/…/quasar/contrib/pathy

core/src/main/scala/quasar/package.scala
```diff
@@ -112,31 +108,4 @@ package object quasar {
...
-      G.map(self.unsized.sequence(ev, G))(bs => Sized.wrap[List[B], N](bs))
-
-    def zip[B](input: Func.Input[B, N]): Func.Input[(A, B), N] =
-      Sized.wrap[List[(A, B)], N](self.unsized.zip(input))
-
-    def unzip3[X, Y, Z](implicit ev: A === (X, (Y, Z))): (Func.Input[X, N], Func.Input[Y, N], Func.Input[Z, N]) =
-      Unzip[List].unzip3[X, Y, Z](ev.subst(self.unsized)) match {
-        case (x, y, z) => (Sized.wrap[List[X], N](x), Sized.wrap[List[Y], N](y), Sized.wrap[List[Z], N](z))
-      }
-  }
```
sellout: And this deleted stuff got moved to foundation/…/quasar/contrib/shapeless

paulp:
Near as I can tell when 150+ files are hit, this looks good. I'm comfortable taking your word for it at this level and zeroing in a little harder at the leaves.

alissapajer:
👍

PR #1437, opened 2016-09-14
Move MarkLogic planner effect to typeclass param and constraints to instance defns

wemrysi:
Rearranges the marklogic planner typeclass, moving the planning effect to a type argument to the typeclass and the various constraints to the instance definitions where they are required.

This makes the capabilities required by the various planners much more obvious.

Hat tip to @sellout for the reminder.


drostron:
Merge script is requesting a merge from master.

PR #1439, opened 2016-09-15
Extract a `connector` subproject from `core`.

sellout:
Individual connectors should now depend on this rather than on core itself.

This is built on #1438.


connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -82,7 +82,7 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
 
     "convert a squashed read" in {
       // "select * from foo"
-      convert(listContents.some, identity.Squash(lpRead("/foo/bar"))) must
+      convert(listContents.some, identity.Squash(lpRead("/foo/bar")).embed) must
```
drostron: Are these embed additions related or just something you noticed while here?

sellout: They’re related – in `core` there is this terrible type alias `FLP` and an implicit conversion that `Fix`es an unfixed thing.Anyway, that is a bad thing to do and a failed experiment at simplifying fixed-point testing. So, I figured rather than dragging it into `connector` for a few tests that shouldn’t use it in the first place, I should just fix the tests.

 (Actually, I think that code _did_ get moved into the `connector` project for the time being, because Mongo tests depend on it, but it needs to die, and I added a comment about that.)

drostron: Gotcha, thanks.

drostron:
LGTM. Will merge upon successful build.

drostron:
`revision.md` disappeared, looks like the merge co-opted `11.4.57.md`.

sellout:
Ah crap. Thanks.

sellout:
Ok, all better. Skipped CI.

PR #1440, opened 2016-09-16
First pass implementation of `ThetaJoin`

jedesah:


marklogic/src/main/scala/quasar/physical/marklogic/qscript/ThetaJoinPlanner.scala
```diff
@@ -17,18 +17,30 @@
...
+      for {
+        l   <- freshVar[F]
+        r   <- freshVar[F]
+        lhs <- rebaseXQuery(lBranch, src)
+        rhs <- rebaseXQuery(rBranch, src)
+        body <- planMapFunc(combine){ case LeftSide => l.xqy case RightSide => r.xqy}
+      } yield {
+        for_(l -> lhs, r -> rhs).
+        return_(body)
+      }
```
sellout: Just to make sure it’s intentional … this is currently ignoring the condition and join type, so every join will be a full cross.

Does MarkLogic support full theta joins (i.e., arbitrary join conditions)? If so, awesome. If not, then you want to implement EquiJoin instead of this – but this impl could be copy/pasted over, because you’re not doing anything with the condition yet.

sellout:
I’m fine with this, just want answers to those questions before I merge.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ThetaJoinPlanner.scala
jedesah: Good catch. Should be fixed now (although it's not being tested).

PR #1441, opened 2016-09-16
Bump version of WartRemover

jedesah:


sellout:
Oooh, is this going to let us enable … uh, NoNeedForMonad finally?

jedesah:
Looks like we could enable it. I enabled it and recompiled and started going through some of the code it's flagging.

mossprescott:
Happy to merge but I'll hold off because it sounds like you might push some more changes.

jedesah:
It looks like `NoNeedForMonad` is still triggering a compiler crash, so I think we should probably hold off on enabling it.

In any case, many of the for-comprehensions it is flagging are arguably not as easy to read if re-written using applicative syntax. Hopefully one day we will have a better alternative syntax to for-comprehensions, and it will be easier to re-write the offending for-comprehensions without sacrificing on readability.

@mossprescott Feel free to merge! :smile:

PR #1442, opened 2016-09-16
Elide ThetaJoin with one side Nullary.

alissapajer:


alissapajer:
@sellout a small step toward fewer `ThetaJoin`s. This removes the outer `ThetaJoin` being generated in `select foo from bar`.

core/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -91,6 +91,12 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
           Free.point(RightSide)))).some)
     }
 
+    "convert a basic select with type checking" in {
+      val lp = fullCompileExp("select foo from bar")
+      val qs = convert(listContents.some, lp)
+      qs must equal(chain(RootR).some) // TODO incorrect expectation
```
alissapajer: I'll update this with the real expectation, happy to do that before merge, just wanted to push first.

core/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -157,6 +157,8 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
     (left.resume, right.resume) match { // if both sides are Constant, we hit the first case
       case (-\/(Constant(_)), _) => rebaseRight[F, G](rebase)(tj, left)
       case (_, -\/(Constant(_))) => rebaseLeft[F, G](rebase)(tj, right)
+      case (-\/(Undefined()), _) => rebaseRight[F, G](rebase)(tj, left)
+      case (_, -\/(Undefined())) => rebaseLeft[F, G](rebase)(tj, right)
       case (_, _) => TJ.inj(tj)
     }
```
sellout: So, I’m actually getting rid of these functions, as soon as I have a passing test I can PR. But basically, there is an infinite number of cases you’d have to match this way – e.g., `IfUndefined(Undefined(), Constant(???) + Now())`. The key to all of them is that `mf.size ≟ 0`. But … even checking for that is unnecessary. There is _another_ normalization I added that does `if (mf.size ≟ 0) src = Unreferenced()`, and then `elideConstantJoin` only has to check if the src of the Map is `Unreferenced`.

sellout: But, I can merge your work for now if you just change this to do the `mf.size ≟ 0` approach, to catch all cases like this.

alissapajer:
@sellout once travis passes this is g2g

alissapajer:
@sellout I merged in master

Issue #1405, opened 2016-09-02
Travis build stalls while compiling


jedesah:
It appears that sometimes the Travis build stalls during the compiling phase with these being the last few lines being printed out by the build:

```
[info] Done updating.
[info] Compiling 18 Scala sources to /home/travis/build/quasar-analytics/quasar/foundation/target/scala-2.11/classes...
[info] Compiling 9 Scala sources to /home/travis/build/quasar-analytics/quasar/effect/target/scala-2.11/classes...
[info] Compiling 5 Scala sources to /home/travis/build/quasar-analytics/quasar/foundation/target/scala-2.11/test-classes...
[info] Compiling 5 Scala sources to /home/travis/build/quasar-analytics/quasar/js/target/scala-2.11/classes...
[info] Compiling 3 Scala sources to /home/travis/build/quasar-analytics/quasar/ejson/target/scala-2.11/classes...
[info] Compiling 2 Scala sources to /home/travis/build/quasar-analytics/quasar/ejson/target/scala-2.11/test-classes...
[info] Compiling 91 Scala sources to /home/travis/build/quasar-analytics/quasar/core/target/scala-2.11/classes...
[info] Compiling 1 Scala source to /home/travis/build/quasar-analytics/quasar/js/target/scala-2.11/test-classes...

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

Since not all sub-projects finished compiling, it seems either Travis is just being way too slow or the Scala compiler is hanging.

If the former, we may want to consider using [travis_wait](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) in order to afford the build more time to compile.

mossprescott:
For what it's worth, I've seen this happen a lot where it turned out that the compile was finishing, and SBT was running tests, and the *test* was stalled. Since there's no output until at least one entire spec finishes, it looks like the compile is what's stuck. Of course I can't say if that's what's happening here.

alissapajer:
@mossprescott @jedesah should we mark this as closed, now that #1459 has landed?

mossprescott:
I'd say so, unless anyone's seen it happen since merging.

alissapajer:
Right, we can wait until the batch of like 10 open PRs all make it in, up to say the most recent #1460, and then if those make it in without seeing a hang, we'll call it fixed.

jedesah:
[...]

PR #1443, opened 2016-09-19
Wip/tightening q script types

alissapajer:
This PR is a clone of #1427. Some builds on that PR are hanging after 10 minutes, so this PR extends the build timeout to 20 minutes to see if we still see the hangs.


alissapajer:
Ok, this _might_ be the solution. The postgres build passed and the 2.6 mongo build failed with OOM, which is at least not a hang... I'm running both those builds again checking for build consistency.

alissapajer:
@sellout ok, so with a 240 min timeout, we consistently run out of memory in the 2.6 mongo build.

https://travis-ci.org/quasar-analytics/quasar/jobs/161169649

alissapajer:
If this works it would address #1405.

alissapajer:
OOM trace: https://gist.github.com/alissapajer/cc29e5c424b4b1452b1b0bc24bcc59d6

sellout:
So, that OOM is interesting – it’s happening in scaladoc generation. Scaladoc has been a pain in many ways. This is just one more. I wonder if we can generate less (e.g., only for the connector, frontend, and interface projects) or what. It’d be nice to move it to a separate build, but 1. it needs to be in the same place as the rest of the jar and 2. we don’t want to publish the jar if we don’t have at least one connector passing integration tests.

alissapajer:
closing - hopefully solved with #1459

PR #1444, opened 2016-09-19
Spark - Proper read

rabbitonweb:
Implement proper read for sparkcore where quasar will not read all content with single read, but do it in batches

### DIFF
--- a/CHANGELOG/feature.md
+++ b/CHANGELOG/feature.md
@@ -1,1 +1,1 @@
-- implement proper read for sparkcore where quasar will not read all content with single read, but do it in batches
+- Implement streaming reads for Spark
--- a/CHANGELOG/feature.md
+++ b/CHANGELOG/feature.md
@@ -1,0 +1,1 @@
+- implement proper read for sparkcore where quasar will not read all content with single read, but do it in batches
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
@@ -28,9 +28,8 @@ import quasar.fs._, FileSystemError._, PathError._
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd._
 import scalaz._, Scalaz._
-import scalaz.concurrent.Task
 
-final case class SparkCursor(rdd: Option[RDD[Data]], pointer: Int)
+final case class SparkCursor(rdd: Option[RDD[(Data, Long)]], pointer: Int)
 
 object readfile {
 
@@ -47,16 +46,14 @@ object readfile {
   def chrooted[S[_]](input: Input[S], prefix: ADir)(implicit
     s0: KeyValueStore[ReadHandle, SparkCursor, ?] :<: S,
     s1: Read[SparkContext, ?] :<: S,
-    s2: MonotonicSeq :<: S,
-    s3: Task :<: S
+    s2: MonotonicSeq :<: S
   ): ReadFile ~> Free[S, ?] =
     flatMapSNT(interpret(input)) compose chroot.readFile[ReadFile](prefix)
 
   def interpret[S[_]](input: Input[S])(implicit
     s0: KeyValueStore[ReadHandle, SparkCursor, ?] :<: S,
     s1: Read[SparkContext, ?] :<: S,
-    s2: MonotonicSeq :<: S,
-    s3: Task :<: S
+    s2: MonotonicSeq :<: S
   ): ReadFile ~> Free[S, ?] =
     new (ReadFile ~> Free[S, ?]) {
       def apply[A](rf: ReadFile[A]) = rf match {
@@ -79,7 +76,7 @@ object readfile {
       rdd <- input.rddFrom(f, offset, limit)
       cur = SparkCursor(rdd.map{ raw =>
         DataCodec.parse(raw)(DataCodec.Precise).fold(error => Data.NA, ι)
-      }.some, 0)
+      }.zipWithIndex.some, 0)
       h <- freshHandle
       _ <- kvs.put(h, cur)
     } yield h
@@ -91,28 +88,25 @@ object readfile {
   }
 
   private def read[S[_]](h: ReadHandle)(implicit
-    kvs: KeyValueStore.Ops[ReadHandle, SparkCursor, S],
-    s1: Task :<: S
+    kvs: KeyValueStore.Ops[ReadHandle, SparkCursor, S]
   ): Free[S, FileSystemError \/ Vector[Data]] = {
 
     // TODO arbitrary value, more or less a good starting point
     // but we should consider some measuring
-    val step = 10000
+    val step = 5000
 
     kvs.get(h).toRight(unknownReadHandle(h)).flatMap {
         case SparkCursor(None, _) =>
           Vector.empty[Data].pure[EitherT[Free[S, ?], FileSystemError, ?]]
         case SparkCursor(Some(rdd), p) =>
 
-        import com.redhat.et.silex.rdd.drop.implicits._
-
-        val collect: Free[S, FileSystemError \/ Vector[Data]] = lift(Task.delay {
-          rdd.drop(p).take(step).toVector.right[FileSystemError]
-        }).into[S]
+        val collect = rdd
+          .filter(d => d._2 >= p && d._2 < (p + step))
+          .map(_._1).collect.toVector.pure[EitherT[Free[S, ?], FileSystemError, ?]]
 
         for {
-          collected <- EitherT(collect)
-          cur = if(collected.isEmpty) SparkCursor(None, 0) else SparkCursor(Some(rdd), p + step)
+          collected <- collect
+          cur = if(collected.isEmpty) SparkCursor(None, 0) else SparkCursor(some(rdd), p + step)
           _ <- kvs.put(h, cur).liftM[FileSystemErrT]
         } yield collected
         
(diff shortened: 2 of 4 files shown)


rabbitonweb:
@jedesah please note also that this PR downgrades version of jackson to `2.4.4` as discussed previously on slack backend channel

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
```diff
@@ -91,15 +91,25 @@ object readfile {
...
-
-          collect <* clear
+        case SparkCursor(Some(rdd), p) =>
+
+        val collect = rdd
+          .filter(d => d._2 >= p && d._2 < (p + step))
+          .map(_._1).collect.toVector.pure[EitherT[Free[S, ?], FileSystemError, ?]]
+
+        for {
+          collected <- collect
```
jedesah: There is not much point in lifting the `Vector[Data]` into `EitherT[Free[S, ?], FileSystemError, ?]]` if we are going to immediately map over it and never use `collect` again. Might as well just leave it outside the monad.

That being said, `collect` is an effectful computation, so it does make sense for it to be inside some kind of delayed execution monad. The way it's happening right now, the effect is happening in supposedly pure code and then being lifted into a Monad after the fact.

All this to say that I suggest that the call to `collect` be wrapped in `Task` and that we lift the Task monad into the `Free[S,?]` Monad where `Task :<: S`. That way, this code will be pure.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
```diff
@@ -91,15 +91,25 @@ object readfile {
...
-        case SparkCursor(Some(rdd)) =>
-
-          val collect = rdd.collect.toVector.pure[EitherT[Free[S, ?], FileSystemError, ?]]
-          val clear = kvs.put(h, SparkCursor(None)).liftM[FileSystemErrT]
-
-          collect <* clear
+        case SparkCursor(Some(rdd), p) =>
+
+        val collect = rdd
+          .filter(d => d._2 >= p && d._2 < (p + step))
```
jedesah: This seems horrible for performance (not to mention somewhat ugly on the eyes). I think it would be considerably better to make use of `drop` and `take`. Unfortunately, it doesn't look like `RDD` has a `drop` method which strikes me as very odd, perhaps it's got a different name.

An alternative which I highly recommend would be to do something like this:
`readFromProcess( (file, opts) => io.iterator(input.rddFrom(file, opts.offset, opts.limit).toLocalIterator))`. This doesn't quite compile, there is a little more type massaging to be done much like [over here](https://github.com/quasar-analytics/quasar/blob/master/marklogic/src/main/scala/quasar/physical/marklogic/fs/readfile.scala#L40). But essentially this one line replaces almost this entire file.

Now I did notice that the Spark doc mentions this about `toLocalIterator`
> The iterator will consume as much memory as the largest partition in this RDD

I don't know whether this is acceptable or not, I sure hope it is or that there is an alternative way of getting some sort of iterator because it would be somewhat ridiculous for the Spark API to force us to do this performance killing filtering in order to accomplish streaming reads.

I also noticed that `rddFrom` uses the same filtering pattern which means we filter the entire dataset twice for every single streaming read! 😄 If `toLocalIterator` is acceptable, then we can just call `take` and `drop` on that iterator to avoid having to do that first filter.

CHANGELOG/feature.md
```diff
@@ -0,0 +1 @@
+- implement proper read for sparkcore where quasar will not read all content with single read, but do it in batches
```
jedesah: I would say "Implement streaming reads for Spark" just to keep it short and to the point, but I won't be offended if you would rather keep it the way it is.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
rabbitonweb: `zip` is somewhat horrible in performance however we must remember that those operations (like `.zip.filter.map`) are "pipelined" and run in parallel (both "zipping" and filtering are happening in the single run on each of the Spark nodes).

We can NOT use `take` as this operation is so called "action" which means it triggers the computation immediately, drop is not implemented for reasons described here http://erikerlandson.github.io/blog/2014/07/27/some-implications-of-supporting-the-scala-drop-method-for-spark-rdds/

I agree that `zip` + `filter` is possibly not the fastest solution [maybe we need to think later about something that was described in the link above) HOWEVER given the time constraints I'm aiming first for correctness then performance

`toLocalIterator` is even worse, please see the code below (from Spark codebase):

```
  def toLocalIterator: Iterator[T] = withScope {
    def collectPartition(p: Int): Array[T] = {
      sc.runJob(this, (iter: Iterator[T]) => iter.toArray, Seq(p)).head
    }
    (0 until partitions.length).iterator.flatMap(i => collectPartition(i))
  }
```

This means that the further you read, the worse your reads gonna be, because in order to perform `.next` on the iterator, further partitions will have to calculated.

`zip` & `filter` aren't that much big of a deal if we consider `caching` the source rdd - which I plan to do in the future HOWEVER as stated before, I target correctness at the moment

Maybe @jdegoes can drop few lines, but at the moment I think this is the right way to go.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
rabbitonweb: Obviously! Thank you for catching this one!

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
jedesah: If `drop` is implemented in a lazy fashion, I am not sure I see where the problem lies. That entire blog post is predicated on the idea that `drop` is strict. I don't understand why they had this idea in their mind that `drop` needed to be strict. Seems to me that if you are implementing a lazy data structure, there is no particularly good reason for `drop` to be strict. Frankly, I am not sure why ANY method needs to be strict except one which is "okay now do the thing". For instance, you mentioned `take` is strict. Why is that? Not blaming you, just ranting/trying to understand this API.

Anyway, the blog post points towards [this PR](https://github.com/apache/spark/pull/1254/) which points to [this one](https://github.com/apache/spark/pull/1839) which points to [this contrib library that has a lazy `drop`](http://silex.freevariable.com/latest/api/#com.redhat.et.silex.rdd.drop.DropRDDFunctions).

I think we should use that lib to make this code cleaner and more performant. Maybe they even have a lazy `take`. It doesn't really matter for `ReadFile` since we just call `take` instead of `collect`, but it's kind of a big deal for `QScript` if I am not missing something.

Also it's worth mentioning that it's not obvious to me in the code snippet you pasted above that `toLocalIterator` is doing something crazy that is undesirable. I mean, I don't know what `runJob` does, so that may very well be why.

rabbitonweb: So the library is basically implementation of the blog post, right?

There are two main concerns  on my side:
1. lack of time for experiments. 
It is a third-party (additional dependency) that I don't quite get. zip+filter is nasty, but it works - 7 days to deadline, this is my issue here
2. lack of lazyness
This is even more important. Each call will trigger the calculation on Spark. While in `ReadFile` algebra this is fine since we need that data anyway - this could be good enough. But in `QueryFile` algebra for example we have `QScriptCore.Take/Drop` and implementing those has to be lazy (folding [cata] over `QScript` structure has to build the final computation unit (the RDD) without running it. In that place the way `drop` is implemented in the links provided will not be sufficient as it triggers calculations for some reason.


As it goes to `toLocalIterator` I don't understand where the benefit would be. The `runJob` is a method from Spark API. Calling it triggers a calculation on a Spark cluster. The code cop-pasted from Spark shows us that in order to create that final iterator the spark will take partiton by partition and run .toIterator on each of the partition -triggering the calculations on spark. To give it an example, imagine we have collection of 10 elements [1,2,3...] divided into 2 partitions: first has elements [1,2,3] second has [4...10]. When you want to drop first 6 elements, using `toLocalIterator` will first fetch first partition, you will discard it's values, then clculate second partition and give u back results. It will trigger two separate jobs.

With `zip` + `filter` it will run single job in parallel. It will be "zipping" the values while applying filter in the same time. It will be a single job run on cluster, not multiple.

jedesah: So the first comment on [this PR](https://github.com/apache/spark/pull/1839) mentions that `drop` is implemented in a lazy fashion.

I think I understand better what you are saying about `toLocalIterator`, it's kind of dump and will load data into local memory before performing the `drop` operation. Agreed that is a no go.

jedesah:
Wow! I never seen that error before. wierd...

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/readfile.scala
```diff
@@ -91,28 +88,25 @@ object readfile {
...
         case SparkCursor(None, _) =>
           Vector.empty[Data].pure[EitherT[Free[S, ?], FileSystemError, ?]]
         case SparkCursor(Some(rdd), p) =>
 
-        import com.redhat.et.silex.rdd.drop.implicits._
-
-        val collect: Free[S, FileSystemError \/ Vector[Data]] = lift(Task.delay {
-          rdd.drop(p).take(step).toVector.right[FileSystemError]
-        }).into[S]
+        val collect = rdd
```
jedesah: This also reverted the change where `collect` was being properly wrapped in a `Task`.

rabbitonweb: crap :)

PR #1447, opened 2016-09-20
Fix bug in autojoin3.

alissapajer:
This bug would affect any query that autojoins 3 sources, for example, any query with a `Guard` node.

This PR now produces correct (though not yet fully optimized) qscript for the query `select foo from bar`.


drostron:
@alissapajer Thank you, the update is appreciated.

sellout:
This looks good – just waiting on Travis.

alissapajer:
@sellout looks like travis just gave it green

PR #1449, opened 2016-09-20
Add `UnshiftArray` support to old MongoDB connector.

sellout:


sellout:
@natefaubion This one’s for you.

it/src/main/resources/tests/unshiftAggregation.test
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "unshift aggregation",
+    "backends": {
+        "marklogic":  "pending",
+        "postgresql": "pending"
+    },
+    "data": "largeZips.data",
+    "query": "select state, city, [pop ...] as pop from largeZips group by state, city",
```
sellout: @natefaubion Here’s the example query, with results below.

it/src/main/resources/tests/unshiftAggregation.test
```diff
@@ -0,0 +1,15 @@
...
+    "predicate": "containsAtLeast",
+    "expected": [
+        { "state": "AK", "city": "ANCHORAGE",   "pop": [32383, 29857] },
+        { "state": "AK", "city": "KETCHIKAN",   "pop": [13886, 422] },
```
mossprescott: Presumably the order of values in the array is undefined, right? We'll get a away with that for now, but it's another case we'll have to del with eventually for new backends.

mossprescott:
:+1: I'll merge if/when it builds.

it/src/main/resources/tests/unshiftAggregation.test
sellout: I think it’s defined to be the order of the values in the set. So, since this query doesn’t have an `order by` before grouping, the order is unknown. We could add sorting to the test to make it deterministic.

natefaubion: Is it sensible to assume that, given multiple unshift aggregations, the arrays should at least "line up" such that you could transpose them?

sellout: @natefaubion You mean in `select … [pop ...], [loc ...] …`, the 0th pop should come from the same record as the 0th loc? Yes, I think that should be true.

This does bring up the one difficulty we’ve had when dealing with undefined values, though – when one record is missing a `loc` field, what do we do? If we skip the undefined values, then they no longer line up. If we _don’t_ skip them, then we need some representation of undefined (which could only ever appear in an array).

In this case, it might be more sensible to do something like `select … [{ "pop": pop, "loc": loc} ...] …`, which would keep them paired, and you might end up with some maps that only have a `"pop"` key or only have a `"loc"` key (if one or the other is undefined, which is not the case in our data set).

mossprescott: You can't just add an `order by` to the query, because sorting is applied _after_ the result is constructed. Probably you could do it with a sub-select, and maybe that's what we'd do to make this test deterministic. Anyway, I just wanted to confirm my understanding.

natefaubion: I was under the assumption that the record would have to contain both to be considered, but I guess that isn't necessarily the case? Does the suggested syntax for pairing work? Keeping them lined up is very important, so if that isn't the case then it might be a show stopper for me (I'll just have to resort to client side).

sellout: Yeah, I think your understanding of the current options are correct, _but_ with SQL² we should eventually be able to insert an ORDER BY before the GROUP BY, although I don’t think we’re there yet (and it does bring up complications with `ORDER BY`’s ability to see field aliases). And, if not that, we can write LPs that do what we want, once we have syntax for it.

sellout:
Arg, tests failed (locally) because I used `zips` for the results, but `largeZips` in the test. Whoops. Fix coming.

sellout:
Ok, fix is pushed – new build to wait for …

sellout:
Grr, failed because of Mongo’s broken field order. @mossprescott Should I fix it by just doing `select state as state0` etc.?

mossprescott:
I think you could have just renamed `pop` (because it happens to be at the end.) But either way is fine with me.

mossprescott:
Seems like this is now a field ordering problem that we might actually want to fix.

sellout:
*sigh* this was supposed to be a quick fix – any idea how the field order is still wrong?

mossprescott:
All I can see is that the result consistently has `pop0`, `state0`, `city0`, which makes me wonder if the pop half got left-right swapped with the state/city half. Do you have the branch handy? Maybe take a look at the plan and see if it 's actually wrong there. 
It's also possible the plan is correct and just suffering from MongoDB's field-ordering "bug".

sellout:
Yeah, it looks like the entire plan is correct, and it’s only when running it that it gets out of order – but I thought that only happened when the fields had names that matched the source collection …

mossprescott:
It doesn't have to be the actual source. It can be fields that were created in an earlier pipeline step.

sellout:
Hey wow – it passed.

PR #1452, opened 2016-09-20
Enable the Product and Serializable warts

sellout:


sellout:
Only the last commit is new here, the others are on #1427.

I’ve fixed things in the `js` project and `frontend`, but there are still two cases in `frontend` that I haven’t figured out:
```
[error] frontend/src/main/scala/quasar/sql/parser.scala:57: Inferred type containing Product
[error]       identifierString ^^ processIdent |
[error]                        ^
[error] frontend/src/main/scala/quasar/std/library.scala:109: Inferred type containing Serializable
[error]     unaryFunctions ++ binaryFunctions ++ ternaryFunctions
[error]                    ^
```

paulp:
I'll look at those two.

sellout:
And I think there were 79 cases I’ve seen so far (so, 77 fixed by the relatively small changes here).

paulp:
The first one is Token, which you are I guess inheriting. Make your own Token which extends that Token with Product with Serializable.

paulp:
The second one is shapeless, and I'm not sure how to handle it generally but you can just annotate the type of the first operand, like `(unaryFunctions: List[GenericFunc[_]]) ++ ...` and it will stop inferring anything more ambitious than that.

sellout:
Ok, the shapeless one is fixed. The `Token` one seems harder, as we are using values of type `Token` so if I try to switch to `SqlToken` (which extends Token with Product with Serializable), then some things aren’t of that subtype.

However, it just occurred to me that I can use `@SuppressWarnings`, and I’ll do that for anything else I can’t figure out until it’s done. Then we can look at those cases in a PR that compiles.

sellout:
Ok, this is updated and should pass. There is one `@SuprressWarnings` in here, but otherwise I fixed them all properly. And I just amended the appropriate commit (since the intermediate one wasn’t compiling anyway), so just look at the latest one.

And don’t merge until #1427 gets in.

sellout:
Ok, merged in master, so it should pass Travis now … and hopefully ready to merge.

paulp:
Merging master into pull requests interacts badly with the (increasingly ill-advised) system of having a file called "revision.md" with ephemeral contents.

PR #1454, opened 2016-09-21
Extract `ExpandDirs` transformation from `DiscoverPath`

sellout:
(Built on #1427)

Now `Read` conversion (via `DiscoverPath`) doesn’t ensure that all paths are files. There is an optional second step, `ExpandDirs` that converts any directories to unions of all the files contained therein.

This also parameterzes `Read` and `ShiftedRead` by the path type.


marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
```diff
@@ -36,23 +37,26 @@ package object qscript {
...
-      new ShiftedReadPlanner[F]
+    implicit def constShiftedReadFile[F[_]: NameGenerator: PrologW]: MarkLogicPlanner[F, Const[ShiftedRead[AFile], ?]] =
+      new UnreachablePlanner[F, Const[ShiftedRead[AFile], ?]]
 
     implicit def projectBucket[F[_]: Applicative, T[_[_]]]: MarkLogicPlanner[F, ProjectBucket[T, ?]] =
-      new ProjectBucketPlanner[F, T]
-
-    implicit def thetajoin[F[_]: NameGenerator: PrologW: MonadPlanErr, T[_[_]]: Recursive: ShowT]: MarkLogicPlanner[F, ThetaJoin[T, ?]] =
-      new ThetaJoinPlanner[F, T]
+      new UnreachablePlanner[F, ProjectBucket[T, ?]]
```
sellout: I wanted to point out that I changed this a bit – rather than having a planner for _every_ case, I consolidated the cases that MarkLogic should never see into a single `UnreachablePlanner`. So now there are three “real” planners, and the other components all use `UnreachablePlanner`.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ShiftedReadPathPlanner.scala
```diff
@@ -26,18 +27,15 @@ import matryoshka._
...
 import scalaz._
 
-private[qscript] final class ShiftedReadPlanner[F[_]: NameGenerator: PrologW]
-  extends MarkLogicPlanner[F, Const[ShiftedRead, ?]] {
-
-  val plan: AlgebraM[F, Const[ShiftedRead, ?], XQuery] = {
-    case Const(ShiftedRead(absFile, idStatus)) =>
-      val asDir = fileParent(absFile) </> dir(fileName(absFile).value)
-      val dirRepr = posixCodec.printPath(asDir)
+private[qscript] final class ShiftedReadPathPlanner[F[_]: NameGenerator: PrologW]
```
wemrysi: I wish it were this simple ;) For the time being, this will have to be `ShiftedRead[AFile]` like everything else as there are a bunch of changes on the filesystem side needed to make this possible.

wemrysi:
LGTM other than the MarkLogic bit, which needs to continue to expand dirs until we have time to change how the filesysytem implementation works.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ShiftedReadPathPlanner.scala
sellout: Ok, that’s reasonable. I do have more questions about the MarkLogic filesystem.

E.g., would your “ideal” representation have `ShiftedRead[ADir]` and `Read[AFile]` in the coproduct (i.e., files contain a single item, directories contain multiple) rather than simply `ShiftedRead[APath]`, where you have to treat files as singleton sets rather than individual values.

wemrysi: I _think_ `ShiftedRead[APath]` is actually preferred as a QScript `DataSet` seems to map well to an [XDM `Sequence`](https://www.w3.org/TR/xpath-datamodel/#sequences) and in the XDM "there is no distinction between an item and a singleton sequence containing that item."

sellout: Ok, cool. That works for me :)

sellout:
Ok, this is updated – hopefully it’ll pass and be ready for merge.

alissapajer:
@sellout what's the status with this PR?

sellout:
Just low-priority unless you need something from it. Turns out it's not usable by MarkLogic until other changes to the file system happen, so I wasn't going to worry about it until after the release.

sellout:
Ok, finally updated this … just a measly 256 commits that had to be merged in.

I wouldn’t be surprised if the merge resulted in some trivial diffs. I’ll try to clean those up.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/UnreachablePlanner.scala
```diff
@@ -16,17 +16,15 @@
...
 import matryoshka._
 import scalaz._, Scalaz._
 
-private[qscript] final class DeadEndPlanner[F[_]: Applicative]
-  extends MarkLogicPlanner[F, Const[DeadEnd, ?]] {
-
-  val plan: AlgebraM[F, Const[DeadEnd, ?], XQuery] = {
-    case Const(Root) => mkSeq_("/".xqy).point[F]
-  }
+private[qscript] final class UnreachablePlanner[M[_]: Applicative, F[_]]
```
wemrysi: How about adding a `case object Unreachable extends MarkLogicPlannerError` sentinel value and failing with that here in order to avoid false-negatives if we ever hit a bug in the planner?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -19,15 +19,13 @@ package quasar.physical.sparkcore.fs
...
 import quasar.fp.ski._
-import quasar.qscript._
-import quasar.contrib.pathy.AFile
-import quasar.qscript.ReduceFuncs._
-import quasar.qscript.SortDir._
+import quasar.qscript._, ReduceFuncs._, SortDir._
 
 import scala.math.{Ordering => SOrdering}
-import SOrdering.Implicits._
+import SOrdering.Implicits._ // FIXME: Where does this come from?
```
wemrysi: What is "this" referring to?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/UnreachablePlanner.scala
wemrysi: This ended up being implemented as part of #1850.

sellout:
@wemrysi Did you address your own comments in your PR against this PR, or should I go through them again and make further changes?

wemrysi:
@sellout The MarkLogic unreachable handling was dealt with in another PR and I'm ok leaving the `ExpandDirs` formulation as-is.

sellout:
Ugh. Erroring with 10-minutes-no-output.

wemrysi:
Yeah, in every build too :( Maybe we were just unlucky? Or have compile times increased too much? There is a big fan-out after `connector` builds, maybe that stage creeps over 10min on travis?

wemrysi:
I think the fan-out is followed by a corresponding fan-in to `interface`, so we have to wait for all the connectors to build before we see any output once they've started.

sellout:
Yeah, it looks like it’s generally waiting on `connector` to finish. Might just need to update `ExpandDirs` to the approach that @paulp has since introduced, which minimizes implicit resolution in the coproduct type classes.

wemrysi:
Hm, early build results don't look promising.

wemrysi:
@sellout Welp, it finally passed. Individual jobs look to be a bit slower on average (though, hard to tell yet given we only have one success and travis is quite volatile) which is to be expected given the added compilation time due to the changes, plus the additional startup time from separate `sbt` processes.

Doesn't look like the max build time increased that much, so hopefully this means elapsed wall-clock time won't be much worse than now.

sellout:
Ok, fingers crossed for developer compile time. I guess we can try inductive implicits now, too (should actually be released by the end of the week).

Is there anything else that needs to be addressed here? And who merges this branch at this point? 😆

wemrysi:
Aye, I am going to try testing this with scalatl 2.12.1 and see how it fares.

As for merging, I don't think there is anything else to address, though I wouldn't be surprised if this needs a merge from `master`.

I think either of us can merge, since I'm the original reviewer, I can do it if you like. If it needs `master` I'll take care of it and kick it back to you for a review.

wemrysi:
@sellout I went ahead and merged in master, PR submitted to your branch.

PR #1455, opened 2016-09-21
Merge sources through pair of LeftShift.

alissapajer:


alissapajer:
@sellout initially went down this path attempting to remove the final theta join from `select foo from bar`. In the end this isn't being hit because source merging happens before optimizations, and it's only the optimized for that has dual `LeftShift` as the left and right sides of the theta join. So that will need another solution.

But this is still useful.

alissapajer:
@sellout this is green!!

PR #1457, opened 2016-09-21
Spark Planner - Union, Take, Drop & LeftShift

rabbitonweb:
- [x] Union
- [x] Take
- [x] Drop
- [x] LeftShift


sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
+          // EitherT[Task, State[SC, A]]
+          // Task[PllE \/ State[SC, A]]
+
+
+          val fromState = freeCataM(from)(interpretM(κ(srcState), algebraM))
+          val countState = freeCataM(count)(interpretM(κ(srcState), algebraM))
+
+          val countEval = countState >>= (rdd => EitherT(Task.delay(rdd.first match {
+            case Data.Int(v) if v.isValidLong => v.toLong.right[PlannerError]
+            case Data.Int(v) => InternalError(s"Unsupported plan").left[Long]
```
alissapajer: I think it would be useful to make this error message more useful. Something like "provided Integer $a is not a Long".

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
-          ???
+          val algebraM = Planner[QScriptTotal[T, ?]].plan(fromFile)
+          val srcState = src.point[StateT[EitherT[Task, PlannerError, ?], SparkContext, ?]]
+
+          // StateT[Task[Planner \/ A]]
+          // EitherT[Task, State[SC, A]]
+          // Task[PllE \/ State[SC, A]]
+
+
+          val fromState = freeCataM(from)(interpretM(κ(srcState), algebraM))
```
alissapajer: Could you ascribe the types to this, `countState`, `countEval`, and the other `val`s you use in the other cases? Personally, I find the code a lot easier to read when all the types are ascribed, and I find it makes future refactorings easier too.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
             }
           )
         case Take(src, from, count) =>
-          ???
+          val algebraM = Planner[QScriptTotal[T, ?]].plan(fromFile)
+          val srcState = src.point[StateT[EitherT[Task, PlannerError, ?], SparkContext, ?]]
+
+          // StateT[Task[Planner \/ A]]
+          // EitherT[Task, State[SC, A]]
+          // Task[PllE \/ State[SC, A]]
```
alissapajer: What are these? Maybe just ascribe them to their appropriate expressions? Or leave a bit more of a comment explaining what they represent?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
+          val fromState = freeCataM(from)(interpretM(κ(srcState), algebraM))
+          val countState = freeCataM(count)(interpretM(κ(srcState), algebraM))
+
+          val countEval = countState >>= (rdd => EitherT(Task.delay(rdd.first match {
+            case Data.Int(v) if v.isValidLong => v.toLong.right[PlannerError]
+            case Data.Int(v) => InternalError(s"Unsupported plan").left[Long]
+            case a => InternalError(s"$a is not a Long number").left[Long]
+          })).liftM[StateT[?[_], SparkContext, ?]])
+          (fromState |@| countEval)((rdd, count) =>
+            rdd.zipWithIndex.filter(_._2 < count).map(_._1))
```
alissapajer: What's the type of `rdd`? Is there a `take` method on it, or is there a reason you went with the `zipWithIndex` approach? `take` would be more efficient since you wouldn't have to `zip`, `filter`, and then `map`.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
+          val srcState = src.point[StateT[EitherT[Task, PlannerError, ?], SparkContext, ?]]
+
+          val fromState = freeCataM(from)(interpretM(κ(srcState), algebraM))
+          val countState = freeCataM(count)(interpretM(κ(srcState), algebraM))
+
+          val countEval = countState >>= (rdd => EitherT(Task.delay(rdd.first match {
+            case Data.Int(v) if v.isValidLong => v.toLong.right[PlannerError]
+            case Data.Int(v) => InternalError(s"Unsupported plan").left[Long]
+            case a => InternalError(s"$a is not a Long number").left[Long]
+          })).liftM[StateT[?[_], SparkContext, ?]])
```
alissapajer: Since `Take` and `Drop` are so similar, pretty much everything in here could be factored out into a `def takeOrDrop` (there's probably a better name...).

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
+          (fromState |@| countEval)((rdd, count) =>
+            rdd.zipWithIndex.filter(_._2 >= count).map(_._1))
+
+        case LeftShift(src, struct, repair) =>
+
+          val structFunc =
+            freeCataM(struct)(interpretM(κ(ι[Data].right[PlannerError]), CoreMap.change))
+
+          def repairFunc: PlannerError \/ ((Data, Data) => Data) ={
```
alissapajer: For readability, I'd consider making this a `case class RepairData(leftSide: Data, rightSide: Data)`. Then it would be just a bit more clear that the input pair to the repair function is the source on the left and the source on the right.

This might be almost more of a style thing... so if you strongly disagree speak your mind :) But I find that case classes generally provide more clarity than tuples.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
+                case Data.Arr(elems) => elems(0)
+                case _ => Data.NA
+              }).right
+              case RightSide => ((x: Data) => x match {
+                case Data.Arr(elems) => elems(1)
+                case _ => Data.NA
+              }).right
+            }, CoreMap.change))
+
+            dd.map(df => (l, r) => df(Data.Arr(List(l, r))))
```
alissapajer: So, it looks like you're creating a `Data` here (the `Data.Arr`) when you don't really need to.

Here you've created a situation where you have to call `elems(0)` and `elems(1)` on a 2-element `List`, which is being constructed right here. So instead of forcing `l` and `r` into a `Data.Arr`, you could use the the `Data` value itself. I think then that you can avoid creating the `Data.Arr` here at all and just use the `Data` that you actually care about. Also `DataRepair` (see previous comment) might be useful too, to contain the `l` and `r`.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -111,13 +107,87 @@ object Planner {
...
+          }
+
+          StateT((sc: SparkContext) =>
+            EitherT((for {
+              df <- structFunc
+              rf <- repairFunc
+              ls = (input: Data) => df(input) match {
+                case Data.Arr(list) => list.map(rf(input, _))
+                case Data.Obj(m) => m.values.map(rf(input, _))
+                case _ => List.empty[Data]
```
alissapajer: @sellout so this implementation of `LeftShift` is consistent with [that in MRA](https://github.com/quasar-analytics/purescript-mra/blob/master/src/MRA.purs#L607), but (at least in mongo) we sometimes use left shifting to break a dataset out into into its individual records (a `LeftShift`ed `Read`). How backend-specific is that to mongo? Is that something we need to think about here too?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
alissapajer: Ah, I was just reading your conversation about `take` over in https://github.com/quasar-analytics/quasar/pull/1444 and it looks like I'll need to read that whole thing :)

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
sellout: So, this should be a correct implementation regardless. If a `Read` precedes the `LeftShift`, that Read will evaluate to a singleton dataset, and this will convert the one value in it to multiple records.

However, the Spark connector implements `ShiftedRead` (rather than `Read`), which means the result of reading is already a dataset of “records”, so there will usually not be a `LeftShift` immediately after reading.

alissapajer:
@rabbitonweb I also wanted to ask about the status of testing spark? Are there any unit tests or are we just using the integration tests? Have you been able to test that some of these cases are implemented correctly, or are there still basic parts missing making it untestable?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
rabbitonweb: that was my comments for later to reason about the code created (stacked monad transformers still blow my mind from time to time :) ) - this is now deleted

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
rabbitonweb: You don't need it. I can make quick intro. RDD operations are divided into two categories:
1. transformations
2. actions

Transforamtions are lazy operations. They don't trigger the computation, they just add next step to the "recipe". The return type of calling any transformation is yet another RDD

Actions are eager operations. They actually "materialize" RDD. They trigger calculations on the Spark cluster. The return type is anything but RDD

`Take` is implemented on RDD but it is an action, not transforamtion.
`Drop` is not implemented at all.

I found a blog about missing "drop", @jedesah followed that blog and found a small lib that implements `drop`  -however there were issues with that lib.

In Q4 I plan to implement my own `drop` and `take` that can be simple transformations. But for now with the time limit I aim for correctness. Thus `zipWithIndex`  +  `filter`

rabbitonweb:
@alissapajer Regarding tests: when I was working on `ReadFile`, `WriteFile` and `ManageFile` I've approached them with TDD fashion. All cases were covered. I've learned later that there were already more general tests written to test any connector that implemented the algebras above. After talking with @wemrysi,  I've decided to delete those tests simply because there was no point keeping doubles.

Same applies for `QueryFile`. @sellout said there are general tests for `QScript` as well. I have not run them yet (algebras are not complete). But I hope they will cover all that has been implemented so far.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
rabbitonweb: But how will I distinguish between the `LeftSide` and the `RightSide`?

I don't get it.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
sellout: This should actually not be `InternalError` – that implies that a connector got a plan that should not exist. It should be something like `UnsupportedPlan("... value too large to handle")`. We don't want to say Long, because our users don't know what Long is.

PR #1459, opened 2016-09-21
Limit concurrent tasks in Travis to 2

mossprescott:
SBT normally limits tasks overall to the reported number of processors/cores. Come to find out, that number is 32(!) when running in Travis.

Since we believe Travis only uses 1-2 CPUs worth of processor time, resetting the parallel task limit to 2 should preserve any actual parallelism while reducing memory footprint and thrashing.

Running on another branch, this change made the `doc` task complete reliably. It's less clear what it does to the build time.

### DIFF
--- a/connector/src/main/scala/quasar/fs/QueryFile.scala
+++ b/connector/src/main/scala/quasar/fs/QueryFile.scala
@@ -174,7 +174,7 @@ object QueryFile {
     val transforms = Transforms[F]
     import transforms._
 
-    /** Returns the path to the result of executing the given [[LogicalPlan]],
+    /** Returns the path to the result of executing the given `LogicalPlan`,
       * using the provided path if possible.
       *
       * Execution of certain plans may return a result file other than the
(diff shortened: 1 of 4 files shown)


connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -174,7 +174,7 @@ object QueryFile {
     val transforms = Transforms[F]
     import transforms._
 
-    /** Returns the path to the result of executing the given [[LogicalPlan]],
+    /** Returns the path to the result of executing the given `LogicalPlan`,
```
sellout: Hrmm, why these changes?

1. the original should just be changed to `[[quasar.LogicalPlan]]` and
2. I turned off `-Xfatal-warnings` in scaladoc specifically so we didn’t have to write less-useful scaladoc just because the scaladoc processor was unhappy.

mossprescott:
I tried just adding the package and there was still a (non-fatal) warning, so this change just cleans up the log slightly. In any case the same types appear in the signatures so when you actually look at the generated docs we're not really losing anything.

mossprescott:
The build is looking good and .1 finished in less than an hour which is consistent with what I was seeing on my fork, and would be a massive improvement over recent builds.

alissapajer:
@mossprescott awesome!!

PR #1460, opened 2016-09-22
Generate new operators in MongoDB 3.0

### LINKED ISSUE #1310: MongoDB 3.0: use new operators
Depends on #1308

There is only one new operator: [`$dateToString`](https://docs.mongodb.com/manual/reference/operator/aggregation/dateToString/#exp._S_dateToString).
- [x] Use it to implement `TimeOfDay` in expr instead of Js (avoiding mapreduce).
- [ ] Possibly add a new date formatting function (possibly following the [PostgreSQL syntax](https://www.postgresql.org/docs/9.2/static/functions-formatting.html)).

Note: selectors now support [`$eq`](https://docs.mongodb.com/manual/reference/operator/query/eq/#op._S_eq), but the benefit of that for generated worklfows is probably not worth the effort to track types in `Selector`.

mossprescott:
Closes #1310.

The upshot here is just that `time_of_day` can now be evaluated in the aggregation pipeline. More interestingly, this lays the groundwork for the 3.2 ops to be generated as well.

### DIFF
--- a/foundation/src/main/scala/quasar/fp/tree/package.scala
+++ b/foundation/src/main/scala/quasar/fp/tree/package.scala
@@ -21,33 +21,31 @@ import quasar.Predef._
 import matryoshka._
 import scalaz._
 
+/** "Generic" types for building partially-constructed trees in some
+  * "functorized" type. */
+// TODO: submit to matryoshka?
 package object tree {
-  /** A tree structure with one kind of hole. See `eval1`. */
+  /** A tree structure with one kind of hole. See [[UnaryOps.eval]]. */
   type Unary[F[_]] = Free[F, UnaryArg]
   object Unary {
     def arg[F[_]]: Unary[F] = Free.pure(UnaryArg._1)
   }
-
-  /** A tree structure with two kinds of holes. See `eval2`. */
-  type Binary[F[_]] = Free[F, BinaryArg]
-  object Binary {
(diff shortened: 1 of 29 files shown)


foundation/src/main/scala/quasar/fp/tree/package.scala
```diff
@@ -0,0 +1,88 @@
...
+import matryoshka._
+import scalaz._
+
+/** "Generic" types for building partially-constructed trees in some
+  * "functorized" type. */
+// TODO: submit to matryoshka?
+package object tree {
+  /** A tree structure with one kind of hole. See [[UnaryOps.eval]]. */
+  type Unary[F[_]] = Free[F, UnaryArg]
+  object Unary {
```
mossprescott: @sellout: The idea is to replace `JsFn` with `Unary[JsCore]`, for example. Interesting?

mossprescott:
Restarted the build ("server failed to start").

wemrysi:
LGTM, I'll merge when green.

jdegoes:
@mossprescott Remember to add at least one integration test for each new operator so we have some end-to-end coverage that benefits other connectors as well.

mossprescott:
@jdegoes: In this case, we're just using the operator to implement an existing function in the pipeline instead of in JS, so one test now runs on the read-only Mongo configuration.

The more interesting operators in 3.2 will come with new tests.

PR #1461, opened 2016-09-22
Lazily load regression test data

wemrysi:
Instead of eagerly loading all regression test data at suite start, this PR changes the behavior to load test data on-demand, if it is missing. This makes it nicer when running a subset of the regression suite as only the required data is loaded.

### DIFF
--- a/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
+++ b/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
@@ -89,7 +89,7 @@ abstract class QueryRegressionTest[S[_]](
   fileSystemShould { fs =>
     suiteName should {
       tests.toList foreach { case (f, t) =>
-        regressionExample(f, t, fs.name, fs.testInterpM)
+        regressionExample(f, t, fs.name, fs.setupInterpM, fs.testInterpM)
       }
 
       step(runT(fs.setupInterpM)(manage.delete(DataDir)).runVoid)
@@ -103,12 +103,13 @@ abstract class QueryRegressionTest[S[_]](
     loc: RFile,
     test: RegressionTest,
     backendName: BackendName,
+    setup: Run,
     run: Run
   ): Fragment = {
     def runTest: Result = {
       val data = testQuery(DataDir </> fileParent(loc), test.query, test.variables)
 
-      (ensureTestData(loc, test, run) *> verifyResults(test.expected, data, run))
+      (ensureTestData(loc, test, setup) *> verifyResults(test.expected, data, run))
         .timed(5.minutes)
         .unsafePerformSync
     }
--- a/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
+++ b/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
@@ -37,7 +37,7 @@ import org.specs2.specification.core.Fragment
 import pathy.Path, Path._
 import scalaz._, Scalaz._
 import scalaz.concurrent.Task
-import scalaz.stream.{merge => pmerge, _}
+import scalaz.stream._
 
 abstract class QueryRegressionTest[S[_]](
   fileSystems: Task[IList[FileSystemUT[S]]])(
@@ -88,8 +88,6 @@ abstract class QueryRegressionTest[S[_]](
 
   fileSystemShould { fs =>
     suiteName should {
-      step(prepareTestData(tests, fs.setupInterpM).unsafePerformSync)
-
       tests.toList foreach { case (f, t) =>
         regressionExample(f, t, fs.name, fs.testInterpM)
       }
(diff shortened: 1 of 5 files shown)


wemrysi:
Ah, looks like this is failing due to the `letInSelectProjection.test` not correctly declaring its data dependencies, will fix.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -132,11 +129,19 @@ abstract class QueryRegressionTest[S[_]](
...
-    query.fileExists(file).map(exists =>
-      if (exists) org.specs2.execute.Success(s"data file exists: ${fileName(file).value}")
-      else org.specs2.execute.Failure(s"data file does not exist: ${fileName(file).value}"))
+  /** Ensures the data for the given test exists in the Quasar filesystem, inserting
+    * it if not. Returns whether any data was inserted.
+    */
+  def ensureTestData(testLoc: RFile, test: RegressionTest, run: Run): Task[Boolean] = {
+    def ensureTestFile(loc: RFile): Task[Boolean] =
+      run(query.fileExists(dataFile(loc))).ifM(
+        false.point[Task],
```
alissapajer: @wemrysi in this case, it would be nice to maintain some sort of helpful error message informing the user that the query failed because the data dir did not exist. Otherwise it could be some tricky debugging.

Here we're using the `Boolean` in `Task` to denote success or failure, but `Task` itself has a built-in failure. So one thought would be that `ensureTestData` returns `Task[Unit]` and fails when something went wrong loading the data. We would then have to turn that `Task` failure into a specs2 `Failure` with the error message from `Task` reused. 

Basically, my primary concern is that we provided descriptive error messages when something goes wrong.

wemrysi: By "data dir did not exist" do you mean on the host filesystem or in quasar? If the latter, then this should not happen as, if it doesn't exist in quasar, it is inserted. If the source data file on the host filesystem does not exist, then the resulting `Task` will be a failure containing the exception thrown when trying to read the non-existent file.

The latter scenario results in something similar to the following in the test output:
```
[error]     ! simple read [./trivial.test]
[error]      java.io.FileNotFoundException: ./it/src/main/resources/tests/no_such.data (QueryRegressionTest.scala:222)
[error] quasar.regression.QueryRegressionTest.quasar$regression$QueryRegressionTest$$$anonfun$30(QueryRegressionTest.scala:222)
```

`Boolean` is actually just used to denote whether any data was written to the quasar filesytem or not (as it may have already existed), it doesn't represent an error.

And while we did return some message before, the purpose of the two functions is different, the previous one checked a condition and reported on it, this one attempts to bring the quasar filesystem into the desired state based its current state.

Unfortunately, we never actually used the nice error messages we were generating as you can see on the previous line `111` where we ignore the result of `verifyDataExists` :(

wemrysi: I'll customize the file loading error message to make it clearer that the IO exception was a result of trying to load one of the data files.

alissapajer: I was refering to the data not existing on the host filesystem. In that case it looks like the file not found error is already being given to the user, as I wanted to ensure.

Also, +1 to any customized error messages.

Oh, and a follow up question: Do we care about the value of the `Boolean`? That is, why does it matter if the data already existed in the quasar filesystem? It looks like we're just throwing away the value when we do `ensureTestData(loc, test, setup) *>`.

wemrysi: No, we don't currently use the value returned by `ensureTestData`, it just seemed like a reasonable return value for that function.

alissapajer:
@wemrysi LGTM - I'll merge once jenkins is green

alissapajer:
@wemrysi so the build for 3.0_RO failed with two instances of
```
[error]      java.lang.RuntimeException: File not found: /regression/cars
```
I actually don't know the difference between the 3.0 build and the 3.0_RO build - what is the difference?

Is there something in the RO build that would cause a file not to be found that _was_ successfully found in the non-RO build?

alissapajer:
Maybe something about the file permissions in read-only (RO) mode caused the file not to be located?

wemrysi:
@alissapajer Yeah, see my earlier comment. The test in question doesn't declare that it depends on the `cars` data set, that's the root of the problem. As for why it happens in RO and not in the other cases, my (unconfirmed, but fairly confident) hypothesis is that another test _does_ depend on `cars` and loads the dataset so that when this test is reached in the non-RO cases the data is there. However, in RO-mode we skip lots of tests that we know can't run with read-only permissions, I suspect the test that was loading `cars` is one of the ones we skip so when we reach the test in question, it attempts to query `cars` and it doesn't exist.

wemrysi:
Come to think of it, we should probably be extracting data set names from the compiled queries (`LogicalPlan`), then we don't need to declare them separately. That's for another PR, however.

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -186,24 +191,24 @@ abstract class QueryRegressionTest[S[_]](
...
-      }
+    val load = write.createChunked(
+      dataFile(dataLoc),
+      testData(dataLoc).chunk(500).translate(injectTask)
+    ).translate[Task](throwFsError.compose[FsErr](runT(run)))
 
-    throwFsError(EitherT(pmerge.mergeN(loads).take(1).runLast.map(_ <\/ (()))))
+    throwFsError(EitherT(load.take(1).runLast map (_ <\/ (())))) handleWith {
+      case err =>
+        val msg = s"Failed loading test data '${posixCodec.printPath(dataLoc)}': ${err.getMessage}"
```
alissapajer: 👍

PR #1462, opened 2016-09-23
Spark - Evaluate plan

rabbitonweb:
Implements EvaluatePlan, More & Close.

EvaluatePlan is implemented in similar fashion as Open in ReadFile. It uses zip+filter. This is probably not the most elegant way to do it (and probably not the most efficient one) - but it works and it uses vanilla spark api. Anything else will require 3rd parties or hacking our own implementations of take & drop. More discussion here https://github.com/quasar-analytics/quasar/pull/1444

TL;DR This is implemented with zipWithIndex + filter. Works and for now I aim for correctness. In Q4 I might try to work on something which will be slightly more efficient.


rabbitonweb:
Wait for `further-execute-plan` to be merged to master before being ready for review

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -14,11 +14,31 @@
  * limitations under the License.
  */
 
+/*
+ * Copyright 2014–2016 SlamData Inc.
```
alissapajer: It looks like we now have two license headers.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -65,6 +89,15 @@ object queryfile {
 
           qs.run.run
         }
+        case QueryFile.EvaluatePlan(lp: Fix[LogicalPlan]) => {
```
alissapajer: This case looks identical to `ExecutePlan`, modulo the call to `evaluatePlan`. Can you factor out the common parts into a separate funciton?

alissapajer:
@rabbitonweb didn't get a chance to finish reviewing this weekend - will finish on Monday.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -96,6 +129,58 @@ object queryfile {
...
+      case RddState(None, _) =>
+        Vector.empty[Data].pure[EitherT[Free[S, ?], FileSystemError, ?]]
+      case RddState(Some(rdd), p) =>
+        for {
+          collected <- lift(Task.delay {
+            rdd
+              .filter(d => d._2 >= p && d._2 < (p + step))
+              .map(_._1).collect.toVector
+          }).into[S].liftM[FileSystemErrT]
+          rddState = if(collected.isEmpty) RddState(None, 0) else RddState(Some(rdd), p + step)
```
alissapajer: `rddState` looks unused.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -96,6 +129,58 @@ object queryfile {
...
+      .map(_.leftMap(planningFailed(lp, _)))
+      .map(mrh => EitherT(mrh.point[Writer[PhaseResults, ?]]))
+  }
+
+  private def more[S[_]](h: ResultHandle)(implicit
+      s0: Task :<: S,
+      kvs: KeyValueStore.Ops[ResultHandle, RddState, S]
+  ): Free[S, FileSystemError \/ Vector[Data]] = {
+
+    val step = 5000
```
alissapajer: Why did you choose this value? Can you leave a comment saying why 5000 is a good number? Should this be passed as a function parameter instead so it's easier to tweak?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
rabbitonweb: Not really. I mean it is doable but it would require a bit of work as those two are a bit different. 
I've started that refactor with @wemrysi during PP and the changes are stashed - but given the time limits, amount of work yet to be done & fact that it is not an easy "extraction" of common parts - I would appreciate if we let this one go with a TODO for Q4

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
rabbitonweb: wow, nice one, thx!

rabbitonweb: Same as ReadFile (were I actually left a comment + a TODO) that this 5000 is arbitrary. There was no reasoning done, nor measurement. Will think about it in Q4 what the value should be (probably there will be some default + ability to configure it).

rabbitonweb: Nevertheless I will provide a function as a starting point - because this number will probably be "connector" specific (different for hdfs, different for cassandra etc)

rabbitonweb:
Ready

PR #1463, opened 2016-09-23
Fix decoding of paths containing "+"

mossprescott:
Fixed problems found by improving Pathy's generator to emit all ASCII characters with reasonable frequency; I'll PR that fix shortly but I don't think these fixes need to wait for a Pathy release; there are hand-written tests for the particular problem cases here.

This also cleared up several `flakyTest`s.


mossprescott:
Having some trouble getting this to build. Timeouts in random builds, possibly when compiling `connector`. I'll keep clicking "restart build" for now because I don't think it's anything this PR is making worse.

mossprescott:
About three more restarts and got lucky. 🎉

web/src/test/scala/quasar/api/UriPathCodecSpec.scala
```diff
@@ -78,5 +80,34 @@ class UriPathCodecSpec extends quasar.Qspec {
         codec.parseRelDir(codec.printPath(path)) must_== Some(path)
       }
     }
+
+    "print and parse through http4s should produce same Path" >> {
+      "absolute file with plus" >> {
+        import pathy.Path._
+        val path = rootDir </> file("a+b/c")
+        val hpath = HPath(codec.printPath(path))
+        AsFilePath.unapply(hpath) must_== Some(path)
```
drostron: `must_=` is available and I believe we're switching to it where possible.

drostron:
Nice to see the removal of all those `!hasDot`s

drostron:
LGTM. Will merge upon conflict resolution.

web/src/test/scala/quasar/api/UriPathCodecSpec.scala
mossprescott: Yes. I was following the pattern from above but I'd be happy to switch the file over.

mossprescott:
Should be ready to go.

drostron:
👍🏽

PR #1466, opened 2016-09-23
Couchbase fs interpreters

drostron:
This is a WIP. There are a few noted and unnoted ToDos that should be revisited once the initial pass of the full Couchbase implementation is in place.

Couchbase isn't currently enabled in the Travis build matrix due to eventual consistency woes causing intermittent integration test failures. I will work with our Couchbase contact on this.

Addresses: #1320


connector/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -75,17 +75,12 @@ object InMemory {
...
-            for {
-              i <- nextSeq
-              h =  ReadHandle(f, i)
-              _ <- readingL(h) := Reading(f, off, lim, 0).some
-            } yield h.right
-
-          case None =>
-            fsPathNotFound(f)
-        }
+        fileL(f).st *>
```
mossprescott: No need to even run `fileL` now, since it was only being used to check for the presence of the file, and now you're discarding that information.

couchbase/src/main/scala/quasar/physical/couchbase/common.scala
```diff
@@ -0,0 +1,89 @@
...
+      .toIList.unite.uncons(
+        FileSystemError.pathErr(PathError.invalidPath(f, "no bucket specified")).left,
+        (h, t) => BucketCollection(h, t.intercalate("/")).right)
+
+  def docIdTypesWithTypePrefix[S[_]](
+    bucket: Bucket,
+    prefix: String
+  ): Task[List[DocIdType]] = Task.delay {
+    val qStr = s"""SELECT meta(`${bucket.name}`).id, type FROM `${bucket.name}`
+                   WHERE type LIKE "${prefix}%"""";
```
mossprescott: Seems like we're vulnerable to SQL injection here with both of these parameters. Can we do something simple about that, like fail if the bucket name contains [[backtick]], or if `prefix` contains`"`?

mossprescott: Relatedly, we could encounter file names with `%` in them, so that should be escaped (as `\%`, probably).

couchbase/src/main/scala/quasar/physical/couchbase/fs/managefile.scala
```diff
@@ -0,0 +1,119 @@
...
+      src       <- EitherT(bucketCollectionFromPath(scenario.src).point[Free[S, ?]])
+      dst       <- EitherT(bucketCollectionFromPath(scenario.dst).point[Free[S, ?]])
+      _         <- EitherT((
+                     if (src.bucket =/= dst.bucket) FileSystemError.pathErr(
+                       PathError.invalidPath(scenario.dst, "different bucket from src path")).left
+                     else
+                       ().right
+                   ).point[Free[S, ?]])
+      bkt       <- EitherT(getBucket(src.bucket))
+      srcDocs   <- lift(docIdTypesWithTypePrefix(bkt, src.collection)).into.liftM[FileSystemErrT]
```
mossprescott: In this version, you're _only_ checking that some data exists, so it seems dangerous to load them all into memory just for that check. How about a variant that runs a different query just to see if any data exists?

Alternatively, can you add "distinct" to that query so it's one result per type/file, rather than one per doc?

couchbase/src/main/scala/quasar/physical/couchbase/fs/managefile.scala
```diff
@@ -0,0 +1,119 @@
...
+      _         <- EitherT((
+                     if (srcDocs.isEmpty)
+                       FileSystemError.pathErr(PathError.pathNotFound(scenario.src)).left
+                     else
+                       ().right
+                   ).point[Free[S, ?]])
+      dstDocs   <- lift(docIdTypesWithTypePrefix(bkt, dst.collection)).into.liftM[FileSystemErrT]
```
mossprescott: Same issue.

couchbase/src/main/scala/quasar/physical/couchbase/fs/managefile.scala
```diff
@@ -0,0 +1,119 @@
...
+    path: APath
+  )(implicit
+    S0: Task :<: S,
+    context: Read.Ops[Context, S]
+  ): Free[S, FileSystemError \/ Unit] =
+    (for {
+      ctx    <- context.ask.liftM[FileSystemErrT]
+      bktCol <- EitherT(bucketCollectionFromPath(path).point[Free[S, ?]])
+      bkt    <- EitherT(getBucket(bktCol.bucket))
+      docIds <- lift(docIdTypesWithTypePrefix(bkt, bktCol.collection)).into.liftM[FileSystemErrT]
```
mossprescott: Same concern, but in this case you're actually doing something with the docs, so I'd say if this is the easiest way to go, it's ok for now.

couchbase/src/main/scala/quasar/physical/couchbase/fs/queryfile.scala
```diff
@@ -0,0 +1,112 @@
...
+    (for {
+      ctx      <- context.ask.liftM[FileSystemErrT]
+      bktNames <- lift(Task.delay(
+                    ctx.manager.getBuckets.asScala.toList.map(_.name)
+                  )).into.liftM[FileSystemErrT]
+      bkts     <- bktNames.traverse(n => EitherT(getBucket(n)))
+      bktCols  <- bkts.traverseM(bkt => lift(Task.delay(
+                    bkt.query(n1qlQuery(s"select type from `${bkt.name}`"))
+                      .allRows.asScala.toList.map(r =>
+                        BucketCollection(bkt.name, new String(r.byteValue)))
```
mossprescott: This will almost certainly explode the VM on any "real" instance. `select distinct type ...` would probably be the best fix, if it's available.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
```diff
@@ -0,0 +1,98 @@
...
+                 )).into.liftM[FileSystemErrT]
+    } yield Cursor(bkt, qResult)).run
+
+  def read[S[_]](
+    cursor: Cursor
+  )(implicit
+    S0: Task :<: S
+  ): Free[S, FileSystemError \/ (Cursor, Vector[Data])] =
+    // TODO: Chunking?
+    if (cursor.result.hasNext) {
```
mossprescott: This is effectful and needs to happen in `Task.delay`. Looks like you can just put the whole thing inside `lift(...)`.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
```diff
@@ -0,0 +1,98 @@
...
+    }
+    else
+      (cursor, Vector.empty[Data]).right.point[Free[S, ?]]
+
+  def close[S[_]](
+    cursor: Cursor
+  )(implicit
+    S0: Task :<: S
+  ): Free[S, Unit] =
+    ().point[Free[S, ?]]
```
mossprescott: Is there really nothing to do to close these Iterators? Maybe a TODO if they should actually be cleaned up?

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
```diff
@@ -0,0 +1,98 @@
...
+import com.couchbase.client.java.document.json.JsonObject
+import eu.timepit.refined.api.RefType.ops._
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+
+object readfile {
+
+  implicit val codec = DataCodec.Precise
+
+  final case class Cursor(bucket: Bucket, result: Iterator[JsonObject])
```
mossprescott: Are we cool with putting a stateful object like this inside a case class which looks like a pure value? I'm actually not sure if we've been careful about that elsewhere. I guess that alternative is to hide it away as some kind of private member and expose only a `Task[Vector[JsonObject]]`. Not sure how that would work.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
```diff
@@ -0,0 +1,98 @@
...
+  )(implicit
+    S1: Task :<: S,
+    context: Read.Ops[Context, S]
+  ): Free[S, FileSystemError \/ Cursor] =
+    (for {
+      ctx     <- context.ask.liftM[FileSystemErrT]
+      bktCol  <- EitherT(bucketCollectionFromPath(file).point[Free[S, ?]])
+      bkt     <- EitherT(getBucket(bktCol.bucket))
+      limit   =  readOpts.limit.map(lim => s"LIMIT ${lim.unwrap}").orZero
+      qStr    =  s"""SELECT d.* FROM `${bktCol.bucket}` d
```
mossprescott: Same question about SQL injection.

effect/src/main/scala/quasar/effect/UUID.scala
```diff
@@ -0,0 +1,41 @@
...
+  final class Ops[S[_]](implicit S: UUID :<: S) extends LiftedOps[UUID, S] {
+    def randomUUID: F[JUUID] = lift(RandomUUID)
+  }
+
+  object Ops {
+    implicit def apply[S[_]](implicit S: UUID :<: S): Ops[S] = new Ops[S]
+  }
+
+  val toTask: UUID ~> Task = λ[UUID ~> Task] {
+    case RandomUUID => Task.delay(JUUID.randomUUID)
```
mossprescott: Seems like this uses `SecureRandom`, which could be problematic in Travis; it will block indefinitely if there's not enough entropy available. Or not, depending on how you call it.

We might want to look into other generators, or at least verify that this one is performing well.

it/src/test/scala/quasar/fs/ReadFilesSpec.scala
```diff
@@ -63,20 +63,17 @@ class ReadFilesSpec extends FileSystemTest[FileSystem](FileSystemTest.allFsUT) {
   def deleteForReading(run: Run): FsTask[Unit] =
     runT(run)(manage.delete(readsPrefix))
 
+  // TODO: Reasonable for these tests at least? Would need Order[Data] to use ISet.
+  //       Could just use must_===?
+  implicit val setDataEq = Equal.equalA[Set[Data]]
```
mossprescott: I see that `Data` is using `equalA`, so this should work for now. Wish there was a way to require that. I don't think writing `Order[Data]` is a good use of time at this moment.

mossprescott:
Mostly minor comments, except for the one piece of effectful code. I do think it's worth doing something about sanitizing paths (even if it's pretty bare bones), and I wonder if `randomUUID` was causing problems in Travis when the tests were running.

I'm concerned about the implications of eliminating errors for missing files, but I feel like we've been back and forth on that issue before and maybe you've already brought it up with others. If not, I think we should do that.

couchbase/src/main/scala/quasar/physical/couchbase/common.scala
drostron: Agreed. If possible I'd like to postpone addressing these until the full pass at initial QScript support is in place. There is already a ToDo for injection and I created an issue so it isn't forgotten, #1468. There is also now an issue for escaping filenames, #1469.

couchbase/src/main/scala/quasar/physical/couchbase/fs/managefile.scala
drostron: Adding a variant for data existence will be quick and useful.

couchbase/src/main/scala/quasar/physical/couchbase/fs/managefile.scala
drostron: It's only retrieving the document id, but the issue remains when the result set is sizable enough. Adding `distinct` to the query is a good suggestion.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
drostron: True! It was a some point in the past but got lost in a refactor.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
drostron: As far as I can tell, no. Do you see something that can be closed? Buckets were being closed until finding the undesired behavior of all bucket connections to the same bucket closing upon the close of any one of them. Maybe we want to have a connection registry of some sort, but I haven't seen any hints of performance implications either way yet?

effect/src/main/scala/quasar/effect/UUID.scala
drostron: Hrm, true. A touch surprising there isn't a non-cryptographic flavor. Will look into alternatives.

drostron:
I wonder too if `randomUUID` was causing problems, although I'm almost certain I saw that issue before making the switch to the `UUID` effect. I'll give those tests another shot after the switch away from `randomUUID`.

I'm also a touch concerned about change in behavior for missing files. It's been discussed a bit in #backend, but there wasn't strong consent or dissent. Thoughts from the team on the matter are welcomed.

drostron:
Created a few issues and pushed a few commits addressing comments.

drostron:
@mossprescott Assuming this passes as [over here](https://travis-ci.org/drostron/quasar/builds/163468304), is there anything that demands further discussion such as missing file behavior?

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
mossprescott: Looking at the docs and the source (https://github.com/couchbase/couchbase-java-client/blob/master/src/main/java/com/couchbase/client/java/query/DefaultN1qlQueryResult.java), this isn't really a cursor, but has all the rows. So I'd say for now you're better off just returning the whole result in a single chunk and then you don't have the issue of holding on to the effectful iterator either.

Also, that's not memory safe, so I'd say we need a TODO to figure out how to stream a large result.

drostron: Yep, I'm cool with that. It's on the aforementioned undocumented list of ToDos. Your suggestion will make it more explicit.

mossprescott:
Looks like this needs a little more merging to build cleanly.

I'd like to see `Vector[JsonObject]` instead of `Iterator[JsonObject]` now that I understand how these results are being returned, and that seems like an easy switch.

After thinking it over, I still don't see any way around changing the semantics for missing vs. empty files, but I'm not comfortable just switching without some consideration of the fallout. Maybe we can take a few minutes during the tech talk slot to talk it over when the front-end folks are there to chime in? What I'm looking for is general agreement on:

1. It's OK for backends to differ about whether empty files are observable or not.
1. SlamData never needs to detect the presence of an empty file to behave correctly.

What do you think?

drostron:
Will merge in master once the PR fixing it is merged.

Switching to `Vector[JsonObject]` sounds good.

Agreed, soliciting feedback on those points at tomorrow's tech talk would be good.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
```diff
@@ -78,15 +79,11 @@ object readfile {
...
-      if (cursor.result.hasNext) {
-        val jsonObjStr = cursor.result.next.toString
-        DataCodec.parse(jsonObjStr).bimap(
-          err => FileSystemError.readFailed(jsonObjStr, err.shows),
-          d => (cursor, Vector(d)))
-      }
-      else {
-        (cursor, Vector.empty[Data]).right
-      }
+      cursor.result.headOption.cata(
```
mossprescott: I think you'd be much better off here just returning the whole result `Vector` in one chunk. Even for fairly modest results you'll see the overhead with 1-doc chunks.

If you want to be fancy you could peel off some fixed-size prefix as a chunk (e.g. `def peelChunk(size: Int): (Vector[JsonObject], Cursor)`.)

mossprescott:
Sorry about the continuing back-and-forth. This looks good other than the one thing, and if you're not in a position to go back to this branch just say so and I'll merge it as is so we can get this into master.

couchbase/src/main/scala/quasar/physical/couchbase/fs/readfile.scala
drostron: Sounds good. I'll go with the unfancy version for the moment and remove bucket from Cursor while there. I have the QScript work happily rebasing off this so it's reasonable to make changes.

mossprescott: Good catch. I completely overlooked that the bucket was also in there.

effect/src/main/scala/quasar/effect/UUID.scala
drostron: Cool, thanks. I've already switched away from it and am using Couchbase's uuid function that I was previously unaware of. The file is deleted but github doesn't appear to be collapsing comments like it used to. Would have been good to combine efforts if it was sticking around.

drostron:
Ready for another quick look.

drostron:
@mossprescott This should be ready.

Issue #1450, opened 2016-09-20
Encoded + in path is decoded as a space


mossprescott:
For example, uploading a file via `PUT /data/fs/a%2Bb` results in a file called `a b`.

Discovered by this failure in an unrelated build, spotted by @alissapajer:
```
[error]       x succeed with view path (45 seconds, 858 ms)
[error]        Falsified after 89 passed tests.
[error]        > ARG_0: FileIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(DirIn(Current,DirName(/tĮnrenh/3cp)),DirName(c8.eq/f)),DirName(r틳ovo)),DirName(䡅)),DirName(blwz)),DirName(jeox6jdtmfm.cx)),DirName(3)),FileName(H))
[error]        > ARG_1: DirIn(DirIn(DirIn(DirIn(DirIn(Root,DirName(i/1MbnO.c2j//fowubfhowqek./.jk/ka.Ov.5g.9z+drdYohqd.if/y/gr..6f..i.v/.)),DirName(g뱿r5lou)),DirName(eXy)),DirName(t)),DirName(j))
[error]        > 'added /i$sep$1MbnO.c2j$sep$$sep$fowubfhowqek.$sep$.jk$sep$ka.Ov.5g.9z drdYohqd.if$sep$y$sep$gr..6f..i.v$sep$./g뱿r5lou/eXy/t/j/$sep$tĮnrenh$sep$3cp/c8.eq$sep$f/r틳ovo/䡅/blwz/jeox6jdtmfm.cx/3/H'
[error]        
[error]         is not equal to 
[error]        
[error]        'added /i$sep$1MbnO.c2j$sep$$sep$fowubfhowqek.$sep$.jk$sep$ka.Ov.5g.9z+drdYohqd.if$sep$y$sep$gr..6f..i.v$sep$./g뱿r5lou/eXy/t/j/$sep$tĮnrenh$sep$3cp/c8.eq$sep$f/r틳ovo/䡅/blwz/jeox6jdtmfm.cx/3/H' (QuasarSpecification.scala:31)
[error] Actual:   ...5g.9z[ ]drdYo...
[error] Expected: ...5g.9z[+]drdYo...
```

paulp:
This indicates it is going through the decoder twice. The first time it translates it to a plus, the second time to a space. Decoding isn't idempotent, unfortunately. I'm not sure where the unexpected decoding is happening.

paulp:
[...]

PR #1473, opened 2016-09-27
update to scala-pathy 0.2.2

mossprescott:
The new generators find bugs in path encoding by covering ASCII much
better than before. There are no such bugs at the moment, but the new
generators should surface them immediately in the future.

See #1450.

_Note: this build is going to fail because there *are* bugs since #1463 is not merged yet. I'll update when that's in so we can see the difference._


drostron:
👍🏾

PR #1474, opened 2016-09-27
1431 connection error detail

### LINKED ISSUE #1431: Request for enhancement: connection troubleshooting
Attempting to connect with Percona's MongoDB server, which admittedly we don't currently list as supported, results in an error:

```Warning: Failed to mount '/c7/' because 'Connection failed: For input string: "8-2"'.```

The only parameters given were hostname, port and mount type.  Connecting via mongo shell was successful with no additional parameters beyond hostname and port.

Quasar / SlamData log files provide no additional information.  MongoDB log file shows that connections were made but then immediately disconnected:

```
2016-09-12T14:40:51.978-0500 I NETWORK  [initandlisten] connection accepted from 127.0.0.1:41742 #24 (1 connection now open)
2016-09-12T14:40:58.592-0500 I NETWORK  [initandlisten] connection accepted from 192.168.138.1:56864 #25 (2 connections now open)
2016-09-12T14:40:58.594-0500 I NETWORK  [initandlisten] connection accepted from 192.168.138.1:56865 #26 (3 connections now open)
2016-09-12T14:40:58.597-0500 I NETWORK  [conn26] end connection 192.168.138.1:56865 (2 connections now open)
2016-09-12T14:40:58.597-0500 I NETWORK  [conn25] end connection 192.168.138.1:56864 (1 connection now open)
```

From a user / customer perspective I don't think there are any other troubleshooting steps that can be taken beyond modifying code to possibly extract more meaningful information.

@mossprescott asked that I submit an enhancement request as several questions should be answered:

- Should we log it?
- How / where to log it?
- Include the entire stack trace?

### PR DESCRIPTION
**Pull request by mossprescott**
Fixes #1431.

### CODE REVIEW

**Comment by mossprescott** [comment]
@wemrysi: what's not captured here is any exception that might be produced by the driver after our own connection timeout has expired and we've already responded to the client with "Connection failed: TimeoutException." I guess our only option there would be to log the exception when it finally does occur, but I'm not sure how to do that with Task; it seems to kill any subsequent steps when the timeout is hit.

However that doesn't seem to be happening in the particular case that @damonLL was looking at, so I think this will be helpful as is.

**Comment by wemrysi** [comment]
@mossprescott Hm, yeah we may need to use a different timeout mechanism if we want to do that, but I agree this is an improvement over what we have.

LGTM, though you have a compilation error that needs attention, I'll merge when green.

**Comment by mossprescott** [comment]
@wemrysi ready for you

**Comment by mossprescott** [comment]
Seems like the broken build was unrelated to the PR. Feel free to merge.

**Comment by damonLL** [comment]
Not sure if I should leave this here or on #1431, apologies in advance.  We get just a tad more detail, but not much.

REPL v 11.5.4 provides:
```
Warning: Failed to mount '/c7/' because 'Connection failed: For input string: "8-2"'.
```

and the new REPL v 11.8.15 provides:
```
Warning: Failed to mount '/c7/' because 'Connection failed: java.lang.NumberFormatException: For input string: "8-2"'.
```

This is testing against Percona v 3.2.8.

Is this at all helpful?

mossprescott:
Perhaps. What is the connectionUri you're using?

wemrysi:
@mossprescott @damonLL I _think_ the issue is due to where we check the version of the mongo server we connect to. I suspect the percona server is returning a version string like `3.2.8-2` and we split that string on `.` and then try and parse each component as an `Int`, which fails on the last component.

I tried just appending the string `-2` to any version and am able to reproduce the connection error above.

@damonLL Would it be possible to ask the customer to execute `db.runCommand({"buildinfo": 1})` at the mongo console connected to Percona and share the output with us so that we can confirm?

mossprescott:
That's a great piece of deduction. As far as I know, @damonLL is testing this on his own, so he should able to run `db.serverBuildInfo()`.

PR #1475, opened 2016-09-27
Cut time to compile QScriptCore from 155 seconds to 20 seconds.

paulp:
The dramatic impact of this change means we should immediately
investigate the other especially slow-to-compile files for
similar opportunities. Every call to freeTransCata is a likely
candidate, but not only those.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/QScriptCore.scala
+++ b/connector/src/main/scala/quasar/qscript/QScriptCore.scala
@@ -317,47 +317,35 @@ object QScriptCore {
     }
 
   // show is needed for debugging
-  implicit def normalizable[T[_[_]]: Recursive: Corecursive: EqualT: ShowT]:
-      Normalizable[QScriptCore[T, ?]] =
-    new Normalizable[QScriptCore[T, ?]] {
-      val opt = new Optimize[T]
+  implicit def normalizable[T[_[_]]: Recursive: Corecursive: EqualT: ShowT]: Normalizable[QScriptCore[T, ?]] = {
+    type QTotal[A] = QScriptTotal[T, A]
+    type QCore[A]  = QScriptCore[T, A]
 
-      def normalize = new (QScriptCore[T, ?] ~> QScriptCore[T, ?]) {
-        def apply[A](qc: QScriptCore[T, A]) = qc match {
-          case Map(src, f) => Map(src, normalizeMapFunc(f))
-          case LeftShift(src, s, r) =>
-            LeftShift(src, normalizeMapFunc(s), normalizeMapFunc(r))
-          case Reduce(src, bucket, reducers, repair) =>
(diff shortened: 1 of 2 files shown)


alissapajer:
+1 for merge ASAP

connector/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -317,47 +317,35 @@ object QScriptCore {
...
-              freeTransCata(from)(liftCo(opt.applyToFreeQS[QScriptTotal[T, ?]])),
-              freeTransCata(count)(liftCo(opt.applyToFreeQS[QScriptTotal[T, ?]])))
-          case Unreferenced() => Unreferenced()
-        }
+  implicit def normalizable[T[_[_]]: Recursive: Corecursive: EqualT: ShowT]: Normalizable[QScriptCore[T, ?]] = {
+    type QTotal[A] = QScriptTotal[T, A]
+    type QCore[A]  = QScriptCore[T, A]
+
+    val opt = new Optimize[T]
+    def freeTC(free: FreeQS[T]): FreeQS[T] = freeTransCata[T, QTotal, QTotal, Hole, Hole](free)(liftCo(opt.applyToFreeQS[QTotal]))
```
alissapajer: @paulp Adding this type parameters here was the change that reduced the compile time?

Did any other changes you made here affect compile times? So I know in order to make analogous changes elsewhere.

paulp: We should pinpoint it more definitively, but yes: it should be adding those type parameters, and also reducing the number of call sites to freeTransCata from 6 to 1. So even if I didn't supply the type parameters I think we'd have seen most of the gain.

alissapajer: Hm, but we're still calling the function `freeTC` that calls `freeTransCata` a total of 6 times. 

So to be clear, this means that when scalac is compiling this file, previously it would have had to do 6 times the work resolving the implicits necessary for `freeTransCata`? Meaning, it does not cache or memoize implicit searches?

paulp: Each of those calls to freeTransCata is a different implicit search. It can't cache those results because it has to infer the type arguments. There would have to be a heavily polymorphic cache. This is a good example of why it's a bad idea to utilize scalac like a constraint solver. It isn't. Everything is ad hoc. Nothing is optimized for this kind of usage.

sellout:
@paulp Yeah, agreed that this is a change that should have been made regardless of performance. But thankfully, such an improvement.

paulp:
I think the failures all stem from the usual mix of unrelated matters.

alissapajer:
merging! build is green!

PR #1476, opened 2016-09-28
improve compile time by reducing Injectable implicit search

mossprescott:
The `coproduct` implicit was making this quadratic in the size of the QScript coproducts. With it no longer implicit, the `Injectable` instance now has to be defined manually for each new coproduct.

The effect is about 2 minutes off the time for `sbt clean connector/compile` (from about 6 minutes down to 4).


sellout:
Doesn’t this also need cases for the connector-specific coproducts?

mossprescott:
The only ones left are MongoQScript(Internal). Not sure why I didn't have to define Injectable's there, but it compiles and the tests run as is.

sellout:
Yeah, if we don’t need it, then ¯\\\_(ツ)_/¯

Merging …

mossprescott:
Seems solid and possibly taking about 10 minutes off each build (modulo Travis weirdness).

PR #1477, opened 2016-09-28
Enable the Equals, NoNeedForMonad, and Overloading warts.

sellout:
Some are still disabled at the project level.


sellout:
Looks like puffnfresh/wartremover#268 is happening when we have coverage on.

drostron:
I'm generally enthusiastic and in favor of additional wart enablement, but am a bit cautious at the moment due to the traversal per wart cost and the slow compile times we've been dealing with lately. It would be great to have some performance insight. Ideally the cost is low?

connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -106,6 +106,9 @@ object MapFunc {
   object ConcatArraysN {
     private implicit def implicitPrio[F[_], G[_]]: Inject[G, Coproduct[F, G, ?]] = Inject.rightInjectInstance
 
+
```
sellout: Well, that’s interesting …

drostron:
Great to see a few `==` and `!=` going away.

foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -317,36 +317,24 @@ trait SKI {
...
-  def κ[A, B](x: B): A => B                                 = _ => x
-  def κ[A, B, C](x: C): (A, B) => C                         = (_, _) => x
-  def κ[A, B, C, D](x: D): (A, B, C) => D                   = (_, _, _) => x
-  def κ[A, B, C, D, E, F, G](x: G): (A, B, C, D, E, F) => G = (_, _, _, _, _, _) => x
+  /** A shorter name for the constant function of 1, 2, 3, or 6 args.
+    * NB: the argument is eager here, so use `_ => ...` instead if you need it
+    *     to be thunked.
+    */
+  def κ[A, B](x: B): A => B                                  = _ => x
+  def κ2[A, B, C](x: C): (A, B) => C                         = (_, _) => x
```
drostron: Of course, that wart escaped the tab title width 😉 .

drostron:
LGTM. Will merge upon a ✅ build.

drostron:
Oh, needs that comment update first too.

drostron:
Did you have ideas for the build failure with coverage on other than disabling `NoNeedForMonad`?

sellout:
Just wait for a patch for the Wart Remover bug – but we can disable it only on the projects that trigger that failure, at least.

sellout:
Updated – hopefully a green build this time around.

drostron:
A merge from master will likely address the build failures.

PR #1479, opened 2016-09-28
Add -> type alias.

paulp:
It should actually be to Product2, not Tuple2, but scala
has been overspecifying Tuple2 for such a long time that
everything in the world expects a Tuple2 in places which
should accept Product2. Que sera sera.


paulp:
The absence of this type alias has slowly driven me mad as I wind up adding it in every branch I create with more than a few lines. The readability of `M[A->B]` vs. `M[(A, B)]` is incomparably better, especially once the nesting of types and interspersion of value argument lists begins.

paulp:
Sigh, will have to come back to.

PR #1484, opened 2016-09-29
fix a flaky test of Config encoding

mossprescott:
A random failure occurred in a build of #1474.

As far as I can tell, this was very unlikely to happen just because `minTestsOk` was set so low that we were only testing small inputs, except when we got very lucky.

It was easy to reproduce by increasing that param, but the results were impossible to analyze because they were simply huge. After limiting the size in the generator, I've run thousands of tests without a failure, and we're certainly testing non-trivial configs of several mounts and variables, based on printing some of them out.

### DIFF
--- a/core/src/test/scala/quasar/fs/mount/MountingsConfigArbitrary.scala
+++ b/core/src/test/scala/quasar/fs/mount/MountingsConfigArbitrary.scala
@@ -23,13 +23,18 @@ import quasar.contrib.pathy.{APath, PathArbitrary}
 //     for scala collections.
 import scala.Predef._
 
-import org.scalacheck.Arbitrary, Arbitrary.arbitrary
+import org.scalacheck.{Arbitrary, Gen}, Arbitrary.arbitrary
 
 trait MountingsConfigArbitrary {
   import MountConfigArbitrary._, PathArbitrary._
 
+  private def scaleSize[A](gen: Gen[A]): Gen[A] =
+    Gen.sized(size => Gen.resize(size/40 + 1, gen))
+
   implicit val mountingsConfigArbitrary: Arbitrary[MountingsConfig] =
-    Arbitrary(arbitrary[Map[APath, MountConfig]] map (MountingsConfig(_)))
+    // NB: this is a map, and MountConfig contains a map, so if the size isn't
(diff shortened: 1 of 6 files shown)


core/src/test/scala/quasar/fs/mount/MountingsConfigArbitrary.scala
```diff
@@ -23,13 +23,18 @@ import quasar.contrib.pathy.{APath, PathArbitrary}
...
 import scala.Predef._
 
-import org.scalacheck.Arbitrary, Arbitrary.arbitrary
+import org.scalacheck.{Arbitrary, Gen}, Arbitrary.arbitrary
 
 trait MountingsConfigArbitrary {
   import MountConfigArbitrary._, PathArbitrary._
 
+  private def scaleSize[A](gen: Gen[A]): Gen[A] =
+    Gen.sized(size => Gen.resize(size/40 + 1, gen))
```
wemrysi: Out of curiosity, would something like `sqrt(size)` or even just `min(size, <something>)` be more appropriate? What motivated the `40`?

If nothing else, maybe a comment giving the rationale (or lack thereof) for the constant would be useful for posterity/future changes.

wemrysi:
One minor comment regarding the constant, otherwise LGTM.

core/src/test/scala/quasar/fs/mount/MountingsConfigArbitrary.scala
mossprescott: I've done all of those things at one time or another. It's 40 because 200 is the default max size and 200/40 is 5, so this will grow up to 5 (or 6?). I hesitate to document such an ugly hack too grandly, but I suppose until I completely lose it and write a scalacheck replacement, I could make this a bit more clear.

How about if I move this somewhere where I can re-use it? https://github.com/quasar-analytics/quasar/blob/7503789a9a7113c466a8884a83424c59f86d359d/connector/src/test/scala/quasar/fs/FileSystemFixture.scala#L85-L93

wemrysi: Aye, that sounds good :+1:

PR #1485, opened 2016-09-29
Cleanups with T[_[_]] types and lambdas.

paulp:
Don't define classes inside of package objects if it can be avoided.
There is a lot of accidental complexity there.

Use polymorphic lambdas more frequently.

Adapt SimplifyProjection methods similarly to Normalize.

Eliminate more construction boilerplate.


paulp:
These tests all passed, then I added a changelog. I've restarted the first job four times now, failed every time. Babysitting travis to demonstrate that the tests still pass after I added three lines of text is a very stupid use of my time and attention, so I'm going to stop.

sellout:
Yeah, I’m fine with that. Also, if you’re just adding a changelog, you can include `[skip ci]` in your commit and it won’t try to build (which I think also uses some Travis “quota” after which our builds take a lot longer to start … but I could be making that up).

paulp:
I merged master in, since the test which wouldn't pass appears to have been fixed in master in the meantime.

sellout:
Umm … 😕 the changelog is in the wrong place. (Submit it with `[skip ci]` and I’ll merge immediately)

PR #1489, opened 2016-09-30
Handle MongoDB version with extra info

### LINKED ISSUE #1488: MongoDB connection fails if the server version contains anything other than x.y.z
See https://github.com/quasar-analytics/quasar/pull/1474#issuecomment-250755716.

Awaiting confirmation from @damonLL that this is causing the issue reported in #1431, but in any case it's clear that version parsing code is unsafe and will throw an exception if the version string is anything other than digits separated by dots.

mossprescott:
Fixes #1488.

Now parsing into `ServerVersion`, or else failing with a message that says we couldn’t parse the version. A proper `Order` instance for versions, and actual tests.

Somewhat confusingly, the resulting errors imply that _MongoDB_ had trouble parsing the version:

```
Failed to mount '/no-auth/' because 'Connection failed: com.mongodb.MongoException: Unable to parse server version: foo'
```

This is just because we're wrapping errors in their exception type for historical reasons. There were already some FIXMEs in the code related to this, and it won’t be simple to unravel, so I left it as is for now.


mongodb/src/test/scala/quasar/physical/mongodb/ServerVersionSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+
+    "compare with different minor" >> prop { (maj: Int, min: Int, rev: Int, extra: String) =>
+      min < Int.MaxValue ==> {
+        ServerVersion(maj, min, Some(rev), extra) must
+          beLessThan(ServerVersion(maj, min + 1, None, ""))
+      }
+    }
+  }
+}
+class ServerVersionFSpec extends org.specs2.scalaz.Spec with ArbitraryServerVersion {
```
wemrysi: I take it wasn't possible to mix `Qspect` with `scalaz.Spec`?

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIOWorkflowExecutor.scala
```diff
@@ -115,7 +115,7 @@ private[mongodb] object MongoDbIOWorkflowExecutor {
   import EnvironmentError._
 
   /** The minimum MongoDbIO version required to be able to execute `Workflow`s. */
-  val MinMongoDbVersion = List(2, 6, 0)
+  val MinMongoDbVersion = ServerVersion(2, 6, None, "")
```
wemrysi: Thoughts on defining constants (maybe in the `ServerVersion` companion or otherwise appropriate location) for this and the other distinguished versions we switch on?

wemrysi:
@mossprescott A couple minor comments, otherwise LGTM.

mongodb/src/test/scala/quasar/physical/mongodb/ServerVersionSpec.scala
mossprescott: Exactly. I stole the trick from another file where we did the same thing.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIOWorkflowExecutor.scala
mossprescott: Sure. Coming up.

mongodb/src/test/scala/quasar/physical/mongodb/ServerVersionSpec.scala
mossprescott: Actually I realized there was one thing I didn't try and it seems to work. One more commit coming.

mossprescott:
I canceled the first build to save Travis’s precious resources. Should be good to go once it builds.

wemrysi:
@mossprescott Oh, interesting it worked that way. I thought `scalaz.Spec` was a trait, surprising it didn't work as `Qspec with scalaz.Spec`

wemrysi:
@mossprescott Not possible to FF, needs master merged in :disappointed:

mossprescott:
It is a trait, so obviously there's more to it than what I said. Here's the confusing error: 
`illegal inheritance; superclass Qspec is not a subclass of the superclass Spec of the mixin trait Spec`

The scalaz one is `trait Spec extends org.specs2.mutable.Spec`, and `org.specs2.mutable` has `abstract class Spec extends SpecLike`. So apparently a trait can extend an abstract class?

The answer is, why yes, of course (http://stackoverflow.com/questions/12854941/why-can-a-scala-trait-extend-a-class#12854994). Funny thing is I'm not surprised that I didn't understand how this works.

PR #1493, opened 2016-10-02
Option normalize

alissapajer:


connector/src/main/scala/quasar/qscript/TTypes.scala
```diff
@@ -103,37 +103,58 @@ class NormalizableT[T[_[_]] : Recursive : Corecursive : EqualT : ShowT] extends
...
-        freeTC(ej.rBranch),
-        freeMF(ej.lKey),
-        freeMF(ej.rKey),
-        ej.f,
-        freeMF(ej.combine)
-      )
-    )
-  )
+    λ[EquiJoin ~> (Option ∘ EquiJoin)#λ](ej =>
+      (freeTCEq(ej.lBranch), freeTCEq(ej.rBranch), freeMFEq(ej.lKey), freeMFEq(ej.rKey), freeMFEq(ej.combine)) match {
```
alissapajer: The `makeNorm` function is specialized to two arguments, but I'd like to generalize it to _n_ arguments to handle this case, and others.

connector/src/main/scala/quasar/qscript/TTypes.scala
```diff
@@ -103,37 +103,58 @@ class NormalizableT[T[_[_]] : Recursive : Corecursive : EqualT : ShowT] extends
...
+  }
   def freeMF[A](fm: Free[MapFunc, A]): Free[MapFunc, A] = {
     freeTransCata[T, MapFunc, MapFunc, A, A](fm)(
       repeatedly(MapFunc.normalize[T, A]) compose orOriginal(MapFunc.foldConstant[T, A])
     )
   }
+  def freeMFEq[A: Equal](fm: Free[MapFunc, A]): Option[Free[MapFunc, A]] = {
+    val fmNormalized = freeMF[A](fm)
+    (fm ≟ fmNormalized).fold(None, fmNormalized.some)
+  }
```
sellout: This and `freeTCEq` are frustrating, but I guess it’s how it’s gotta be until we have mutual-rec.

connector/src/test/scala/quasar/qscript/Optimize.scala
```diff
@@ -65,7 +66,7 @@ class QScriptOptimizeSpec extends quasar.Qspec with CompilerHelpers with QScript
       val run =
         (SimplifyProjection[QSI, QSI].simplifyProjection(_: QSI[Fix[QSI]])) ⋙
           liftFF(repeatedly(Coalesce[Fix, QSI, QSI].coalesce(idPrism))) ⋙
-          Normalizable[QSI].normalize ⋙
+          repeatedly(Normalizable[QSI].normalize(_: QSI[Fix[QSI]])) ⋙
           liftFF(repeatedly(Coalesce[Fix, QSI, QSI].coalesce(idPrism))) ⋙
           liftFF(repeatedly(opt.compactQC(_: QScriptCore[Fix, Fix[QSI]])))
```
sellout: I think that if there’s a reason to test at some level between an individual normalization and `applyAll`, then that should probably be reflected in the non-test code. I.e., this should either apply only one transformation or use `applyAll`. Not least because the current approach allows it to get out of sync with the order of the non-test transformations.

connector/src/main/scala/quasar/qscript/TTypes.scala
```diff
@@ -142,23 +163,75 @@ class NormalizableT[T[_[_]] : Recursive : Corecursive : EqualT : ShowT] extends
...
+          }
+
+        val orderNormOpt: Option[List[(FreeMap[T], SortDir)]] =
+          if (orderOpt.map(_.toList).flatten.isEmpty)
+            None
+          else
+            (orderOpt.zip(order).map {
+              case (Some(norm), _) => norm
+              case (None, red)     => red
+            }).some
```
alissapajer: Because `zipWith` is also defined in `Traverse` [here](https://github.com/scalaz/scalaz/blob/9eeccf5f2634c6dd91127ccca3409d7aec4998e3/core/src/main/scala/scalaz/Traverse.scala#L132), which requires the RHS to be handled with an `Option`, we have to explicitly invoke `Zip[List]` in order to use the [`zipWith`](https://github.com/scalaz/scalaz/blob/ea856759e60d0d3fbf2becc7b4e1918ecdf70085/core/src/main/scala/scalaz/Zip.scala#L29) that truncates the shorter of the lists. 

```scala
Zip[List].zipWith(orderOpt, order)(_.getOrElse(_)).some
```

But that's definitely what we want.

foundation/src/main/scala/quasar/Predef.scala
```diff
@@ -36,6 +36,10 @@ class Predef extends LowPriorityImplicits {
    */
   type EndoK[F[X]] = scalaz.NaturalTransformation[F, F]
 
+  // TODO generalize this and matryoshka.Delay into
+  // `type KleisliK[M[_], F[_], G[_]] = F ~> (M ∘ G)#λ`
+  type NTComp[F[X], G[Y]] = scalaz.NaturalTransformation[F, matryoshka.∘[G, F]#λ]
```
alissapajer: @sellout how is `EndoK` different? Meaning, do you think that `EndoK` should be here but `NTComp` should be in `quasar.fp`?

connector/src/test/scala/quasar/qscript/Optimize.scala
alissapajer: Sounds reasonable to me, I'll remove the `repeatedly` in the tests and use `orOriginal` here instead.

foundation/src/main/scala/quasar/Predef.scala
sellout: Oh, I totally didn’t see that `EndoK` was here. I think that should also be in `quasar.fp` (or even `quasar.contrib.scalaz`?).

In general, I don’t think that anything should be _defined_ in `Predef`, it should just re-export things that are defined elsewhere.

connector/src/main/scala/quasar/qscript/TTypes.scala
alissapajer: Yeah this is the best way I could figure :/

connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -667,11 +667,14 @@ object MapFuncs {
     def apply[T[_[_]]: Corecursive, A](): Free[MapFunc[T, ?], A] =
       Free.roll(Constant[T, Free[MapFunc[T, ?], A]](EJson.fromCommon[T].apply(ejson.Null[T[EJson]]())))
 
-    def unapply[T[_[_]]: Recursive, A](mf: Free[MapFunc[T, ?], A]): Boolean = mf.resume.fold ({
-      case Constant(ej) => CommonEJson.prj(ej.project).fold(false) {
+    def isNull[T[_[_]]: Recursive](ej: T[EJson]): Boolean =
```
alissapajer: Good call.

alissapajer: Moved to `ejson.EJson`.

alissapajer:
@sellout back in your court. Addresses PR comments and adds a second commit generalizing `applyAll`.

alissapajer:
Dangit, I'll resolve those conflicts :)

alissapajer:
@sellout merge conflicts resolved.

connector/src/test/scala/quasar/qscript/Optimize.scala
```diff
@@ -66,7 +66,7 @@ class QScriptOptimizeSpec extends quasar.Qspec with CompilerHelpers with QScript
       val run =
         (SimplifyProjection[QSI, QSI].simplifyProjection(_: QSI[Fix[QSI]])) ⋙
           liftFF(repeatedly(Coalesce[Fix, QSI, QSI].coalesce(idPrism))) ⋙
-          repeatedly(Normalizable[QSI].normalize(_: QSI[Fix[QSI]])) ⋙
+          orOriginal(Normalizable[QSI].normalize(_: QSI[Fix[QSI]])) ⋙
```
sellout: Ah yeah, this makes sense.

PR #1487, opened 2016-09-30
Improvements related to LeftShift/Reduce cancelation.

sellout:
- handle ZipMapKeys when canceling;
- simplify `Reduce(LeftShift)` pairs;
- restrict `MapFunc.foldConstant` to guarantee that it can only return a
  constant;
- rename `MapFunc.normalize` to `rewrite`, and create a new `normalize`
  that composes `rewrite` with `foldConstant`;
- upgrade `ReduceFunc` from `Traverse` to `Traverse1`; and
- remove some duplication – including adding `::\::` and `::/::` for
  Injectable instances.

PR #1494, opened 2016-10-02
MarkLogic Planner support for QScriptCore.Reduce

wemrysi:
Adds support for `Reduce` nodes to the MarkLogic planner.

"Support" in this case is functionally complete, but not very performant as index usage is minimal. Future efforts will be required to speed this up.

I think a few of the regression tests might be tenable once #1487 is merged, but definitely not until then.


marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
```diff
@@ -95,4 +110,53 @@ private[qscript] final class QScriptCorePlanner[F[_]: NameGenerator: PrologW: Mo
...
+
+  def reduceFuncCombine(rf: ReduceFunc[FreeMap[T]]): F[XQuery] = rf match {
+    case Avg(fm)              =>
+      combiner(fm)((st, x) => mapFuncXQuery[T, F](fm, x) >>= (qscript.incAvg[F].apply(st, _)))
+
+    case Count(fm)            => combiner(fm)((c, _) => (c + 1.xqy).point[F])
+    case Max(fm)              => combiner(fm)((x, y) => fn.max(mkSeq_(x, y)).point[F])
+    case Min(fm)              => combiner(fm)((x, y) => fn.min(mkSeq_(x, y)).point[F])
+    case Sum(fm)              => combiner(fm)((x, y) => fn.sum(mkSeq_(x, y)).point[F])
+    case Arbitrary(fm)        => combiner(fm)((x, _) => x.point[F])
```
alissapajer: This looks like it will return the _same_ `Arbitrary` value each time this is run. Is that consistent with what we do in mongo?

mossprescott: For MongoDB, it's `$first`, but I could see changing it to `$sample(1)` now that that's available. The idea was not to constrain the choice at all so the backend can use whatever's cheapest.

wemrysi: I thought `Arbitrary` meant arbitrary, so it shouldn't matter, right?

sellout: The intent of `Arbitrary` is to return the “lowest-cost” option (this is distinct from what something like `Random` (which doesn’t exist) might do. It’s not explicitly part of the spec, but I _think_ `Arbitrary` is generally deterministic in as much as the query itself is (i.e., modulo sharding, `order by`, etc.).

`Arbitrary` can (I think) be explicitly written in SQL², but it’s generally added to the LP when we know that all the values in the reduction must be identical, so it doesn’t matter which is selected.

sellout: @mossprescott I would say `$sample(1)` would be used for some (hypothetical) `Random`, not `Arbitrary` (just because it sounds more expensive).

Also, at what level do we determine that the value should be pulled from `_id`? Isn’t that done for `Arbitrary` when possible? If so, it’s not _always_ `$first` in Mongo, because we have an even _more_ efficient approach sometimes.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -50,25 +127,40 @@ object qscript {
...
-                     .let_(n -> xs.string(fn.nodeName(c.xqy)))
-                     .return_(kvEnt)
-          zMap    <- ejson.mkObject[F] apply entries
-        } yield zMap
+        asMKey(bucket fnapply (x.xqy)) map { theKey =>
+          let_(
+            m -> map.map(),
+            o -> for_ (x -> xs) .let_ (
+                   k -> theKey,
+                   v -> if_(map.contains(m.xqy, k.xqy))
```
alissapajer: @wemrysi so the `bucket` of a qscript `Reduce` might not just refer to a map or an array, but could refer to a value in a map or an array. Looking at this implementation here, it wasn't clear to me what this `map` is, and if this allows for buckets to reference any value type, or just something specific to maps?

wemrysi: That map is in-memory state that holds the bucketed reductions, it looks like
```
hashkey_of_bucket_value_1 -> (reduction1, reduction2, ..., reductionN),
hashkey_of_bucket_value_2 -> (reduction1, reduction2, ..., reductionN),
...
hashkey_of_bucket_value_N -> (reduction1, reduction2, ..., reductionN)
```

alissapajer:
@wemrysi can you resolve the conflicts?

alissapajer:
Also, #1487 is merged, so maybe that regression test is tenable? I'm happy to merge without enabling that first though.

alissapajer:
LGTM - I'll merge when it's ready.

wemrysi:
@alissapajer Once I've finished the rest of the `MapFunc`s (PR inc soon) I'll focus on seeing if I can get some of the regression tests enabled, I still need to get a MarkLogic environment setup in travis, anyhow.

Issue #1481, opened 2016-09-29
Exclude ids from `ShiftedRead`s when they aren't used


wemrysi:
When datum ids are not needed for a plan, use `ExcludeId` in `ShiftedRead` to ease the burden on the backend.

wemrysi:
Closed via #1495.

PR #1495, opened 2016-10-03
Optimization for `ShiftedRead`

jedesah:
- Add `shiftRead` convenience function
- Add optimization for `ShiftedRead` and apply in new `shiftRead` function

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,2 @@
+- Add `shiftRead` convenience function
+- Add optimization for `ShiftedRead` and apply in new `shiftRead` function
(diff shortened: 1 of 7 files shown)


wemrysi:
@jedesah @alissapajer Does this close #1481, or are there other cases that need handling?

alissapajer:
@wemrysi looking at the ticket, this should close it. 

To be clear, in the simple qscript cases this normalization will be successfully applied. But I can imagine some complicated qscript that is not fully normalized for _other_ reasons, and so this specific normalization case won't be hit, even though it should be. But I think those cases should just be filed as separate bugs.

PR #1497, opened 2016-10-03
Bump default maximum heap size in sbt launcher

jedesah:
Was running into consistent `OutOfMemoryException` attempting to publish.


jedesah:
@paulp Curious to hear your thoughts on this. Maybe this isn't the best place to make this change, I figured it would be nice for new contributors to not have to mess with this stuff if they pull down quasar.

On the flip side, probably this change will get wiped next time we pull down an updated version of the launcher.

At one point I had added a `.sbtopts` file to quasar to attempt to solve this problem in a way that does not require using this particular launcher and would not get wiped away by an update of the launcher script, but you removed that I file and I forgot the details of the conversation we had around why you did that.

paulp:
@jr 1536m was chosen because it is small enough not to prevent 32bit jvms from starting. I don''t know if there are 32bit jvms anymore.

The only line in the .sbtopts I removed was `-J-XX:MaxMetaspaceSize=512m` which wouldn't have helped you to not run out of heap.

jedesah:
Okay, so maybe a better solution would be to add a `.sbtopts` with `-Xmx3g` under the assumption that most `quasar` contributors are running on 64bit vms and leave the launcher script untouched.

jedesah:
@paulp Do you with my above statement? If so, I'll update this PR with it.

paulp:
Yes, sorry to neglect. I think leaving the launcher script untouched and adding .sbtopts is better from a futureproofing standpoint. I might suggest something smaller than 3g though. If you only hit the current limit when publishing then you're already in "niche usage" territory.

paulp:
Sorry to only think of this now, but the more direct way to do this is to add `.jvmopts` with `-Xmx2g` rather than routing it through the runner.

paulp:
Superseded by #1511.

PR #1501, opened 2016-10-04
Implement `union` for the old and new Mongo connoctors.

sellout:


mossprescott:
The `union` test needs to be `pending` for `mongodb_read_only`.

Otherwise, LGTM.

PR #1507, opened 2016-10-04
Centralize creation of common value.

paulp:
The value ExprOpCoreF.fixpoint[Fix, ExprOp] was being created
from 17 distinct call sites.


alissapajer:
LGTM. These builds are really taking their time to start though.

PR #1508, opened 2016-10-04
Eliminate the separate compilation diverging implicit error.

paulp:
It's just black magic. Import only what you need and certain
kinds of bugs go away for a while.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- work around scala separate compilation bug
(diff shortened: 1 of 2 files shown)


sellout:
😭

sellout:
This could probably use a couple comments on the new imports so I don’t accidentally revert this next time I clean up imports in here. (`[skip ci]`)

paulp:
One really gets a skewed picture looking at diffs. Looking at the whole file, it turns out the two imports I added were in the same scope. So I went with not skipping ci since it's a whole new patch.

PR #1511, opened 2016-10-05
Add `jvmopts` with larger maximum heap

paulp:


paulp:
Failure was actually legitimate. I updated.

PR #1512, opened 2016-10-05
decode MongoDB index with Double instead of Integer

mossprescott:
Ran into this when preparing for the 3.1 demo. No idea why I didn't see it previously using the same version of MongoDB.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- more lenient decoding of MongoDB indexes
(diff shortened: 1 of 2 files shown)


paulp:
I guess the PR title means "Number" not "Double". Although if mongodb indices can really be doubles, that's completely awesome.

mossprescott:
The values I was expecting to be `java.lang.Integer` can apparently come back as `java.lang.Double`. Both are subtypes of `java.lang.Number`, so this change is the simplest way to deal with both.

These values indicate the index type, and can be `1` or `-1` (meaning ascending or descending), apparently `1.0` or `-1.0`, but also things like `"hashed"` and some other strings that we don't need to recognize. So your basic untyped Java.

mossprescott:
Build failed with a 50 minute timeout at the very last step, so I’m gonna restart and hope.

sellout:
This looks good. Trying to get it to pass, though.

PR #1514, opened 2016-10-05
Eliminate the elusive ThetaJoin.

sellout:
Also,
- have Coalesce push Take/Drop later, rather than earlier
- add an `optimize` transformation to finalize QScript (which pulls Take/Drop earlier)
- implement Mergable for ProjectBucket
- fix conversions of Take and Drop from LP

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Optimize.scala
+++ b/connector/src/main/scala/quasar/qscript/Optimize.scala
@@ -119,17 +119,6 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
       coEnvHtraverse(λ[QScriptTotal[T, ?] ~> (Option ∘ F)#λ](FI.project(_))).apply)
       .map(targ => (targ >> srcCo.fromCoEnv).toCoEnv[T])
 
-  def rebaseTEnv[F[_]: Traverse](
-    target: FreeQS[T])(
-    srcEnv: T[EnvT[Ann[T], F, ?]])(
-    implicit FI: Injectable.Aux[F, QScriptTotal[T, ?]]):
-      Option[T[EnvT[Ann[T], F, ?]]] =
-   freeCataM(
-      target)(
-      interpretM[Option, QScriptTotal[T, ?], Hole, T[EnvT[Ann[T], F, ?]]](
-        κ(srcEnv.some),
-        FI.project(_) ∘ (qs => EnvT((EmptyAnn[T], qs)).embed)))
-
   private val UnrefedSrc =
     Inject[QScriptCore[T, ?], QScriptTotal[T, ?]].inj(Unreferenced[T, FreeQS[T]]())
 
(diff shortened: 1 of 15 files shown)


connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -93,8 +95,18 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
+        QC.inj(LeftShift((),
+          HoleF,
+          Free.roll(MakeMap(
+            StrLit("foo"),
+            Free.roll(Guard(
+              RightSideF,
+              Type.Obj(ScalaMap(),Some(Type.Top)),
+              ProjectFieldR(RightSideF, StrLit("foo")),
+              Free.roll(Undefined())))))))).some)
+    }
```
sellout: Check it out, everyone.

foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -480,18 +480,11 @@ package object fp
...
-    reflNT[F]
-  )
+    reflNT[F])
 
   def coenvPrism[F[_], A] = PrismNT[CoEnv[A, F, ?], F](
     λ[CoEnv[A, F, ?] ~> λ[α => Option[F[α]]]](_.run.toOption),
-    λ[F ~> CoEnv[A, F, ?]](fb => CoEnv(fb.right[A]))
-  )
-
-  def envTPrism[T[_[_]], F[_], A](empty: A) = PrismNT[EnvT[A, F, ?], F](
```
sellout: No – it’s not actually a legit Prism.

alissapajer: because it requires the `empty: A`, it's not legit?

connector/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -304,6 +289,15 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
       F[T[F]] => F[T[F]] =
     applyNormalizations[F, F](idPrism, rebaseT)
 
+  /** Should only be applied after all other QScript transformations. This gives
+    * the final, optimized QScript for conversion.
+    */
+  def optimize[F[_], G[_]: Functor]
```
alissapajer: @sellout it looks like we're not calling this anywhere?

alissapajer:
@sellout aside from question about if we're calling `optimize` anywhere, LGTM. I'll merge once that's addressed.

connector/src/main/scala/quasar/qscript/Optimize.scala
sellout: Right – we just had an optimization hanging around and getting called, incorrectly, as part of normalization. I just put it in this function to give it a more obvious label. The connectors _should_ call this function, but I didn’t add it to any yet.

alissapajer:
@sellout I got a "fatal: Not possible to fast-forward, aborting." Can you merge in master?

sellout:
Done.

PR #1516, opened 2016-10-05
Refactor mounter

mossprescott:
Defines a `PathStore` trait with the necessary operations, with an implementation for `KeyValueStore.

### DIFF
--- a/core/src/main/scala/quasar/fs/mount/Mounter.scala
+++ b/core/src/main/scala/quasar/fs/mount/Mounter.scala
@@ -114,7 +114,7 @@ object Mounter {
       case Unmount(path) =>
         handleUnmount(path)
           .flatMapF(_ =>
-            refineType(path).swap.fold[F[Set[APath]]](κ(Set.empty.point[F]), store.descendants)
+            refineType(path).swap.foldMapM(store.descendants)
               .flatMap(_.toList.traverse_(handleUnmount(_)).run.void))
           .toRight(pathNotFound(path))
           .run
--- a/core/src/main/scala/quasar/fs/mount/Mounter.scala
+++ b/core/src/main/scala/quasar/fs/mount/Mounter.scala
@@ -39,8 +39,6 @@ object Mounter {
     /** Associate a value with a path that is not already present, or else do
       * nothing. Yields true if the write occurred. */
     def insert(path: APath, value: V): F[Boolean]
-    /** Put the store back into a usable state after a failed `insert`. */
-    def recover: F[Unit]
     /** Remove a path and its value, if present, or do nothing */
     def delete(path: APath): F[Unit]
   }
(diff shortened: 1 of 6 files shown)


core/src/main/scala/quasar/fs/mount/Mounter.scala
```diff
@@ -123,23 +114,23 @@ object Mounter {
       case Unmount(path) =>
         handleUnmount(path)
           .flatMapF(_ =>
-            refineType(path).swap.foldMap(forPrefix)
+            refineType(path).swap.fold[F[Set[APath]]](κ(Set.empty.point[F]), store.descendants)
```
wemrysi: I think you could use `foldMapM` here.

core/src/main/scala/quasar/fs/mount/Mounter.scala
```diff
@@ -28,91 +28,82 @@ import scalaz._, Scalaz._
...
-  def apply[F[_], S[_]](
-    mount: MountRequest => F[MountingError \/ Unit],
-    unmount: MountRequest => F[Unit]
-  )(implicit
-    S0: F :<: S,
-    S1: MountConfigs :<: S
-  ): Mounting ~> Free[S, ?] = {
+  /** A kind of key-value store where the keys are absolute paths, with an
+    * additional operation to efficiently look up nested paths. */
+  trait PathStore[F[_], V] {
```
wemrysi: Mind adding some documentation for the methods, especially `insert` and `recover` (i.e. what the `Boolean` represents and what the semantics of `recover` is)?

wemrysi:
This looks good in general, I think the `recover` function is a bit smelly, though I don't have a great alternative (maybe a more declarative way of specifying an atomic region `atomic[A](fa: F[A]): F[A]`?).

mossprescott:
I can't argue with your assessment of `recover`. In fact, I can't say for sure that it's even being used correctly because the tests passed before it was implemented and I only included it in the hope of preserving the previous behavior. I believe it may be necessary only in the case of concurrent conflicting `mount` operations, but looking at it again I'm not sure I understand how it would be necessary to rollback the transaction because it doesn't seem like any subsequent operations are done on `store`. It may depend on how the resulting program (containing `ConnectionIO`) is interpreted.

core/src/main/scala/quasar/fs/mount/Mounter.scala
mossprescott: I will. Alternatively, I could change the signature to `insert(path, value): MntErrT[F, Unit]` which is probably somewhat more self-explanatory. But I think I'll just add a comment.

mossprescott:
I'm not 100% sure, but I think I can eliminate `recover` and just put the rollback directly into the `insert` action. Maybe we can look at it together tomorrow and decide if that seems correct.

wemrysi:
@mossprescott I _think_ (if memory serves) that the `rollback` in QAdv was due to how the PostgreSQL java lib dealt with constraint violations. We are relying on the `PRIMARY KEY` constraint in the `Mounts` table to tell us if a mount already exists, but I think it puts the `Connection` in a state where every subsequent op fails with an error that the transaction is closed (or something like that), so we `rollback` to abort that transaction and start a new one, does that sound familiar?

If that is the case, then I think you're right that the `rollback` can be moved into the `insert` impl for the `MetaStoreMounter`.

PR #1518, opened 2016-10-06
Consistent Injectable, more type aliases

paulp:
- Lower Free[MapFunc...] noise with type alias.
- Utilize FreeQS alias more places.
- More consistent definition of Injectable.


alissapajer:
LGTM

PR #1517, opened 2016-10-05
Finish MarkLogic planning of MapFunc

wemrysi:
* Includes support for all terms except `Undefined`, `IfUndefined` and `Guard` (`Guard` expressions are currently replaced by their `cont` expression, effectively ignoring them for now).
* Also fixes the plan for `QScriptCore.Unreferenced` to have the correct cardinality (1).

PR #1510, opened 2016-10-05
Fix REPL debug output.

sellout:
This also eliminates the implicit Show automatically provided by RenderTree.

PR #1522, opened 2016-10-06
Remove unused import.

alissapajer:
After cleaning the travis cache, this unused import was causing builds to fail.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Remove unused import.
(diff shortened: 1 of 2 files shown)


sellout:
This change is also in #1510, which will _hopefully_ be merged RSN.

alissapajer:
@sellout ah ok cool. I'll keep this open to see if it builds, but if #1510 lands soon, we can close this without merging. The unused import is now blocking other PRs, so we should get this change in ASAP one way or the other.

wemrysi:
Fixed in #1517.

sellout:
This was fixed by #1517.

PR #1523, opened 2016-10-06
Fix a bug in Take/Drop.

sellout:
Take/Drop would lose the location of the result, but now they don’t.

Also some cleanup of EJson pattern-matching, QScript merging, and
Mongo-QScript undefined.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Transform.scala
+++ b/connector/src/main/scala/quasar/qscript/Transform.scala
@@ -115,9 +115,9 @@ class Transform
   /** Contains a common src, the MapFuncs required to access the left and right
     * sides, and the FreeQS that were unmergeable on either side.
     */
-  type MergeResult = (T[F], FreeMap[T], FreeMap[T], Free[F, Hole], Free[F, Hole])
+  type MergeResult = (T[F], FreeQS[T], FreeQS[T])
 
-  def merge0(left: T[F], right: T[F]): MergeResult = {
+  def merge(left: T[F], right: T[F]): MergeResult = {
     val lLin = left.cata(linearize).reverse
     val rLin = right.cata(linearize).reverse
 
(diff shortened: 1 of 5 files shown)


rabbitonweb:
Keeping my fingers crossed for this one :)

jdegoes:
Would ideally have a test that would fail on the old behavior, but demonstrate the new behavior is correct.

alissapajer:
@sellout is there a test you can add for this in qscript spec?

sellout:
@jdegoes Agreed, but this was blocking people test case (and generalizing this fix) coming.

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -556,26 +533,14 @@ class Transform
...
-        (QC.inj(Take(src,
-          lfree.mapSuspension(FI.inject),
-          rfree.mapSuspension(FI.inject))).embed,
-          a1._1.provenance,
-          HoleF,
-          HoleF))
+      val (src, lfree, rfree) = merge(a1._2, a2._2)
 
-      (Ann(buckets, HoleF), qs).right
+      (a1._1, QC.inj(Take(src, lfree, rfree)).embed).right
```
alissapajer: This is a lot cleaner +1

sellout: Yeah, and that’s the main fix right there – we were throwing away the pointer to the value (`a1._1.values`), and so the final result then included the provenance information. This actually happens in a couple other places (although less frequently hit), and that’s the generalization of this fix I mentioned in the other comment.

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -164,45 +155,35 @@ class Transform
...
-              rann.provenance.map(_ >> racc)),
-            lann.values >> lacc,
-            rann.values >> racc)
-        } (qs =>
-          (qs.embed,
-            lann.provenance.map(_ >> lacc),
-            lann.values >> lacc,
-            rann.values >> racc))
-    }
+    val (src, lBranch, rBranch) = merge(left._2, right._2)
```
alissapajer: @sellout since some of this diff is just moving stuff around (like this chunk, I _think_) could you point out where you actually made the change that changed the behavior of `Take` and `Drop`?

alissapajer: Ah, I think you answered this question literally as I was typing it :)

sellout: I did – just two minutes before this comment 😆

sellout: Ok, I feel like GitHub has a sync problem!

alissapajer:
@sellout will merge once it's green.

sellout:
Figure @alissapajer may be done for the night, and I promised @rabbitonweb I’d try to get this in, so since it’s been approved, unless I hear otherwise in the next couple minutes, I’m going to merge it myself.

PR #1524, opened 2016-10-07
Fix handling of “result” pointers.

sellout:
Sometimes we would throw away necessary information for finding the
result of a QScript node. We now apply it properly.

This also has some other improvements to simplifying `ShiftedRead` nodes
and normalization in general.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Coalesce.scala
+++ b/connector/src/main/scala/quasar/qscript/Coalesce.scala
@@ -23,20 +23,32 @@ import quasar.fp.ski._
 import quasar.qscript.MapFunc._
 import quasar.qscript.MapFuncs._
 
-import matryoshka._
+import matryoshka._, Recursive.ops._
+import matryoshka.patterns._
 import scalaz._, Scalaz._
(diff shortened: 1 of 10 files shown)


sellout:
BTW, this includes the test for the change in the previous PR.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -153,6 +232,52 @@ object Coalesce {
...
+        case Drop(src, from, count) =>
+          makeBranched(from, count)(freeQCEq[T])(Drop(src, _, _))
+        case Take(src, from, count) =>
+          makeBranched(from, count)(freeQCEq[T])(Take(src, _, _))
+        case Union(src, from, count) =>
+          makeBranched(from, count)(freeQCEq[T])(Union(src, _, _))
+        case _ => None
+      }
+
+      def coalesceSR[F[_]: Functor]
```
wemrysi: Nice! I started looking into this myself, glad to see you beat me to it.

PR #1525, opened 2016-10-07
Remove $conforms from Predef, and boilerplate reduction.

paulp:


connector/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -16,6 +16,7 @@
 
 package quasar.fs
 
+import scala.Predef.$conforms
```
paulp: `$conforms` is the overwhelming leader for implicits which look plausible enough to satisfy some type that they must be explored, but then don't.

Almost every place where it's needed in the current codebase is a call to either `unzip` or `flatten`, both of which restrict eligibility by requiring a `<:<` which shows the element type to be a tuple or to be nested. But as you know, the right way to accomplish that sort of thing is with a type class. `<:<` is sort of the "subtyping type class" which is dangerously general.

Unzip and flatten methods which are simply called directly with appropriate types would solve the same problem with much less indirection, and most of those imports would go away.

PR #1527, opened 2016-10-07
Tests for date, math, and relations, and fixes for JavaScript

mossprescott:
First part of #1311.

Adds fairly complete tests for `date`, `math`, and `relation` to `StdLibSpec`, and includes all of the fallout from seeing how broken things were.

Breaks out the `Extract` function into a distinct LP Func and QS MapFunc for each field.

Pulls out JavaScript compilation for both MongoDB planners into common code.

If there was a clean way to break this into 3 or so PRs, I would have, but it seemed sensible to start with the tests and so this is what I ended up with.

### DIFF
--- a/frontend/src/main/scala/quasar/SemanticError.scala
+++ b/frontend/src/main/scala/quasar/SemanticError.scala
@@ -90,6 +90,9 @@ object SemanticError {
   final case class InvalidPathError(path: Path[_, File, _], hint: Option[String]) extends SemanticError {
     def message = "Invalid path: " + posixCodec.unsafePrintPath(path) + hint.map(" (" + _ + ")").getOrElse("")
   }
+  final case class UnexpectedDatePart(part: String) extends SemanticError {
+    def message = "Invalid selector for DATE_PART: " + part + " (expected \"century\", \"day\", etc.)"
+  }
 
   // TODO: Add other prisms when necessary (unless we enable the "No Any" wart first)
   val genericError: Prism[SemanticError, String] =
--- a/frontend/src/main/scala/quasar/std/date.scala
+++ b/frontend/src/main/scala/quasar/std/date.scala
@@ -17,8 +17,9 @@
 package quasar.std
 
 import quasar.Predef._
-import quasar.{Data, Func, UnaryFunc, BinaryFunc, GenericFunc, Mapping, Type, SemanticError}, SemanticError._
+import quasar.{Data, Func, UnaryFunc, GenericFunc, Mapping, Type, SemanticError}, SemanticError._
 import quasar.fp.ski._
+
 import org.threeten.bp.{Duration, Instant, LocalDate, LocalTime, Period, ZoneOffset}
 import scalaz._, Validation.success
 import shapeless.{Data => _, _}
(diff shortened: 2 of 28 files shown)


frontend/src/main/scala/quasar/SemanticError.scala
```diff
@@ -90,6 +90,9 @@ object SemanticError {
   final case class InvalidPathError(path: Path[_, File, _], hint: Option[String]) extends SemanticError {
     def message = "Invalid path: " + posixCodec.unsafePrintPath(path) + hint.map(" (" + _ + ")").getOrElse("")
   }
+  final case class UnexpectedDatePart(part: String) extends SemanticError {
```
wemrysi: :tada:

frontend/src/main/scala/quasar/std/date.scala
```diff
@@ -57,15 +58,58 @@ trait DateLib extends Library {
...
+  private def extract(name: String, help: String) =
+    UnaryFunc(
+      Mapping, name, help,
+      Type.Numeric,
+      Func.Input1(Type.Temporal),
+      noSimplification,
+      constTyper[nat._1](Type.Numeric),
+      basicUntyper)
+
+  val ExtractCentury      = extract("extract_century",
```
wemrysi: Appreciate all the docs!

marklogic/src/main/scala/quasar/physical/marklogic/fs/queryfile.scala
```diff
@@ -72,8 +70,8 @@ object queryfile {
         Vector(PhaseResult.detail("XQuery", main.render))
 
       def extractErr(partName: String, msg: String): FileSystemError =
-        FileSystemError.planningFailed(lp, QPlanner.UnsupportedPlan(
-          LogicalPlan.InvokeF(Extract, Input2(partName, "<date/time>")), Some(msg)))
+        FileSystemError.planningFailed(lp, QPlanner.UnsupportedFunction(
```
wemrysi: This could go back inline now that `UnrecognizedDatePart` is gone

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
```diff
@@ -58,30 +58,27 @@ object MapFuncPlanner {
...
+    case ExtractIsoDayOfWeek(time)    => xdmp.weekdayFromDate(xs.date(time)).point[F]
+    case ExtractIsoYear(time)         => MonadPlanErr[F].raiseError(MarkLogicPlannerError.unsupportedDatePart("isoyear"))
+    case ExtractMicroseconds(time)    => mkSeq_(fn.secondsFromDateTime(xs.dateTime(time)) * 1000000.xqy).point[F]
+    case ExtractMillenium(time)       => mkSeq_(mkSeq_(fn.yearFromDateTime(xs.dateTime(time)) mod 1000.xqy) + 1.xqy).point[F]
+    case ExtractMilliseconds(time)    => mkSeq_(fn.secondsFromDateTime(xs.dateTime(time)) * 1000.xqy).point[F]
+    case ExtractMinute(time)          => fn.minutesFromDateTime(xs.dateTime(time)).point[F]
+    case ExtractMonth(time)           => fn.monthFromDateTime(xs.dateTime(time)).point[F]
+    case ExtractQuarter(time)         => xdmp.quarterFromDate(xs.date(time)).point[F]
+    case ExtractSecond(time)          => fn.secondsFromDateTime(xs.dateTime(time)).point[F]
+    // case ExtractTimezone(time)       => qscript.timezoneOffsetSeconds[F] apply (xs.dateTime(time))
```
wemrysi: These are commented because those cases don't yet exist, right? Also, why is that?

wemrysi:
Why are the various examples in the math tests commented out?

Looks good other than that, had a few other comments but nothing major. Will be great to begin taking advantage of these in other connectors.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
mossprescott: These were never implemented for MongoDB (because it knows nothing about time zones), so I just dropped them. I'll add them back since MarkLogic can handle them intelligently.

mossprescott:
It was a struggle to get the math tests to run consistently for the entire domain of `Int` and `Dec` on all three implementations currently being tested, so I ended up commenting some that just wouldn't cooperate. They can't be marked pending in general because they pass on one impl, and the tests aren't yet set up so that a particular impl can choose to make something pending, or even to skip a particular test unless it can be identified by the argument types. So the shortest path was just to leave them in and commented for now. I'll make a pass and add TODOs with an indication of what the problem might be.

PR #1530, opened 2016-10-07
Bring DataServiceSpec test description for non-existent file into agreement with test

drostron:


paulp:
Brilliant! (Would still like to change these semantics though.)

drostron:
Updating the semantics is fine with me. I at least wanted to get rid of this confusion until further decisions are made.

paulp:
@drostron I'm glad you eliminated the confusion; I only meant the comment as an aside. (When I wrote it I thought I was simultaneously merging/closing it, then I ran into the changelog issue, then we had that whole discussion...)

PR #1532, opened 2016-10-07
avoid theta join creation if one side is constant

### LINKED ISSUE #1529: avoid creating one-sided joins in autojoin
Currently we create joins (in `autojoin`) even when one side is a constant value (e.g. `Map(Unreferenced, Constant(Int(12)))`). Then later we attempt to normalize them away with `Optimize.elideOneSidedJoin`. In all the `autojoin` cases we can avoid creating them in the first place.

We will still keep `elideOneSidedJoin` around to catch potential other cases, but it will share most of its logic with the `ThetaJoin` avoidance logic.

May help us toward fixing #1411 since we hit the "unsupported order function" because the constant string value representing the ordering is hidden inside a `ThetaJoin`.

alissapajer:
fixes #1529

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+Avoid creation of a ThetaJoin in autojoin if one side is a constant value.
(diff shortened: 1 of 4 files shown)


sellout:
This looks good – just needs to merge in master.

PR #1533, opened 2016-10-07
various qscript renamings

alissapajer:


connector/src/main/scala/quasar/qscript/Normalizable.scala
```diff
@@ -24,20 +24,20 @@ import scalaz._, Scalaz._
 import simulacrum.typeclass
 
 @typeclass trait Normalizable[F[_]] {
-  def normalize: NTComp[F, Option]
+  def normalizeF: NTComp[F, Option]
```
alissapajer: `normalize` -> `normalizeF`

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -30,7 +30,7 @@ import matryoshka._,
 import matryoshka.patterns._
 import scalaz.{:+: => _, Divide => _, _}, Scalaz._, Inject._, Leibniz._
 
-class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
+class Rewrite[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
```
alissapajer: `Optimize` -> `Rewrite`

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -245,31 +245,31 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
+  def normalizeCoEnv[F[_]: Traverse: Normalizable](
     implicit C:  Coalesce.Aux[T, F, F],
              QC: QScriptCore[T, ?] :<: F,
              TJ: ThetaJoin[T, ?] :<: F,
              FI: Injectable.Aux[F, QScriptTotal[T, ?]]):
       F[T[CoEnv[Hole, F, ?]]] => CoEnv[Hole, F, T[CoEnv[Hole, F, ?]]] =
     applyNormalizations[F, CoEnv[Hole, F, ?]](coenvPrism, rebaseTCo)
 
-  def applyAll[F[_]: Traverse: Normalizable](
+  def normalize[F[_]: Traverse: Normalizable](
```
alissapajer: `applyAll` -> `normalize`

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -245,31 +245,31 @@ class Optimize[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] {
...
+    repeatedly(Normalizable[F].normalizeF(_: F[T[G]])) ⋙
       liftFG(injectRepeatedly(elideNopJoin[F].apply[T[G]])) ⋙
       liftFG(injectRepeatedly(elideOneSidedJoin[F, G](rebase))) ⋙
       repeatedly(C.coalesceQC[G](prism)) ⋙
       liftFG(injectRepeatedly(C.coalesceTJ[G](prism.get))) ⋙
       liftFF(repeatedly(compactQC(_: QScriptCore[T, T[G]]))) ⋙
       (fa => QC.prj(fa).fold(prism.reverseGet(fa))(elideNopQC[F, G](prism.reverseGet)))
 
-  def applyToFreeQS[F[_]: Traverse: Normalizable](
+  def normalizeCoEnv[F[_]: Traverse: Normalizable](
```
alissapajer: `applyToFreeQS` -> `normalizeCoEnv`

connector/src/main/scala/quasar/qscript/Normalizable.scala
alissapajer: I still don't particularly like the overlap in how we use `normalize` (here and in `Rewrite`). Maybe there is something better we can rename this typeclass `Normalizable` to. Because this just represents normalizations within a type, so maybe `NormalizableWithin`? Kinda verbose though...

alissapajer: So, aside from the tests, the only call site of `normalizeF` looks like `repeatedly(Normalizable[F].normalizeF(_: F[T[G]]))` in `applyNormalizations`. The benefit I see of having it a typeclass is that we can individually see what the transformations for each qscript component look like. But, if we just had `F ~> Option[F]`, then we'd define all those transformations in a single place, in that `~>` definition. Then we wouldn't have to file-hop to find all the different ways `normalizeF` is implemented. _And_ it would be easier to println debug.

paulp: Right.

sellout: I do think we should put all instances in the typeclass companion object, so moving them would be great. But I think we need a type class. That’s how you work with coproducts without knowing which coproduct you have.

This way normalize can be applied to QScriptTotal, or MongoQScript, or whatever. I don’t know how else you’d do that.

alissapajer: Though, we would lose the guarantee that a `QScriptCore` node transformation to a `QScriptCore` node, for example, because we'd be implementing this `~>` for a larger `F`. But what is in `QScriptCore` as opposed to another component is actually arbitrary, from the point of view of normalizations... So we already don't guarantee at the type-level that a `Map` transforms to a `Map` and not a `LeftShift`, since those are both `QScriptCore` components.

So I think this idea of a "within" normalization (a normalization within a component) might not be as meaningful as I originally thought.

paulp: I'll stop with mild suggestions for now, but to prep you for the direction I will argue for in the future - I see no reason there can't be an interface which you must implement to play in the coproduct arena. It would lower the complexity a lot to have a `Thingy[S[_]]` rather than an open-ended bundle of coproducts. And once you have a little bit of structure, there's a place to put methods which you might otherwise build a typeclass around. You don't have to forfeit generality to do it this way.

Raw coproducts are hard to deal with for the same reason raw tuples are. You can write a type like `(String, Int, List[Int], Double, Int => Int)`, but there are a lot of good reasons not to. And the performance characteristics of open coproducts and implicit `:<:`s is bad in the best case and unpredictable to boot.

sellout: @alissapajer We don’t particularly care about the guarantee that `QScriptCore ~> QScriptCore`, etc., what we care about is `MongoQScript ~> MongoQScript` or `QScriptTotal ~> QScriptTotal`, without having to reimplement the same things for different combinations, and without connector authors needing to compose operations any more than is necessary (something we already have a problem with).

alissapajer: @sellout right, the whole reason we have these different components _is_ so that backends can choose to only implement a certain selection of nodes.

sellout:
I think this, too, looks good. I think that the Normalizable type class should be renamed – `Rewrite.normalize` is much closer to actual normalization – and we should in general keep instances in the typeclass companion rather than the type companion (at least for these QScripty things, where the instances are really pieces of a bigger function).

PR #1534, opened 2016-10-08
Many improvements to QScript joins.

sellout:
- complete implementation for Mergable\[*Join\] (avoids autojoins after a join)
- fix a bug where we included the provenance in the join condition
- fix join combine func to use the result pointers

Also,
- combine `Take`/`Drop` into `Subset` (and add `Sample` case)
- fix handling of `ProjectField` in Mongo-QScript planner
- handle `ShiftedRead`’s `IncludeId` in Mongo-QScript planner
- fix `Reduce`’s repairFunc handling in Mongo-QScript planner
- normalize Mongo-specific QScript

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
@@ -891,7 +891,7 @@ object MongoDbQScriptPlanner {
       case RightSide => a2
     } ∘ (JsFn(JsFn.defaultName, _))
 
-  def meh[M[_]: Functor: Plus, A](a: A)(exf: A => M[Fix[ExprOp]], jsf: A => M[JsFn])
+  def exprOrJs[M[_]: Functor: Plus, A](a: A)(exf: A => M[Fix[ExprOp]], jsf: A => M[JsFn])
       : M[Expr] =
     exf(a).map(_.right[JsFn]) <+> jsf(a).map(_.left[Fix[ExprOp]])
 
@@ -899,13 +899,13 @@ object MongoDbQScriptPlanner {
     (funcHandler: FuncHandler[T, EX], fm: FreeMap[T])
     (implicit ev: EX :<: ExprOp)
       : OutputM[Expr] =
-    meh(fm)(getExpr(funcHandler)(_), getJsFn[T])
+    exprOrJs(fm)(getExpr(funcHandler)(_), getJsFn[T])
 
   def handleRedRepair[T[_[_]]: Recursive: ShowT, EX[_]: Traverse]
     (funcHandler: FuncHandler[T, EX], jr: FreeMapA[T, ReduceIndex])
     (implicit ev: EX :<: ExprOp)
       : OutputM[Expr] =
-    meh(jr)(getExprRed(funcHandler)(_), getJsRed[T])
+    exprOrJs(jr)(getExprRed(funcHandler)(_), getJsRed[T])
 
   def getExprRed[T[_[_]]: Recursive: ShowT, EX[_]: Traverse]
     (funcHandler: FuncHandler[T, EX])
--- a/mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
@@ -26,12 +26,12 @@ import quasar.fs.{FileSystemError, QueryFile}
 import quasar.javascript._
 import quasar.jscore, jscore.{JsCore, JsFn}
 import quasar.namegen._
-import quasar.physical.mongodb.WorkflowBuilder._
+import quasar.physical.mongodb.WorkflowBuilder.{Subset => _, _}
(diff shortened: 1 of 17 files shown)


mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -876,9 +891,32 @@ object MongoDbQScriptPlanner {
       case RightSide => a2
     } ∘ (JsFn(JsFn.defaultName, _))
 
+  def meh[M[_]: Functor: Plus, A](a: A)(exf: A => M[Fix[ExprOp]], jsf: A => M[JsFn])
```
wemrysi: That name, though ;)

sellout: Ahahaha, yeah … I do have it changed locally. It’s my default “I don’t yet know what this function does” name.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -1001,8 +1037,14 @@ object MongoDbQScriptPlanner {
       // NB: right now this only outputs one phase, but it’d be cool if we could
       //     interleave phase building in the composed recursion scheme
       opt <- log("QScript (Mongo-specific)")(
-        shiftRead(qs).transCata[MongoQScript](
-          SimplifyJoin[T, QScriptShiftRead[T, ?], MongoQScript].simplifyJoin(idPrism.reverseGet)).point[M])
+        shiftRead(qs)
+          .transCata[MongoQScript](SimplifyJoin[T, QScriptShiftRead[T, ?], MongoQScript].simplifyJoin(idPrism.reverseGet))
+          .transAna(
```
wemrysi: Should this coalescing be done by anything using `SimplifyJoin`? If so, would it be possible to encapsulate this into a function to be used in place of `SimplifyJoin` directly?

wemrysi:
Regarding the `Subset` term, what are your thoughts on something like `Slice[T[_[_]], A](src: A, from: FreeQS[T], range: FreeQS[T] \&/ FreeQS[T])` and then a standalone `Sample[T[_[_]], A](src: A, from: FreeQS[T], count: FreeQS[T])` where the two structures in `range` represent "start" and "end"?

MarkLogic supports selecting a "slice" of a dataset directly, but I'm not sure how common this will be. The `Slice` term would allow for expressing that and other backends that don't support this could translate to an "offset -> limit" internally or whatever.

Conversely, the MarkLogic planner could also just implement a transformation that looks for a `Subset(_, _, Take, _)` followed by a `Subset(_, _, Drop, _)` (or vice versa) and construct the appropriate range. I don't really have a notion of which is preferable, just curious if you'd considered it.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
sellout: I’m trying to think of how to do this well.

There are some constraints –
1. you should only have to normalize then optimize once, after you’ve done any of your transformations.
2. the set of normalizations that can be applied depends on your coproduct.

So, while it’d be nice to normalize as we go, say, composed with `simplifyJoin`, the problem is that we don’t know which coproduct you have. We could get away with applying `coalesceEJ` alongside `simplifyJoin`, but you’d still need to explicitly call the others and if `shiftRead` (for example) were called afterward, you’d have to reapply it anyway.

I mentioned to @mossprescott that this is a discussion we need to have – how to handle recursive coproducts.

sellout:
Yeah, I think Slice/Sample sounds good.

sellout:
Ok, so I think Slice might be the right way forward, but not in this PR. If you’d like, I can revert the Subset bits (but I do think they clean some things up a little).

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
wemrysi: I see, ok, this is fine then.

An (unfiltered) thought off the top of my head might be to see if we can come up with a TC with instances for the common subset of the cross product of the various coproducts (and then maybe a low-priority catch-all "identity" optimization). It will be a bunch of boiler-plate in one spot, but then it'd only be in one spot and the compiler would resolve which coproduct you have and apply the appropriate normalizations.

Anyhow, it isn't going to be solved here, regardless so :+1:

wemrysi:
Leaving `Subset` is fine, LGTM otherwise modulo the build. I'll merge when green.

alissapajer:
@sellout would it be cool if we start using the QScript tag in our PRs? Then it's easy to scan the list of open PRs to see which ones are QScript related.

sellout:
Yep, sorry, just forget most of the time. Also, working on merging now …

alissapajer:
@sellout if you haven't merged in yet, my PR #1538 is green now, will likely cause merge conflicts, so you can consider merging that in to master first, then into this PR.

alissapajer:
@sellout it's merged, you'll need to fetch again :)

PR #1536, opened 2016-10-10
Spark onejar & spark runtime issues

rabbitonweb:


rabbitonweb:
This PR is needed to spark-hdfs work on real Spark cluster. 

A bit of theory behind it is following: When our calculations will be serialized and send over to Spark, Spark will deserialize them and (for the whole process to work) it needs all
classes (including lambdas) to be on the classpath.
SparkContext comes with a handy method .addJar which allows us to send over to Spark our classes in a single jar. There is however a restriction on that jar - it can not contain any dependencies to org.apache.spark itself.

This PR addresses the above. However it will require the way we build & ship Quasar

1. we need to build a special version sparkcore.jar by calling `sbt sparkcore/assembly -DbuildSparkCore=yes`
2. our "executable" can not longer be just a jar holding application, it also needs the `sparkcore.jar` in it and while installing user will have to set a env. variable `QUASAR_HOME`

// CC @jdegoes

jdegoes:
This requirements are OK for Friday, but we need to remove (2) as a requirement (it should not be necessary) and automate (1).

wemrysi:
@rabbitonweb So should the build process be updated in this PR then? Sounds like some documentation in the README is also warranted to explain what must be done in order to connect to spark.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -79,13 +79,19 @@ package object hdfs {
 
   final case class SparkHdfsFSDef[S[_]](run: Free[Eff, ?] ~> Free[S, ?], close: Free[S, Unit])
 
+  private def fetchSparCoreJar: Task[String] = Task.delay {
```
wemrysi: Minor typo? `fetchSparCoreJar` -> `fetchSparkCoreJar`

mossprescott:
@rabbitonweb What's the status with this?

Seems like the build is failing just because of a missing import.

Are we still planning to build a separate installer for Spark for this week? I think we should be able to embed sparkcore.jar inside the "big" jar, but I'm not sure exactly what will be involved so putting that off makes sense.

rabbitonweb:
There is more to it then fixing that import. PR is in progress. I hope to get this done today. Will keep u posted :)

rabbitonweb:
In this PR we have to do two things:
1. give ability to build quasar for spark (this means excluding netty from mongo connector
2. give ability to build sparkcore.jar - this is a small jar that holds classes used in the spark calculations (mostly lambdas)

For 1. we will use `quasar4Spark` system property
For 2. we will use `buildSparkCore` system property

In other words:
1. to assembly quasar for mongo and other connectors (as before) we run `web/assembly`
2. to assembly quasar for spark we take few steps:
a) run `sbt web/assembly -Dquasar4Spark=yes`
b) run `sbt web/assembly -DbuildSparkCore=yes`
c) installation means putting both JARs in a single folder and pointing `QUASAR_HOME` to that folder

rabbitonweb:
This PR allows us to build Quasar for both Spark & other connectors. However it uses system properties to do it (see previous comments).

@paulp said 

```
@rabbit there's no reason to try to incur complicated behavior changes with system properties. In sbt you have the entirety of scala at your command to generate build definitions.

[1:28]  
if there are two similar projects, make a method which abstracts the difference and create two projects.

[1:30]  
In my own build.sbt I have stuff like
```lazy val ygg      = project |> setup |> Ygg.ygg
lazy val jsonfile = project |> setup |> Ygg.jsonfile
```
Where the things beyond project are `Project => Project` methods which configure the project.

[1:31]  
System properties should be a last resort, and are really only suitable for situations where unknown persons might need to change something in unknown ways. Here you are dealing with known quantities.
````

So I understand that there is a better way to do it. Given time constraints & fact I don't have now a clue how to do it other way I ask either for

a) review current solution
b) help (best pair programming session) to do it without system properties

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
mossprescott: @rabbitonweb wouldn't hurt to address this.

rabbitonweb:
I've added https://github.com/quasar-analytics/quasar/issues/1589 and I'm voting to push current way of building quasar as it is. I will work on https://github.com/quasar-analytics/quasar/issues/1589 as soon as possible

rabbitonweb:
About documentation I've added separate issue https://github.com/quasar-analytics/quasar/issues/1590

PR #1540, opened 2016-10-10
Filter noise from travis output.

paulp:


paulp:
I really wanted to achieve [per-spec folds](https://twitter.com/extempore2/status/784888171993542656)  but I couldn't get it to work consistently. Along the way I was reminded of how much noise there is in the output, so that's what this attacks.

paulp:
[Here is a sample log](https://travis-ci.org/paulp/quasar/jobs/166511952).

paulp:
[Here is a better one](https://travis-ci.org/quasar-analytics/quasar/jobs/166548694). It's 1771 lines.

[Same job in master](https://travis-ci.org/quasar-analytics/quasar/jobs/166549733) is 3715 lines.

I do not think there is anything of value in those 2000 filtered lines. (But on the off chance there is, it's also logging all output to a file.)

alissapajer:
@paulp in the two examples you gave above, the old one had colored output (little green `+` symbols, etc) while the new one has no color. Is the color something you removed and could put back?

alissapajer:
Also, is it possible to access the file that you are writing _all_ the output to in jenkins?

paulp:
There's color! I did disable the normal colors because they complicate the string matching. I could restore them but since I like it this way better I don't want to do the color work until I know what the consensus is. When there are errors the error lines are in red.

As I understand things there's a way to retrieve individual build files but I haven't tracked down that mechanism yet.

screen shot 2016-10-12 at 8 15 26 am

alissapajer:
Re the colors:  fine by me. I personally prefer having the little `+` symbols also colored, but I'm happy to merge it with the current colorings.

alissapajer:
@mossprescott thoughts? I'm happy to merge this, but I do think we need to track down how to access the full build log files, just in case there is some strange build problem that is only printed there.

mossprescott:
I understand the urge to get this output under control, but I'm not sure this approach will save us time and energy in the long run.

This type of filtering inevitably creates doubt about what's being lost. Since there's no way to access files after the fact, we'll end up pushing test branches to access the full log, which is no way to live.

I think it's great that the log isn't literally half "Resolving ...", but you can't really scan a 1700 line log either. I for one will still end up doing find-in-page-"[err" to figure out what went wrong. So I don't think this fundamentally fixes the problem.

Finally, this is a fairly complex script with a long list of search patterns that will need to be maintained over time and that just seems like a distraction to me. I think that effort would be better put into improving SBT. For example, just reducing the "Resolving" messages to debug level would solve 90% of the problem.

bin/sbt-travis.sh
```diff
@@ -0,0 +1,72 @@
...
+Attempting
+Formatting
+Reapplying
+downloading
+\[SUCCESSFUL\s*\]
+Passed: Total 0,
+0 example,
+ScalaCheck
+not simplified
+not implemented in aggregation
```
mossprescott: These messages are output from our own tests. If we think they're too noisy, we could do something about that at the source. Actually, I think the problem is just the way specs2/SBT logs them on their own lines.

bin/sbt-travis.sh
```diff
@@ -0,0 +1,72 @@
...
+Server started listening
+Stopped last server
+TODO
+log4j:
```
mossprescott: I believe our customers see this output, so I think we should fix it by configuring log4j or whatever it is that we're not doing.

paulp:
Maybe I'll come back to it.

PR #1543, opened 2016-10-10
Enable integration tests for MarkLogic

wemrysi:
Adds an installer script for MarkLogic and adds it to the travis build matrix. Only the filesystem integration tests are enabled atm, all regression tests are skipped.

Also fixes a couple issues discovered thanks to the improved `QScript` generation.

### DIFF
--- a/marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
+++ b/marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
@@ -323,10 +323,10 @@ object qscript {
         val n = "$name"
 
         for {
-          kelt    <- ejson.mkArrayElt[F] apply n.xqy
-          velt    <- ejson.mkArrayElt[F] apply c.xqy
+          kelt    <- ejson.mkArrayElt[F](n.xqy)
+          velt    <- ejson.mkArrayElt[F](c.xqy)
           kvArr   <- ejson.mkArray_[F](mkSeq_(kelt, velt))
-          kvEnt   <- ejson.mkObjectEntry[F] apply (n.xqy, kvArr)
+          kvEnt   <- ejson.renameOrWrap[F] apply (n.xqy, kvArr)
           entries =  for_ (c -> elt `/` child.element())
                      .let_(n -> fn.nodeName(c.xqy))
                      .return_(kvEnt)
--- a/marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
+++ b/marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
@@ -158,7 +158,7 @@ object qscript {
       }
     }
 
-  // qscript:inc-avg-state($cntavg as map:map, $x as item()*) as map:map
+  // qscript:inc-avg-state($cnt as xs:integer, $avg as xs:double) as map:map
   def incAvgState[F[_]: PrologW]: F[FunctionDecl2] =
     qs.name("inc-avg-state").qn[F] map { fname =>
       declare(fname)(
--- a/marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
+++ b/marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
@@ -136,8 +136,8 @@ object qscript {
         val (c, a, y) = ("$c", "$a", "$y")
         incAvgState[F].apply(c.xqy, y.xqy) map { nextSt =>
           let_(
-            c -> (map.get(st, "cnt".xqy) + 1.xqy),
-            a -> map.get(st, "avg".xqy),
+            c -> (map.get(st, "cnt".xs) + 1.xqy),
+            a -> map.get(st, "avg".xs),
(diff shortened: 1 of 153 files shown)


wemrysi:
@mossprescott Looks like it worked, back in your court.

scripts/installMarkLogic
```diff
@@ -0,0 +1,50 @@
...
+ML_DOCKER_DIR="scripts/marklogic"
+ML_IMG_NAME="marklogic:${ML_VERSION}"
+ML_CNT_NAME="ml8"
+
+ML_ADMIN_PREFIX="http://localhost:8001/admin/v1"
+ML_USERNAME="marklogic"
+ML_PASSWORD="marklogic"
+
+# Login to developer.marklogic.com
+curl -X POST --data "email=${ML_DOWNLOAD_EMAIL}&password=${ML_DOWNLOAD_PASSWORD}" -c $ML_DEV_COOKIES $ML_DEV_LOGIN
```
mossprescott: Before we unleash this in master, I think it would be a good idea to add the binary to Travis's cache.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -120,6 +120,20 @@ object qscript {
       }
     }
 
+  // qscript:element-left-shift($elt as element()) as item()*
```
mossprescott: What are these comments meant to convey? Is this how you call it? What it expands to?

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -120,6 +120,20 @@ object qscript {
...
+      declare(fname)(
+        $("elt") as SequenceType("element()")
+      ).as(SequenceType("item()*")) { elt: XQuery =>
+        isArr(elt) map { eltIsArray =>
+          if_ (eltIsArray)
+          .then_ { elt `/` child(aelt) `/` child.node() }
+          .else_ { elt `/` child.node() }
+        }
+      }
+    }.join
```
mossprescott: As far as I can tell, there's no unit test coverage for these functions (and I guess inspecting the XQuery syntax would be painful), and there's no integration test to show that they actually perform as expected in ML. Is there a plan to get something like that going?

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -149,7 +149,7 @@ object qscript {
     qs.name("inc-avg-state").qn[F] map { fname =>
       declare(fname)(
         $("cnt") as SequenceType("xs:integer"),
-        $("avg") as SequenceType("xs:decimal")
+        $("avg") as SequenceType("xs:double")
```
mossprescott: I think the comment on this fn needs some attention.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
mossprescott: And how is that useful? If it's useful to a Quasar/Scala programmer, then I'd say it should be a scaladoc comment, but in that case I think it would also be more useful to write it in terms of the Quasar/Scala types.

It also seem like something that could easily be generated from the declaration (indeed, I suppose it is, when the Prolog is serialized. Not sure why it's also being written by hand for each one.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
wemrysi: See #1570

mossprescott:
Really happy to be getting this into the build 😀 

I'm obviously just getting acquainted with this codebase, and of course at this point there are lots of areas for future attention. You can take my questions about testing as suggestions for that.

So, just the one comment that seems wrong and the caching thing and I'll merge.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
mossprescott: Is any regression test already enabled, or do you mean in a subsequent PR? It's hard to tell in the diff.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
wemrysi: It is useful when applying them while constructing plans as one would like to think in terms of XQuery and not xquery-embedded-in-scala. If the scala declaration syntax can be improved to be closer to that of the xquery, then they can probably go away, as-is it is still a bit verbose and if all you care about is what this ends up looking like in XQuery, its a lot of noise to see through.

No argument that it is duplication and suffers from all the drawbacks thereof, examples of which you have already identified.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
wemrysi: No, that is for an imminent PR as I also need to implement `ExecutePlan` in order to enable a regression test.

PR #1565, opened 2016-10-11
no more tuples in qscript Transform

alissapajer:


connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -115,7 +114,7 @@ class Transform
   /** Contains a common src, the MapFuncs required to access the left and right
     * sides, and the FreeQS that were unmergeable on either side.
     */
-  private type MergeResult = (T[F], FreeQS, FreeQS)
+  private case class MergeResult(src: T[F], lval: FreeQS, rval: FreeQS)
```
alissapajer: This is now a case class.

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -134,25 +133,27 @@ class Transform
...
         .get(rTail.reverse.ana[T, CoEnv[Hole, F, ?]](delinearizeTargets[F, ExternallyManaged] >>> (CoEnv(_)))).mapSuspension(FI.inject)
 
-    (common.reverse.ana[T, F](delinearizeInner),
+    MergeResult(common.reverse.ana[T, F](delinearizeInner),
       rebaseBranch(leftF, lMap),
       rebaseBranch(rightF, rMap))
   }
 
+  private case class AutoJoinResult(src: T[F], buckets: List[FreeMap], lval: FreeMap, rval: FreeMap)
+  private case class AutoJoin3Result(src: T[F], buckets: List[FreeMap], lval: FreeMap, cval: FreeMap, rval: FreeMap)
```
alissapajer: We now have these return types for `autojoin`. I don't like how much they share but wasn't convinced that a second case class for `src` and `buckets` would clean this up... maybe an abstract class would be better?

paulp: I would do `private case class AutoJoin3Result(cval: FreeMap, rest: AutoJoinResult)`.

paulp: (Possibly also naming `AutoJoinResult` something which isn't misleading if used in both ways.)

connector/src/main/scala/quasar/qscript/package.scala
```diff
@@ -99,6 +99,16 @@ package object qscript {
       Show.show(ann => Cord("Ann(") ++ ann.provenance.show ++ Cord(", ") ++ ann.values.show ++ Cord(")"))
   }
 
+  @Lenses final case class Target[T[_[_]], F[_]](ann: Ann[T], value: T[F])
```
paulp: It's highly desirable not to define classes inside package objects. Pretty much only implicit classes should be defined that way (since you have no choice.)

alissapajer: In this case, when we want a class accessible to all of qscript, and to a backend as well, where would be a good place to put it?

alissapajer: Though, it does just need to be accessed in `Transform` and then `Transform` exposes `lpToQScript` which returns a `Target`.

paulp: You can put it in the qscript package. Just not the qscript package object. You can even do it in the same file if there's a motivation to do so, like:
```scala
package object qscript {
  ... blah 
}
package qscript {
  @Lenses final case class Ann[T[_[_]]](provenance: List[FreeMap[T]], values: FreeMap[T])

  object Ann {
    implicit def equal[T[_[_]]: EqualT]: Equal[Ann[T]] =
      Equal.equal((a, b) => a.provenance ≟ b.provenance && a.values ≟ b.values)

    implicit def show[T[_[_]]: ShowT]: Show[Ann[T]] =
      Show.show(ann => Cord("Ann(") ++ ann.provenance.show ++ Cord(", ") ++ ann.values.show ++ Cord(")"))
  }

  @Lenses final case class Target[T[_[_]], F[_]](ann: Ann[T], value: T[F])

  object Target {
    implicit def equal[T[_[_]]: EqualT, F[_]: EqualF]: Equal[Target[T, F]] =
      Equal.equal((a, b) => a.ann ≟ b.ann && a.value ≟ b.value)

    implicit def show[T[_[_]]: ShowT, F[_]: ShowF]: Show[Target[T, F]] =
      Show.show(target => Cord("Target(") ++ target.ann.show ++ Cord(", ") ++ target.value.show ++ Cord(")"))
  }
}
```

connector/src/main/scala/quasar/qscript/Transform.scala
alissapajer: I decided on having `private case class AutoJoinBase(src: T[F], buckets: List[FreeMap])`. 

If we had `private case class AutoJoin3Result(cval: FreeMap, rest: AutoJoinResult)` that wouldn't represent what `autojoin3` really does, since the third value is joined in on the right, meaning we'd really have a second `rval: FreeMap`, which I think would be more confusing.

alissapajer:
@paulp changes pushed in response to review.

paulp:
LGTM after travis approves.

PR #1566, opened 2016-10-11
Remove dead code.

paulp:
If experience is any guide there will be some resistance
to removing one or more of these methods. I propose that
all such resistance is misguided, because:
- none of it was important enough to use
- none of it was important enough to test

Ergo, none of it is important enough to weigh down master.


mossprescott:
Some of this stuff might have been important enough to use downstream, so it would be worth doing a `publishLocal` and building Q-A to make sure we're not losing something we need. If so, I completely agree that there should be a test in quasar.

sellout:
And two other bits – I’d like to make sure none of the things in qscript _should_ be used before this gets merged, and the stuff in `contrib` is often instances and stuff that should be pushed to (and tested in) an upstream library.

paulp:
> I’d like to make sure none of the things in qscript should be used before this gets merged

Indeed part of the purpose of a PR such as this is to force that analysis.

paulp:
And yes I should have stated clearly up front I'm in no rush to merge this (unlike most of my other PRs), but corrective action should be taken with respect to anything we wish to keep.

jedesah:
This all looks pretty reasonable to me although I agree with @mossprescott's comment.

foundation/src/main/scala/quasar/contrib/matryoshka/package.scala
```diff
@@ -33,11 +33,6 @@ package object matryoshka extends CoEnvInstances {
       Show[T[F]] =
     T.showT[F](F)
 
-  def elgotM[M[_]: Monad, F[_]: Traverse, A, B](a: A)(φ: F[B] => M[B], ψ: A => M[B \/ F[A]]):
```
alissapajer: I think that everything in a contrib dir should stay there so it can eventually be moved to its new home. At that point, when the contrib file is wiped clean, decisions should be made whether or not a function is worth moving to the library or should just  be deleted. The decision _not_ to move something to a library is a different concern than if something is used in quasar.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
```diff
@@ -26,27 +26,12 @@ import scalaz.std.iterable._
 
 @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
 object cts {
-  def andQuery(query: XQuery, queryOrOptions: XQuery*): XQuery =
```
alissapajer: Since the marklogic stuff was added so recently, we should confirm with @wemrysi that these functions won't be used in the future. Or maybe they are here for completeness, since we may be able to make the xquery stuff its own library at some point. In that case, maybe a contrib dir of some sort would be appropriate?

alissapajer: And same for everything in the marklogic package.

foundation/src/main/scala/quasar/contrib/matryoshka/package.scala
alissapajer: Here's an idea - what do you folks think of this?

We have two contribs per library: `contrib` and `quasar-contrib`. In `quasar-contrib` lives everything that we _actually_ call from quasar. In `contrib` lives other things that at one point were probably called from quasar, but aren't anymore. And then when you import from a contrib, you only import from the `quasar-contrib`.

Also, that way when we're porting the contribs over to their future library homes, we know that we don't want to delete any functions in the `quasar-contrib` because we're using them.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
paulp: "It might be used in the future" is said of all dead code. If it isn't used and it isn't tested then it doesn't belong in the repository.

foundation/src/main/scala/quasar/contrib/matryoshka/package.scala
paulp: The way you know you're using a function is that there's a test which exercises it. If it isn't important enough to test, it isn't important enough to keep.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
alissapajer: A library doesn't call the majority of its own functions (but it _does_ test them). It exposes them so other projects can call them. That's the point I was making here. These functions may be relevant if we aim to create an xquery library. But with that said, I do agree that a test should be written for any such function, and that test could be written in lieu of deleting the function.

Though, then you have the problem of dead code which is only called in tests. We might not need those functions around, but we have to draw the line somewhere, so I think drawing it at a test is a reasonable place, because it will at least allow us to delete a bunch of truly (even less used) code.

alissapajer:
@paulp I'd like to get a version of this merged. Maybe after #1780 gets merged?

paulp:
@alissapajer sure - I'm sure it needs a total reboot, but I'll do it if there is consensus that whatever's still dead can go.

PR #1572, opened 2016-10-12
Planner tests

jedesah:


sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -466,24 +430,11 @@ class PlannerSpec extends quasar.Qspec with QScriptHelpers with DisjunctionMatch
...
-    }
-    masterAndRoot <- uri match {
-      case Data.Str(s) => s.point[OptionT[Task, ?]]
-      case _ => OptionT.none[Task, String]
-    }
-  } yield {
-    val master = masterAndRoot.split('|')(0)
-    val config = new SparkConf().setMaster(master).setAppName(this.getClass().getName())
-    new SparkContext(config)
+  private def withSparkContext[A](f: SparkContext => Task[A]): Task[A] = {
```
rabbitonweb: You've removed functionalty to ONLY run Spark Planner tests when QUASAR_SPARK_LOCAL is set.

This is crucial because we can not run spark tests on local-mode ("local[*]") because of lack of ability to clean up resources after tests.

rabbitonweb:
I do apologies this review took so long.

jedesah:
@rabbitonweb I am not married to some of the cosmetic changes in case you really don't like them

rabbitonweb:
All looks good. I've seen a lot of code disappear :)

LGTM once CI is green

jedesah:
Haha, looks like the Spark tests are broken again. Another non-serializable thing found it's way into a closure. I'll have a look at what it might be.

Hopefully, we can finally get this thing merged so it doesn't happen again!

jedesah:
My previous comment I was laughing because the Spark tests were broken again by some apparently inconspicuous change. I am not laughing anymore though since it's proving much harder this time than the previous time to figure out what is causing the failure to serialize.

Which is why I am marking the failing tests as `pendingUntilFixed` and going to investigate later so that we can get the rest of the tests running on Travis as quickly as possible. It's much easier to diagnose what happened when you know the change that broke serialization rather than try and figure out after the fact.

Right now my hunch is that it's the upgrade to Matryoshka that might have broke serialization, simply because Spark is complaining about `Fix` not being `Serializable`. I doubt `Fix` was ever `Serializable` so it's probably more subtle than that, but that's what I have got to work on right now. I have pretty much eliminated the hypothesis of a local modification being the culprit. @sellout Any idea what change in the latest version of Matryoshka might have triggered this issue?

sellout:
Fix should be Serializable, all case classes automatically are, IIRC.

Matryoshka has definitely changed a lot – @rabbitonweb has mentioned wanting to ale parts of Matryoshka Serializable – namely the RecursiveT type class. Maybe that'll help?

jedesah:
@rabbitonweb I added a few extra changes, but this should be ready to merge whenever you get around to having one last look at it.

jedesah:
Looks like the shadowing has caused an error to be produced when `sbt assembly` is called. Putting back to "in progress" and I'll have a look at what's happening.

jedesah:
@rabbitonweb This is ready finally! Can you have a final look at it today?

rabbitonweb:
@jedesah did you have a chance to it on real cluster (not local-mode) ?

jedesah:
No, I didn't do that.

rabbitonweb:
Code LGTM, tomorrow (on a plane) I will check if all works on real cluster (I don't  trust local-mode with ability to check for dependency hell  - all runs on a single JVM). If all works I will merge to master.

jedesah:
Cool! If it doesn't work on the real cluster, let me know what the problem is.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -553,43 +549,19 @@ class PlannerSpec
...
-    newSc.flatMap {
-      case Some(sc) =>
-        run(sc).map(_.toResult)
-            .onFinish(κ(Task.delay {
-              sc.stop()
-            }))
-      case None => Task.now(skipped("skipped because QUASAR_SPARK_LOCAL is not set"))
-    }.unsafePerformSync
+  private def withSpark[A](f: SparkContext => Task[A]): A = {
+    val config = new SparkConf().setMaster("local[*]").setAppName("PlannerSpec")
```
sellout: @rabbitonweb All “pure” tests of sparkcore code (i.e., not requiring a running Spark cluster), should be run any time `sparkcore/test` is executed. We _do_ want to run these regardless of the environment variables. Ideally, we would run all tests every time, but that currently requires users having set up specific stuff on their machines, so we let them indicate which things they have set up. This is especially important for contributors, where we want people (including ourselves) to get as much confidence from `sbt test` as possible before submitting a PR.

As far as CI goes, we can do all sorts of fine-grained things if we need to without having to build it into specific test suites. It is a problem that Travis wants to build & test everything on every job, but that is a Travis problem, not a test suite problem. Better build systems (like GitLab) let us build once and run all of the pure tests, then run only impure connector-specific tests in forked jobs.

rabbitonweb: @sellout correct but PlannerSpec is NOT pure unit tests - it requires environment (even if it is emulated Spark). I will not argue over it, but I'm seeing inconsistency here when compared to other connectors.

rabbitonweb:
@jedesah `revision.md` should be deleted right?

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
sellout: @rabbitonweb Ah, ok. Well, we _should_ be able to test the bulk of the planner without any environment, right? Although, I have a feeling that you’re going to say “RDDs are hard to test”, which will make me sad.

sellout:
@rabbitonweb If this gets merged before 4.1 is cut, then it can still be merged with the old process. But we don’t want to merge anything post-4.1 with a CHANGELOG file. And no new PRs should be created with a CHANGELOG file.

rabbitonweb:
@sellout it's ready to be merged, just need to delete `revison.md` & it's ready

sellout:
@rabbitonweb It _doesn’t_ need revision.md removed if it gets in 4.1. If you have the latest version of devtools, just use `git checkout HEAD~3; bin/merge ...; git checkout -` and that’ll merge with the old version of the script, which we won’t need after today.

rabbitonweb:
@jedesah @sellout ok, i will, the only change now required is to remove `pendingUntilFixed` in `PlannerSpec` since those serialization issues are fixed in master.

To sum up this conversation:
1. git merge master
2. rm `revision.md`
3. `PlannerSpec` - remove .pendingUntilFixed

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
jedesah: I would say that an emulated Spark is a "pure" environment. It doesn't require anything external in order to run. If any of our other backends had in-memory emulators, I would consider testing against those to be pure and could be done in their respective projects. I see `it` as a way of separating tests that require some external setup and may spuriously fail depending on the nature of the configured instances.

Issue #1568, opened 2016-10-12
Use of 'union' results in MatchError in Transforms


wemrysi:
The use of `union` in SQL² triggers the following match error during `QScript` generation, the example query in `it/src/main/resources/tests/union.test` triggers it as does the query `select foo from /example/data union select bar from /example/data`.

```
[error]     ! union [./union.test]
[error]      scala.MatchError: Distinct((Ann(List(),Return(SrcHole)),Fix(Coproduct(-\/(Union(Fix(Coproduct(-\/(LeftShift(Fix(Coproduct(\/-(Coproduct(-\/(BucketField(Fix(Coproduct(\/-(Coproduct(-\/(BucketField(Fix(Coproduct(\/-(Coproduct(-\
/(BucketField(Fix(Coproduct(\/-(Coproduct(-\/(BucketField(Fix(Coproduct(\/-(Coproduct(\/-(Coproduct(\/-(Const(Root)))))))),Gosub(Gosub(Return(SrcHole),<function1>),<function1>),Gosub(Gosub(Gosub(Suspend(Constant(Fix(Coproduct(\/-(Str(quasa
r-test)))))),<function1>),<function1>),<function1>))))))),Gosub(Gosub(Return(SrcHole),<function1>),<function1>),Gosub(Gosub(Gosub(Suspend(Constant(Fix(Coproduct(\/-(Str(run_cbe6793c)))))),<function1>),<function1>),<function1>))))))),Gosub(
Gosub(Return(SrcHole),<function1>),<function1>),Gosub(Gosub(Gosub(Suspend(Constant(Fix(Coproduct(\/-(Str(regression)))))),<function1>),<function1>),<function1>))))))),Gosub(Gosub(Return(SrcHole),<function1>),<function1>),Gosub(Gosub(Gosub(
Suspend(Constant(Fix(Coproduct(\/-(Str(zips)))))),<function1>),<function1>),<function1>))))))),Gosub(Suspend(ZipMapKeys(Return(SrcHole))),<function1>),Gosub(Suspend(ConcatArrays(Gosub(Suspend(MakeArray(Gosub(Suspend(ConcatArrays(Gosub(Susp
[...]

PR #1574, opened 2016-10-12
A collection of QScript fixes.

sellout:
This is a bunch of small fixes to QScript motivated by one of two things
1. the included changes to the Mongo-QScript connector or
2. a simplification in the way we handle Constants in qscript.Transform.

The latter exposed a number of places we weren’t properly dealing with
the value pointer in the annotation. Basically, that change broke a
number of tests (when it should have been semantically equivalent), and
the subsequent fixes were needed to make them pass again. In general, I
also took advantage of the value pointer more often, to simplify some
Transform cases and avoid creating intermediate Map nodes.

In the Mongo-QScript connector, we now identify arrays and maps with
statically-known indices/keys to take advantage of ArrayBuilder and
DocBuilder.

### DIFF
--- a/connector/src/main/scala/quasar/fs/QueryFile.scala
+++ b/connector/src/main/scala/quasar/fs/QueryFile.scala
@@ -64,7 +64,7 @@ object QueryFile {
     //       so we don’t duplicate work.
     lp.transCata[LogicalPlan](orOriginal(Optimizer.elideLets[T]))
       .cataM[PlannerError \/ ?, (Ann[T], T[QS])](newLP => transform.lpToQScript(newLP.map(_ ∘ (_.transCata(eval)))))
-      .map(qs => QC.inj((transform.reifyResult(qs._1, qs._2))).embed.transCata(eval))
+      .map(qs => QC.inj((transform.reifyResult(qs._1, qs._2))).embed.transAna(eval))
   }
 
   def simplifyAndNormalize
@@ -85,8 +85,8 @@ object QueryFile {
     _.transAna(SP.simplifyProjection)
       // TODO: Rather than explicitly applying multiple times, we should apply
       //       repeatedly until unchanged.
-      .transCata(rewrite.normalize)
-      .transCata(rewrite.normalize)
+      .transAna(rewrite.normalize)
+      .transAna(rewrite.normalize)
   }
 
   /** The shape of QScript that’s used during conversion from LP. */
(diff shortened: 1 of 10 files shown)


sellout:
I know this already needs a merge from master, but I’m going to try to get a few other PRs in the queue merged (and maybe deal with PR feedback) before updating it.

alissapajer:
@sellout should any pending tests be enabled in this PR? (Some of the "expected" results might be a touch off, causing them to still fail, meaning the `pendingUntilFixed` wouldn't alert us.)

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -85,8 +85,8 @@ object QueryFile {
     _.transAna(SP.simplifyProjection)
       // TODO: Rather than explicitly applying multiple times, we should apply
       //       repeatedly until unchanged.
-      .transCata(rewrite.normalize)
-      .transCata(rewrite.normalize)
+      .transAna(rewrite.normalize)
```
alissapajer: So, this changes the recursion from bottom up to top down. Why? If this change was worth making, can you leave a comment in the code about this?

sellout:
Yep, I just enabled one of the pending tests.

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: So, in this case, I just think it will converge more quickly. For example, if you tried to do constant-folding top-down, you’d only be able to fold one level per pass, whereas applied bottom-up, you will always succeed in one pass. I think I can revert these back to `transCata` in this PR, if you’d prefer.

I do think we want to eventually expose just a fold (rather than the algeabras), but the problem is in the coproducts. There is a discussion to be had (post-release) on how we can best apply the right combination of operations given an arbitrary coproduct. We have one approach in QScript, @mossprescott has another (or two) in the Mongo connector. And I hope we can find something that works everywhere (and better than any of the current approaches).

Related: backends (other than MarkLogic) can’t call `rewrite.normalize` anyway, because it requires ThetaJoins in the coproduct.

alissapajer:
@sellout regarding the pending test, there should be an issue for that, so make sure to find it and close it here! :)

alissapajer:
@sellout ok LGTM, feel free to merge in master.

sellout:
Ok, master is merged in. Also, notice there’s a new commit to review – it adds a fix to #1568.

alissapajer:
Fixes #1568. 

(According to [the docs](https://help.github.com/articles/closing-issues-via-commit-messages/) I think we need a keyword next to the issue number for it to be closed automatically.)

Issue #1539, opened 2016-10-10
self-joins get elided


sellout:
Joining a table against itself gets treated as an auto-join and elided, when it shouldn’t be.

alissapajer:
@sellout can you confirm this example triggers the bug you have in mind?
```
select a1.foo, a1.bar from ack a1, ack a2 where a1.xyz = a2.xyz
```

alissapajer:
A slightly better example, and one that compiles in SQL^2 (the `as` for the table alias is required).

```
select z1.foo, z2.bar from zips as z1, zips as z2 where z1.xyz = z2.pqr
```

alissapajer:
Note to self: this query should be an integration test if it's not already.

alissapajer:
This requires a first-class representation of joins in `LogicalPlan`. That way we can reference the left and right sides separately, even though they may be the same dataset.

alissapajer:
Pertinent self-join information is currently being held in let-bindings in the logical plan, which we elide in QScript.

alissapajer:
Fixable after #1943 is fixed.

alissapajer:
Verified that with #1943 fix "selfjoinWithComplexCondition.test" and "selfjoinConstantOnLeft.test" both pass in Spark.

PR #1576, opened 2016-10-13
MarkLogic ExecutePlan support and regression tests

wemrysi:
Adds an implementation of `ExecutePlan` for MarkLogic and enables a handful of regression tests.

Some of the regression tests that were using the implicit numerical field names were modified to use explicit aliases to avoid a mismatch in expectations due to the current handling of numeric field names, which seemed reasonable as the field names themselves aren't what is being tested.


marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
```diff
@@ -83,25 +83,25 @@ object MapFuncPlanner {
 
     // math
     case Negate(x)      => (-x).point[F]
-    case Add(x, y)      => (x + y).point[F]
-    case Multiply(x, y) => (x * y).point[F]
-    case Subtract(x, y) => (x - y).point[F]
-    case Divide(x, y)   => (x div y).point[F]
-    case Modulo(x, y)   => (x mod y).point[F]
+    case Add(x, y)      => binOp[F](x, y)(_ + _)
```
mossprescott: Is this about evaluation order? Cleaning up the xquery syntax? Again, mostly curious and not a blocker.

mossprescott:
Looks good to me and awesome to see a dozen or so tests passing out of the gate.

Unfortunately, it now needs a merge from master. 😞

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
wemrysi: Yes, as `x` and `y` can be arbitrary expressions, there was some ambiguity in certain situations. However, it is a bit more hammer than is needed in many situations (like constants or other basic expressions), that can be something dealt with later as this change should only improve correctness possibly at the expense of clarity.

wemrysi:
@mossprescott Ok, conflicts resolved and building. Unfortunately, I had to disable three of the tests that were passing prior to the recent commits to master as they now fail :disappointed: The ones involving joins fail with a "non theta join condition found" error (first I've heard of this one), not sure why the third is failing.

sellout:
@wemrysi Hrmm, is that a regression _since_ my join fixes the other day? (It’s hard to tell from the PR what was new in the merge.)

Also, why “skip” rather than “pending”?

wemrysi:
@sellout Yes, the passing build above was late in the day yesterday CDT and merging in master just now caused the `projectIndexWithObjectFlatten`, `3WayJoin` and `innerEquiJoin`tests to fail.

wemrysi:
And "skip" because running all of them as "pending" locally just takes too long at the moment.

wemrysi:
@sellout Sorry, read too quickly, no it looks like both the success and the failure included your join fixes, the interesting `QScript` related commits between the success and the failure (determined by limiting to your and @alissapajer's commits and filtering out merges, changelog additions) are
```
ab3c5d30b389e6e55384cf2af403b558a9f92b97 Implement Distinct & fix Union.
a48370f61ad66d8ef354f77af30a465db270bcb9 Clean up Normalizable.
ca836b2b418dd1657ce6738b851b65f7afca0b7c A collection of QScript fixes.
7cebbd640c31008202cf492d25a040dec829fc36 Refactor Coalesce and Normalizable.
```

sellout:
Ah! The join tests hit #1539.

I’m not sure why the third is failing – it at least generates QScript. At a quick glance it looks reasonable (definitely would take some time to dig through the provenance details) – it has a Join in it, but it should, since it has to join the collection with a flattened version of the collection (which is not the same as the self-joins in #1539).

mossprescott:
Going to assume these issues with specific queries can be worked out in the future and merging this as is.

PR #1580, opened 2016-10-13
Useful RenderTree instances for the QScript types

mossprescott:
Breaks out each QScript node in the rendered tree, so no manual re-indenting is needed to make sense of them.

Caveat: embedded `FreeQS` terms are not properly expanded here, because I couldn't figure out how to handle the recursive type (e.g. `ThetaJoin` contains `FreeQS` contains `QScriptTotal` contains `ThetaJoin` ...). Unfortunately that means simple joins and even plans with `Subset` aren't improved that much.

Before:
```
LeftShift((),SrcHole,ConcatMaps(MakeMap(Constant(Str(city)), ProjectField(Guard(RightSide, Obj(Map(),Some(Top)), RightSide, Undefined()), Constant(Str(city)))), MakeMap(Constant(Str(state)), ProjectField(Guard(RightSide, Obj(Map(),Some(Top)), RightSide, Undefined()), Constant(Str(state))))))
╰─ Read(Read(/test/zips))
```

After:
```
LeftShift
├─ Read(Read(/test/zips))
├─ Struct
│  ╰─ ○
╰─ Repair
   ╰─ ConcatMaps
      ├─ MakeMap
      │  ├─ Constant(Str(city))
      │  ╰─ ProjectField
      │     ├─ Guard
      │     │  ├─ RightSide
      │     │  ├─ Type(Obj(Map(),Some(Top)))
      │     │  ├─ RightSide
      │     │  ╰─ Undefined
      │     ╰─ Constant(Str(city))
      ╰─ MakeMap
         ├─ Constant(Str(state))
         ╰─ ProjectField
            ├─ Guard
            │  ├─ RightSide
            │  ├─ Type(Obj(Map(),Some(Top)))
            │  ├─ RightSide
            │  ╰─ Undefined
            ╰─ Constant(Str(state))
```


sellout:
Do we not have a `RenderTree[Free[F, A]]` given `RenderTree[A]` and `Delay[RenderTree, F]`? I thought we did, but maybe that’s only for `Show`. But that would give us a handler for `FreeQS`.

mossprescott:
We didn't, but I implemented it. Maybe it doesn't work because it requires `Functor[F]`? Not sure if I can get rid of that.

sellout:
But we have Functor for all of the QScript stuff. Probably just need to sort out why it can’t find some implicits. In any case, this is a big improvement.

mossprescott:
No doubt the implicits are the problem. I was asking about RenderTreeT earlier because the `Delay[Show, ...]` instances all require only `ShowT`, but I could not get the same trick to work with RenderTreeT. I'm probably missing something simple, but it will be easy to improve this later on. Just hoping to get something in place for everyone who's having to stare at these plans.

connector/src/main/scala/quasar/qscript/JoinSide.scala
```diff
@@ -25,4 +28,6 @@ final case object RightSide extends JoinSide
 object JoinSide {
   implicit val equal: Equal[JoinSide] = Equal.equalRef
   implicit val show: Show[JoinSide] = Show.showFromToString
+  implicit val renderTree: RenderTree[JoinSide] =
+    RenderTree.make(t => Terminal(List(t.shows, "JoinSide"), None))
```
sellout: Don’t we actually have a `RenderTree.fromShow` or something for this case?

sellout: Checked. We do, but it does `Terminal(List("JoinSide"), t.shows.some)` … is the distinction important?

mossprescott: `fromShow` is a little more verbose; in this case it would look like `╰─ JoinSide(LeftSide)` whereas my version looks like `╰─ LeftSide`. I could go either way.

sellout: I’m fine with either – but there are other cases in this PR that are identical to `fromShow` (I think).

connector/src/main/scala/quasar/qscript/JoinType.scala
```diff
@@ -27,4 +30,6 @@ final case object RightOuter extends JoinType
 object JoinType {
   implicit val equal: Equal[JoinType] = Equal.equalRef
   implicit val show: Show[JoinType] = Show.showFromToString
+  implicit val renderTree: RenderTree[JoinType] =
+    RenderTree.make(t => Terminal(List(t.shows, "JoinType"), None))
```
mossprescott: @sellout: Going through these; this one is similar to the other, which suggests an alternative to `fromShow` which just formats this way instead of the other way since I think each version is useful in different circumstances.

connector/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -47,6 +47,10 @@ object ReduceIndex {
       case ReduceIndex(idx) =>
         Cord("ReduceIndex(") ++ idx.show ++ Cord(")")
     }
+
+  implicit def renderTree: RenderTree[ReduceIndex] = RenderTree.make {
+    case ReduceIndex(idx) => Terminal(List("ReduceIndex"), idx.shows.some)
```
mossprescott: This one is equivalent to fromShow, and should be done in this style, so I'll switch it.

mossprescott: Arg. Correction; this one is another where fromShow is redundant: `ReduceIndex(ReduceIndex(...))`.

mossprescott:
@sellout: Here's another take with less duplication. And merged from master preemptively.

sellout:
Thanks. Looks great. Will merge when green.

Issue #1578, opened 2016-10-13
order by query should compile to QScript


alissapajer:
```scala
       val lp = fullCompileExp("select * from zips order by city") 
```

This is a pending test and should be enabled.

alissapajer:
Closed by the merge of #1582

Issue #1569, opened 2016-10-12
Elide identity mappings


wemrysi:
A request to elide `Map((), SrcHole)` nodes from `QScript` as seen in the plan for the following query

`select * from /ml8/census/some-zips`
```
Logical Plan:
Let('tmp0)
├─ Read(/ml8/census/some-zips)
╰─ Let('tmp1)
   ├─ Invoke(SQUASH)
   │  ╰─ Free('tmp0)
   ╰─ Free('tmp1)

Optimized:
Invoke((LIMIT))
├─ Invoke(SQUASH)
│  ╰─ Read(/ml8/census/some-zips)
╰─ Constant(Int(11))

Typechecked:
Invoke((LIMIT))
├─ Invoke(SQUASH)
│  ╰─ Read(/ml8/census/some-zips)
╰─ Constant(Int(11))

QScript:
List
├─ Unreferenced
╰─ Subset((),LeftShift(Read(/census/some-zips),SrcHole,RightSide),Take,Map(Unreferenced,Constant(Int(11))))
   ╰─ ExternallyManaged(Extern)

QScript (ShiftRead):
List
├─ Unreferenced
╰─ Subset((),Map(ShiftedRead(/census/some-zips, ExcludeId),SrcHole),Take,Map(Unreferenced,Constant(Int(11))))
   ╰─ ExternallyManaged(Extern)

QScript (Optimized):
List
├─ Unreferenced
├─ Subset((),ShiftedRead(/census/some-zips, ExcludeId),Take,Map(Unreferenced,Constant(Int(11))))
│  ╰─ ExternallyManaged(Extern)
╰─ Map((),SrcHole)
   ╰─ ExternallyManaged(Extern)
```

sellout:
@wemrysi This is a normalization that exists. I can make the change, but I’m not set up to run tests against MarkLogic. If you want to make the change yourself, here it is:

```diff
modified   marklogic/src/main/scala/quasar/physical/marklogic/fs/queryfile.scala
@@ -65,6 +65,7 @@ object queryfile {
       // TODO[scalaz]: Shadow the scalaz.Monad.monadMTMAB SI-2712 workaround
       import WriterT.writerTMonad
       val rewrite = new Rewrite[Fix]
+      val C = Coalesce[Fix, MLQScript, MLQScript]
 
       def phase(main: MainModule): PhaseResults =
         Vector(PhaseResult.detail("XQuery", main.render))
[...]

Issue #1545, opened 2016-10-10
Normalization around “shape-preserving” operations.


sellout:
Take/Drop are fine, but since Sort and Filter rely on the shape of the “common src”, it is harder to move them around.

Some potential directions (not mutually-exclusive):
- add normalizations to treat `X(Filter(Y))` the way we would treat `X(Y)`
- add normalizations to push Filter/Sort after other operations – the way we do with Take/Drop.
- add a parameter to Filter/Sort for the result expression (otherwise, these ops generally need a subsequent `Map` operation, even after being pushed to the end of the plan)

PR #1582, opened 2016-10-14
Fixes for Filter & Sort

### LINKED ISSUE #1542: QScript basic `Filter` query
```scala
  "convert a filter" in {
      // "select * from foo where bar between 1 and 10"
      convert(
        listContents.some,
        StdLib.set.Filter(
          lpRead("/bar"),
          relations.Between(
            structural.ObjectProject(lpRead("/bar"), LP.Constant(Data.Str("baz"))).embed,
            LP.Constant(Data.Int(1)),
            LP.Constant(Data.Int(10))).embed).embed) must
      equal(chain(
        ReadR(rootDir </> file("bar")),
        QC.inj(LeftShift((),
          HoleF,
          Free.point(RightSide))),
        QC.inj(Filter((),
          Free.roll(Between(
            ProjectFieldR(HoleF, StrLit("baz")),
            IntLit(1),
            IntLit(10)))))).some)
    }
```

The above query should compile as shown to qscript.

sellout:
Fixes #1542, #1569, and #1578.

Also some improvement on #1545, but only the most basic case so far.


connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -562,7 +562,7 @@ class Transform
       Target(a1.ann, QC.inj(Subset(merged.src, merged.lval, Drop, Free.roll(FI.inject(QC.inj(reifyResult(a2.ann, merged.rval)))))).embed).right
 
     case LogicalPlan.InvokeFUnapply(set.OrderBy, Sized(a1, a2, a3)) =>
-      val AutoJoin3Result(base, dataset, keys, directions) = autojoin3(a1, a2, a3)
+      val AutoJoinResult(base, dataset, keys) = autojoin(a1, a2)
```
sellout: Ah, that makes sense – must have been when I merged in all the type changes, and we didn’t yet have a non-pending sort test, because of the over-complicated provenance.

But, I don‘t think the `autojoin3` would have broken anything – just made provenance even _more_ complicated. Only noticed it as I was cleaning up provenance in the tests.

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -108,11 +107,129 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
     }
 
+    // TODO: This would benefit from better normalization around Sort (#1545)
     "convert a basic order by" in {
       val lp = fullCompileExp("select * from zips order by city")
       val qs = convert(listContents.some, lp)
-      qs must equal(chain(RootR).some) // TODO incorrect expectation
-    }.pendingUntilFixed
+      qs must equal(chain(
+        ReadR(rootDir </> file("zips")),
```
alissapajer: @sellout were you able to verify the correctness of the following qscript+provenance by hand? This qscript is still way to complex. I do think it's a good idea to have tests with non-optimal qscript results, with an issue open for making the qscript better, but only if we can actually verify the correctness of the qscript.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -246,6 +246,23 @@ class CoalesceT[T[_[_]]: Recursive: Corecursive: EqualT] extends TTypes[T] {
                 Free.roll(Inject[QScriptCore, QScriptTotal].inj(Map(lb, mf))),
                 sel,
                 rb).some
+            case Filter(Embed(innerSrc), cond) => FToOut.get(innerSrc) >>= QC.prj >>= {
+              case Map(doubleInner, mfInner) =>
+                Map(
+                  FToOut.reverseGet(QC.inj(Filter(doubleInner, cond >> mfInner))).embed,
+                  mf >> mfInner).some
```
alissapajer: _The following is me convincing myself that this is correct :) posted for the sake of posterity_

@sellout so the hole in `mfInner` originally points to `doubleInner`. When we do `mf >> mfInner` here, the hole in that entire expression points to the `Filter` on line 252. I'm not convinced that that is correct. 

`Filter(doubleInner, cond >> mfInner)` makes sense, because we have changed the source of `Filter` from `innerSrc` to `doubleInner`, and so we apply `cond >> mfInner`. But why do we need to call `mf >> mfInner` on line 253? The hole in `mf` points to the original `Filter`. But `mfInner`'s hole points to `doubleInner`. And even though the new `Filter` has `doubleInner` as its source, we already apply `mfInner` on line 252 when the `Filter` gets run... 

_But_ by its nature, the `Filter` on line 252 is essentially `doubleInner`, just filtered. So it makes sense that we do `mf >> mfInner` here, since `mfInner` points to `doubleInner`.

alissapajer:
@sellout LGTM, except for my question about the complex expected results in two of the test queries. Once that's answered/addressed, +1 to merge

connector/src/test/scala/quasar/qscript/QScript.scala
sellout: So, I somewhat disagree. E.g., if we had had a non-verified version of the order-by test in, we wouldn’t have had the `autojoin3` regression.

Obviously, correct tests are better, but seeing something fail when results change is helpful in itself. I did read through the provenance, and it seemed sensible (e.g., I noticed the issue with `autojoin3` because of it), but I definitely didn’t trace through it, and I wouldn’t be surprised if it had the same issue as #1583.

Eliminating the static provenance info from identity is important, too – I see that as job #1 post-release. Then we’ll actually be able to show connector plans to users without being embarrassed. And will actually be able to tell if we have bugs in provenance more easily, too.

alissapajer: Ok, that is a good point about noticing if a query changes. The change could either be for the better (then we update the expected result) or for the worse (then we figure out the bug we introduced). For example, some query could suddenly spring an `AutoJoin` where it didn't previously have one, and this approach would catch that.

So I think you have me convinced :)

alissapajer: But I do think we should have a comment above tests like this that says something like "TODO: this result is more complex than desired, because of problems X and Y. And due to the complexity, the results are not completely verifiable by hand."

@sellout so could you drop in a line like that for these two queries?

alissapajer:
@sellout ok feel free to merge in master now :)

connector/src/test/scala/quasar/qscript/QScript.scala
sellout: BTW, I did verify that the provenances here _don’t_ suffer from #1583, and in fact I think that bug is already fixed.

alissapajer:
@sellout there are still conflicts with master

sellout:
Yep – tests are still running locally.

alissapajer:
fixes #1542 
fixes #1569 
fixes #1578 

Github requires each issue number to be prefaced with a keyword.

alissapajer:
It looks like the fixes/closes triggers don't hit if you add them post-merge. I'll manually close the issues from the above list that are still open.

PR #1563, opened 2016-10-11
Preliminary Couchbase QScript planner

drostron:
Stated and unstated ToDos remain. A few QScript terms are unimplemented due to recent arrival or unclear mapping to N1QL. 

Addresses: #1321 and #1322

PR #1584, opened 2016-10-14
Add ability to ignore field order in query regression tests

drostron:
- Enable trivial.test against Couchbase

Merge after: #1563


sellout:
Looks good – will merge when green.

sellout:
If you have time to kill (😉) you could switch more `select *` queries to use this.

drostron:
Heh, true. I figure Couchbase will motivate that transition as those queries are enabled against it.

mossprescott:
This is missing revision.md, otherwise I would be ready to merge it.

PR #1585, opened 2016-10-14
Utilize bulk operations for Couchbase insertion

drostron:
Merge after: #1584


mossprescott:
Needs a changelog, but otherwise LGTM.

I think we should look into registering with the Observable, as opposed to blocking on it, but this is no doubt a step in the right direction. Will merge as soon as #1584 is in.

drostron:
I like the idea of not blocking too, but I'm unsure how to map the needs of the interpreter to the capabilities of Observable. Interested to hear your ideas on approaching the matter at some point.

PR #1588, opened 2016-10-14
Run StdLib tests on sparkcore's MapFunc implementation

mossprescott:
This runs all the tests, skipping the Extract functions and the tests for `Power` since the implementation is limited to small values. 

Other than that, it found only a few small problems which are fixed here.

I tried to factor out the parts that should be common to QScript/MapFunc implementations, and included a few TODO comments about how that might work.

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/CoreMap.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/CoreMap.scala
@@ -178,6 +178,7 @@ object CoreMap extends Serializable {
     }).right
     case ConcatArrays(f1, f2) => ((x: Data) => (f1(x), f2(x)) match {
       case (Data.Arr(l1), Data.Arr(l2)) => Data.Arr(l1 ++ l2)
+      case (Data.Str(s1), Data.Str(s2)) => Data.Str(s1 ++ s2)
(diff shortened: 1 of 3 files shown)


sparkcore/src/main/scala/quasar/physical/sparkcore/fs/CoreMap.scala
```diff
@@ -352,19 +353,19 @@ object CoreMap extends Serializable {
 
   private def between(d1: Data, d2: Data, d3: Data): Data = (d1, d2, d3) match {
     case (Data.Int(a), Data.Int(b), Data.Int(c)) =>
-      Data.Bool(a <= b && b <= c)
+      Data.Bool(b <= a && a <= c)
```
rabbitonweb: Then it means that function semantics were `a b c   =>     b  a  c` - wow, that was confusing! good you wrote those tests!

mossprescott: Yes, it's mirroring SQL's syntax: `a between b and c`.

mossprescott:
@rabbitonweb can you go ahead and merge?

mossprescott:
Merged latest from master. Will ping you when it's built.

Issue #1559, opened 2016-10-10
compile a basic wildcard inner join to QScript


alissapajer:
```scala
    // an example of how logical plan expects magical "left" and "right" fields to exist
    "convert magical query" in {
      // "select * from person, car",
      convert(
        None,
        LP.Let('__tmp0,
          StdLib.set.InnerJoin(lpRead("/person"), lpRead("/car"), LP.Constant(Data.Bool(true))).embed,
          identity.Squash(
            structural.ObjectConcat(
              structural.ObjectProject(LP.Free('__tmp0), LP.Constant(Data.Str("left"))).embed,
              structural.ObjectProject(LP.Free('__tmp0), LP.Constant(Data.Str("right"))).embed).embed).embed)) must
      equal(chain(RootR).some) // TODO incorrect expectation
    }
```

This test is `pendingUntilFixed`.

#1556 eliminates the `"left"` and `"right"` fields, but this ticket will likely be completed before that one is resolved.

rabbitonweb:
Copy-paste conversation with @sellout 


for
```select * from `/rabbit/hadoop/test4.txt` as left join `/rabbit/hadoop/test2.txt` as right on left.country = right.country```
I get 0 results


sellout [1:06]  
I simplified it down to this:
```Subset(
  Unreferenced,
  EquiJoin(
    Unreferenced,
      Map(ShiftedRead(/rabbit/hadoop/test4.txt, ExcludeId),
        Guard(SrcHole, Obj(Map(),Some(Top)), SrcHole, Undefined())),
      Map(ShiftedRead(/rabbit/hadoop/test2.txt, ExcludeId),
        Guard(SrcHole, Obj(Map(),Some(Top)), SrcHole, Undefined())),
      MakeArray(SrcHole),
      MakeArray(SrcHole),
      Inner,
      Guard(RightSide, Obj(Map(),Some(Top)), Guard(LeftSide, Obj(Map(),Some(Top)), ConcatMaps(LeftSide, RightSide), Undefined()), Undefined())),
  Take,
  Map(Unreferenced,Constant(Int(11))))
```

[1:06]  
Which looks right except that the keys are `MakeArray(SrcHole)` when they should be `MakeArray(ProjectField(SrcHole, StrLit("country")))`.

[1:07]  
I.e., it’s only going to return a result when both sides are equal.



Long version of generated QS

```
QScript:
List
├─ Unreferenced
[...]

Issue #1550, opened 2016-10-10
simplify Provenance representation


sellout:
Provenance is reified into QScript to be used by bucketed operations, autojoin, etc.

The current representation is verbose and includes unnecessary components. We need a new representation that separates the parts that we need during QScript compilation (for correctly generating autojoins) and the parts that need to be reified for bucketing.

alissapajer:
@sellout wondering if this ticket could be broken into multiple tickets. 

This ticket was partially completed with the merging of #1592, which reifies only the provenance information that is needed past the compile phase.

So I'm thinking, perhaps it makes sense to close this ticket since #1592 is merged, and then add new tickets giving the details of the remaining provenance simplification work? Since only the reification work was initially described here, we could use descriptions of the remaining work.

alissapajer:
@sellout after this release is out, could you describe the remaining work in this ticket? I'm not sure exactly what further work you have in mind to be completed.

PR #1592, opened 2016-10-17
Simplify runtime Provenance representation.

### LINKED ISSUE #1558: compile a shift/unshift array query to QScript
```scala
    "convert a shift/unshift array" in {
      // "select [loc[_:] * 10 ...] from zips",
      convert(
        None,
        makeObj(
          "0" ->
            structural.UnshiftArray(
              math.Multiply(
                structural.ShiftArrayIndices(
                  structural.ObjectProject(lpRead("/zips"), LP.Constant(Data.Str("loc"))).embed).embed,
                LP.Constant(Data.Int(10))).embed))) must
      equal(chain(
        RootR,
        QC.inj(LeftShift((),
          ProjectFieldR(HoleF, StrLit("zips")),
          Free.point[MapFunc, JoinSide](RightSide))),
        QC.inj(LeftShift((),
          Free.roll(DupArrayIndices(
            ProjectFieldR(HoleF, StrLit("loc")))),
          Free.roll(Multiply(Free.point(RightSide), IntLit(10))))),
        QC.inj(Reduce((),
          HoleF, // FIXME provenance needs to be here
          List(ReduceFuncs.UnshiftArray(HoleF[Fix])),
          Free.roll(MakeMap[Fix, FreeMapA[ReduceIndex]](
            StrLit[Fix, ReduceIndex]("0"),
            Free.point(ReduceIndex(0))))))).some)
    }
```

This query is `pendingUntilFixed`.

sellout:
Previously we reified _all_ of provenance into QScript, however much of
it is static and is only required during compilation. So we now only
reify the dynamic portion required for bucketing, autojoin, etc. Which
simplifies queries greatly.

This also fixes #1558 and #1559 (tests that were pending). And also
improves both typechecker error messages and QScript phase results.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Provenance.scala
+++ b/connector/src/main/scala/quasar/qscript/Provenance.scala
@@ -14,80 +14,172 @@
  * limitations under the License.
  */
 
-package quasar.qscript
+package quasar.qscript.provenance
 
 import quasar.Predef._
 import quasar.contrib.matryoshka._
+import quasar.ejson.EJson
+import quasar.fp._
+import quasar.qscript._
 import quasar.qscript.MapFuncs._
 
+import scala.Predef.$conforms
+
 import matryoshka._
+import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
-// NB: Should we use char lits instead?
-class Provenance[T[_[_]]: Corecursive: EqualT] {
-  private def tagIdentity[A](tag: String, mf: Free[MapFunc[T, ?], A]): FreeMapA[T, A] =
-    Free.roll(MakeMap[T, FreeMapA[T, A]](StrLit(tag), mf))
+// TODO: Convert to fixed-point
+sealed abstract class Provenance[T[_[_]]]
+@Lenses final case class Nada[T[_[_]]]() extends Provenance[T]
+@Lenses final case class Value[T[_[_]]](expr: FreeMap[T]) extends Provenance[T]
+@Lenses final case class Proj[T[_[_]]](field: T[EJson]) extends Provenance[T]
+@Lenses final case class Both[T[_[_]]](l: Provenance[T], r: Provenance[T])
+    extends Provenance[T]
+@Lenses final case class OneOf[T[_[_]]](l: Provenance[T], r: Provenance[T])
+    extends Provenance[T]
+@Lenses final case class Then[T[_[_]]](l: Provenance[T], r: Provenance[T])
+    extends Provenance[T]
 
-  // provenances:
-  // projectfield: f
-  // projectindex: i
-  // join:         j [] // should be commutative, but currently isn’t
-  // union:        u [] // should be commutative, but currently isn’t
-  // nest:         n []
-  // shiftmap:     m
-  // shiftarray:   a
-  def projectField[A](mf: FreeMapA[T, A]) = tagIdentity("f", mf)
-  def projectIndex[A](mf: FreeMapA[T, A]) = tagIdentity("i", mf)
-  def shiftMap[A](mf: FreeMapA[T, A]) = tagIdentity("m", mf)
-  def shiftArray[A](mf: FreeMapA[T, A]) = tagIdentity("a", mf)
-
-  def join[A](left: FreeMapA[T, A], right: FreeMapA[T, A]) =
-    tagIdentity("j",
-      Free.roll(ConcatArrays(
-        Free.roll(MakeArray[T, FreeMapA[T, A]](left)),
-        Free.roll(MakeArray[T, FreeMapA[T, A]](right)))))
-
-  def union[A](left: FreeMapA[T, A], right: FreeMapA[T, A]) =
-    tagIdentity("u",
-      Free.roll(ConcatArrays(
-        Free.roll(MakeArray[T, FreeMapA[T, A]](left)),
-        Free.roll(MakeArray[T, FreeMapA[T, A]](right)))))
-
-  def nest[A](car: FreeMapA[T, A], cadr: FreeMapA[T, A]) =
-    tagIdentity("n",
-      Free.roll(ConcatArrays(
-        Free.roll(MakeArray[T, FreeMapA[T, A]](car)),
-        Free.roll(MakeArray[T, FreeMapA[T, A]](cadr)))))
-
-  def joinProvenances(leftBuckets: List[FreeMap[T]], rightBuckets: List[FreeMap[T]]):
-      List[FreeMap[T]] =
-    leftBuckets.alignWith(rightBuckets) {
-      case \&/.Both(l, r) => if (l ≟ r) l else join(l, r)
-      case \&/.This(l)    => join(l, NullLit())
-      case \&/.That(r)    => join(NullLit(), r)
+object Provenance {
+  // TODO: This might not be the proper notion of equality – this just tells us
+  //       which things align properly for autojoins.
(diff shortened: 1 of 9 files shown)


connector/src/main/scala/quasar/qscript/Provenance.scala
```diff
@@ -14,80 +14,172 @@
...
+      case (_, _) => None
+    }
+
+  def rebase0(newBase: FreeMap): Provenance => Option[Provenance] = {
+    case Value(expr) => Value(expr >> newBase).some
+    case Both(l, r)  => (rebase0(newBase)(l), rebase0(newBase)(r)) match {
+      case (None,     None)     => None
+      case (None,     Some(r0)) => Both(l, r0).some
+      case (Some(l0), None)     => Both(l0, r).some
+      case (Some(l0), Some(r0)) => Both(l0, r0).some
```
mossprescott: Seems like there's probably a simpler way to express this repeated pattern, but at least you could extract it as `def rebaseBoth(l, r, f: (FreeMap, FreeMap) => Provenenance): Option[Provenenance]`.

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -119,118 +110,39 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
-                                          ejsonProjectField(ejsonStr("zips"))),
-                                      ejsonNull)))))))))),
-                              IntLit(1))))),
-                          Free.roll(Constant(ejsonNullArr)))))))))))))),
-              // }}}
+                  ProjectIndexR(
+                    ProjectIndexR(
+                      ProjectIndexR(ProjectIndexR(RightSideF, IntLit(0)), IntLit(0)),
+                      IntLit(0)),
+                    IntLit(0)))))),
```
mossprescott: Nice.

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -500,24 +404,42 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
+          Free.roll(ConcatArrays(
+            Free.roll(ConcatArrays(
+              Free.roll(MakeArray(
+                Free.roll(ConcatArrays(
+                  Free.roll(MakeArray(RightSideF)),
+                  Free.roll(MakeArray(
+                    ProjectIndexR(
+                      ProjectIndexR(LeftSideF, IntLit(0)),
+                      IntLit(0)))))))),
+              Free.roll(MakeArray(RightSideF)))),
```
mossprescott: Looks like you could use something like `ConcatArraysM`, if that's feasible in this setting.

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -534,67 +456,54 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
-                        ejsonJoin(
-                          ejsonJoin(
-                            ejsonProjectField(ejsonStr("bar")),
-                            ejsonNull),
-                          ejsonNull),
-                        ejsonNull)))))))))),
-              // }}}
-              Free.roll(MakeArray(Free.roll(ProjectIndex(RightSideF, IntLit(1))))))),
+                Free.roll(MakeArray(ProjectIndexR(RightSideF, IntLit(0)))),
+                // FIXME: This line is wrong.
```
mossprescott: To be fixed in this PR?

frontend/src/main/scala/quasar/types.scala
```diff
@@ -207,7 +207,30 @@ trait TypeInstances {
...
+    case Id => "Id"
+    case Arr(types) => "Arr(" + types.shows + ")"
+    case FlexArr(min, max, mbrs) =>
+      "FlexArr(" + min.shows + ", " + max.shows + ", "  + mbrs.shows + ")"
+    case Obj(assocs, unkns) =>
+      "Obj(" + assocs.shows + ", " + unkns.shows + ")"
+    case cp @ Coproduct(_, _) =>
+      val cos = cp.flatten.map(_.shows)
+      cos.init.mkString(", ") + ", or " + cos.last
+  }
```
mossprescott: Happy to see this for a start, but I also think it might be worth handling some of the common cases which are hard to interpret, since these do show up in errors that users see. For starters `AnyArray` and `AnyObject`.

Small thing: your Oxford comma is going to be wrong when Coproduct has two types.

paulp: Most of these would print exactly that way if you used their toString method - they're case objects.
```
case x @ (Top | Bottom | Null | Str | Int | Dec | ...) => x.toString
```

paulp: (Or, y'know, a default `case x => "" + x`)

sellout: Yeah – I started to use more colloquial names, but then switched back. I can simplify that.

connector/src/main/scala/quasar/qscript/Provenance.scala
alissapajer: Using `scalaz.These` could be one approach.

And in any case, the pattern of these three cases can be extracted out to a function that will take `Both`, `OneOf` or `Then`, since the pattern matches on the optional pairs are otherwise identical.

connector/src/main/scala/quasar/qscript/Provenance.scala
```diff
@@ -14,80 +14,172 @@
...
+    leftBuckets.reverse.alignWith(rightBuckets.reverse) {
+      case \&/.Both(l, r) => if (l ≟ r) l else Both(l, r)
+      case \&/.This(l)    => Both(l, Nada())
+      case \&/.That(r)    => Both(Nada(), r)
+    }.reverse
+
+  def unionProvenances(leftBuckets: List[Provenance], rightBuckets: List[Provenance]):
+      List[Provenance] =
+    leftBuckets.reverse.alignWith(rightBuckets.reverse) {
+      case \&/.Both(l, r) => OneOf(l, r)
```
alissapajer: Would it be simpler here to check for equality (like you did in the `joinProvenance` case above)? And then just return the provenance itself if they are equal? _Somewhere_ we should simplify `OneOf` provenances where both sides are equal.

connector/src/main/scala/quasar/qscript/Provenance.scala
```diff
@@ -14,80 +14,172 @@
...
+      case \&/.This(l)    => OneOf(l, Nada())
+      case \&/.That(r)    => OneOf(Nada(), r)
```
alissapajer: Why is it important to preserve that one side came from `Nada()`? In that case, we just have a value on one side, so why do we preserve/remember the fact that there _could_ have been something on the other side but wasn't? 

And this question goes for the the `joinProvenance` case above.

connector/src/main/scala/quasar/qscript/Provenance.scala
sellout: Yeah, that’s probably an improvement, but it’s still not the correct approach. This is one part of #1550 that’s not addressed in this PR.

connector/src/main/scala/quasar/qscript/Provenance.scala
sellout: We have to preserve that it’s a union, so we have to have _something_ on the other side, and that’s what Nada is for. So that when we later try to union something else into this union. The better version of this is in purescript-mra, which is what we need to get to.

PR #1596, opened 2016-10-17
Add Sort for Spark

rabbitonweb:


paulp:
@rabbitonweb I pushed this a little further, [code here](https://github.com/paulp/quasar/tree/spark_sort). Though I think sticking an Order on Data is a pretty serious thing to be doing and probably warrants more consideration than doing it just because you need it for spark.

sellout:
@paulp We _do_ need a total ordering of our data model, simply because we don’t want to fail on datasets with inconsistent data. It’s been something we’ve known about, but Mongo already does this internally, so we haven’t had to codify our own yet. Eventually, we do want consistent ordering across connectors, but for now, we’ve been happy to just use whatever total ordering may exist.

I think our goal is an ordering that looks like `null < bool < byte < char < num < array < map`, so Order[Data] could go for that.

Also, I have personally wanted to keep the influence of that total ordering as constrained as possible. I.e., only available via `order by`, not via the usual comparison operators (where `null < 3` would be undefined), but I’m not sure if that’s a view shared by @jdegoes or anyone else.

rabbitonweb:
@paulp I don't see the changes, are they local to your fork?

rabbitonweb:
This is now RDY for review.
I've moved 'Order[Data]` to `sparkcore.fs.Planner` and restored previously deleted `Equal` instance for `Data`. In other words ordering is implemented currently only within sparkconnector.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -213,8 +217,63 @@ object Planner {
                 }
             }).map((sc, _)).point[Task])
           )
-        case Sort(src, bucket, order) =>
-          unimplemented("sort")
+        case Sort(src, bucket, orders) =>
+
+          implicit def ordering: Order[Data] = new Order[Data] with Serializable {
```
sellout: I’m a fan of tight scoping, but I feel like defining this within match case is maybe a bit _too_ tight. Maybe define it in the `Planner` object? Also, it should probably be called `dataOrder` (that’s our usual convention unless its in a companion object). Finally, it should have a comment explaining why it’s not yet defined in the `Data` companion object.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -213,8 +217,63 @@ object Planner {
...
+
+          /*
+           * Copy-paste from scalaz's `toScalaOrdering`
+           * Copied because scalaz's ListInstances are not Serializable
+           */
+          implicit val ord: SOrdering[Data] = new SOrdering[Data] {
+            def compare(x: Data, y: Data) = ordering.order(x, y).toInt
+          }
+          
+          implicit val listOrd = new SOrdering[List[Data]] {
```
sellout: These should be rescoped, too.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -62,10 +62,10 @@ class PlannerSpec extends quasar.Qspec with QScriptHelpers with DisjunctionMatch
           case rdd =>
             val results = rdd.collect
             results.size must_== 1
-            results(0) must_== Data.Obj(ListMap(
+            results(0) must_== Data.Arr(List(Data.Int(0), Data.Obj(ListMap(
               "name" -> Data.Str("tom"),
               "age" -> Data.Int(28)
-            ))
+            ))))
```
sellout: Can you explain why this result changed? The previous one looks right, AFAICT.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
paulp: You don't have to write this.
```scala

scala> List(1, 2, 3) < List(3, 2)
<console>:12: error: value < is not a member of List[Int]
       List(1, 2, 3) < List(3, 2)
                     ^

scala> import Ordering.Implicits._
import Ordering.Implicits._

scala> List(1, 2, 3) < List(3, 2)
res1: Boolean = true
```

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
rabbitonweb: Ok, will change right away.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
rabbitonweb: That's change after `ShiftedRead` was fixed in separate PR (took `idStatus` into consideration)  - but this test was not fixed

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
rabbitonweb: @paulp Oh wait, it is actually from Scala. I was mistaken because something similiar exists in Scalaz for `Order`. Thx for the tip, updating now :)

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
rabbitonweb: i've reverted to old version

sellout:
Tried to merge – needs a merge from master.

rabbitonweb:
RDY to merge

PR #1602, opened 2016-10-18
CB: ProjectBucket and ConstDeadEnd as unreachable

drostron:


sellout:
Two things here. First, I’d suggest defining a single `UnreachablePlanner[F[_], QS[_]]`, which you can then instantiate with `QS` = `Const[DeadEnd, ?]` and `ProjectBucket`. Then you also should remove ThetaJoinPlanner and ConstReadPlanner, replacing them with instantiations of UnreachablePlanner.

drostron:
Pushed a commit addressing suggestion.

sellout:
This looks great. I’ll merge when it’s green.

PR #1604, opened 2016-10-19
Restrict `Coalesce` to SQL².

sellout:
Previously, `Coalesce` had to be implemented by connectors, but it’s a SQL-ism.

This also extracts the mapping of SQL/LP functions from the StdLib,
which should also make it easier to implement aliases without expanding
them directly in the parser.


mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -327,7 +327,13 @@ class PlannerSpec extends org.specs2.mutable.Specification with org.specs2.Scala
...
          $read(collection("db", "foo")),
          $project(
-           reshape("0" -> $ifNull($field("bar"), $field("baz"))),
+           reshape("0" ->
+             // TODO: Would be nice to identify this pattern and use $ifNull,
+             //       since MongoDB has it.
+             $cond(
+               $eq($field("bar"), $literal(Bson.Null)),
+               $field("baz"),
+               $field("bar"))),
```
mossprescott: This doesn't preserve the behavior, because `$eq: [..., null]` only matches `null`, while `$ifNull` also matches missing fields and `undefined`. I'm not sure which is the intended behavior of Coalesce.

frontend/src/main/scala/quasar/std/relations.scala
```diff
@@ -240,40 +240,6 @@ trait RelationsLib extends Library {
     },
     untyper[nat._3](t => success(Func.Input3(Type.Bool, t, t))))
 
-  val Coalesce = BinaryFunc(
-    Mapping,
```
mossprescott: Removing this from StdLib means we can't test it in StdLibSpec, which is kind of a bummer because testing behavior around nulls was the next thing on my list.

But I suppose that's a lost cause because eventually we'll have only QS-based backends and we'll convert StdLibSpec to use FreeMap anyway, so it probably doesn't make sense to keep _any_ LP Mapping Func that doesn't have a corresponding MapFunc.

Of course, we can test the translation to LP, and we can test the funcs it translates to.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
sellout: Oh, good, then this fixes a bug in the Mongo connector.

The current semantics of Coalesce are only supposed to match null values. This is the discussion we were having in Slack the other day – whether SQL’s `is null` (and by extension `coalesce`) should have more SQL-like semantics rather than their current MRA-like semantics.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -180,50 +180,157 @@ trait Compiler[F[_]] {
...
+      "FLATTEN_MAP"             -> structural.FlattenMap,
+      "FLATTEN_ARRAY"           -> structural.FlattenArray,
+      "{*:}"                    -> structural.FlattenMapKeys,
+      "[*:]"                    -> structural.FlattenArrayIndices,
+      "SHIFT_MAP"               -> structural.ShiftMap,
+      "SHIFT_ARRAY"             -> structural.ShiftArray,
+      "{_:}"                    -> structural.ShiftMapKeys,
+      "[_:]"                    -> structural.ShiftArrayIndices,
+      "({...})"                 -> structural.UnshiftMap,
+      "[...]"                   -> structural.UnshiftArray)
```
mossprescott: Hmm. This is duplicating the name found in the Funcs themselves. Is that no longer used anywhere?

sellout: Yep, at this point it’s only used by `toString` and `RenderTree`, and should be eliminated.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -523,18 +630,27 @@ trait Compiler[F[_]] {
         findUnaryFunction(name).flatMap(compileFunction[nat._1](_, Func.Input1(a1)))
 
       case InvokeFunction(name, List(a1, a2)) =>
-        findBinaryFunction(name).flatMap(compileFunction[nat._2](_, Func.Input2(a1, a2)))
+        (name.toLowerCase ≟ "coalesce").fold((CompilerState.freshName("c") ⊛ compile0(a1) ⊛ compile0(a2))((name, c1, c2) =>
+          LP.Let(name, c1,
+            relations.Cond(
+              // TODO: Ideally this would use `is null`, but that doesn’t makes it
+              //       this far (but it should).
+              relations.Eq(LP.Free(name), LP.Constant(Data.Null)).embed,
```
mossprescott: Right, so what _is_ the semantics of `Eq(x, Const(null)`, exactly? And what is the intended semantics of `Coalesce`? Where is it specified, especially now that Coalesce exists only in SQL²?

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -597,25 +713,25 @@ object Compiler {
...
-          InvokeF[Cofree[LogicalPlan, Boolean], nat._1](func, Func.Input1(strip(arg)))
+          InvokeF[Cofree[LP, Boolean], nat._1](func, Func.Input1(strip(arg)))
 
         case _ =>
           if (t.head) InvokeF(agg.Arbitrary, Func.Input1(strip(t)))
           else t.tail
       }
     }
-    ann.ana[Fix, LogicalPlan](rewriteƒ)
+    ann.ana[Fix, LP](rewriteƒ)
```
mossprescott: By the way, would have appreciated a separate commit for this refactoring, and one for the removal of the list of std lib functions.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -1043,17 +1043,6 @@ abstract class StdLibSpec extends Qspec {
...
-      "Coalesce" >> {
-        "null" >> prop { (x: Data) =>
-          binary(Coalesce(_, _).embed, Data.Null, x, x)
-        }
-
-        "non-null" >> prop { (x: Data, y: Data) =>
-          x != Data.Null ==>
-            binary(Coalesce(_, _).embed, x, y, x)
-        }
-      }
```
mossprescott: 😞

sql/src/main/scala/quasar/sql/compiler.scala
sellout: So, I feel like the semantics of `Eq(x, Const(null))` is clear – it’s just like the equality semantics of any other value.

The semantics of Coalesce is defined right here – it is exactly `Cond(Eq(x, Null), y, x)`. I _do_ think (as the comment says) that it should be defined as (mixing SQL and LP ASTs a bit here) `Cond(IsNull(x), y, x)`, because it should be based on the semantics of `IsNull`, but currently `is null` is expanded to `= null` in the parser directly. Adding IsNull to the SQL AST will allow us to make that relationship more explicit than a comment. And then, if we change the semantics of `is null` (as I am inclined to), we don’t also have to change the definition of `coalesce`.

sql/src/main/scala/quasar/sql/compiler.scala
sellout: Yeah, sorry about that. Did this in the frustration of typing `LogicalPlan` repeatedly, didn’t think about its impact on the diff.

mossprescott:
This raises similar questions about some other Func/MapFuncs that have obvious translations:

- `Between(a, b, c)` === `b <= a and a <= c`
- `Negate(x)` === `x * -1`
- `Null(str)` === `if (str == "null") null else undefined`
- `Boolean(str)` (similar)

At least `Between` really feels like a SQL weirdness that doesn't need to make it to QS, and it's semantics is weird enough that one backend has already gotten it wrong. On the other hand, `Negate` might be worth something in case its behavior is interesting for floating point values.

Also I think it's been discussed that `Guard` could be implemented with a simpler `Match` primitive.

What principle are we using to decide what merits a MapFunc? Do we intend it to be a minimal set? Is it OK to include things that aren't primitive just because _some_ backend might want to recognize them for special handling?

sellout:
I agree that some of those may be viable, and we should discuss whether they are or not. For the case of `coalesce`, we were giving special privilege to `null` as a value that we didn’t have for anything else, whereas for `Null(str)` and `Boolean(str)`, they exist for consistency with the other string-parsing functions  (but it might make sense to expand them between LP and QS). Additionally, the SQL semantics of `coalesce` are certainly not something we want to represent in LP or QS directly (granted, we‘re currently not representing them in SQL² either).

sql/src/main/scala/quasar/sql/compiler.scala
mossprescott: Agreed after you reminded me that `Coalesce` === `_ = null` and `??` does what I was thinking of.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -180,50 +180,157 @@ trait Compiler[F[_]] {
...
+      "date"                    -> date.Date,
+      "time"                    -> date.Time,
+      "timestamp"               -> date.Timestamp,
+      "interval"                -> date.Interval,
+      "time_of_day"             -> date.TimeOfDay,
+      "to_timestamp"            -> date.ToTimestamp,
+      "date"                    -> date.Date,
+      "SQUASH"                  -> identity.Squash,
+      "oid"                     -> identity.ToId,
+      "(+)"                     -> math.Add,
```
mossprescott: I think there's an opportunity here to get rid of the "()" hack here and just have separate mappings for binary operators and binary functions. That was always just an artifact of the name being encoded into Func.

mossprescott:
So I guess the only thing I'd like to see done here is to pull the duplicative `name` field out of `Func`, and separate the operators and funcs. I think if you do that, you'll find that sql.pprint is being a bit naughty, but that would be worth calling out.

sql/src/main/scala/quasar/sql/compiler.scala
sellout: Yeah, and we actually have an enumeration for the operators, so those cases don’t need to match on strings.

frontend/src/main/scala/quasar/std/library.scala
```diff
@@ -99,12 +99,4 @@ trait Library {
       self(args) ||| that(args)
     }
   }
-
-  // TODO exhaustiveness checking for each arity
-  def unaryFunctions: List[GenericFunc[nat._1]]
-  def binaryFunctions: List[GenericFunc[nat._2]]
-  def ternaryFunctions: List[GenericFunc[nat._3]]
```
sellout: Yep, they seemed to only be used to build up the list of all functions. I’m guessing we had used them in the compiler for matching functions, but then realized we didn’t want to say “couldn’t find function” when the real problem was that it just had the wrong number of arguments.

sellout:
Ok, this has been updated with 1. `GenericFunc#name` gone and 2. SQL operators handled a bit more nicely than functions.

frontend/src/main/scala/quasar/logicalplan.scala
```diff
@@ -148,9 +148,7 @@ object LogicalPlan {
 
   final case class InvokeF[A, N <: Nat](func: GenericFunc[N], values: Func.Input[A, N]) extends LogicalPlan[A] {
     override def toString = {
-      val funcName = if (func.name(0).isLetter) func.name.split('_').map(_.toLowerCase.capitalize).mkString
-                      else "\"" + func.name + "\""
-      funcName + "(" + values.mkString(", ") + ")"
+      func.shows + "(" + values.mkString(", ") + ")"
```
mossprescott: 👍

frontend/src/main/scala/quasar/sql/package.scala
```diff
@@ -225,11 +225,10 @@ package object sql {
       }
       case Ident(name) => _qq("`", name)
       case InvokeFunction(name, args) =>
-        import quasar.std.StdLib.string
         (name, args) match {
-          case (string.Like.name, (_, value) :: (_, pattern) :: (Embed(StringLiteral("\\")), _) :: Nil) =>
+          case ("(like)", (_, value) :: (_, pattern) :: (Embed(StringLiteral("\\")), _) :: Nil) =>
```
mossprescott: ☹️ I think we should just have `Like(…) extends Sql` and get rid of this, but I won't insist on it here.

frontend/src/main/scala/quasar/sql/package.scala
```diff
@@ -226,9 +226,9 @@ package object sql {
       case Ident(name) => _qq("`", name)
       case InvokeFunction(name, args) =>
         (name, args) match {
-          case ("(like)", (_, value) :: (_, pattern) :: (Embed(StringLiteral("\\")), _) :: Nil) =>
+          case ("like", (_, value) :: (_, pattern) :: (Embed(StringLiteral("\\")), _) :: Nil) =>
```
mossprescott: Alright, this I can live with, it being no worse than what we do with the ordinary function syntax.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -224,45 +222,19 @@ trait Compiler[F[_]] {
...
-      "(<>)"                    -> relations.Neq,
-      "(<)"                     -> relations.Lt,
-      "(<=)"                    -> relations.Lte,
-      "(>)"                     -> relations.Gt,
-      "(>=)"                    -> relations.Gte,
       "BETWEEN"                 -> relations.Between,
-      "(??)"                    -> relations.IfUndefined,
-      "(AND)"                   -> relations.And,
-      "(OR)"                    -> relations.Or,
-      "NOT"                     -> relations.Not,
```
mossprescott: Unintentional?

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -224,45 +222,19 @@ trait Compiler[F[_]] {
...
-      "(IF_THEN_ELSE)"          -> relations.Cond,
-      "(LIMIT)"                 -> set.Take,
-      "(OFFSET)"                -> set.Drop,
-      "(..)"                    -> set.Range,
       "ORDER BY"                -> set.OrderBy,
```
mossprescott: I don't think this actually gets used. Doesn't the parser have to handle these specially, if they have spaces in them?

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -277,51 +249,11 @@ trait Compiler[F[_]] {
       "MAKE_ARRAY"              -> structural.MakeArray,
```
mossprescott: Likewise, is there any need to support writing these in SQL²?

mossprescott:
I really like the direction this is going, but it seems like there is still some dodgy stuff happening. See comments on the most recent commit (which github is folding for some reason).

sql/src/main/scala/quasar/sql/compiler.scala
sellout: I considered removing them, but that would bump this from a revision to a breaking change. There are some other breaking SQL² changes I’d like to make, e.g. `is null` semantics, and there’s actually a stupid workaround in the parser for still supporting `is_null` as a _function name_, even after it was removed from the LP.

sql/src/main/scala/quasar/sql/compiler.scala
sellout: Yep, I should remove any with spaces. No way to call them (unless they’re hard-coded in the parser).

sql/src/main/scala/quasar/sql/compiler.scala
sellout: Oh … I did it _intentionally_ because `not` is a unary operator, but it’s true – unary ops that are strings can also be called as functions, so technically this would be a breaking change.

sellout:
The function parts are still dodgy – we basically parse any identifier and say “maybe it’s a function?” until we try to compile it to LP. I do agree that it’s getting better, though. Still a lot of improvements to SQL² to be had.

sql/src/main/scala/quasar/sql/compiler.scala
sellout: Oh wait – that’s not true. They can be called with backticks. So, breaking change to remove them.

sellout: And … the parenthesized operators can also be called with backticks, e.g.
```
pop - loc[0]
```
can be written
```
`(-)`(pop, loc[0])
```
so removing _anything_ from the function list is technically a breaking change. So … maybe I should bite the bullet and break all of these things?

mossprescott: I don’t think `` `(-)` `` is useful, so I say let's go ahead and break that. Likewise `` `ORDER BY`(...)``, since I don't think there’s anything you can do with it that you can't do with parens in some other way.

I'd say anything you can remove without breaking a test should go. And anything that you can't bring yourself to remove that doesn’t have a test needs one. We shouldn’t have to debate what the syntax is supposed to be every time we touch the code.

sellout: So, then should I make this a breaking PR? Or just claim that anything not documented/tested can be removed without it being a breaking change?

sql/src/main/scala/quasar/sql/compiler.scala
mossprescott: Well, yes, but I don't think there's actually much cost to making _this_ breaking change, since there is likely no one using these functions and as far as I know no documentation of them to be updated. When you "fix" `is null`, that will be the one that we want to be clear about; in that case, I suspect users will be affected and we just need to help them understand the new, better way.

In short, even if this goes in as 13.0.0, I don't think it will be a major disruption, and is 14.0.0 comes soon, that should be fine. As long as we're tracking the core, SQL², and the connectors all together with semver, that's life.

sellout:
Updated again.

mossprescott:
This looks good to me. Will merge once it builds.

sellout:
Argh – a bit overzealous in my change, and underzealous in my testing. Should I amend the last commit so they all run cleanly?

mossprescott:
I would have voted for removing those tests, but this is fine too. Merging.

mossprescott:
@sellout: Needs merge from master.

sellout:
Merged master (cleanly). I don’t know why this didn’t trigger a build?

Issue #1603, opened 2016-10-18
Couchbase Connector: Utilize DefinitionError in FileSystemDef interpreter


drostron:
Currently some errors are being swallowed as Task failures and therefore not communicated via the API.

PR #1606, opened 2016-10-19
Utilize DefinitionError in FileSystemDef interpreter

drostron:
addresses: #1603


couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -62,38 +54,51 @@ package object fs {
...
-        uri     <- Task.fromDisjunction(Uri.fromString(connectionUri.value))
-        cluster =  CouchbaseCluster.fromConnectionString(uri.renderString)
-        user    <- Task.fromMaybe(uri.params.get("username").toMaybe)(
-                     new RuntimeException("No username in ConnectionUri"))
-        pass    <- Task.fromMaybe(uri.params.get("password").toMaybe)(
-                     new RuntimeException("No password in ConnectionUri"))
-        manager <- clusterManager(cluster, user, pass)
-      } yield (cluster, manager)
+        uri     <- EitherT.fromDisjunction[Task](
+                     Uri.fromString(connectionUri.value).leftMap(e => physErr(e.message))
```
wemrysi: This should be a configuration error (left side of `DefinitionError`).

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -62,38 +54,51 @@ package object fs {
...
+                   )
+        cluster <- EitherT(Task.delay(
+                     CouchbaseCluster.fromConnectionString(uri.renderString).right
+                   ).handle { case e: Exception => unhandledFSError(e).left })
```
wemrysi: Does `fromConnectionString` just validate the configuration somehow, or does it actually attempt to connect to CouchBase? If the former then this should also probably be a config error, otherwise maybe an `EnvironmentError` (right side of `DefinitionError`).

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -62,38 +54,51 @@ package object fs {
...
+        user    <- EitherT.fromDisjunction[Task](
+                     uri.params.get("username") \/> physErr("No username in ConnectionUri")
```
wemrysi: Should be a config error.

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -62,38 +54,51 @@ package object fs {
...
+                   )
+        pass    <- EitherT.fromDisjunction[Task](
+                     uri.params.get("password") \/> physErr("No password in ConnectionUri")
```
wemrysi: Should be a config error.

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -62,38 +54,51 @@ package object fs {
...
+                   )
+        ctx     <- EitherT(Task.delay(
+                     Option(cluster.clusterManager(user, pass)) \/>
+                       physErr("Unable to obtain a ClusterManager with provided credentials.")
```
wemrysi: Does one of the `EnvironmentError`s make sense here?

wemrysi:
This is definitely heading in the right direction, I just think we have more appropriate methods to report some of these errors. In general, `PhysicalError` is intended to be the `RuntimeException` for backends, any errors we know at mount time should be communicated more precisely and only those truly exceptional cases during operation be reported via `PhysicalError`.

We can also probably catch many things in a general way during the interpretation of `PhysicalEff` (i.e. a transformation that `handle`s any `Exceptions` in `Task` by replacing them with a `PhysErr`.

drostron:
Suggestions appreciated. Ready for another look.

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -45,55 +46,61 @@ package object fs {
...
-      } yield (cluster, manager)
+        uri     <- liftDT(
+                     Uri.fromString(connectionUri.value).leftMap(_.message)
+                   )
+        cluster <- EitherT(Task.delay(
+                     CouchbaseCluster.fromConnectionString(uri.renderString).right
+                   ).handle {
+                     case e: Exception => e.getMessage.wrapNel.left[EnvironmentError].left
+                   })
+        user    <- liftDT(
```
wemrysi: You could validate the presence of both the username and password via `Validation` and then convert the result to a disjunction. It doesn't buy you much now, but I just wanted to mention in case there are more properties to validate in the future (it is why the lhs of `DefinitionError` is a `NonEmptyList`). I'll merge either way.

wemrysi:
LGTM, will merge once conflicts are resolved.

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
drostron: Dig it. I'll update and merge master.

paulp:
> In general, PhysicalError is intended to be the RuntimeException for backends, any errors we know at mount time should be communicated more precisely and only those truly exceptional cases during operation be reported via PhysicalError.

@wemrysi By this time I had pretty much worked that out, but is what you've just said documented somewhere? In particular, there's zero documentation in PhysicalError.scala itself. It can be a bit frustrating having to speculate about the role of such classes.

wemrysi:
@paulp Agreed and there isn't any comprehensive documentation about this anywhere. There is a small amount of docs on `DefinitionError` attempting to identify the two cases, so maybe if you found that first it has a chance of making sense, but no docs on `PhysicalError`. I am about to work on a similar refactoring for MarkLogic (#1608) and will add docs as part of it.

drostron:
Realized that `clusterManager` isn't validating passed credentials. I'm going to update with an additional call that does.

drostron:
Ready for another look.

Issue #1612, opened 2016-10-20
Towards a more expressive representation of file existence


drostron:
Connectors differ in their ability to represent file and directory existence. 

A short term solution was recently put in place where file existence is indistinguishable from an empty file. This creates confusion and suggests the need for a longer term solution.

So far discussion around this has been a bit ad-hoc. This exists as a central point to capture discussion and solutions.

paulp:
There are some fundamental questions which must be grappled with.

1) Which structural elements have identity - an independent existence, a lifespan potentially longer than that which they enclose - and which ones are simply consequences?

2) Where there are opportunities for inconsistent answers, what are we going to do?

In git, there is no notion of a directory. A directory "exists" if a file exists which includes that directory path. If you delete the last file which includes a directory, you "delete" the directory as a consequence.

It sounds like the situation with couchbase is analogous: that there is no notion of a file, only of a document. If you truncate a document to 0 bytes, you have "deleted" the file as a consequence, because the file never existed from couchbase's POV.

Asking whether something exists is a very specific question which generalizes to: what is known **about** what's at this path? And there's no way to know what we might know - or might want to know - about what's at a path, nor to anticipate what a future backend might want to be able to expose. But we CAN provide a means for it to do so, and model this question and more.

```scala
// AttrMap is a heterogeneous map
// The key encodes the attribute and the type of the value
final case class Metadata(path: APath) extends QueryFile[AttrMap]

```
This can subsume at least ListContents and FileExists from QueryFile, but the possibilities go very much further.

As soon as you have a method like `list` and an overlapping one like `exists` you have to decide how they interact.

- Can a file exist, but not be listed? 
- Can a file be listed, but not exist?

Both combinations have uses, but the first especially is almost inevitable. If you start considering computation on demand, one might want an infinite number of paths to translate into data depending on the path - but the path doesn't exist until it's asked about. So to `list` the contents of the directory is really just a particular metadata query: what are the doors which leave from here?

Unix filesystems don't let us answer that question in any way other than an exhaustive enumeration of the possibilities. But you can imagine a world where the answer could be `[0-9]+` and not `0, 1, 2, 3, 4, ... [ls never terminates]`.

I realize this probably seems off in the weeds relative to the question of how to deal with whether something does or does not exist. But having spent a lot of time with these questions I found that they cannot be avoided. They infect everything.

jdegoes:
[...]

PR #1613, opened 2016-10-20
Update errant test description

drostron:
@paulp And yes, looking at this test description makes the need for #1612 apparent 😄


paulp:
I would rather pursue reverting the changes which made this necessary (something I've already done in my precog branch) than fix the test description.

drostron:
That works.

PR #1614, opened 2016-10-20
Attribute new constructs with their EJSON type

### LINKED ISSUE #1575: Attribute new elements with their EJSON type in MarkLogic
When constructing new structures (maps, arrays) the EJSON type of the value needs to be determined and attributed to the element via the `ejson:type` attribute in order to deserialize as anything other than a string.

wemrysi:
When constructing new `Array`s or `Map`s, adds the appropriate `ejson:type` attribute so that values are properly deserialized to types other than `String`.

Also a couple other fixes, making `Data` decoding a bit more accepting, fixing an issue with how sort direction was rendered and left shift for arrays.

Enables another set of regression tests that pass due to the aforementioned improvements with the same caveats of explicit field aliases. One test is also enabled for MarkLogic and marked as pending for Mongo due to the presence of the `value` field (as including the field is not the correct behavior, as I understand it). If this is contentious, I can a) revert it or b) add another test that is skipped by all but Mongo and includes the `value` field.

Fixes #1575.

### DIFF
--- a/marklogic/src/main/scala/quasar/physical/marklogic/xquery/package.scala
+++ b/marklogic/src/main/scala/quasar/physical/marklogic/xquery/package.scala
@@ -39,7 +39,7 @@ package object xquery {
   type PrologW[F[_]] = MonadTell[F, Prologs]
 
   sealed abstract class SortDirection {
-    override def toString = this match {
+    def asOrderModifier: String = this match {
       case SortDirection.Ascending  => "ascending"
       case SortDirection.Descending => "descending"
     }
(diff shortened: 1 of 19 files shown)


marklogic/src/main/scala/quasar/physical/marklogic/xquery/package.scala
```diff
@@ -38,7 +38,12 @@ package object xquery {
   type Prologs = ISet[Prolog]
   type PrologW[F[_]] = MonadTell[F, Prologs]
 
-  sealed abstract class SortDirection
+  sealed abstract class SortDirection {
+    override def toString = this match {
+      case SortDirection.Ascending  => "ascending"
```
mossprescott: If this value is being used in constructing queries (which I can't spot in this diff, but I'm guessing), then I think it would be better to define it as a function and make that clear.

mossprescott:
Just the one issue and otherwise looks good.

wemrysi:
Good suggestion, updated.

PR #1615, opened 2016-10-20
Improve reporting of MarkLogic mount errors

### LINKED ISSUE #1608: Report configuration errors as such when mounting MarkLogic
Configuration and other errors encountered at MarkLogic mount-time aren't reported properly, resulting in useless `500` errors at the web api. The various failures need to be properly categorized as configuration or "environmental" errors to provide useful feedback to users.

wemrysi:
Adds more fine-grained reporting of MarkLogic mount errors along with some documentation.

Also adds a transformation that reifies all non-fatal errors from any connector as `PhysicalError`s, ensuring they are reported properly.

Fixes #1608.

### DIFF
--- a/interface/src/main/scala/quasar/main/package.scala
+++ b/interface/src/main/scala/quasar/main/package.scala
@@ -25,6 +25,7 @@ import quasar.fs._
 import quasar.fs.mount._
 import quasar.fs.mount.hierarchical._
 
+import eu.timepit.refined.auto._
 import monocle.Lens
 import pathy.Path.posixCodec
 import scalaz.{Failure => _, Lens => _, _}, Scalaz._
@@ -47,7 +48,7 @@ package object main {
     quasar.physical.mongodb.fs.mongoDbFileSystemDef[PhysFsEff],
     quasar.physical.mongodb.fs.mongoDbQScriptFileSystemDef[PhysFsEff],
     quasar.physical.postgresql.fs.definition[PhysFsEff],
-    quasar.physical.marklogic.fs.definition[PhysFsEff],
+    quasar.physical.marklogic.fs.definition[PhysFsEff](10000L),
     quasar.physical.sparkcore.fs.local.definition[PhysFsEff],
     quasar.physical.sparkcore.fs.hdfs.definition[PhysFsEff],
     quasar.physical.couchbase.fs.definition[PhysFsEff]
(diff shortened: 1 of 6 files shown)


interface/src/main/scala/quasar/main/package.scala
```diff
@@ -47,7 +50,7 @@ package object main {
     quasar.physical.mongodb.fs.mongoDbFileSystemDef[PhysFsEff],
     quasar.physical.mongodb.fs.mongoDbQScriptFileSystemDef[PhysFsEff],
     quasar.physical.postgresql.fs.definition[PhysFsEff],
-    quasar.physical.marklogic.fs.definition[PhysFsEff],
+    quasar.physical.marklogic.fs.definition[PhysFsEff](10000L),
```
paulp: When magic numbers are dropped into source code it's super helpful to use named argument style, like
```
quasar.physical.marklogic.fs.definition[PhysFsEff](readChunkSize = 10000L)
```

marklogic/src/main/scala/quasar/physical/marklogic/fs/package.scala
```diff
@@ -54,44 +56,62 @@ package object fs {
...
-      (runML, Task.delay(csource.getConnectionProvider.shutdown(null)))
+      GenUUID.type1
+    ).tupled.liftM[DefErrT].flatMap { case (whandles, rhandles, qhandles, seq, genUUID) =>
+      contentSource.flatMapF { cs =>
+        val runCSIO = ContentSourceIO.runNT(cs)
+        val runSIO  = runCSIO compose ContentSourceIO.runSessionIO
+        val runML   = reflNT[Task] :+: runSIO   :+:
+                      runCSIO      :+: genUUID  :+:
+                      seq          :+: rhandles :+:
+                      whandles     :+: qhandles
```
paulp: It's no big deal but I'm doubtful about the wisdom of any kind of formatting which doesn't either put all the arguments on one line or each argument on its own line. Ideally in any multiline expression you should surround the whole expression with parentheses, because it prevents the situation where you get a variation on 
```scala
val x = 1 + 2
  + 3
```
with a truncated expression and an orphan value.

paulp:
Other than those minor comments looks great.

wemrysi:
Thanks for the comments, updated.

PR #1616, opened 2016-10-20
Utilize `=` instead of `like` for files

drostron:
Hopefully this will alleviate some of the connection timeouts seen in Travis.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Couchbase: Utilize `=` instead of `like` for files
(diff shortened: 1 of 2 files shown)


mossprescott:
Guess that wasn't it.

drostron:
Yeah, more hope than reason on this attempt. Looks like times for `quasar.fs.*` were similar with and without this change too.

mossprescott:
I'm going to merge this since it seems like a good idea anyway, even if it doesn't fix the flaky build. I'll restart that last one just to give it a chance to build cleanly before I do that.

And I would rather have this running in Travis even it's not 100% stable, especially since only @dave is running CB locally as far as I know.

mossprescott:
Arg. Should have reloaded the page before commenting. I'm OK with that as well, and I concur on the times not looking better.

Issue #1556, opened 2016-10-10
Eliminate hard-coded “left” and “right” for join access.


sellout:
In LP, it’s implied that a join creates a Map with the fields “left” and “right”. Then they can later be projected as necessary. QScript has to handle these magical names.

In QScript, we prefer to handle this situation by having a `JoinFunc` in join nodes that contains an expression representing how the left and right sides of the join should be combined.

alissapajer:
Related to #1539 in that an improved representation of the combine func in logical plan will also be a useful as a representation of the join condition.

alissapajer:
This ticket will be complete when we have a first  class representation of joins in `LogicalPlan`. This ticket is not a high priority and so will be put off until later.

PR #1617, opened 2016-10-21
Containment of "left" and "right" sides in LogicalPlan.

alissapajer:
The first step toward #1556, which eliminates the hard-coded "left" and "right" fields in join access.

In order to complete that ticket, it will be useful to easily see all the places "left" and "right" are actually used.

### DIFF
--- a/frontend/src/main/scala/quasar/sql/JoinDir.scala
+++ b/frontend/src/main/scala/quasar/sql/JoinDir.scala
@@ -22,22 +22,15 @@ import quasar.std.StdLib._
 
 import matryoshka.Fix
 
-sealed trait JoinDir {
+sealed abstract class JoinDir(val name: String) {
   import structural.ObjectProject
 
-  val name: String
-
-  lazy val data: Data = Data.Str(name)
-  lazy val const: Fix[LP] = LP.Constant(data)
+  val data: Data = Data.Str(name)
+  val const: Fix[LP] = LP.Constant(data)
   def projectFrom(lp: Fix[LP]): Fix[LP] = Fix(ObjectProject(lp, const))
 }
 
 object JoinDir {
-  final case object Left extends JoinDir {
-    val name: String = "left"
-  }
-
-  final case object Right extends JoinDir {
-    val name: String = "right"
-  }
+  final case object Left extends JoinDir("left")
+  final case object Right extends JoinDir("right")
 }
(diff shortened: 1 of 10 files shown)


mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -26,7 +26,7 @@ import quasar.physical.mongodb.expression._
 import quasar.physical.mongodb.planner._
 import quasar.physical.mongodb.workflow._
 import quasar.qscript.SortDir
-import quasar.sql.{fixpoint => sql, _}
+import quasar.sql.{fixpoint => sql, JoinDir, _}
```
sellout: This shouldn’t need `JoinDir` should it? I figure that’s covered by the `_`.

frontend/src/main/scala/quasar/sql/JoinDir.scala
```diff
@@ -16,25 +16,28 @@
...
+import quasar.Predef.String
 import quasar.std.StdLib._
 
 import matryoshka.Fix
 
 sealed trait JoinDir {
-  def projectFrom(t: Fix[LogicalPlan]): Fix[LogicalPlan]
+  import structural.ObjectProject
+
+  val name: String
```
sellout: While you’re in this area, I would define this as `sealed abstract class JoinDir(name: String)` then below, `final case object Left extends JoinDir("left")`.

Other than `abstract class` vs `trait`, I think it’s just personal preference.

mossprescott:
No additional comments from me. Good to see this getting cleaned up; I thought we had pulled all these out ages ago.

frontend/src/main/scala/quasar/sql/JoinDir.scala
sellout: Oh yeah – and I _think_ this would allow you to make the `val`s below non-`lazy`?

(and I modified the previous comment, because I left out a `val` I think is probably necessary)

alissapajer: Aside from removing the need for the `lazy val`, what benefit do you see to using `sealed abstract class` over `sealed trait`?

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
alissapajer: Yup, thanks!

alissapajer:
@mossprescott changes pushed.

PR #1620, opened 2016-10-21
Report errors from the workflow phase

mossprescott:
This produces a much more comprehensible error than the one the refers to JS phase.

With that in hand, it was possible to debug and enable a more comprehensive `date_part` regression test that had been pending.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,2 @@
+- [#1526] improve error reporting for cases that aren't handled in the MongoDB planner (those requiring and auto-self-join).
+- Fix some date_part cases, mostly when evaluated in MongoDB's map-reduce.
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,2 +1,2 @@
-- [#1526] improve error reporting for cases that aren't handled in the MongoDB planner (those requiring and auto-self-join).
+- [#1526] improve error reporting for cases that aren't handled in the MongoDB planner (those requiring an autojoin).
 - Fix some date_part cases, mostly when evaluated in MongoDB's map-reduce.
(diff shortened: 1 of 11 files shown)


mossprescott:
Finally, A legitimate conflict!

it/src/main/resources/tests/temporal/datePartsConverted.test
```diff
@@ -1,45 +1,42 @@
...
-         Also `doy` and `week` are missing because we currently have no JS implementation.",
+  "nb": "`doy`, `week`, and `quarter` are missing because we currently have
+         no JS implementation.",
   "query": "select
-              date_part(\"millennium\", timestamp(commit.committer.date)) as millenium,
+              date_part(\"millennium\", timestamp(commit.committer.date)) as millennium,
               date_part(\"century\", timestamp(commit.committer.date)) as century,
               date_part(\"decade\", timestamp(commit.committer.date)) as decade,
               date_part(\"year\", timestamp(commit.committer.date)) as year,
-              date_part(\"quarter\", timestamp(commit.committer.date)) as quarter,
```
sellout: I see there are a lot of fixes to `quarter` handling, but it’s still been removed here. It _seems_ like it’d be a straightforward translation from the ExprOp one to get a JS one.

CHANGELOG/revision.md
```diff
@@ -0,0 +1,2 @@
+- [#1526] improve error reporting for cases that aren't handled in the MongoDB planner (those requiring and auto-self-join).
```
sellout: “_and_ auto-_self_-join” => “_an_ autojoin”

(Technically not a self-join, because there is a reduction on one side, meaning the provenance of the two datasets differs, even though they started from the same collection.)

PR #1624, opened 2016-10-21
Implement DistinctBy compilation in QScript.

### LINKED ISSUE #1623: Distinct + Order results in a MatchError when converting LogicalPlan to QScript
The query `select distinct city from zips order by pop desc limit 5` (from `it/src/main/resources/tests/citiesByLargestZip.test`) results in a `MatchError` in `Transform#lpToQScript`.

It looks like the `distinct` + `order by` is what causes it as I tried `select distinct city from zips` and `select distinct city from zips limit 5` and both of those work.

sellout:
Fixes #1623.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- add missing DistinctBy compilation.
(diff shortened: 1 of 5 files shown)


sellout:
Took longer than I thought, because actually _testing_ it led to me having to avoid another autojoin.

PR #1632, opened 2016-10-24
Preserve Undefined values in arrays in MarkLogic

wemrysi:
Fixes an issue where undefined values were being collapsed in Arrays meaning they wouldn't be as large as expected and cause `ProjectIndex` to fail in those cases.

Also improve handling of inputs to qscript:length to detect the input type more reliably.

These fixes allow another handful of regression test to pass for MarkLogic.


it/src/main/resources/tests/distinctStar.test
```diff
@@ -9,5 +9,6 @@
     "data": "cities.data",
     "query": "select distinct * from cities where city = \"BOSTON\"",
     "predicate": "equalsExactly",
+    "NB": "This will only ever pass for MongoDB as other backends will consider the '_id' field.",
```
mossprescott: This is currently true, but the upshot is we need to straighten out how we store the data so we can test `distinct *`, which is the point here. For one thing, we could just switch to a data set that doesn’t include `_id`.

Would this test pass if we did that?

sellout:
Dealing with undefined values in arrays is an open problem – we don’t store or return undefined values (well `Data` allows us to, unfortunately – and I think there’s a comment somewhere stating that `NA` should only show up in arrays), but at the same time, we can’t just remove their entries like we can in a map (since it changes the indices of any later values).

it/src/main/resources/tests/countDistinctStar.test
```diff
@@ -3,11 +3,10 @@
   "backends": {
     "mongodb_read_only": "pending",
     "postgresql":        "pending",
-    "marklogic":         "skip",
     "couchbase":         "skip"
   },
   "data": "olympics.data",
-  "query": "select count(distinct *) from olympics",
+  "query": "select count(distinct *) as cnt from olympics",
```
wemrysi: They are currently realiased as `_N`, I have a note to look into something more comprehensive, but it will likely involve an encoding of some sort which I'd like to discuss with others given all the pain we've had in `pathy` with encodings.

Another option is not to use the names as element names but as element values (or try out the built-in `map` type in MarkLogic, which essentially does just that), the drawback here (among others) is the discrepancy between these synthesized structures and how customer data will look, the current scheme keeps them similar.

it/src/main/resources/tests/distinctStar.test
wemrysi: Yes, we should also decide what to do about the numeric values that are truncated when loaded as a 64-bit double.

it/src/main/resources/tests/idNotAliased.test
```diff
@@ -6,11 +6,10 @@
...
         "mongodb_3_2":       "pending",
         "postgresql":        "pending",
-        "marklogic":         "skip",
         "couchbase":         "skip"
     },
     "data": "zips.data",
-    "query": "select _id from zips where city = 'BOULDER' and state = 'CO'",
+    "query": "select _id from zips where city = \"BOULDER\" and state = \"CO\"",
     "predicate": "containsExactly",
     "expected": [{ "_id": "80301" },
```
wemrysi: Yeah, I think placing any MongoDB specific tests in their own directory would be helpful (there are some related to OIDs as well).

it/src/main/resources/tests/selectWildcardAndField.test
```diff
@@ -1,22 +1,25 @@
 {
     "name": "select wildcard and field",
     "backends": {
-        "postgresql": "pending",
-        "marklogic":  "skip",
-        "couchbase":  "skip"
+        "mongodb_2_6":       "pending",
+        "mongodb_3_0":       "pending",
+        "mongodb_read_only": "pending",
+        "mongodb_3_2":       "pending",
```
mossprescott: :-1: 

I don't think we should let tests like this go for MongoDB, even temporarily.

Options I can see:

- Add a test parameter `nestedUnderValue` that directs it to match the result produced by Mongo (and is ignored for other backends.)
- Just change the matcher to accept results under `value` regardless.
- Make a copies of the test for MongoDB and non-MongoDB

it/src/main/resources/tests/idNotAliased.test
wemrysi: Not to mention there are some expectations that expect how the data is stored in Mongo (I think), expecting `{"foo": {"$timestamp": ...}}` for example.

it/src/main/resources/tests/countDistinctStar.test
sellout: @wemrysi Is the problem that it’s not even possible to re-alias them on the way back out? I.e., will we need the fallback evaluator to fix the keys _after_ reading the results?

NB: `EvaluatePlan` _could_ fix this stuff up internally, but `ExecutePlan` has this problem where we have to store it, and we don’t necessarily have a way to apply some transformation to the plan upon reading. This is the problem Mongo has with `{ "value": <actual result> }`, etc.

it/src/main/resources/tests/selectWildcardAndField.test
wemrysi: Hm, ok. I've suggested the third option in other PRs and it was suggested to me to just make MongoDB pending like we do the other connectors that can't pass a test.

I'm fine with adding another test until the `value` bit is fixed.

it/src/main/resources/tests/countDistinctStar.test
mossprescott: @sellout brings up the trickiest point. If the front-end stores the result of a query (as it will when you use the "cache" card (or something like that), we won't have any information about the query when the data is actually read.

mossprescott:
Nice progress.

A couple of the sticky issues are coming up now, and I think we need to start addressing them.

it/src/main/resources/tests/countDistinctStar.test
wemrysi: If we had a faithful encoding, then we could fixup during the conversion to `Data`.

sellout: I don’t think we can truly have a faithful encoding, can we? I mean, whatever we choose, it’s _possible_ that a user would choose to name something that way in data that wasn’t created by Quasar.

But, until we have a fallback evaluator, that might be the best option.

sellout: Oh, I guess if we use a `quasar` namespaced attribute or something?

wemrysi: Yeah, that's why I've punted on it thus far. Though I hadn't thought of leveraging an attribute, that's interesting, and would work at the serialization boundary, but might get tricky in the rest of the plan as we have to know how to handle projects of the aliased fields that might happen later on.

it/src/main/resources/tests/idNotAliased.test
mossprescott: `$timestamp` is the Precise JSON encoding for `Data.Timestamp`, and it should be read into the correct value by the test machinery. See also `$date`, `$time`, etc.

On the other hand, there are some cases where things come back in a form that’s peculiar to MongoDB and isn't mapped to the right type in `Data`. For example, `Data.Time` is represented as string, and `Data.Interval` is a number (milliseconds). This is another thing we need to fix up on the way out, and comes with the same issues as `value` nesting.

wemrysi: Ah, ok I didn't realize that. So if the `Data` from the backend is typed properly then it should match?

it/src/main/resources/tests/countDistinctStar.test
wemrysi: Issue #1642 was created to track this issue.

it/src/main/resources/tests/idNotAliased.test
mossprescott: It _should_, yes. Subject to reality.

wemrysi:
@mossprescott Finally got the green light, I've shuffled around the mongodb tests and #1642 to track the marklogic field name issue (which I'll solve in a separate PR).

PR #1635, opened 2016-10-24
CB: N1QL gen improvements, additional it enablement

drostron:


sellout:
Looks good – will merge when green.

PR #1636, opened 2016-10-24
`connector` no longer depends on `sql`.

sellout:
Other things still to be changed in this vein:
- `core` shouldn’t depend on `sql` (involves extricating SQL from views)
- `common` shouldn’t depend on `js` (move `Data => JS` to mongo
  connector)
- `connector` shouldn’t depend on `frontend` (need to get rid of old
  mongo so fs can expect QScript rather than LP)

Also note that there is a `"test->compile"` dependency from connector to sql. I think this is fine – it means we can use SQL to test QScript.


mossprescott:
Can we avoid splitting packages across projects here?

How about just moving `PathError.scala` and `PhaseResult.scala` to the `common` dir and changing the package declaration in `PathError` (`PhaseResult` is already in the common package, but the file is not in the right place.)

mossprescott:
What's the vision for moving stuff to `common`? Because as is it's hardly pulling its weight.

Do QScript and the filesystem algebras eventually go there, when we break the dependency from connector to frontend? Is something stopping that from happening now?

It's ok if it's just too painful/disruptive to take on right now, but maybe we can record the intention so it's not confusing why there's such a tiny project with an important sounding name.

sellout:
So, I don’t know if you prefer this or not – I do want to move things to packages matching their projects, _but_ I want things to be in the packages for projects they _should_ be in. E.g., parts of `quasar.sql` are still in the frontend project, but they should get moved when file dependencies are detangled.

Common is intended to be things that are shared by frontend and connector, both dependencies and code, so that includes foundation & ejson as dependencies, and for code, it will eventually include Data (until it dies), MapFunc, Type, etc. … but QScript and the filesystem algebras belong solidly in the connector project. Nothing in frontend should know about them.

Overall, this is another incremental step toward de-spaghetti-ing our code. I mentioned some of the others in the description, but yeah, it’s a few steps until common holds all it should. I could perhaps be a bit more proactive – e.g., moving `Data` to common, but that isn’t yet forced, because connector has to depend on frontend until the old mongo planner is gone.

mossprescott:
I’m not super happy about creating messes here and now (packages split across multiple projects) for the sake of future cleanup (things in the right packages when they get moved to the right projects someday). Why not just put a "TODO: move to pkg X in project Y"?

Furthermore, I'm still having trouble grasping the goal state here. We have a lot of types and a lot of projects, and I think it would help to have a way of getting an overview of what talks to what and in terms of what types. E.g SQL² -> LP+Data -> QScript+EJson -> [physical plan], but which projects are involved in each step, and which interfaces (web services; FS algebras; ???)

At the moment, one can’t figure this out by looking at the code, because some of it is in your head (and maybe some others’ heads as well).

mossprescott:
For this PR, I’d like to see the package and location for PhaseResult.scala synced up and the rest we can tackle over time.

Also, you've got conflicts.

common/src/main/scala/quasar/PhaseResult.scala
```diff
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package quasar
+package quasar.common
```
sellout: Yep, this was an oversight..

sellout:
For the goal state of what talks to what – we have a project-level dependency tree in ASCII art in build.sbt, with comments showing how the goal differs from the current state of things.

If I’ve done any _new_ splitting of packages across projects, that’s a mistake. I made new packages – common, connector, frontend, that hold the pieces I’ve moved to those projects. I haven’t done a wholesale moving of everything into project-named directories.

As far as “SQL² -> LP+Data -> QScript+EJson -> [physical plan]”, the goal is to have the first three stages there all use EJson – no `Data`. So, EJson is in the “common” project. LP is in the “frontend” project, because it’s what specific frontends (only SQL² so far) depend on. QScript (and FS algebras) is in the “connector” project, because it’s what connectors  depend on. The transition between LP and QScript is in the “core” project – it’s the core of the compiler. So, yeah, this stuff should be comments in build.sbt. I’ll add those as I make the other changes.

mossprescott:
The comments are helpful. I wish there was a better place for this than build.sbt, but I guess it is what is is. Merged.

sellout:
Once we align package names with projects, we can Scaladoc the packages. Also, if we switch from .sbt build files to .scala ones, we may be able to get Scaladoc out of the build project itself.

PR #1638, opened 2016-10-24
Use type aliases for CoEnv in MapFunc.

alissapajer:


connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -50,41 +50,50 @@ sealed abstract class Ternary[T[_[_]], A] extends MapFunc[T, A] {
 object MapFunc {
   import MapFuncs._
 
+  type CoMF[T[_[_]], A, B] = CoEnv[A, MapFunc[T, ?], B]
+
+  type TCoMF[T[_[_]], A] = T[CoMF[T, A, ?]]
+  type CoMFR[T[_[_]], A] = CoMF[T, A, TCoMF[T, A]]
```
paulp: I am on board with the idea, but these aliases cross some imaginary line in my head where they become too difficult to understand. To me the major benefit of type aliases is common subexpression elimination and limiting the nesting of square bracket expressions. Shortening names is more mixed in its impact. What do you think about these being `CoMapFunc`, `TCoMapFunc`, and `CoMapFuncR` ?

connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -50,41 +50,50 @@ sealed abstract class Ternary[T[_[_]], A] extends MapFunc[T, A] {
...
+
+  private def makeCoMF[T[_[_]], A, B](mf: A \/ MapFunc[T, B]): CoMF[T, A, B] =
+    CoEnv[A, MapFunc[T, ?], B](mf)
+
+  private def makeCoMFR[T[_[_]], A](mf: A \/ MapFunc[T, TCoMF[T, A]]): CoMFR[T, A] =
+    makeCoMF[T, A, TCoMF[T, A]](mf)
+
   /** Returns a List that maps element-by-element to a MapFunc array. If we
```
paulp: And then I might call these methods `coMapFunc` and `coMapFuncR`.

connector/src/main/scala/quasar/qscript/MapFunc.scala
alissapajer: Those names work for me. What matters to me is that I can easily easily see that the `CoEnv[A, MapFunc[T, ?], T[CoEnv[A, MapFunc[T, ?], ?]]]` in one place is the same as the `CoEnv[A, MapFunc[T, ?], T[CoEnv[A, MapFunc[T, ?], ?]]]` in some other place, and doesn't vary in some slight way.

paulp: Right, that's what I mean by common subexpression elimination (where subexpression can be the whole expression.)

alissapajer:
@paulp changes pushed

paulp:
Looks good if travis approves.

Issue #1607, opened 2016-10-19
CB: Enable remaining query integration tests


drostron:
This is a bit of a catch all and will likely spawn new issues as additional tests are being enabled.

Test | Status
--- | ---
caseVariations.test | distinct/ordering bug (ordering lost on reduce step)
caseWithGroup.test | distinct/ordering bug
citiesByLargestZip.test | distinct/ordering bug
different-flattens.test | #1951 qscript provenance bug
flattenArrayLeftWithUnflattened.test | #2360 CB RuntimeException
flattenArrayWithUnflattened.test | #2360 CB RuntimeException
flattenGroupedObject.test | qscript bug - unnecessary join
groupByExpression.test | fixed pending merge of #2106 
groupByFlatten.test | qscript bug - unnecessary join
joins/3WayJoin.test | unimplemented EquiJoin
joins/constantOnLeft.test | unimplemented EquiJoin
joins/crossJoin.test | unimplemented EquiJoin
joins/crossJoinWithOneSidedConditions.test | unimplemented EquiJoin
joins/groupedJoin.test | unimplemented EquiJoin
joins/innerEquiJoin.test | unimplemented EquiJoin
joins/innerEquiJoinWithWildcard.test | unimplemented EquiJoin
joins/joinSubSelects.test | unimplemented EquiJoin
joins/joinWithComplexCondition.test | unimplemented EquiJoin
joins/joinWithMultipleConditions.test | unimplemented EquiJoin
joins/joinWithReversedCondition.test | unimplemented EquiJoin
joins/nonJsJoinCondition.test | unimplemented EquiJoin
joins/orderJoinByAlias.test | unimplemented EquiJoin
joins/sameFieldName.test | unimplemented EquiJoin
joins/symmetricNonJsJoinCondition.test | unimplemented EquiJoin
locationCount.test | qscript bug - unnecessary join
mongodb/distinctStar.test | NA
mongodb/fakeObjectId.test | NA
mongodb/fieldOrder.test | NA
mongodb/filterOnContains.test | NA
mongodb/projectIndexFromGroup.test | NA
mongodb/selectWildcardAndField.test | NA
mongodb/unshiftAggregation.test | NA
multistageFlatten.test | qscript bug - unnecessary join
multitypeFlatten.test | qscript bug - unnecessary join
null/nullTestExprsWithMissing.test | returning incorrect results
projectFieldWithFlatten.test | investigation required - likely CB bug
[...]

PR #1643, opened 2016-10-25
CB: Enable additional query integration tests

drostron:
Towards #1607

### DIFF
--- a/it/src/main/resources/tests/letInSelectProjection.test
+++ b/it/src/main/resources/tests/letInSelectProjection.test
@@ -7,7 +7,7 @@
 
     "data": ["cars.data", "smallZips.data"],
 
-    "query": "select (zips := select * from cars; select * from zips) as value from smallZips",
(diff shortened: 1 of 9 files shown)


it/src/main/resources/tests/letInSelectProjection.test
```diff
@@ -2,13 +2,12 @@
...
-        "postgresql": "pending",
-        "couchbase":  "skip"
+        "postgresql": "pending"
     },
 
-    "data": ["cars.data", "zips.data"],
+    "data": ["cars.data", "smallZips.data"],
 
-    "query": "select (zips := select * from cars; select * from zips) as value from zips",
+    "query": "select (zips := select * from cars; select * from zips) as value from smallZips",
```
sellout: You have to be careful with this one – I think the goal is to make sure the names are scoped properly, so you need to make sure the same name is used in all three places.

drostron: Ahh, OK.

drostron:
Ready for another look.

PR #1640, opened 2016-10-25
Fix merging of QScript Reduce nodes.

sellout:
Fixes #1637.

PR #1645, opened 2016-10-26
Fix MarkLogic issue with Reduce initialization

wemrysi:
Fixes an issue related to how `Reduce` nodes are planned in MarkLogic and enables a handful of working reduction/grouping tests thanks to #1640.


marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -74,6 +74,17 @@ object qscript {
       }
     }
 
+  // qscript:combine-apply($fns as (function(item()) as item())*) as function(item()) as item()*
+  def combineApply[F[_]: PrologW]: F[FunctionDecl1] =
+    qs.name("combine-apply").qn[F] map { fname =>
+      declare(fname)(
+        $("fns") as SequenceType("(function(item()) as item())*")
+      ).as(SequenceType("function(item()) as item()*")) { fns: XQuery =>
```
alissapajer: So it looks like `combineApply` and `zipApply` are identical up to this point? Could you factor this first bit out into another function?

alissapajer: And maybe other functions in this file could use this new function too?

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -74,6 +74,17 @@ object qscript {
...
+        val (f, x) = ("$f", "$x")
+        func(x) { fn.map(func(f) { f.xqy fnapply x.xqy }, fns) }
```
alissapajer: Discerning the difference between this and `zipApply`. In `zipApply` you index into `fns` from `1... length(fns)` whereas here you simply don't? Is there another key difference that I'm missing?

alissapajer:
@wemrysi random question: Do you think that all the work you've done with `XQuery` (for example marklogic/src/main/scala/quasar/physical/marklogic/xquery/expr.scala) would be useful as a standalone library for others querying marklogic through Scala?

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
wemrysi: `zipApply` applies a sequence of functions to a sequence of arguments, pointwise, producing a sequence of results whereas `combineApply` applies a sequence of functions to a single argument, producing a sequence of results. So in the former, we're indexing into `fns` and `x`.

wemrysi:
@alissapajer Yeah, I think so, probably the stuff in `quasar.physical.marklogic.xcc` as well. I think extracting these bits out would be worthwhile once they're a bit more mature (some of the expressions, for example, will probably become nodes in the `XQuery` type and we might want to convert the `XQuery` type itself to fixpoint style).

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
wemrysi: @alissapajer What do you think of

```scala
def combineApply[F[_]: PrologW]: F[FunctionDecl1] =
  qs.declare("combine-apply") map (_(
    $("fns") as ST("(function(item()) as item())*")
  ).as(ST("function(item()) as item()*")) { fns =>
    val (f, x) = ("$f", "$x")
    func(x) { fn.map(func(f) { f.xqy fnapply x.xqy }, fns) }
  })
```

it gets rid of a bit of boilerplate and (mostly) preserves the common `declare <name> (args...) as ...` syntax (which mimics `XQuery` and works for many arities). I also tried defining some helpers for specific arities like `declareUnary` and `qs.unaryFn` which worked, but then became inconsistent when you needed more arities.

`ST` is an alias for `SequenceType` just for brevity. `SequenceType` itself could use some attention when time permits, maybe to provide more structure and maybe some syntax for the various qualifiers (zero-or-one, zero-or-many, etc).

Does this improve the current situation? Did you have something else in mind?

wemrysi: @alissapajer Aye, that is feasible, but at the expense of uniformity, so I'm having a hard time deciding whether it is an improvement.

The separate type to hold the arguments seems a bit superfluous they're only ever used at the declaration site and can be recovered from the `FunctionDecl` if needed.

The helper for declaring a unary function reduces some repetition, but then we'd probably want one for binary functions as well (and maybe other arities?), which begins to defeat the purpose of the uniform syntax of `declare` (which works for many arities).

```scala
val combineApply = qs.unaryFn(
  "combine-apply",
  "fns",
  ST("(function(item()) as item())*"),
  ST("function(item()) as item()*")
) { fns => 
  ... 
}

vs

val combineApply = qs.declare("combine-apply") map (_(
  $("fns") as ST("(function(item()) as item())*")
).as(ST("function(item()) as item()*")) { fns =>
  ...
})

val deleteField = qs.binaryFn(
  "delete-field",
  "src",
  ST("element()"),
  "field",
  ST("xs:QName")
  ST("element()")
) { (src, field) =>
  ...
}

vs

val deleteField = qs.declare("delete-field") map (_(
  $("src") as ST("element()"),
  $("field") as ST("xs:QName")
).as(ST("element()")) { (src, field) =>
  ...
}
```
we could use the `$("name") as ST("type")` syntax for the helper functions also, but then they become even less compelling, I think.

In the way of background, part of the goal of the `declare` syntax is that it mimics `XQuery`, so to stick with the `combine-apply` example

XQuery
```xquery
declare function qscript:combine-apply(
  $fns as (function(item()) as item())*
) as function(item()) as item()* {
  function($x) { fn:map(function($f) { $f($x) }, $fns) }
};
```

Scala
```scala
qs.declare("combine-apply") map (_(
  $("fns") as ST("(function(item()) as item())*")
).as(ST("function(item()) as item()*")) { fns =>
  val (f, x) = ("$f", "$x")
  func(x) { fn.map(func(f) { f.xqy fnapply x.xqy }, fns) }
})
```
this property has been fairly useful to me in authoring functions as it makes it reasonably possible (once you get used to seeing through some of the Scala) to visualize what the function will translate to in XQuery and, probably more importantly, port an XQuery implementation to Scala (something like an `xqy"<some xquery>"` string interpolator would be even better, but that's a ways off). So my concern is that we'd be trading the similarity with XQuery to eliminate some of the repetition.

Sorry if I've missed the point of your suggestions, having stared at this for the last couple months I'm sure there is some tunnel-vision at play. Are there other advantages I've overlooked? Do you think maintaining the similarity is beneficial? A vain pursuit? I really appreciate the feedback.

alissapajer: @wemrysi and I chatted about this. In the end we decided that this can be merged after the change to use `qs.declare` is made. That way the bug fix can get in ASAP, and some of the boilerplate/duplication is removed. 

And then a future cleanup project can entail removing further duplication, possibly with helper functions specializing on each arity. Thinking about it now, another option would be to use `shapeless.nat` to declare the arity, and then you just need one helper function. Though that would involve more compilation overhead, so that cost would have to be weighed.

wemrysi:
@alissapajer I've switched to using the `declare` syntax where possible, back in your court. Let me know if there is anything else.

alissapajer:
LGTM - merging!

PR #1646, opened 2016-10-26
remove LP Fix constructors

alissapajer:
Broken into one commit per constructor.


frontend/src/test/scala/quasar/helpers.scala
```diff
@@ -0,0 +1,44 @@
...
+package quasar.frontend
+
+import quasar.contrib.pathy._
+import quasar.Predef._
+import quasar.{LogicalPlan => LP, _}
+
+import matryoshka.Fix
+import shapeless.Nat
+
+trait LogicalPlanHelpers {
```
alissapajer: So, this was the easiest thing to do to get the tests passing, but I'm not completely sold on this approach.

(1) I don't like the naming - maybe `letF` would be better than `fixLet`?
(2) Outside of this PR I'd like to remove the ` implicit def toFix[F[_]](unFixed: F[Fix[F]]): Fix[F] = Fix(unFixed)`. To do that, we'll need some sort of machinery along these lines. What I was experimenting with was:
```
 46   def fixUnary(func: UnaryFunc)(v1: Fix[LP]): Fix[LP] =                                                      
 47     Fix[LP](func[Fix[LP]](v1))                                                                               
 48 
 49   def fixBinary(func: BinaryFunc)(v1: Fix[LP], v2: Fix[LP]): Fix[LP] =                                       
 50     Fix[LP](func[Fix[LP]](v1, v2))
 51 
 52   def fixTernary(func: TernaryFunc)(v1: Fix[LP], v2: Fix[LP], v3: Fix[LP]): Fix[LP] =                        
 53     Fix[LP](func[Fix[LP]](v1, v2, v3)) 
```
but I'm sure there are other approaches.

alissapajer: So, the test helpers also does not need to have `Fix` hard-coded; nstead, we could create a `LogicalPlanHelpers(Fix)` so that we can easily swap in something else for `Fix`.

sellout: 5-minute window …

Yeah, that implicit was a failed experiment. Doesn't even simplify things, really.

If you look at the wiki page, you can see my proposal for non-specific fixed-point helpers. That wiki page is very open to discussion.

Also, you can see on the wiki, that I've been preferring a `R` suffix, as `F` is commonly used for functors and `T` for transformers.

alissapajer: For the `UnaryFunc`, `BinaryFunc` and `TernaryFunc` I'd like to avoid having a function here for _each_ function in those categories (read 10s, maybe 100+), which is what drove me to `fixUnary` etc.

alissapajer:
Failed with a real errors in the render specs - will fix.

alissapajer:
Realized I should have made a ticket for this. So I did now.

Fixes #1651.

alissapajer:
And the ticket for removing the `implicit` embedding is #1652.

frontend/src/test/scala/quasar/helpers.scala
wemrysi: @alissapajer I like your suggestion, we use a similar technique in `fs` for lifting types into `Free` and also the `fixpoint` type in `jscore` and `mongodb` that @mossprescott mentioned.

sellout: So, I think this can move to the main code, not test, so we can take advantage of both the already-fixed values as well as perhaps some compile-time benefit by only having a single `val lpr = new LogicalPlanR[Fix]` in a file.

I was also thinking that perhaps the names in `LogicalPlanR` should be simply `constant` rather than `constantR` (i.e., departing from what I suggested in the wiki), then we can can avoid importing and use `lpr.constant(…)`, etc. This also makes it easy to have, say `val lpf = new LogicalPlanR[Fix]` and `val lpn = new LogicalPlanR[Nu]` in the same file.

Finally, why not just use `def invoke[N <: Nat](…)` here rather than separating into `unary`/`binary`/`ternary`? I feel like if we separate them, we should separate the case classes as well. Also, the functions already have an apply method so, while not pre-fixed, you can write `ObjectProject(a, b).embed` directly without wrapping in `invoke`.

it/src/test/scala/quasar/physical/mongodb/MongoDbStdLibSpec.scala
```diff
@@ -39,7 +40,7 @@ import shapeless.{Nat}
 /** Test the implementation of the standard library for one of MongoDb's
   * evaluators.
   */
-abstract class MongoDbStdLibSpec extends StdLibSpec {
+abstract class MongoDbStdLibSpec extends StdLibSpec with LogicalPlanHelpers {
```
sellout: I know I’ve read something about it, but what is the benefit here of `with LogicalPlanHelpers` over
```scala
val lpr = new LogicalPlanHelpers
import lpr._
```
I know the former will exports the members of `LogicalPlanHelpers` as part of `MongoDbStdLibSpec`, but does that help us here, or is there something else?

sellout: NB: I have the `import` here just for the equivalent behavior – as I mentioned above, I’m thinking of disrecommending the import that’s currently in the wiki page.

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -105,7 +105,7 @@ object Sql {
     extends Sql[A]
 @Lenses final case class Unop[A] private[sql] (expr: A, op: UnaryOperator) extends Sql[A]
 @Lenses final case class Ident[A] private[sql] (name: String) extends Sql[A]
-@Lenses final case class InvokeFunction[A] private[sql] (name: String, args: List[A])
+@Lenses final case class Invokeunction[A] private[sql] (name: String, args: List[A])
```
alissapajer: Yeah, looks this was replaced globally. I'll fix it!

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -57,15 +57,15 @@ object view {
       } yield h
 
     def openView(f: AFile, off: Natural, lim: Option[Positive]): FileSystemErrT[Free[S, ?], ReadHandle] = {
-      val readLP = addOffsetLimit(LogicalPlan.Read(f), off, lim)
+      val readLP = addOffsetLimit(Fix(LP.Read[Fix[LP]](f)), off, lim)
```
alissapajer: @sellout I think I'm on board with your suggestions here. Summarizing, to make sure we're on the same page: 

`LogicalPlanHelpers` will move to the main code and be renamed `LogicalPlanR`. It will be `class LogicalPlanR[T[_[_]]: Corecursive]`, and then we can have `val lpr = new LogicalPlanR[Fix]`, like you said.

Then all the places (in main code and in tests) will be calling `lpr.whatever`, like in the example right above. Meaning the fixed-point wrappers will be used everywhere. I like when something can be used everywhere; that sits well with me :)

I'm still torn between naming the functions with an `R` suffix or not, for example `constant` versus `constantR`. If we _don't_ recommend `import lpr._` then we'd be calling `lpr.constant` versus `lpr.constantR`. So I  suppose the `R` in the second example there is redundant... Personally I find imports like `import lpr._` confusing, and prefer just to use the prefix, because then it's obvious where something is importing from.

sellout: Yeah, the lack of `R` is explicitly to discourage the import. And the reason to discourage the import, beyond just generally preferring to see where things come from, is to make it easy to use additional fixed-points alongside the original one. Getting stuck on `Fix` is something that has been a pain, so I’m trying to discourage that as much as possible.

sellout: And yeah, everything else sounds good.

it/src/test/scala/quasar/physical/mongodb/MongoDbStdLibSpec.scala
alissapajer: I think you're right that importing (or just using the `val` for) `LogicalPlanR` is better here, because we are not attempting to create a dependency. We just want access to the functions in `LogicalPlanR`, that's it.

alissapajer:
@sellout @wemrysi I have made the discussed changes.

@sellout let me know if there is more (or less) I can do to make the merge with your work easier.

frontend/src/test/scala/quasar/sql/fixpoint.scala
```diff
@@ -21,6 +21,8 @@ import quasar.Predef._
 import matryoshka._
 
 package object fixpoint {
+  val lpf = new quasar.frontend.LogicalPlanR[Fix]
```
alissapajer: I put the test `lpf` in here, but it shouldn't really be in the `sql` package; it should be in the `frontend` package because that's where `LogicalPlan` lives. Is this file on its way into the sql project, and just living in the frontend project for the time being?

sellout: It can be here for now, as long as there are no references to `lpf` outside of the sql package. This file is exactly what should become `class SqlR[T[_[_]]]` anyway, at which point `lpf` becomes `new LogicalPlanR[T]`.

connector/src/test/scala/quasar/fs/QueryFileSpec.scala
```diff
@@ -22,17 +22,17 @@ import quasar.{Data, DataArbitrary}
...
 import pathy.Path._
 import pathy.scalacheck.PathyArbitrary._
 import scalaz._, Scalaz._
 import scalaz.scalacheck.ScalazArbitrary._
 
-class QueryFileSpec extends quasar.Qspec with FileSystemFixture with LogicalPlanHelpers {
+class QueryFileSpec extends quasar.Qspec with FileSystemFixture {
   import InMemory._, FileSystemError._, PathError._, DataArbitrary._
   import query._, transforms.ExecM
+  import quasar.sql.fixpoint.lpf
```
alissapajer: Also, I'd like to have just one place for the `lpf` test instance, if possible.

frontend/src/test/scala/quasar/sql/fixpoint.scala
sellout: So, in my upcoming changes, I reference it a couple places – I didn’t try to make one central value to access. But yeah, it _seems_ like it should live in the frontend package, but since that’s not really possible (because of the singular `package object` problem), putting it in some object called `fixpoint` or whatever in that package seems fine.

alissapajer: @sellout Ok. It will probably just be a new file. For file naming, which are we preferring:
`/frontend/src/test/scala/quasar/frontend/fixpoint.scala`
or 
`/frontend/src/test/scala/quasar/fixpoint.scala`
?

sellout: The former – I’m not about to give up on Ensime yet (at least @rabbitonweb and I rely on it), and even if I were, we need to get everything in the right packages first before we change any file structure.

sellout: Yeah, it can’t live there because that’s not in `test`, and we don’t ever write `Fix` in `main` code anymore 😉

alissapajer: Oh haha yeah, I would have come to that conclusion like 10 seconds later :)

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -302,13 +302,13 @@ trait Compiler[F[_]] {
           stepName <- CompilerState.freshName("tmp")
           current  <- current
           bc        = relations match {
-            case ExprRelationAST(_, name)        => BindingContext(Map(name -> LP.Free(stepName)))
-            case TableRelationAST(_, Some(name)) => BindingContext(Map(name -> LP.Free(stepName)))
-            case id @ IdentRelationAST(_, _) => BindingContext(Map(id.aliasName -> LP.Free(stepName)))
+            case ExprRelationAST(_, name)        => BindingContext(Map(name -> Fix(LP.Free(stepName))))
```
wemrysi: Is it possible to leverage `LogicalPlanR` here and in elsewhere in this file?

alissapajer: Yes indeed! I'll do that right now.

sellout: Yeah, check the PR diff for anywhere that added `Fix` or `.embed`. The latter is fine if it’s not pervasive enough to justify `LogicalPlanR`, but the former (including `.unFix`) shouldn’t be there (except for maybe some type annotations).

alissapajer: Change pushed.

alissapajer: I'm going through the diff now - I'll post back when I've complete that.

alissapajer: Done.

alissapajer:
@sellout @wemrysi I've pushed everything I have planned to push. Let me know if there is anything else!

wemrysi:
@alissapajer LGTM, nice work!

frontend/src/main/scala/quasar/fixpoint.scala
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package quasar.frontend
+
+import matryoshka.Fix
+
+object fixpoint {
+  // TODO remove any dependency on `Fix` from main
+  val lpf = new LogicalPlanR[Fix]
+}
```
sellout: This shouldn’t be in `main`.

frontend/src/main/scala/quasar/optimizer.scala
```diff
@@ -35,28 +35,29 @@ object Optimizer {
   import quasar.std.StdLib._
   import set._
   import structural._
+  import quasar.frontend.fixpoint.lpf
```
sellout: Ok, so, this should use use
```scala
val lpr = new LogicalPlanR[Fix]
```
for now (note the `r` not `f`). In my changes, `Optimizer` is parameterized over `T` anyway, so this will need to create a `LogicalPlanR` for whatever `T` that is.

frontend/src/main/scala/quasar/sql/JoinDir.scala
```diff
@@ -24,9 +24,10 @@ import matryoshka.Fix
 
 sealed abstract class JoinDir(val name: String) {
   import structural.ObjectProject
+  import quasar.frontend.fixpoint.lpf
```
sellout: I’ve also parameterized `const` (and `projectFrom`) over `T`, so it’s going to need it’s own `lpr`, too – but locally I just did `constant(data).embed`, since there was pretty minimal need for it.

sellout:
Ok, a few more comments – mostly just gotta get rid of `lpf` from `main`.

frontend/src/main/scala/quasar/fixpoint.scala
alissapajer: Dangit. I really did not mean to put it there...

frontend/src/main/scala/quasar/optimizer.scala
alissapajer: @sellout are you saying in just this file I should rename it to `lpr`, or can I just use `lpr` everywhere in place of `lpf`?

frontend/src/main/scala/quasar/fixpoint.scala
alissapajer: Oh and I also managed to name it wrong, without the second "frontend" in the fie path...

alissapajer: @sellout no wait, this is the old one, I put this in main on purpose, because we still use `Fix` in main so this lets us clean it up.

alissapajer: Otherwise we have to either wrap `Fix` everywhere or call `.embed` but I thought this was nicer because it pointed out the places we're using `Fix` to make future search/replace easier.

frontend/src/main/scala/quasar/optimizer.scala
sellout: Well, you can’t use `fixpoint.lpf` in here, because this is in main, and you shouldn’t define `lpf` in main. It’s not about making the merge easier, it’s about reducing references to `Fix` in main.

In my local changes currently, I have
```scala
val lpr = new LogicalPlanR[T]
import lpr._
```
and since we want to ditch the import anyway, there’s going to be merge stuff to clean up. But yeah, since having to specialize on `Fix` is just a temporary bug, renaming `lpf` to `lpr` (at least in all the main code) would be good.

alissapajer: I didn't have to make a `LogicalPlanR[Fix]` in main, but I did because I thought it would be easier to them  (in your PR) remove all reference to `Fix`.

alissapajer: OH I think I just got what you're saying! The `f` in `lpf` refers to `Fix` and we're not going to have `Fix` in main very soon, so I should name `LogicalPlanR[Fix]` to `lpr` just because in your PR it will be `LoticalPlanR[T]`, so the `f` would be silly.

sellout: Well, I’d rather see the file containing the fixpoint object go away, and have us call `val lpr = new LogicalPlan[Fix]` instead of importing it. But, I can make that happen in my PR, so just renaming it is fine.

alissapajer: I should be able to globally search/replace the import with the `val`. So I'll do that and then remove the `fixpoint.scala` file from main.

alissapajer:
@sellout `lpf` deleted from main - we now use a `val lpr = new LogicalPlanR[Fix]` in each file in main that requires it.

alissapajer:
Back in your court again! :)

sellout:
Cool. Will merge when green.

PR #1647, opened 2016-10-26
Couchbase: Increase query timeout

drostron:
It's tough to say whether or not this will decrease the number of timeouts we are seeing since we don't know what state Couchbase server is getting into on Travis builds at this point. Despite that I think it's worth a try.


drostron:
This failure leads me to think that a longer timeout is not sufficient for the Couchbase server to recover from whatever state it gets into leading to timeouts. Closing.

Issue #1601, opened 2016-10-18
CB: Additional QScript terms


drostron:
- [x] Arbitrary #1653 
- [x] EJson: Meta, Byte, Char #1686 
- [x] Const(Root) #1602
- [x] TimeOfDay, Extract* #1686 
- [x] IfUndefined #1696
- ~~Coalesce~~ #1604
- ~~DupMapKeys, DupArrayIndices, ZipMapKeys, ZipArrayIndices~~ #1692
- [x] BucketField, BucketIndex #1602

sellout:
You don’t have to provide a real implementation for `Const(Root)` or `BucketField`/`BucketIndex`. If you look at another connector, you’ll see that those (as well as two other cases that depend on your flavor of QScript) are implemented with something like `InternalError("unreachable").left` (depending on the monad of the planner).

I’m also going to remove `Coalesce` – that’s another SQL-ism that should be eliminated prior to LogicalPlan.

Issue #1621, opened 2016-10-21
CB: Handle error responses from CB


drostron:
Currently some error responses from CB are ignored. These should be intercepted and turned into appropriate Quasar errors.

zeeiyerWork:
Also look at slamdata/slamdata#1457

nicflores:
Fixed in PR #2112. Closing.

PR #1653, opened 2016-10-27
CB: Impls for ConcatMaps and Arbitrary

drostron:
Addresses: #1600 and #1601

### DIFF
--- a/couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
+++ b/couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
@@ -223,15 +223,17 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
     case ConcatArrays(a1, a2)                     =>
       partialQueryString(s"array_concat(${n1ql(a1)}, ${n1ql(a2)})").point[M]
     case ConcatMaps(a1, a2)                       =>
-      val a1N1ql = n1ql(a1)
-      val a2N1ql = n1ql(a2)
+      val a1N1ql  = n1ql(a1)
+      val a2N1ql  = n1ql(a2)
+      val n1qlStr = s"object_concat(${n1ql(a1)}, ${n1ql(a2)})"
       prtell[M](Vector(Detail(
         "N1QL ConcatMaps",
         s"""  a1:   $a1N1ql
            |  a2:   $a2N1ql
-           |  n1ql: ???""".stripMargin('|')
-      ))) *>
-      unimplementedP("ConcatMaps")
+           |  n1ql: $n1qlStr""".stripMargin('|')
+      ))).as(
+        partialQueryString(n1qlStr)
+      )
     case ProjectField(PartialQueryString(a1), a2) =>
       val a2N1ql  = n1ql(a2)
       val n1qlStr = s"$a1.[$a2N1ql]"
--- a/couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
+++ b/couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
@@ -219,9 +219,29 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
         partialQueryString(n1qlStr)
       )
     case MakeMap(a1, a2)                          =>
-      partialQueryString(s"""object_add({}, ${n1ql(a1)}, ${n1ql(a2)})""").point[M]
+      val a1N1ql  = n1ql(a1)
+      val a2N1ql  = n1ql(a2)
+      val n1qlStr = s"""object_add({}, ${n1ql(a1)}, ${n1ql(a2)})"""
+      prtell[M](Vector(Detail(
+        "N1QL MakeMap",
+        s"""  a1:   $a1N1ql
+           |  a2:   $a2N1ql
+           |  n1ql: $n1qlStr""".stripMargin('|')
+      ))).as(
+        partialQueryString(n1qlStr)
+      )
     case ConcatArrays(a1, a2)                     =>
-      partialQueryString(s"array_concat(${n1ql(a1)}, ${n1ql(a2)})").point[M]
+      val a1N1ql  = n1ql(a1)
+      val a2N1ql  = n1ql(a2)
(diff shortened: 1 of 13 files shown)


couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -219,9 +219,29 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
         partialQueryString(n1qlStr)
       )
     case MakeMap(a1, a2)                          =>
-      partialQueryString(s"""object_add({}, ${n1ql(a1)}, ${n1ql(a2)})""").point[M]
+      val a1N1ql  = n1ql(a1)
+      val a2N1ql  = n1ql(a2)
+      val n1qlStr = s"""object_add({}, ${n1ql(a1)}, ${n1ql(a2)})"""
```
mossprescott: Should be `s"""object_add({}, $a1n1ql, $a2N1ql)`, or else remove the `val`s.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -219,9 +219,29 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
...
+           |  a2:   $a2N1ql
+           |  n1ql: $n1qlStr""".stripMargin('|')
+      ))).as(
+        partialQueryString(n1qlStr)
+      )
     case ConcatArrays(a1, a2)                     =>
-      partialQueryString(s"array_concat(${n1ql(a1)}, ${n1ql(a2)})").point[M]
+      val a1N1ql  = n1ql(a1)
+      val a2N1ql  = n1ql(a2)
+      val n1qlStr = s"array_concat(${n1ql(a1)}, ${n1ql(a2)})"
```
mossprescott: Similar. I guess you need the `val`s for logging.

it/src/main/resources/tests/boulderLikePop.test
```diff
@@ -13,6 +12,8 @@
 
   "predicate": "containsExactly",
 
+  "ignoreFieldOrder": true,
```
mossprescott: Why ignore the field order here? This looks correct; I think we only want to use that flag in cases where we know MongoDB gets it wrong.

mossprescott:
Needs changelog, by the way

mossprescott:
Not really an issue for this PR, but the way PhaseResults are being produced here isn't useful, I don’t think (without being able to run the planner I haven’t actually looked at the result).

If you run in the REPL with `debug = 2`, each PhaseResult should be a verbose but comprehensible summary of one stage of compilation. It looks like you're producing one result per node. I guess that's ok if it’s useful for debugging a the moment, but it's something we'll want to straighten out at some point. Have you actually been using the log?

jdegoes:
@drostron The other thing is we really need to check on mount or mount access that the version of Couchbase used is sufficient for the features we're using. Probably not an immediate priority since Couchbase will probably generate some error (assuming we propagate it).

drostron:
@mossprescott I've been using PhaseResults frequently for debugging at the repl. What's there was a remedy for ad-hoc printlns. Moving towards N1QL generation as a single PhaseResult sounds good, #1656.

drostron:
@jdegoes sounds good. Once #1621 is in place we'll propagate errors from Couchbase more consistently. I also created #1657.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
drostron: Right, just missed updating the n1qlStrs after PhaseResult additions.

it/src/main/resources/tests/boulderLikePop.test
drostron: Up until now, I was assuming that N1QL offered the ability to specify result set field order but a further look finds this is not the case. e.g.

```
cbq> select v.z, v.y from (select value { "z": 1, "y": 2 }) v;
{
    ⋮
    "results": [
        {
            "y": 2,
            "z": 1
        }
    ],
    ⋮
}
```

drostron:
I was hoping a test using `zips.data` would be possible on Travis, but it is timing out consistently. That test is pretty tailored to `zips.data` so I'll switch it out for another.

drostron:
Ready for another look. We need to decide how to handle N1QL not supporting result set field ordering.

drostron:
I propose we find a solution to the field order concern in another PR via #1668.

couchbase/src/test/scala/quasar/physical/couchbase/BasicQueryEnablementSpec.scala
```diff
@@ -94,17 +94,13 @@ class BasicQueryEnablementSpec
...
       "select name from `beer-sample` limit 1",
-      """select value v from (select value _4 from (select value _2[0:_1[0]] from (select value (select value {"name": _5.["name"]} from (select value ifmissing(v.`value`, v) from `beer-sample` v) as _5) from (select value (select value [])) as _0) as _2 let _1 = (select value 1 let _6 = (select value []))) as _3 unnest _3 _4) as v""")
+      """select value v from (select value _4 from (select value _2[0:least(array_length(_2), _1[0])] from (select value (select value {"name": _5.["name"]} from (select value ifmissing(v.`value`, v) from `beer-sample` v) as _5) from (select value (select value [])) as _0) as _2 let _1 = (select value 1 let _6 = (select value []))) as _3 unnest _3 _4) as v""")
 
     testSql2ToN1qlPending(
       "select name from `beer-sample` order by name",
       pending("#1545"))
 
     testSql2ToN1ql(
-      """select address from `beer-sample` where name = "Brasserie de Silly"""",
```
drostron: The tests in this file are fragile and intended for removal as they are covered by query integration tests. Currently it's too easy for the generated N1QL strings to change, still be valid and correct, yet trigger a test failure. There is however opportunity and value for unit style tests against a richer N1QL representation when it is available.

drostron: Agreed, and looking forward to it. The current representation was a path towards getting everything initially stood-up and we'll have increased motivation for the richer representation as we require things like N1QL specific transformations or finer grained tests.

alissapajer:
LGTM will merge when jenkins gives green.

PR #1660, opened 2016-10-28
Add support for `Avg` in spark-core

rabbitonweb:


sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -127,6 +128,36 @@ class PlannerSpec extends quasar.Qspec with QScriptHelpers with DisjunctionMatch
...
+          case rdd =>
+            val results = rdd.collect
+            results.size must_== 2
+            results(1) must_== Data.Dec(28)
+            results(0) must_== Data.Dec(23)
+        }
+        sc.stop
+      }).run.unsafePerformSync
+      ok
+    }
```
mossprescott: Does this test actually fail if the result is wrong? It isn't clear to me and I can’t make the existing tests fail locally by changing their assertions.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -127,6 +128,36 @@ class PlannerSpec extends quasar.Qspec with QScriptHelpers with DisjunctionMatch
...
+          Data.Obj(ListMap() + ("age" -> Data.Int(24)) + ("country" -> Data.Str("Poland"))),
+          Data.Obj(ListMap() + ("age" -> Data.Int(32)) + ("country" -> Data.Str("Poland"))),
+          Data.Obj(ListMap() + ("age" -> Data.Int(28)) + ("country" -> Data.Str("Poland"))),
+          Data.Obj(ListMap() + ("age" -> Data.Int(23)) + ("country" -> Data.Str("US")))
+        ))
+
+        def bucket: FreeMap = ProjectFieldR(HoleF, StrLit("country"))
+        def reducers: List[ReduceFunc[FreeMap]] = List(Avg(ProjectFieldR(HoleF, StrLit("age"))))
+        def repair: Free[MapFunc, ReduceIndex] = Free.point(ReduceIndex(0))
+        val reduce = Reduce(src, bucket, reducers, repair)
```
mossprescott: There are many interesting cases not being tested here:

- Dec values
- mixed `Int` and `Dec` values
- non-numeric values, with or without numeric values also present
- "undefined" values (missing field, I guess)
- no values/buckets
- more or less than 2 buckets
- more than one reducer
- non-trivial repair

Some of these cases could probably be covered by simply switching to a property-based test, while others might be covered in some other way (but I don't see it), or might call for a whole different kind of test.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -209,7 +216,13 @@ object Planner {
                 src.map(d => (partitioner(d), Data.Arr(trans.map(_(d))) : Data))
                   .reduceByKey(merge(_,_, reducersFuncs))
                   .map {
-                  case (k, v) => repair(v)
+                  case (k, Data.Arr(vs)) =>
+                    val v = Zip[List].zipWith(vs, reducers) {
+                      case (Data.Arr(List(Data.Dec(sum), Data.Int(count))), Avg(_)) => Data.Dec(sum / BigDecimal(count))
```
wemrysi: While theoretically correct, this implementation isn't numerically stable. An incremental mean [like this one](http://math.stackexchange.com/questions/106700/incremental-averageing#106720) will generally give more precise results.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
rabbitonweb: I've deleted my previous LONG comments on this one so I will provide for now shorter versions of the original (this apply to everything below):

Test will run only if QUASAR_LOCAL env variable is set, like `QUASAR_SPARK_LOCAL={"sparklocal":{"connectionUri":"local[*]|/Users/rabbit/"}}`

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
rabbitonweb: I can add those tests. Those will be integration tests, but we can run them just in case

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
rabbitonweb: If I understand correctly, you mean other formula then simple `sum / count`

rabbitonweb: No, wait - you actually mean completely different way to calculate the average. Incrementally. But what is wrong with current approach?

rabbitonweb: Does it mean that `sum / count` is less precise  ?

wemrysi: Usually you have to worry about overflowing the sum for large datasets or losing precision on the division when the mean is very small and the dataset is very large. But, I forgot you're using `BigDecimal` and `BigInt` here, do you know if they suffer from these issues? I think the sum should be ok but I'm not sure about the divide (I've never used either of these types).

So this may not be an issue, give the types you're using.

rabbitonweb:
RDY

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -71,448 +72,570 @@ class PlannerSpec extends quasar.Qspec with QScriptHelpers with DisjunctionMatch
       }).run.unsafePerformSync
       ok
     }
+    "core" should {
+      "map" in {
+        newSc.map ( sc => {
```
wemrysi: The running of these tests is a bit awkward (running the `Task` and then returning `ok` is a bit confusing) and one always has to remember to `stop` the spark context everywhere, how about introducing a new function similar to this (I haven't tried compiling it):
```scala
def withSpark[A: AsResult](f: SparkContext => Task[A]): Result =
  newSc
    .flatMapF(sc => f(sc).onFinish(_ => Task.delay(sc.stop)).map(AsResult(_)))
    .getOrElse(skipped)
    .unsafePerformSync
```
and then using it like so:
```scala
withSpark { sc =>
  ...

  state.eval(sc).run map (_ must beRightDisjunction.like {
    case rdd => ...
  })
}
```

rabbitonweb:
Tests has been cleaned.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -631,13 +563,19 @@ class PlannerSpec extends quasar.Qspec with QScriptHelpers with DisjunctionMatch
...
           }
-          sc.stop
-        }).run.unsafePerformSync
-        ok
+        })
       }
     }
   }
 
+  private def withSpark[T](run: SparkContext => T) = {
```
wemrysi: This is a nice improvement, but I'd still like to see us leveraging `Task` more here as this forces callers to run `Task`s and also will not clean up the spark context if the `run` function throws an exception.

It also still suffers from returning the `ok` which shouldn't be necessary.

rabbitonweb:
And tests are now cleaning resources! :)

PR #1661, opened 2016-10-28
Change the Spark MapFunc algebra from `Data => Data`  to `A => Data`.

sellout:


paulp:
@sellout [See also some of my aspiration to eliminate a lot of this duplication in a standard way](https://github.com/paulp/quasar/blob/precog/jsonfile/src/main/scala/quasar/physical/jsonfile/fs/Fallback.scala#L127-L146).

paulp:
Can anyone comment on why this algebra operates on a function and not on the incoming type directly? Why `M[Data => Data] => (Data => Data)` vs. `M[Data] => Data` (or sub in A as appropriate) ?

jdegoes:
@paulp Oops, sorry about not getting back to this sooner. I believe the difference is that one is an actual (analytical) computation whereas the other is building a description of a computation that can be executed later. If you start crunching numbers inside an algebra, then you'll force the entire computation there, which is not really the goal of Spark's `RDD` and could lead to some serialization problems.

paulp:
There are some pretty severe drawbacks to stitching together a function from peephole views of other functions, like pathologically bad performance. A description of a computation to be executed later sounds good - but I don't think that execution is likely to finish very often if the entire ADT for representing the description is opaque Function1.

jdegoes:
@paulp Agreed, but if you don't want to try serializing all of Matryoshka, then the Spark connector has few choices. An alternative would be to compute an `RDD[Data]`, but that wouldn't gain you any performance (actually you'd lose some) since Spark's `RDD.map` would have to capture all the functions and stuff them in data structures to be later serialized.

The fallback evaluator won't have this performance issue, because its version of `change` would produce not `Data => Data`, but `TransSpec`, which at a higher-level, ultimately becomes a table. Because the overhead of interpreting `TransSpec` is per chunk, which can be made arbitrarily large, it's negligible. Albeit the Precog analytics engine, while quite fast, still has some performance issues caused by its laziness (it needs to eagerly compute chunks using loops & arrays; but right now defers as much computation as possible resulting in a buildup of composed functions from row to value). Though these are easily solved in a future project.

sellout:
Ok, just noticed that this has been sitting here forever. @rabbitonweb, can I get a review?

sellout:
@rabbitonweb *cough*

sellout:
@rabbitonweb Ahem.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/Planner.scala
```diff
@@ -240,30 +224,17 @@ object Planner {
...
-                  case _ => Data.NA
-                }).right
-              }, CoreMap.change))
-
-            dd.map(df => (l, r) => df(Data.Arr(List(l, r))))
-          }
+            CoreMap.changeFreeMap(struct)
+
+          def repairFunc: PlannerError \/ ((Data, Data) => Data) =
+            CoreMap.changeJoinFunc(repair)
```
sellout: This is one of the important parts of the change. Previously, Everything had to be `Data => Data`, which meant that to handle `(Data, Data) => Data`, we had to create a `Data.Arr(List(_, _))`, and then handle it along with a catchall case that resulted in `Data.NA`, but we statically know we have a pair of elements and we throw that info away by wrapping it in some `Data`.

So, with `A => Data`, we can handle the `Data => Data` case (FreeMap), the `(Data, Data) => Data` case (JoinFunc), and the `List[Data] => Data` case (ReduceFunc) using a single implementation while maintaining the static knowledge.

sellout:
Could really use a review before I resolve conflicts …

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/CoreMap.scala
```diff
@@ -16,28 +16,49 @@
...
-import org.threeten.bp.{Instant, ZoneOffset}
 import matryoshka.{Hole => _, _}, Recursive.ops._
+import org.threeten.bp.{Instant, ZoneOffset}
 import scalaz.{Divide => _, _}, Scalaz._
 
 object CoreMap extends Serializable {
 
   private val undefined = Data.NA
 
+  def changeFreeMap[T[_[_]]: Recursive](f: FreeMap[T])
```
rabbitonweb: `changeFreeMap`, `changeJoiFunc`, `changeReduceFunc` - shouldn't they go to their own package? changeFreeMap is used also in `Planner`

sellout: ¯\_(ツ)_/¯ They’re all used in Planner. `change` defined here was used from Planner before. I think it’d be good to make `change` private, which then makes it harder to move the others.

sellout:
Ok, @rabbitonweb, this is updated. Conflicts resolved.

rabbitonweb:
LGTM

PR #1663, opened 2016-10-30
Eliminating `Fix` from `LogicalPlan`.

sellout:
The tests don’t yet compile, but I wanted to give @alissapajer something to work from.

The other thing I’d like to do is move `quasar.logicalPlan` to `quasar.frontend.logicalPlan`.


alissapajer:
@sellout perfect thanks!

alissapajer:
@sellout is there a reason you named the package `logicalPlan` with a capital "P"? I strongly prefer package names to be strictly lower case. Would you be okay with calling it `logicalplan` instead?

With that said, our naming of `.scala` files (main and test) is also very inconsistent, especially when it comes to casing. And I'd like to deal with that at some point, but I'm not going to open that can of worms now :)

frontend/src/main/scala/quasar/LogicalPlan.scala
```diff
@@ -1,488 +0,0 @@
...
-
-  final case class Read[A](path: FPath) extends LogicalPlan[A] {
-    override def toString = s"""Read("${path.shows}")"""
-  }
-
-  final case class Invoke[A, N <: Nat](func: GenericFunc[N], values: Func.Input[A, N]) extends LogicalPlan[A] {
-    override def toString = {
-      func.shows + "(" + values.mkString(", ") + ")"
-    }
-    override def equals(that: scala.Any): Boolean = that match {
```
alissapajer: @sellout was this removed for a reason?

`Func.Input1(4) == Func.Input1(4)` evaluates to `false` so we need to run equality on the `unsized` version.

frontend/src/main/scala/quasar/logicalPlan/package.scala
```diff
@@ -0,0 +1,56 @@
...
+
+package object logicalPlan {
+  def read[A] =
+    Prism.partial[LogicalPlan[A], FPath] { case Read(p) => p } (Read(_))
+
+  def constant[A] =
+    Prism.partial[LogicalPlan[A], Data] { case Constant(p) => p } (Constant(_))
+
+  def invoke[A] =
+    Prism.partial[LogicalPlan[A], (GenericFunc[Nat], Func.Input[A, Nat])] {
```
alissapajer: This `Prism` is incorrect because we need a guarantee that all references to `Nat` are the same `N <: Nat`. Will change to three specialized versions `invoke1`, `invoke2`, and `invoke3`.

alissapajer: And then we'll be able to use the appropriate `Prism` to construct `Invoke` instances everywhere.

alissapajer: And a useful link [Faking-Sum-Types](https://github.com/quasar-analytics/quasar/wiki/Faking-Sum-Types)

frontend/src/main/scala/quasar/LogicalPlan.scala
sellout: Yeah – we should be using Equal[LogicalPlan] instead. Implementing this with `scala.sys.error` should show everywhere that needs to be changed over.

alissapajer: Also, this is why all the tests were failing.

alissapajer:
Closing in favor of #1679 . Had to cherry-pick @sellout's original commit because commits got muddled.

frontend/src/main/scala/quasar/logicalPlan/package.scala
alissapajer: This is a TODO in #1679 - I wasn't able to hack it with `invoke1` etc.

Issue #1295, opened 2016-07-27
Too many CI failures


paulp:
I wish to suggest that the spurious failure rate on travis is too high - especially when all four jobs must succeed. The damage it does to one's flow to have to perpetually check back in on factors out of one's control in order to make forward progress should not be underestimated.

[Here's my most recent example](https://travis-ci.org/quasar-analytics/quasar/builds/147578627). Three out of four jobs succeed, one fails because:

```
[error]     x be 400 if attempting to move a dir into a file
[error]      Gave up after only 1 passed tests. 10 tests were discarded. (DataServiceSpec.scala:54)
```

I'm not sure what to make of that actually - why did it give up after only 10 discards?

Here's a similar pattern from a different spurious failure:

```
[error]     x and list of children for existing nonempty directory
[error]      Gave up after only 1 passed tests. 10 tests were discarded. (MetadataServiceSpec.scala:65)
```

mossprescott:
Yes, we're all very aware of how disruptive this is and we've made a couple of attempts to fix it with incomplete success.

It's somehow related to the behavior of scalacheck when `minTestsOK` is set to a small value, which we do in some cases because the tests otherwise run very slow.

mossprescott:
For some context, see #1172. I believe there was another PR and a JIRA ticket, but I can't find either at the moment.

paulp:
Right, I didn't mean to suggest you weren't aware. I'll see what I can figure out.

paulp:
Collecting more examples for reference. Probably a completely separate underlying cause, failing right at the end:
```
 Quasar Version: 11.0.4
 Launching Quasar JAR...
 Quasar PID: 8076
 JarClassLoader: Warning: quasar/LowPriorityImplicits.class in lib/quasar-foundation-internal_2.11-11.0.4.jar is hidden by lib/quasar-core- internal_2.11-11.0.4.jar (with different bytecode)
 Read config from /home/travis/.config/quasar/quasar-config.json
 Killing Quasar JAR...
 Status code from killing Quasar: 0
 Quasar failed to start
```

paulp:
```
[error]     x be 204 with existing dir
[error]      Gave up after only 1 passed tests. 10 tests were discarded. (DataServiceSpec.scala:54)
```

paulp:
[An interesting one](https://travis-ci.org/quasar-analytics/quasar/jobs/147583994):
```
[...]

PR #1670, opened 2016-10-31
CB: Handle connection refused

drostron:
addresses: slamdata/slamdata#1159

### DIFF
--- a/couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
+++ b/couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
@@ -97,7 +97,7 @@ package object fs {
                    ).as(
                      ().right
                    ).handle {
-                     case ex: InvalidPasswordException =>
+                     case _: InvalidPasswordException =>
                        invalidCredentials(
                          "Unable to obtain a ClusterManager with provided credentials."
                        ).right[NonEmptyList[String]].left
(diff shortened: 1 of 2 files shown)


couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -87,10 +95,15 @@ package object fs {
                      cm.info
                    ).as(
                      ().right
-                   ).handle { case ex: InvalidPasswordException =>
-                     invalidCredentials(
-                       "Unable to obtain a ClusterManager with provided credentials."
-                     ).right[NonEmptyList[String]].left
+                   ).handle {
+                     case ex: InvalidPasswordException =>
```
alissapajer: `ex` is not used

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -50,6 +51,13 @@ package object fs {
     KeyValueStore[ResultHandle, Cursor, ?]
   )#M[A]
 
+  object CBConnectException {
+    def unapply(ex: Throwable): Option[Throwable] = ex.getCause match {
```
drostron: This would be better as `Option[ConnectException]d` — will update.

alissapajer:
Do we have the test infrastructure to test something like this right now?

drostron:
The necessary infrastructure exists. As far as I know testing a server startup scenario such as this would be new however.

alissapajer:
@drostron was just checking on the build here, had a failure in `PrettifySpecs`. Added it to #1295 . I don't see how that could be related to this PR, but I'll let you restart that build to confirm.

drostron:
Thanks for adding it to #1295 and agree that it's unrelated — _restarted_.

PR #1681, opened 2016-11-03
Add Specification for `KeyValueStore` effect

jedesah:


effect/src/test/scala/quasar/effect/KeyValueStoreSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+          ops.put(key, value) *> ops.put(otherKey, otherValue) *> ops.keys
+        ) must_= Vector(key, otherKey)
+      }
+    }
+  }
+}
+
+object DefaultEmptyImpl extends KeyValueStoreSpec {
+  def eval[A](program: Free[S, A]) =
+    program.foldMap(KeyValueStore.impl.empty.unsafePerformSync).unsafePerformSync
```
wemrysi: How about `KeyValueStore.impl.empty.flatMap(program foldMap _).unsafePerformSync`?

jedesah: I was on the fence already 😄

PR #1682, opened 2016-11-03
Sql ast uses scalaz Equal.

### LINKED ISSUE #1678: scalaz.Equal for Sql
Create a `matryoshka.Delay[scalaz.Equal, Sql]` and delete all `override def equals` methods.

alissapajer:
Also breaks Sql ast into multiple files.

Fixes #1678


drostron:
👍🏾

PR #1683, opened 2016-11-03
CB: Update explain, speed up ListContents, list empty buckets

drostron:


couchbase/src/main/scala/quasar/physical/couchbase/common.scala
```diff
@@ -41,6 +41,8 @@ object common {
 
   final case class DocIdType(id: String, tpe: String)
```
alissapajer: How many different values of `id` and `tpe` are there? I'm wondering if it would be possible to make an algebra of these values so you could use a type other than `String`?

Then same for the `v` in `DocType` below.

couchbase/src/main/scala/quasar/physical/couchbase/common.scala
```diff
@@ -41,6 +41,8 @@ object common {
...
 
+  final case class DocType(v: String)
```
drostron: Yep, same as `tpe` above, but not being too rigorous about these things yet, however now is a reasonable time to start moving that direction.  Only added it to avoid having `List[String]` in a few places. I almost changed `DocTypeId` as you say but didn't for reasons that now seem trite.

Sure.

drostron: In fact quite trite since `DocIdType` isn't used anymore 😄.

alissapajer: Deleting something is always even better! :)

drostron: Oh, and `v` can be any string here. Quasar stores documents with a pathlike type but a bucket can have documents stored otherwise.

drostron:
Ready for another look.

alissapajer:
LGTM.

PR #1684, opened 2016-11-03
Faster keyvaluestore implementation

jedesah:


jedesah:
This PR is based on #1681. I still need to add a `revision.md` that is appropriate for the changes in this PR.

jedesah:
This is ready for review now

drostron:
Was there a particular usage that was slow and motivated this? If so, did you notice an improvement?

drostron:
@jedesah just a minor comment and a quick question.

jedesah:
Sorry. Forgot about this. No particular usage that was slow. Just kind of always bugged me that the default implementation was a sequential one. i.e. that concurrent threads need to wait for each other even if they are dealing with separate keys. This new default implementation should not suffer from this problem. I don't have any proof that it's faster in practice though.

drostron:
Context appreciated. 👍🏼

jedesah:
Oh, I was actually going to change that `exactly` thing you mentioned 😄

drostron:
Ah, too quick on the trigger. I'm ok with it either way 😄.

PR #1685, opened 2016-11-04
Use Equal for LogicalPlan in as many places as possible.

### LINKED ISSUE #1677: scalaz.Equal for LogicalPlan
Create a `scalaz.Equal[LogicalPlan]` and delete all `override def equals` methods.

alissapajer:
Closes #1677 . Sort of. So I didn't get to go quite as far with this PR as I wanted. The goal was to delete the `override def equals` method on ` logicalplan.Invoke`, but because of the pervasive use of `==` in external dependencies and scalac, this was not possible. But this PR _does_ rely on the `scalaz.Equal` for `LogicalPlan` more than before, so it's an improvement.

### DIFF
--- a/frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
+++ b/frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
@@ -31,20 +31,17 @@ import scalaz._, Scalaz._
 import shapeless.{Nat, Sized}
 import pathy.Path.posixCodec
 
-sealed abstract class LogicalPlan[A] extends Product with Serializable
+sealed abstract class LogicalPlan[A] extends Product with Serializable {
+  override def equals(that: scala.Any): Boolean = that match {
+    case lp: LogicalPlan[A] => LogicalPlan.equal(Equal.equalA[A]).equal(this, lp)
+    case _                  => false
+  }
+}
 
 final case class Read[A](path: FPath) extends LogicalPlan[A]
 final case class Constant[A](data: Data) extends LogicalPlan[A]
 final case class Invoke[A, N <: Nat](func: GenericFunc[N], values: Func.Input[A, N])
-    extends LogicalPlan[A] {
-  // TODO this should be removed, but usage of `==` is so pervasive in
-  // external dependencies (and scalac) that removal may never be possible
-  override def equals(that: scala.Any): Boolean = that match {
-    case that @ Invoke(_, _) =>
-      (this.func == that.func) && (this.values.unsized == that.values.unsized)
-    case _ => false
-  }
-}
+    extends LogicalPlan[A]
 // TODO we create a custom `unapply` to bypass a scalac pattern matching bug
 // https://issues.scala-lang.org/browse/SI-5900
 object InvokeUnapply {
--- a/frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
+++ b/frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
@@ -32,6 +32,8 @@ import shapeless.{Nat, Sized}
 import pathy.Path.posixCodec
 
 sealed abstract class LogicalPlan[A] extends Product with Serializable {
+  // TODO this should be removed, but usage of `==` is so pervasive in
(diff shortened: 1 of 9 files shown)


frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
```diff
@@ -37,10 +37,8 @@ final case class Read[A](path: FPath) extends LogicalPlan[A]
 final case class Constant[A](data: Data) extends LogicalPlan[A]
 final case class Invoke[A, N <: Nat](func: GenericFunc[N], values: Func.Input[A, N])
     extends LogicalPlan[A] {
-  override def toString = {
-    func.shows + "(" + values.mkString(", ") + ")"
-  }
-  // TODO remove this #1677
+  // TODO this should be removed, but usage of `==` is so pervasive in
+  // external dependencies (and scalac) that removal may never be possible
   override def equals(that: scala.Any): Boolean = that match {
```
alissapajer: This is what I wanted gone. It would even be an improvement to call the `Equal[LogicalPlan[A]]` here, but that's not possible without requiring an `Equal[A]` in the construction of `Invoke`, which is not something I think is reasonable. @paulp suggested we could solve this by storing an `Option[Equal[A]]` in a `LogicalPlan[A]`, but at this point I don't think it's worth the time to pursue such a route. So, this stands as is :/

frontend/src/test/scala/quasar/frontend/logicalplan/LogicalPlanSpecs.scala
```diff
@@ -25,18 +25,26 @@ import quasar.std
...
 import shapeless.contrib.scalaz.instances._
 import pathy.Path._
 
-class LogicalPlanSpecs extends Spec {
+class LogicalPlanSpecs extends Spec with ScalazMatchers {
+  // `Gen.oneOf` calls `Gen.suchThat` which in turn calls `==`
+  // We cannot compare `LogicalPlan` instances with `==` and instead prefer `scalaz.Equal`.
+  // Removing the call to `Gen.suchThat` may have a negative performance impact or
+  // may result in nonoptimal example simplification.
+  def oneOfNoEquals[T](gs: Gen[T]*): Gen[T] = {
```
alissapajer: See the comment here. Is this hack worth keeping? Because I didn't delete the `override def equals`, we _could_ just use `Gen.oneOf` here and the tests would pass.

frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
wemrysi: What about removing this and overriding `equals` in `LogicalPlan` using `Equal.equalA` for the `A`. It isn't ideal, but it is  a small improvement over what we have as we won't be duplicating the definition for `Equal[LogicalPlan[A]]` and `equals`.

frontend/src/test/scala/quasar/frontend/logicalplan/LogicalPlanSpecs.scala
wemrysi: I guess if we aren't really reaping its benefits, then may as well delete it.

We use `Gen.oneOf` pervasively throughout the project and sounds like we might want to sort this out for all our usages at some point (if possible, I haven't looked into what that entails).

alissapajer: Yeah, I agree. I'll delete this.

alissapajer: At some point we could also consider [scalaprops](https://github.com/scalaprops/scalaprops), but  now isn't the time for that.

frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlan.scala
alissapajer: That's a good idea. Just  pushed the change. I opted to directly call the implicits rather than let them be found.

wemrysi:
LGTM, I'll merge when green.

PR #1686, opened 2016-11-04
CB: Proto datetime

drostron:
This is meant as an interim solution. Ultimately we'll want to preserve datetime information in the N1QL representation so we can be more informed and precise when generating comparisons. That way we can avoid recovering it within the generated N1QL which is a bit more noisy than ultimately necessary.

Addresses part of: #1601


couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -34,67 +36,145 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
...
+       when regexp_contains($exprN1ql, "$regex") then { "$key": $exprN1ql }
+       else null
+       end)"""
+    ).point[M]
+  }
+
+  def rel(a1: N1QL, a2: N1QL, op: String): M[N1QL] = {
+    val a1N1ql = n1ql(a1)
+    val a2N1ql = n1ql(a2)
+    val l = s"""
```
drostron: I'll methodize these to avoid the duplication.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -34,67 +36,138 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
...
+  }
+
+  def rel(a1: N1QL, a2: N1QL, op: String): M[N1QL] = {
+    val a1N1ql = n1ql(a1)
+    val a2N1ql = n1ql(a2)
+    def unwrap(expr: String) = s"""
+      (case
+       when $expr.["$DateKey"]      is not null then $expr.["$DateKey"]
+       when $expr.["$TimeKey"]      is not null then $expr.["$TimeKey"]
+       when $expr.["$TimestampKey"] is not null then $expr.["$TimestampKey"]
```
alissapajer: Are these three cases mutually exclusive? Or is it possible that, for example, both the `$DateKey` and the `$TimeKey` cases could match, which would mean that the order of the `when` clauses affects the result.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -34,67 +36,138 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
...
+        ).point[PR]
+      ))
     case Undefined() =>
       partialQueryString("null").point[M]
 
     // array
     case Length(a1) =>
-      partialQueryString(s"array_length(${n1ql(a1)})").point[M]
+      val a1N1ql = n1ql(a1)
+      partialQueryString(s"ifnull(length($a1N1ql), array_length($a1N1ql))").point[M]
```
alissapajer: Why do you need `ifnull` for some functions but not others? Is there something special about those functions or would it be beneficial to use `ifnull` everywhere by default?

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -118,17 +191,17 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
...
+      rel(a1, a2, "<")
     case Lte(a1, a2)         =>
-      partialQueryString(s"(${n1ql(a1)} <= ${n1ql(a2)})").point[M]
+      rel(a1, a2, "<=")
     case Gt(a1, a2)          =>
-      partialQueryString(s"(${n1ql(a1)} > ${n1ql(a2)})").point[M]
+      rel(a1, a2, ">")
     case Gte(a1, a2)         =>
-      partialQueryString(s"(${n1ql(a1)} >= ${n1ql(a2)})").point[M]
+      rel(a1, a2, ">=")
```
alissapajer: `rel` looks like a nice abstraction :+1:

it/src/main/resources/tests/convertToFromString.test
```diff
@@ -3,8 +3,7 @@
     "backends": {
         "mongodb_read_only": "pending",
         "postgresql":        "pending",
-        "marklogic":         "skip",
-        "couchbase":         "skip"
```
alissapajer: Awesome to have a few more tests enabled!

alissapajer:
@drostron back to you with a couple questions!

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
drostron: They should be mutually exclusive based on their existence due to `DataCodec.Precise`. There could be existing data not loaded by quasar that happens to use the same keys, but this is not meant to handle that situation.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
drostron: Ideally `length` would handle more types, but since it doesn't handle arrays we need to defer to `array_length`. There's also `object_length` which I'm adding in an upcoming PR. It's not great, but works and is more concise than a case expression.

drostron:
@alissapajer back to you with a few answers!

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
alissapajer: Could you use `ifmissingornull` to replace this entire case statement? Is the "missing" component implicit in the `when` clause? I was looking at some docs [here](http://developer.couchbase.com/documentation/server/current/n1ql/n1ql-language-reference/condfununknown.html).

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -34,67 +36,138 @@ final class MapFuncPlanner[F[_]: Monad: NameGenerator, T[_[_]]: Recursive: ShowT
...
+       when $expr.["$DateKey"]      is not null then $expr.["$DateKey"]
+       when $expr.["$TimeKey"]      is not null then $expr.["$TimeKey"]
+       when $expr.["$TimestampKey"] is not null then $expr.["$TimestampKey"]
+       else $expr
+       end)"""
+    def trunc(v: String) = s"""
+      (case
+       when ifmissing($a1N1ql.["$DateKey"], $a2N1ql.["$DateKey"]) is not null
+       then date_trunc_millis(millis($v), "day")
+       else $v
```
alissapajer: It looks like it would be nice to call some function here (something kind of like `ifmissingornull`) to replace the `when`-`then`-`else` pattern, but the `then` cause throws this for an extra loop, and I didn't see any functions in the docs that match this pattern.

alissapajer:
LGTM will merge when green.

PR #1687, opened 2016-11-05
Finish and test* MarkLogic MapFunc implementation

### LINKED ISSUE #1502: MarkLogic representation and handling of "Undefined" values
The `QScript` data model includes `Undefined` values which need a representation in MarkLogic as well as handling of the `MapFunc`s `Undefined` and `IfUndefined`.

Other functions may need modification to deal with undefined values as well once the semantics of undefined have been clarified.

wemrysi:
Finishes the planning of the last few unimplemented `MapFunc`s for MarkLogic and adds the necessary infrastructure to get the first set of `StdLibSpec`s running.

Also includes a handful of fixes discovered via more testing.

[*] Only unary tests are enabled currently, the binary and ternary will be enabled as soon as time permits.

Closes #1502

### DIFF
--- a/it/src/main/resources/tests/groupAndWildcard.test
+++ b/it/src/main/resources/tests/groupAndWildcard.test
@@ -9,6 +9,7 @@
   "query": "select * from zips where _id like \"8030%\" group by city",
 
   "predicate": "containsExactly",
+  "ignoredFields": ["_id"],
 
   "ignoredFields": ["_id"],
 
--- a/it/src/main/resources/tests/groupAndWildcard.test
+++ b/it/src/main/resources/tests/groupAndWildcard.test
@@ -9,7 +9,6 @@
   "query": "select * from zips where _id like \"8030%\" group by city",
 
   "predicate": "containsExactly",
-  "ignoredFields": ["_id"],
 
   "ignoredFields": ["_id"],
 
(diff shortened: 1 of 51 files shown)


it/src/main/resources/tests/mongodb/simpleProjectWithOneRenamed.test
```diff
@@ -6,14 +6,14 @@
     "mongodb_3_0":       "pending",
     "mongodb_read_only": "pending",
     "mongodb_3_2":       "pending",
```
sellout: I know this isn’t new with this PR, but any idea why this is pending on Mongo? The query seems simple, and the kind of thing I do frequently.

it/src/main/resources/tests/groupAndWildcard.test
```diff
@@ -9,6 +9,7 @@
   "query": "select * from zips where _id like \"8030%\" group by city",
 
   "predicate": "containsExactly",
+  "ignoredFields": ["_id"],
```
wemrysi: So it does, will fix.

it/src/main/resources/tests/mongodb/simpleProjectWithOneRenamed.test
wemrysi: No idea, I haven't actually tried running it on Mongo, maybe it works now?

wemrysi:
@sellout Aye, that sounds reasonable. Also useful if we ever change the format of the generated names (which is actually an argument to not rely on them in the other specs as we'll have to update all of them if the format changes).

wemrysi:
@sellout Ok, I think this is ready to go.

Issue #1690, opened 2016-11-09
FLWOR fusion for ThetaJoin when both branches consist only of mappings over a common src


wemrysi:
When both branches of a `ThetaJoin` are FLWOR expressions representing a mapping (i.e. no filtering or sorting) over the data emitted from `src` they can be fused into a single FLWOR expression.

PR #1691, opened 2016-11-09
Basic XQuery FLWOR expression fusion for MarkLogic Planner

### LINKED ISSUE #1625: Fuse multiple FLWOR expressions
Multiple QScript nodes are implemented as XQuery FLWOR expressions, in many cases these can be consolidated into a single expression and avoid traversing the dataset multiple times.

wemrysi:
Implements basic fusion for FLWOR expressions in the MarkLogic planner and converts `ShiftedRead` plan from a function call to a FLWOR expression in order to benefit from fusion.

The fusion implemented relies on contextual information within the planner and isn't completely general, though it could be made so once we have a complete fixpoint AST for XQuery.

Fusion for `ThetaJoin` branches sharing a non-trivial common src isn't implemented here, but is tracked in #1690.

Closes #1625

### DIFF
--- a/marklogic/src/main/scala/quasar/physical/marklogic/xquery/FunctionDecl.scala
+++ b/marklogic/src/main/scala/quasar/physical/marklogic/xquery/FunctionDecl.scala
@@ -124,7 +124,7 @@ object FunctionDecl {
     def as(rType: SequenceType): FunctionDecl1Dsl = copy(rt = rType)
 
     def apply[F[_]: Functor](body: XQuery => F[XQuery]): F[FunctionDecl1] =
-      body(p1.ref) map (FunctionDecl1(fn, p1, rt, _))
+      body(~p1) map (FunctionDecl1(fn, p1, rt, _))
 
     def apply(body: XQuery => XQuery): FunctionDecl1 =
       apply[Id](body)
@@ -134,7 +134,7 @@ object FunctionDecl {
     def as(rType: SequenceType): FunctionDecl2Dsl = copy(rt = rType)
 
     def apply[F[_]: Functor](body: (XQuery, XQuery) => F[XQuery]): F[FunctionDecl2] =
-      body(p1.ref, p2.ref) map (FunctionDecl2(fn, p1, p2, rt, _))
+      body(~p1, ~p2) map (FunctionDecl2(fn, p1, p2, rt, _))
 
     def apply(body: (XQuery, XQuery) => XQuery): FunctionDecl2 =
       apply[Id](body)
@@ -144,7 +144,7 @@ object FunctionDecl {
     def as(rType: SequenceType): FunctionDecl3Dsl = copy(rt = rType)
 
     def apply[F[_]: Functor](body: (XQuery, XQuery, XQuery) => F[XQuery]): F[FunctionDecl3] =
-      body(p1.ref, p2.ref, p3.ref) map (FunctionDecl3(fn, p1, p2, p3, rt, _))
+      body(~p1, ~p2, ~p3) map (FunctionDecl3(fn, p1, p2, p3, rt, _))
 
     def apply(body: (XQuery, XQuery, XQuery) => XQuery): FunctionDecl3 =
       apply[Id](body)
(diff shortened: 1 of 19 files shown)


sellout:
Any idea how this affects the performance problems caused by the unnecessary provenance? I think @alissapajer is close to a PR for that, anyway, but I wonder if this eliminates the performance aspect of it.

wemrysi:
It has a moderate effect, but wasn't the panacea I hoped it would be. I suspect all the extra structure, especially capturing things like predicate results and then projecting them out in the filter, defeats the MarkLogic optimizer.

It seems like creating the structure itself is expensive, maybe the current implementation is suboptimal in that regard. Prior to using the `StaticArray`/`StaticMap` extractors the profiler said that just building and concatenating all the structure was taking 50% of the execution time, after those it is down to around 30%, but that still seems high.

jdegoes:
Close to time to pull out the profiler...

marklogic/src/main/scala/quasar/physical/marklogic/xquery/FunctionDecl.scala
```diff
@@ -92,79 +92,79 @@ object FunctionDecl {
...
       FunctionDecl5Dsl(fname, p1, p2, p3, p4, p5, SequenceType.Top)
   }
 
-  final case class FunctionDecl1Dsl(fn: QName, p1: TypedBinding, rt: SequenceType) {
+  final case class FunctionDecl1Dsl(fn: QName, p1: TypedBindingName, rt: SequenceType) {
     def as(rType: SequenceType): FunctionDecl1Dsl = copy(rt = rType)
 
     def apply[F[_]: Functor](body: XQuery => F[XQuery]): F[FunctionDecl1] =
-      body(p1.name.xqy) map (FunctionDecl1(fn, p1, rt, _))
+      body(~p1) map (FunctionDecl1(fn, p1, rt, _))
```
alissapajer: @wemrysi how does the `~` syntax work? I've seen various different expressions (e.g. `a.xqy` or `a.name.zqy` or `a.ref`) get replaced with `~a`. Maybe I missed it in the diff, but how is this working?

marklogic/src/main/scala/quasar/physical/marklogic/xquery/package.scala
```diff
@@ -81,16 +94,17 @@ package object xquery {
       Show.showFromToString
   }
 
-  final case class TypedBinding(name: BindingName, tpe: SequenceType) {
-    def render: String = s"${name.xqy} as $tpe"
+  final case class TypedBindingName(name: BindingName, tpe: SequenceType) {
```
alissapajer: A general question so I can learn a bit more about this: In what situations don't you have a `SequenceType`, meaning you'd need to use `BindingName` instead of `TypedBindingName`? Is this at all related to the type check information provided by QScript?

alissapajer:
@wemrysi LGTM - just a couple questions on things I was curious about.

alissapajer:
Also when I was searching through the diff, I found it pretty much impossible to type "Flwor" and typed "Flower" every single time :)

marklogic/src/main/scala/quasar/physical/marklogic/xquery/FunctionDecl.scala
wemrysi: It is a reference to a binding name in XQuery. If you create a binding name, typed or not as `val x = ...` then `~x` represents the reference to that binding in XQuery. Some of the places you're seeing it replace other forms is also due to more pervasive use of `BindingName`/`TypedBindingName` than previously (like in FLWOR expressions).

marklogic/src/main/scala/quasar/physical/marklogic/xquery/package.scala
wemrysi: It isn't currently related to the type check information provided by QScript, we might be able to leverage that eventually.

There are many places where the `SequenceType` is optional and `item()*` (which is XQuery for `Top`) is implied in those cases. Positional variables do not allow a type (the `$i` in `for $x at $i in ...`) as it is known that it will always be `xs:integer`.

wemrysi:
@alissapajer Haha, yeah I kept typing flowr instead of flwor myself

PR #1697, opened 2016-11-12
Add casting to MarkLogic planner where beneficial

### LINKED ISSUE #1504: Cast atomic `Data` values when projecting fields and indices in MarkLogic
In order for most aggregations and other operations to behave as one would expect, we want atomic values to have proper XDM types. Since we don't have a schema for `Data` encoded as XML, we'd like to cast values when projecting them according to their `ejson:type` attribute.

wemrysi:
Adds type casts based on the `ejson:type` attribute of elements to the MarkLogic planner where they would be beneficial (operands to various operators, sort keys, etc) resulting in another handful of passing regression tests.

Also improves `Constant` planning by using the `EJson => Data` algebra and unifies date/time result parsing for XQuery literals and elements with the corresponding `ejson:type` attribute.

Closes #1504.

### DIFF
--- a/it/src/main/resources/tests/flattenArray.test
+++ b/it/src/main/resources/tests/flattenArray.test
@@ -2,11 +2,17 @@
     "name": "flatten array",
     "backends": {
         "postgresql": "pending",
-        "marklogic":  "skip",
         "couchbase":  "skip"
     },
     "data": "zips.data",
-    "query": "SELECT loc[*] AS loc FROM zips LIMIT 1",
-    "predicate": "containsExactly",
-    "expected": [{ "loc" : -72.622739}]
+    "query": "SELECT loc[*] AS loc FROM zips",
+    "predicate": "containsAtLeast",
+    "expected": [
+      { "loc" : -72.622739 },
+      { "loc" :  34.933613 },
+      { "loc" : -78.504109 },
+      { "loc" :  34.994081 },
+      { "loc" : -79.261843 },
+      { "loc" :  34.588664 }
+    ]
 }
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,3 @@
+- [1504] Cast operands of various operations in MarkLogic planner
+  - Unify parsing of date/time XQuery literals and elements
+  - Update MarkLogic planner for Constants to go through Data
--- a/it/src/main/resources/tests/complex-vars.test
+++ b/it/src/main/resources/tests/complex-vars.test
@@ -2,8 +2,7 @@
   "name": "variable with a non-trivial value",
 
   "backends": {
-    "postgresql": "pending",
-    "marklogic":  "skip"
+    "postgresql": "pending"
   },
 
   "data": "days.data",
--- a/it/src/main/resources/tests/filterByJsWithOnlyReducing.test
+++ b/it/src/main/resources/tests/filterByJsWithOnlyReducing.test
@@ -4,16 +4,15 @@
   "backends": {
       "mongodb_read_only": "pending",
       "postgresql":        "pending",
-      "marklogic":         "skip",
       "couchbase":         "skip"
   },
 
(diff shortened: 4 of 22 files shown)


wemrysi:
Hm, looks like the MarkLogic build is taking too long to output results for travis, restarted twice to no avail :disappointed:

it/src/main/resources/tests/flattenArray.test
```diff
@@ -2,11 +2,17 @@
     "name": "flatten array",
     "backends": {
         "postgresql": "pending",
-        "marklogic":  "skip",
         "couchbase":  "skip"
     },
     "data": "zips.data",
-    "query": "SELECT loc[*] AS loc FROM zips LIMIT 1",
```
alissapajer: Should we keep this test as it was (with the `LIMIT`) and then write and second test without `LIMIT`? That way we aren't removing a test, but can still show that marklogic passes the test without the `LIMIT`. 

(I was assuming this test didn't work in marklogic? Or was there another reason you removed the `LIMIT`?)

alissapajer:
@wemrysi do you suspect that the new tests you enabled here are slowing down the build, causing it to time out? If so, in order to get this merged, we could re-skip the slow tests. And then open tickets to figure out why the tests are so slow.

it/src/main/resources/tests/flattenArray.test
wemrysi: I'm not sure this test would work in anything but `mongodb` as it assumes the output order without sorting, which seemed brittle to me.

wemrysi:
@alissapajer Yes, I think that is a good idea, I'm sure one of these new tests is slowing things down. I'll narrow it down tomorrow, re-skip it and open an issue like you suggest.

it/src/main/resources/tests/flattenArray.test
alissapajer: Do we then have other tests that test `LIMIT`, sorting the data first, then taking `LIMIT`, so that we expect the same results for each backend?

alissapajer: An `ack` of this directory shows that only this `flattenArray.test` is using `LIMIT`. I think we should still have an integration test testing `LIMIT` so what do you think of  adding a second test that sorts first, then takes the limit?

wemrysi: Yes, a couple examples are `citiesByLargestZip`, `citiesByTotalPopulation` and `filterSortLimit`. There are also some others that use the `LIMIT` without sorting that should be fixed, some also currently expect data nested in a `value` field, another `mongodb` peculiarity.

alissapajer: Oh, I just `ack`ed for `LIMIT` and not `limit` - I see all the other tests now!

alissapajer: The case I was concerned about looks well-covered in our test suite, so I'm fine keep this change as you made it.

alissapajer:
@wemrysi it looks like the marklogic build is still timing out

wemrysi:
Aye, looks like I'll have to disable some more =/

wemrysi:
@alissapajer Success! After discussing some ideas with @drostron, I decided to try adding a basic [progress indicator](https://travis-ci.org/quasar-analytics/quasar/jobs/176458319#L3200) to the regression tests so that travis would receive some indication that progress is being made. This also allowed me to keep all the tests enabled.

alissapajer:
@wemrysi great! Looks like the marklogic build is coming in at 1hr 2min. Performance is a top qscript priority, so we should be able to see that number go down. But being able to show correctness of the queries today is great so we know we don't regress!

Merging away.

alissapajer:
@wemrysi got a `fatal: Not possible to fast-forward, aborting.` Can you merge master?

wemrysi:
@alissapajer Merged and back in Travis' court.

PR #1702, opened 2016-11-17
Fix MarkLogic array concatenation and enable all MapFunc tests

### LINKED ISSUE #1503: Generalize MarkLogic array handling/operations to work on any XML element
Now that the `Data` representation of `EJSON`/`Data` arrays is more "natural", generalize how arrays are handled in the planner to work with any XML element. In other words, every XML element can be treated as an array of its children.

wemrysi:
Fixes array concat to work with strings and arrays in MarkLogic. Also enables all the `StdLibSpec`s for MarkLogic and adds a few new ones exercising `ArrayConcat`.

In the process, also relaxed various array operations to work on any XML element by not requiring child elements to have a particular name.

Fixes #1503.

### DIFF
--- a/frontend/src/main/scala/quasar/std/structural.scala
+++ b/frontend/src/main/scala/quasar/std/structural.scala
@@ -149,19 +149,21 @@ trait StructuralLib extends Library {
     Func.Input2(AnyArray ⨿ Str, AnyArray ⨿ Str),
     noSimplification,
     partialTyperV[nat._2] {
-      case Sized(t1, t2) if t1.arrayLike && t2.contains(AnyArray ⨿ Str) => ArrayConcat.tpe(Func.Input2(t1, FlexArr(0, None, Top)))
-      case Sized(t1, t2) if t1.contains(AnyArray ⨿ Str) && t2.arrayLike => ArrayConcat.tpe(Func.Input2(FlexArr(0, None, Top), t2))
-      case Sized(t1, t2) if t1.arrayLike && t2.arrayLike                => ArrayConcat.tpe(Func.Input2(t1, t2))
+      case Sized(t1, t2) if t1.arrayLike && t2.contains(AnyArray ⨿ Str)     => ArrayConcat.tpe(Func.Input2(t1, FlexArr(0, None, Top)))
+      case Sized(t1, t2) if t1.contains(AnyArray ⨿ Str) && t2.arrayLike     => ArrayConcat.tpe(Func.Input2(FlexArr(0, None, Top), t2))
+      case Sized(t1, t2) if t1.arrayLike && t2.arrayLike                    => ArrayConcat.tpe(Func.Input2(t1, t2))
 
       case Sized(Const(Data.Str(str1)), Const(Data.Str(str2)))              => success(Const(Data.Str(str1 ++ str2)))
       case Sized(t1, t2) if Str.contains(t1) && t2.contains(AnyArray ⨿ Str) => success(Type.Str)
       case Sized(t1, t2) if t1.contains(AnyArray ⨿ Str) && Str.contains(t2) => success(Type.Str)
       case Sized(t1, t2) if Str.contains(t1) && Str.contains(t2)            => StringLib.Concat.tpe(Func.Input2(t1, t2))
 
-      case Sized(t1, t2) if t1 == t2 => success(t1)
+      case Sized(t1, t2) if t1 == t2                                        => success(t1)
 
-      case Sized(t1, t2) if Str.contains(t1) && t2.arrayLike => failure(NonEmptyList(GenericError("cannot concat string with array")))
-      case Sized(t1, t2) if t1.arrayLike && Str.contains(t2) => failure(NonEmptyList(GenericError("cannot concat array with string")))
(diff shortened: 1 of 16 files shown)


it/src/main/resources/tests/mongodb/wildcardWithExtraProjections.test
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "wildcard with additional projections and filtering (MongoDB)",
+  "backends": {
+    "mongodb_read_only": "pending",
+    "postgresql":        "skip",
+    "marklogic":         "skip",
+    "couchbase":         "skip"
```
alissapajer: Just confirming that this test is meant to be skipped for all backends?

alissapajer: Ok looked a bit further. It looks like this version is a mongo-only version. And then we have a test of the same name that's slightly different that's _not_ in the mongo test dir. And that test is enabled for marklogic, but not for any other backends.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/ejson.scala
```diff
@@ -273,26 +244,22 @@ object ejson {
...
+  // ejson:singleton-array($value as item()*) as element()
   def singletonArray[F[_]: PrologW]: F[FunctionDecl1] =
-    (ejs.name("singleton-array").qn[F] |@| ejsonN.qn) { (fname, ename) =>
-      declare(fname)(
-        $("value") as ST.Top
-      ).as(ST(s"element($ename)")) { value: XQuery =>
-        mkArrayElt[F](value) flatMap (mkArray[F].apply(ename.xqy, _))
-      }
-    }.join
+    ejs.declare("singleton-array") flatMap (_(
```
alissapajer: +1 to all the cases of simplification like this using `declare`

it/src/main/resources/tests/mongodb/wildcardWithExtraProjections.test
wemrysi: The idea behind the mongo-specific test dir is to preserve passing regression tests for the legacy mongo planner that cannot be run on other backends because they assert some implementation-specific behavior (i.e. expecting results to all be under some special `value` field).

The goal would be that the QScript-based MongoDB planner would pass all the general tests and, once the legacy planner is retired, we can delete the `mongodb` dir.

There are a few in there that can be removed now that we have per-backend `ignoreFieldOrder` settings.

wemrysi:
@alissapajer I'd forgotten about the typing + constant folding for `ConcatOp`, I've remedied that and its ready for another look.

frontend/src/main/scala/quasar/std/structural.scala
```diff
@@ -149,19 +149,21 @@ trait StructuralLib extends Library {
     Func.Input2(AnyArray ⨿ Str, AnyArray ⨿ Str),
     noSimplification,
     partialTyperV[nat._2] {
-      case Sized(t1, t2) if t1.arrayLike && t2.contains(AnyArray ⨿ Str) => ArrayConcat.tpe(Func.Input2(t1, FlexArr(0, None, Top)))
-      case Sized(t1, t2) if t1.contains(AnyArray ⨿ Str) && t2.arrayLike => ArrayConcat.tpe(Func.Input2(FlexArr(0, None, Top), t2))
-      case Sized(t1, t2) if t1.arrayLike && t2.arrayLike                => ArrayConcat.tpe(Func.Input2(t1, t2))
+      case Sized(t1, t2) if t1.arrayLike && t2.contains(AnyArray ⨿ Str)     => ArrayConcat.tpe(Func.Input2(t1, FlexArr(0, None, Top)))
+      case Sized(t1, t2) if t1.contains(AnyArray ⨿ Str) && t2.arrayLike     => ArrayConcat.tpe(Func.Input2(FlexArr(0, None, Top), t2))
+      case Sized(t1, t2) if t1.arrayLike && t2.arrayLike                    => ArrayConcat.tpe(Func.Input2(t1, t2))
```
alissapajer: In the case when we have `(Const(Data.Arr(xs)), Const(Data.Arr(ys)))`, do we have constant folding for that? (That might be a subset of this case?)

alissapajer: As a side note, where constant folding happens in something we'll have to take a look at at some point down the road. QScript handles some of it, some is handled here. There may be other places (in specific backends maybe?). So we'll probably want to consolidate that as much as possible.

wemrysi: Agreed, regarding consolidation.

As for the "both arrays" case, that is handled in the `ArrayConcat` typer which `ConcatOp` delegates to when both are arrays. I guess I could just handle the `String -> [Char]` transformation here and let `ArrayConcat`'s constant folding concatenate them? We're also duplicating constant string folding here and in `StringLib.Concat`.

alissapajer: Yeah, at this point I'm not sure that it matters much where the "both arrays" case is handled, only because constant folding is already so spread out that there is not really a standard place for it.

So I think it's fine keeping it here for now.

alissapajer:
LTGM - merging.

PR #1705, opened 2016-11-17
CB: Enable additional integration tests

drostron:
Also splits out additional tests where Mongo results end up under value.

Towards: #1607


it/src/main/resources/tests/citiesByTotalPopulation.test
```diff
@@ -3,11 +3,12 @@
     "backends": {
         "postgresql": "pending",
         "marklogic":  "skip",
-        "couchbase":  "skip"
+        "couchbase":  "skipCI"
```
sellout: I don’t recall seeing this before – I’m assuming it’s because it’s too slow to run on Travis, but we want to run it locally?

it/src/main/resources/tests/couchbase/temporal/time.test
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "filter on time_of_day (Couchbase)",
+
+  "NB": "Couchbase's millis_to_utc doesn't show fractional part when milliseconds are zero",
```
sellout: I think the place where we call `millis_to_utc` should then be changed to `if millis % 1000 = 0 then millis_to_utc + ".000" else millis_to_utc` so we can use the regular tests.

it/src/main/resources/tests/filterOnContains.test
```diff
@@ -5,12 +5,11 @@
...
         "mongodb_3_2":       "pending",
-        "postgresql":        "pending",
-        "couchbase":         "skip"
+        "postgresql":        "pending"
     },
     "data": "zips.data",
     "query": "select * from zips where 43.058514 in loc[_]",
     "predicate": "equalsExactly",
-    "ignoreFieldOrder": [ "marklogic" ],
+    "ignoreFieldOrder": [ "couchbase", "marklogic" ],
```
sellout: This is a query where we _really_ want `"ignoreFieldOrder": *`, but I’m assuming we don’t yet have a way to express that.

it/src/main/resources/tests/wildcardWithExtraProjections.test
```diff
@@ -5,12 +5,12 @@
...
     "mongodb_read_only": "skip",
     "mongodb_3_2":       "skip",
-    "postgresql":        "pending",
-    "couchbase":         "skip"
+    "postgresql":        "pending"
   },
   "data": "zips.data",
   "query": "select *, concat(city, concat(\", \", state)) as city_state, city = \"BOULDER\" as boulderish from zips where city like \"BOULDER%\"",
   "predicate": "containsExactly",
+  "ignoreFieldOrder": [ "couchbase" ],
```
sellout: This should also … kind of … ignore field order everywhere. If we could actually extract field order from the query, perhaps we could specify that it should _end_ with `city_state` and `boulderish`, but the fields from the `*` are of unknown order.

it/src/main/resources/tests/couchbase/temporal/time.test
drostron: Agreed and have already made that change in #1709. Need to switch back to the regular test there however — will do so.

it/src/main/resources/tests/filterOnContains.test
drostron: Right, don't have `*` for `ignoreFieldOrder` at the moment. Adding it should be quick though.

it/src/main/resources/tests/citiesByTotalPopulation.test
drostron: Correct.

drostron:

it/src/main/resources/tests/couchbase/temporal/toFromString.test
```diff
@@ -1,8 +1,5 @@
 {
     "name": "convert dates to/from strings (Couchbase)",
-
-    "NB": "Couchbase's millis_to_utc doesn't show fractional part when milliseconds are zero",
-
```
sellout: Is this one still distinct from the non-Couchbase-specific one?

If so, I’ll merge.

drostron: Yes, as mentioned [here](https://github.com/drostron/quasar/blob/a7b2eb8821431f49b41f18f1bf7b7bd79a413de3/it/src/main/resources/tests/temporal/toFromString.test#L12), although if this result set is what we generally want then that should go under mongodb and this can become the general one. Opinion?

sellout: Ah, yeah – I think you should do that. This one definitely has the correct types.

drostron:
Ready

PR #1709, opened 2016-11-21
CB: CouchbaseStdLibSpec

drostron:


drostron:
@wemrysi Finally ready for a look.

it/src/main/resources/tests/temporal/time.test
```diff
@@ -2,19 +2,23 @@
...
+    "mongodb_3_2":       "pending",
     "postgresql":        "pending",
     "marklogic":         "skip"
   },
 
   "data": "../days.data",
 
   "query": "select day, time_of_day(ts) as tod, time_of_day(day) as `not a date`, time_of_day(missing) as missing from `../days`
-    where time_of_day(ts) >= time(\"08:00\") and time_of_day(ts) < time(\"10:20:30.400\")",
+    where time_of_day(ts) >= time(\"08:00\") and time_of_day(ts) < time(\"10:20:30.400\") limit 20",
```
wemrysi: Is the `limit` necessary? Looks like there are only 7 entries in that data set.

it/src/main/resources/tests/temporal/time.test
```diff
@@ -2,19 +2,23 @@
...
 
   "predicate": "containsExactly",
   "ignoreFieldOrder": [ "couchbase" ],
   "expected": [
-      { "day": "Tuesday",   "tod": "08:00:00.000", "not a date": {"$na": null}, "missing": {"$na": null} },
-      { "day": "Wednesday", "tod": "09:00:00.000", "not a date": {"$na": null}, "missing": {"$na": null} },
-      { "day": "Thursday",  "tod": "10:00:00.000", "not a date": {"$na": null}, "missing": {"$na": null} }]
+      { "day": "Tuesday",   "tod": { "$time": "08:00:00.000" }, "not a date": {"$na": null}, "missing": {"$na": null} },
```
wemrysi: Interesting, does this mean MongoDB returns a `String` for `tod`?

it/src/test/scala/quasar/physical/couchbase/CouchbaseStdLibSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+        MonotonicSeq.fromZero.unsafePerformSync :+:
+        Read.constant[Task, Context](ctx)
+      ).unsafePerformSync._2
+
+      (r must beRightDisjunction.like { case (q, Vector(d)) =>
+        d must beCloseTo(expected).updateMessage(_ ⊹ s"\nquery: $q")
+      }).toResult
+  }
+
+  def runner(ctx: Context) = new MapFuncStdLibTestRunner {
```
wemrysi: Looks like all the impls have to `translate(...)` with one of the `{Unary, Binary, Ternary}Arg` types. Could we DRY that up by adding new abstract members like

```
unaryMapFunc(prg: FreeMap[UnaryArg], arg: Data, expected: Data): Result
binaryMapFunc(prg: FreeMap[BinaryArg], arg1: Data, arg2: Data, expected: Data): Result
...
```

to `MapFuncStdLibTestRunner` and implement the existing `unary, binary, ...` using them?

it/src/main/resources/tests/temporal/time.test
drostron: A missed leftover from troubleshooting another issue — shears in hand.

it/src/main/resources/tests/temporal/time.test
drostron: It appears so, yes.

wemrysi: @drostron It looks like it should be a `Data.Time` (according to the `StdLibSpec` and the defn of `TimeOfDay` in `quasar.std.date`), so `$time` object it is.

drostron:
Ready for another look.

PR #1720, opened 2016-11-22
Add support for projecting prefixed fields in MarkLogic

wemrysi:
Fixes `QName` validation in the MarkLogic planner to support prefixed names.

Also ceases use of `XMLChar.java` (borrowed from Xerces) in favor of a local definition of `Name`/`NCName` as the former is for an older version of the XML spec.


marklogic-validation/src/main/scala/quasar/physical/marklogic/validation/IsName.scala
```diff
@@ -16,18 +16,25 @@
...
-  * @see https://www.w3.org/TR/xml/#NT-Name
+  * @see https://www.w3.org/TR/2008/REC-xml-20081126/#NT-Name
   */
 final case class IsName()
 
-object IsName {
+// TODO: This is a false-negative as we aren't overloading anything here.
+@SuppressWarnings(Array("org.wartremover.warts.Overloading"))
+object IsName extends (String => Boolean) {
+  def apply(s: String): Boolean =
```
sellout: I wonder if the wart is hit because you’re overriding normal function application, but not using `override`? I wonder if adding the keyword would eliminate the wart.

sellout: I wonder if the wart is hit because you’re overriding normal function application, but not using `override`? I wonder if adding the keyword would eliminate the wart.

wemrysi: Ah! `IsName` is a `case class`, I didn't realize. I'll remedy shortly.

wemrysi:
@sellout Back in your court.

sellout:
Ah, right – cool. Yeah. Looks good.

sellout:
Just noticed this failed to merge – needs a merge from master.

wemrysi:
@sellout Hopefully mergable now.

PR #1721, opened 2016-11-22
Relocate the jawn-based ejson.Json parser to ejson package

wemrysi:


paulp:
Can this one also be moved? 
```scala
  object JsonParser extends SupportParser[Data] {
    implicit val facade: Facade[Data] =
      new SimpleFacade[Data] {
        def jarray(arr: List[Data])         = Data.Arr(arr)
        def jobject(obj: Map[String, Data]) = Data.Obj(ListMap(obj.toList: _*))
        def jnull()                         = Data.Null
        def jfalse()                        = Data.False
        def jtrue()                         = Data.True
        def jnum(n: String)                 = Data.Dec(BigDecimal(n))
        def jint(n: String)                 = Data.Int(BigInt(n))
        def jstring(s: String)              = Data.Str(s)
      }
  }
```

wemrysi:
@paulp Yessir.

wemrysi:
@paulp Ready for another look.

paulp:
Looks great thanks. I'll merge.

PR #1726, opened 2016-11-28
Notify when `sparkcore` tests are being skipped

jedesah:
I am still not a fan of an environment variable being used for tests outside of integration, but at least with this it will be more obvious to developers that the tests are indeed being skipped.


jedesah:
I think @rabbitonweb might be on vacation or something, so will assign someone else in the interest of shorting open PR time.

wemrysi:
@jedesah LGTM, needs a merge from master. I'll merge once that's done and green.

PR #1733, opened 2016-11-29
CB: Document known limitations

drostron:


drostron:
Just noticed I missed the MarkLogic mount header.

drostron:
@sellout Ready

PR #1734, opened 2016-11-29
CB: Retrieve queryTimeout from connectionUri params

drostron:


couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -66,31 +66,40 @@ package object fs {
...
     val cbCtx: DefErrT[Task, Context] =
       for {
         uri     <- liftDT(
                      Uri.fromString(connectionUri.value).leftMap(_.message.wrapNel)
                    )
+        params  <- liftDT((
+                     uri.params.get("username").toSuccessNel("No username in ConnectionUri") |@|
+                     uri.params.get("password").toSuccessNel("No password in ConnectionUri") |@|
+                     (
+                       uri.params.get("queryTimeout") ∘ (s => parseLong(s)) | 300L.success
```
alissapajer: Is the `300L` here the default timeout (in seconds)? That might be more clear if that were pulled out as `val defaultTimeout: Long = 300`.

drostron: How about switching the param to "queryTimeoutSeconds"?

alissapajer: I think putting "Seconds" (or just "Secs") at the end of the name of the param would make the units part more clear.

I also was suggesting moving the default out as a `val` so it's clear when reading the the code that it's the default value. I was thinking that it would be easier to change its value in the future if you didn't have to look into the middle of a line of code to find it.

drostron: Sounds good. I'll go with "queryTimeoutSecs" and pull the default out as a `val`.

couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -42,6 +42,8 @@ import scalaz.concurrent.Task
 package object fs {
   val FsType = FileSystemType("couchbase")
 
+  val defaultQueryTimeout: Duration = Duration.ofSeconds(300)
```
alissapajer: :+: even better using `Duration`!

alissapajer:
LGTM - will merge when green

alissapajer:
@drostron `revision.md` is in the wrong directory

drostron:
Oh wow, that's a new one — foiled by the save dialog. @alissapajer should be good now.

Issue #1729, opened 2016-11-28
Eliminate usage of org.threeten.bp


paulp:
Since quasar requires java8 this library is buying us nothing except a lot of otherwise unnecessary code and error prone conversions.

One might be tempted to use implicits and convert between the various similar types, but this turns out to be a lot less trivial than one might anticipate. Just switching to `java.time` is a better use of effort.

jdegoes:
👍 This would most likely be a 10% time project.

paulp:
It turned out to be almost painless. Merged in 79ca389b33.

PR #1737, opened 2016-11-30
Eliminate dependency on org.threeten.bp.

paulp:
Since we require java 8 now, the library serves no purpose.


paulp:
#1729. It turned out to be only very slightly more effort than a search-and-replace to switch. I've spent 10x that much time looking up how to convert between the types already.

PR #1742, opened 2016-12-01
Support arbitrary XML children when decoding XML as 'Data'

wemrysi:
This fixes decoding of XML to `Data` when "encoded" elements (i.e. those that are faithful representations of `Data` values in XML) contain arbitrary XML. This arises, for example, when maps or arrays are constructed as part of a query with arbitrary values projected from documents at their leaves.


marklogic/src/test/scala/quasar/physical/marklogic/xml/ToDataSpec.scala
```diff
@@ -69,6 +69,18 @@ class ToDataSpec extends quasar.Qspec {
...
+
+      "element with just attributes" in {
+        val sample = <foo type="baz" quux="bar" />
+        val expected = Obj(
+          "foo" -> Obj(
+            "_attributes" -> Obj(
+              "type" -> Str("baz"),
+              "quux" -> Str("bar")
+            ),
+            "_text" -> Str("")))
```
sellout: So, these underscored things generated by the connector should be formatted according to the EJson spec – `_xml.attributes` (or `_ml.` or something). Also, we’ll eventually want attributes to be stored in the metadata once the connector is using EJson instead of Data, right?

So then we’d have something like
```
{ "foo": "" } @ { "_xml.attributes": { "type": "baz", "quux": "bar" } }
```
right?

sellout: A different approach might be to have an encoding of XML (all of it! 😆) inside EJson, so then something like
```
{ "tag": "foo",
  "attributes": { "type": "baz", "quux": "bar" },
  "content": { "text": "" }
} @ { "_ejson.type": "_xml.element" }
```
Not that I like that better – I think the less special processing the better – mapping as much to EJson as possible seems good, with namespaced metadata at the parts that need it.

So, a simpler encoding than the first one could even have `"type"` and `"quux"` directly as metadata, rather than under `"_xml.attributes"` perhaps.

wemrysi: Yes, we'll want to leverage `EJson`'s metadata for attributes once they're available, I'd prefer attempting the direct approach first.

I'll update the synthetic keys to use the `_xml.` prefix, I didn't realize there was any specification regarding those. Is the spec available anywhere?

PR #1743, opened 2016-12-01
[1706] Make an array when a projection results in multiple items.

### LINKED ISSUE #1706: Treat multiple child elements having the same name as an array
In addition to synthesized arrays and treating all children of an element as an array, we also need to treat multiple elements with the same name as an array.

If there are three child elements named `foo` and `foo` is then projected, the result should be an array of three elements.

wemrysi:
It is common in XML for an element to have multiple child elements with the same name, when such an element is projected in a query the result should be an array of all the matching elements. This adds support to the MarkLogic planner to support this behavior.

Fixes #1706.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- [1706] Make an array when a projection results in multiple items.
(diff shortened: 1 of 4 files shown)


wemrysi:
@drostron Same here, about time I started addressing these, more to come!

PR #1744, opened 2016-12-01
Spark fix & organize itests

rabbitonweb:
Purpose of this PR is to:

1. set failing tests to `skipped` (not spark connector fault they are failing) and `pending` (spark connector fault OR still unknown why failing)
2. keep checked list (in org-mode) of tests that are still `skipped` or `pending` with decent description of why that is

In the process few of the tests were already fixed. Other require more investigation.
Further work is to actually make all the `pending` tests to pass.


rabbitonweb:
@sellout why changed to 'inreview' ?

it-tests-failing.org
```diff
@@ -0,0 +1,68 @@
+* SKIPPED - failures not Spark related
```
drostron: Did you mean to include this?

it/src/main/resources/tests/LocationCount.test
```diff
@@ -7,7 +7,9 @@
         "mongodb_3_2":       "pending",
         "postgresql":        "pending",
         "marklogic":         "skip",
-        "couchbase":         "skip"
+        "couchbase":         "skip",
```
drostron: Indentation and alignment throughout.

rabbitonweb: yes, I want to document all tests that are skipped but they fail not because of Spark

it-tests-failing.org
rabbitonweb: yes, I want to document all tests that are skipped but they fail not because of Spark

it/src/main/resources/tests/LocationCount.test
rabbitonweb: this is going to be a pain :)

rabbitonweb: just to be sure: the issue here is that the "skips" are not in single column, correct?

drostron: Was just observing that the additions didn't follow current alignment format where connector names are indented by the same amount and directives are aligned. It's likely reasonable to auto-format these at some point to avoid the manual formatting headache.

it-tests-failing.org
rabbitonweb: I'm happy to push it as issues

rabbitonweb:
RDY

drostron:
There are still a few misalignments, but I'm ok leaving them as motivation to put an auto-formatter of some sort in place. LGTM, will merge upon Travis's 👌🏽.

drostron:
Needs a `revision.md`

it/src/main/resources/tests/LocationCount.test
rabbitonweb: I see different files being fomratted differently

drostron: I was referring to alignment throughout which you've mostly addressed. This is a case where a tool should be applied for consistent formatting. There's already at least indentation differences among the test files. We've been following the formatting within a particular file for the most part. That said, it's a better use of our time automating formatting rather than manually doing so. I'm ok with leaving things as is as mentioned in another comment.

Issue #1745, opened 2016-12-01
Remove unused array elements in QScript


alissapajer:
QScript stores provenance information in arrays. Sometimes this provenance information ultimately goes unused, meaning it can be removed from the QScript.

In the following example, only indices 2 and 3 are dereferenced from the array, meaning the array can be rewritten and the derefs can be reindexed.
```scala
    "convert a filter" in {
      // "select * from foo where bar between 1 and 10"
      convert(
        listContents.some,
        StdLib.set.Filter(
          lpRead("/bar"),
          relations.Between(
            structural.ObjectProject(lpRead("/bar"), lpf.constant(Data.Str("baz"))).embed,
            lpf.constant(Data.Int(1)),
            lpf.constant(Data.Int(10))).embed).embed) must
      beSome(beQScript(chain(
        ReadR(rootDir </> file("bar")),
        QC.inj(LeftShift((),
          HoleF,
          IncludeId,
          Free.roll(ConcatArrays(
            Free.roll(ConcatArrays(
              Free.roll(ConcatArrays(
                Free.roll(MakeArray(ProjectIndexR(RightSideF, IntLit(0)))),
                // FIXME: #1622
                Free.roll(MakeArray(ProjectIndexR(ProjectIndexR(ProjectIndexR(ProjectIndexR(RightSideF, IntLit(0)), IntLit(0)), IntLit(0)), IntLit(0)))))),
              Free.roll(MakeArray(ProjectIndexR(RightSideF, IntLit(1)))))),
            Free.roll(MakeArray(
              Free.roll(Between(
                ProjectFieldR(
                  ProjectIndexR(RightSideF, IntLit(1)),
                  StrLit("baz")),
                IntLit(1),
                IntLit(10))))))))),
        QC.inj(Filter((), ProjectIndexR(HoleF, IntLit(3)))),
QC.inj(Map((), ProjectIndexR(HoleF, IntLit(2)))))))
```

alissapajer:
This relates to #1622 in that sometimes we don't reference the excessive multiple array projections reported there, and so they can be pruned in this way. But when they cannot be pruned, that bug will still be hit.

alissapajer:
Closed by merge of #1747

Issue #1622, opened 2016-10-21
QScript for Filter with Reduce or multiple projections attempts ProjectIndex excessively


wemrysi:
Queries involving a `Filter` and a `Reduce` or a `Filter` and multiple projections results in QScript that attempts to `ProjectIndex` too far into the data from the backend. Included is an example of the former as it is more focused (to trigger the latter, a query such as `select city, loc from /explain/data where pop > 15000` suffices, but the output is about twice as long).

Interestingly, projecting a single field or `*` does not suffer from this problem.

If I'm reading the QScript correctly, the initial `LeftShift` produces a dataset where each datum has the shape `[<id>, <value>]` which the `Repair` expression then attempts to `ProjectIndex(ProjectIndex(ProjectIndex(RightSide, 0), 0), 0)`. 

**`select count(*) from /explain/data where pop > 1000`**

```
[info] Logical Plan:
[info] Let('tmp0)
[info] ├─ Read(/explain/data)
[info] ╰─ Let('tmp1)
[info]    ├─ Invoke(WHERE)
[info]    │  ├─ Free('tmp0)
[info]    │  ╰─ Invoke((>))
[info]    │     ├─ ObjectProject('tmp0["pop"])
[info]    │     ╰─ Constant(Int(1000))
[info]    ╰─ Let('tmp2)
[info]       ├─ Invoke(SQUASH)
[info]       │  ╰─ Invoke(MAKE_OBJECT)
[info]       │     ├─ Str("0")
[info]       │     ╰─ Invoke(COUNT)
[info]       │        ╰─ Free('tmp1)
[info]       ╰─ Free('tmp2)
[info]
[info] Optimized:
[info] Let('__tmp0)
[info] ├─ Read(/explain/data)
[info] ╰─ Invoke(SQUASH)
[info]    ╰─ Invoke(MAKE_OBJECT)
[info]       ├─ Str("0")
[info]       ╰─ Invoke(COUNT)
[info]          ╰─ Invoke(WHERE)
[info]             ├─ Free('__tmp0)
[info]             ╰─ Invoke((>))
[info]                ├─ ObjectProject('__tmp0["pop"])
[info]                ╰─ Constant(Int(1000))
[info]
[info] Typechecked:
[info] Let('__tmp0)
[info] ├─ Let('__check0)
[info] │  ├─ Read(/explain/data)
[info] │  ╰─ Typecheck(Obj(Map[], Some(Top)))
[info] │     ├─ Free('__check0)
[...]

PR #1747, opened 2016-12-01
Remove unused array elements from qscript arrays.

alissapajer:


connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,375 @@
...
+  def findIndices[F[_]: Functor]
+    (implicit P: PruneArrays.Aux[T, F, G], F: F :<: G)
+      : CoalgebraM[ArrayState, ArrayEnv[F, ?], T[G]] = tqs => {
+    F.prj(tqs.project) match {
+      case Some(ftg) =>
+        State(state => {
+          val Annotation(newState, newEnv) = P.find(state, ftg)
+          (newState, EnvT((newEnv, ftg)))
+        })
+      case None => ??? // TODO we can't get a `F` for the `ArrayEnv` so we cannot apply the transformation
```
alissapajer: @sellout this TODO needs to be addressed before merge. We can pass along the `Option` but we have to handle it somewhere.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,375 @@
...
+    }
+  }
+
+  // TODO can we be more efficient?
+  def indexMapping(state: StateAcc): Option[Mapping] =
+    state.map(_.toList.sorted.zipWithIndex.map(_.rightMap(BigInt(_))).toMap)
+}
+
+// TODO `find` and `remap` impls should be returning a free algebra
+// which is interpreted separately
```
alissapajer: The state-machiney-ness of the more complicated impls of `find` and `remap` below  are really bugging me and could be much improved if we separated out the interpretation of the effects. Just haven't had time to do that yet.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,375 @@
...
+            Output(newState, Coproduct.rightc(out))
+          })
+    }
+
+  implicit def read[T[_[_]], G[_]]: Aux[T, Const[Read, ?], G] = default
+  implicit def shiftedRead[T[_[_]], G[_]]: Aux[T, Const[ShiftedRead, ?], G] = default
+  implicit def deadEnd[T[_[_]], G[_]]: Aux[T, Const[DeadEnd, ?], G] = default
+
+  // TODO examine branches
+  implicit def thetaJoin[T[_[_]], G[_]]: Aux[T, ThetaJoin[T, ?], G] = default
```
alissapajer: For now we just examine `FreeMap` and not `FreeQS`. Meaning, that it's possible to rewrite a layer with branches, but I'm going to do that in another PR (with another ticket). This PR covers enough major cases to stand on its own.

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -549,138 +545,182 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
-  "convert distinct by" in {
-    val lp = fullCompileExp("select distinct(city) from zips order by pop")
-    val qs = convert(listContents.some, lp)
-    qs must beSome(beQScript(chain(
-      ReadR(rootDir </> file("zips")),
-      QC.inj(LeftShift((),
-        HoleF,
-        IncludeId,
-        Free.roll(ConcatArrays(
+    "convert a filter followed by a reduce" in {
```
alissapajer: new test from the original example in #1622

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -549,138 +545,182 @@ class QScriptSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers
...
+      val lp = fullCompileExp("select count(*) from zips where pop > 1000")
+      val qs = convert(listContents.some, lp)
+
+      qs must beSome(beQScript(chain(
+        ReadR(rootDir </> file("zips")),
+        QC.inj(LeftShift((),
+          HoleF,
+          IncludeId,
           Free.roll(ConcatArrays(
-            // FIXME: #1622
```
alissapajer: Also one of the existing tests got fixed - yay!

alissapajer: There are still a couple tests that exhibit #1622 though. I assigned this work under #1745 .

connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,440 @@
...
+import quasar.Predef.List
+import quasar.fp._
+import quasar.qscript.MapFuncs._
+import quasar.sql.CompilerHelpers
+
+import matryoshka._
+import pathy.Path._
+import scalaz._, Scalaz._
+
+class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScriptHelpers {
```
alissapajer: Basic unit tests live here.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,375 @@
...
+  implicit def projectBucket[T[_[_]]: Recursive: Corecursive, G[_]]
+      : Aux[T, ProjectBucket[T, ?], G] =
+    new Aux[T, ProjectBucket[T, ?], G] {
+
+      val helpers = new PAHelpers[T]
+      import helpers._
+
+      def find(state: StateAcc, in: ProjectBucket[T[G]]): Annotation = in match {
+        case BucketField(_, value, name) =>
+          annotateEmpty(findIndicesInFunc(value) |++| findIndicesInFunc(name) |++| state)
```
alissapajer: For example, it's really important that we include `state` in the accumulation here. So if this returned effects to be interpreted later, we could then explicitly test that the effect that represents accumulating `state` is returned, rather than implicitly testing its presence by testing the whole `BucketField` case.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,375 @@
...
+  def indexMapping(state: StateAcc): Option[Mapping] =
+    state.map(_.toList.sorted.zipWithIndex.map(_.rightMap(BigInt(_))).toMap)
+}
+
+// TODO `find` and `remap` impls should be returning a free algebra
+// which is interpreted separately
+object PruneArrays {
+  import PATypes._
+
+  sealed trait Aux[T[_[_]], IN[_], G[_]] extends PruneArrays[IN] {
```
sellout: What’s the benefit of this being a `sealed trait` rather than a `type` alias?

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -77,21 +77,24 @@ object QueryFile {
...
       QS[_]: Traverse: Normalizable]
     (implicit
-      CI:  Coalesce.Aux[T, IQS, IQS],
+      CI: Coalesce.Aux[T, IQS, IQS],
+      CQ: Coalesce.Aux[T, QS, QS],
       SP: SimplifyProjection.Aux[IQS, QS],
-      CQ:  Coalesce.Aux[T, QS, QS],
-      QC:  QScriptCore[T, ?] :<: QS,
-      TJ:    ThetaJoin[T, ?] :<: QS,
+      PA: PruneArrays.Aux[T, QS, QS],
```
sellout: So, we use this sort of Aux pattern in two ways – one, like `ShiftRead`, changes the coproduct in some way, so you end up with `ShiftRead.Aux[T, InQS, OutQS]`, but the other, like `Coalesce` and `PruneArrays` doesn’t – so we always duplicate the same functor for `F` and `G`. I feel like in these cases, we should change the Aux alias like:
```scala
type Aux[T[_[_]], F[_]] extends PruneArrays[F] {
  type IT[H[_]] = T[H]
  type IG[A] = F[A]
}
```
Might clean up a few annoying places.

sellout: Actually, I guess it has to be _two_ aliases, because we do need the current `Aux` in order to _define_ the instances, but then we want the one where F & G are the same when _resolving_ the instances.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
alissapajer: If we have `type Aux` instead of the `sealed trait` then when we want to actually create a `Aux[T, F, G]` we have to type:
```scala
new PruneArrays[F] {
  type IT[H[_]] = T[H]
  type IG[A] = G[A]
  // etc.
}
```
where as with this `trait` we can just type
```scala
new Aux[T, F, G] {
  // etc.
}
```
and don't have to repeat the boilerplate of defining `type IT[H[_]]` and `type IG[A]` for each typeclass instance. I think this could probably be a `class` just  as  well as a `trait`, thinking about it now. 

But when `type` is used, you get a "class type required" error if you try to make a `new Aux`.

sellout: Oh, that makes sense. It seems totally reasonable here, but I wonder if other cases will run into problems like slamdata/matryoshka#44.

E.g., with `Recursive.Aux[T, F]` sometimes we can use `T: Recursive` if the `F` is unconstrained, and I don’t know, if `Aux` is a trait, whether the non-Aux superclass would resolve. But now I want to try it 😄

Making it a `class` only gets in the way when you want to do something like `implicit val instance: Foo[A] with Bar[A]`, right? At least, it seemed like there was _some_ reason that type classes are `trait` and not `abstract class`.

alissapajer:
@sellout still addressing some of your initial comments, but I just pushed a commit removing the `???`. The main reason I pushed this as a wip branch was that `???` existed. So, we have a couple choices of how to proceed here.

(1) After the build passes and you (as the reviewer) give the a-okay, I rebase this branch, squashing away the commit with the `???`. The branch name will still have "wip" in it, but that's fine.

(2) We leave this as is, and treat it as a feature branch, and I will not rebase it again.

Do you have a preference or another idea? I'd rather have a single clean initial commit, and then make more review changes on top of that, so I guess that's a vote for (1). But since this is already in the review process, maybe (2) makes more sense.

connector/src/main/scala/quasar/fs/QueryFile.scala
alissapajer: I think defining as follows makes it pretty nice. This way we just require the `AuxR` outside of where we are actually defining the instances. 
```scala
type AuxR[T[_[_]], IN[_]] = Aux[T, IN, IN]
```
Change is pushed with this done.

sellout: Cool – but I would name the “public” one `Aux`. The “Aux pattern” is fairly well-known (at least among the FP-in-Scala crowd), so following the naming convention in the public API would be nice.

alissapajer:
@sellout this has been rebased and will not be rebased again! :)

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,371 @@
...
+   * Returns `None` if a non-static non-integer index was found.
+   * Else returns all indices of the form `ProjectIndex(SrcHole, IntLit(_))`.
+   */
+  def findIndicesInFunc(func: FreeMap): StateAcc = {
+    type FreeMapCoEnv[A] = CoEnv[Hole, MapFunc, A]
+
+    def accumulate: MapFunc[(T[FreeMapCoEnv], StateAcc)] => StateAcc = {
+      case ProjectIndex((src, Some(acc1)), (value, Some(acc2))) =>
+        (src.project.run, value.project.run) match {
+          case (-\/(SrcHole), \/-(Constant(ej))) => IntLit.unapplyEJson(ej).map((acc1 ++ acc2) + _)  // static integer index
```
sellout: Can we use the `IntLit` unapply here?

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,371 @@
...
+
+    func.toCoEnv.para[StateAcc](galg)
+  }
+
+  /**
+   * Remap all indices in `func` in structures like `ProjectIndex(SrcHoe, IntLit(_))`
+   * according to the provided `mapping`.
+   */
+  def remapIndicesInFunc(func: FreeMap, mapping: Mapping): FreeMap =
+    freeTransCata[T, MapFunc, MapFunc, Hole, Hole](func) {
```
sellout: There is a lot of stuff like this that will be cleaner once the new Matryoshka is used.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,371 @@
...
+    val galg: GAlgebra[(T[FreeMapCoEnv], ?), FreeMapCoEnv, StateAcc] =
+      _.run.fold(κ(Set().some), accumulate)
+
+    func.toCoEnv.para[StateAcc](galg)
+  }
+
+  /**
+   * Remap all indices in `func` in structures like `ProjectIndex(SrcHoe, IntLit(_))`
+   * according to the provided `mapping`.
+   */
```
sellout: Minor style thing, but we write comments like
```scala
/** Remap all indices in `func` in structures like `ProjectIndex(SrcHoe, IntLit(_))`
  * according to the provided `mapping`.
  */
```

alissapajer: Oh, huh, starting the text on the first line _is_ the official [scaladoc](http://docs.scala-lang.org/style/scaladoc.html)! I always thought it was how I did it :)

alissapajer: Though the style guide actually aligns comments like this:
```scala
/** Remap all indices in `func` in structures like `ProjectIndex(SrcHoe, IntLit(_))`
 *  according to the provided `mapping`.
 */
```

sellout: Yeah, so the Scala Style Guide uses that, but says you can align like we do as well. And then the Scala wiki (https://wiki.scala-lang.org/display/SW/Syntax) says to use our style.

¯\\\_(ツ)_/¯

connector/src/main/scala/quasar/qscript/PruneArrays.scala
alissapajer: @sellout so, the `IntLit.unapply` (as it stands) expects a `FreeMapA[T, A]` but in the RHS of the `Either` here we expect a `MapFunc[T, T[FreeMapCoEnv]]`. So one proposal is as follows:

Delete the `unapplyEJson` function that this PR adds, and instead have the following unapplys:
```scala
888   object IntLit {
889     def apply[T[_[_]]: Corecursive, A](i: BigInt): FreeMapA[T, A] =
890       Free.roll(Constant[T, FreeMapA[T, A]](EJson.fromExt[T].apply(ejson.Int[T[EJson]](i))))
891 
892     def unapply[T[_[_]]: Recursive, A](mf: FreeMapA[T, A]): Option[BigInt] =
893       mf.resume.fold(IntLitMF.unapply(_), _ => None)
894   }
895 
896   object IntLitMapFunc {
897     def unapply[T[_[_]]: Recursive, A](mf: MapFunc[T, A]): Option[BigInt] = mf match {
898       case Constant(ej) => ExtEJson.prj(ej.project).flatMap {
899         case ejson.Int(i) => i.some
900         case _ => None
901       }
902       case _ => None
903     }
904   }
905 
906   object IntLitCoEnv {
907     def unapply[T[_[_]]: Recursive](coenv: T[CoEnv[Hole, MapFunc[T, ?], ?]]): Option[BigInt] =
908       IntLit.unapply(coenv.fromCoEnv)
909   }
```
Then this case matches nicely as:
```scala
case (-\/(SrcHole), \/-(IntLitMapFunc(idx))) => ???
```
and another case in this file matches nicely with the `IntLitCoEnv`.

What do you think of this approach?

This also illustrates complications in pattern matching on this stuff, when we have `CoEnv` or `Free` or neither, all used to more or less represent the same underlying structure.

alissapajer: @sellout I pushed the change that does the above because I had it all compiling and I think it's an improvement on what I had before. But I'm happy to make further changes after further discussion!

sellout: I _think_ you could actually put all of them in the same companion object, just might need to suppress the Overloading wart.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -0,0 +1,371 @@
...
+   * @param out the value which replaces the current focus
+   */
+  @Lenses final case class Output[T[_[_]], F[_], G[_]](newState: StateAcc, out: F[T[G]])
+}
+
+@typeclass trait PruneArrays[F[_]] {
+  import PATypes._
+
+  type IT[H[_]]
+  type IG[A]
```
alissapajer: These types are here for a couple reasons:

1. I wanted to require, at the level of the typeclass definition, that the `in` parameter for both `find` and `remap` and the `out` in `remap` are all of _exactly_ the same type, namely `F[IT[IG]]`. If we move the types down to each of the operations, then we lose the information that the `IT` and the `IG` are necessarily the same in both `find` and `remap`.

2. Another idea would be to remove `IG`, and require `F[IT[F]]` instead of `F[IT[IG]]`. In practice, meaning when we actually call instances of the typeclass, `F` and `IG` are always equal. But when defining the `Coproduct` instance, it seemed natural (and maybe necessary?) to have an inner type `IT[IG]` different from the outer type `F[_]`. Because then we just require instances for each of the `Coproduct` sides, and we're done. Now, the fact that a `Coproduct` instance exists at all should not influence the actual definition of the typeclass, which is exactly what I just argued should happen... 

So with all that said, we could do:
```scala
 59   def find[A](state: StateAcc, in: F[A]): Annotation
 60   def remap[A](env: StateAcc, state: StateAcc, in: F[A]): Output[F, A]
```
and then `A` would be `T[G]` at the call sites for `find` and `remap`. Looking at that, I'm not sure my point 1. above is all that valid... So, what do you think of the above? It lets us nicely define the `Coproduct` instance, and we remove the `IT` and `IG` types entirely.

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -77,21 +77,24 @@ object QueryFile {
...
     val rewrite = new Rewrite[T]
+    val pa = new PAFindReplace[T, QS]
 
     // TODO: This would be `transHylo` if there were such a thing.
     _.transAna(SP.simplifyProjection)
       // TODO: Rather than explicitly applying multiple times, we should apply
       //       repeatedly until unchanged.
       .transAna(rewrite.normalize)
       .transAna(rewrite.normalize)
+      .hyloM(pa.remapIndices, pa.findIndices).run(None)._2
```
alissapajer: Also, this line is duplicated here and in the class `PAFindReplace` so I want to address that.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
alissapajer: Also, then we don't need any `Aux` types at all.

sellout: Yeah, I think the most parametric function that works.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
alissapajer: I'd prefer to keep them in separate objects because I like to avoid overloading, and that forces the user to know what function they are calling, rather than have the correct function just get found.

sellout: Yeah, I’m fine with that. The only reason I was thinking the other way in this case is that the Matryoshka update should eliminate `IntLitCoEnv` and mutual-recursion should eliminate `IntLit`, leaving only `IntLitMapFunc`.

alissapajer:
@sellout I've pushed changes that address your comments. I still want to remove that `hyloM` duplication and I have a few more unit tests in progress, so when I'm done with what I have planned to push I'll let you know.

wemrysi:
@alissapajer Nice work, also looks like this has a healthy effect on performance if the build times are any indication!

connector/src/main/scala/quasar/qscript/PruneArrays.scala
sellout: Yeah, I think that sounds good. And no more `Aux` is so nice 😄

connector/src/main/scala/quasar/fs/QueryFile.scala
sellout: Right, I meant to comment on that and forgot. Gotta take more advantage of the new review format where I can make a pending comment and delete it later if it turns out to be answered elsewhere in the code.

Oh, also, about moving this earlier for performanc … does it work to do so? I.e., we don’t depend on the structure being normalized for PruneArrays to work properly?

alissapajer:
Oh yeah, the couchbase build is down from about 1:45 to 0:55. Cool.

alissapajer:
@selout ok I've pushed a few more tests and some cleanup. Back in your court.

alissapajer:
Fixes #1745

PR #1756, opened 2016-12-05
[1717] Add a META function providing access to any metadata associated with a value

### LINKED ISSUE #1701: Ability to address XML attributes in MarkLogic
Enable projecting and predicates on XML attributes in queries against MarkLogic.

### LINKED ISSUE #1717: Add META(d) function
The `META(d)` function lifts up the metadata associated with `d` (possibly `undefined`, if there is no such metadata) into the value-level, where it may be accessed and processed like other data.

For accessing attributes on XML data (#1701), we can use `META(node).className` for example. I believe metadata on XML nodes will always be a map of string-valued keys.

Ultimately, the pattern matching syntax for metadata can be implemented on top of this, although an alternate function `SET_METADATA` will be required to actually set metadata on a node (which would be used for metadata native syntax, `foo @ bar`).

wemrysi:
The `META` function returns any metadata associated with the argument or `Undefined` otherwise.

Implementations for Couchbase (via the `meta()` function) and MarkLogic are included, MongoDB and Spark didn't seem to have an obvious (to me) implementation (@sellout @rabbitonweb feel free to suggest one).

I've added tests for the MarkLogic implementation but testing this in general via `StdLibSpec` requires `EJson` in `LogicalPlan` as we don't have a way to represent metadata in our `Data` type.

Closes #1717.
Closes #1701.

### DIFF
--- a/frontend/src/test/scala/quasar/std/StdLibSpec.scala
+++ b/frontend/src/test/scala/quasar/std/StdLibSpec.scala
@@ -1035,6 +1035,11 @@ abstract class StdLibSpec extends Qspec {
           binary(ConcatOp(_, _).embed, Data._str(x), Data._str(y), Data._str(x + y))
         }
       }
+
+      "Meta" >> {
+        // FIXME: Implement once we've switched to EJson in LogicalPlan.
+        "returns metadata associated with a value" >> skipped("Requires EJson.")
+      }
     }
   }
 }
--- a/CHANGELOG/feature.md
+++ b/CHANGELOG/feature.md
@@ -1,0 +1,3 @@
+- [1717] Add a META function for accessing metadata associated with a value
+  - [1701] Implement for MarkLogic which returns attributes of XML elements.
+  - Implement for Couchbase using the meta() function.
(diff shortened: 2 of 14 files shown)


wemrysi:
@drostron Let me know if the Couchbase implementation makes any sense and if there are any tests that should/could be added.

@sellout If you could have a look at the `Meta` typer/untyper I'd appreciate it, there didn't seem to be anything else to do there but it'd be good to have confirmation of this.

drostron:
@wemrysi The Couchbase implementation makes sense. I'm planning on enabling N1QL's key joins via `EquiJoin` and `Meta` soon and that will bring some testing to the mix, so ok to leave it without for the moment. Thanks for adding `Meta` — will be nice to finally bring a Join facet to CB.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -1035,6 +1035,11 @@ abstract class StdLibSpec extends Qspec {
           binary(ConcatOp(_, _).embed, Data._str(x), Data._str(y), Data._str(x + y))
         }
       }
+
+      "Meta" >> {
+        // FIXME: Implement once we've switched to EJson in LogicalPlan.
```
sellout: Hopefully we’ll be able to define StdLibSpec directly in MapFunc before LP uses EJson.

wemrysi: Ah, yeah, or that ;)

wemrysi:
@drostron I think this is ready to merge, @rabbitonweb confirmed with me out-of-band that he is fine with the changes (@rabbitonweb feel free to add anything here if I've misrepresented you).

PR #1765, opened 2016-12-07
[1759] Executing a query should overwrite an existing file in MarkLogic

### LINKED ISSUE #1759: Executing a query to an existing file should overwrite contents
Currently, if the results destination for a query already exists an error is raised, instead the results should replace the existing file.

This behavior isn't specified anywhere (will be remedied as part of this fix), but is the current behavior for the legacy MongoDB connector.

wemrysi:
This doesn't seem to ever have been specified, so I've added a test to `QueryFilesSpec` that does so, essentially codifying the existing MongoDB connector behavior.

SlamData currently expects this behavior (for the cache card, which is how I discovered it), so if we want to change it we should probably do so in a breaking release.

Fixes #1759.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+[1759] Executing a query should overwrite an existing file in MarkLogic
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ For compressed output use `Accept-Encoding: gzip`.
 
 Executes a SQL² query, contained in the request body, on the backend responsible for the request path.
 
-The `Destination` header must specify the *output path*, where the results of the query will become available if this API successfully completes.
+The `Destination` header must specify the *output path*, where the results of the query will become available if this API successfully completes. If the output path already exists, it will be overwritten with the query results.
 
 All paths referenced in the query, as well as the output path, are interpreted as relative to the request path, unless they begin with `/`.
 
--- a/connector/src/main/scala/quasar/fs/QueryFile.scala
+++ b/connector/src/main/scala/quasar/fs/QueryFile.scala
@@ -174,8 +174,9 @@ object QueryFile {
     }
   }
 
-  /** The result of the query is stored in an output file
-    * instead of being returned to the user immediately.
+  /** The result of the query is stored in an output file, overwriting any existing
+    * contents, instead of being returned to the user immediately.
+    *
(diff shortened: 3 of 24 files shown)


wemrysi:
Hm, looks like I need to figure out how to skip postgres and the mongo read-only connection as well, maybe a broader predicate than just "is external".

@drostron Looks like Couchbase appends to the existing file when it exists, I don't disagree with this behavior, but would it be possible to have it overwrite if we decide to preserve the existing behavior?

If we think appending is the proper behavior, then we can see what sort of implications that has for the frontend and go from there.

I think you can achieve similar effects with both behaviors, e.g. appending multiple queries to the same output file with the "appends" behavior is straightforward, but you could also do a `UNION` query and achieve the same effect when the "replace" behavior.

We could also do neither and just error if the file already exists, or add some sort of "strategy" parameter like we have with move.

Lastly, we probably want the http verb to match the semantics, so if we go with append we can keep `POST` but if we go with replace we should probably change to a `PUT` for this case.

Reminds me that I should also update comments and docs with whatever we decide on as well.

drostron:
Offering a way for a connector declare what it supports might prove useful, but for now a broader predicate would suffice.

@wemrysi Overwrite is possible and fine for Couchbase depending on what we decide. The current overwriting behavior appears to suit the needs of users therefore staying with that and updating the verb to match sounds reasonable.

wemrysi:
@drostron I went ahead and added a way for test backends to declare their capabilities, it is fairly coarse at the moment but could be refined if needed.

I also updated the Couchbase `ExecutePlan` impl to overwrite and added docs.

After looking at the http api, I think it should remain as-is. I forgot that the URI path in the query is not the destination of the results, but the context for the query so I think a `POST` is still the appropriate verb.

wemrysi:
Still awaiting travis (pushing a `[skip ci]` commit appears to trash the latest build state).

drostron:
@wemrysi Looks like `BackendCapability` did the trick nicely. Good to see a few adhoc exclusions cleaned up.

Thanks for updating Couchbase's `ExecutePlan` and appreciate the additional clarity in the docs.

Leaving the verb as is makes sense in light of what you mention.

drostron:
[build pass](https://travis-ci.org/quasar-analytics/quasar/builds/182436913)

PR #1767, opened 2016-12-08
Fix nested array deref bug.

alissapajer:
There should be some integration tests that can be enabled with this fix. I'd like to see what happens when integration is run - not sure if tests will come to life (they're pending) or if I need to go in there and turn them on (they're skipped) or if I need to add some integration tests (the necessary ones don't exist yet). 

In any case, a good example query of what now comes to life is: `select distinct(city) from zips order by pop`. That query requires actually using the reified bucketing information in order to compute the sorting buckets. (There is just one bucket, but we have to give that information to the backend!)

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Provenance.scala
+++ b/connector/src/main/scala/quasar/qscript/Provenance.scala
@@ -128,7 +128,6 @@ class ProvenanceT[T[_[_]]: Corecursive: EqualT] extends TTypes[T] {
   def genBuckets(ps: List[Provenance]): Option[(List[Provenance], FreeMap)] =
     ps.traverse(genBucket).eval(0).unzip.traverse(_.join match {
       case Nil      => None
-      case h :: Nil => Free.roll(MakeArray[T, FreeMap](h)).some
       case h :: t   =>
         t.foldLeft(
           Free.roll(MakeArray[T, FreeMap](h)))(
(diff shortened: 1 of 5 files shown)


alissapajer:
This aims to take care of #1622.

connector/src/main/scala/quasar/qscript/Provenance.scala
```diff
@@ -128,7 +128,7 @@ class ProvenanceT[T[_[_]]: Corecursive: EqualT] extends TTypes[T] {
   def genBuckets(ps: List[Provenance]): Option[(List[Provenance], FreeMap)] =
     ps.traverse(genBucket).eval(0).unzip.traverse(_.join match {
       case Nil      => None
-      case h :: Nil => h.some
+      case h :: Nil => Free.roll(MakeArray[T, FreeMap](h)).some
```
sellout: This case can just be removed then, right? The foldLeft should handle it.

alissapajer: Oh, ha, yes, good point! :)

alissapajer:
@sellout as might have been clear from the initial comment, even if the build passes, I'd still like to investigate the integration tests and make sure this is tested there before merge.

alissapajer:
Ok there are a couple real couchbase+marklogic integration test errors which I'll look into!

rabbitonweb:
@alissapajer I can tests my `select distinct` query that was failing.

alissapajer:
@rabbitonweb please do and let me know! What is the full query statement?

it/src/main/resources/tests/guardedExpression.test
```diff
@@ -1,6 +1,7 @@
 {
   "name": "regex on non-string field",
   "backends": {
+      "couchbase":         "skip",
```
alissapajer: Spoke with @drostron about this, and the pertinent code is being rewritten, so we decided to skip this test for now. 

This test is passing the marklogic build, and the qscript looks correct, so it's likely a couchbase issue.

alissapajer:
That last commit should fix the failing ML test. It turns out that that test was passing because the two bugs fixed in this PR were negating each other... We effectively generated a "null" bucketing by generating an impossible bucketing that ML turned to undefined.

alissapajer:
@sellout green build - this is ready for review now

PR #1770, opened 2016-12-09
Get your N1QL fix

drostron:
@sellout There are easily a few matryoshka-isms available that I've missed. Please let me know if/where you see those opportunities.


couchbase/src/main/scala/quasar/physical/couchbase/N1QL.scala
```diff
@@ -17,111 +17,142 @@
...
-    unnest: Option[(String, String)],
-    orderBy: Option[String]
-  ) extends N1QL {
-    def n1ql: N1QL = this
+import quasar.{Data => QData}
+import quasar.common.SortDir
+
+import scalaz._, NonEmptyList.nels, OneAnd.oneAnd
+
+sealed trait N1QL[A]
```
sellout: This should be a `sealed abstract class`

couchbase/src/main/scala/quasar/physical/couchbase/N1QL.scala
```diff
@@ -17,111 +17,142 @@
...
+
+@SuppressWarnings(Array("org.wartremover.warts.Overloading"))
```
sellout: This needs a comment as to why/what is being overloaded

couchbase/src/main/scala/quasar/physical/couchbase/N1QLRenderQueryInstance.scala
```diff
@@ -0,0 +1,209 @@
...
+import scalaz._, Scalaz._
+
+trait N1QLRenderQueryInstance {
+  import N1QL._, Case._, Select._
+
+  implicit def renderQuery[T[_[_]]: Recursive: Corecursive]: RenderQuery[N1QLT[T]] =
+    new RenderQuery[N1QLT[T]] {
+      implicit val codec = common.CBDataCodec
+
+      def compact(a: N1QLT[T]): PlannerError \/ String = {
```
sellout: You should pull out the algebra as `AlgebraM[PlannerError \/ ?, N1QL, String]`, then define a function that’s just 
```scala
def compactTree(n1ql: T[N1QL]): PlannerError \/ String = n1ql.cataM(compact)
```

couchbase/src/main/scala/quasar/physical/couchbase/N1QLRenderQueryInstance.scala
```diff
@@ -0,0 +1,209 @@
...
+        val q = a.embed.cataM[PlannerError \/ ?, String] {
```
sellout: You shouldn’t have to call `.embed` before a fold – in general, you should be passing around values of `T[N1QL]`, not `N1QL[T[N1QL]]` You usually only see the projected form inside algebras.

couchbase/src/main/scala/quasar/physical/couchbase/RenderQuery.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package quasar.physical.couchbase
+
+import quasar.Predef._
+import quasar.Planner.PlannerError
+
+import scalaz._
+import simulacrum.typeclass
+
+@typeclass trait RenderQuery[A] {
```
sellout: Why make this a type class? It seems like there is only intended to be one instance.

couchbase/src/main/scala/quasar/physical/couchbase/fs/queryfile.scala
```diff
@@ -50,8 +53,16 @@ object queryfile {
...
+  implicit class LiftF[F[_]: Monad, A](p: F[A]) {
+    val liftF = p.liftM[PhaseResultT].liftM[FileSystemErrT]
+  }
+
+  implicit class LiftFE[A](d: FileSystemError \/ A) {
+    def liftFE[S[_]]: Plan[S, A] = EitherT(d.point[Free[S, ?]].liftM[PhaseResultT])
+  }
+
+  implicit class liftPE[A](d: PlannerError \/ A) {
+    def liftPE[S[_]]: Plan[S, A] = d.leftMap(FileSystemError.qscriptPlanningFailed(_)).liftFE
```
sellout: For the future … might be worth thinking about MTL classes.

couchbase/src/main/scala/quasar/physical/couchbase/package.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.physical
+
+package object couchbase {
+
+  type N1QLT[T[_[_]]] = N1QL[T[N1QL]]
```
sellout: Need for this may go away once you switch to `T[N1QL]` more often.

couchbase/src/main/scala/quasar/physical/couchbase/fs/queryfile.scala
```diff
@@ -190,18 +197,19 @@ object queryfile {
     } yield exists).exists(ι)
 
   def n1qlResults[S[_]](
-    n1ql: N1QL
+    n1ql: N1QLT[Fix]
```
sellout: More parametricity! 😉 Would be nice to see `Fix` gone as much as possible.

couchbase/src/main/scala/quasar/physical/couchbase/planner/EquiJoinPlanner.scala
```diff
@@ -36,41 +31,10 @@ import scalaz._, Scalaz._
...
-                      |  rKey:    $rkN1ql
-                      |  f:       $f
-                      |  combine: $cN1ql
-                      |  n1ql:    ???""".stripMargin('|'))))
-    } yield ()) *>
-    unimplementedP("EquiJoin")
+final class EquiJoinPlanner[T[_[_]]: Recursive: Corecursive: ShowT, F[_]: Monad: NameGenerator]
+  extends Planner[T, F, EquiJoin[T, ?]] {
 
+  def plan: AlgebraM[M, EquiJoin[T, ?], N1QLT[T]] = {
```
sellout: Ah, and this is maybe the root of the N1QLT problem – you may want to redefine `plan` from an algebra to a transformation.

As an algebra, you want `AlgebraM[M, EquiJoin[T, ?], T[N1QL]]` (`EquiJoin[T, T[N1QL]] => M[T[N1QL]]` rather than the current `EquiJoin[T, N1QL[T[N1QL]]] => M[N1QL[T[N1QL]]]]`), which means putting `.embed` at the end somehow (either in each case, or `>>> (_.embed)` or whatever) rather than calling `.embed` on each argument on the way in. But if you define it as a transformation, you can do `EquiJoin[T, T[N1QL]] => M[N1QL[T[N1QL]]]`, which is pretty close to what you have (except you get to skip the extra `.embed`) calls, then you just use `transCataM` instead of `cataM` when you apply it.

couchbase/src/main/scala/quasar/physical/couchbase/planner/Planner.scala
```diff
@@ -23,60 +23,64 @@ import quasar.physical.couchbase._
...
 
-abstract class Planner[F[_], QS[_]] {
+abstract class Planner[T[_[_]], F[_], QS[_]] {
   type M[A]  = CBPhaseLog[F, A]
   type PR[A] = PhaseResultT[F, A]
 
-  def plan: AlgebraM[M, QS, N1QL]
+  implicit class PointM(v: N1QLT[T])(implicit M: Applicative[M]) {
+    def ηM: M[N1QLT[T]] = v.η[M]
+  }
```
sellout: This only seems to be used in one place, I’ll comment there more …

couchbase/src/main/scala/quasar/physical/couchbase/planner/ReduceFuncPlanner.scala
```diff
@@ -17,60 +17,24 @@
...
-      )
+  def plan: AlgebraM[M, ReduceFunc, N1QLT[T]] = {
+    case RF.Arbitrary(a1)      => Min(a1.embed).ηM
+    case RF.Avg(a1)            => Avg(a1.embed).ηM
+    case RF.Count(a1)          => Count(a1.embed).ηM
+    case RF.Max(a1)            => Max(a1.embed).ηM
+    case RF.Min(a1)            => Min(a1.embed).ηM
+    case RF.Sum(a1)            => Sum(a1.embed).ηM
+    case RF.UnshiftArray(a1)   => ArrAgg(a1.embed).ηM
+    case RF.UnshiftMap(a1, a2) => Obj(Map(a1.embed -> a2.embed)).ηM
```
sellout: Yeah, see, here are some of the places with the unnecessary `.embed`s.

sellout: I think it would be much nicer to see this defined as
```scala
val planʹ: AlgebraicTransform[T, ReduceFunc, N1QL] = { … } // I probably got the type params wrong
def plan = planʹ >>> (_.point[M])
```
that shows that although there are instances that can fail, this particular one is total.

drostron:
@sellout Excellent, feedback appreciated.

drostron:

Issue #1762, opened 2016-12-07
Spark can not compare timestamp and dates


rabbitonweb:
Tests affected are: temporal/filterOnDate.test, complex-vars.test

rabbitonweb:
It was not a bug, but a feature :)

rabbitonweb:
It was not a bug, but a feature :)

PR #1771, opened 2016-12-09
Allow compareing Timestamp & Date in queries

rabbitonweb:
Changes needed to fix https://github.com/quasar-analytics/quasar/issues/1762


rabbitonweb:
This is being closed as implementation is correct. The tests were incorrect (you can not compare date and timestamp)

PR #1777, opened 2016-12-10
Enable some integration tests for marklogic & couchbase.

alissapajer:
Now that #1745 and #1767 have merged, some more integration tests for marklogic and couchbase may pass.

Additionally since couchbase tests are running faster, I removed the all the "skipCI" flags to see if any of those now run fast enough.


alissapajer:
Just a note: I didn't look at all the integration tests in detail to determine which ones should/could be passing. I just grabbed a bunch that looked promising, and we'll take it from there.

sellout:
So, like pending in specs2, pending in regression should fail if the test passes … unfortunately, we disable running of pending tests in Travis, so we don't get alerted when a test is fixed.

That wasn't a problem when there was one connector that we routinely tested locally, but it's more of a problem now.

alissapajer:
@sellout yeah, ideally we could  have _all_ the integration tests pending, and then we'd just be alerted when they pass by a failing build. If we had that, a PR like this wouldn't be necessary at all. (Except we still might need to skip tests on travis that run too slowly but eventually pass.)

alissapajer:
Have we disabled running pending tests because they were throwing exceptions and causing failures? Or because they were running too slowly?

Would it be feasible to re-enable the running of pending tests on Travis, and then still allow the "skipCI" flag for tests that are troublesome when run?

sellout:
We did it for coverage correctness – there's something like a `PendingWithCorrectCoverage` class. But … we (at least for now) only track coverage on the Mongo 2.6 build. So I think we could adjust the PWCC class to only disable tests on that one build.

rabbitonweb:
Does it also include spark? I know we don't have it yet on CI, but maybe I can also check Spark on my side?

alissapajer:
@rabbitonweb this does not include spark because we're not running it on travis yet. If more spark tests work locally, great :) But they won't be run regularly until they are run on travis, meaning if someone breaks one inadvertently, they won't know.

alissapajer:
Closing to open with single commit.

alissapajer:
Reopened as #1779

PR #1779, opened 2016-12-13
Enable some more CB and ML integration tests.

alissapajer:
Removes "skipCI" from a handful of CB tests, as well as enabling some others.

Performance improved at least in part from #1745 and #1767


wemrysi:
:+1: LGTM, always nice to see more tests passing.

PR #1780, opened 2016-12-14
Update to the latest version of Matryoshka.

sellout:
There have been a ton of upstream changes (notably directly-recursive
support), so this is not a small update.


sellout:
There are a ton of small changes, the lazy ones are like `Recursive` -> `RecursiveT`, but that sort of change is also necessary for the way we fake mutual recursion. The less lazy ones are like `[T[_[_]]: Recursive, F[_]]` -> `[T, F[_]](implicit T: Recursive.Aux[T, F])`, which we’ll be pushing through the code incrementally.

Also, all the conversion between `Free[F, A]` and `T[CoEnv[A, F, ?]]` is gone now (except for one case). So no more `toCoEnv` or `freeCataM` or whatever.

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -1770,68 +1846,68 @@ object WorkflowBuilder {
...
-              }) :: nt)))
-        case SpliceBuilderF(src, structure) =>
-          NonTerminal("SpliceBuilder" :: nodeType, None,
-            render(src) :: structure.map(RC.render))
-        case ArraySpliceBuilderF(src, structure) =>
-          NonTerminal("ArraySpliceBuilder" :: nodeType, None,
-            render(src) :: structure.map(RC.render))
-      }
-    }
+  // implicit def WorkflowBuilderRenderTree[F[_]: Coalesce: Functor]
```
drostron: Any reason to keep this around as a comment?

drostron:
A spiffy update.

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -169,8 +169,8 @@ object Sql {
     }
   }
 
-  implicit def toExprAlgebraOps[A](a: Algebra[Sql, A]): AlgebraOps[Sql, A] =
-    toAlgebraOps[Sql, A](a)
+  // implicit def toExprAlgebraOps[A](a: Algebra[Sql, A]): AlgebraOps[Sql, A] =
```
sellout: Yep. This was a workaround for a bug in an old version of Matryoshka, and yay, it’s fixed.

frontend/src/main/scala/quasar/std/structural.scala
```diff
@@ -352,10 +359,13 @@ trait StructuralLib extends Library {
     Func.Input2(Top, Top),
     noSimplification,
     // new Func.Simplifier {
```
sellout: Good question … I think this is a valid simplification if the first argument (currently ignored) is actually `ShiftMapKeys` of the same input map. So,
```scala
case InvokeUnapply(
  Sized(Embed(InvokeUnapply(ShiftMapKeys, Sized(Embed(map1))))),
  Sized(Embed(InvokeUnapply(ShiftMap, Sized(Embed(map2))))))
    if map1 ≟ map2 =>
  map1.some
```
But I don’t think that’s something I want to play with in this PR.

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
sellout: Nope. It should be gone.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -77,11 +78,11 @@ object queryfile {
       for {
         qs <- QueryFile.convertToQScriptRead[Fix, FileSystemErrT[PhaseResultT[Free[S, ?],?],?], QScriptRead[Fix, ?]](lc)(lp)
           .map(simplifyRead[Fix, QScriptRead[Fix, ?], QScriptShiftRead[Fix, ?], SparkQScript].apply(_))
-        optQS = qs.transAna(
+        optQS = qs.transHylo(
```
sellout: Ok, I _might_ have solved this. If so, it’ll be in this PR.

sellout: Ok, so I pushed a fix … we’ll see if it breaks any of the connectors …

drostron:
Nice to see `finalizeQScript`. 👍🏾, will merge upon Travis approval.

sellout:
Ok, reverted that over-optimistic fix 😢

drostron:
I'll keep hope for that idea's future. Stale comment pruning was also reverted.

drostron:
Needs a changelog.

sellout:
*sigh* Ok, changelog’d

PR #1785, opened 2016-12-15
Fix non-static array projection prefix

### LINKED ISSUE #1688: QScript missing constant array for `projectFromConcattedArray.test`
While looking at enabling `projectFromConcattedArray.test` I noticed that `[1, 2]` is missing from the generated `QScript`.

SQL²:
```
select (loc || [1, 2])[1] from largeZips
```

lp:
```
Invoke(Take)
├─ Invoke(Squash)
│  ╰─ Invoke(MakeObject)
│     ├─ Str("0")
│     ╰─ Let('__check0)
│        ├─ Read(/quasar-test/largeZips.json)
│        ╰─ Typecheck(Obj(Map[], Some(Top)))
│           ├─ Free('__check0)
│           ├─ Let('__check1)
│           │  ├─ ObjectProject('__check0{"loc"})
│           │  ╰─ Typecheck(FlexArr(0, None, Top))
│           │     ├─ Free('__check1)
│           │     ├─ Invoke(ArrayProject)
│           │     │  ├─ Invoke(ArrayConcat)
│           │     │  │  ├─ Free('__check1)
│           │     │  │  ╰─ Constant(Arr(List(Int(1), Int(2))))
│           │     │  ╰─ Constant(Int(1))
│           │     ╰─ Constant(NA)
│           ╰─ Constant(NA)
╰─ Constant(Int(11))
```

QScript:
```
Subset(Take)
├─ Unreferenced
├─ From
│  ╰─ LeftShift
│     ├─ Read(/quasar-test/largeZips.json)
│     ├─ Struct
│     │  ╰─ ○
│     ╰─ Repair
│        ╰─ MakeMap
│           ├─ Constant(Str(0))
│           ╰─ Guard
│              ├─ RightSide
│              ├─ Type(Obj(Map[], Some(Top)))
│              ├─ Guard
│              │  ├─ ProjectField
│              │  │  ├─ RightSide
│              │  │  ╰─ Constant(Str(loc))
│              │  ├─ Type(FlexArr(0, None, Top))
│              │  ├─ Constant(Int(2))
│              │  ╰─ Undefined
│              ╰─ Undefined
╰─ Count
   ╰─ Map
      ├─ Unreferenced
      ╰─ Constant(Int(11))
```

alissapajer:
Fixes #1688

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Coalesce.scala
+++ b/connector/src/main/scala/quasar/qscript/Coalesce.scala
@@ -58,23 +58,23 @@ trait Coalesce[IN[_]] {
 }
 
 trait CoalesceInstances {
-  def coalesce[T[_[_]]: Recursive: Corecursive: EqualT] = new CoalesceT[T]
+  def coalesce[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] = new CoalesceT[T]
 
-  implicit def qscriptCore[T[_[_]]: Recursive: Corecursive: EqualT, G[_]]
+  implicit def qscriptCore[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, G[_]]
     (implicit QC: QScriptCore[T, ?] :<: G)
       : Coalesce.Aux[T, QScriptCore[T, ?], G] =
     coalesce[T].qscriptCore[G]
 
-  implicit def projectBucket[T[_[_]]: Recursive: Corecursive: EqualT, F[_]]
+  implicit def projectBucket[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, F[_]]
       : Coalesce.Aux[T, ProjectBucket[T, ?], F] =
     coalesce[T].projectBucket[F]
 
-  implicit def thetaJoin[T[_[_]]: Recursive: Corecursive: EqualT, G[_]]
+  implicit def thetaJoin[T[_[_]]: Recursive: Corecursive: EqualT: ShowT, G[_]]
     (implicit TJ: ThetaJoin[T, ?] :<: G)
       : Coalesce.Aux[T, ThetaJoin[T, ?], G] =
     coalesce[T].thetaJoin[G]
 
-  implicit def equiJoin[T[_[_]]: Recursive: Corecursive: EqualT, G[_]]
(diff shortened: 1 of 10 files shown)


connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -107,20 +89,38 @@ object MapFunc {
         Option[List[TCoMapFunc[T, A]]] =
       mf match {
         case ConcatArraysN(as) =>
-          as.foldRightM[List[TCoMapFunc[T, A]] \/ ?, List[TCoMapFunc[T, A]]](
+          as.foldLeftM[List[TCoMapFunc[T, A]] \/ ?, List[TCoMapFunc[T, A]]](
```
alissapajer: The bug was here. If we `foldRightM` then we associate from the right, and look for failures (i.e. `Either.left`) right to left. This would produce the static suffix instead of the static prefix.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -58,23 +58,23 @@ trait Coalesce[IN[_]] {
 }
 
 trait CoalesceInstances {
-  def coalesce[T[_[_]]: Recursive: Corecursive: EqualT] = new CoalesceT[T]
+  def coalesce[T[_[_]]: Recursive: Corecursive: EqualT: ShowT] = new CoalesceT[T]
```
sellout: I feel like we should avoid this proliferation of `Show` everywhere for debugging. If we have a Recursive constraint, we can do `Recursive.show.show(qs)` when we want to debug something without having to thread Show through. Similarly with `RenderTree.recursive.render(qs)`.

Of course, that still requires a `Delay[Show, F]` for whatever the type of `qs` is. In these cases, I think that’s fine, because it’s all concrete types. It’s only in cases where we have some unknown `F[_]` that that crops up. But we should think about a less pervasive way to enable that.

sellout:
I’ll merge once the MarkLogic test is fixed.

alissapajer:
Spoke with @wemrysi and decided to skip the failing ML test for now. It's returning an empty results, so there is not a quick thing to try, though a SQL alias will probably be at least part of the solution.

alissapajer:
@sellout this is ready now.

PR #1788, opened 2016-12-15
Improvements to `fileSystemShould`

jedesah:
- List all backends for which the integration tests were skipped for lack of a proper testing environment
- Do not fail integration tests if no external backend is configured
- run tests on internal backends, even if no external backend is configured


jedesah:
I would have liked to print a message like `"marklogic was skipped"` instead of just the string `"marklogic"` with nothing after it. We'll see if [this SO question](http://stackoverflow.com/questions/41174047/how-to-skip-an-implementation-with-specs2) gets a satisfactory response.

jedesah:
So I followed the suggestion in the answer to the SO question and [this](https://github.com/jedesah/Quasar/commit/1713d07749d38024553c7e7cdd58adc027545f2b) is what I cam up with. Not sure if it's an improvement.

The output is nicer, but the implementation is not so nice.

jedesah:
@sellout Let me know what you think

jedesah:
@sellout ping

it/src/test/scala/quasar/fs/FileSystemTest.scala
```diff
@@ -59,14 +59,16 @@ abstract class FileSystemTest[S[_]](
...
-      ()
-    }).unsafePerformSync
+  def fileSystemShould(examples: FileSystemUT[S] => Fragment): Fragments =
+    fileSystems.map { fss =>
+      Fragments.foreach(fss.toList)(fs =>
+        fs.impl.map { f =>
+          s"${fs.ref.name.shows} FileSystem" >>
+            Fragments(examples(f), step(f.close.unsafePerformSync))
+        } getOrElse {
+          Fragments(s"${fs.ref.name.shows} FileSystem" >> skipped("Environment not setup to test this file system"))
```
sellout: Could you add the particular environment variable name to this message?, Like “Set the $QUASAR_MONGODB_3_2 environment variable to test this file system,” perhaps.

jedesah: I thought of that but I shied away from it for now because it's not easy to do in a DRY way. In any case, I want to explore getting rid of environment variables in favor of sbt `SettingKey`. However, I just realized that I should probably not sacrifice "functionality" for that reason. I'll make the non-DRY change and revisit it later. I might even be able to make it DRY more easily than I expect.

jedesah:
Nevermind. Turned out to be way easier than I had somehow imagined it was.

This should be ready to merge now.

PR #1795, opened 2016-12-19
Mtl for planner

rabbitonweb:
Refactor Planner to use MTL. 

The `M[A]` here is for now a `StateT[EitherT[Task, PlannerError, ?], SparkContext, A]` but this will be soon changed to `Free`

Also moved `MonadError_`


rabbitonweb:
@here I guess I will close this MTL branch now. There is nothing I can do about it for now. I will try to provide Kryo serializer to our Spark connector so all future 'Serializable' issues should go away.

Any thoughts before I close this branch?

sellout:
Looks like you never closed this – is that still the plan, or did you sort out the Serializable stuff?

rabbitonweb:
I'm closing this PR. MTL brings much more readability but with current Spark serialization issues (Scalaz is not serializable) it's impossible to apply MTL.

I might do a second approach with this after I will introduce a Kryo serialization to sparkcore - see https://github.com/quasar-analytics/quasar/issues/1926

PR #1799, opened 2016-12-20
Various bug fixes.

### LINKED ISSUE #1541: Verify basic analytic query in QScript
Verify that QScript generated by a basic analytic query is optimized correctly.

``` sql
SELECT 
  state, 
  COUNT(*) AS count, SUM(pop) AS sum, AVG(pop) AS avg, MIN(pop) AS min, MAX(pop) AS max
FROM 
  `/zips/`
WHERE 
  pop > 100000
GROUP BY
  state
ORDER BY
  max DESC, state
OFFSET 1
LIMIT 100
```

### LINKED ISSUE #1710: Expose SAMPLE(d, n), NOW(), UNION [ALL]
- [ ] `SAMPLE(d, n)` — Should attempt to retrieve a sample of N elements from the Dataset. If this cannot be done in a performant fashion, then implementation is free to return a non-random sample.
- [ ]  `NOW()` — The current date/time. This should not be converted in the LP / QS compilers, but rather pushed all the way to the physical plan. For rolling queries, NOW() will have a different interpretation and result ina  stream of result sets, so we should be thinking forward ahead to that.
- [ ] `UNION`
- [ ] `UNION ALL`

### LINKED ISSUE #1711: Treat single column, unaliased selection expressions as SELECT VALUE
When a single projection expression appears in the SELECT clause, and that expression is not aliased, we should treat it equivalent to a SELECT VALUE, and not build a singleton object with "0" and the expression. This just confuses the frontend and it's not necessary in a system that can return sets of raw values (without any decoration in an object or array).

### LINKED ISSUE #1749: DATE_PART doesn't work for ISO_DATE
I have a collection with following "schema" 
```javascript
{ "_id" : ObjectId("57f53249b746cc464b9fac1f"), "dt" : ISODate("1926-08-11T21:00:00Z"), "ts" : -1369278000, "value" : 0.4842846439255769 }
{ "_id" : ObjectId("57f53249b746cc464b9fac20"), "dt" : ISODate("1993-03-21T21:00:00Z"), "ts" : 732747600, "value" : 0.47271090326168097 }
{ "_id" : ObjectId("57f53249b746cc464b9fac21"), "dt" : ISODate("1968-02-05T21:00:00Z"), "ts" : -60058800, "value" : 0.020977698066937056 }
{ "_id" : ObjectId("57f53249b746cc464b9fac22"), "dt" : ISODate("1978-07-23T21:00:00Z"), "ts" : 270075600, "value" : 0.5733601616341653 }
{ "_id" : ObjectId("57f53249b746cc464b9fac23"), "dt" : ISODate("1955-03-16T21:00:00Z"), "ts" : -466916400, "value" : 0.7161196150201754 }
{ "_id" : ObjectId("57f53249b746cc464b9fac24"), "dt" : ISODate("1924-07-04T21:00:00Z"), "ts" : -1435633200, "value" : 0.11387358349052235 }
{ "_id" : ObjectId("57f53249b746cc464b9fac25"), "dt" : ISODate("1937-06-20T21:00:00Z"), "ts" : -1026615600, "value" : 0.9649450316634763 }
{ "_id" : ObjectId("57f53249b746cc464b9fac26"), "dt" : ISODate("1917-03-05T21:00:00Z"), "ts" : -1667012400, "value" : 0.16969064247916554 }
{ "_id" : ObjectId("57f53249b746cc464b9fac27"), "dt" : ISODate("1911-04-26T21:00:00Z"), "ts" : -1851908400, "value" : 0.19975430639659764 }
{ "_id" : ObjectId("57f53249b746cc464b9fac28"), "dt" : ISODate("1940-05-31T21:00:00Z"), "ts" : -933649200, "value" : 0.6063859894813727 }
```

This query 
```sql
select distinct date_part("year", dt), date_part("month", dt), dt, value 
from `/mongo/testDb/dates` 
```
Returns nulls in first two columns for every date before 1970.

sellout:
- Fixes #1541 – add a test for a “basic analytic query”
- Fixes #1710 – expose `sample` and `now()` operations (which necessitated adding `NullaryFunc` to `LogicalPlan`)
- Fixes #1711 – treat an unaliased single-field select as `select value`
- Fixes #1749 – handle dates before 1970 correctly


sellout:
The problem that currently exists here is one of precision – all of the fixes are fine, but the test for the basic analytic query has different floating point precision on MarkLogic and Mongo.

sellout:
I have a couple issues to add for MarkLogic and Couchbase tests that I marked pending (existing issues that were exposed by changes to the tests in here).

Also, this makes a few changes to our tests – we now run `pending` tests on Travis _unless_ it’s also a coverage run and we truncate BigDecimals to 13 digits when comparing them, to make things more consistent across connectors.

foundation/src/test/scala/quasar/QuasarSpecification.scala
```diff
@@ -44,7 +44,7 @@ trait QuasarSpecification extends AnyRef
 
   // Fail fast and report all timings when running on CI.
   if (scala.sys.env contains "TRAVIS") {
-    args(stopOnFail = ArgProperty(true))
+    // args(stopOnFail = ArgProperty(true))
```
wemrysi: Is this intentional? If so, should probably just be deleted.

it/src/main/resources/tests/temporal/convertTimestamp.test
```diff
@@ -2,6 +2,7 @@
   "name": "convert epoch milliseconds value to timestamp",
 
   "backends": {
+    "marklogic":  "pending",
```
wemrysi: Out of curiosity, do you remember why this one failed for ML?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -159,6 +163,31 @@ abstract class QueryRegressionTest[S[_]](
...
+
+  val TestContext = new MathContext(13, RoundingMode.DOWN)
+
+  /** This helps us get identical results on different connectors, even though
+    * they have different precisions for their floating point values.
+    */
+  val reducePrecision =
+    λ[EndoK[ejson.Common]](ejson.dec.modify(_.round(TestContext))(_))
+
+  val normalizeJson: Json => Option[Json] =
```
wemrysi: Nice :+1:

it/src/main/resources/tests/temporal/convertTimestamp.test
sellout: Argh, I tried to keep track of these, but I don’t remember – I know there was a memory issue with one of the ML tests, but then it should have been changed to `skip` so it’s probably not this one. There was also one that had an interval formatting issue with negative intervals, but this test doesn’t involve intervals, so also probably not the culprit.

sellout:
Updated.

sellout:
Should have done `[skip ci]`, since I just removed a commented line.

wemrysi:
@sellout Needs a merge from master, unfortunately :disappointed:

sellout:
Of course. Merged.

PR #1800, opened 2016-12-21
Attempt to reduce amount of memory allocated by MarkLogic during tests

wemrysi:
Some builds were running up against the `SVC-MEMALLOC` error from MarkLogic when running regression tests. According to a developer [e-mail thread](http://markmail.org/message/i7vewefftvnijpwa#query:+page:1+mid:xgpdbv3qm62doxkv+state:results) reducing the list and tree memory limits may help avoid this.


wemrysi:
The changes didn't seem to have an effect on the observed issue, however said issue no longer appears to be affecting builds. Closing this until we notice it again as I wasn't able to reproduce it locally, maybe due to something in the Travis environment?

PR #1815, opened 2016-12-28
Some generalizations of the SQL² implementation.

sellout:
- Use MTL classes for SQL² (which required moving and expanding our variants from MarkLogic)
- Move from `Fix[LP]` and `T[LP]` to `[T](implicit Recursive.Aux[T, LP])`
- A few random changes from Fix-specific operations to {co}recursive operations

(All of the added lines are for our variants on the Scalaz MTL classes – some aren’t used here, but I use them on other upcoming changes, and so I just added them all at once.)


sellout:
Argh, needed to do a full recompile. Some stuff was missed in incremental apparently.

alissapajer:
@sellout LGTM - just missing a changelog

sellout:
Changelog is in.

PR #1817, opened 2016-12-29
Update README

jedesah:


README.md
```diff
@@ -30,12 +31,6 @@ See the instructions below for running and configuring these JARs.
 
 **Note**: This requires Java 8 and Bash (Linux, Mac, or Cygwin on Windows).
 
-### Checkout
```
alissapajer: Why did you remove this section? I think it's useful to explicitly say that the repo needs to be cloned.

README.md
```diff
@@ -62,14 +56,13 @@ A docker container running mongodb operates for the duration of the test run.
...
+##### Testing option 2
 
-Currently Quasar only supports MongoDB so in order to run the integration
-tests, you will need to provide a URL to a MongoDB. If you have a hosted MongoDB instance handy, then you can simply point to it, or else
+In order to run the integration tests for a given backend, you will need to provide a URL to it. For instance, in the case of MongoDB, If you have a hosted MongoDB instance handy, then you can simply point to it, or else
 you probably want to install MongoDB locally and point Quasar to that one. Installing MongoDB locally is probably a good idea as it will
 allow you to run the integration tests offline as well as make the tests run as fast as possible.
 
-In order to install MongoDB locally you can either use something like Homebrew or simply go to the MongDB website and follow the
+In order to install MongoDB locally you can either use something like Homebrew (on OS X) or simply go to the MongDB website and follow the
```
alissapajer: typo: "MongDB"

README.md
```diff
@@ -51,7 +46,6 @@ To compile the project and run tests, execute the following command:
 Note: please note that we are not using here a system wide sbt, but our own copy of it (under ./sbt). This is primarily
  done for determinism. In order to have a reproducible build, the helper script needs to be part of the repo.
 
-This will lead to failures in the integration test project (`it`) because there is no configured "backend" available.
```
alissapajer: I think this line is useful. It tells someone the most likely reason their fresh checkout just failed lots of tests.

README.md
jedesah: I think there are very few users of GitHub that would not know how to clone a GitHub/git repository and I question whether it's the role of each individual README to explain how to do that. It's also not common practice, if it were, I would perhaps be in favor of having it there like everyone else. Ultimately, though, I don't think it's a huge deal to have it there either, so if I haven't convinced you, I am happy to revert this change.

README.md
jedesah: nice catch!

README.md
jedesah: It was very useful! I added it myself 😄 . Fortunately, though, it's no longer what happens.

alissapajer: Ah right, we have the more useful error message now!

README.md
alissapajer: Maybe we can delete this section like you've done, and then below in "Basic Compile & Test" have something like:

"To compile the project and run tests, first clone the quasar repo and then execute the following command:"

I think that makes it just a touch more beginner friendly.

jedesah: I think that's a fair compromise!

PR #1820, opened 2016-12-30
CB: Verify sufficient version

drostron:
Fixes: #1657


couchbase/src/main/scala/quasar/physical/couchbase/fs/package.scala
```diff
@@ -96,6 +99,10 @@ package object fs {
                  })
       cm      =  cluster.clusterManager(params.user, params.pass)
       _       <- EitherT(Task.delay(
+                   (cm.info.getMinVersion.compareTo(minimumRequiredVersion) >= 0).unlessM(
+                     s"Couchbase Server must be ${minimumRequiredVersion}+"
+                       .wrapNel.left[EnvironmentError].left)))
+      _       <- EitherT(Task.delay(
```
wemrysi: Is this second call to `info` still necessary? Can the `handle { ... }` be moved to the new one?

drostron: A useful observation. Intended error messaging is lost unless order is reversed or they are combined _(will roll with that)_.

PR #1827, opened 2017-01-02
remove dead code

alissapajer:
This is a revival of @paulp's PR #1566 - the only new commit is merging in master.

@wemrysi there are a handful of marklogic files touched here. Can you take a look at those to see if you'll need anything that's going away? I'm happy to revert any of those files even if the code isn't used yet; I think getting in the other dead code deletions is important and I'd like to get #1566  out of the queue.


mongodb/src/test/scala/quasar/physical/mongodb/selector.scala
```diff
@@ -102,7 +102,7 @@ class SelectorSpec extends quasar.Qspec  {
 
     "constructors" should {
       "define nested $and and $or" in {
-        val cs =
+        val _ =
```
wemrysi: What is being tested here?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/managefile.scala
```diff
@@ -86,7 +86,7 @@ object managefile {
       dstParent <- toPath(fileParent(dst))
     } yield {
       val deleted = hdfs.delete(dstPath, true)
-      val parentsCreated = hdfs.mkdirs(dstParent)
+      val _ = hdfs.mkdirs(dstParent)
```
wemrysi: `deleted` doesn't appear to be used either, maybe just convert them to statements for clarity? Looks like line 109 should get the same treatment.

connector/src/main/scala/quasar/connector/EnvironmentError.scala
```diff
@@ -40,12 +40,6 @@ object EnvironmentError {
       case _ => None
     } (ConnectionFailed(_))
 
-  val insufficientPermissions: Prism[EnvironmentError, String] =
```
alissapajer: Awesome, yes it can! :)

connector/src/main/scala/quasar/fs/impl.scala
```diff
@@ -95,19 +95,6 @@ object impl {
     )
   }
 
-   def readFromDataCursor[S[_], F[_]: Functor, C](
```
alissapajer: If it's okay with you, I'll opt for deleting it here. Then it will be (re)added to the history when it's used.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/managefile.scala
alissapajer: So, according to [the docs](https://hadoop.apache.org/docs/r2.7.1/api/index.html?org/apache/hadoop/fs/FileSystem.html), both those lines are returning  `Boolean`. 

If we want to keep the code as is, we can either keep the `val`s, or use statements and then ignore the non-unit-statements wart we have enabled.

I think the correct thing to do, though, is actually check the values of the returned `Boolean`s to see if the actions completed successfully, and error otherwise. I think that's out of the scope of this PR, so I'd suggest asking @rabbitonweb about this and then possibly opening a ticket.

What do you think?

mongodb/src/test/scala/quasar/physical/mongodb/selector.scala
alissapajer: Hm, so it looks like we're testing the construction itself of nested `And` and `Or`. Maybe there was a bug in the constructors at some point?  `git blame` attributes this test to @mossprescott on 2015-08-18.

I generally don't like deleting tests. So, I could test the `negate` of `cs`, like other tests in this file do. That way we're still testing the construction of the thing, as well as a bit more.

I can also change the `1 must_== 1` to `ok`.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/managefile.scala
wemrysi: That sounds good, so we'll just leave the code as-is and bring it to @rabbitonweb's attention?

alissapajer: Yes :+1:

mongodb/src/test/scala/quasar/physical/mongodb/selector.scala
wemrysi: Maybe, still seems like a tautology to me as nothing is being asserted about the result so we're just testing that it compiles, but the compiler already does that. Maybe it would be a bit less confusing if the value was just defined in the body of the class and not in an example? Then at least we're not pretending any assertions are happening.

Either way, my intent isn't to make a big deal out of this, so if you think it should stay that's fine by me.

alissapajer:
@wemrysi back to you - I've pushed some changes.

wemrysi:
Looks great, will merge when green.

PR #1828, opened 2017-01-02
ServiceSpec not fail when no env vars set.

alissapajer:


drostron:
Merge script is requesting an update from master.

PR #1829, opened 2017-01-02
Upgrade WartRemover to version `1.2.1`

jedesah:
- Iron out all the new warts it found in the codebase :smile:

### DIFF
--- a/frontend/src/main/scala/quasar/sql/ast.scala
+++ b/frontend/src/main/scala/quasar/sql/ast.scala
@@ -24,9 +24,7 @@ import matryoshka._
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
-sealed trait Sql[A] {
-  def prettyPrint: String = "SQL AST" // TODO: Flesh out pretty printing of ast
-}
+sealed trait Sql[A]
 object Sql {
   implicit val equal: Delay[Equal, Sql] =
     new Delay[Equal, Sql] {
--- a/frontend/src/main/scala/quasar/sql/ast.scala
+++ b/frontend/src/main/scala/quasar/sql/ast.scala
@@ -24,7 +24,9 @@ import matryoshka._
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
-sealed trait Sql[A]
+sealed trait Sql[A] {
+  def prettyPrint: String = "SQL AST" // TODO: Flesh out pretty printing of ast
+}
 object Sql {
   implicit val equal: Delay[Equal, Sql] =
     new Delay[Equal, Sql] {
--- a/frontend/src/main/scala/quasar/std/structural.scala
+++ b/frontend/src/main/scala/quasar/std/structural.scala
@@ -434,7 +434,9 @@ trait StructuralLib extends Library {
 
   object MakeArrayN {
     def apply[T](args: T*)(implicit T: Corecursive.Aux[T, LP]): LP[T] =
-      args.map(MakeArray(_)).reduceOption((x, y) => ArrayConcat(x.embed, y.embed)).getOrElse(Constant(Data.Arr(Nil)))
+      args.map(MakeArray(_))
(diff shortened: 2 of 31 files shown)


frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -24,7 +24,9 @@ import matryoshka._
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
-sealed trait Sql[A]
+sealed trait Sql[A] {
+  def prettyPrint: String = "SQL AST" // TODO: Flesh out pretty printing of ast
```
sellout: There’s already a `pprint` defined in the Sql companion. As a bonus, it actually prints the SQL ;)

jedesah: Ah yes, that is a nice bonus! 😄  Thx for having a look!

mongodb/src/main/scala/quasar/physical/mongodb/workflow/WorkflowOpCore.scala
```diff
@@ -601,7 +601,7 @@ final case class $SimpleMapF[A](src: A, exprs: NonEmptyList[CardinalExpr[JsFn]],
           }(ident("value")),
           Js.Return(Js.Ident("rez"))))
 
-    body(exprs.toList.zipWithIndex.map(("each" + _).second))
+    body(exprs.toList.zip(Stream.from(0).map("each" + _.toString)))
```
alissapajer: Was the change away from `zipWithIndex` driven by a wart? Just curious because I couldn't find reference to such a wart.

jedesah: Yes, but it's not because of `zipWithIndex` itself. It's because there was an implicit conversion to `String` and adding `.toString` to the first version suddenly made the compiler no longer able to infer the type so I would have had to do something like: `body(exprs.toList.zipWithIndex.map((x: Int) => ("each" + x.toString).second))` or `body(exprs.toList.zipWithIndex.map{ case (expr, index) => (expr, "each" + index.toString)})`. At first I was annoyed at scala's spotty inference, but I think I prefer the new implementation now anyway, it only seems longer because of the extra `.toString`, otherwise it would be shorter and I think it's actually easier to grok at a glance.

frontend/src/main/scala/quasar/std/structural.scala
```diff
@@ -434,11 +434,7 @@ trait StructuralLib extends Library {
 
   object MakeArrayN {
     def apply[T](args: T*)(implicit T: Corecursive.Aux[T, LP]): LP[T] =
-      args.map(x => MakeArray(x)) match {
-        case Nil      => Constant(Data.Arr(Nil))
-        case t :: Nil => t
-        case mas      => mas.reduce((t, ma) => ArrayConcat(t.embed, ma.embed))
-      }
+      args.map(MakeArray(_)).reduceOption((x, y) => ArrayConcat(x.embed, y.embed)).getOrElse(Constant(Data.Arr(Nil)))
```
alissapajer: `reduceOption` (and actually `reduce` as well) do not guarantee that the reduce operation will be performed in consistent order. See [scaladoc](http://www.scala-lang.org/api/current/scala/collection/Seq.html#reduceOption[A1%3E:A](op:(A1,A1)=%3EA1):Option[A1]). Though, with that said, it's not clear to me from the docs if the _overall_ order of the `args` will remain the same, and it's just the association within that is potentially nondeterministic...

But in any case, it looks like `reduceLeftOption` (or `reduceRightOption`) is the better choice since it guarantees to return the same result each run (because we have an ordered collection).

jedesah: Nice catch! Updated.

alissapajer:
LGTM - will merge when green

alissapajer:
@jedesah needs merge from  master

PR #1831, opened 2017-01-03
CB: Single quote strings in query rendering

drostron:
Fixes: #1751

### DIFF
--- a/couchbase/src/main/scala/quasar/physical/couchbase/RenderQuery.scala
+++ b/couchbase/src/main/scala/quasar/physical/couchbase/RenderQuery.scala
@@ -41,7 +41,7 @@ object RenderQuery {
   }
 
   val alg: AlgebraM[PlannerError \/ ?, N1QL, String] = {
-    case Data(s @ QData.Str(v)) =>
+    case Data(QData.Str(v)) =>
       ("'" ⊹ v.flatMap { case ''' => "''"; case v   => v.toString } ⊹ "'").right
     case Data(v) =>
       DataCodec.render(v).leftAs(NonRepresentableData(v))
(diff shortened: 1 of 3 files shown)


couchbase/src/main/scala/quasar/physical/couchbase/RenderQuery.scala
```diff
@@ -41,6 +41,8 @@ object RenderQuery {
   }
 
   val alg: AlgebraM[PlannerError \/ ?, N1QL, String] = {
+    case Data(s @ QData.Str(v)) =>
```
alissapajer: Looks like `s` isn't used below.

drostron: That's true — a leftover ready for ✂️.

couchbase/src/main/scala/quasar/physical/couchbase/RenderQuery.scala
```diff
@@ -41,6 +41,8 @@ object RenderQuery {
   }
 
   val alg: AlgebraM[PlannerError \/ ?, N1QL, String] = {
+    case Data(QData.Str(v)) =>
+      ("'" ⊹ v.flatMap { case ''' => "''"; case v   => v.toString } ⊹ "'").right
```
alissapajer: A question about `case '''`: is this the intention to turn a single `'` into `''`?
```
scala> "'".flatMap { case ''' => "''"; case v => v.toString }
res1: String = ''

scala> "'ab'c".flatMap { case ''' => "''"; case v => v.toString }
res2: String = ''ab''c
```

alissapajer: Noting that we'd include an additional `"'"` on either end, returning `''''` and `'''ab''c'`, respectively.

drostron: Correct that is the intention and N1QL's way of quoting `'`s within single quoted strings.

alissapajer: Ok cool!

alissapajer:
LGTM - will merge when green

PR #1834, opened 2017-01-04
Use configuration instead of environment variables for `it` tests

jedesah:
`it` tests are now configured in the `it/testing.conf` file instead of global environment variables


jedesah:
The main benefit of doing things this way besides being more local (inside the git repo as opposed to somewhere in the user's `bash` or `fish` init script) is that this allows one to change what FileSystem to test within the same sbt session as opposed to having to exit and set/unset an environment variable.

jedesah:
@alissapajer I think you will probably like this change as even without something like `testCore`, this makes it really easy to comment out all the backends and just test "pure" code.

drostron:
@jedesah From a quick glance it looks like we'll end up with a vcs diff when updating configuration in our dev envs. Is there a way to specify a different config file outside of vcs for this?

wemrysi:
@drostron I think the inclusion of `application.conf` in the `.gitignore` should prevent the vcs diff, but I agree it'd be nice to be able to specify the location of the config.

drostron:
@wemrysi @jedesah It's been a while since I've used typesafe config, but would this be a case where reference.conf is in vcs and application.conf is used for customization (not sure you can remove enablement that way though)? Actually, just having application.conf outside vcs seems to fit our needs for the moment. Maybe offering an example.conf would be useful. We should think about the future case where we want it to be easy to pull down the repo and test with zero config. This direction seems reasonable since whatever is responsible for setting up the docker environment could also supply the appropriate application.conf.

jedesah:
Perhaps it would be a better idea to rename `application.conf` to `reference.conf` in order to make it easier to override those values with `application.conf`. I can't really think of a particularly compelling use case where we would need to do that. Maybe if we were to ship the `it` tests jar to some client so that they could run the test suite in their environment.

What's the advantage of having an `application.conf.example` as opposed to the examples commented out in the `application.conf` that comes with a fresh checkout? It might avoid a git conflit if the "reference" `appliation.conf` is changed in the upstream repo, but perhaps that's desirable to call attention to the fact that something has changed that might make you want to update your local copy as well.

drostron:
It could be nice for folks to pull down a fresh copy, update a config per connectors available in their env, run tests, and still have a clean vcs diff. We could use `.gitignore`, but that incurs some friction when we want to update it in vcs. If we use `application.conf.example` then all the user has to do is prune `.example` and update per available connectors.

drostron:
To clarify the suggestion:
- Leave `application.conf` in `.gitignore`
- Rename `application.conf` to `application.conf.example`

Then the aforementioned scenario becomes smooth. Thoughts?

wemrysi:
Maybe we even look for `testing.conf` to make it clear that this isn't configuration for Quasar, per se, just integration testing. I was thinking that `testing.conf.example` would just live in the top level, or someplace easier to find than buried in the `resources` directory. We could still `.gitignore` `testing.conf` so that a developer can just copy the example and uncomment things in the root of the project without bothering vcs.

drostron:
👍🏿 for the `testing.conf` suggestion.

jedesah:
I like the idea of it being called `testing.conf` and living in the top level directory even though I could just as easily be convinced that it belongs inside of the `it` directory because it only relates to that project.

The thing though is that it's probably going to require some weird contortions to get typesafe-config to find a file in the top level directory with that name. I don't know if it's even possible at all. I can look into whether it's possible or not or whether there is another config library that might make such a thing easier (Verizon's Knobs perhaps). I have to admit that I don't feel strongly enough that `/testing.conf` is sufficiently better in order to motivate going through the trouble (assuming it's non-trivial).

Since the discussion is open, might be worth mentioning at this point that another possibility I considered for getting rid of the environment variables is that of using sbt `SettingKey`s instead. With such an approach, it should be easy to have a top-level `testConfig.sbt` which overrides the default values of those setting keys. The disadvantage of such an approach is that it would require a `reload` after changing the file or require using the `sbt set` command which is probably slower than just commenting out one or many lines in the configuration file. It also would mean the jar would be hardcoded for a specific uri, so it would not be possible to ship the test jar to some client like I mentioned earlier (or at least make the process less flexible). The advantage would be "typesafe" configuration keys and autocomplete in the `sbt` console. Thoughts on whether you think this could actually be the more desirable approach?

wemrysi:
@jedesah Having the `testing.conf` live at `it/testing.conf` seems fine to me. The typesafe config docs imply that as long as that file makes it on the classpath, it should be found by the library. I think the `it/` dir should be on the classpath for the `it` project by default (if I understand the [sbt docs](http://www.scala-sbt.org/0.13/docs/Classpaths.html) correctly). If not, it looks like it is possible to add directories to the classpath ad-hoc.

As for a config file vs `sbt` setting keys, I think the config file is easier to use, especially for those new to `sbt`, so I'm in favor of sticking with it.

sellout:
I don’t know if this is already handled in this PR, but one thing that bugs me with the current approach is that setting the vars doesn’t mark the it project “dirty”.

So you can run `it/testQuick`, and it skips almost everything (because no var set), then you set a var (and restart sbt), but then `it/testQuick` does nothing, because it thinks the tests were successful. You have to run `it/test` to force it to run the newly-enabled regression tests.

So, it’d be nice if modifying testing.conf reset the it tests.

jedesah:
@sellout I think that's a good idea and perhaps I can have a look at how to accomplish that with `sbt`. Unfortunately, this PR doesn't do that. Since it's a related but orthogonal feature, I think it should be done in another PR anyhow.

jedesah:
Looks like I got all the integration tests working on Travis now except for Mongo read only.

I am pretty sure the problem is with how the environment variable `BACKEND_TO_TEST` is being expanded on line 355 of the build. Anyone have an idea what might be the best way to fix this? If someone would like to suggest a better approach in general to supplying the test configuration to the travis build jobs, I am all ears.

scripts/build
```diff
@@ -38,15 +38,13 @@ if [[ ${LOCAL_MONGODB:-} == "true" ]] ; then
...
-    export QUASAR_MONGODB_3_0="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_0}\"}}"
+    export BACKEND_TO_TEST="-Dmongodb_3_0=\"mongodb://${QUASAR_MONGODB_HOST_3_0}\""
   elif [[ "$QUASAR_MONGODB_RELEASE" == "3_0_RO" ]]; then
-    export QUASAR_MONGODB_READ_ONLY="{\"mongodb\": {\"connectionUri\": \"mongodb://quasar-read:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\"}}"
-    export QUASAR_MONGODB_READ_ONLY_INSERT="{\"mongodb\": {\"connectionUri\": \"mongodb://quasar-dbOwner:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\"}}"
+    export BACKEND_TO_TEST="-Dmongodb_read_only=\"mongodb://quasar-read:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\" -Dmongodb_read_only_insert=\"mongodb://quasar-dbOwner:quasar@${QUASAR_MONGODB_HOST_3_0}/${QUASAR_MONGODB_TESTDB}\""
   elif [[ "$QUASAR_MONGODB_RELEASE" == "3_2" ]]; then
-    export QUASAR_MONGODB_3_2="{\"mongodb\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_2}\"}}"
-    export QUASAR_MONGODB_Q_3_2="{\"mongodbq\": {\"connectionUri\": \"mongodb://${QUASAR_MONGODB_HOST_3_2}\"}}"
+    export BACKEND_TO_TEST="-Dmongodb_3_2=\"mongodb://${QUASAR_MONGODB_HOST_3_2}\""
```
sellout: This seems to drop the `mongodb_q` case.

jedesah:
Oh sorry! I was wondering what mongo_q was! Will restore with a comment
explaining what Q represents

jedesah:
Hurray! 🎆 @wemrysi What do you think of the latest changes?

.gitignore
```diff
@@ -1,4 +1,5 @@
 target/
+/it/src/test/resources/application.conf
```
wemrysi: This should probably be `/it/testing.conf` now, no?

.travis.yml
```diff
@@ -43,12 +43,12 @@ before_cache:
...
   - MONGO_RELEASE=2_6
-  - QUASAR_COUCHBASE='{"couchbase":{"connectionUri":"couchbase://0.0.0.0?username=Administrator&password=password"}}'
-  - QUASAR_MARKLOGIC='{"marklogic":{"connectionUri":"xcc://marklogic:marklogic@localhost:8000/Documents"}}'
+  - QUASAR_COUCHBASE='"couchbase://0.0.0.0?username=Administrator&password=password"'
+  - QUASAR_MARKLOGIC='"xcc://marklogic:marklogic@localhost:8000/Documents"'
   - MONGO_RELEASE=3_0
   - MONGO_RELEASE=3_0_RO
   - MONGO_RELEASE=3_2
-  - QUASAR_POSTGRESQL='{"postgresql":{"connectionUri":"jdbc:postgresql://localhost/quasar-test?user=postgres&password=postgres"}}'
+  - BACKEND_TO_TEST='-Dpostgresql="jdbc:postgresql://localhost/quasar-test?user=postgres&password=postgres"'
```
wemrysi: It looks like the `postgresql` tests are being skipped since the change to use the config file instead of system properties.

CHANGELOG/revision.md
```diff
@@ -0,0 +1 @@
+Use configuration instead of environment variables for `it` tests
```
wemrysi: This can be deleted now that we're using the github-metadata-based merge script.

wemrysi:
@jedesah Is this ready to merge, or were you planning on adding something else (asking as it is still labeled as `in progress`)?

jedesah:
@wemrysi Nope, ready to merge. I was just waiting to see if the build passed to switch over to "in review"

wemrysi:
@jedesah Ok, also you'll need to add a description in order to merge.

jedesah:
@wemrysi done

PR #1835, opened 2017-01-04
Add `testCore` aggregate project for testing "pure" code

jedesah:


build.sbt
```diff
@@ -187,6 +187,7 @@ lazy val githubReleaseSettings =
 lazy val isCIBuild        = settingKey[Boolean]("True when building in any automated environment (e.g. Travis)")
 lazy val isIsolatedEnv    = settingKey[Boolean]("True if running in an isolated environment")
 lazy val exclusiveTestTag = settingKey[String]("Tag for exclusive execution tests")
+lazy val testCore         = project aggregate (foundation, ejson, js, common, frontend, effect, sql, connector, marklogicValidation, core, couchbase, marklogic, mongodb, postgresql, skeleton, sparkcore, interface, web, repl)
```
alissapajer: The `aggregate` below is identical to this one, except for `it`. Meaning that whenever someone adds a new project, it would needed to be added in two places.

I wonder if instead we can just `aggregate` in one place. I like how we show the dependency tree below with comments, so maybe that can be a separate `val` (minus `it`) which we can then use in `root` and `testCore`?

sellout: Or we could put the bulk of the dependency tree in testCore, then have `root … .aggregate(testCore, it)`? (Might want to come up with a different name for `testCore`, though)

alissapajer: Also, I agree with another name for `testCore`. In sbt you'd run `testCore/test`, so maybe `pure` or `basic` or `base` or `unit`... not necessarily sold on any of those though...

jedesah: I agree with the above discussion. I am tempted to close this PR though because I think #1834 sort of supersedes it. Thoughts?

alissapajer: I'm fine with closing this for now. 

I do think it would be nice to have a way to run just the unit tests, without even attempting to run the integration tests. But we should be able to do that more cleanly in our build environment (e.g. `sbt test` vs `sbt it/test`) or something like that.

PR #1838, opened 2017-01-04
Allow variables with quoted names like :`start time`

### LINKED ISSUE #1837: Support expanded variable name syntax
In the frontend we need to be able to naturally namespace variables, and extending the allowed set of variable name characters is probably the most conservative change. Since variables are already prefixed by `:`, could we allow any series of non-whitespace characters as a variable. That, or ```:`foo.bar baz` ``` like syntax.

```
:foo.bar.baz
:foo:bar:baz
:foo/bar/baz
:`foo bar baz`
```

Would these all not be unambiguous from a SQL^2 grammar standpoint?

sellout:
Fixes #1837

Also fixes a bug where we could error in `bson.Timestamp.fromInstant` .

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/bson.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/bson.scala
@@ -21,13 +21,13 @@ import quasar.fp._
 import quasar.javascript._
 import quasar.jscore, jscore.JsFn
 
+import java.time.Instant
 import scala.Any
 import scala.collection.JavaConverters._
 
 import monocle.Prism
 import org.bson._
 import org.bson.types
-import java.time.Instant
 import scalaz._, Scalaz._
 
 /**
@@ -60,7 +60,7 @@ object Bson {
     case rex:  BsonRegularExpression => Regex(rex.getPattern, rex.getOptions)
     case str:  BsonString            => Text(str.getValue)
     case sym:  BsonSymbol            => Symbol(sym.getSymbol)
-    case tms:  BsonTimestamp         => Timestamp.fromInstant(Instant.ofEpochSecond(tms.getTime.toLong), tms.getInc)
+    case tms:  BsonTimestamp         => Timestamp(tms.getTime, tms.getInc)
     case _:    BsonUndefined         => Undefined
       // NB: These types we can’t currently translate back to Bson, but we don’t
       //     expect them to appear.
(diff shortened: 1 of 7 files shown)


mongodb/src/main/scala/quasar/physical/mongodb/bson.scala
```diff
@@ -202,8 +202,11 @@ object Bson {
     override def toString = s"Timestamp(${Instant.ofEpochSecond(epochSecond.toLong)}, ${ordinal.shows})"
   }
   object Timestamp {
-    def fromInstant(instant: Instant, ordinal: Int): Timestamp =
-      Timestamp((instant.toEpochMilli/1000).toInt, ordinal)
+    def fromInstant(instant: Instant, ordinal: Int): Option[Timestamp] = {
```
wemrysi: Is this still used anywhere? Given the change to the return type I'd expect all existing uses to show up in this PR and you appear to have eliminated them.

sellout: I don’t think it is. I left it in, because I imagine this eventually being releasable as some sort of FP-friendly Mongo driver, and it seems converting from an Instant is generally useful. Actually … I’m surprised it’s not being used in the Data => Bson conversion that already exists …

sellout: Ah, because we’re converting through EJson.

wemrysi: Ok, we should probably come to a consensus about unused code as, while I'm fine with leaving this, we just merged a commit recently that deleted all (most?) of the unused code in the codebase.

wemrysi: Or you could add a test or two (especially since this was fixing a defect in the previous version.)

sellout: Actually, I’m going to delete it. Because I was like “oh, this should be a Prism,” but then realized it doesn’t follow the laws when `millis != 0`, and having the conversion fail in between seconds would be terrible, and just tossing the millis (and not making it a Prism) seemed sloppy at best.

sellout:
Ok, pushed that.

wemrysi:
@sellout Needs a merge from master :disappointed:

sellout:
Ok, merged it in.

Issue #1802, opened 2016-12-22
Add TYPEOF function


jdegoes:
I couldn't find an issue for this (possibly due to limitations of Github search).

 * `TYPEOF({} @ {"_ejson.type": "identity"}) = "identity"`

Otherwise:

 * `TYPEOF(true) = "boolean"`
 * `TYPEOF(1) = "integer"`
 * `TYPEOF(1.3) = "decimal"`
 * `TYPEOF(null) = "null"`
 * `TYPEOF("foo") = "array"`
 * `TYPEOF([]) = "array"`
 * `TYPEOF({}) = "map"`
 * `TYPEOF(undefined) = undefined` (i.e. normal undefined semantics)

PR #1842, opened 2017-01-07
Allow users to compare types of values.

sellout:
- Add a `TypeOf` operation (not a fix for #1802 yet, because it doesn’t implement it for any connector)
- adds a new PrimaryType type that reflects the simpler notion of type used with “patterns” as we go forward (and also the notion of types used by `TypeOf`)
- use the sbt-travisci plugin
- convert `Bson.Date` to use a `Long` instead of `Instant` (since `Instant` encodes more values than `Bson.Date`)


drostron:
Merge script requests a merge from master.

sellout:
Merged master.

PR #1843, opened 2017-01-08
Add TemporalTrunc and StartOfDay

### LINKED ISSUE #1735: CB: QScript Interval

### LINKED ISSUE #1773: Add `date_trunc` function
A few integration tests involve queries comparing dates to timestamps. A function such as `date_trunc` is needed in order to avoid coercion that is currently assumed due to mongo behavior. PostgreSQL and a few other SQL flavors offer `date_trunc`.

### LINKED ISSUE #1774: Integration tests incorrectly compare dates to timestamps
Queries involving a date and timestamp should not depend on coercion and instead utilize conversion in order to compare the same types.

Affected tests:
- `complex-vars.test` 
- `filterOnDate.test`

Leverage #1773, #1775

e.g. The query in `filterOnDate.test`
```sql
select day from `../days`
where ((ts > date("2014-08-17") and ts <= date("2014-08-20")) and ts != date("2014-08-19")
```

would become

```sql
select day from `../days`
where (date_trunc("day", ts) > start_of_day(date("2014-08-17")) and date_trunc("day", ts) <= start_of_day(date("2014-08-20"))) and date_trunc("day", ts) != start_of_day(date("2014-08-19"))
```

### LINKED ISSUE #1775: Add `start_of_day` function
Required in addition to #1773 in order to support queries such as those in #1774.

### LINKED ISSUE #1784: CB: Remove unnecessary date/time guards
See: #1774

drostron:
- Fixes #1774 Integration tests incorrectly compare dates to timestamps
- Fixes #1773 Add `date_trunc` function
- Fixes #1775 Add `start_of_day` function
- Fixes #1735 CB: QScript Interval
- Fixes #1784 CB: Remove unnecessary date/time guards


drostron:
@wemrysi I added you since the ML impl ended up being somewhat move involved than expected. Let me know if there are simplifications, improvements, or something that I missed.

connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -764,9 +775,11 @@ object MapFuncs {
   @Lenses final case class Time[T[_[_]], A](a1: A) extends Unary[T, A]
   @Lenses final case class Timestamp[T[_[_]], A](a1: A) extends Unary[T, A]
   @Lenses final case class Interval[T[_[_]], A](a1: A) extends Unary[T, A]
+  @Lenses final case class StartOfDay[T[_[_]], A](a1: A) extends Unary[T, A]
+  @Lenses final case class TemporalTrunc[T[_[_]], A](a1: TemporalPart, a2: A) extends MapFunc[T, A]
```
drostron: Extending `Binary` here expects `a1` to be `A`. Thoughts?

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -578,6 +578,12 @@ class Transform
             prov.genBuckets(base.buckets.drop(1)).fold(NullLit[T, Hole]())(_._2),
             os)).embed))
 
+    case lp.TemporalTrunc(part, src) =>
```
drostron: I'm unfamiliar with `Target`, `Ann` — correction appreciated if incorrect.

frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlanR.scala
```diff
@@ -284,9 +290,10 @@ final class LogicalPlanR[T]
           unifyOrCheck(inf, in.inferred, let(name, appConst(value, constant(Data.NA)), appConst(in, constant(Data.NA))))
         // TODO: Get the possible type from the LetF
         case Free(v) => emit(ConstrainedPlan(inf, Nil, free(v)))
-
         case Sort(expr, ords) =>
           unifyOrCheck(inf, expr.inferred, sort(appConst(expr, constant(Data.NA)), ords map (_ leftMap (appConst(_, constant(Data.NA))))))
+        case TemporalTrunc(part, src) =>
```
drostron: Correct?

frontend/src/main/scala/quasar/std/date.scala
```diff
@@ -170,6 +257,24 @@ trait DateLib extends Library {
...
+    Type.Timestamp,
+    Func.Input1(Type.Date),
+    noSimplification,
+    partialTyper[nat._1] {
+      case Sized(Type.Const(Data.Date(v))) =>
+        Type.Const(Data.Timestamp(startOfDayInstant(v)))
+      case Sized(Type.Date) =>
+        Type.Timestamp
+    },
+    untyper[nat._1] {
```
drostron: Correct?

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
```diff
@@ -346,6 +350,109 @@ object qscript {
...
+          xdmp.formatNumber(sec, "01".xs), ".".xs,
+          xdmp.formatNumber(ms, "001".xs), "Z".xs)
+
+      def fmtDateTime(x: XQuery): XQuery = fn.formatDateTime(x, dateTimeFmt)
+
+      val (dateTime, wrap) = ($("dateTime"), $("wrap"))
+      let_(
+        dateTime := typeswitch(item)(
+          $("dateTime") as ST("xs:dateTime") return_ (ι),
+          $("date") as ST("xs:date") return_ (xs.dateTime(_, xs.time("00:00:00-00:00".xs))),
```
drostron: @wemrysi Is there a way to leverage `startOfDay` here?

connector/src/main/scala/quasar/qscript/MapFunc.scala
sellout: I think you can extend `Unary` here, like how `Guard` extends `Ternary`, despite taking four args (it has three `A` args).

drostron: Ah, didn't notice that precedent — will go with `Unary`.

frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlanR.scala
sellout: Nope – you don’t want to call `appConst` – that will insert the `Typecheck` nodes around the argument to `TemporalTrunc`. You want to treat `TemporalTrunc` as if `handleGenericInvoke` were called on a func with the `Mapping` effect – might be possible to pull out that case to not only work for `GenericFunc`?

frontend/src/main/scala/quasar/std/date.scala
sellout: Yeah, but if you wanted, you could actually run the Const case, no? Something like
```scala
case Type.Const(Data.Timestamp(ts)) => success(Func.Input1(Type.Const(Data.Date(ts.datePart))))
```
Not necessary, though ;)

sellout:
I’ll do more of a review shortly, just wanted to address what you had called out.

frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlanR.scala
drostron: OK, will give that a whirl.

frontend/src/main/scala/quasar/std/date.scala
drostron: Ahh, 👍

marklogic/src/main/scala/quasar/physical/marklogic/xquery/qscript.scala
wemrysi: @drostron Aye, you need to do something like

```scala
val (date, dateTime, wrap) = ($("date"), $("dateTime"), $("wrap"))

startOfDay[F].apply(~date) map { sod =>
  let_(
    dateTime := typeswitch(item)(
      ...
      date as ST("xs:date") return_ κ(sod)
      ...
    )  
  }
```

And then change the `declare` line to use a `flatMap` instead of a `map`.

I just realized that it'd be a lot nicer if there was syntax for `FunctionDeclN` to obtain the XQuery-level function, i.e. something like

```scala
final implicit class FunctionDecl1Ops(val func: FunctionDecl1) extends scala.AnyVal {
  def fn[F[_]](implicit F: PrologW[F]): F[XQuery => XQuery] =
    F.writer(ISet singleton Prolog.funcDecl(func), func.name(_))
}
```

as this would allow you to write

```scala
startOfDay.fn[F] map { sod =>
  let_(
    dateTime := typeswitch(item)(
      ...
      $("date") as ST("xs:date") return_ sod
      ...
    )  
  }
```

which is much more natural.

drostron: Excellent, will roll with your suggestion and 👍🏾 to `FunctionDeclN` syntax at some point.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -66,6 +69,15 @@ abstract class StdLibSpec extends Qspec {
         runner.stringDomain.map(Data.Str(_)))
     }
 
+    // TODO
```
wemrysi: Would it help to add a `dateDomain` to `StdLibTestRunner` like we have for some of the other types?

drostron: Yes, that's useful — will update accordingly.

drostron:
@sellout Ready for review. The ML failure should be addressed by #1850.

drostron:
@sellout Comments addressed and ready for another look.

sellout:
Needs a merge from master (if anyone wants to figure out the magic incantation to line up the merge script merges with the GitHub auto-merges, that’d be great.)

PR #1846, opened 2017-01-10
Re-add not so dead code used downstream in QAdv

drostron:

### DIFF
--- a/connector/src/main/scala/quasar/config/FsPath.scala
+++ b/connector/src/main/scala/quasar/config/FsPath.scala
@@ -120,6 +120,7 @@ object FsPath {
     if (os.isWin) parseWinAbsAsDir(s) else parseOther(codecForOS(os))
   }
 
+  // TODO: Add test (Used downstream in QAdv)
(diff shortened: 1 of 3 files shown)


alissapajer:
Should this code live in QAdv if it's only used there?

drostron:
It would be inconvenient to add `interpretT3` and `runLogT3`  to QAdv as currently implemented. `parseSystemFile` seems to go along with its siblings here. It does however make sense to move `genOption`.

alissapajer:
Maybe it would make sense to add comments to the functions that are staying in quasar, saying they are used in QAdv, only so they don't get wiped again in a few months when we remove dead code again?

Ideally we'd have tests for them too, so they couldn't get deleted, so maybe a TODO for that?

drostron:
I'm in favor or tests since parts of quasar are and will be used as a library therefore having pieces that are only used downstream. Since this is just to facilitate a downstream release I'll add Todos for tests with a mention of their downstream usage.

PR #1848, opened 2017-01-10
Implement TypeOf for MongoDB.

### LINKED ISSUE #1802: Add TYPEOF function

sellout:
Fixes #1802

Also eliminates undefined values from result sets. (This misbehaves
slightly in a couple tests, where results that _should_ be an empty
object are eliminated, but this is better than returning bad results to
a user.)

### DIFF
--- a/frontend/src/main/scala/quasar/codec.scala
+++ b/frontend/src/main/scala/quasar/codec.scala
@@ -53,7 +53,7 @@ object DataEncodingError {
 }
 
 trait DataCodec {
-  def encode(data: Data): DataEncodingError \/ Option[Json]
+  def encode(data: Data): Option[Json]
   def decode(json: Json): DataEncodingError \/ Data
 }
 object DataCodec {
@@ -62,8 +62,8 @@ object DataCodec {
   def parse(str: String)(implicit C: DataCodec): DataEncodingError \/ Data =
     \/.fromEither(Parse.parse(str)).leftMap(ParseError.apply).flatMap(C.decode(_))
 
-  def render(data: Data)(implicit C: DataCodec): DataEncodingError \/ String =
-    C.encode(data).map(_.fold("Undefined")(_.pretty(minspace)))
+  def render(data: Data)(implicit C: DataCodec): Option[String] =
+    C.encode(data).map(_.pretty(minspace))
 
   val Precise = new DataCodec {
     val TimestampKey = "$timestamp"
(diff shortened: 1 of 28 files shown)


frontend/src/main/scala/quasar/codec.scala
```diff
@@ -53,7 +53,7 @@ object DataEncodingError {
 }
 
 trait DataCodec {
-  def encode(data: Data): DataEncodingError \/ Json
+  def encode(data: Data): DataEncodingError \/ Option[Json]
```
paulp: I'm very doubtful about a type of this form. Why isn't the `None` case on the right a member of the ADT on the left? If it isn't an "error" then `DataEncodingError` could be renamed to something more inclusive. Structuring the type like this means there are two layers of wrapping to traverse just to take the happy path.

paulp: ...I think that's the whole review because close to every other line in the patch follows from that type.

sellout: Yeah, good point. `DataEncodingError` should really go away. Part of the problem is that `Data` is just terribly broken, but hasn’t yet gone away.

sellout:
Ok, that was a good simplification (although I just want `Data` to be gone already). Back to you.

it/src/main/resources/tests/undefinedJsInExpr.test
```diff
@@ -1,6 +1,7 @@
 {
     "name": "propagate undefined and null in JS",
     "backends": {
+        "couchbase":         "pending",
```
sellout: @drostron I had to disable this test, because now that `Data.NA` is no longer exposed via JSON, it can’t be used the way you use it in N1QL. So, I replaced it with `Null()` (see the rest of this commit), but that gives the wrong results in this one test.

drostron: @sellout 👌🏽

sellout:
@paulp Could you merge this when you get a change? I’d like to have it in for 4.1.

Issue #1793, opened 2016-12-16
PruneArrays for non-LeftShift arrays


alissapajer:
Currently we only look for arrays produced by the `repair` in a `LeftShift`. We can generalize this to also look for arrays in other `MapFunc`s, for example in the `repair` of a `Reduce`.

 Consider the qscript in #1757 as an example.

alissapajer:
We may need this in `select city, loc[*] from zips` as well. QScript is a `LeftShift` within a `LeftShift`

alissapajer:
^^ The above `LeftShift` example now has its own ticket #1851

alissapajer:
I'm closing this because it's way too general and as such can never actually be completed. When we come across examples that motivate additional array pruning, then we should open individual tickets for those examples.

PR #1849, opened 2017-01-11
Remove extra theta join

### LINKED ISSUE #1783: Generated QScript for query could omit join
Given query:

```
SELECT
 last_name || ", " || first_name AS Name,
 city as City,
 state as State,
 codes[*] AS Code
FROM `/myserver/db_name/patients`
```

produces QScript with a `Join` in it. That `Join` could be omitted with `LeftShift`

alissapajer:
Fixes #1783 - removes unnecessary theta join from query with flatten.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Rewrite.scala
+++ b/connector/src/main/scala/quasar/qscript/Rewrite.scala
@@ -108,6 +108,38 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
               case LeftSide  => mf1
               case RightSide => mf2
             })).some
+          // left side maps over the data while the right side shifts the same data
+          case (Some(Map(\/-(SrcHole), mf1)), Some(LeftShift(-\/(values), struct, status, repair))) =>
+            FI.project(values) >>= QC.prj match {
+              case Some(Map(src2, mf2)) if src2 ≟ HoleQS =>
+                QC.inj(LeftShift(src,
+                  struct >> mf2,
+                  status,
+                  combine >>= {
+                    case LeftSide  => mf1 >> LeftSideF  // references `src`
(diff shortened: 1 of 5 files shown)


connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -880,5 +880,65 @@ class QScriptSpec
         implicitly, Corecursive[Fix[QS], QS])))
 
     }
+
+    "convert a flattened array" in {
+      val lp = fullCompileExp("select city, loc[*] from zips")
+      val qs = convert(listContents.some, lp)
+
+      qs must beSome(beTreeEqual(chain(
```
alissapajer: arrays need to be pruned in this qscript (separate issue, mentioned in #1793, probably its own ticket later)

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -108,6 +108,38 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
...
+              case Some(Map(src1, mf1)) if src1 ≟ HoleQS =>
+                QC.inj(LeftShift(src,
+                  struct >> mf1,
+                  status,
+                  combine >>= {
+                    case LeftSide  => repair >>= {
+                      case LeftSide  => mf1 >> LeftSideF
+                      case RightSide => RightSideF
+                    }
+                    case RightSide => mf2 >> LeftSideF  // references `src`
```
alissapajer: The bug I fixed this morning was here. `RightSide` in `combine` is referencing the `Map`, but after we create the new `LeftShift`, we access the source of that original `Map` through the `LeftSide` of that new `LeftShift`.

alissapajer:
@sellout this is good to go!

PR #1853, opened 2017-01-12
CB: Collapse inlining within ConcatArrays

### LINKED ISSUE #1752: CB: ViewReadQueryRegressionSpec failure
quasar.ViewReadQueryRegressionSpec

```
[error]     x filter on field comparison [./filterOnFieldComparison.test] (14 seconds, 191 ms)
[error]      unexpected value 'Some({"city":"JACKSON","state":"MS","pop":0,"_id":"39269","loc":[-90.188503,32.30085]})' is not None (file:1)
```

drostron:
When possible collapse inlining within `ConcatArrays` for improved CB query engine performance due to reduced query size. `order by` is added to the query in `filterOnFieldComparison.test` to avoid related intermittent failures seen on Travis.

Fixes #1752

### DIFF
--- a/couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
+++ b/couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
@@ -370,29 +370,56 @@ final class MapFuncPlanner[T[_[_]]: BirecursiveT: ShowT, F[_]: Monad: NameGenera
           ).wrapNel,
           Keyspace(a2, id1.some).some,
           unnest  = none,
+          let     = nil,
           filter  = none,
           groupBy = none,
-          orderBy = Nil).embed,
+          orderBy = nil).embed,
       Obj(Map(a1 -> a2)).embed))
     case MF.ConcatArrays(a1, a2) =>
-      IfNull(
-        ConcatStr(a1, a2).embed,
-        ConcatArr(
-          Case(
-            WhenThen(
-              IsString(a1).embed,
-              Split(a1, str("")).embed)
-          )(
-            Else(a1)
-          ).embed,
-          Case(
-            WhenThen(
-              IsString(a2).embed,
-              Split(a2, str("")).embed)
-          )(
-            Else(a2)
-          ).embed).embed
-      ).embed.η[M]
+      def containsAgg(v: T[N1QL]): Boolean = v.cataM[Unit \/ ?, Unit] {
+        case Avg(_) | Count(_) | Max(_) | Min(_) | Sum(_) | ArrAgg(_) => ().left
+        case _                                                        => ().right
+      }.isLeft
+
+      (containsAgg(a1) || containsAgg(a2)).fold(
+        IfNull(
+          ConcatStr(a1, a2).embed,
+          ConcatArr(
+            Case(
+              WhenThen(
+                IsString(a1).embed,
+                Split(a1, str("")).embed)
+            )(
+              Else(a1)
+            ).embed,
+            Case(
+              WhenThen(
+                IsString(a2).embed,
+                Split(a2, str("")).embed)
+            )(
+              Else(a2)
+            ).embed).embed
+        ).embed.η[M],
+        (genId[T[N1QL], M] ⊛ genId[T[N1QL], M]) { (id1, id2) =>
+          SelectElem(
+            Select(
+              Value(true),
+              ResultExpr(
+                IfNull(
+                  ConcatStr(id1.embed, id2.embed).embed,
+                  ConcatArr(Split(id1.embed, str("")).embed, id2.embed).embed,
+                  ConcatArr(id1.embed, Split(id2.embed, str("")).embed).embed,
+                  ConcatArr(id1.embed, id2.embed).embed).embed,
+                none
+              ).wrapNel,
+              keyspace = none,
+              unnest   = none,
+              List(Binding(id1, a1), Binding(id2, a2)),
+              filter   = none,
+              groupBy  = none,
+              orderBy  = nil).embed,
+            int(0)).embed
+        })
     case MF.ConcatMaps(a1, a2) =>
       ConcatObj(a1, a2).embed.η[M]
     case MF.ProjectField(a1, a2) =>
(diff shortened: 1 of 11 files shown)


it/src/main/resources/tests/filterOnFieldComparison.test
```diff
@@ -1,13 +1,12 @@
...
     "name": "filter on field comparison",
     "backends": {
-        "couchbase":         "skip",
         "marklogic":         "skip",
         "mongodb_read_only": "pending",
         "postgresql":        "pending"
     },
     "data": "zips.data",
-    "query": "select * from zips where pop < loc[1] limit 10",
+    "query": "select * from zips where pop < loc[1] order by _id limit 10",
```
alissapajer: Wanted to clarify something about the addition of `order by`. Is this separate from the work on `ConcatArrays` in that the query would still pass consistently with the inclusion of `order by` and without the work on `ConcatArrays`?

Follow up: In a query like this, does CB not provide consistently ordered results (without the `order by`)?

drostron: The `ConcatArrays` change was motivated by timeouts that occurred for this test when `order by` was added.

Yes, in general CB doesn't offer ordering guarantees without a specified order.

sellout: Yeah, I’m curious about the `order by` addition, too. That shouldn’t change whether or not Couchbase passes this test – the `containsExactly` predicate doesn’t care about order. If you do add `order by`, you should also change the predicate to `equalsExactly`.

drostron: There a `limit` though right?

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -370,29 +370,56 @@ final class MapFuncPlanner[T[_[_]]: BirecursiveT: ShowT, F[_]: Monad: NameGenera
...
-          Case(
-            WhenThen(
-              IsString(a2).embed,
-              Split(a2, str("")).embed)
-          )(
-            Else(a2)
-          ).embed).embed
-      ).embed.η[M]
+      def containsAgg(v: T[N1QL]): Boolean = v.cataM[Unit \/ ?, Unit] {
+        case Avg(_) | Count(_) | Max(_) | Min(_) | Sum(_) | ArrAgg(_) => ().left
```
alissapajer: Would it work to make a `N1QLAgg` type, so that where you define the constructors for the aggregate functions they extend `N1QLAgg` instead of `N1QL`. Then if a new aggregate function gets added you don't have to update this here?

it/src/main/resources/tests/filterOnFieldComparison.test
drostron: @sellout Could you clarify you `limit` works with what you mentioned.

alissapajer: The `limit` is in the sql though, so that should be okay?

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
drostron: Yeah, at some point I like to Coproductify N1QL, but didn't feel like tackling that in this PR.

it/src/main/resources/tests/filterOnFieldComparison.test
alissapajer: But I do agree that the test should be able to pass consistently without the `order by`. Maybe the work on `ConcatArrays` improves performance, and we don't need `order by` at all?

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
alissapajer: Sounds good, I agree that can be a separate issue.

it/src/main/resources/tests/filterOnFieldComparison.test
drostron: I was under the impression that if results weren't ordered then any 10 could be the set of 10 returned. If that's the case then it could potentially be a different set of 10 per query result. At that point order or not they would disagree with the specified result set. Interested in clarity here.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
```diff
@@ -370,29 +370,56 @@ final class MapFuncPlanner[T[_[_]]: BirecursiveT: ShowT, F[_]: Monad: NameGenera
...
+        case _                                                        => ().right
+      }.isLeft
+
+      (containsAgg(a1) || containsAgg(a2)).fold(
```
alissapajer: So I think I might be missing something in regards to how the let binding (or aggregation) works overall. Why does at least one side containing an aggregation function preclude the use of a let binding for each side of the `ConcatArrays`?

it/src/main/resources/tests/filterOnFieldComparison.test
sellout: @drostron Ah, right – yeah, the test was previously broken – it should have had the `order by` (and `equalsExactly`) to ensure the correct results. Because the `limit` could end up with any 10 otherwise.

@alissapajer Without `order by`, we can’t assume any ordering.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
drostron: N1QL limits where aggregation functions can appear and this is a case where they are disallowed.

it/src/main/resources/tests/filterOnFieldComparison.test
drostron: @sellout Clarity appreciated and will update to `equalsExactly`.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncPlanner.scala
sellout: @drostron Why not `Option[Unit]` here?

drostron: @sellout Almost went with that — I take it `Option[Unit]` is more idiomatic?

sellout: Yeah, I think just because `Option` is a “smaller” type than `\/`. To me, `\/` means I’m short-circuiting with some value, whereas `Option` means I just want to know _if_ I short-circuited.

drostron: A reasonable perspective.

PR #1859, opened 2017-01-17
array pruning for LeftShift

### LINKED ISSUE #1851: PruneArrays for LeftShift
```
select city, loc[*] from zips
```
We need to examine the `struct` of a `LeftShift` for array derefs.

alissapajer:
Prunes arrays in QScript `LeftShift`.

Fixes #1851

### DIFF
--- a/connector/src/main/scala/quasar/qscript/PruneArrays.scala
+++ b/connector/src/main/scala/quasar/qscript/PruneArrays.scala
@@ -66,7 +66,13 @@ class PAHelpers[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
   /** Returns `None` if a non-static non-integer index was found.
     * Else returns all indices of the form `ProjectIndex(SrcHole, IntLit(_))`.
     */
-  def findIndicesInFunc(func: FreeMap): StateAcc = {
+  def findIndicesInFunc(func: FreeMap): StateAcc =
+    if (func ≟ HoleF)
+      None
+    else
+      findIndicesInStruct(func)
+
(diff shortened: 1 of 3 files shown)


connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -78,19 +84,31 @@ class PAHelpers[T[_[_]]: BirecursiveT] extends TTypes[T] {
     }
 
     // CoEnv[Hole, MapFunc, (T[CoEnv[Hole, MapFunc, ?]], StateAcc)] => StateAcc
-    val galg: GAlgebra[(FreeMap, ?), CoEnv[Hole, MapFunc, ?], StateAcc] =
+    def galg: GAlgebra[(FreeMap, ?), CoEnv[Hole, MapFunc, ?], StateAcc] =
```
sellout: Why change this to a `def`?

sellout:
(Oh, but let me know if you’re going to ignore the comment so I can merge or not.)

connector/src/main/scala/quasar/qscript/PruneArrays.scala
alissapajer: Ah, that's left over from a previous version of this function when I only sometimes used `galg` in the following lines.

alissapajer:
@sellout if you're fine merging with the `def` go ahead and I'll probably make it a `val` again next time I'm in that file. If not I'll revert it in this PR.

sellout:
Yep, merging. (Wasn’t going to bother to write that, but the wifi is so slow, it’s taking longer to actually merge than I expected.)

PR #1863, opened 2017-01-17
SQL^2 pretty-printer fixes

### LINKED ISSUE #1845: Writes view mount config that can't be read back
I started up with a clean `config.json` and ran a workspace (which creates view mounts). When I restarted, the config could not be parsed.

```
Error in configuration file, using default configuration: Failed to parse ./tmp/quasar/config.json: operator ')' expected; `.': CursorHistory(List(El(CursorOpDownField(connectionUri),true), El(CursorOpDownField(view),true), El(CursorOpDownField(/srv/Untitled Workspace 3.slam/.tmp/outdbce9980-7d77-47e3-b6a1-4e508fab8115),true), El(CursorOpDownField(mountings),true)))
```

Which appears to point to this entry in the config:

```
    "/srv/Untitled Workspace 3.slam/.tmp/outdbce9980-7d77-47e3-b6a1-4e508fab8115": {
      "view": {
        "connectionUri": "sql2:///?var.A.results=%60/srv/Untitled%20Workspace%203.slam/.tmp/outeab1f5d1-923c-4012-a170-8340588da3dc%60&q=%28select%20%2A%20from%20%3AA.results%29&var.Foo=%22Baz%22&var.results=%28%60/srv/Untitled%20Workspace%203.slam/.tmp/outf31cb03c-bd15-4e49-8dda-e0f008d27c14%60%2C%60/srv/Untitled%20Workspace%203.slam/.tmp/out2e81bff8-3f5d-44f5-a1eb-6340454a713f%60%29&var.Name=%22Adam%22&var.A.Car=%22Camaro%22"
      }
    },
```

May be related to the liberal variable names.

### LINKED ISSUE #1855: Metastore corruption when multi-file workspace gets converted to single file
From customer:
```
If I create a pivot table in 4.1.0-prerelease2 with quasar-advance 0.6.0
and add a show chart card then this shows no data and also corrupts
the metastore so that when I close the workspace neither the mount
points nor their content is displayed. To recover from this I need to
delete the Metastore. The same happens if I open an old multifile
workspace, once it’s converted to a single file workspace the metastore
is corrupted
```

Appears the common denominator is when a workspace is opened and then saved, going from multi-file format to single-file format, the metastore gets corrupted.

wemrysi:
Fixes pretty printing for complex identifiers and unshift operations so that they emit valid SQL^2. This was causing view mounts that used these features to become corrupted when saved as we use the pretty printer to serialize the view query.

I also updated `ExprArbitrary` to emit complex identifiers and additional operations so the generated expressions better represent the range of possible expressions.

Fixes #1845.
Fixes #1855.


wemrysi:
@alissapajer What do you think about removing the catch-all case in the pretty printer for unary and binary operations in favor of enumerating all of them so that this sort of thing doesn't happen in the future (as we'd benefit from exhaustiveness checking in the compiler). It'd be at the cost of some duplication, but I'm not sure how else to prevent this sort of thing other than constant vigilance in future code reviews.

alissapajer:
@wemrysi since our pretty printers are actually being used to serialize queries (and not just to make a query print nicely) then I think it makes sense to enumerate all the operations so we can be sure to do the right thing on a case by case basis. I agree that constant vigilance in the future isn't really an option.

Since this PR should land ASAP in 4.1 release, I think enumerating all the operations can be done in a separate PR.

alissapajer:
Also, I wonder if we should rename the pretty printers from `pprint` to `printSerializable` or something like that?

When I see a pretty printer, I assume its output is just meant for human eyes and will never be fed into another machine process. Though maybe my assumptions are wrong? :)

wemrysi:
Or maybe we should be serializing using some other mechanism? The pretty printer is used as we only have the parsed expression tree in memory, we could, instead, hold onto the original query provided by the user and then serialize it verbatim (assuming it parsed), eliminating reliance on the pretty printer.

We didn't want to serialize the expression tree in order to remain closed to changes in its representation.

alissapajer:
Eliminating reliance on the pretty printer actually sounds even better. Then we're not required to print something that will serialize. If we can just keep the original query and serialize that, then we're removing an entire step in the serialization process, which would be an improvement.

sellout:
So, to me, `pprint` is the dual to `parse` (like `Show` is supposed to be the dual to `Read`, not that Scalaz has a `Read` type class). But … there are a few things to consider
1. we can formalize this relationship (and eliminate the dual implementations of the parser and pretty-printer) using invertible syntax, which we should move to at some point and
2. once we have position info (which invertible syntax will also give us), pretty-printing will be used less. Right now, any errors need to be pprinted from the AST, but with position info, we’d just `substring` the original query (and, of course, that also allows us to report errors better even after we’ve gotten to some deeper representation like QScript or whatever).

alissapajer:
@wemrysi once the build is green, good to go for merge for 4.1. Then we can figure out the other parsing/printing stuff separately.

alissapajer:
@sellout even if we unify `parse` and `print` with invertible syntax, do you agree it would still make sense to just serialize the original query, rather than pretty print it _first_ and then serialize that? Is there a reason we weren't doing that before?

sellout:
I’m not sure what you mean … if by “original query” you mean the `String`, then it’s already serialized. If you mean the AST, then `pprint` is the serializer. If you mean `Serializable` – I hope there’s nowhere we’re actually serializing any representation of SQL².

The reason we currently can’t just print the original query string is that for an error we want to print the part of the string that was wrong, like `select 3 + "s" from zips` => “mismatched types in `3 + "s"`.” To do that we either need to pprint the portion of the AST that’s incorrect as we currently do, or have access to position information from parsing (which we’d like to do).

NB: I keep talking about “substring”, but we’d probably do something more like scalac to maintain context, like
```
error: mismatched types in operation
error: select 3 + "s" from zips
error:        |_____|
```
but that’s too much to write out every time.

alissapajer:
@sellout ah ok that makes perfect sense - I meant the "original query" and was missing the important part where we are printing "substrings" of that query, which come to us (currently) in AST form.

sellout:
Oh, sorry – I see my confusion. I missed the part about this being for view mounts. Yes, we should be maintaining the original query string in view mounts. They used to be there, but they were eliminated because it seemed redundant to hold both the SQL string and the SQL AST in the mount. However, the mount really should never have had the SQL AST in it at all.

The view mount needs to have 1. the _QScript_ AST (but the best we can do is the LP AST until old Mongo is truly gone), 2. the original query string (only so a user can edit the mount), and 3. ideally some way to indicate the frontend to be used to convert that string into LP (currently we only have SQL², so this can be postponed until we actually add another frontend).

Holding the SQL AST in the view mount has a few problems – 1. it means a user trying to edit the view might see a different query than they originally wrote (since we pprint it), 2. we don’t get SQL->LP->QS errors at the time the view mount is created, but rather when it’s used, and 3. once we have other frontends, we’ll have no way to get to a SQL AST from those frontends.

sellout:
So yeah, a “simple” first step would just be adding the query string to the view mount, and not pprinting anywhere in that code. Then next we can replace SQL AST with QScript AST (once old Mongo is gone). And finally, abstract over the parser, so views can be written using an arbitrary frontend syntax.

alissapajer:
To summarize: since the view mount currently requires a SQL AST in order to be created, any SQL parse errors will prevent the view mount from successfully being created, and those parse errors will be reported using the pretty printer on a subset of the SQL AST. But once we have a view mount, we also have a valid SQL AST. And the problem being fixed here lay in the fact that reverting the entire SQL AST back to a `String` using pretty print was creating invalid SQL^2.

@sellout does that all sound right?

alissapajer:
@wemrysi looks like there are conflicts now :(

PR #1864, opened 2017-01-18
Fix MarkLogic 'union' implementation.

wemrysi:
Arguments to the XQuery `union` operator weren't being grouped properly which could lead to ambiguity in certain cases, wrapping them in a sequence solved the issue.

### DIFF
--- a/it/src/main/resources/tests/union.test
+++ b/it/src/main/resources/tests/union.test
@@ -2,11 +2,9 @@
     "name": "union",
     "backends": {
         "mongodb_read_only": "pending",
-        "mongodb_q_3_2": "pending",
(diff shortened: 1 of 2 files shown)


it/src/main/resources/tests/union.test
```diff
@@ -2,11 +2,9 @@
     "name": "union",
     "backends": {
         "mongodb_read_only": "pending",
-        "mongodb_q_3_2": "pending",
+        "mongodb_q_3_2":     "pending",
         "postgresql":        "pending",
-        "couchbase":         "skip",
-        "marklogic_json":    "pending",
-        "marklogic_xml":     "pending"
```
alissapajer: yay! :)

PR #1865, opened 2017-01-18
integration join tests rewritten to be non-self-join

### LINKED ISSUE #1860: Rewrite join integration tests to be non-self-joins
The majority of the join integration tests use self-joins. This was only done originally due to a lack of datasets with common fields.

This ticket is to rewrite all the self-join integration tests to be non-self-joins. And then additionally include one or two self-join integration tests.

This will provide many more join integration tests that backends can test against, without requiring self-joins, which are really a special case.

alissapajer:
Translates all the join integration tests to be non-self-join, preserving a couple tests as self-join tests as well.

Expands the autojoin simplification condition for theta joins.

Fixes #1860


alissapajer:
Some ML join tests are timing out. I believe the QScript is correct for at least most of these, if not all; and QScript performance is next on my plate after QScript join correctness.

One ML test is failing to autojoin in the QScript compiler, which is a bug in QScript. (Separate ticket incoming.)

wemrysi:
As unfortunate as they are, I'm not terribly surprised about the timeouts as the ML join impl is very naive.

alissapajer:
@wemrysi yeah, I think some the time is being spent on the QScript side as well, so we'll probably meet somewhere in the middle :)

PR #1873, opened 2017-01-20
Update matryoshka-core to 0.16.4

rabbitonweb:
With Matryoshka 0.16.4 we get RecursiveT to be serializable, which fixes
Spark issues. See https://github.com/slamdata/matryoshka/pull/57 for details


rabbitonweb:
@jedesah PR is green on PR, this change will bring spark-core back to normal

Issue #1861, opened 2017-01-17
Add means of controlling the content disposition filename


natefaubion:
As per http://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

https://github.com/slamdata/slamdata/issues/1369

I blindly tried adding the appropriate headers to the `request-headers` query param, crossing my fingers that it would magically work, to no avail 😆

Right now the UI lets you type a filename, but its an obnoxious noop.

drostron:
When using mitmproxy to intercept a request and update the `request-headers` query param to `{"Accept":"application/ldjson;disposition=\"attachment; filename=random_file\""}` the response contains `Content-Disposition:  attachment; filename="random_file"` and Chrome downloads the file as the specified filename. @natefaubion Is that workable/reasonable?

natefaubion:
That should be workable as long as I can do the UTF-8 encoding as in the linked stackoverflow post. Specific test case would be `Content-Disposition: attachment; filename*=UTF-8''Na%C3%AFve%20file.txt`.

drostron:
Currently quotes are being added to the `filename*` value and that appears to throw off at least Chrome. I'll look into it further.

drostron:
The behavior stems from Http4s supporting RFC 2616 and not RFC 5987.

PR #1879, opened 2017-01-24
Handle non-ascii in disposition's filename

drostron:
In lieu of crafting a PR to support RFC 5987 in http4s this is a quick workaround due to the ASAP nature of the originating issue.

Fixes: #1861

### DIFF
--- a/web/src/main/scala/quasar/api/MessageFormat.scala
+++ b/web/src/main/scala/quasar/api/MessageFormat.scala
@@ -75,8 +75,26 @@ trait Decoder {
   }
 }
 
+// [#1861] Workaround to support a small slice of RFC 5987 for this isolated case
+// NonUnitStatements due to http4s's Writer
+@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+final case class ContentDispositionʹ(dispositionType: String, parameters: Map[String, String])
+  extends Header.Parsed {
+  import org.http4s.util.Writer
+  override def key = `Content-Disposition`
+  override lazy val value = super.value
+  override def renderValue(writer: Writer): writer.type = {
+    writer.append(dispositionType)
+    parameters.foreach(p =>
+      p._1.endsWith("*").fold(
+        writer << "; " << p._1 << "=" << p._2,
+        writer << "; " << p._1 << "=\"" << p._2 << '"'))
+    writer
+  }
+}
+
 sealed trait MessageFormat extends Decoder {
-  def disposition: Option[`Content-Disposition`]
+  def disposition: Option[ContentDispositionʹ]
   def encode[F[_]](data: Process[F, Data]): Process[F, String]
   protected def dispositionExtension: Map[String, String] =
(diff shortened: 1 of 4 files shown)


drostron:
I'm interested in alternatives if there are ideas.

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -75,8 +75,26 @@ trait Decoder {
   }
 }
 
+// [#1861] Workaround to support a small slice of RFC 5987 for this isolated case
```
wemrysi: Since it appears the issue is isolated to properly rendering RFC 5987 extended parameter values, could we close the rest of the system to this change by implementing as an `HttpMiddleware`, swapping any `Content-Disposition` headers in the response with this implementation?

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -75,8 +75,26 @@ trait Decoder {
...
+// NonUnitStatements due to http4s's Writer
+@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+final case class ContentDispositionʹ(dispositionType: String, parameters: Map[String, String])
```
wemrysi: As this is a workaround for lack of RFC 5987 support, what do you think of naming it something like `Rfc5987CompliantContentDisposition` or something similarly egregious to justify its existence?

If we end up going the `HttpMiddleware` route, it could even be private/local to that implementation.

web/src/main/scala/quasar/api/MessageFormat.scala
drostron: I considered and like the `HttpMiddleware` approach, but didn't initially see the path. Now that this is working I'll take another look.

web/src/main/scala/quasar/api/MessageFormat.scala
drostron: Sounds good, but maybe sans `Compliant` in case that's to strong.

drostron:
@wemrysi Appreciate the `HttpMiddleware` reminder/suggestion — much nicer.

web/src/main/scala/quasar/api/MessageFormat.scala
```diff
@@ -205,9 +205,9 @@ object MessageFormat {
   // FIXME: I don’t know why this is triggering here.
   @SuppressWarnings(Array("org.wartremover.warts.NoNeedForMonad"))
   def fromMediaType(mediaType: MediaRange): Option[MessageFormat] = {
-    val disposition = mediaType.extensions.get("disposition").flatMap { str =>
-      HttpHeaderParser.CONTENT_DISPOSITION(str).toOption
-    }
+    val disposition = mediaType.extensions.get("disposition").flatMap(str =>
```
wemrysi: This looks like a vestige from the previous implementation, or is it still necessary?

drostron: Vestige indeed.

PR #1881, opened 2017-01-24
Remove double-application of projection in `distinct`.

### LINKED ISSUE #1856: Select value with order by results in empty objects
```
SELECT DISTINCT(state) FROM :results ORDER BY state
```

Results in a list of empty objects, instead of `{ value: foo }` objects. Removing the ORDER BY yields correct results.

sellout:
Fixes #1856.


sellout:
I expect the new test will fail on some other connectors. I’ll mark it pending once I see the Travis results.

sellout:
Realized I broke the test with a naïve find/replace meant to clean it up. Fixed now.

PR #1887, opened 2017-02-02
Provide `ReadSupport[Data]` for Parquet support

rabbitonweb:
Aim of this PR is to provide implementation of `ReadSupport[Data]`. `ReadSupport` is used by Parquet API to transform stream of data from parquet file format to destination type (here to `Data`).

This implementation is so horribly non FP that I cried when I wrote it. HOWEVER at this moment I have no clue how to make it pure FP since there is so much imperative thinking implemented within the parquet protocol.


build.sbt
```diff
@@ -241,6 +241,13 @@ lazy val ejson = project
   .settings(commonSettings)
   .enablePlugins(AutomateHeaderPlugin)
 
+lazy val parquet = project
```
drostron: Did you want to remove the project and move respective files now that they're within the `sparkcore` package?

parquet/src/main/scala/quasar/parquet/reader.scala
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.parquet.hadoop.api.ReadSupport.ReadContext
+import org.apache.parquet.io.api._
+import org.apache.parquet.schema.{OriginalType, MessageType, GroupType}
+import scalaz._
+
+/**
+  * Parquet Mutable Dragons
+  * 
+  * Due to the nature of java-like parquet-mr API we need to deal
+  * with mutable datastrucutres and "protocol" that requries
```
drostron: s/datastrucutres/data structures/
s/requries/requires/

rabbitonweb:
@drostron ready

project/Dependencies.scala
```diff
@@ -94,11 +94,9 @@ object Dependencies {
     "org.tpolecat" %% "doobie-contrib-postgresql" % doobieVersion % "compile, test"
   )
 
-  def parquet = Seq(
-    "org.apache.parquet" % "parquet-format" % "2.3.1",
-    "org.apache.parquet" % "parquet-hadoop" % "1.9.0",
-    "org.apache.hadoop" % "hadoop-client" % "2.3.0" % "provided"
-  )
+  // def parquet = Seq(
```
drostron: Why leave this here and commented out?

PR #1891, opened 2017-02-03
Move typechecks out of join condition in logical plan.

### LINKED ISSUE #1892: Remove typechecks from LogicalPlan join condition
In the `LogicalPlan` for the query
```
select * from zips join smallZips on zips.x = smallZips.foo.bar
```
the `Typecheck` in the join condition references both the left and right sides of the join. This makes it not possible (or at least not easy) to transform this plan into QScript. And makes other connectors (like Couchbase) more difficult to implement as well.

In the old mongo backed, situations like this were handled by deleting the `Typecheck` entirely. This ticket is to handle this  properly as a `LogicalPlan` transformation, moving the `Typecheck` nodes _before_ the join as `Filter`s.
```
Typechecked:
Let('__tmp0)
├─ Let('__check0)
│  ├─ Read(/mongo/test/zips)
│  ╰─ Typecheck(Obj(Map[], Some(Top)))
│     ├─ Free('__check0)
│     ├─ Free('__check0)
│     ╰─ Constant(NA)
╰─ Let('__tmp1)
   ├─ Let('__check1)
   │  ├─ Read(/mongo/test/smallZips)
   │  ╰─ Typecheck(Obj(Map[], Some(Top)))
   │     ├─ Free('__check1)
   │     ├─ Free('__check1)
   │     ╰─ Constant(NA)
   ╰─ Let('__tmp2)
      ├─ Invoke(InnerJoin)
      │  ├─ Free('__tmp0)
      │  ├─ Free('__tmp1)
      │  ╰─ Let('__check2)
      │     ├─ ObjectProject('__tmp1{"foo"})
      │     ╰─ Typecheck(Obj(Map[], Some(Top)))
      │        ├─ Free('__check2)
      │        ├─ Invoke(Eq)
      │        │  ├─ ObjectProject('__tmp0{"x"})
      │        │  ╰─ ObjectProject('__check2{"bar"})
      │        ╰─ Constant(NA)
      ╰─ Invoke(Take)
         ├─ Invoke(Squash)
         │  ╰─ Let('__check4)
         │     ├─ ObjectProject('__tmp2{"right"})
         │     ╰─ Typecheck(Obj(Map[], Some(Top)))
         │        ├─ Free('__check4)
         │        ├─ Let('__check3)
         │        │  ├─ ObjectProject('__tmp2{"left"})
         │        │  ╰─ Typecheck(Obj(Map[], Some(Top)))
         │        │     ├─ Free('__check3)
         │        │     ├─ Invoke(ObjectConcat)
         │        │     │  ├─ Free('__check3)
         │        │     │  ╰─ Free('__check4)
         │        │     ╰─ Constant(NA)
         │        ╰─ Constant(NA)
         ╰─ Constant(Int(11))
```

alissapajer:
Moves typechecks outside of the join condition in a `LogicalPlan`.

Fixes #1892


frontend/src/main/scala/quasar/frontend/logicalplan/Optimizer.scala
```diff
@@ -214,9 +214,7 @@ final class Optimizer[T: Equal]
 
   // FIXME: Make this a transformation instead of an algebra.
   val elideTypeCheckƒ: Algebra[LP, T] = {
-    case Let(n, b, Embed(Typecheck(Embed(Free(nf)), _, cont, _)))
-        if n == nf =>
-      lpr.let(n, b, cont)
+    case Typecheck(_, _, cont, _) => cont
```
alissapajer: We want to actually elide all the let bindings.

frontend/src/main/scala/quasar/frontend/logicalplan/Optimizer.scala
```diff
@@ -235,27 +233,47 @@ final class Optimizer[T: Equal]
     }
 
     def toComp(left: T, right: T)(c: T):
```
alissapajer: Some more important cases added below, compared to @drostron's original diff. Namely a fallback case for `Typecheck` and a way to preserve components through an `And`.

it/src/main/resources/tests/joins/joinWithComplexCondition.test
```diff
@@ -14,11 +14,17 @@
     "query": "select l.sha as child, l.author.login as c_auth, r.sha as parent, r.author.login as p_auth
               from `../slamengine_commits` as l join `../slamengine_commits_dup` as r
               on r.sha = l.parents[0].sha and l.author.login = r.author.login
-              where l.author.login || \",\" || r.author.login = \"selloutsellout\"",
-    "predicate": "containsAtLeast",
```
alissapajer: This query was actually broken on master but is working now. The problem was in the old mongo planner.

mongodb/src/main/scala/quasar/physical/mongodb/planner/MongoDbPlanner.scala
```diff
@@ -459,6 +461,11 @@ object MongoDbPlanner {
                   Selector.Doc(f -> Selector.Type(BsonType.Int32)),
                   Selector.Doc(f -> Selector.Type(BsonType.Int64)),
                   Selector.Doc(f -> Selector.Type(BsonType.Dec))))
+            case Type.FlexArr(_, _, _) ⨿ Type.Str =>
```
alissapajer: This was the key match missing that was making the integration test (silently) fail in master. @sellout I wanted to bring this to your attention since it's a mongo thing. This case was missing, and so the match was failing. The mongo planner was then just silently dropping the entire typecheck node, which of course changed the semantics of the query.

What's the expected behavior when a type doesn't match here? Do you expect all the possible types to be listed here? Or do you expect a more sane default case, namely using the continuation in the `Typecheck` and hoping it's the correct type, rather than just dropping the whole `Typecheck`?

mongodb/src/main/scala/quasar/physical/mongodb/planner/MongoDbPlanner.scala
```diff
@@ -899,6 +907,13 @@ object MongoDbPlanner {
     }
   }
 
+  def removeTypecheckFilters:
```
alissapajer: Old mongo expects these to not be here so we need to get rid of them.

sql/src/test/scala/quasar/sql/compiler.scala
```diff
@@ -98,6 +98,320 @@ class CompilerSpec extends quasar.Qspec with CompilerHelpers {
...
+                      lpf.constant(Data.Bool(false))))),
+                lpf.let('__tmp6,
+                  lpf.invoke2(Filter, // filter right side types
+                    lpf.free('__tmp2),
+                    lpf.typecheck(
+                      lpf.invoke2(ObjectProject, lpf.free('__tmp2), lpf.constant(Data.Str("author"))),
+                      Type.Obj(Map(), Some(Type.Top)),
+                      lpf.constant(Data.Bool(true)),
+                      lpf.constant(Data.Bool(false)))),
+                  lpf.invoke3(InnerJoin, // join post type filters
```
alissapajer: Before this join you can see two `Filter`s, one for each side of the join. These check that the correct types will be joined, moving the `Typecheck` nodes out of the join condition. That was the goal of this PR.

alissapajer:
@drostron I'll create a github issue for this in the morning, showing the `LogicalPlan` for a problematic query, so it's clear what the problem was.

sql/src/test/scala/quasar/sql/compiler.scala
```diff
@@ -98,6 +98,320 @@ class CompilerSpec extends quasar.Qspec with CompilerHelpers {
...
+                          lpf.typecheck(
+                            lpf.free('__tmp12),
+                            Type.Obj(Map(), Some(Type.Top)),
+                            lpf.invoke2(ObjectProject, lpf.free('__tmp12), lpf.constant(Data.Str("login"))),
+                            lpf.constant(Data.NA))),
+                        lpf.constant(Data.NA))))))))))
+    }
+
+    "compile with typecheck in multiple join condition followed by filter" in {
+      testTypedLogicalPlanCompile("select l.sha as child, l.author.login as c_auth, r.sha as parent, r.author.login as p_auth from slamengine_commits as l join slamengine_commits_dup as r on r.sha = l.parents[0].sha and l.author.login = r.author.login where r.author.login || \",\" || l.author.login = \"jdegoes,jdegoes\"",
```
drostron: In the direction of simplifying the largish LP a bit, can this query be whittled down at all and still capture the desired test condition?

frontend/src/main/scala/quasar/frontend/logicalplan/Optimizer.scala
```diff
@@ -334,4 +358,23 @@ final class Optimizer[T: Equal]
       lpr.normalizeTempNames
 
     ).foldLeft1(_ >>> _)
+
+  val rewriteJoins: T => T =
```
drostron: This could benefit from a comment around the commonality and absence of `normalizeLets` compared to `optimize`.

mongodb/src/main/scala/quasar/physical/mongodb/planner/MongoDbPlanner.scala
sellout: I wouldn’t worry about it now, but the default case should probably fail to generate a `Selector`, forcing the `Typecheck` to be handled in an `Expression` or JavaScript. However, you can’t actually typecheck an array in Mongo’s `Selector`. Typechecking in `Selector` is broken to some extent anyway, because what `BsonType.Text` actually means in that context is “either a string or an array containing any number/type of values, but at least one string in there somewhere.” This is even worse in the array case, where Mongo _only_ matches if the value is an array that contains at least one array (https://docs.mongodb.com/manual/reference/operator/query/type/#arrays). So, the `Selector.Type(BsonType.Arr)` case will never match when we want it to.

But, I think that’s fine [dog in fire]. Old Mongo is about gone, and it’s not worth spending more time on it.

mongodb/src/main/scala/quasar/physical/mongodb/planner/MongoDbPlanner.scala
alissapajer: All of `MongoDbPlanner` will be deleted in favor of `MongoDbQScriptPlanner` when we remove old mongo, but I'll add a comment saying that we won't be doing this in new mongo, in case any of this file gets ported over.

sql/src/test/scala/quasar/sql/helpers.scala
```diff
@@ -44,11 +44,17 @@ trait CompilerHelpers extends TermLogicalPlanMatchers {
...
   val lpr = optimizer.lpr
 
-  // Compile -> Optimize -> Typecheck
+  // Compile -> Optimize -> Typecheck -> Rewrite Joins
   val fullCompile: String => String \/ Fix[LP] =
-    q => compile(q).map(optimizer.optimize).flatMap(lp =>
-      lpr.ensureCorrectTypes(lp)
-        .disjunction.leftMap(_.list.toList.mkString(";")))
+    q => compile(q).flatMap { lp =>
+      val withErrors = for {
```
sellout: Definitely can’t move `preparePlan` to foundation. I _think_ it’s in core, which is where it belongs. sql/test can depend on core/compile, though, if needed.

drostron: Yeah, it's in core. @sellout, for future reference, what's the criteria for location? Ah, sure, if the de-duplication is compelling enough then the dependency you mention seems right.

sql/src/test/scala/quasar/sql/compiler.scala
alissapajer: So the only whittling I would consider is selecting only two objects instead of four, one from each side. The complexity in both the join condition and the filter condition are part of the desired test condition.

With that said, this query was the most troublesome one when I was getting the tests to pass, failing in both the mongo planner and integration tests. So for that reason I chose it as a unit test. Even though I agree that the LP is large, I think I'd like to have the LP documented for this query to prevent future (unintentional) changes to the LP that might break the query down the line.

sql/src/test/scala/quasar/sql/helpers.scala
sellout: Well, for one, everything else depends on foundation, so you can’t refer to LP in foundation without moving all of frontend (and its dependencies) into the foundation subproject. Core is where all of the IR transformations live – it is aware of both LP and QScript, and contains things (like `preparePlan`) that operate on LP, but that a frontend (e.g. SQL²) author doesn’t need access to.

But it’s expected that tests for frontends and connectors can depend on core, so that SQL² can test against the normalized LP rather than the naïve LP generated by the SQL² compiler, and Mongo can (eventually) test compilation from easy-to-write (eventually) LP rather than from QScript.

alissapajer: Yeah, I think that the de-duplication is compelling, because what I'm really trying to test here is `preparePlan`. I'll see what I can do.

drostron: @sellout Background and intent on project structure appreciated.

sql/src/test/scala/quasar/sql/compiler.scala
drostron: Was mostly interested if the cost had sufficient value and it sounds like it does. My only concern would be a false failure due to an acceptable but different LP, but that goes along with the nature of this style of test and is in the realm of acceptable maintenance.

sql/src/test/scala/quasar/sql/helpers.scala
alissapajer: So, we already have `core/compile` depending on `sql/compile`. So we cannot make `sql/test` depend on `core/compile` because that creates a recursive value `sql` in the `build.sbt`.

Logically, we'd attempt to compile `sql/test`, determine we need `core/compile`, determine we need `sql/compile`(which we need anyway), which could theoretically be okay, but I don't know if sbt can handle that sort of situation.

Ultimately we don't want `core` to depend on `sql` at all, but until that day we might be stuck here. Unless anyone knows a way around this, I can just make a comment here about using `preparePlan` when we can.

alissapajer:
@drostron changes pushed.

PR #1893, opened 2017-02-04
Organize and consolidate instances in the fp package

wemrysi:
Some reorganization and tidying of `foundation`, moving most of the instances contained in the `fp` package object to appropriate companion objects or under `contrib.scalaz`.


connector/src/test/scala/quasar/fs/ManageFileSpec.scala
```diff
@@ -18,8 +18,9 @@ package quasar.fs
 
 import quasar.Predef._
 import quasar.Data
-import quasar.fp._, eitherT._
-import quasar.fp.ski._
+import quasar.contrib.scalaz.eitherT._
+import quasar.contrib.scalaz.stateT._
+import quasar.fp._, ski._
```
sellout: `ski` should be a separate import, as it’s a separate package.

foundation/src/main/scala/quasar/RenderTree.scala
```diff
@@ -19,13 +19,13 @@ package quasar
 import quasar.Predef._
 import quasar.fp._
 import quasar.fp.ski._
+import RenderTree.make, RenderTree.ops._
```
sellout: These need to be qualified – either `import quasar.RenderTree.{make, ops}, ops._` or ```scala
import quasar.RenderTree.make
import quasar.RenderTree.ops._
```
Really, `RenderTree` should be in its own package, or at _least_ moved to `quasar.fp`, but that’s not something that needs to be in this PR.

File: foundation/src/main/scala/quasar/RenderTree.scala [hunk 3]
```diff
@@ -186,29 +184,25 @@ object RenderTree extends RenderTreeInstances {
...
 
-  implicit def const[A: RenderTree]: Delay[RenderTree, Const[A, ?]] =
-    Delay.fromNT(λ[RenderTree ~> DelayedA[A]#RenderTree](_ =>
-      make(_.getConst.render)))
-
   /** For use with `<|`, mostly. */
   def print[A: RenderTree](label: String, a: A): Unit =
     println(label + ":\n" + a.render.shows)
 
+  // TODO: Should this be necessary given RenderTreeT?
```
sellout: Yes – there are `Recursive` types that aren’t `RecursiveT`, which means they can’t have `RenderTreeT`.

foundation/src/main/scala/quasar/RenderTree.scala
```diff
@@ -231,4 +223,90 @@ sealed abstract class RenderTreeInstances {
...
+    new RenderTree[((A, B), C)] {
+      def render(t: ((A, B), C)) =
+        NonTerminal("tuple" :: Nil, None,
+          RA.render(t._1._1) ::
+          RB.render(t._1._2) ::
+          RC.render(t._2)    ::
+          Nil)
+    }
+
+  implicit def fix[F[_]: Functor](implicit F: Delay[RenderTree, F]): RenderTree[Fix[F]] =
```
sellout: Although … yeah, these particular instances could perhaps be defined using `RenderTreeT` … although I’d rather see `RenderTreeT` go away, as I’d also like to see `RecursiveT` go away.

foundation/src/main/scala/quasar/RenderTree.scala
wemrysi: These are imports from an object defined in this file, why do they need to be qualified?

foundation/src/main/scala/quasar/RenderTree.scala
wemrysi: Ok, I'll remove the comment and we can address the existence of `RenderTreeT` and `RecursiveT` elsewhere.

wemrysi:
@sellout Thanks for the comments, I've pushed a couple commits addressing them.

PR #1902, opened 2017-02-10
`it/testing.conf` as `.example`

drostron:


alissapajer:
@drostron do we want to add `it/testing.conf` to `.gitignore`? (I'm fine with a [skip ci] commit).

alissapajer:
@drostron also, this will need a PR description before it can be merged.

drostron:
@alissapajer Yes, good to have `it/testing.conf` in `.gitignore` and was put in place prior to this PR. The descriptive " " has been added to please our merge script.

alissapajer:
@drostron Ah, so it looks like the `.gitignore` file is incorrect. It has `/it/testing.conf` when it should have `it/testing.conf`, which is why the file wasn't getting ignored before this PR.

drostron:
Perhaps unexpectedly, a leading `/` for paths in `.gitignore` are relative to the `.gitignore` location. Also, I verified earlier that `it/testing.conf` is ignored after this update. The reason it wasn't ignored before this update was due to being tracked.

PR #1903, opened 2017-02-10
ThetaJoin merging

### LINKED ISSUE #1866: failed autojoin when joining sub-queries
In the integration test "join the results of a pair of sub-queries" (`joins/joinSubSelects.test`) we fail to `autojoin` in QScript.

This behavior is exhibited after the merge of #1865.

```
             select t1.city, t1.popCA, t2.popMA from
              (select city, sum(pop) as popCA from `../zips`
                where state = "CA" group by city
                order by sum(pop) desc limit 200) as t1
              join
              (select city, sum(pop) as popMA from `../largeZips`
                where state = "MA" group by city
                order by sum(pop) desc limit 200) as t2
              on t1.city = t2.city
```

### LINKED ISSUE #1867: failed autojoin when joining a complex condition
In the integration test "join with complex condition" (`joins/joinWithComplexCondition.test`) we fail to `autojoin` in QScript.

This behavior is exhibited after the merge of #1865.

```
               "select l.sha as child, l.author.login as c_auth, r.sha as parent, r.author.login as p_auth
                from `../slamengine_commits` as l join `../slamengine_commits_dup` as r
                on r.sha = l.parents[0].sha and l.author.login = r.author.login
                where l.author.login || \",\" || r.author.login = \"selloutsellout\""
```

alissapajer:
Fixes #1867 

Implements ThetaJoin source merging.


alissapajer:
@sellout so this PR doesn't yet have unit tests for the queries in the two tickets it is claiming to fix. This is because (a) the queries take too long to return a result in the integration tests and (2) the qscript is complex enough (with some unused components) that it's not ideal for a unit test.

Basically, this PR is fixing the fact that those two queries used to explode pretty quickly in qscript. Now they just take forever. I'm not sure yet the best way to test this...

alissapajer:
@sellout this PR makes some organizational changes (many of which were needed to get access to some existing functionality `mergeBranches`). The work actually making the `ThetaJoin` merging is in a separate commit.

it/src/main/resources/tests/joins/sameFieldNameComplex.test
```diff
@@ -0,0 +1,18 @@
...
+    "backends": {
+        "couchbase":         "skip",
+        "mongodb_read_only": "pending",
+        "mongodb_q_3_2": "pending",
+        "postgresql":        "pending",
+        "spark_local": "pending",
+        "spark_hdfs": "pending"
+    },
+    "data": ["../owners.data", "../cars.data"],
+    "query": "SELECT owner.name, car.name from `../owners` as owner join `../cars` as car on car._id = owner.carId and owner.year = car.year[0]",
```
alissapajer: This test is one that passes in marklogic and was failing in the old branch. It's a minimal example of #1867.

And as I think about it now, I think I can write a minimal example of #1866 using the same data set that _should_ pass in marklogic, so I'll add a commit for that.

I'd still like to test the original queries though because they are complex enough that minimal examples might not cover _all_ the problems the complex query actually tests.

connector/src/main/scala/quasar/qscript/ThetaJoin.scala
```diff
@@ -96,22 +96,43 @@ object ThetaJoin {
...
+              func.flatMap {
+                case LeftSide => left.as(LeftSide)
+                case RightSide => right.as(RightSide)
+              }
+
+            merge.tryMergeBranches(rewrite)(left1, right1, left2, right2).toOption.map {
+              case (resL, resR) => {
+                val onL: JoinFunc[IT] = updateJoin(o1, resL.lval, resR.lval)
+                val onR: JoinFunc[IT] = updateJoin(o2, resL.rval, resR.rval)
+                val on: JoinFunc[IT] = (onL ≟ onR).fold(onL, Free.roll(MapFuncs.And(onL, onR)))
```
alissapajer: All the tests I've run have verified `onL ≟ onR` (as well as `o1 ≟ o2` actually) so we're not actually hitting this case in any tests yet. The idea here is that if our join conditions are different, that's okay as long as we satisfy both of them.

The `o1` and `o2` conditions are `Constant(Bool(true))` even though we actually join on an `Eq` condition in the final QScript (gets inserted later). I think we'd need a pretty complex case to hit the case when `o1` and `o2` are not equal. Maybe nested joins or something...

alissapajer:
@sellout as an update, this PR does not actually fix #1866 (that was the query I was adding a test for today).

alissapajer:
The new integration test is failing in `marklogic_json` (but passing in `marklogic_xml`) so I disabled it.
```
[error]      java.lang.RuntimeException: XDMP-EXPR: (err:XPTY0004) json:object(<json:object xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:json="http://marklogic.com/xdmp/json"><json:entry key="_ejson.value"><json:value xsi:type="xs:double">...</json:object>) + 1 -- Invalid expression on line 105
```

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -33,26 +33,71 @@ import scalaz.{:+: => _, Divide => _, _},
...
+              S: ShiftRead.Aux[T, F, G],
+              C: Coalesce.Aux[T, G, G],
+              N: Normalizable[G])
+      : T[F] => T[G] = {
+    _.codyna(
+      normalize[G] >>> liftFG(injectRepeatedly(C.coalesceSR[G, APath](idPrism))) >>> (_.embed),
+      ((_: T[F]).project) >>> (S.shiftRead(idPrism.reverseGet)(_)))
+  }
+
+  def simplifyJoins[F[_]: Functor, G[_]: Traverse, H[_]: Functor]
```
sellout: This name isn’t much better than the last one (definitely not worse, though), so I think it still needs a comment.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -188,7 +188,7 @@ class CoalesceT[T[_[_]]: BirecursiveT: EqualT: ShowT] extends TTypes[T] {
       case (l,    r)    => f(l.getOrElse(lOrig), r.getOrElse(rOrig)).some
     }
 
-  def rewrite(elem: FreeMap): Option[FreeMap] = {
+  private def sub(elem: FreeMap): Option[FreeMap] = {
```
sellout: Can this get a better name, like, `eliminateRightSideProjection` or something like that?

connector/src/main/scala/quasar/qscript/Rewrite.scala
alissapajer: Do you have a suggestion for a name that wouldn't require a FIXME? I'd like to get rid of that FIXME if possible :)

alissapajer: I guess this name covers the `simplifyJoin` portion but not the `shiftRead` portion. We could be verbose and say `simplifyJoinOnShiftRead`?

sellout: Yeah, I’m fine with that.

connector/src/main/scala/quasar/qscript/ThetaJoin.scala
alissapajer: @sellout hm, yeah so that example is exactly what I was going for.

I think it might be helpful to think about this in terms of filtered cross-joins. The second query represents the full cross-join of `l` and `r` filtered by `l.a = r.b and l.c = r.d`. The first query represent the join of two cross-joins of `l` and `r`, one filtered by `l.a = r.b` and the other by `l.c = r.d` 

The first query is finding the intersection of two sets where `l.a = r.b` in one and `l.c = r.d` in the other, and in that intersection both conditions will hold. So a result of the first query has the property `l.a = r.b and l.c = r.d` meaning it's contained in the results of the second query. Furthermore, the first query represents the set of _all_ results for which `l.a = r.b and l.c = r.d` (which is the second query) because if a result has that property, it will be a result of both `lj` and `rj` and hence pass the `lj = rj` filter.

So, at least for your specific example, I think this is true. But of course that's just an example and our situation could be more complex. I haven't been able to think up any counterexamples though.

If we're unsure of this, we always require `o1 ≟ o2`, but I think eventually we will need to handle this case for actual joined theta joins (as opposed to "fake" joins created by `QScript` via `autojoin` that will definitely be merged out).

What do you think? I'm feeling reasonable confident about it because if you substitute generic join conditions for `l.a = r.b` and `l.c = r.d`, the logic in my argument still holds. The condition `lj = rj` could be different though, would that change anything?

sellout: Yeah, the `lj = rj` condition would really be the provenance of the two sides, because it’s an autojoin, I was just winging it.

But yeah, I think it seems reasonable.

alissapajer: I'll add a comment explaining it.

alissapajer:
@sellout PR is updated with discussed changes. Back to you.

alissapajer:
Fixing conflicts...

PR #1904, opened 2017-02-10
CB: Basic N1QL key join

drostron:
Fixes: #1786


connector/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -147,6 +147,9 @@ object ReduceIndex {
     extends QScriptCore[T, A]
 
 object QScriptCore {
+  def unapply[T[_[_]], F[_], A](fa: F[A])(implicit C: QScriptCore[T, ?] :<: F): Option[QScriptCore[T, A]] =
+    C.prj(fa)
```
drostron: Indeed

PR #1906, opened 2017-02-11
Add incremental timing to PhaseResult

### LINKED ISSUE #1868: Incremental timings included with PhaseResults
We'd like to time how long we spend computing each `PhaseResult`. 

When `PhaseResults` print in the REPL, we'd print the timing of each phase along with the `RenderTree`. If we're not in debug mode, we'd print the sum of the times of each phase (not just the time of the last phase).

(Note to self: does anything of value happen _not_ during a `PhaseResult`?)

This will greatly aid in debugging performance issues in queries as we'll know which phases are the culprits just by running the query in the REPL.

alissapajer:
Fixes #1868 

Adds incremental timing to `PhaseResult`. 

Additionally renames `F` to `FreeS` in `LiftedOps`.


repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -248,7 +249,12 @@ object Repl {
   ): Free[S, Option[String]] =
     M.lookupType(path).map(_.fold(_.value, "view")).run
 
-  def showPhaseResults: PhaseResults => String = _.map(_.shows).mkString("\n\n")
+  def showPhaseResults: PhaseResults => String = _.foldLeft((None: Option[Long], "")) {
+    case ((prev, acc), res) =>
+      val elapsed: Option[Duration] = prev.map(t => Duration.ofNanos(res.time - t))
+      val timing: String = elapsed.fold("Unavailable")(t => (t.toMillis/1000.0).toString + "s")
```
alissapajer: Upon creation of a `PhaseResult` we record the system time. (This relies on the fact that we actually have a result when we create the `PhaseResult`.) Then when we get the `PhaseResults` (a `Vector[PhaseResult]`) we subtract adjacent times.

This means that we don't have the timing for the first phase that is run. I'd like to have that timing, and we could make an empty `PhaseResult` to stick at the beginning, but that seemed a bit hacky, so I didn't do that. The important thing immediately is that we have `LogicalPlan`, `QScript`, and connector transformation timings.

effect/src/main/scala/quasar/effect/LiftedOps.scala
```diff
@@ -24,9 +24,8 @@ import scalaz._
   * monads over effect algebras.
   */
 abstract class LiftedOps[G[_], S[_]](implicit S: G :<: S) {
-  // FIXME: Rename this to something that wont be confused with a type parameter
-  type F[A] = Free[S, A]
```
alissapajer: The mysterious `F` that I kept assuming was a type parameter.

common/src/main/scala/quasar/common/PhaseResult.scala
```diff
@@ -25,30 +25,37 @@ import scalaz.syntax.show._
...
 object PhaseResult {
-  final case class Tree(name: String, value: RenderedTree) extends PhaseResult
-  final case class Detail(name: String, value: String)     extends PhaseResult
+  private final case class Tree(name: String, time: Long, value: RenderedTree) extends PhaseResult
+  private final case class Detail(name: String, time: Long, value: String)     extends PhaseResult
 
   def tree[A: RenderTree](name: String, value: A): PhaseResult =
-    Tree(name, value.render)
-  def detail(name: String, value: String): PhaseResult = Detail(name, value)
+    Tree(name, java.lang.System.nanoTime, value.render)
```
wemrysi: While I appreciate the ease of reading the system clock here, it breaks referential transparency which I think we should avoid, even when it is for debugging/logging purposes.

One alternative would be a monadic interface, perhaps something along the lines of

```scala
trait PhaseLog[F[_]] {
  def string[A](label: String, value: F[(String, A)]): F[A]
  
  def renderedTree[A](label: String, value: F[(RenderedTree, A)]): F[A]
  
  def detail(label: String, value: F[String])(implicit F: Functor[F]): F[String] =
    string(label, F.map(value)(_.squared))
    
  def message(label: String, msg: String)(implicit F: Applicative[F]): F[Unit] =
    string(label, F.point((msg, ())))
  
  def tree[A: RenderTree](label: String, value: F[A])(implicit F: Functor[F]): F[A] =
    renderedTree(label, F.map(value)(a => (a.render, a)))
}
```

which will also allow for timing actions directly in the implementation instead of relying on the accuracy of adjacent results.

jdegoes: 👍 No production code should break referential transparency.

alissapajer: @wemrysi thanks, I agree. I'll you know how it goes!

alissapajer:
@wemrysi it's been tedious keeping this branch up to date with merge conflicts. So I was thinking that I would close this PR, and then open a separate PR with just the commit 99f40d9c77d9018254fd266e3e499c5025b3118f (renaming of `F` to `FreeS`). And then I'll continue the work locally for the phase logging, meaning that the other commit on this branch 7f4ff429226ab3d794f68d9bb2f79b08f2e7709e for logging won't make it into master.

And since the type-renaming commit came second, I'll have to cherry-pick that commit as well.

Does that sound okay to you?

wemrysi:
@alissapajer Sounds good to me.

PR #1911, opened 2017-02-15
Fix bug in PruneArrays so we check for entire array usage.

alissapajer:
Fix bug in PruneArrays so we check for entire array usage.

### DIFF
--- a/connector/src/test/scala/quasar/qscript/PruneArrays.scala
+++ b/connector/src/test/scala/quasar/qscript/PruneArrays.scala
@@ -207,6 +207,25 @@ class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScr
       initial.pruneArrays must equal(initial)
     }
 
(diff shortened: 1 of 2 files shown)


connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -658,7 +677,7 @@ class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScr
           initialSrc,
           ProjectIndexR(HoleF, IntLit[Fix, Hole](2)),
           ExcludeId,
-          MakeMapR(StrLit("xyz"), Free.point(LeftSide)))).embed
+          MakeMapR(StrLit("xyz"), Free.point(RightSide)))).embed
```
alissapajer: There was an incorrect test case here that would have caught this bug had it not been for the incorrect expectation. So I fixed it and then added more tests.

Issue #1674, opened 2016-11-02
Support Parquet file format - Phase 1


rabbitonweb:
- [x] support primitive types
- [x] support documented logical types
- [x] support undocumented logical types
- [x] implement ParquetRDD
- [x] bind ParquetRDD to readfile & queryfile

rabbitonweb:
Phase 2 in https://github.com/quasar-analytics/quasar/issues/2215

PR #1913, opened 2017-02-16
Parquet missing logical types

rabbitonweb:
Some logical types are not documented or poorly documented. This PR includes support for those types as well.


rabbitonweb:
This PR is part of https://github.com/quasar-analytics/quasar/issues/1674

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
```diff
@@ -135,11 +143,45 @@ private[parquet] class DataReadSupport extends ReadSupport[Data] with Serializab
...
+    }
+
+    override def getConverter(fieldIndex: Int): Converter = converters.apply(fieldIndex)
+    override def start(): Unit = {}
+    override def end(): Unit = {
+      val na = ("n/a" -> Data.NA)
+      val emptyObj = Data.Obj(ListMap[String, Data]())
+      def normalize = values.map {
+        case Data.Obj(lm) => lm.toList match {
+          case ("key" -> Data.Str(k)) :: ("value" -> v) :: Nil => (k -> v)
```
wemrysi: Defining these special strings (as well as "element" for lists) as constants would make it clearer they're part of the encoding as well as keep them opaque in this code.

Defining extractors for them might also be useful so one could write `case MapKey(k) :: MapValue(v) :: Nil`.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
```diff
@@ -135,11 +143,45 @@ private[parquet] class DataReadSupport extends ReadSupport[Data] with Serializab
...
+          case ("value" -> v) :: ("key" -> Data.Str(k)) :: Nil => (k -> v)
+          case _ => na
```
wemrysi: I'm a bit skeptical of this case, as in the presence of multiple failures they will be collapsed into a single `n/a -> NA` entry and, in the unlikely but possible case there is a real entry with a key of `n/a` and a failure in the same map, the real entry will be overwritten with this value.

I wonder if it'd be better to just `lm.toList collect { ... }` and ignore malformed entries in this case? Or just return `NA` for the entire object if any of its entries are malformed?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
```diff
@@ -135,11 +143,45 @@ private[parquet] class DataReadSupport extends ReadSupport[Data] with Serializab
...
+        }
+        case o => ("key_value", o)
```
wemrysi: Is this an expected case? Or would the source data be considered malformed in this scenario?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
```diff
@@ -135,11 +143,45 @@ private[parquet] class DataReadSupport extends ReadSupport[Data] with Serializab
     override def getConverter(fieldIndex: Int): Converter = converters.apply(fieldIndex)
     override def start(): Unit = {}
     override def end(): Unit = {
-      parent.save(name, Data.Arr(values.toList))
+      val normalize = values.toList.map {
+        case Data.Obj(lm) if lm.size == 1 && lm.isDefinedAt("element") => lm("element")
```
wemrysi: Another case where an extractor could be useful.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupportSpec.scala
```diff
@@ -63,42 +63,100 @@ class DataReadSupportSpec extends Qspec {
...
+              Data.Str("scala"),
+              Data.Str("FP"),
+              Data.Str("spillikins")
+            ))
+          )
         )
-      )
+      }
+
+      "be read even if data is not following LIST specification" in {
```
rabbitonweb: In general parquet has primitve types only and logical types are build on top off those primitives.

For example LIST has a specific format that it needs to follow HOWEVER documentation is saying that IF it not follows the specification then reading should be working readless of it

More under https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
```diff
@@ -135,11 +143,45 @@ private[parquet] class DataReadSupport extends ReadSupport[Data] with Serializab
...
+    def save: (String, Data) => Unit = (name: String, data: Data) => {
+      values += data
+      ()
+    }
+
+    override def getConverter(fieldIndex: Int): Converter = converters.apply(fieldIndex)
+    override def start(): Unit = {}
+    override def end(): Unit = {
+      val na = ("n/a" -> Data.NA)
+      val emptyObj = Data.Obj(ListMap[String, Data]())
```
rabbitonweb: It will be deleted, forgot about it

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
rabbitonweb: I'm trying to adhere to backward-compatibilty rules for MAPs that do not follow specification https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules-1

wemrysi: But if we can't somehow extract a key and value from the datum, then is it really a map entry? Seems like we should skip these or error as, similarly to the `n/a` case above, if there are multiple of these they'll overwrite each other in the resulting map.

What if we added another case to the `Obj` handling above that look for a string and a value, ignoring the field names to support backwards compatibility, something like `case (_, Data.Str(k)) :: (_, v) :: Nil => (k -> v)`?

rabbitonweb:
@wemrysi Map is fixed now. So the idea is simple: if MAP adheres to the specification 100% we will get a `Data.Obj` if does not adhere in 100% the parquet interpret it as a normal "group".

I've added a test to show that behaviour.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/parquet/DataReadSupport.scala
```diff
@@ -134,11 +149,77 @@ private[parquet] class DataReadSupport extends ReadSupport[Data] with Serializab
...
+      def normalize = values.map {
+        case Data.Obj(lm) => lm.toList match {
+          case MapKey(k) :: MapValue(v) :: Nil => (k -> v)
+          case MapValue(v) :: MapKey(k) :: Nil => (k -> v)
+          case _ => na
+        }
+        case o => na
+
+      }
+      val data = if(values.size === valid.size) Data.Obj(normalize:_*) else Data.Arr(values.toList)
```
wemrysi: If I understand this correctly, if all the values collected are valid map entries, then we build an object out of them, otherwise we return them all as an array.

There is lots of duplication between `MapEntries` and `normalize`, if we instead define a function to optionally extract a map entry, we can combine extraction and validation into one step and eliminate all the duplication:

```
def mapEntry(d: Data): Option[(String, Data)] = ???
val data = values.traverse(mapEntry).cata(entries => Data.Obj(entries: _*), Data.Arr(values.toList))
```

rabbitonweb: Yes - this is exactly the case. If the format adheres to `MAP` specification 100% we treat the whole 'thing' as a Map (we put all data to `Data.Obj`) if not we rollback to default primitive approach. I've added test to explain scenario (see one of the last commits).

I know there is a lot of similarity but I found impossible to make compiler happy with generalized code.

PR #1917, opened 2017-02-17
SQL compiler now accepts `FunctionDecl` which represent user defined functions

jedesah:
- Simplified and generalized compiler to handle all function invocations in on `case` and support functions of arbitrary arity
- Introduced `Symbol` to avoid having to remember to call `toLowerCase` in many places in compiler
- Compiler will now return `WrongArgumentCount` instead of `FunctionNotFound` if user attempts to call `coalesce` with more or less than 2 arguments


core/src/main/scala/quasar/fs/mount/MountRequestHandler.scala
```diff
@@ -16,7 +16,7 @@
 
 package quasar.fs.mount
 
-import quasar.Predef.{None, Unit}
+import quasar.Predef.{None, Unit, Nil}
```
sellout: This should just be `import quasar.Predef._`.

core/src/main/scala/quasar/package.scala
```diff
@@ -57,8 +59,9 @@ package object quasar {
       substAst <- phase("Variables Substituted",
                     Variables.substVars(ast, vars) leftMap (_.wrapNel))
       absAst   <- phase("Absolutized", substAst.mkPathsAbsolute(basePath).right)
+      annFuncs <- phase("Annotated Functions", scope.traverse(_.transformBodyM(AllPhases[Fix[Sql]])))
```
sellout: This isn’t really a phase, since it doesn’t include the full query.

foundation/src/main/scala/quasar/Predef.scala
```diff
@@ -40,6 +40,14 @@ class Predef extends LowPriorityImplicits with quasar.pkg.PackageAliases with qu
...
   type Char = scala.Char
   type Double = scala.Double
+  trait Function[-A, +B] { self =>
+    def arity: Int
+    def apply(args: List[A]): Option[B]
+    def andThen[C](f: B => C) = new Function[A, C] {
+      def arity = self.arity
+      def apply(args: List[A]) = self.apply(args).map(f)
+    }
+  }
```
sellout: This should be defined elsewhere then re-exported via Predef.

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -172,6 +192,17 @@ object Sql {
   }
 }
 
+final case class Symbol private[sql](value: String)
```
sellout: Why not use Scala’s `Symbol` for this?

If we don’t I’d suggest a new name, like `CIString` or `CIName` or something, as we already use Scala’s `Symbol`.

frontend/src/main/scala/quasar/SemanticError.scala
```diff
@@ -39,8 +39,8 @@ object SemanticError {
     def message = s"The data '${data.shows}' did not fall within its expected domain" + hint.map(": " + _).getOrElse("")
   }
 
-  final case class FunctionNotFound(name: String) extends SemanticError {
-    def message = s"The function '$name' could not be found in the standard library"
+  final case class FunctionNotFound(name: Symbol) extends SemanticError {
+    def message = s"The function '${name.shows}' could not be found in the standard library"
```
sellout: Ah, one distinction here (which can be handled differently once we have source locations) is that the new code doesn’t display the name the same way the user typed it. It’ll be displayed as all lower-case.

So, if you _don’t_ use Scala’s own `Symbol`, then you could have your `Symbol` hold the original string, but compare it case-insensitively.

foundation/src/main/scala/quasar/Predef.scala
sellout: Hrmm, looking at this again, I’m not sure it’s appropriately general – it makes sense in our ASTs, but it requires a function that takes some number of arguments of the same type. And it’s really more of a PartialFunction than a Function, since it can be applied to the wrong number of arguments.

Especially considering that `Predef` is becoming something that’s used across all SlamData projects and not just Quasar, it seems too specific.

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -20,12 +20,32 @@ import quasar.Predef._
 import quasar._, RenderTree.ops._
 import quasar.fp._
 
+import argonaut._, Argonaut._
 import matryoshka._
+import matryoshka.data.Fix
```
sellout: Ah ha! I knew moving this to its own package would help. No new `Fix`!

frontend/src/main/scala/quasar/functions.scala
```diff
@@ -114,13 +119,21 @@ sealed abstract class GenericFunc[N <: Nat] {
...
+
   final def untpe(tpe: Func.Codomain): Func.VDomain[N] =
     untyper0((domain, codomain), tpe)
 
   final def tpe(args: Func.Domain[N]): Func.VCodomain =
     typer0(args)
 
   final def arity: Int = domain.length
+
+  def toFunction[A]: Function[A, LP[A]] = new Function[A, LP[A]] {
```
sellout: So, what does `toFunction` buy us here? `arity` is never used, AFAICT. We could just define `apply` on GenericFunc directly, no? And then it also has composition operators.

sql/src/test/scala/quasar/sql/helpers.scala
```diff
@@ -35,12 +35,21 @@ trait CompilerHelpers extends TermLogicalPlanMatchers {
...
     for {
-      select <- fixParser.parse(Query(query)).leftMap(_.toString)
-      attr   <- AllPhases(select).leftMap(_.toString)
-      cld    <- Compiler.compile[Fix[LP]](attr).leftMap(_.toString)
+      attr   <- parseAndAnnotate(query)
+      cld    <- Compiler.compile[Fix[LP]](attr, Nil).leftMap(_.toString)
     } yield cld
   }
 
+  val parseAndAnnotate: String => String \/ Cofree[Sql, SemanticAnalysis.Annotations] = query => {
```
sellout: Hrmm, isn’t there an error type we can use rather than `String` here?

sql/src/test/scala/quasar/sql/helpers.scala
```diff
@@ -35,12 +35,21 @@ trait CompilerHelpers extends TermLogicalPlanMatchers {
...
+    fixParser.parse(Query(query)).leftMap(_.toString).flatMap { select =>
+      AllPhases(select).leftMap(_.toString)
+    }
+  }
+
+  val parseAndAnnotateUnsafe: String => Cofree[Sql, SemanticAnalysis.Annotations] = query => {
+    parseAndAnnotate(query).toOption.getOrElse(throw new RuntimeException("False assumption in test, could not parse and annotate: "))
```
sellout: And here it looks like
```scala
parseAndAnnotate(query).valueOr(err => throw new RuntimeException(s"False assumption in test, could not parse and annotate: $err"))
```
would be more informative.

frontend/src/main/scala/quasar/sql/ast.scala
jedesah: I don't think scala's `Symbol` is case insensitive so it wouldn't be useful for this use case.

I wouldn't mind calling it `CIName`.

Why are we using `scala.Symbol` anyway?

frontend/src/main/scala/quasar/sql/ast.scala
jedesah: I'm confused by this comment :smile:

frontend/src/main/scala/quasar/SemanticError.scala
jedesah: Ah, yes, wasn't sure what the intended behavior was here. I think I already made this change later to get tests to pass so I'll just include that in this PR along with the other fixes.

foundation/src/main/scala/quasar/Predef.scala
jedesah: I think you can see it in two ways. Either a `PartialFunction` which actually knows a little about its domain, i.e. it can tell you its `arity` and if you respect that, it will be total. A `PartialFunction` is like, just call me and hope for the best which unfortunately does not suit the problem here.

The other way of seeing it is like `Function1[A, B] extends Function[A,B]` and defining apply appropriately (although in that case, it might make sense to rename `apply` defined in `Function` to make it less confusing with `Function1`'s own apply. But you are right, that for `Function2`, the two argument types need to be the same for the "extends" relationship to make sense unless you just say `Function[A1, A2, B] extends Function[Any, B]` but that kind of sucks.

Anyway, I agree that it's pretty borderline generally useful and I am okay defining it somewhere more local. Can always pull it back out if it proves to be useful in another context.

sql/src/test/scala/quasar/sql/helpers.scala
jedesah: This is just the signature of the logic extracted from `compile`.

It's because the error types returned by the parser and the compiler are different

sql/src/test/scala/quasar/sql/helpers.scala
jedesah: Agreed

frontend/src/main/scala/quasar/sql/ast.scala
sellout: Sorry, I got excited 😄

I mean that we shouldn’t reference `Fix` anywhere in the SQL. I managed to make it fixed-point agnostic, and it should stay that way. That’s why I moved specific fixed-points to their own package in matryoshka – to make it more obvious when they are being used.

`FunctionDeclF` should just be `FunctionDecl`, and there should be no type alias for the `Fix` variant.

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -20,12 +20,32 @@ import quasar.Predef._
...
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
+@Lenses final case class FunctionDeclF[BODY](name: Symbol, args: List[Vari[Nothing]], body: BODY) {
+  def transformBody[B](f: BODY => B): FunctionDeclF[B] =
+    FunctionDeclF(name, args, f(body))
+  def transformBodyM[M[_]: Monad, B](f: BODY => M[B]) =
+    f(body).map(FunctionDeclF(name, args, _))
```
sellout: I missed this before … but these are just `map` and `traverse`, and should be implemented as such.

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -20,12 +20,32 @@ import quasar.Predef._
 import quasar._, RenderTree.ops._
 import quasar.fp._
 
+import argonaut._, Argonaut._
 import matryoshka._
+import matryoshka.data.Fix
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
 
+@Lenses final case class FunctionDeclF[BODY](name: Symbol, args: List[Vari[Nothing]], body: BODY) {
```
sellout: Ok, so `String` isn’t the _best_ type, but this `args` should probably use the same time as `Let` bindings (but feel free to come up with a different type than `String` to use in both places). `Vari` is meant to be a variable reference, not introduction.

Probably worth switching them to the same `CIName` type as for function names.

frontend/src/main/scala/quasar/sql/ast.scala
sellout: Symbol just gives you fast equality for strings, so we use it for case-sensitive variable names in LP, and possibly elsewhere.

frontend/src/main/scala/quasar/functions.scala
jedesah: `arity` is being used in the compiler in the case `InvokeFunction`.

jedesah:
@sellout I think I have addressed all your concerns. Let me know if there is anything else

frontend/src/main/scala/quasar/sql/ast.scala
```diff
@@ -192,15 +174,22 @@ object Sql {
...
+    case _                  => false
+  }
+
+  override def hashCode: Int = value.toLowerCase.hashCode
+}
 
-object Symbol {
-  def fromString(s: String) = new Symbol(s.toLowerCase)
+object CIName {
+  def fromString(s: String) = new CIName(s)
```
sellout: Do you still want this function?

jedesah: Nah, kind of pointless. Just pushed a change that also puts `CIName` in it's own file

PR #1918, opened 2017-02-18
Update wartremover to 2.0.2

alissapajer:
Update wartremover to version 2.0.2.


alissapajer:
@drostron I'd like to include `sbt-wartremover-contrib` so we can get `NoNeedForMonad` back, as well as some other warts, but I'm getting an error in the `build.sbt` that I haven't resolved yet. So I figured I'd just get this out to see what jenkins thinks of it.

build.sbt
```diff
@@ -67,15 +67,14 @@ lazy val buildSettings = Seq(
...
   scalacOptions in (Compile, doc) -= "-Xfatal-warnings",
   // NB: Some warts are disabled in specific projects. Here’s why:
-  //   • AsInstanceOf   – puffnfresh/wartremover#266
-  //   • NoNeedForMonad – puffnfresh/wartremover#268
+  //   • AsInstanceOf   – wartremover/wartremover#266
   //   • others         – simply need to be reviewed & fixed
   wartremoverWarnings in (Compile, compile) ++= Warts.allBut(
-    Wart.Any,                   // - see puffnfresh/wartremover#263
-    Wart.NoNeedForMonad,        // - Causes issues compiling with scoverage
-    Wart.ExplicitImplicitTypes, // - see puffnfresh/wartremover#226
```
alissapajer: This issue was resolved in wartremover, so we were able to enable this wart completely.

build.sbt
```diff
@@ -67,15 +67,14 @@ lazy val buildSettings = Seq(
...
+    Wart.Any,                   // - see wartremover/wartremover#263
+    Wart.PublicInference,       // - needs to be reviewed
+    Wart.ImplicitParameter,     // - needs to be reviewed
```
alissapajer: The two warts above create a large number of compiler errors. I was thinking that we can enable these incrementally and at our discretion.

alissapajer:
I solved the problem in build.sbt, so I'll push a change with `sbt-wartremover-contrib` enabled after I fix the issues with the warts! :)

For the record, the problem was that you need `wartremover.Warts.allBut` (using `Warts.allBut` isn't enough) in the build.sbt because it looks to resolve `Warts.allBut` in `wartremover.contrib.Warts.allBut`.

alissapajer:
Well, launching the quasar jar is causing the stack to overflow on this branch. I've seen it twice now in travis. At the bottom of the [full build text](https://s3.amazonaws.com/archive.travis-ci.org/jobs/202976446/log.txt) you can see the `StackOverflowError`. I haven't investigated this at all yet, but I haven't seen this behavior previously.

drostron:
@alissapajer It's recursively attempting to resolve using that definition. Looks like that line isn't necessary anymore though.

drostron:
Oh, [the offending line](https://github.com/quasar-analytics/quasar/pull/1918/files#diff-274022642a21e100a0308a0d117fb924R165) _(had a comment there already so forgot to mention the location 😄 )_.

alissapajer:
@drostron ah yeah that makes sense. I'll wait for the rest of your review before I push any changes.

js/src/main/scala/quasar/jscore/package.scala
```diff
@@ -102,7 +102,7 @@ package object jscore {
     }
     case IfF(Literal(Js.Bool(cond)), cons, alt) =>
       (if (cond) cons else alt).project.some
-    case IfF(cond0, If(cond1, cons, alt1), alt0) if alt0 == alt1 =>
+    case IfF(cond0, If(cond1, cons, alt1), alt0) if alt0 ≟ alt1 =>
```
alissapajer: Yes as far as I can tell, it was just missing these. We ignore `Wart.Equals` in some projects, but not in this one.

connector/src/main/scala/quasar/planner.scala
```diff
@@ -50,8 +50,9 @@ object Planner {
   final case class UnsupportedJoinCondition(cond: Fix[LogicalPlan]) extends PlannerError {
     def message = s"Joining with ${cond.shows} is not currently supported"
   }
+  @SuppressWarnings(Array("org.wartremover.warts.ToString"))
   final case class UnsupportedPlan(plan: LogicalPlan[_], hint: Option[String]) extends PlannerError {
-    def message = "The back-end has no or no efficient means of implementing the plan" + hint.map(" (" + _ + ")").getOrElse("")+ ": " + plan
+    def message = "The back-end has no or no efficient means of implementing the plan" + hint.map(" (" + _ + ")").getOrElse("")+ ": " + plan.toString
```
alissapajer: Ah yes, so the problem here is that to require a `Show` for `plan` here we'd need `class UnsupportedPlan[A: Show](plan: LogicalPlan[A], ...)`. And threading that through to the call sites found cases where we didn't have a `Show` for `A`. Which is why I skipped this for now.

build.sbt
drostron: Agreed

frontend/src/main/scala/quasar/sql/parser.scala
```diff
@@ -452,11 +452,12 @@ private[sql] class SQLParser[T[_[_]]: BirecursiveT]
         keyword("intersect")                  ^^^ (Intersect(_: T[Sql], _: T[Sql]).embed)    |
         keyword("except")                     ^^^ (Except(_: T[Sql], _: T[Sql]).embed))
 
+  @SuppressWarnings(Array("org.wartremover.warts.ToString"))
   def parseExpr(exprSql: String): ParsingError \/ T[Sql] =
     phrase(expr)(new lexical.Scanner(exprSql)) match {
       case Success(r, q)        => \/.right(r)
       case Error(msg, input)    => \/.left(GenericParsingError(msg))
-      case Failure(msg, input)  => \/.left(GenericParsingError(msg + "; " + input.first))
+      case Failure(msg, input)  => \/.left(GenericParsingError(msg + "; " + input.first.toString))
```
alissapajer: If `toString` is explicitly overridden, then the wart will allow a call to `toString`. I'm not actually sure what the type of `input` is offhand though, to verify that that's the case here.

drostron: Ah, yep. That's sounds right.

alissapajer:
@drostron changes pushed - back to you.

build.sbt
drostron: Updating comments to what you say here would be good.

alissapajer: Will do!

PR #1919, opened 2017-02-20
Run sparkcore tests in parallel + remove obsolete test

rabbitonweb:
`Sparkcore` was running test in parallel tasks - this was turned off
`ReadFileSpec` is obsolete - so it was deleted


drostron:
What motivated going with sequential?

rabbitonweb:
@drostron `SparkContext` is like the Ring from the Lord of the Rings. It is like immortal from "Highlander".

There can be only one (per JVM)

drostron:
@rabbitonweb Appreciate that explanation 😄.

So, tests _could_ be revamped to use the same ctx, but I'm ok making that change later if/ever non-sequential is desired.

rabbitonweb:
@damonLL yes

rabbitonweb:
@drostron yes - in single test we can reuse `SparkContext`
             and no - each test file has to run one after the other (right?)

drostron:
@rabbitonweb Sure, it sounds like cost of trying to reuse a `SparkContext` (if even possible) isn't worth the trouble.

drostron:
Needs a semantic version label.

rabbitonweb:
Done

PR #1921, opened 2017-02-20
Rewrite MarkLogic filesystem to be a direct mapping to Quasar filesystem

### LINKED ISSUE #1847: Option to Treat a MarkLogic File as a Quasar File
We currently assume that a MarkLogic File is the equivalent of a single document or single row in a different database.  @wemrysi and I ran into a use case where a MarkLogic File was actually a collection of many rows / documents.

This resulted in unexpected behavior when querying the Quasar File (which I believe is currently equivalent to a MarkLogic Directory)

For instance, we might expect the `/Users` directory in MarkLogic (which is presented as a Quasar File) to contain 100 MarkLogic files, perhaps labeled `User1.xml`, `User2.xml` and so on.  So performing a `SELECT * FROM Users` might return multiple entries, one per user.

Instead, if that same MarkLogic directory contained a single MarkLogic File named `Users.xml` which stored multiple users, when we perform a `SELECT *` on that file, rather than getting one entry per user, we get a single entry with an array containing all the users.  @wemrysi keep me honest here, please.

This is a request to enhance the MarkLogic connector to allow for this situation

wemrysi:
A significant rewrite of the MarkLogic filesystem implementation as well as the read bits of the planner to support queries on file and directory paths. This does away with the previous encoding of Quasar files as MarkLogic directories in favor of a direct mapping between Quasar files and dirs to MarkLogic files and dirs.

Also updates all the filesystem operations to respect the mount format so that operations on files of one format don't affect files in other formats as well as restricting the view of the filesystem to files of the configured format to improve user experience.

Closes #1847.


it/src/test/scala/quasar/physical/marklogic/fs/DirectoryQueriesSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+import pathy.Path._
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+import scalaz.stream.Process
+
+final class DirectoryQueriesSpec extends MultiFormatFileSystemTest {
+
+  def multiFormatFileSystemShould(js: FileSystem ~> Task, xml: FileSystem ~> Task) = {
+    "Querying directory paths" >> {
+      "results in a dataset comprised of immediate child documents" >> {
```
alissapajer: This test seems to be testing the original issue reported in #1847, is that correct?

web/src/test/scala/quasar/api/Http4sPathEncodingSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import org.http4s.dsl.{Path => HPath}
+import pathy.Path._
+import pathy.scalacheck.PathyArbitrary._
+import scalaz.Scalaz._
+
+class Http4sPathEncodingSpec extends quasar.Qspec {
+  val codec = UriPathCodec
+
+  "print and parse through http4s should produce same Path" >> {
```
wemrysi: Yes, see `quasar.contrib.pathy.UriPathCodecSpec` from which this was extracted when the former was moved into `foundation`.

it/src/test/scala/quasar/physical/marklogic/fs/DirectoryQueriesSpec.scala
wemrysi: This is testing that queries on Quasar directory paths work as expected. The common regression test suite tests the new feature described in #1847 as it is focused on creating and querying individual Quasar files.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/ShiftReadDir.scala
```diff
@@ -0,0 +1,129 @@
...
+import matryoshka.{Hole => _, _}
+import matryoshka.data._
+import matryoshka.implicits._
+import matryoshka.patterns._
+import scalaz._, Scalaz._
+
+/** Like [[ShiftRead]], but only applies when the path is a directory, leaving
+  * file reads untouched.
+  */
+trait ShiftReadDir[F[_]] {
```
wemrysi: Yeah, good point, I'll move it.

alissapajer:
@wemrysi I also wanted to confirm that "version:feature" is the correct tag for this. Would this be considered "breaking" for a user because their ML data may be treated differently (i.e. not written into an array but written into records)?

sellout:
I think this should be version:revision (although I haven’t reviewed the code), because we’re currently only versioning the Web API.

wemrysi:
@alissapajer Hm, that's a good question. I had been thinking about it from the view of existing queries, which _should_ still work without modification, however I hadn't considered the versioning implications of the uploaded data.

@sellout `revision` is fine with me as well.

it/src/test/scala/quasar/physical/marklogic/fs/DirectoryQueriesSpec.scala
alissapajer: So, I'm still confused about one thing. This PR treats ML files as Quasar files (and dirs as dirs). So is there an example of a query that previously would have treated a ML file with multiple rows as a single record that now treats that file as containing multiple records?

wemrysi: No, but there was no way to actually reference an ML file directly in a query before as all file paths were interpreted as directories in ML. Additionally, all files are treated as a single record as required by MarkLogic, which enforces that XML and JSON documents have exactly one root node, which coincidentally maps nicely to MRA's view of data as well.

marklogic/src/main/scala/quasar/physical/marklogic/fs/ops.scala
```diff
@@ -17,234 +17,250 @@
...
+      case _                            => false
     }
   }
 
-  def deleteFile[F[_]: Monad: Capture: SessionReader: XccErr](file: AFile): F[Unit] = {
-    val uri = pathUri(asDir(file))
+  /** Creates a new file containing the given contents, overwriting any
+    * existing file. Returns any errors encountered, either with the
+    * contents or during the process of creation itself.
+    */
```
wemrysi: This function would indeed allow one to overwrite by mistake, but it is sort of intended to in the sense that it is exposing the low-level primitive from MarkLogic which has this behavior. I think I'll rename it to `insertFile` which is a bit more evocative of the "make it so" semantics.

Collecting warnings could provide some useful feedback, though we don't really have the notion of a warning at the filesystem api level, just success or failure. In cases where one is concerned about overwriting something, we have derived `create` operations that fail if the destination file exists.

marklogic/src/main/scala/quasar/physical/marklogic/xcc/Xcc.scala
```diff
@@ -34,19 +34,37 @@ import simulacrum.typeclass
 
 @typeclass
 trait Xcc[F[_]] extends MonadError_[F, XccError] {
+  /** Returns the most recent system commit timestamp. */
```
alissapajer: +1 to more doc.

alissapajer:
@wemrysi build failure in 2.6 mongo build. Looking at the [raw logs](https://s3.amazonaws.com/archive.travis-ci.org/jobs/204303811/log.txt), it looks like marklogic/compile for Scaladoc generation failed. At least that's the only error that I found...

wemrysi:
@alissapajer Yeah, I never think to test docs locally, I'll take a look.

alissapajer:
:+1: yeah I don't test docs locally either, since failures there are so rare.

alissapajer:
Also, there are some conflicts now :(

alissapajer:
Looks like there's still a Scaladoc problem in the mongo 2.6 build.

wemrysi:
Yeah, I haven't figured out how to fix it yet, it _seems_ like compilation in the doc step for the marklogic project isn't finding the simulacrum-generated operations on the `Xcc` typeclass, but I don't know why, or what is different about "regular" compilation and doc compilation that would cause this.

Or it could be an implicits issue as well (maybe the `transact` syntax is being shadowed somehow in doc compilation)?

wemrysi:
Avoiding simulacrum on the `Xcc` typeclass seems to have solved the problem locally, we'll see if travis agrees.

PR #1932, opened 2017-02-28
Include physical input paths to query in ExecutionPlan

### LINKED ISSUE #1920: Allow backends to provide the set of physical inputs to a query
Responses from the `/compile` endpoint include the physical set of inputs to the compiled query in addition to the compilation output itself.

This set of inputs is currently determined from `LogicalPlan`, assuming each path represents a file in the underlying backend(s), however this may not always be the case (if a path represents a physical directory, or something deeper than a file, for instance). Instead the set of physical inputs should be provided by the backend(s) themselves.

wemrysi:
Adds an `ISet[APath]` to `ExecutionPlan` representing the physical inputs to the query planned. This allows connectors to provide the input paths, which is necessary as paths in `SQL^2` don't necessarily represent physical paths.

Closes #1920


connector/src/main/scala/quasar/qscript/ExtractPath.scala
```diff
@@ -0,0 +1,122 @@
...
+  ): ExtractPath[Coproduct[F, G, ?], P] =
+    new ExtractPath[Coproduct[F, G, ?], P] {
+      def extractPath[H[_]: ApplicativePlus]: Algebra[Coproduct[F, G, ?], H[P]] =
+        _.run.fold(F.value.extractPath[H], G.value.extractPath[H])
+    }
+
+  implicit def constDeadEnd[P]: ExtractPath[Const[DeadEnd, ?], P] =
+    new ExtractPath[Const[DeadEnd, ?], P] {
+      def extractPath[G[_]: ApplicativePlus] = {
+        case Const(Root) => mempty[G, P]
```
sellout: Not certain about this, but inclined to think that this should be an error (or `None`). `{Shifted}Read` is generated by the backend to represent physical paths, but the existence of `Root` in the QScript means we haven’t resolved all physical paths.

wemrysi:
@sellout I managed to hide the `ExtractPath` instances for `DeadEnd` and `ProjectBucket` so that they exist for the impl of `QScriptTotal` but not generally, so now attempts to `ExtractPath[F, P]` when `Const[DeadEnd, ?] :<: F` fails to compile.

sellout:
I was going to argue against this, but talked myself into it being a good idea. I don’t think `ProjectBucket` needs to be restricted, but I’m fine either way.

foundation/src/main/scala/quasar/contrib/pathy/order.scala
```diff
@@ -0,0 +1,41 @@
+/*
```
drostron: Looks like pathy 0.2.8 will be in master soon — #1936. Up to you whether or not to wait so we don't have to prune later.

wemrysi: I guess we can either delay this or prune in #1936, depending on which is merged first.

djspiewak: I sincerely hope #1936 gets blessed and merged shortly, since Quasar isn't buildable on a clean machine without it!  :-)

drostron:
Needs a merge from master.

foundation/src/main/scala/quasar/contrib/pathy/order.scala
drostron: Was about to say this is prunable now, but slamdata/scala-pathy#40 is still open. Version in the comment is a touch stale.

wemrysi: Yeah, now that #1936 has been merged, it looks like versions agree enough to use the latest version of pathy. So I'll update this once slamdata/scala-pathy#40 is merged.

foundation/src/main/scala/quasar/contrib/pathy/order.scala
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package quasar.contrib.pathy
+
+import quasar.Predef._
+
+import pathy.Path, Path._
+import scalaz._, Scalaz._
+
+// FIXME: Remove once we're able to upgrade to pathy-0.2.7+
```
djspiewak: Is this stale now that master is on 0.2.8?

foundation/src/main/scala/quasar/contrib/pathy/order.scala
```diff
@@ -0,0 +1,41 @@
...
+object order {
+  implicit val dirNameOrder: Order[DirName] =
+    Order.orderBy(_.value)
+
+  implicit val fileNameOrder: Order[FileName] =
+    Order.orderBy(_.value)
+
+  implicit def pathOrder[B,T,S]: Order[Path[B,T,S]] =
```
djspiewak: Presumably we can just remove this by updating to pathy 0.2.9?

marklogic/src/main/scala/quasar/physical/marklogic/fs/queryfile.scala
```diff
@@ -150,14 +151,15 @@ object queryfile {
                      N.normalizeF(_: MLQ[T[MLQ]]))))
       _       <- logPhase(PhaseResult.tree("QScript (Optimized)", optmzed))
       main    <- plan(optmzed)
+      inputs  =  optmzed.cata(ExtractPath[MLQ, APath].extractPath[DList])
```
djspiewak: BTW, `s/optmzed/optimized/g`, unless the `i` key is frowned upon.

PR #1934, opened 2017-03-01
[WIP] Spark cassandra

marekklis:
This PR introduces a new connector called `spark-cassandra`. This connector is based on sparkcore and it gives ability to mount Quasar to Cassandra instance.
Keyspaces are representing folders, tables are representing files.
Current limitation: connector works with single column tables that hold JSON String as a value.

TODO:
- [x]  All tests (unit & integration) are passing except for one: `QueryFilesSpec."executing query to an existing file overwrites with results"`
- [ ]  Cassandra's keyspaces (here folders in FS) work only with alphanumeric characters - we need to handle errors when user provides a character out of the alphanumeric set 

Possible further improvements:

- [ ] Remove the "single JSON column" limitation & be able to represent any Cassandra table
- [x]  `cassandra.common` should be represented as algebra (e.g `CassandraDDL[A]`) so that it can be later used and interpreted as `Free[CassandraDDL, A]`


interface/src/main/scala/quasar/main/package.scala
```diff
@@ -57,7 +57,8 @@ package object main {
     postgresql.fs.definition[PhysFsEff],
     skeleton.fs.definition[PhysFsEff],
     sparkcore.fs.hdfs.definition[PhysFsEff],
-    sparkcore.fs.local.definition[PhysFsEff]
+    sparkcore.fs.local.definition[PhysFsEff],
+    sparkcore.fs.cassandra.definition[PhysFsEff]
```
sellout: A minor comment – in places where we list multiple connectors, cassandra should be listed before hdfs – using alphabetical order makes it easier to scan and edit places where we don’t remember what order might have been used in other cases.

rabbitonweb:
This PR has to be closed (and reopened under new PR https://github.com/quasar-analytics/quasar/pull/2582) because of Travis bug (Travis is somehow reading old configuration trying to execute script (initConnector) that no longer exists). https://github.com/quasar-analytics/quasar/pull/2582 is the same PR as this one, but is not effected by the Trvis bug

PR #1937, opened 2017-03-02
Add support for queries with function definitions to query api

jedesah:
- Extend parser
- Thread `Blob` through in order to expose new compiler functionality through the web api
- Beginnings of "module" support
- Extend example config to show how to configure views and modules


core/src/main/scala/quasar/fs/mount/MountConfig.scala
```diff
@@ -50,21 +57,16 @@ object MountConfig {
...
   implicit def equal: Equal[MountConfig] =
-    Equal.equalBy(MountConfig.toConfigPair)
-
-  val toConfigPair: MountConfig => (String, ConnectionUri) = {
-    case ViewConfig(query, vars) =>
-      "view" -> ConnectionUri(viewCfgAsUri(query, vars))
-    case FileSystemConfig(typ, uri) =>
-      typ.value -> uri
-  }
+    Equal.equalA
```
wemrysi: This isn't safe as it ignores the `Equal` instances defined for all the other types used in `MountConfig`, you could do something like `Equal.equalBy[MountConfig, (Option[List[Statement[Fix[Sql]]], Option[(Fix[Sql], Variables)], Option[(FileSystemType, ConnectionUri)])](...)`.

frontend/src/main/scala/quasar/sql/Blob.scala
```diff
@@ -20,13 +20,25 @@ import quasar.Predef._
...
+    f(expr).map(Blob(_, scope))
+}
 
 object Blob {
   implicit def renderTree[T:RenderTree]: RenderTree[Blob[T]] =
     new RenderTree[Blob[T]] {
       def render(blob: Blob[T]) =
         NonTerminal("Sql Blob" :: Nil, None, List(blob.scope.render, blob.expr.render))
     }
+  implicit val functor: Functor[Blob] = new Functor[Blob] {
```
wemrysi: No need for `Functor` if you have `Traverse`.

web/src/test/scala/quasar/api/services/query/QueryFixture.scala
```diff
@@ -91,4 +95,9 @@ object queryFixture {
     liftMT[Task, ResponseT]             :+:
     failureResponseOr[FileSystemError]  :+:
     liftMT[Task, ResponseT].compose(fs)
+
+  def compileQuery(query: quasar.sql.Query): Fix[LogicalPlan] =
```
wemrysi: Is this used anywhere?

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -733,8 +733,8 @@ final class Compiler[M[_], T: Equal]
         val lpFunc = new HomomorphicFunction[T, T] {
           def arity = func.args.size
           def apply(args: List[T]): Option[T] = {
-            val argsMap = func.args.map(arg => scala.Symbol(arg.value)).zip(args).toMap
-            if (func.args.size != args.size) None else lpr.bindFree(argsMap)(body).toOption
+            val argsMap = func.args.zip(args).toMap
```
wemrysi: Using `Align[List]` here would make this a bit more straightforward and eliminate the need for the separate size check.

wemrysi:
Looks like this needs another merge from master.

jedesah:
@wemrysi Do you have any idea what's up with that first build failing after 3 minutes? It must have run a total of like 20 times now with all the restarts and merge commits I've pushed since the PR was done pretty much.

I haven't even changed something in the build so I can't fathom what code change would trigger a failure like this on only one build.

Do you think I should open another PR in case this is some kind of remanent of the AWS failure? We'll lose all the review history and stuff but it's probably worth it if you agree there is no other obvious reason why this could be failing.

wemrysi:
@jedesah I'm not sure, we saw a swath of those sorts of failures for many builds recently, but they seem to have cleared up for the most part. Sounds like trying a new PR couldn't hurt, just include a reference this one.

jedesah:
@wemrysi Cool. I'll try that just as soon as I convince myself it's not the change to the example config that somehow is causing this despite the confusing travis log.

PR #1945, opened 2017-03-07
Take advantage of Scalaz's Inject#unapply.

sellout:
This eliminates a ton of `unapply` methods on various functors that are intended to project a component of a Coproduct. This is now provided directly by the `Inject` instance.

There are still 16(?) unapply methods in the MongoDB connector. These are a bit more entrenched, so aren't being removed in this PR. All other `unapply`s in this style have been removed.


sellout:
This needs a review.

PR #1947, opened 2017-03-07
Add Order instances for EJson.

wemrysi:
Adds `OrderT` and `Delay[Order, Common|Obj|Extension]` instances. Also extracts `EJson` arbitrary instances from the `EJson` tests so they can be used elsewhere.

### DIFF
--- a/ejson/src/main/scala/quasar/ejson/EJson.scala
+++ b/ejson/src/main/scala/quasar/ejson/EJson.scala
@@ -16,23 +16,26 @@
 
 package quasar.ejson
 
-import quasar.Predef._
+import quasar.Predef.{Byte => SByte, Char => SChar, Map => SMap, _}
+import quasar.contrib.matryoshka._
 import quasar.fp._
 
 import matryoshka._
 import scalaz._, Scalaz._
 
-sealed trait Common[A]
+sealed abstract class Common[A]
 final case class Arr[A](value: List[A])    extends Common[A]
 final case class Null[A]()                 extends Common[A]
 final case class Bool[A](value: Boolean)   extends Common[A]
 final case class Str[A](value: String)     extends Common[A]
 final case class Dec[A](value: BigDecimal) extends Common[A]
 
-object Common {
+object Common extends CommonInstances {
   def unapply[F[_], A](fa: F[A])(implicit C: Common :<: F): Option[Common[A]] =
     C.prj(fa)
+}
 
+sealed abstract class CommonInstances extends CommonInstances0 {
   implicit val traverse: Traverse[Common] = new Traverse[Common] {
     def traverseImpl[G[_], A, B](
       fa: Common[A])(
@@ -48,20 +51,12 @@ object Common {
       }
   }
 
-  implicit val equal: Delay[Equal, Common] =
-    new Delay[Equal, Common] {
-      def apply[α](eq: Equal[α]) = Equal.equal((a, b) => (a, b) match {
-        case (Arr(l),       Arr(r))  => listEqual(eq).equal(l, r)
-        case (Arr(_),       _)       => false
-        case (Null(),       Null())  => true
-        case (Null(),       _)       => false
-        case (Bool(l),      Bool(r)) => l ≟ r
-        case (Bool(_),      _)       => false
-        case (Str(l),       Str(r))  => l ≟ r
-        case (Str(_),       _)       => false
-        case (Dec(l),       Dec(r))  => l ≟ r
-        case (Dec(_),       _)       => false
-      })
+  implicit val order: Delay[Order, Common] =
+    new Delay[Order, Common] {
+      def apply[α](ord: Order[α]) = {
+        implicit val ordA: Order[α] = ord
+        Order.orderBy(generic)
+      }
     }
 
   implicit val show: Delay[Show, Common] =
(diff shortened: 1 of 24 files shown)


ejson/src/main/scala/quasar/ejson/EJson.scala
```diff
@@ -76,32 +71,68 @@ object Common {
     }
 }
 
+sealed abstract class CommonInstances0 {
+  implicit val equal: Delay[Equal, Common] =
```
sellout: Could this have a comment re: how `Delay` breaks the normal inheritance stuff?

ejson/src/main/scala/quasar/ejson/EJson.scala
```diff
@@ -76,32 +71,68 @@ object Common {
...
-        mapShow(stringInstance, shw).show(a.value))
+      def apply[α](shw: Show[α]) = {
+        implicit val shwA: Show[α] = shw
+        Show.show(o => (o.value: SMap[String, α]).show)
+      }
+    }
+}
+
+sealed abstract class ObjInstances0 {
+  implicit val equal: Delay[Equal, Obj] =
```
sellout: Same.

ejson/src/main/scala/quasar/ejson/EJson.scala
```diff
@@ -165,7 +192,26 @@ object Extension {
         case Int(v)     => Cord(s"Int($v)")
       })
     }
+}
 
-  def fromObj[A](f: String => A): Obj[A] => Extension[A] =
-    obj => Map(obj.value.toList.map(_.leftMap(f)))
+sealed abstract class ExtensionInstances0 {
+  implicit val equal: Delay[Equal, Extension] =
```
sellout: Same.

ejson/src/main/scala/quasar/ejson/package.scala
```diff
@@ -27,31 +27,40 @@ import scala.math.{BigDecimal, BigInt}
...
+
+  def nul[A] =
+    Prism.partial[Common[A], Unit] { case Null() => () } (κ(Null()))
+
+  def str[A] =
+    Prism.partial[Common[A], String] { case Str(s) => s } (Str(_))
 
   def obj[A] =
-    Prism.partial[Obj[A], ListMap[String, A]] { case Obj(o) => o } (Obj(_))
+    Iso[Obj[A], ListMap[String, A]](_.value)(Obj(_))
```
sellout: Good catch.

foundation/src/main/scala/quasar/contrib/matryoshka/OrderT.scala
```diff
@@ -0,0 +1,38 @@
...
+@typeclass
+trait OrderT[T[_[_]]] {
+  def order[F[_]: Functor](tf1: T[F], tf2: T[F])(implicit del: Delay[Order, F]): Ordering
+
+  def orderT[F[_]: Functor](delay: Delay[Order, F]): Order[T[F]] =
+    Order.order[T[F]](order[F](_, _)(Functor[F], delay))
+}
+
+object OrderT {
+  implicit def recursiveT[T[_[_]]: RecursiveT]: OrderT[T] =
```
sellout: So, I’ve avoided type class instances like this. I tend to make this instance non-implicit, then create implicits for `Fix`, etc. specifically. I just get nervous about magically creating instances for types we don’t know about yet.

I can be talked out of it perhaps, though.

wemrysi:
@sellout I think this may no longer be correct after the most recent discussion with @jdegoes as we want `Extension.Map` to preserve order, but not consider it during equality. Correct? This implementation considers the key order in equality.

If you agree, I'll push an update.

foundation/src/main/scala/quasar/contrib/matryoshka/OrderT.scala
wemrysi: Sure, that's fine, I'll remove the implicitness and add the explicit instances.

sellout:
I’m not sure – I saw that order is important, but that it doesn’t affect type. Don’t recall whether or not it affects equality (although my guess is that it doesn’t). @jdegoes?

jdegoes:
@sellout It should not affect type or equality. I would preserve order in so far as it's economically possible to do so; I'd regard order-preserving as part of the semantics, but equality should ignore order, and types should compare equal regardless of order.

ejson/src/main/scala/quasar/ejson/EJson.scala
wemrysi: It does prevent a `Delay[Order, F]` from being used as a `Delay[Equal, F]` since `Order` appears on both sides of the function, but I think we'd still want to define both regardless as they depend on the type parameter to `F`, so `Delay[Order, F]` doesn't help you when there is no `Order[A]` for `F[A]`, similarly to why both `Equal[A] => Equal[List[A]]` and `Order[A] => Order[List[A]]` are both defined.

sellout: Ah, that’s a good point.

wemrysi:

foundation/src/main/scala/quasar/contrib/matryoshka/OrderT.scala
```diff
@@ -29,10 +30,17 @@ trait OrderT[T[_[_]]] {
...
-  implicit def recursiveT[T[_[_]]: RecursiveT]: OrderT[T] =
+object OrderT extends OrderTInstances {
+  def recursiveT[T[_[_]]: RecursiveT]: OrderT[T] =
     new OrderT[T] {
       def order[F[_]: Functor](tf1: T[F], tf2: T[F])(implicit del: Delay[Order, F]): Ordering =
         del(orderT[F](del)).order(tf1.project, tf2.project)
     }
 }
+
+sealed abstract class OrderTInstances {
```
sellout: These cyclic dependencies absolutely amaze me.

PR #1949, opened 2017-03-07
Eliminate `sealed trait`.

sellout:
Most have been converted to `sealed abstract class`. One was actually a type class, so I just unsealed it.

There are still a few remaining in Mongo (I feel like I write this comment in every PR), which are actually used with multiple inheritance, so not eliminable yet.


drostron:
Merge script requests a merge from master

sellout:
Passed.

PR #1957, opened 2017-03-10
Add support for queries with function definitions to query api

jedesah:
- Extend parser
- Thread Blob through in order to expose new compiler functionality through the web api
- Beginnings of "module" support
- Extend example config to show how to configure views and modules


jedesah:
Same as #1937. Reopening in the hopes it makes the build happy

jedesah:
Okay, so still failing. I'll have to look into this further later

example-quasar-config.json
```diff
@@ -4,6 +4,15 @@
...
             }
+        },
+        "/myview": {
+            "view": {
+                "connectionUri": "sql2://q=\"select bar from biz\""
+            }
+        },
+        "/mymodule": {
+            "module":
+                "CREATE FUNCTION ARRAY_LENGTH(:foo) BEGIN COUNT(:foo[_]) END; CREATE FUNCTION USER_DATA(:user_id) BEGIN SELECT * FROM `/root/path/data/` WHERE user_id = :user_id END"
```
jedesah: @wemrysi Do you think I should make this a JSON array?

jedesah:
@wemrysi Sorry for the huge delay on this one. I finally figured out that it was probably a compiler crash that was causing the first build to fail and I think I fixed it. We'll see! :crossed_fingers:

example-quasar-config.json
wemrysi: I'm not sure what the advantage an array would bring due to there being syntax for separating statements, thus each array element wouldn't necessarily correspond to a single statement. Maybe if we didn't have such syntax an array would be useful (necessary?) to support multiple statements.

Let me know if I've missed the point here, or if you had some other thoughts why an array would be useful.

jedesah: We could use a parser that only parses a single statement instead of the one we are using now. The problem is that JSON doesn't support multiline strings. Making the value of `module` a json array would probably result in easier to read config files.

Don't know if this is a very serious problem though...

jedesah:
😞 Now I am hitting some kind of coverage minimum. I thought my additions were relatively well tested. Perhaps not...

example-quasar-config.json
wemrysi: @drostron can confirm, but I think we're moving the metastore to quasar and mounts will reside therein instead of in the config file, so readability probably isn't much of a concern.

drostron: Yes, mount configuration will soon live in the metastore.

jedesah:
@wemrysi This should be ready to merge if you think the latest changes are reasonable. In particular, I disabled the coverage minimum for the `common` project because of the `scoverage` upgrade (which I thought might help fix the compiler crash and included anyway despite it not doing so)

Issue #1900, opened 2017-02-09
Support configuration for production Spark clusters


jdegoes:
TODO:

- [x] Propose`connectionURI` specification for front-end team
- [x] Implement new `connectionURI`
- [x]  New `connectionURI` implemented on front-end side
- [x] BUG: on backend wrong sparkUri is created

rabbitonweb:
We need to add ability to give our users way to configure connection with Spark. Right now it is available only via environment variables, but we might have clients willing to keep that configuration internal (within the quasar).

rabbitonweb:
Things we need to provide as part of configuration for our users to set:
(details at http://spark.apache.org/docs/latest/configuration.html)

**For Executors**
- spark.executor.memory
- spark.executor.cores
- spark.executor.extraJavaOptions
- spark.default.parallelism
- spark.files.maxPartitionBytes

**For Diver**
- spark.driver.cores
- spark.driver.maxResultSize
- spark.driver.memory
- spark.local.dir


**Shuffle & Memory** 
- spark.reducer.maxSizeInFlight
- spark.reducer.maxReqsInFlight
- spark.shuffle.file.buffer
- spark.shuffle.io.retryWait
- spark.memory.fraction
- spark.memory.storageFraction

**Other**
- spark.cores.max
- spark.speculation
- spark.task.cpus

TODO:

- [ ] Make those configurations part of `connectionURI`
- [ ] Update SlamData to give ability to provide those values

rabbitonweb:
Specification for new `connectionUri`

PR #1958, opened 2017-03-10
Change the way spark-hdfs connectionUri is defined, add spark-cluster configs

rabbitonweb:
Part of https://github.com/quasar-analytics/quasar/issues/1900

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
@@ -30,6 +30,7 @@ import quasar.physical.sparkcore.fs.hdfs.writefile.HdfsWriteCursor
 import java.net.URI
 import scala.sys
 
+import org.http4s.{ParseFailure, Uri}
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem => HdfsFileSystem}
 import org.apache.spark._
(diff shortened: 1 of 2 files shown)


rabbitonweb:
I'm proud of the fact that I took advantage of `Applicative` ;)

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -52,25 +53,63 @@ package object hdfs {
...
-      case _ =>
-        error("Missing master and prefixPath (seperated by |)" +
-          " e.g spark://spark_host:port|hdfs://hdfs_host:port|/user/")
-    }
+  val parseUri: ConnectionUri => DefinitionError \/ (SparkConf, SparkFSConf) = (connUri: ConnectionUri) => {
+
+    def liftErr(msg: String): DefinitionError = NonEmptyList(msg).left[EnvironmentError]
+
+    def master(uri: Uri): State[SparkConf, Unit] =
+      State[SparkConf, Unit](c => (c.setMaster(s"spark://${uri.host}:${uri.port}"), ()))
```
alissapajer: You can use [`modify`](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/State.scala#L31) here and in a couple places below, if you prefer.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -52,25 +53,63 @@ package object hdfs {
...
+        config("spark.memory.fraction", uri) *>
+        config("spark.memory.storageFraction", uri) *>
+        config("spark.cores.max", uri) *>
+        config("spark.speculation", uri) *>
+        config("spark.task.cpus", uri)
+      ).exec(new SparkConf())
+    )
+
+    val hdfsUrlOrErr: DefinitionError \/ String = uriOrErr.flatMap(uri =>
+      uri.params.get("hdfsUrl").fold(liftErr("'hdfsUrl' parameter not provider").left[String])(_.right[DefinitionError])
```
alissapajer: typo: s/provider/provided/g

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -52,25 +53,63 @@ package object hdfs {
...
+    ) 
+    val rootPathOrErr: DefinitionError \/ ADir =
+      uriOrErr
+        .flatMap(uri =>
+          uri.params.get("rootPath").fold(liftErr("'rootPath' parameter not provider").left[String])(_.right[DefinitionError])
```
alissapajer: Same typo.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -52,25 +53,63 @@ package object hdfs {
...
+          posixCodec.parseAbsDir(pathStr)
+            .map(sandboxAbs)
+            .fold(liftErr("'rootPath' is not a valid path").left[ADir])(_.right[DefinitionError])
+        )
+
+    for {
+      sparkConf <- sparkConfOrErr
+      hdfsUrl <- hdfsUrlOrErr
+      rootPath <- rootPathOrErr
+    } yield (sparkConf, SparkFSConf(sparkConf, hdfsUrl, rootPath))
```
alissapajer: Since `sparkConf` is contained within the `SparkFSConf` could you just return the `SparkFSConf` here instead of the tuple?

rabbitonweb: Let's see if I can explain this :)

Each connector that is based on sparkcore is required to return a pair `(SparkConf,T)` - see https://github.com/quasar-analytics/quasar/blob/master/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/package.scala#L35

Here that T is `SparkFSConf` but sparkcore does not really care about what is inside of that `T` - it needs ability to fetch `SparkConf`... maybe there is a way to say "`T` is some type that has `SparkConf` in it" - but I could not think of the right syntax - that's why I'm returning a pair

rabbitonweb: btw I'm awayre that `SparkFSConf` & `SparkConf` are confusing and I'm planning to rename `SparkFSCons` to something more meaningful in near future

rabbitonweb:
@alissapajer `modify` added (thx for hint), typos fixed

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
alissapajer: Ah ok, that makes sense. If `T` will always have a `SparkConf` in it (for all connectors based on sparkcore), then maybe you could pass a `SparkConf` and a `T` to `fsInterpret` in `fs.definition`, instead of just a `T`. But I think is fine for now - just something to look out for if/when more connectors based on sparkcore are added.

rabbitonweb: 👍

PR #1959, opened 2017-03-11
Enable integration test with LIKE and variable

### LINKED ISSUE #1695: Query involving 'LIKE' results in a scalac MatchError
Attempting to run the query with variables specified in the `likeWithConstantFoldablePattern` regression test results in a Scala `MatchError` at `MapFunc.scala:749`.

alissapajer:
Closes #1695.

### DIFF
--- a/it/src/main/resources/tests/likeWithConstantFoldablePattern.test
+++ b/it/src/main/resources/tests/likeWithConstantFoldablePattern.test
@@ -1,18 +1,14 @@
 {
     "name": "LIKE with constant-foldable pattern",
     "backends": {
-        "couchbase":      "skip",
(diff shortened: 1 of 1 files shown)


it/src/main/resources/tests/likeWithConstantFoldablePattern.test
```diff
@@ -1,18 +1,14 @@
 {
     "name": "LIKE with constant-foldable pattern",
     "backends": {
-        "couchbase":      "skip",
-        "marklogic_json": "pending",
-        "marklogic_xml":  "pending",
```
alissapajer: The pending tests didn't come alive because we had to ignore the "_id" field.

Issue #1966, opened 2017-03-14
MongoDbQJsStdLibSpec Century TemporalTrunc failure


drostron:
[Travis log](https://travis-ci.org/quasar-analytics/quasar/jobs/210648450#L711)
```
[info] TemporalTrunc
[error] x timestamp (941 ms)
[error]  Falsified after 12 passed tests.
[error]  > ARG_0: 2000-03-01T06:15:45.204Z
[error]  > ARG_1: Century
[error]  > List(Obj(Map(value -> Timestamp(1900-01-01T00:00:00Z)))) is a single result but Timestamp(1900-01-01T00:00:00Z) does not match Timestamp(2000-01-01T00:00:00Z) (MongoDbQStdLibSpec.scala:133)
```

PR #1967, opened 2017-03-14
Fix JsFuncHandler TemporalTrunc Century handling

drostron:
Fixes: #1966


frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -605,6 +605,19 @@ abstract class StdLibSpec extends Qspec {
...
+
+          truncZonedDateTime(y, t).fold(
+            e => Failure(e.shows),
+            tt => unary(
+              TemporalTrunc(y, _).embed,
+              Data.Timestamp(x),
+              Data.Timestamp(tt.toInstant)))
+        }
+
         "timestamp" >> prop { (x: Instant, y: TemporalPart) =>
```
sellout: Would probably be a good idea to explicitly property-test each of the values in `TemporalPart`.

drostron: Agreed.

drostron:
Re recent discussion, I'm reminded that we might be able to leverage 47deg's scalacheck-datetime now that our scalacheck version has been updated. I'll give that a quick whirl while in the vicinity.

drostron:
`scalacheck-datetime` doesn't have the distribution shaping I was looking for so leaving `DateArbitrary` as is for now.

drostron:

PR #1968, opened 2017-03-14
Rely on sbt-slamdata and slamdata-predef

drostron:
Depends on slamdata/sbt-slamdata/pull/16


project/project/plugins.sbt
```diff
@@ -1,2 +1,4 @@
+resolvers += Resolver.sonatypeRepo("releases")
+
 addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-M15")
-addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.5.0")
+addSbtPlugin("com.slamdata"    % "sbt-slamdata" % "0.0.12")
```
djspiewak: This doesn't need to be here, unless I'm missing something that sbt-slamdata offers.  You don't need it in the meta project.

drostron: It provides `commonScalacOptions_2_10` used in `project/plugins.sbt`.

https://github.com/drostron/quasar/blob/2e092d998ec761f3b7631efce1bd29556fa1a990/project/plugins.sbt#L11

At some point I'd like to share at least plugin versions via `sbt-buildinfo` or similar.

djspiewak: Oh, because we have special options for our project compiler.  That's… weird.  At some point we should probably split these plugins, but it's fine for now.

drostron: Yeah, definitely room for sbt-slamdata evolution as we start to live with it for a bit.

drostron:
@djspiewak Travis appeased and ready for another look.

djspiewak:
@drostron What was wrong with the `TravisCIPlugin`?

drostron:
@djspiewak When enabled at that level supposedly it was defaulting to Scala 2.10 (due to absent `.travis.yml`), yet resolving occurred for sbt-slamdata against 2.11 which was failing.

djspiewak:
Oh, this is because of using sbt-slamdata in the meta project.  RIP

Just so we don't get confused about the level at which this is being disabled, I would actually move that line to a separate `project/build.sbt`.  It will have the same effect, but it's a bit more explicit that we're disabling things *in the meta project*.  As it stands, it looks (especially to someone who doesn't know SBT well) that the travisci plugin is being disabled in the primary project.

drostron:
Agreed — will do.

drostron:
Ready

project/Dependencies.scala
```diff
@@ -5,68 +5,65 @@ import java.lang.{String, System}
...
   // TODO: Upgrade to `0.15.2a` (or above) once we can figure out a fix for:
   // https://github.com/quasar-analytics/quasar/issues/1852
   // Although this issue will be closed by the current commit that downgrades it,
   // it's still an issue that needs to be considered for anyone attempting to upgrade
+  // NB: Argonaut RC2 conflicts with http4s 0.14.1a, but we're evicting to it anyway
   private val http4sVersion     = "0.14.1a"
   private val jawnVersion       = "0.8.4"
   private val jacksonVersion    = "2.4.4"
-  private val monocleVersion    = "1.3.2"
+  private val matryoshkaVersion = "0.16.4"
```
sellout: Just curious if you tried upgrading Matryoshka. 0.16.10 in particular is the version that added sbt-slamdata support.

drostron: Gave it a whirl just now and ran into the issue described in slamdata/matryoshka#75. I'll submit a fix for that momentarily. Think I'll wait on upgrading Matryoshka here though, since updating to the latest will require more than a version bump.

PR #1969, opened 2017-03-14
Fix various things in the Mongo-QScript connector.

sellout:
- LeftShift
- Reduce
- elide Guards more often
- handling of array concat

Fixes ~30 integration tests.

Also
- adds `First` and `Last` reductions, using `First` for `DistinctBy`
- eliminates `Sort` before `Reduce` when none of the reductions are order-dependent.


mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -853,15 +857,15 @@ object MongoDbQScriptPlanner {
                 getReduceBuilder[T, M, WF, EX](
                   funcHandler)(
                   red.map(_.sequence).sequence.fold(
-                    κ(GroupBuilder(ArrayBuilder(src, red.unite), // FIXME: Doesn’t work with UnshiftMap
+                    κ(GroupBuilder(DocBuilder(src, red.unite.zipWithIndex.map(_.map(i => BsonField.Name("f" + i.toString)).swap).toListMap), // FIXME: Doesn’t work with UnshiftMap
```
wemrysi: It'd be worth encapsulating how the reduction field name is constructed in a function like `reduceField(i: Int): String` to dry up all the call sites.

wemrysi:
@sellout As it isn't obvious to me, what does the change from arrays to documents in the `GroupBuilder` for reductions improve?

sellout:
There are two things with the Array => Doc change. One is that MongoDB doesn’t handle arrays as a “top level” value – ArrayBuilder is useful along the way if you don’t know what’s going to happen to that array – it could end up as a value inside a DocBuilder or something, and that’d be fine. But if we reach the end of the line and it’s still an ArrayBuilder, then we actually have to convert it to a document, and other tricky stuff to make sure references to it are updated, etc. In this case, we’re building a GroupBuilder immediately after, so we know it’ll stay an ArrayBuilder. So, using ArrayBuilder isn’t helpful there.

But, it’s actually problematic here, because there are times when we add values to Builders, and it can affect the indices. And we _should_ be tracking that and updating references, but there’s a bug somewhere. At least we notice it and give a runtime failure, but it’s in the guts of the nastiest code in Quasar that should finally go away with the old Mongo.

sellout:
Updated, but unfortunately we don’t have the faster builds yet …

PR #1970, opened 2017-03-15
Ignore fewer warts at the project level

alissapajer:
Ignore fewer warts at the project level. This forces us to explicitly recognize when we are disabling a wart.


marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
```diff
@@ -24,7 +24,9 @@ import java.lang.SuppressWarnings
 import scalaz.{Order => _, _}
 import scalaz.std.iterable._
 
-@SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+@SuppressWarnings(Array(
+  "org.wartremover.warts.DefaultArguments",
+  "org.wartremover.warts.Overloading"))
```
sellout: I would prefer to see these on the specific methods.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/fn.scala
```diff
@@ -23,7 +23,9 @@ import java.lang.SuppressWarnings
 
 import eu.timepit.refined.auto._
 
-@SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+@SuppressWarnings(Array(
+  "org.wartremover.warts.DefaultArguments",
+  "org.wartremover.warts.Overloading"))
```
sellout: Same.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/map.scala
```diff
@@ -20,6 +20,7 @@ import quasar.Predef._
 
 import scalaz.{Foldable, IList}
 
+@SuppressWarnings(Array("org.wartremover.warts.Overloading"))
```
sellout: Et cetera.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/syntax.scala
```diff
@@ -72,6 +72,7 @@ object syntax {
       this := expression
   }
 
+  @SuppressWarnings(Array("org.wartremover.warts.Overloading"))
```
sellout: @wemrysi What is with all the overloading and default arguments? 😛

wemrysi: Its a dsl designed to closely resemble `XQuery`, which includes lots of both.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
alissapajer: @sellout yeah I agree that's better. I'll make the change throughout.

If we have overloaded say `def foo` twice, it's enough (to make the compiler happy) to put the warning suppression on just one of the function defs. Do you if there is a preferred practice to try to put the warning suppression on _all_ overloaded defs?

alissapajer:
@sellout also, btw, I decided not to touch the mongodb project with regards to warts since that's all changing.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
sellout: I’m fine with whichever. In the past I’ve labeled one of the two (don’t think I’ve come across three before).

sellout:
Good call, because that would also lead to hundreds of suppressions, probably 😆 But yeah, I’m happy to attempt that once we’ve switched over.

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
alissapajer: I'll just label the minimum necessary to make it compile, since we cannot enforce labeling more anyway.

alissapajer:
And then in the build.sbt there are still 5 warts remaining that are disabled for the entire repo (some because of upstream bugs). We can address those separately when possible.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/StructuralPlanner.scala
```diff
@@ -32,6 +33,7 @@ import scalaz._, Scalaz._
   * @tparam F   the effects employed by the library.
   * @tparam FMT type index representing the data format supported by the library.
   */
+@SuppressWarnings(Array("org.wartremover.warts.Overloading"))
```
wemrysi: Sigh, probably due to `toString`, which isn't intentionally overloaded. Mind just renaming it (maybe `asString`)?

alissapajer: Will do!

marklogic/src/main/scala/quasar/physical/marklogic/xquery/FunctionDecl.scala
```diff
@@ -101,6 +101,7 @@ object FunctionDecl {
     def parameters = NonEmptyList(param1, param2, param3, param4, param5)
   }
 
+  @SuppressWarnings(Array("org.wartremover.warts.Overloading"))
```
alissapajer: For these few case classes with each a handful of overloaded `apply` methods I put the warning suppression at the start of the case class instead of before each `apply` method.

alissapajer:
@sellout and @wemrysi addressed your comments.

marklogic/src/main/scala/quasar/physical/marklogic/xml/NCName.scala
```diff
@@ -26,11 +26,12 @@ import scalaz.std.string._
 import scalaz.syntax.show._
 import scalaz.syntax.std.either._
 
-final case class NCName(value: String Refined IsNCName) {
+final case class NCName private (value: String Refined IsNCName) {
```
wemrysi: Is the `private` modifier related to the overloading?

alissapajer: Not directly. I noticed that for some case classes we had `apply` methods in the companion object and figured that if the constructor for the class could  be `private`, it should be.

wemrysi: Ah, ok. I'm pretty sure that for case classes, marking the constructor `private` doesn't change the access level of the auto-generated `apply` method in the companion, so it might be a bit confusing.

This is another case where the overloaded apply method should probably just be renamed to `fromString` or something, sorry I missed it earlier. Though if this is ready to merge otherwise, I can submit another PR to fix it.

alissapajer: Ah yeah, there is still the auto-gen `apply` that the case class creates (which is what causes the overloading in the first place :))

I'll make the change here in this PR to rename all the overloaded `apply` methods in this situation (there are 3 maybe?) and then remove the `private` modifier. Might as well have these changes all together :)

alissapajer:
@wemrysi added you as a reviewer so you can officially sign off on the PR when you're ready. Should have done that in the first place since it touches marklogic a bunch :)

alissapajer:
@sellout @wemrysi ready for you to take another look.

alissapajer:
@drostron this should be good to go now!

PR #1972, opened 2017-03-15
Order of EJson Meta nodes should only consider the value component.

wemrysi:
This fixes a regression I introduced when adding the `Order` instances for `EJson` in #1947, I've fixed the regression and added a test that verifies it.

I also updated `Arbitrary[Extension[A]]` to actually generate `Meta` values.


ejson/src/main/scala/quasar/ejson/EJson.scala
```diff
@@ -182,11 +182,12 @@ sealed abstract class ExtensionInstances {
 
   ////
 
+  // NB: This elides metadata as it isn't considered for purposes of equality.
   private[ejson] def generic[A: Order](e: Extension[A]) = (
```
alissapajer: It seems that someone could use this method `generic` by accident to construct some instance, without wanting to actually elide the metadata within.

Maybe it would be better to elide the metadata inside the creation of the `Delay[Equal, Common]` implicit rather than here, to make sure this method is actually "generic".?

wemrysi: Good idea, will do!

wemrysi:
@alissapajer Ready for another look.

Issue #1716, opened 2016-11-21
Basic Query Analysis


jdegoes:
**Query Analysis**. Users have no idea the impact of their queries. Even basic query analysis could help provide sensible feedback to the SlamData user-interface, as well as the user, on whether they can expect to wait seconds, minutes, or hours; and on the load exerted on the data source by the query.

SlamData can make much better decisions if it knows how complex a query will be to execute.

* **Instant** — There is negligible cost to execute the query.
* **Interactive** — There is some cost to execute the query, but only in the range of a few seconds to low tens of seconds.
* **Semi-Interactive** — The query will execute in minutes.
* **Batch** — The query requires hours or days to execute.
* **Unknown** — There is no possible way to estimate how much time executing the query would require, without actually executing it and measuring the result. **This should be used sparingly by connectors as it provides no information to the front-end.**

This functionality has to be pushed into the connector, since there's no way anything higher could know how long a given query would take.

This should be incorporated into the explain-type API which compiles the query.

- [x] QA typeclasss to collect info about qscript template
- [x] Implement Cardinality typeclass
- [x] Implement generic Cost typeclass
- [x] change `pathCard` to return `M[Int]`
- [x] Implement generic `pathCard` that uses `QueryFile` interpreter & does simple `select * from x`
- [x] mapping cost value to `AnalysisResult` - a manual, explicit tuning
- [x] add REST endpoint /analysis

rabbitonweb:
- [ ] QA typeclasss to collect info aobut qscript template

- [ ] Implement QA typeclass

- [ ] Implement QA algebra for sparkcore

- [ ] Implement QA algebra for makrlogic

- [ ] Implement QA algebra for couchbase

rabbitonweb:
This issue is closed.

With a rise of brand new `BackendModule` we will need to add implementation for it. For details look in https://github.com/quasar-analytics/quasar/issues/2213

PR #1977, opened 2017-03-17
Query Analysis first attempt: Cardinality & Generic Cost

rabbitonweb:
Part of #1716


rabbitonweb:
You should see this comment in Slack....

connector/src/main/scala/quasar/qanalysis/package.scala
```diff
@@ -0,0 +1,149 @@
+/*
```
alissapajer: Testing PR review comment for Slack.

connector/src/main/scala/quasar/qanalysis/package.scala
```diff
@@ -0,0 +1,149 @@
...
+ * Copyright 2014–2017 SlamData Inc.
```
alissapajer: Testing PR review comment for Slack take 2.

connector/src/main/scala/quasar/qanalysis/package.scala
```diff
@@ -0,0 +1,149 @@
...
+ *
```
alissapajer: test comment 3

rabbitonweb:
Missing elements (will be part of separate PR):

1. change return type of `Cardinality` to return range `(Int, Option[Int]) where `fst` is minimal and `snd` is `None` if unknown or Some(max)
2. change `pathCard` to return `M[Int]` instead of `Int` where `M : Monad`

rabbitonweb:
@sellout pls review

connector/src/main/scala/quasar/qanalysis/package.scala
```diff
@@ -0,0 +1,227 @@
...
+    implicit def qscriptCore[T[_[_]]: RecursiveT: ShowT]: Cost[QScriptCore[T, ?]] =
+      new Cost[QScriptCore[T, ?]] {
+        def evaluate(pathCard: APath => Int): GAlgebra[(Int, ?), QScriptCore[T, ?], Int] = {
+          case qscript.Map((card, cost), f) => card + cost
+          case Reduce((card, cost), bucket, reducers, repair) => card + cost
+          case Sort((card, cost), bucket, orders) => card + cost
+          case Filter((card, cost), f) => card + cost
+          case Subset((card, cost), from, sel, count) => card + cost
+          case LeftShift((card, cost), struct, id, repair) => card + cost
+          case Union((card, cost), lBranch, rBranch) => card + cost
```
sellout: This case should look like the Join cases.

connector/src/main/scala/quasar/qanalysis/package.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  @typeclass
+  trait Analyzer[F[_]] {
+    def analyze[T](t: T)(pathCard: APath => Int)(implicit
+      rec: Recursive.Aux[T, F],
+      func: Functor[F],
+      cardinalty: Cardinality[F],
+      cost: Cost[F]
+    ): Int = rec.zygo(t)(cardinalty.calculate(pathCard), cost.evaluate(pathCard))
```
sellout: Actually, why is this even in a type class? You can just expose the function directly.

connector/src/main/scala/quasar/qanalysis/package.scala
```diff
@@ -0,0 +1,227 @@
...
+  case object Unknown extends AnalysisResult
+
+  @typeclass
+  trait Cardinality[F[_]] {
+    def calculate(pathCard: APath => Int): Algebra[F, Int]
+  }
+
+  @typeclass
+  trait Cost[F[_]] {
+    def evaluate(pathCard: APath => Int): GAlgebra[(Int, ?), F, Int]
```
rabbitonweb: @sellout `evaluate` needs `pathCard` becaue when we deal with `EquiJoin` and `ThetaJoin` we need to call `zygo` on both left and right branch

rabbitonweb:
@sellout i will go with range in separate PR. I don't think I will put any probabilistic for `4.2`

rabbitonweb:
@sellout this is ready for review

sellout:
@rabbitonweb There are still a bunch of comments you haven’t addressed.

connector/src/main/scala/quasar/qanalysis/package.scala
rabbitonweb: Why is `analyze` a type class? I was thinking that we could have a single entry point for connector creators to use

sellout: A single entry point how? Do you expect connectors to override `analyze`? It definitely doesn’t need to be parameterized on `F`, as if it is, you’ll need to create empty(?) instances for each component of the coproduct.

sellout:
@rabbitonweb Just want to make sure you’re aware this build isn’t passing.

PR #1980, opened 2017-03-20
Add log4j.properties killing Spark verbosness

rabbitonweb:
Missing log4j properties to kill the spark's verbose output


sellout:
I see this is in the `repl` project – does this also kill the logging output in the tests … maybe in the `spark` project? Or is it in `it`?

rabbitonweb:
@sellout the one in testes are already killed, this is the last missing piece that also kills them in repl

sellout:
Ah, ok. Cool 😄

alissapajer:
It looks like these configs are already duplicated in `sparkcore/src/test/resources/log4j.properties` and `it/src/test/resources/log4j.properties`. Could we put the log4j configs in only one file and then provide a path to them at all the usage places? Alternatively we could use symlinks, but I'm not sure if we want symlinks in our repo; it might be better to maintain out-of-the-box compatibility with all file systems.

rabbitonweb:
troll_on symlinks in repo? what about windows users  troll_off

we might want to have different configs for repl, web and tests - i dont mind duplication but if u guys think we should have a single file with other files pointing to it - i'm fine with it. however i still prefer the explicit approach where each "context" has its own config

alissapajer:
@rabbitonweb I don't think we should have duplicate identical configs, if we can avoid it. Can we take the single file approach here, having other files pointing to the main config? Then if we want to adjust the logging config everywhere we only have to do it in one place. And that should still allow us to add specific configs to specific contexts, if necessary.

alissapajer:
@rabbitonweb can you take a look at this again?

rabbitonweb:
@alissapajer first thing tomorrow in the morning

alissapajer:
@rabbitonweb _ping_

sellout:
@rabbitonweb I would say that `sparkcore` may be a better place than `common`, since (at least for now) all this logging stuff is Spark-specific.

Although if we later have other things that use log4j, I don’t know what the solution is, because if multiple connectors include it in their resources, they’ll trample each other, right?

rabbitonweb:
@sellout `sparkcore` would be good for now, but as u mentioned, if sombody would like to addd something to log4j not spark related, he would have to move the file to more general project. `common` seem the right place for me

sellout:
@rabbitonweb There are two problems with common, though – 1. backward references – i.e., it refers to spark, which doesn’t exist at this level, and 2. any log4j stuff is going to be only in the connectors – common shouldn’t know anything about log4j.

So, the _most_ I would push it up is to the `connector` project. But even that still has both problems (just slightly less so, since it’s lower in the dependency graph). Also, in the near future, sparkcore will be in a separate repository from common and connector, separating the logging config from the logging dependency even more.

Hopefully we won’t have any other connectors that require us to deal with log4j, so I don’t really want to increase coupling to avoid maybe having to think about how to resolve that conflict on the off chance it happens in future.

rabbitonweb:
@sellout I can move it to sparkcore for now. And if log4j will be needed some place else, we will come back to this issue

PR #1996, opened 2017-03-21
Cache target directories, but ONLY for PR builds

djspiewak:
Also disables coverage on PRs (for now) since it interferes with caching.


scripts/build
```diff
@@ -53,22 +53,34 @@ if [[ ${LOCAL_MONGODB:-} == "true" ]] ; then
   echo "Using local MongoDB config"
 fi
 
+# if we're building a pull request, DON'T clean and use the cache instead
```
alissapajer: I expect that at some point we'll hit a situation where we'll need to `clean` when running a PR, due to the contents of the PR. Would it be possible to allow a user to specify for a particular PR that `clean` should be run? Or actually, will clearing the travis cache be enough to trigger a `clean`?

alissapajer: Well, clearing the travis cache would effectively be triggering a clean by deleting targets without actually calling `clean` itself, which should be fine.

build.sbt
```diff
@@ -173,6 +184,7 @@ lazy val foundation = project
   .settings(name := "quasar-foundation-internal")
   .settings(commonSettings)
   .settings(publishTestsSettings)
+  .settings(targetSettings)
```
sellout: Should we just `commonSettings ++= targetSettings` somewhere, or are there some projects that don’t use `targetSettings`?

scripts/build
```diff
@@ -81,10 +93,10 @@ else
   # Compiling `connector` can take awhile, so allow it some extra time when
   # running on travis to avoid having the build killed due to lack of output.
   if [[ ${TRAVIS:-} ]]; then
-    travis_wait $TRAVIS_MAX_CONNECTOR_COMPILE_MINS "$SBT" -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION connector/test:compile
+    travis_wait $TRAVIS_MAX_CONNECTOR_COMPILE_MINS "$SBT" -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION $CLEAN connector/test:compile
   fi
 
-  # Not the first job in Travis; run only integration tests (no sense
+  # Not the first job in Travis; run only integration tests (no sensek
```
sellout: Inserted a typo in the comment.

djspiewak: …or is it? 😎

build.sbt
djspiewak: `root` doesn't use `targetSettings`

PR #1997, opened 2017-03-21
Add Modules to Filesystem

jedesah:
- Modules appear in directory listings
- Module functions appear in filesystem as file and metadata lists names of arguments to function
- Modules can be mounted, unmounted, moved
- Modules can be configured directly in config file


jedesah:
@drostron Sorry for the delay on addressing those minor issues!

drostron:
@jedesah No worries! Now we just need to get Travis to weigh in.

beckyconning:
Ooooh! : D

drostron:
@jedesah Needs a semantic label

jedesah:
@drostron I'm on my phone but I think feature is appropriate so if you
agree go ahead and add that label before running the merge script

drostron:
@jedesah Could it be `version: breaking` as in adding functionality in a backwards compatible way, or is there an aspect that breaks backwards compatibility?

jedesah:
@drostron I'm confused. Did you mean `version: feature`. That's what I was thinking since I don't think it breaks backwards compatibility.

drostron:
@jedesah Indeed, had them switched. Maybe a simple legend is in order somewhere. `version: feature` it is.

drostron:
Updated our [Versioning](https://github.com/quasar-analytics/quasar/wiki/Versioning) to have a clearer correlation between our labels and semver's.

PR #2023, opened 2017-03-28
Basic join branch array pruning

### LINKED ISSUE #1794: PruneArrays for ThetaJoin
Consider the qscript in #1757 as an example. 

Both `LeftSide` and `RightSide` branches are arrays, which are dereferenced in the `combine` (or in other examples the `on`) of the `ThetaJoin`. We can perform array pruning in this case, removing unused elements from the arrays and reindexing the derefs.

alissapajer:
If the left and/or right branch of a qscript join is a `LeftShift` with a `combine` func that is an array, then in some cases we can rewrite the array produced by that `LeftShift`, namely when only some (but not all) elements of that array are used in the join condition and join combine.

Closes #1794

### DIFF
--- a/connector/src/main/scala/quasar/qscript/PruneArrays.scala
+++ b/connector/src/main/scala/quasar/qscript/PruneArrays.scala
@@ -154,6 +154,22 @@ class PAHelpers[T[_[_]]: BirecursiveT: OrderT: EqualT] extends TTypes[T] {
   def rewriteRepair(array: ConcatArrays[T, JoinFunc], seen: SeenIndices): JoinFunc  =
     arrayRewrite(array, seen.map(_.toInt).toSet)
 
+  def rewriteBranch(branch: FreeQS, seen: SeenIndices): FreeQS =
+    branch.resume match {
+      case -\/(qs) =>
+        Inject[QScriptCore, QScriptTotal].prj(qs) match {
+          case Some(LeftShift(src, struct, id, repair)) =>
+            repair.resume match {
+              case -\/(array @ ConcatArrays(_, _)) =>
+                Free.roll(Inject[QScriptCore, QScriptTotal].inj(
+                  LeftShift(src, struct, id, rewriteRepair(array, seen))))
+              case _ => branch
+            }
+          case _ => branch
+        }
+      case _ => branch
+    }
+
   // TODO: Can we be more efficient? - can get rid of `.sorted`, but might be
   //       non-deterministic, then.
   val indexMapping: SeenIndices => IndexMapping =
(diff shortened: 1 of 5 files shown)


connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -606,78 +537,280 @@ class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScr
...
-    "rewrite left shift with array referenced through struct" in {
-      val initialSrc: Fix[QST] =
-        QCT.inj(LeftShift(
+    "not rewrite equi join with entire array branch referenced in key" in {
+      val initial: Fix[QST] =
+        EJT.inj(EquiJoin(
+          UnreferencedRT.embed,
+	  arrayBranch3,
+          HoleQS,
+	  HoleF,  // reference entire left branch
```
alissapajer: There are some formatting issues in this file, vim randomly decided to use tabs everywhere, I'll push a fix.

connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -606,78 +537,280 @@ class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScr
...
+          MakeMapR(
+	    StrLit("xyz"),
+	    LeftSideF))).embed  // reference entire left branch
+
+      initial.pruneArrays must equal(initial)
+    }
+
+    // we'd like to be able to rewrite this
+    // but it might require more work in addition to array pruning
+    "not rewrite equi join with array branch referenced outside of join" in {
```
alissapajer: This is an example of future work that I will probably do soon, since in practice we are ending up with `UnshiftMap` inside `LeftShift` in join branches that could be elided.

connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -606,78 +537,280 @@ class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScr
...
+            ConcatArraysR(MakeArrayR(LeftSideF), MakeArrayR(RightSideF)))).embed,
+	  ProjectIndexR(
+	    ProjectIndexR(HoleF, IntLit[Fix, Hole](0)),
+	    IntLit[Fix, Hole](2)))).embed
+
+      initial.pruneArrays must equal(initial)
+    }
+
+    // this can be rewritten - we just don't support that yet
+    "not rewrite theta join with filtered left shift as branch" in {
```
alissapajer: This is a second example of an issue that needs to be addressed soon.

alissapajer:
Quick status update on this: the CB build is failing for not-yet-understood reasons (have seen same behavior in another PR build too). Discussed with @drostron and we're going to collect some more data and see what other PR builds do.

alissapajer:
Cleared PR build cache (so build used master cache) and build has fixed itself. Unclear what the problem was exactly, but it appears to have been using a corrupt cache.

alissapajer:
@sellout could I get a review on this when you can?

connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -554,20 +507,9 @@ class QScriptPruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScr
 
     // this can be rewritten - we just don't support that yet
```
sellout: Can these comments all be prefixed with “FIXME:”, since they’re actually testing the reverse of the desired behavior (I understand why this is better than just writing the result we want and adding `pendingUntilFixed`, though).

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -32,21 +32,47 @@ import simulacrum.typeclass
 object PATypes {
   type SeenIndices = Set[BigInt]
   type KnownIndices = Option[SeenIndices]
+  type Indices = ScalaMap[Hole \/ JoinSide, KnownIndices]
```
sellout: Having `Hole \/ JoinSide` here makes me nervous. Do I understand correctly that we only ever have _one_ of the two types in the map at a time, but since we have to expose them in the algebras, we need to unify the types? I wonder if there is some trick to get back to `ScalaMap[A, KnownIndices]` … but nothing comes to mind.

connector/src/test/scala/quasar/qscript/PruneArrays.scala
alissapajer: I think FIXME is a good idea - makes it more clear that we need to support this rewrite. :+1:

connector/src/main/scala/quasar/qscript/PruneArrays.scala
alissapajer: So, consider a `ThetaJoin`. The `Hole`s in the left and right branches (`FreeQS`) are referencing the `src` of the `ThetaJoin`. It's possible that this `src` is say a `LeftShift` with an array as its `repair`. Since the rest of the `ThetaJoin` (join function and combine function) are referencing the branches, then if there were unused array elements in the `src` (unused by either branch) then in some cases we could rewrite that array and those branches with array pruning. (e.g. If the branch is filtering the `src` we couldn't rewrite (at least not without inspecting the subsequent join funcs) but if the branch is mapping the `src` then we could.)

This rewrite isn't supported yet (there are tests for it though). We don't hit this case commonly, as usually the `src` is `Unreferenced` because we are joining two distinct data sets. But, for example, in a self join I think we would hit this rewrite case.

So long story short, when we support this `src` rewriting, we will have both `Hole` and `JoinSide` keys in the `Indices` here for branched cases like `ThetaJoin` and `EquiJoin`, so I believe that the disjunction is the proper encoding.

PR #2027, opened 2017-03-30
Fix way master and port are fetched from URI

rabbitonweb:
There was a bug in how sparkUri was created. This PR fixes this bug.

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
@@ -35,6 +35,7 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem => HdfsFileSystem}
 import org.apache.spark._
 import pathy.Path._
+import pathy.Path.posixCodec
 import scalaz._, Scalaz._
 import scalaz.concurrent.Task
 
@@ -57,7 +58,8 @@ package object hdfs {
 
     def liftErr(msg: String): DefinitionError = NonEmptyList(msg).left[EnvironmentError]
 
-    def master(uri: Uri): State[SparkConf, Unit] = State.modify(_.setMaster(s"spark://${uri.host}:${uri.port}"))
+    def master(host: String, port: Int): State[SparkConf, Unit] =
+      State.modify(_.setMaster(s"spark://$host:$port"))
 
     def appName: State[SparkConf, Unit] = State.modify(_.setAppName("quasar"))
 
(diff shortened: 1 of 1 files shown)


sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -66,8 +68,12 @@ package object hdfs {
 
     val uriOrErr: DefinitionError \/ Uri = Uri.fromString(connUri.value).leftMap((pf: ParseFailure) => liftErr(pf.toString))
 
-    val sparkConfOrErr: DefinitionError \/ SparkConf = uriOrErr.map(uri => 
-      (master(uri) *> appName *>
+    val sparkConfOrErr: DefinitionError \/ SparkConf = for {
+      uri <- uriOrErr
+      host <- uri.host.fold(NonEmptyList("host not provided").left[EnvironmentError].left[Uri.Host])(_.right[DefinitionError])
+      port <- uri.port.fold(NonEmptyList("port not provided").left[EnvironmentError].left[Int])(_.right[DefinitionError])
+    } yield {
```
djspiewak: The above is rather hard to follow.  Are there nested eithers here?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -57,7 +58,8 @@ package object hdfs {
 
     def liftErr(msg: String): DefinitionError = NonEmptyList(msg).left[EnvironmentError]
 
-    def master(uri: Uri): State[SparkConf, Unit] = State.modify(_.setMaster(s"spark://${uri.host}:${uri.port}"))
+    def master(host: String, port: Int): State[SparkConf, Unit] =
```
djspiewak: I'm not clear on exactly what the bug was here?

rabbitonweb: Bug was that `Uri.host` returns `Option[Uri.Host]` and `Uri.port` returns `Option[Int]`

I thought that the returned `String`s

Method `master` was expecting two `String`s, Option was implicitly casted to `String` (because Java ¯\_(ツ)_/¯)

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
rabbitonweb: As above, `uri.host` & `uri.port` return `Option` thus I'm folding them to `DefinitionError` if `None`

rabbitonweb: `type DefinitionError  = NonEmptyList[String] \/ EnvironmentError`

djspiewak: Ah, I see.  Well, just as a mental TODO, I think it would probably be better if `DefinitionError` were a proper ADT, rather than encoding it via a coproduct.  Generally better API and error messages that way.  Doesn't hold up the PR though.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
djspiewak: In general, I don't like replacing semantic types (e.g. `Uri`) with non-semantic, non-newtyped "equivalents".  In this case though, the difference is pretty minor, and `Uri` was only being used to carry around those two values.  👍 from me.

Issue #1907, opened 2017-02-13
Test for Spark robustness


rabbitonweb:
- [x] Enable all integration tests for Spark that are not effected by QScript issues
- [x] Upgrade to latest version of Spark
- [x] set-up 2.1.0 cluster for tests
- [x] spark-connectors should fail gracefully when mounted more then one
- [x] unmounting spark connector should free resources (there is a bug most probably in front-end)

rabbitonweb:
Upgrade in https://github.com/quasar-analytics/quasar/pull/1908

rabbitonweb:
**Should be fixed in SparkCore:**
`typeOf.test` - `MapFunc` `TypeOf` must be implemented in `CoreMap`
`multitypeFlatten.test`, `likeWithConstantFoldablePattern.test`, `different-flattens.test` - gives wrong results, problems with LIKE, must investigate

**QScript bugs:**
`statesByShortestFirstCity.test`, `singleFieldOrderedDistinct.test`, `LocationCount.test` - issue is that any `select foo from bar group by foo order by baz` will create QScript like `Reduce(Sort(Read(..)))` but Spark will loose order after reduce, we need `Sort(Reduce(Sort(Read(..))))`. This is fixed on some mysterious @sellout branch but pushing this to master requires removal of old mongo connector - so we are waiting patiently for that to happen :)
`tripleFlatten.test` - known bug #732
`nonJsJoinCondition.test` - self JOINs
`servlet.test` - pending #465, #467
`skipCount.test` - unknown
`multitypeFlatten.test` - unknown

alissapajer:
@rabbitonweb can you make individual tickets for the QScript bugs that don't have tickets already?

rabbitonweb:
@alissapajer I will double check if tickets exits

rabbitonweb:
FINISHED

Issue #1712, opened 2016-11-21
Modify installers project to install Quasar Spark JAR


jdegoes:
TODO:

- [x] Modify script to generate sparkcore.jar
- [x] Modify building scripts to include sparkcore.jar
- [x]  Either set instructions to include QUASAR_HOME or set `addJar` to place where sparkcore.jar is (some vodoo magic needed)

jdegoes:
@rabbitonweb Don't forget about this one!

rabbitonweb:
@jdegoes In order for `spark-hdfs` to work we need to include `sparkcore.jar` in the installers. That jar is very light (few kilobytes).
Do we want a separate installers for Quasar and Quasar that works with `spark-hdfs` OR do we want to always include `sparkcore.jar` in the installers (regardless of whether somebody is downloading Quasar to use with Spark or not)

Please note that `spark-local` (connector that will be used for https://github.com/quasar-analytics/quasar/issues/1899) does NOT require `sparkcore.jar`)

jdegoes:
Always include it in the installers.

rabbitonweb:
TODO:

- [ ] Modify script to generate sparkcore.jar
- [ ] Modify building scripts to include sparkcore.jar
- [ ]  Either set instructions to include QUASAR_HOME or set `addJar` to place where sparkcore.jar is (some vodoo magic needed)

jdegoes:
@zeeiyerWork This too is a priority in order for people to use Spark without manual (and currently undocumented) configuration.

rabbitonweb:
@jdegoes the configuration is documented but only for ppl building their own version of quasar (or did you mean something else?)

zeeiyerWork:
@rabbitonweb, @damonLL, @veryfatboy -> I believe we are talking about documentation for end-users , we are getting reports from the field about problems.

zeeiyerWork:
@rabbitonweb -> This would imply that you get familiar with our installers.

rabbitonweb:
@zeeiyerWork yes, this requires undestanding how installers works

rabbitonweb:
CLOSED

PR #2029, opened 2017-03-31
Remove need for QUASAR_HOME

rabbitonweb:
Part of https://github.com/quasar-analytics/quasar/issues/1712 & https://github.com/quasar-analytics/quasar/issues/1907

This change brings two unrelated functionalities
1. Remove need for QUASAR_HOME
2. Handle more gracefully errors when user tries to connect second spark based connector

Ad. 1 Remove need for QUASAR_HOME
Every spark based connector need to know where `sparkcore.jar` is placed. Previously that information was provided with `QUASAR_HOME` env. variable. Now the need for `QUASAR_HOME` is completely removed. ill Path of quasar-web.jar as a root to find sparkcore.jar

Ad.2 Handle errors

When running Spark only one `SparkContext` per JVM is possible. Thus if we try to connect a second spark-based connector, we will get an error. It caused not human-friendly error message on UI. Now we fail with more detailed message, hiding the details.


rabbitonweb:
@drostron after recent commit generation of `SparkContext` is unified among all spark based connectors

drostron:
Also, there's still a reference to `QUASAR_HOME` in `README.md`.

drostron:
@rabbitonweb Needs a merge from master

drostron:
@rabbitonweb Needs a semantic version label

PR #2030, opened 2017-03-31
Remove unnecessary index creation to avoid CB timeout on Travis

drostron:


nicflores:
Are there any tests the rely on this index? Otherwise this looks good to me.

drostron:
`leftKeyJoin.test` and `rightKeyJoin.test` rely on the beer-sample bucket, but do not require the index and display reasonable performance without it.

nicflores:
sounds good dave.

PR #2033, opened 2017-04-04
improved QScript merging through Filter and Sort

### LINKED ISSUE #1915: QScript generates unnecessary ThetaJoin in `caseVariations.test`
In the `caseVariations.test` integration test, the resulting `QScript` contains a `ThetaJoin` when it should not.

The query:
```
💪 $ explain select distinct case pop when 0 then "nobody" when 1 then "one" when 2 then "a couple" when 3 then "a few" else "more" end as cardinal, case pop when 1 then 0 when 10 then 1 end as power, case when pop % 2 = 0 then "even" when pop = 1 or pop = 9 then "odd" else "prime" end as parity, case when pop > 5 then pop - 5 end as grade from zips where pop <= 10 order by pop
```
This query can be simplified by removing nearly all (or perhaps all) the `case` statements, and an unnecessary join will still be produced in `QScript`.

### LINKED ISSUE #1924: QScript fails to merge through a mapped sort
The QScript for the query `select distinct city from zips where pop <= 10 order by pop` contains an unnecessary `ThetaJoin`. (It should contain no `ThetaJoin`s at all.)

The cause of the problem is clear if we look at a subset of the `LogicalPlan`.
```
      ╰─ Let('__tmp5)
         ├─ Sort
         │  ├─ Free('__tmp4)
         │  ├─ ObjectProject('__tmp4{"__sd__0"})
         │  ╰─ LogicalPlan(Ascending)
         ╰─ Invoke(Take)
            ├─ Invoke(DeleteField)
            │  ├─ Invoke(DistinctBy)
            │  │  ├─ Free('__tmp5)
            │  │  ╰─ Invoke(DeleteField)
            │  │     ├─ Free('__tmp5)
            │  │     ╰─ Str("__sd__0")
            │  ╰─ Str("__sd__0")
            ╰─ Constant(Int(11))
```
We attempt to merge the `QScript` forms of `__tmp5` and `DeleteField(__tmp5, "__sd__0")`. In `QScript` this is merging a `Sort(Map(Hole, _), _, _)` and a `Map(Sort(Hole, _, _), _)`. This merge currently fails in `QScript`, returning a `ThetaJoin`.

One solution to this problem is to add an additional parameter to `Sort` (and `Filter`) which is a a `MapFunc` operation that we perform after the `Sort` (or `Filter`). This would allow us to fold a `Map(Sort, _)` into just a `Sort` node, thus allowing the sources in the example to cleanly merge. This partially addresses #1545 as well.

Another idea is to interleave identity `Map` nodes before performing source merging in `QScript` to allow cases like this to merge. I don't think we need to do this yet, and shouldn't without examples showing why we need it, but I just wanted to mention it.

alissapajer:
Fixes #1915 
Fixes #1924

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Merge.scala
+++ b/connector/src/main/scala/quasar/qscript/Merge.scala
@@ -45,17 +45,11 @@ class Merge[T[_[_]]: BirecursiveT: OrderT: EqualT: ShowT] extends TTypes[T] {
   private def linearize[G[_]: Traverse]: Algebra[G, List[G[EM]]] =
     fl => fl.as[EM](Extern) :: fl.fold
 
-  private def delinearizeInner[G[_]: Traverse, A](
-    implicit DE: Const[DeadEnd, ?] :<: G):
+  private def delinearizeInner[G[_]: Functor, H[_], A]
+    (HtoG: H ~> G)
+    (implicit DE: Const[DeadEnd, ?] :<: H):
       Coalgebra[G, List[G[A]]] = {
-    case Nil    => DE.inj(Const[DeadEnd, List[G[A]]](Root))
-    case h :: t => h.as(t)
-  }
-
-  private def delinearizeInnerCoEnv[A](
-    implicit DE: Const[DeadEnd, ?] :<: QScriptTotal):
-      Coalgebra[CoEnv[Hole, QScriptTotal, ?], List[CoEnv[Hole, QScriptTotal, A]]] = {
-    case Nil    => CoEnv(\/-(DE.inj(Const[DeadEnd, List[CoEnv[Hole, QScriptTotal, A]]](Root))))
+    case Nil    => HtoG(DE.inj(Const[DeadEnd, List[G[A]]](Root)))
     case h :: t => h.as(t)
   }
 
@@ -65,6 +59,7 @@ class Merge[T[_[_]]: BirecursiveT: OrderT: EqualT: ShowT] extends TTypes[T] {
     case h :: t => h.as(t).right
   }
 
+  // FIXME unify with more general target delinearization
   private def delinearizeTargetsCoEnv[A]:
       Coalgebra[CoEnv[Hole, QScriptTotal, ?], List[CoEnv[Hole, QScriptTotal, A]]] = {
     case Nil    => CoEnv(-\/(SrcHole))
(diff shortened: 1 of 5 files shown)


connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -864,6 +865,96 @@ class QScriptSpec
 
     }
 
+    "convert an ordered filtered distinct" in {
```
alissapajer: Previously these two tests produced `ThetaJoin` in the `QScript`. So even though there are the FIXMEs here for fixing `Sort`, it is important to have these tests.

I'll make a ticket for the `Sort` issue and then update the FIXME comments here with the issue number.

connector/src/main/scala/quasar/qscript/Merge.scala
```diff
@@ -94,53 +89,75 @@ class Merge[T[_[_]]: BirecursiveT: OrderT: EqualT: ShowT] extends TTypes[T] {
...
+      ZipperSides(HoleF[T], HoleF[T]),
+      ZipperTails(makeList[S, G, H](left), makeList[S, G, H](right))))(consZipped[G], zipper[G])
+
+  def mergePair[S[_[_]], G[_]: Traverse, H[_]]
+    (left: S[H], right: S[H])
+    (HtoG: H ~> G, unify: List[G[EM]] => FreeQS)
+    (implicit
+      mergeable: Mergeable.Aux[T, G],
+      C: Corecursive.Aux[S[H], G],
+      R: Recursive.Aux[S[H], G],
```
alissapajer: Using `Corecursive` and `Recursive` until `Birecursive` is available. Could add a comment here referencing https://github.com/slamdata/matryoshka/issues/44.

connector/src/main/scala/quasar/qscript/Merge.scala
```diff
@@ -45,17 +45,11 @@ class Merge[T[_[_]]: BirecursiveT: OrderT: EqualT: ShowT] extends TTypes[T] {
   private def linearize[G[_]: Traverse]: Algebra[G, List[G[EM]]] =
     fl => fl.as[EM](Extern) :: fl.fold
 
-  private def delinearizeInner[G[_]: Traverse, A](
-    implicit DE: Const[DeadEnd, ?] :<: G):
+  private def delinearizeInner[G[_]: Functor, H[_], A]
+    (HtoG: H ~> G)
```
sellout: 👍🏾

connector/src/main/scala/quasar/qscript/Merge.scala
sellout: I don’t think it’s necessary – this is something that is _everywhere_ in Matryoshka and Quasar code. We’ll fix it globally once we manage to define `Birecursive`.

connector/src/main/scala/quasar/qscript/Merge.scala
alissapajer: Yeah, this is what was needed to generalize the `CoEnv`.

PR #2039, opened 2017-04-08
Generalize `CoEnv` usage

alissapajer:
Generalizes usage of `CoEnv`.


foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -276,6 +276,20 @@ package object fp
...
     λ[CoEnv[A, F, ?] ~> λ[α => Option[F[α]]]](_.run.toOption),
     λ[F ~> CoEnv[A, F, ?]](fb => CoEnv(fb.right[A])))
+
+  def coenvBijection[T[_[_]]: BirecursiveT, F[_]: Functor, A]:
+      Bijection[Free[F, A], T[CoEnv[A, F, ?]]] =
+    bijection[Id, Id, Free[F, A], T[CoEnv[A, F, ?]]](
+      _.convertTo[T[CoEnv[A, F, ?]]],
+      _.convertTo[Free[F, A]])
+
+  def applyBijection[A, B](bij: Bijection[A, B])(modify: B => B): A => A =
```
alissapajer: I am open to alternate names for this function; the current name isn't very specific.

djspiewak: It really surprises me that this function doesn't exist already.  `liftApplyRight` seems like an appropriate name if it were defined on `Bijection`.  As a standalone… not sure.  Something with `Right` (or `R`) is probably best though, since there are two directions you could go with it.

alissapajer: Yeah it's not defined in `BijectionT` in scalaz. 

I agree that it's a good idea to differentiate that we're running `A => B => B => A` rather than `B => A => A => B`.  scalaz names the functions of the bijection `to: A => B` and `from: B => A`, so I'm thinking we could name this `applyFrom`. Though should we be naming it based on the domain or codomain of the bijection functions? I think domain is reasonable, but not sure if there is a standard :)

Also since this is defined here with `Id` everywhere we can omit the `lift` in the name.

djspiewak: `applyFrom` seems reasonable.  I wouldn't use domain or codomain, since it's really easy to get confused as to which is which on `Bijection`.

alissapajer: By domain and codomain I was referring to the fact that we are `modify`ing `B` and `B` is the codomain of `to` and the domain of `from` and so by naming this function `applyFrom` we are referencing the domain of `from` whereas if we named it `applyTo` we'd be referencing the codomain of `to`.

djspiewak: Right, but I mean I get confused with domain and codomain on a bijection, whereas "from" and "to" have a directionality in the type parameters.

alissapajer: I'm not talking about putting "domain" or "codomain" in the name. I'm bringing up the idea that we could either name this `applyTo` or `applyFrom`. How do we decide which to name it? I was using the idea of domain and codomain to attempt to make a decision between those two choices of names.

alissapajer: I agree that domain and codomain would be confusing in the name because `A` and `B` each appear in both positions.

djspiewak: Oh, ok I was confused.  :-)  I think any name which references the directionality of the type parameters in `Bijection` should be ok.  Alternatively, something that remains uniform with other functions which have to make a similar choice.  So `applyFrom`, `applyRight`, etc.

alissapajer: :+1: I think we're on the same page.

PR #2043, opened 2017-04-12
Add new "invoke" endpoint

jedesah:
New endpoint to invoke functions inside of a module.


web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -288,6 +288,8 @@ class DataServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s {
         response.headers.get(`Content-Disposition`) must_== Some(disposition)
       }.set(minTestsOk = 10).flakyTest("scalacheck: Gave up after only 2 passed tests. 12 tests were discarded.")  // NB: this test is slow because NonEmptyDir instances are still relatively large
       "what happens if user specifies a Path that is a directory but without the appropriate headers?" >> todo
+      "description of the function if the file is a module function" in todo
```
wemrysi: Are these reminders? Would it make sense to implement them now?

core/src/main/scala/quasar/fs/mount/module/Module.scala
```diff
@@ -0,0 +1,172 @@
...
+          s"Encountered the following semantic errors while attempting to invoke function: ${e.shows}"
+        case ArgumentsMissing(missing) =>
+          s"The following arguments are missing: $missing"
+        case ParsingErr(e) => e.shows
+      }
+  }
+
+
+  final case class InvokeModuleFunction(path: AFile, args: Map[String, String])
+    extends Module[Error \/ QueryFile.ResultHandle]
```
wemrysi: This should probably be its own handle type, reusing the one from `QueryFile` is leaking implementation details.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
jedesah: Ya, they are reminders of what would be desirable behavior on the data endpoint side of things but I wanted to get this PR up before they get implemented so that @beckyconning can start using this new API to implement "parameterized views" sooner rather than later

core/src/main/scala/quasar/fs/mount/module/Module.scala
jedesah: I don't see it that way. Invoking a function in a module is analogous to executing a query so it doesn't seem surprising to receive a `QueryFile.ResultHandle`.

I would be inclined to extract `ResultHandle` from `QueryFile` though and let it stand on its own to be used by both if you think that is a good idea

jedesah:
@wemrysi We did talk about this but no worries, happy to go over it again to make sure I am not making any false assumptions! I would have gladly made it just a function and saved myself the trouble of creating a new algebra and corresponding interpreter :smile:. The reason I made it an algebra though is for authorization in quasar-advanced. The only way to get the intended authorization behavior was to either create this new algebra or modify the `ReadFile` algebra and you voted for new algebra when we discussed it :smile:

core/src/main/scala/quasar/fs/mount/module/Module.scala
wemrysi: Isn't `InvokeModuleFunction` supposed to be an abstraction around the notion of invocation? If so it shouldn't assume a particular implementation.

Ideally, the result handles would be existential, as clients don't need to know what they are, just that they can be obtained and then used to request data. In lieu of this, having separate types helps to convey they aren't interchangeable.

wemrysi:
Ah, ok thanks for the refresher, I think the algebra is still the way to go then.

core/src/main/scala/quasar/fs/mount/module/Module.scala
jedesah: Okay, I'm convinced

PR #2046, opened 2017-04-12
push filter before union in qscript

### LINKED ISSUE #1381: optimization: push other operations before Union
This is similar to how we push single-side filters ahead of joins. We simply apply the same operation to both sides of the Union.

alissapajer:
Closes #1381.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Optimize.scala
+++ b/connector/src/main/scala/quasar/qscript/Optimize.scala
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript
+
+import slamdata.Predef._
+import quasar.fp._
+
+import matryoshka._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+class Optimize[T[_[_]]: BirecursiveT] extends TTypes[T] {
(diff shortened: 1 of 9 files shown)


connector/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -0,0 +1,68 @@
...
+        case Map(fromInner, mf) =>
+          Map(FtoG(QC.inj(Subset(src, fromInner, sel, count))).embed, mf).some
+        case _ => None
+      }
+    case _ => None
+  }
+
+  private def filterBeforeUnion[F[_]: Functor](implicit QC: QScriptCore :<: F)
+      : QScriptCore[T[F]] => Option[QScriptCore[T[F]]] = {
+    case Filter(Embed(src), fm) =>
```
alissapajer: @sellout `Filter` is the only operation that seemed reasonable to push before `Union`, at least without knowing any more information about the underlying data/backend. Was there anything else you had in mind to close out #1381?

sellout: I think that’s probably right. If anything else comes up, we can address it then.

connector/src/main/scala/quasar/qscript/Optimize.scala
```diff
@@ -0,0 +1,68 @@
...
+  /** Should only be applied after all other QScript transformations. This gives
+    * the final, optimized QScript for conversion.
+    */
+  def optimize[F[_], G[_]: Functor](FtoG: F ~> G)(
+    implicit
+      QCF: QScriptCore :<: F,
+      QCG: QScriptCore :<: G)
+      : F[T[G]] => F[T[G]] =
+    liftFF[QScriptCore, F, T[G]](repeatedly(subsetBeforeMap[F, G](FtoG))) >>>
+      liftFF[QScriptCore, F, T[G]](repeatedly(filterBeforeUnion[G]))
```
sellout: Here we should probably do what we do with `coalesce` – `repeatedly(applyTransforms(subsetBeforeMap, filterBeforeUnion))`.

sellout: I’m not sure if the difference is testable with the current set of optimizations … you’d basically need a case where `subsetBeforeMap` could only be applied after `filterBeforeUnion` was applied. But future optimizations may work better with the `applyTransforms` versions.

alissapajer:
@sellout ready for another look.

PR #2049, opened 2017-04-15
Upgrade sbt

jedesah:


drostron:
The "Maven version range improvement" is interesting, especially in light of the version range issue we ran into the other day with QAdv.

PR #2050, opened 2017-04-15
Improve error messages from `compile` and `query` api endpoints when variables are missing

jedesah:


web/src/test/scala/quasar/api/services/query/QueryServiceSpec.scala
```diff
@@ -79,6 +81,57 @@ class QueryServiceSpec extends quasar.Qspec with FileSystemFixture {
...
+                        "detail" -> Json(
+                          "message" := "There is no binding for the variable :baz",
+                          "varName" := "baz"
+                        )
+                      )
+                    )
+                  )
+                )
+              )
+            )
```
alissapajer: What would you think of moving these parentheses all up to the end of line 124? Since we're constructing various `Json` and not creating a context that needs to be opened and then closed (as if often the case with `{ ... }`) then I'd prefer all the parentheses on the same line.

frontend/src/main/scala/quasar/variables.scala
```diff
@@ -67,10 +68,23 @@ object Variables {
     case x => x.embed.right
   }
 
+  def allVariablesƒ: Algebra[Sql, List[VarName]] = {
```
sellout: This should just be called `allVariables`. The `ƒ` suffix is a remnant from before we had nice types describing algebras, and should generally go away.

frontend/src/main/scala/quasar/variables.scala
```diff
@@ -67,10 +68,23 @@ object Variables {
...
+    case Vari(name)                             => List(VarName(name))
+    case sel @ Select(_, _, rel, _, _, _) =>
+      rel.toList.collect { case VariRelationAST(vari, _) => VarName(vari.symbol)} ++
+      (sel: Sql[List[VarName]]).toList.join
+    case other                                  => other.toList.join
```
sellout: It looks like the `Vari` and `other` cases need to have their `=>` realigned.

frontend/src/main/scala/quasar/variables.scala
```diff
@@ -67,10 +68,23 @@ object Variables {
     case x => x.embed.right
   }
 
+  def allVariablesƒ: Algebra[Sql, List[VarName]] = {
+    case Vari(name)                             => List(VarName(name))
+    case sel @ Select(_, _, rel, _, _, _) =>
+      rel.toList.collect { case VariRelationAST(vari, _) => VarName(vari.symbol)} ++
+      (sel: Sql[List[VarName]]).toList.join
```
sellout: `.toList.join` can be replaced with `.fold`, no? (Although, IIRC, there is some hoop necessary to make sure you get the one from `Foldable` and not the one from the stdlib.)

sellout: Oh, and needs a space before the closing `}`.

frontend/src/main/scala/quasar/variables.scala
```diff
@@ -67,10 +68,23 @@ object Variables {
...
+  }
+
   // FIXME: Get rid of this
   def substVars(expr: Fix[Sql], variables: Variables)
-      : SemanticError \/ Fix[Sql] =
-    expr.cataM[SemanticError \/ ?, Fix[Sql]](substVarsƒ(variables))
+      : SemanticErrors \/ Fix[Sql] = {
+    val allVars = expr.cata(allVariablesƒ)
+    val errors = allVars.map(variables.lookup(_)).collect { case -\/(semErr) => semErr }.toNel
+    errors.cata(
```
sellout: Can you use `.fold` here instead of `.cata`? I just find `cata` confusing because of Matryoshka, no technical reason for the change.

web/src/test/scala/quasar/api/services/query/QueryServiceSpec.scala
sellout: Yes please – no dangling parens.

frontend/src/main/scala/quasar/variables.scala
jedesah: I just hate those damn parentheses!

Issue #1549, opened 2016-10-10
Implement common subexpression elimination


sellout:
This is an optimization that would pull out more common substructures to eliminate duplicated work.

This would be handled differently depending on exactly where the common expressions are found, and the abilities of the connector. E.g., if a connector supports `Let`, then we could use them within a FreeMap. Otherwise, we could de-normalize a `qscript.Map` into two Maps, one setting up values to be used by the second. And there may be other representations for other dimensional situations.

alissapajer:
@djspiewak see this old ticket for tracking `Let` bindings in `FreeMap`, or whatever solution we decide on.

PR #2052, opened 2017-04-17
LeftShift cancels out static array.

### LINKED ISSUE #1382: normalization: LeftShift cancels out statically-known structure
`LeftShift(Map(x, MakeArray(y))) ≟ Map(x, y)`

alissapajer:
Closes #1382.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Rewrite.scala
+++ b/connector/src/main/scala/quasar/qscript/Rewrite.scala
@@ -278,6 +278,33 @@ class Rewrite[T[_[_]]: BirecursiveT: OrderT: EqualT] extends TTypes[T] {
     case _                                 => None
   }
 
+  def compactLeftShift[F[_]: Functor, G[_]: Functor]
+    (FToG: PrismNT[G, F])
+    (implicit QC: QScriptCore :<: F)
+      : QScriptCore[T[G]] => Option[F[T[G]]] = {
+    case qs @ LeftShift(Embed(src), struct, ExcludeId, joinFunc) =>
+      (FToG.get(src) >>= QC.prj, struct.resume) match {
+        // LeftShift(Map(_, MakeArray(_)), Hole, ExcludeId, _)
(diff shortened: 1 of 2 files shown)


connector/src/test/scala/quasar/qscript/Rewrite.scala
```diff
@@ -330,5 +334,51 @@ class QScriptRewriteSpec extends quasar.Qspec with CompilerHelpers with QScriptH
...
+          ConcatMapsR(
+            MakeMapR(StrLit("right"), RightSideF),
+            MakeMapR(StrLit("left"), LeftSideF)))).embed
+
+      val expected: Fix[QS] =
+        QC.inj(Map(
+          RootR.embed,
+          ConcatMapsR(
+            MakeMapR(StrLit("right"), AddR(HoleF, IntLit(3))),
+            MakeMapR(StrLit("left"), MakeArrayR(AddR(HoleF, IntLit(3))))))).embed
```
alissapajer: We end up with `AddR(HoleF, IntLit(3))` duplicated. We could pull that out into a second `Map` but then `Coalesce` would just combine the two `Map`s and we'd be back to where we started. To avoid this we need common subexpression elimination #1549.

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -278,6 +278,33 @@ class Rewrite[T[_[_]]: BirecursiveT: OrderT: EqualT] extends TTypes[T] {
...
+        case (Some(Map(innerSrc, fm)), \/-(SrcHole)) =>
+          fm.resume match {
+            case -\/(MakeArray(value)) =>
+              QC.inj(Map(innerSrc, joinFunc >>= {
+                case LeftSide => fm
+                case RightSide => value
+              })).some
+            case _ => None
+          }
+        // LeftShift(_, MakeArray(_), ExcludeId, _)
```
sellout: Nice catch on this other case.

PR #2053, opened 2017-04-17
Update Quasar readme with integration testing instructions

nicflores:
Quasar README update with instructions on creating containerized backend datastores using the scripts located in docker/scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,41 +46,62 @@ To compile the project and run tests, first clone the quasar repo and then execu
 Note: please note that we are not using here a system wide sbt, but our own copy of it (under ./sbt). This is primarily
  done for determinism. In order to have a reproducible build, the helper script needs to be part of the repo.
 
-Running the full test suite can be done in two ways:
+Running the full test suite can be done using docker containers for various backends:
 
-##### Testing option 1 (prerequisite: docker)
+##### Full Testing (prerequisite: docker and docker-compose)
 
-A docker container running mongodb operates for the duration of the test run.
+In order to run integration test for various backends the `docker/scripts` are provided to easily create dockerized backend data stores.
 
-```bash
-./bin/full-it-tests.sh
-```
+Of particular interest are the following two scripts:
 
-##### Testing option 2
+  1. `docker/scripts/setupContainers`
+  2. `docker/scripts/assembleTestingConf`
 
-In order to run the integration tests for a given backend, you will need to provide a URL to it. For instance, in the case of MongoDB, If you have a hosted MongoDB instance handy, then you can simply point to it, or else
-you probably want to install MongoDB locally and point Quasar to that one. Installing MongoDB locally is probably a good idea as it will
-allow you to run the integration tests offline as well as make the tests run as fast as possible.
 
-In order to install MongoDB locally you can either use something like Homebrew (on OS X) or simply go to the MongoDB website and follow the
-instructions that can be found there.
-
-Once we have a MongoDB instance handy, we need to configure the integration tests
-in order to inform Quasar about where to find the backends to test.
-
-Simply copy `it/testing.conf.example` to `it/testing.conf` and change the values. For example:
+Supported backends datastores are:
 
 ```
-mongodb_3_2="mongodb://<mongoURL>"
-mongodb_3_0="mongodb://<mongoURL>"
-mongodb_2_6="mongodb://<mongoURL>"
+quasar_mongodb_2_6
+quasar_mongodb_3_0
+quasar_mongodb_read_only
+quasar_mongodb_3_2
+quasar_mongodb_3_4
+quasar_metastore
+quasar_postgresql
+quasar_marklogic_xml
+quasar_marklogic_json
+quasar_couchbase
 ```
 
-where <mongoURL> is the url at which one can find a Mongo database. For example <mongoURL> would probably look
-something like `localhost:27017` for a local installation. This means the integration tests will be run against
-both MongoDB versions 2.6, 3.0, and 3.2. Alternatively, you can choose to install only one of these and run the integration
-tests against only that one database. Simply omit a version in order to avoid testing against it. On the integration
-server, the tests are run against all supported filesystems.
+Knowing which backend datastores are supported you can create and configure docker containers using `setupContainers`. For example the command:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,41 +46,62 @@ To compile the project and run tests, first clone the quasar repo and then execu
...
  done for determinism. In order to have a reproducible build, the helper script needs to be part of the repo.
 
-Running the full test suite can be done in two ways:
+Running the full test suite can be done using docker containers for various backends:
 
-##### Testing option 1 (prerequisite: docker)
+##### Full Testing (prerequisite: docker and docker-compose)
 
-A docker container running mongodb operates for the duration of the test run.
+In order to run integration test for various backends the `docker/scripts` are provided to easily create dockerized backend data stores.
```
drostron: Could also mention the MetaStore.

README.md
```diff
@@ -46,41 +46,62 @@ To compile the project and run tests, first clone the quasar repo and then execu
...
 ```
 
-##### Testing option 2
+Knowing which backend datastores are supported you can create and configure docker containers using `setupContainers`. For example the command:
 
-In order to run the integration tests for a given backend, you will need to provide a URL to it. For instance, in the case of MongoDB, If you have a hosted MongoDB instance handy, then you can simply point to it, or else
-you probably want to install MongoDB locally and point Quasar to that one. Installing MongoDB locally is probably a good idea as it will
-allow you to run the integration tests offline as well as make the tests run as fast as possible.
+```
+./setupContainers -u quasar_mongodb_3_0,quasar_postgresql,quasar_marklogic_xml,quasar_couchbase
```
drostron: We might want to add quasar_metastore here as well since there's a MetaStore related integration test.

README.md
```diff
@@ -46,41 +46,62 @@ To compile the project and run tests, first clone the quasar repo and then execu
...
+postgresql="jdbc:postgresql://192.168.99.101:5433/quasar-test?user=postgres&password=postgres"
+mongodb_3_0="mongodb://192.168.99.101:27019"
 ```
 
-where <mongoURL> is the url at which one can find a Mongo database. For example <mongoURL> would probably look
-something like `localhost:27017` for a local installation. This means the integration tests will be run against
-both MongoDB versions 2.6, 3.0, and 3.2. Alternatively, you can choose to install only one of these and run the integration
-tests against only that one database. Simply omit a version in order to avoid testing against it. On the integration
-server, the tests are run against all supported filesystems.
+IP's will vary depending on your docker enviornment. In addition the scripts assume you have docker and docker-compose installed.
```
drostron: s/enviornment/environment/

drostron:
Oh, an esthetic nit — we've typically been adding `[skip ci]` at the end of the commit message.

jedesah:
Minor comment. We tend to write commit messages in the present tense. So in this case "update README..." instead of "updated README..."

alissapajer:
Looking through our commit history, we have commits in both present and past tense. As far as I know, we haven't standardized a tense for commit messages.

jedesah:
@alissapajer Oh, I was under the impression this was discussed at some point a while ago. I guess it never got put into written form or I simply imagined it. I would vote for standardizing on the present tense but would not be strongly offended if there was no desire to standardize on a relatively minute issue like this one. I'll just continue to write my own commit message in the present tense until consensus is achieved :smile:

README.md
nicflores: What would be a good way to word it? Would something like: 
"...to run integration tests or the MetaStore the `docker/scripts` are provided..."

README.md
nicflores: that line is meant to be just an example, but I could add `quasar_metasore` in there too. I just wanted users to have an example of how to run `setupContainers` with the `-u` flag.

drostron: Sure. My motivation for adding `quasar_metastore` is that it's always needed for running `it` tests. Up to you.

README.md
```diff
@@ -50,15 +50,15 @@ Running the full test suite can be done using docker containers for various back
...
+In order to run integration tests for various backends the `docker/scripts` are provided to easily create dockerized backend data stores.
 
 Of particular interest are the following two scripts:
 
   1. `docker/scripts/setupContainers`
   2. `docker/scripts/assembleTestingConf`
 
 
-Supported backends datastores are:
+Qusar supports the following backends datastores:
```
drostron: s/Qusar/Quasar/

alissapajer: s/backends datastores/datastores/

or something like that

README.md
```diff
@@ -46,41 +46,63 @@ To compile the project and run tests, first clone the quasar repo and then execu
...
-In order to install MongoDB locally you can either use something like Homebrew (on OS X) or simply go to the MongoDB website and follow the
-instructions that can be found there.
+Once backends are ready we need to configure the integrations tests in order to inform Quasar about where to find the backends to test. 
+This information is conveyed to Quasar using the file `it/testing.conf`. Using the `assembleTestingConf` script you can generate a `testing.conf`
+file based on the currently running containerizd backends using the following command:
 
-Once we have a MongoDB instance handy, we need to configure the integration tests
-in order to inform Quasar about where to find the backends to test.
+```
+./assembleTestingConf -a
```
alissapajer: To include spark we need the `-t` addition. Maybe include that information here or nearby?

alissapajer:
@nic thinking that we should also add docs for adding mounts in the metastore (unless those docs exist and I missed them?) In particular I didn't see any docs for the step when we started the web server `sbt web/run` and then ran:
```
curl -v -X PUT http://localhost:8080/mount/fs/mongoq/ -d '{ "mongodbq": { "connectionUri": "mongodb://localhost:27021" } }'
```

nicflores:
@alissapajer,
There's a "Metadata Store" heading a little further below in the README. Maybe that's where that `curl` info should live?

alissapajer:
@nicflores that seems like a good place to me, since that `curl` info isn't needed for the integration tests. We should probably link to the metastore setup in the "REPL Usage" section too, since people need metastore for REPL.

drostron:
@alissapajer I added you as a reviewer and will merge upon your approval.

drostron:
@nicflores Needs a merge from master

PR #2054, opened 2017-04-18
Unnecessary tx shutdown and optional conf file mountings

drostron:


repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -167,10 +167,10 @@ object Main {
                    Task.now, MetaStoreConfig.configOps.default
                  ).liftM[MainErrT]
       tx      <- metastoreTransactor(msCfg)
-      _       <- EitherT((opts.cmd match {
+      _       <- opts.cmd match {
                    case Start               => start(tx)
                    case InitUpdateMetaStore => initUpdateMigrate(Schema.schema, tx.transactor, cfgPath)
-                 }).run.onFinish(κ(tx.shutdown)))
```
alissapajer: When we exit the REPL, will the metastore still get shutdown properly? Is it okay that we're not shutting it down at all anymore?

drostron: From some quick testing and observation of threads the MetaStore shuts down sufficiently. Also, the behavior you ran into is averted.

interface/src/main/scala/quasar/main/metastore.scala
```diff
@@ -136,15 +136,17 @@ object metastore {
...
       mountingsConfigDecodeJson.decodeJson(jCfg).fold(
         { case (e, _) => taskToConnectionIO(Task.fail(new RuntimeException(e.shows))) },
-        m => m.toMap.toList.traverse {
-          case (p, m) => MetaStoreAccess.insertMount(p, m)
-        }.void)
+        _.cata(
+          m => m.toMap.toList.traverse {
+            case (p, m) => MetaStoreAccess.insertMount(p, m)
+          }.void,
+          ().η[ConnectionIO]))
```
alissapajer: :+1:

PR #2058, opened 2017-04-19
TypeF: A fixpoint representation for Quasar types

### LINKED ISSUE #1944: Fixpoint representation of LogicalPlan Types
Convert our representation of types for `LogicalPlan` into fixpoint style in order to be able to use and annotate them for Statistical Structural Types (#1715) as well as take advantage of the machinery matryoshka provides.

If time permits will replace the existing type representation with this one, otherwise the goal would be to define `matryoshka.Recursive|Corecursive` instances for the existing representation that use the new representation as the pattern functor so that we can share as much implementation as possible until we're able to fully adopt the new representation.

wemrysi:
Introduces the `TypeF` pattern functor that forms the base of a fixpoint representation of Quasar types. Includes definitions and tests for the lattice operations (glb, lub), normalization, and a partial order defined on subtyping.

Also fixes a bug in `EJson` ordering that testing uncovered where we weren't actually ignoring metadata when testing for equality/order in all cases.

Closes #1944.


frontend/src/main/scala/quasar/tpe/TypeF.scala
```diff
@@ -0,0 +1,400 @@
...
+
+  def coproduct[J, A]: Prism[TypeF[J, A], (A, A)] =
+    Prism.partial[TypeF[J, A], (A, A)] {
+      case Union(x, y, INil()) => (x, y)
+    } { case (x, y) => Union(x, y, INil()) }
+
+  def union[J, A]: Prism[TypeF[J, A], (A, A, IList[A])] =
+    Prism.partial[TypeF[J, A], (A, A, IList[A])] {
+      case Union(x, y, zs) => (x, y, zs)
+    } { case (x, y, zs) => Union(x, y, zs) }
```
alissapajer: I haven't tried it out, but it looks like we should be able to use [`GenIso`](https://github.com/julien-truffaut/Monocle/blob/cb428cc0b409850b72cd8e1755e5ccd336ad2813/docs/src/main/tut/optics/iso.md#iso-generation) for many of these cases, and then convert that `asPrism`.

frontend/src/main/scala/quasar/tpe/TypeF.scala
```diff
@@ -0,0 +1,400 @@
...
+          case NonEmptyList(x,       INil()) => x
+        }
+    }
+  }
+}
+
+private[quasar] sealed abstract class TypeFInstances {
+  import TypeF._
+
+  def boundedDistributiveLattice[J: Order, T](
```
sellout: 🤘🏽

frontend/src/main/scala/quasar/tpe/TypeF.scala
```diff
@@ -0,0 +1,400 @@
...
+          _.transCata[T](
+                reduceToBottom[J, T]
+            >>> normalizeEJson[J, T]
+            >>> lowerConst[J, T]
+            >>> coalesceUnion[J, T]
+            >>> reduceToTop[J, T]
+            >>> elideBottom[J, T]
+          )
+
+        normalization.normalize[J]((x, y).apo[T](glbƒ[J, T] <<< normArg.product))
```
sellout: If you expose the underlying `Transform` in `normalize`, you could fuse it with the `apo` here. Something like,
```scala
(x, y).ghylo(distCata, distApo[T], normalizeƒ >>> (_.embed), glbƒ[J, T] <<< normArg.product)
```

frontend/src/main/scala/quasar/tpe/TypeF.scala
```diff
@@ -0,0 +1,400 @@
...
+
+sealed abstract class TypeF[J, A] extends Product with Serializable
+
+object TypeF extends TypeFInstances {
+  import SimpleType._
+
+  final case class Bottom[J, A]()                                        extends TypeF[J, A]
+  final case class Top[J, A]()                                           extends TypeF[J, A]
+  final case class Simple[J, A](simpleType: SimpleType)                  extends TypeF[J, A]
+  final case class Const[J, A](ejson: J)                                 extends TypeF[J, A]
```
sellout: Argh, we need mutual-rec again.

frontend/src/test/scala/quasar/tpe/TypeFSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+import org.specs2.scalacheck._
+import org.specs2.scalaz._
+import scalaz._, Scalaz._
+import scalaz.scalacheck.ScalazProperties._
+import scalaz.scalacheck.ScalazArbitrary._
+
+final class TypeFSpec extends Spec with TypeFArbitrary with EJsonArbitrary {
+  // TODO: Anything above 5 seems to explode (even '6'), quick thread dump
+  //       landed in scalacheck.Gen, so probably need to tune the Arbitrary
+  //       instance for TypeF (and maybe EJson).
```
sellout: You might try using `Nu` instead of `Fix` for these tests, then – it’s lazy (but there are still plenty of things that force it, so it may not help).

foundation/src/main/scala/quasar/contrib/matryoshka/EqualR.scala
```diff
@@ -0,0 +1,31 @@
...
+import matryoshka.implicits._
+import scalaz._
+
+object EqualR {
+  def equal[T, F[_]: Functor](a: T, b: T)(implicit T: Recursive.Aux[T, F], E: Delay[Equal, F]): Boolean =
+    E(equalR[T, F](E)).equal(a.project, b.project)
+
+  def equalR[T, F[_]: Functor](eqF: Delay[Equal, F])(implicit T: Recursive.Aux[T, F]): Equal[T] =
+    Equal.equal[T](equal[T, F](_, _)(Functor[F], T, eqF))
+}
```
wemrysi: It isn't, I just hadn't thought to look for such a thing :wink:

foundation/src/main/scala/quasar/contrib/matryoshka/OrderT.scala
```diff
@@ -1,46 +0,0 @@
...
-        del(orderT[F](del)).order(tf1.project, tf2.project)
-    }
-}
-
-sealed abstract class OrderTInstances {
-  import OrderT._
-  implicit def fixOrderT: OrderT[Fix] = recursiveT[Fix]
-  implicit def muOrderT: OrderT[Mu]   = recursiveT[Mu]
-  implicit def nuOrderT: OrderT[Nu]   = recursiveT[Nu]
-}
```
wemrysi: Probably, though we might be able to avoid it more often by avoiding reference to `T[_[_]]`.

sellout: Yeah – I was wondering about that – QScript and TypeF differ in that they both carry an additional parameter, but in QScript it’s the `T[_[_]]` that is used to covert the EJson functor to fixed-point, and for TypeF, it’s `J` that is the fixed-point of EJson. I kind of like your way, but I haven’t thought it through enough yet.

Your way does let us avoid `OrderT`, but we (at least sometimes) will need to pass `Order[J]` instead. Also, I don’t think if it work work for faking mutual recursion in general. Both QScript and TypeF are more hierarchical … QScript contains MapFuncs which contain EJson … but something like `Diff` in Matryoshka alternates between Diff nodes and F nodes.

wemrysi: Yeah, but `Order[J]` is what we're actually interested in, `OrderT` is implementation machinery used to materialize instances of the former, it _seems_ like it shouldn't have to show up in constraints.

frontend/src/test/scala/quasar/tpe/TypeFSpec.scala
wemrysi: Appreciate the idea, but it didn't seem to have an effect on the performance :disappointed:

foundation/src/main/scala/quasar/contrib/matryoshka/OrderT.scala
wemrysi: Also, I think you're right about this not being a general strategy for faking muturec.

wemrysi:
@sellout I've pushed a few new commits that should address your comments.

frontend/src/main/scala/quasar/tpe/TypeF.scala
```diff
@@ -131,7 +131,11 @@ object TypeF extends TypeFInstances {
             >>> elideBottom[J, T]
           )
 
-        normalization.normalize[J]((x, y).apo[T](glbƒ[J, T] <<< normArg.product))
+        (x, y).ghylo[Id, T \/ ?](
+          distCata[TypeF[J, ?]],
+          distApo[T, TypeF[J, ?]],
+          normalization.normalizeƒ[J, T] >>> (_.embed),
+          glbƒ[J, T] <<< normArg.product)
```
sellout: It’d be nice to have a macro that expands this from
```scala
(x, y)
  .apo[T](glbƒ[J, T] <<< normArg.product)
  .cata(normalization.normalizeƒ[J, T] >>> (_.embed))
```
eh?

wemrysi: A "fold-optimizer" macro, that _would_ be nice.

frontend/src/main/scala/quasar/tpe/TypeF.scala
alissapajer: Yeah, it's out of the scope of this PR since we do this everywhere, but I've been meaning to try it out!

alissapajer: So tried it out, the problem is this:
```scala
sealed abstract class Bar
final case class Foo(a: Int) extends Bar

// what we get
val prismFoo: Prism[Foo, Int] = GenIso.fields[Foo].asPrism

// what we need
val prismBar: Prism[Bar, Int] = ???
```
I wonder if there's a way around that...

wemrysi: @alissapajer How about `GenPrism[Bar, Foo] composeIso GenIso.fields[Foo]`?

alissapajer: @wemrysi I was just looking at `composeIso`, trying to find the right thing to compose with to deal with the subtyping, and I think you've found it! :) But, here's the catch: `isInstanceOf` (and `asInstanceOf`) are used within the macro and we get `[wartremover:IsInstanceOf] isInstanceOf is disabled` error upon compile in quasar. I tried:
```scala
@SuppressWarnings(Array("org.wartremover.warts.isInstanceOf")) 
val prism: Prims[Bar, Int] = GenPrism[Bar, Foo] composeIso GenIso.fields[Foo]
```
but that doesn't have an effect. So we'd need a way to ignore warts within a macro...

alissapajer: Ok the problem was that I mistyped the string. s/isInstanceOf/IsInstanceOf/ solves the problem and it compiles!

PR #2070, opened 2017-04-21
Json codec for EJson

### LINKED ISSUE #2066: A Json encoding for EJson
In order to serialize `EJson` as `Json`, we will encode the extended parts of `EJson` as follows:

1. All `JSON` compatible values are encoded verbatim.

2. Extended values are encoded as
```
Meta -> { ∃value: ..., ∃meta: ...}
Map  -> { ∃map: [{∃key: ..., ∃value: ...}, ...] }
Byte -> { ∃byte: 42 }
Char -> { ∃char: "x" }
Int  -> { ∃int: 2345 }
```

3. A `Map` where the keys are all strings will be encoded as a `Json` object with encoded `EJson` values.

4. Any map keys in a source `EJson` value that begin with `∃` need to be prefixed with an additional `∃` when encoding (e.g. `∃foo` becomes `∃∃foo`) and, conversely, any map keys that begin with `∃` when decoding, and aren't one of the special keys defined above, need to have a single `∃` removed (e.g. `∃∃bar` becomes `∃bar`).

wemrysi:
Implements the `Json` encoding for `EJson` described in #2066.

I had to pull in the `Arbitrary` machinery for corecursive types from `matryoshka` as it still depends on a snapshot version of `scalacheck`. Once we're able to upgrade to the latest `matryoshka`, we can get rid of it.

Closes #2066.

### DIFF
--- a/ejson/src/test/scala/quasar/ejson/JsonCodecSpec.scala
+++ b/ejson/src/test/scala/quasar/ejson/JsonCodecSpec.scala
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.ejson
+
+import slamdata.Predef._
+import quasar.Qspec
+import quasar.contrib.matryoshka._
+import quasar.contrib.matryoshka.arbitrary._
+import quasar.fp._
+
+import matryoshka.{equalTEqual => _, _}
+import matryoshka.data.Fix
+import matryoshka.implicits._
+import org.specs2.scalacheck._
+import scalaz._, Scalaz._
+
+final class JsonCodecSpec extends Qspec with EJsonArbitrary {
+  import JsonCodec.DecodingFailed
+
+  implicit val params = Parameters(maxSize = 10)
+
+  type E = Fix[EJson]
+  type J = Fix[Json]
+
+  val decodeMƒ: CoalgebraM[DecodingFailed[J] \/ ?, EJson, J] =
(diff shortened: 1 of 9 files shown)


ejson/src/test/scala/quasar/ejson/JsonCodecSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+    val (k1, k2) = (JsonCodec.Sigil.toString + k10, JsonCodec.Sigil.toString + k20)
+    val m = ExtEJson(Map(List(
+      CommonEJson(str[E](k1)).embed -> v1,
+      CommonEJson(str[E](k2)).embed -> v2
+    ))).embed
+
+    roundtrip(m).toEither must beRight(equal(m))
+  }
+
+  "map keys equal to one of the codec keys are preserved" >> prop { v: E =>
```
alissapajer: Could you also add an explicit test for `ExtKeys.toList.map(sigild)` (creating keys that look like `∃∃meta`, `∃∃map` etc.) Since this was an example that drove us to prefix _all_ string keys that start with `∃` with another `∃`, it would be good to have an explicit test for it.

wemrysi: Will do!

ejson/src/test/scala/quasar/ejson/JsonCodecSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+    _.cata(JsonCodec.encodeƒ[J]).anaM[E](decodeMƒ)
+
+  val soloKeys = JsonCodec.ExtKeys \\ ISet.fromList(List(JsonCodec.KeyK, JsonCodec.ValueK, JsonCodec.MetaK))
+
+  "faithfully roundtrip EJson" >> prop { e: E =>
+    roundtrip(e).toEither must beRight(equal(e))
+  }
+
+  soloKeys.toList foreach { k =>
+    s"fail when singleton map with `$k` key maps to unexpected value" >> {
```
wemrysi: Oh, that's actually string interpolation as it creates an example for each key in `soloKeys`, so it will result in examples titled like "fail when singleton map with `∃int` key maps to unexpected value".

alissapajer: Oh! I see that now, cool. I had seen `$key` used in place of `∃key` in some of the earlier discussion, so I just read it like that. :)

wemrysi: I did the same thing, it was only until I went to make the change that I realized it :wink:

PR #2071, opened 2017-04-22
knobs version updated

nicflores:
Knobs version updated to 4.0.30

### DIFF
--- a/it/testing.conf.example
+++ b/it/testing.conf.example
@@ -1,5 +1,3 @@
-# This property is never used, but it's here so that Knobs doesn't fail to parse the file if all other
-# parameters are commented out
 #postgresql_metastore = "{\"host\":\"localhost\",\"port\":5432,\"database\":\"metastore_test\",\"userName\":\"postgres\",\"password\":\"postgres\"}"
 #mongodb_2_6       ="mongodb://<host>:<port>"
 #mongodb_3_0       ="mongodb://<host>:<port>"
--- a/it/testing.conf.example
+++ b/it/testing.conf.example
@@ -1,6 +1,5 @@
 # This property is never used, but it's here so that Knobs doesn't fail to parse the file if all other
 # parameters are commented out
-# TODO: Remove once PR #49 of Knobs gets merged and released (https://github.com/Verizon/knobs/pull/49)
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -137,5 +137,5 @@ object Dependencies {
     CommonDependencies.argonaut.monocle                         % Test,
     CommonDependencies.http4s.blazeClient                       % Test,
     CommonDependencies.refined.scalacheck                       % Test,
-    "io.verizon.knobs" %% "core"                % "3.12.27a"    % Test)
+    "io.verizon.knobs" % "core_2.11"  % "4.0.30-scalaz-7.2"     % Test)
```
jedesah: Why the change to `%` over `%%`?

nicflores: I tried staying sticking with our `%%` pattern but the only thing that worked was what is posted [here](http://search.maven.org/#artifactdetails%7Cio.verizon.knobs%7Ccore_2.11%7C4.0.30-scalaz-7.1%7Cjar). Using the original style:

```
"io.verizon.knobs" %% "core"  % "4.0.30-scalaz-7.2"         % Test
```

kept doubling up the core version

```
[trace] Stack trace suppressed: run last it/*:coursierResolution for the full output.
[error] (it/*:coursierResolution) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error]   io.verizon.knobs:core_2.11_2.11:4.0.30-scalaz-7.2:
```

Let me know if I'm missing something.

drostron: The one you posted in your comment is going to be what you want. I just gave it a whirl and seems ok. Maybe you had both `%` and `core_2.11` when you encountered the error.

nicflores: Thanks @drostron and @jedesah for the heads up. What I posted above works so I went ahead and changed it.

drostron:
@nicflores [This](https://github.com/quasar-analytics/quasar/blob/56ff1b6e7643a5ae5c87e68ac4f6c344b9527071/it/testing.conf.example#L1-L4) can now be removed.

it/testing.conf.example
```diff
@@ -1,7 +1,6 @@
 # This property is never used, but it's here so that Knobs doesn't fail to parse the file if all other
```
jedesah: The comments above should be removed as well

PR #2072, opened 2017-04-24
Brush up MetaStore schema migration

drostron:


alissapajer:
@drostron were you ever able to replicate the issue I had with my migration (where none of the entries ended up in the metastore)? I unfortunately don't have the original config file that was migrated from :/

drostron:
@wemrysi I've tested a variety of migration paths including existing schema and data for both Quasar and QAdv.

drostron:
@alissapajer I've been unable to reproduce the behavior you ran into. I seem to remember a few mounts made it to the metastore and a few didn't. There's a chance the the over eager transaction close that's been fixed caused the issue.

Issue #1852, opened 2017-01-12
Incorrect mimetype being sent for font files


natefaubion:
Quasar is currently sending font files as `text/plain` which breaks in Chrome, so no icons show up.

sellout:
And it seems like this is something that changed with the recent update to HTTP4S, since it was working properly in a previous Quasar release, right?

natefaubion:
Correct.

natefaubion:
I was previously using QA 0.5.4, which is based on 13.something, and everything worked correctly then.

drostron:
Once I add an Origin header to a request I'm able to reproduce the issue. It appears to be something related to CORS. Investigation continues...

beckyconning:
This is also affecting videos in Safari.

drostron:
The current plan is to downgrade http4s until fixed. @jr offered to submit a PR for this.

drostron:
The current plan is to downgrade http4s until fixed. @jedesah offered to submit a PR for this.

PR #2081, opened 2017-04-24
Upgrade to `http4s` version `0.15.9a`

jedesah:
- Upgrade `jawn` to version `0.10.4`
- Decouple `quasar` from `sbt-slamdata`'s version of `http4s`
- Remove code that worked around a `jawn` bug

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,12 +9,13 @@ import slamdata.CommonDependencies
 
 object Dependencies {
   private val disciplineVersion        = "0.5"
-  private val jawnVersion              = "0.8.4"
+  private val jawnVersion              = "0.10.4"
   private val jacksonVersion           = "2.4.4"
   private val matryoshkaVersion        = "0.18.3"
   private val pathyVersion             = "0.2.9"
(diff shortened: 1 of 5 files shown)


jedesah:
I just got so fed up with staring at that `bugFreeArgonautFacade` that I decided to see if the fixed version of `jawn` was released which it was but it required an upgrade to `http4s` as well. I then remembered #1852 which is likely not fixed but perhaps an open PR will catch our attention to get around to fixing that bug so that we don't get too far behind on `http4s` (and our dependencies in general). A quick glance at the `http4s` changelog tells me that substantial amount of bugs were fixed since `0.14.1` so if it doesn't fix #1852 then at least it fixes a bunch of others.

@drostron Since you were able to reproduce the issue for #1852, can you see if you can reproduce it on this branch and perhaps post the reproduction steps if you can.

Also, I didn't test any connectors, so I wouldn't be all that surprised if there is some kind of binary incompatibility there.

jedesah:
Revision label assumes we are still versioning according to the web api

jedesah:
@drostron Oh, awesome! That's good news.

PR #2082, opened 2017-04-24
Spew out a better error message when `quasar_metastore` is missing when running `it` tests

jedesah:


nicflores:
Verified message shows up if you don't have the `quasar_metastore` running in the background:

```
[trace] Stack trace suppressed: run last it/test:test for the full output.
[error] Could not run test quasar.metastore.PostgreSqlAccessSpec: org.specs2.specification.dsl.mutable.SpecificationCreationException:
[error] An exception was raised during the creation of the specification: You must configure the quasar_metastore backend as described in the README in order to run this test without encountering a failure (yes, we know it would be better if we could tell specs2 to mark it as skipped when that backend is not present, but that's tricky because of the way doobie provides us with some fixtures that are inheritance based.
[error] This means that you have some code which should be enclosed in an example. Instead of writing:
```

PR #2084, opened 2017-04-25
Refactor code that sets up interpreters for tests in `web`

jedesah:
- Make some re-usable interpreters more easily accessible for dependent projects like `quasar-advanced`


core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -230,4 +231,13 @@ object Mounting {
     implicit def apply[S[_]](implicit S: Mounting :<: S): Ops[S] =
       new Ops[S]
   }
+
+  object impl {
+    def constant: Mounting ~> State[Map[APath, MountConfig], ?] = {
```
alissapajer: This appears to only be used in the test file `core/src/test/scala/quasar/fs/mount/Fixture.scala`. You could probably just put this `def` there inside `object Fixture`.

jedesah:
@alissapajer I did the change you suggested and also added a bit more code that is useful in `quasar-advanced` but is general enough to be in quasar

alissapajer:
Regarding the addition of `runLogCatch` and `eitherTThrowableCatchable`, I don't think we should be adding things to quasar that aren't used in quasar. If the only use case is in quasar-advanced, then we should just keep them there. Otherwise we end up maintaining functions that may not be used anywhere.

Basically, imo, if a function has no call sites and no test calling it, it should be deleted, and these new additions would qualify.

jedesah:
@alissapajer I think you are right. Although I was pretty sure I could make use of `runLogCatch` in quasar but instead of just debating that point I decided to see if that was actually true hence the additional commit which makes use of it for the filesystem tests. I think you will be relatively happy with the amount of code it removes elsewhere for us not to maintain :smile:

jedesah:
@wemrysi Feel free to have a look as well at my last commit as I made some pretty extensive changes to the filesystem tests and found what I believe to be a couple bugs with them.

alissapajer:
@jedesah is the implicit `eitherTThrowableCatchable` in use? If not, then could we remove that?

connector/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -100,55 +95,25 @@ trait FileSystemFixture {
...
+  // TODO[scalaz]: Shadow the scalaz.Monad.monadMTMAB SI-2712 workaround
+  import StateT.stateTMonadState
 
-    def runLogEmpty[A](p: Process[FileSystemErrT[F,?],A]): Task[FileSystemError \/ IndexedSeq[A]] =
-      runLogE(p).run.eval(emptyMem)
+  object Mem extends Interpreter[FileSystem, InMemoryFs](interpretTerm = fileSystem) {
+    def runEmpty[A](program: Free[FileSystem, A]) = run(program).eval(emptyMem)
+  }
+  object ReadWriteMem extends Interpreter[FileSystem, InterceptInMemoryFs](interpretTerm = readWrite) {
+    def runWithWriteErrors[A](program: Free[FileSystem, A], errors: List[Vector[FileSystemError]]) =
```
alissapajer: Can you include the return type on this and on `runEmpty` above?

alissapajer:
@jedesah also could you point out which tests are the ones you found bugs in?

foundation/src/main/scala/quasar/fp/free/Interpreter.scala
```diff
@@ -33,72 +31,39 @@ import scalaz.stream.Process
 class Interpreter[F[_], M[_]: Monad](val interpretTerm: F ~> M) {
   type Program[A] = Free[F,A]
 
-  def interpret: Program ~> M =
```
jedesah: Ya, I can get behind that. I changed this mostly because we use `run` in a lot of places in the code base for this purpose. I guess if we decide `interpret` is better, we can slowly migrate those occurrences into `interpret` as we run into them (no pun intended). I think @wemrysi tends to use `run` but I could be wrong about that. Would be curious to hear your insight on that @wemrysi. `run` does have the advantage of being shorter but it can also easily be confused with all the other ways we use `run`.

I don't feel too strongly about which one we choose but it would be nice to kind of decide instead of having instances of `interpret` and instances  of `run` depending on who last touched the code 😄

connector/src/test/scala/quasar/fs/FileSystemFixture.scala
```diff
@@ -100,55 +95,25 @@ trait FileSystemFixture {
...
-      type T1[M[_],A] = EitherT[M,E,A]
-      type T2[M[_],A] = WriterT[M,L,A]
-      interpretT2[T1,T2].apply(term).run.run
-    }
-  }
-
-  val hoistTask: InMemoryFs ~> MemStateTask =
-    Hoist[StateT[?[_], InMemState, ?]].hoist(pointNT[Task])
+  type FreeFS[A]            = Free[FileSystem, A]
+  type InterceptInMemoryFs[A]     = ReadWriteT[InMemoryFs, A]
```
jedesah: @alissapajer How about `InjectErrorsMemFs` or something like that? The objective of this `Monad` is to "inject" read and write errors into the interpretation of the program so that we can test how our code reacts to those failures. My gripe with `ReadWriteInMemoryFs` is that it describes the type (not all that well since it doesn't say what we are writing and reading) instead of the intention.

connector/src/test/scala/quasar/fs/FileSystemFixture.scala
jedesah: Will do

jedesah:
@alissapajer It's used by `AugmentedProcess`

jedesah:
@alissapajer Good idea about pointing out the bugs. Will do. I thought about doing that but then I forgot so thx for pointing it out.

connector/src/test/scala/quasar/fs/WriteFileSpec.scala
```diff
@@ -85,42 +79,45 @@ class WriteFileSpec extends org.specs2.mutable.Specification with org.specs2.Sca
...
+        val program = write *> query.fileExists(f)
 
-        MemTask.runLogEmpty(p).unsafePerformSync must_=== \/-(Vector(true))
+        Mem.run(program).eval(emptyMem) must_=== true
       }
 
       s"$n should leave existing file untouched on failure" >> prop {
         (f: AFile, xs: Vector[Data], ys: Vector[Data]) => (xs.nonEmpty && ys.nonEmpty) ==> {
           val err = writeFailed(Data.Str("bar"), "")
-          val ws = xs.as(Vector()) :+ Vector(err)
```
jedesah: I thought there was a bug here but I just finally understood how it was working before and it's fine there was no bug, I'd just argue that it was harder to understand, in part because the program itself was performing setup and checking final state instead of leaving that to the interpretation and assertion down below.

jedesah:
@alissapajer Ya, so, as per my previous line comment, there were no bugs with the tests, after all, I just had trouble understanding what was going on.

foundation/src/main/scala/quasar/fp/free/Interpreter.scala
alissapajer: Yeah, `run` is generally overused, or at least used consistently in various unrelated contexts, so in general I'm in favor of using words that aren't "run" wherever we can. 

I did a quick grep through quasar and it looks like we often use `run` to define a natural transformation like this. But we also use `run` in various other contexts. So I think it could be good to use some other word like "interpret" only so "run" isn't overused.

connector/src/test/scala/quasar/fs/FileSystemFixture.scala
alissapajer: I think in general I prefer to name type aliases according to what they _are_ rather than what they _do_, and then encode what they do in the types themselves. In this case we are creating a `StateT[InMemoryFs, (Reads, Writes), A]` which could potentially have other uses than what we've used it for, so I'm hesitant to name it with "inject" or "intercept".

It looks like this type is actually only used in one place, so I would be fine omitting the type alias altogether and just inlining the type, rather than giving it an alias.

jedesah: I kind of see it like I would `type Age = Int`. I guess one might argue not to do `type Age = Int` but to do instead `case class Age(value: Int) extends AnyVal` (along with associated instances for `Show`, `Equal`, etc). But in some cases, I think it's just not worth doing all that (small scope, test code, etc) so I would say that `type Age = Int` is fine in some cases and by extension so is `InjectErrorsMemFs` (or `InMemoryFs` for that matter).

I'd be somewhat okay with inlining, but unfortunately, it's used in three places as far as I can see.

> encode what they do in the types themselves

What do you mean by this?

alissapajer: It essentially has only one usage place. It is used in one place which is directly called in another. That's what I meant.

I don't see any name helping me figure out what this type does, including something like `InjectErrorMemFs` because I'd still have to go to the call sites and see how it's used. So if you want to call it that, that's fine with me. My point was just that ultimately I would have to look at the type itself and figure out how it's used, so I didn't see a name like that being helpful, in my opinion. But I'm not going to object to it.

jedesah: I think I could be convinced that `InMemStateT` and `ReadWriteInMemStateT` would be better names perhaps for both of these aliases. I'd say until we change `InMemFs`, it would be nice to stay consistent though.

jedesah:
@alissapajer Okay, I just removed the alias after all :smile:

foundation/src/main/scala/quasar/contrib/scalaz/State.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.contrib.scalaz
+
+import scalaz._, Scalaz._
+
+object State {
+  def evalNT[A, M[_]:Monad](value: A): State[A, ?] ~> M =
```
wemrysi: A slightly more general version of this exists as `quasar.fp.evalNT`.

foundation/src/main/scala/quasar/fp/free/Interpreter.scala
wemrysi: I'd agree that `run` is probably overused, it has sort of a de facto convention of being "the function that eliminates this type" for newtypes and monad transformers, at least. These natural transformations similarly eliminate the lhs, so that was my thought process there. Anyhow, I don't have a preference either way, `interpret` is fine with me.

PR #2045, opened 2017-04-12
Generic path card

rabbitonweb:
Part of https://github.com/quasar-analytics/quasar/issues/1716

This provides generic implementation of `pathCard` that will be available for ALL connectors. It relies on `QueryFile` algebra and calculates cardinality by running `select count(*) from apath` for a given `APath`

PR #2086, opened 2017-04-25
AnalyticalFileSystem +  Analyze Query endpoint in REST API

rabbitonweb:
@jedesah This PR is ready for review.

We introduce `Analyze` which is a new algebra containing (currently) only one verb - `QueryCost`. It is responsible to take existing `LogicalPlan` and evaluate its costs/weight.

Originally `QueryCost` was supposed to be part of `QueryFile` algebra, but had to be moved to its own separate algebra. To understand why that is, we need to look at two typeclasses `Cardinality` and `Cost` (they are not part of this PR, they are already in master). Those two typeclasses provide functionality necessary to complete 'query analysis' functionality.

`Cardinality` has instance that can be used by all connectors.
`Cost` ideally would be implemented by each connector separately (each connector has its own internal knowledge that can be used to evaluate cost of the query). However we provide instance (very naive one) that can be used by any connector.

Finally both `Cardinality` and `Cost` need `pathCard` function to work. It is a function that evaluates cardinality for specific path. This ideally should also be implemented by each of the connectors separately, but there is a default implementation that just reuses `QueryFile.evaluate` and calculates cardinality for given `afile` by running query `select count(*) from afile`. Because of *THAT* `QueryCost` had to be moved to its own algebra, because we needed `QueryFile ~> M` interpreter in the scope, to write `Analyze ~> M`. Without it (if `QueryCost` was part of `QueryFile`) the interpreter would have to use recursion calls - which was unfortunate.

All stuff described above is already in the master (most of it). This PR was supposed to be a simple one: putting everything together - which should be easy, right?

Well, it was not.

To make everything working one of two options could be chosen

1. Add `Analyze` as part of `FileSystem`
2. Create new effect called `AnalyticalFileSystem` (not best name, I know, see disclamiers)

ad.1 it seemed wrong because:
a) `Analyze` has very little to do with file system as a concept - thise seemed as wrong abstraction
b) it would break tons of tests were it really should not (smell of a wrong abstraction again)

ad.2 Required a bit less changes but it was challenging anyway. Opened a door for improvement in separating in the future different concerns (analyzing, quering, file system).

TL;DR 
1. had to introduce `Analyze`
2. introduced `AnalyticalFileSystem` as `Analyze` + `FileSystem`
3. everything else is attempt to glue those together

Disclaimers:
1.  @jedesah please note that there is one PR (https://github.com/quasar-analytics/quasar/pull/2045) not yet merged to master - this PR is based on it. However this should not be a blocker, I hope to have #2045 in master tomorrow in the morning 

2. `Analyze` was combined with `FileSystem` to form `AnalyticalFileSystem` - that name is not the best one we could choose. There is a separate issue planned for May2007 to fix those naming controversies (and discuss them if needed). https://github.com/quasar-analytics/quasar/issues/2113


rabbitonweb:
@jedesah I've added long description for this PR plus TLDR

connector/src/main/scala/quasar/qscript/analysis/package.scala
```diff
@@ -46,11 +47,22 @@ package object analysis {
...
-  ): M[Int] = rec.zygoM(t)(cardinalty.calculate(pathCard), cost.evaluate(pathCard))
+  def pathCard[S[_]](
+    implicit queryOps: QueryFile.Ops[S]
+  ): APath => FileSystemErrT[Free[S, ?], Int] = (apath: APath) => {
+    val afile: Option[AFile] = peel(apath).map {
+      case (dirPath, \/-(fileName)) => dirPath </> file1(fileName)
+      case (dirPath, -\/(dirName))  => dirPath </> file(dirName.value)
+    }
+
+    val lpFromPath: FPath => Fix[LogicalPlan] = (p: FPath) => Fix(agg.Count(Fix(LPRead(p))))
```
jedesah: I don't think @sellout will be very happy that this is hardcoded to `Fix` especially since we have `Recursive.Aux` above.

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
```diff
@@ -47,13 +50,15 @@ package object fs {
 
   final case class TmpPrefix(run: String) extends scala.AnyVal
 
+
```
jedesah: This can be removed

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
```diff
@@ -91,13 +98,17 @@ package object fs {
 
   val QScriptFsType = FileSystemType("mongodbq")
 
+  val toQS: Fix[LogicalPlan] => FileSystemErrT[Free[QueryFile, ?], Fix[MongoQScript[Fix, ?]]] =
+    (lp: Fix[LogicalPlan]) => ???//toMongoQScript(lp, ???).mapT(_.value)
```
jedesah: Was this intentionally left like this? Seems like it warrants an explanation at least

skeleton/src/main/scala/quasar/physical/skeleton/fs/package.scala
```diff
@@ -25,17 +25,19 @@ import scalaz.concurrent.Task
 package object fs {
   val FsType = FileSystemType("skeleton")
 
+  def analyze[S[_]]: Analyze ~> Free[S, ?] = Empty.analyze[Free[S, ?]]
   def query[S[_]]: QueryFile ~> Free[S, ?] = Empty.queryFile[Free[S, ?]]
   def read[S[_]]: ReadFile ~> Free[S, ?] = Empty.readFile[Free[S, ?]]
   def write[S[_]]: WriteFile ~> Free[S, ?] = Empty.writeFile[Free[S, ?]]
   def manage[S[_]]: ManageFile ~> Free[S, ?] = Empty.manageFile[Free[S, ?]]
 
+
```
jedesah: Extra line

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -49,10 +49,12 @@ object RestApi {
         S7: MountingFailure :<: S,
         S8: PathMismatchFailure :<: S,
         S9: Module :<: S,
-        S10: Module.Failure :<: S
+        S10: Module.Failure :<: S,
+        S11: Analyze :<: S
       ): Map[String, QHttpService[S]] =
     ListMap(
       "/compile/fs"  -> query.compile.service[S],
+      "/analysis/cost/fs"  -> query.analysis.service[S],
```
jedesah: How about calling this `"/estimate/fs"`? If this was specified in the ticket then so be it or else it would be nice to be similar to the other ones

web/src/test/scala/quasar/api/services/RestApiSpecs.scala
```diff
@@ -34,26 +34,28 @@ import org.specs2.matcher.TraversableMatchers._
...
         Task.fail(new RuntimeException("unimplemented"))
     }
 
-    val fs =
-      runFs(InMemState.empty)
-        .map(MountingFileSystem.interpret(mount, _))
+    val analyze = Empty.analyze[Task]
 
-    val eff: Task[Eff ~> Task] = fs map { runFs =>
+    type MountingFileSystem[A] = Coproduct[Mounting, FileSystem, A]
```
jedesah: I don't remember seeing where this was removed so this must be a duplicate

connector/src/main/scala/quasar/fs/Analyze.scala
```diff
@@ -0,0 +1,59 @@
...
+import matryoshka.data.Fix
+import scalaz._
+
+sealed abstract class Analyze[A]
+
+object Analyze {
+
+  final case class QueryCost(lp: Fix[LogicalPlan]) extends Analyze[FileSystemError \/ Int]
+
+  // TODO rename to Ops
```
rabbitonweb: I should yes :)

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -303,6 +309,7 @@ object QueryFile {
         .map(cs => Process.emitAll(cs) flatMap (Process.emitAll(_)))
     }
 
+
```
rabbitonweb: 👍

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
rabbitonweb: 👍

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
rabbitonweb: @sellout said it should be left like this since this code will be removed in 4.2 (old mongo), @sellout can you confirm?

skeleton/src/main/scala/quasar/physical/skeleton/fs/package.scala
rabbitonweb: 👍

web/src/main/scala/quasar/api/services/RestApi.scala
rabbitonweb: I'm ok with renaming if needed but `analysis` was chosen for a reason.
There will be other things here in the future that represent some analysis. This name was chosen with @sellout & @wemrysi as a consensus :) Maybe they could comment on it, becasue I can not recall at the moment what other "analysis" might be here in the future

rabbitonweb:
@jedesah regarding https://github.com/quasar-analytics/quasar/pull/2086#pullrequestreview-35534111

The idea is to no longer grow `FileSystem` into something that it is not. `Analyze` should not be part of `FileSystem` nor should most probably `QueryFile`. As disclamier to this PR I've added short info about it: "Analyze was combined with FileSystem to form AnalyticalFileSystem - that name is not the best one we could choose. There is a separate issue planned for May2007 to fix those naming controversies (and discuss them if needed). #2113". The idea is to finally at some point call it something like `Connector`, haveing `FileSystem` holding only FS realted algebras. Adding `Analyze` to `FileSystem` is move in the wrong direction.

@sellout @wemrysi could you guys add comment on your side about this?

web/src/main/scala/quasar/api/services/RestApi.scala
wemrysi: I like `/estimate/fs`, we may have other analyses in the future (missing indexes might be an example), but how we organize their implementation doesn't necessarily need to be reflected in the HTTP api.

wemrysi:
@rabbitonweb You pretty much said it all ;) But yeah, the idea is that `AnalyticalFileSystem` is temporary and we eventually come up with a name for the set of effects for a backend, maybe we call it `Backend` or `Connector`, and that will include the filesystem, analysis and whatever other effects we end up needing.

jedesah:
Okay, cool, I like `Connector`. `AnalyticalFileSytem` is a good temporary as we will probably be fairly motived to change it!

web/src/test/scala/quasar/api/services/RestApiSpecs.scala
rabbitonweb: That was moved from production code. Let me check where that was..

rabbitonweb: there it is https://github.com/quasar-analytics/quasar/pull/2086/files#diff-cd0fb92878c764a7808667b47d223131L51

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
jedesah: Looks like this code will not be removed in 4.2 after all no?

connector/src/main/scala/quasar/qscript/analysis/package.scala
rabbitonweb: Well this *is* `Fix` because I'm constructing here `LP` so I can use it on line `60` and then `62`.

`lpFromPath` is not some function that becomes part of API. It is local variable used in `pathCard` method.

To build bigger context: 
Both `Cardinalty` and `Cost` typeclasses take as a argument a function `APath => M[Int]`. The idea is that each connector will provide body for this signature (because maybe storage specific mechanism can give fast estimate algorithm e.g cassandra is using HyperLogLog algorithm). For connectors that don't provide that body, we created a "generic" implementation. It creates `LogicalPlan` that represents `select count(*) from afile`.

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
rabbitonweb: @sellout I'm provide Empty.analyze for old mongo

connector/src/main/scala/quasar/qscript/analysis/package.scala
sellout: I think this should still be parameterized like
```scala
def lpFromPath[LP](p: FPath)(implicit LP: Corecursive.Aux[LP, LogicalPlan]): LP =
  agg.Count(LPRead(p).embed).embed
```
(as should the things below)

If some of them need to be specifically `Fix` (because they’re calling something else that isn’t yet abstracted), then the bodies should still use `.embed`/`.project` instead of `Fix(_)`/`.unFix`.

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
rabbitonweb: @jedesah well this is a bit embarrassing but this turned out to be the NEW mongo connector (not the old) - we just did not finished it with @sellout 

I will revert this one to `Empty.analyze` so adding interpeter for new monog connector can be achieved in separate PR

👍  or 👎  ?

jedesah: That sounds fine

connector/src/main/scala/quasar/qscript/analysis/package.scala
rabbitonweb: But this is ... local variable... it is used in once place (with Fix). Should be abstraction used when it's needed (giving functionality to the user). Here I'm using it locally, what's the point? Staying abstract for the sake of being abstract?

connector/src/main/scala/quasar/fs/transformPaths.scala
```diff
@@ -202,11 +202,11 @@ object transformPaths {
   ): S ~> Free[S, ?] = {
     import Analyze._
 
-    val U = Analyze.Unsafe[S]
+    val O = Analyze.Ops[S]
```
jedesah: I don't think `O` is a very good name for an identifier here. Perhaps `A` or `analyze`.

rabbitonweb: You are right, you can see why I did that:

```
  def queryFile[S[_]](
    inPath: EndoK[AbsPath],
    outPath: EndoK[AbsPath],
    outPathR: EndoK[RelPath]
  )(implicit
    S: QueryFile :<: S
  ): S ~> Free[S, ?] = {
    import QueryFile._

    val Q = QueryFile.Ops[S]
    val U = QueryFile.Unsafe[S]
```

connector/src/main/scala/quasar/qscript/analysis/package.scala
rabbitonweb: @sellout ^^ ?

rabbitonweb: @sellout @jedesah regarding above (that's the last thing that is not closed for this review).

What do you think about my comment (it being local variable and making it too abstract is adding magic with little value).

If you feel strongly about keeping this abstract, when I can get `.embed` from (probably some implicit conversion that I'm not aware of)

rabbitonweb: @sellout @jedesah I will push changes making `lpFromPath` more abstract but I would really like to understand the benefit of it when it is just a local value

rabbitonweb:
@jedesah this is ready for review

connector/src/main/scala/quasar/qscript/analysis/package.scala
jedesah: My initial comment was based on a very summary look at this code. If there is no way of getting rid of `Fix` entirely in this snippet of code because some client code is hardcoded, I would tend to agree there is little value in making a local variable more abstract then it needs to be. Once we can get rid of `Fix` completely in this snippet then things will need to be made more abstract but I don't think we need to jump the gun on that. If Scala made it just as concise (or more) to make it generic then that's great and hurray for parametricity, but since that's not the case, I think we need to be careful where we introduce boilerplate-heavy generality to be sure it's actually providing value or else we are just making the codebase harder to understand for ourselves and for new contributors without any counterbalancing benefit.

PR #2089, opened 2017-04-25
Implement first-class LogicalPlan joins.

### LINKED ISSUE #1539: self-joins get elided

### LINKED ISSUE #1866: failed autojoin when joining sub-queries

### LINKED ISSUE #1943: First-class joins in LogicalPlan
Implement first-class joins in `LogicalPlan`. Instead of joins being represented as yet another function, we will represent them using their own structure, which will allow us to reference the left and right side branches. This branch information will then be passed along to `QScript`, which will mean that `QScript` will no longer have to perform lots of work to determine which branch is which and if/how the branches merge.

alissapajer:
Closes #1943 
Fixes #1539 
Fixes #1866


frontend/src/main/scala/quasar/frontend/logicalplan/Optimizer.scala
```diff
@@ -136,6 +137,16 @@ final class Optimizer[T: Equal]
     case _ => None
   }
 
+  // TODO delete this when old mongo is deleted
+  val reconstructOldJoins: Algebra[LP, T] = {
```
alissapajer: This is what we do to convert the new LP joins back to the old ones to satisfy old mongo.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -2774,13 +2774,7 @@ class PlannerSpec extends
               $unwind(DocField(JoinHandler.LeftName)),
               $unwind(DocField(JoinHandler.RightName)),
               $project(
-                reshape("value" ->
-                  $cond(
-                    $and(
-                      $lte($literal(Bson.Doc()), $field(JoinDir.Right.name)),
-                      $lt($field(JoinDir.Right.name), $literal(Bson.Arr()))),
-                    $field(JoinDir.Right.name, "city"),
-                    $literal(Bson.Undefined))),
```
alissapajer: We don't need this typechecking anymore because we check that the right-hand side of the join (`zips2`) is an object in the join condition, and don't need to check that again.

it/src/main/resources/tests/joins/nonJsJoinCondition.test
```diff
@@ -10,7 +10,6 @@
         "spark_local": "pending",
         "spark_hdfs": "pending"
     },
-    "NB": "Skipped for all connectors due to LogicalPlan/QScript self-join bug #1539.",
```
sellout: So, good that this isn’t the problem anymore, but what is now?

it/src/main/resources/tests/joins/selfjoinConstantOnLeft.test
```diff
@@ -1,17 +1,13 @@
...
-
     "backends": {
         "couchbase":         "skip",
         "marklogic_json":    "skip",
         "marklogic_xml":     "skip",
         "mongodb_read_only": "pending",
         "mongodb_q_3_2":     "pending",
-        "postgresql":        "pending",
-        "spark_local":       "skip",
-        "spark_hdfs":        "skip"
```
sellout: At least with these, we have Spark working, but I wonder why not on the previous one, and what’s in the way of the other connectors.

it/src/main/resources/tests/skipCount.test
```diff
@@ -7,10 +7,17 @@
...
+        "spark_hdfs":     "skip",
+        "spark_local":    "skip"
     },
 
+    "NB": "Skipped in spark due to ListMap serialization stack overflow, where
+           the ListMap is used in Data.Obj and is contained in a QScript UnshiftMap.
+
+           QScript can (and needs to) remove the UnshiftMap as it is unnecessary.
+           And spark needs to support UnshiftMap, which it will when it moves from
+           Data to EJson.",
```
sellout: This should mostly be in an issue, with a short comment and the issue number.

And is this the same reason the other tests switched to "skip"? If so, they should have the same issue reference.

alissapajer:
@sellout sounds good, I know what's wrong with the integration tests that still don't pass, so I'll add comments to those accordingly!

it/src/main/resources/tests/joins/selfjoinConstantOnLeft.test
alissapajer: ML join tests are disabled because they timeout on large(ish) datasets. We have a handful of ML join tests enabled for the tests that run on small datasets. If there are other underlying ML issues, we have been unable to detect them because the tests simply time out.

CB currently only supports key-joins, whose tests live in `it/src/main/resources/tests/couchbase/`. As such, all other joins are disabled for CB.

Mongoq does not yet support joins on master.

The remaining spark join tests are disabled for a couple issues which I'll add comments about in the test themselves, referencing github issues.

it/src/main/resources/tests/joins/nonJsJoinCondition.test
alissapajer: #2110

alissapajer:
@sellout ready for another look!

alissapajer:
@wemrysi your #2058 beat this in by seconds and now there is a merge conflict! :) :)

In any case, I'll push the conflict-resolution change, and then this should be good to go!

alissapajer:
@wemrysi this should be good to go!

PR #2099, opened 2017-04-26
Fix lack of URL decode bug

rabbitonweb:
That was ... interesting bug to catch. Because of it spark-hdfs acted like spark-local


sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/package.scala
```diff
@@ -175,7 +175,7 @@ package object hdfs {
     def fileSystem: Task[HdfsFileSystem] = Task.delay {
       val conf = new Configuration()
       conf.setBoolean("fs.hdfs.impl.disable.cache", true)
-      HdfsFileSystem.get(new URI(sparkFsConf.hdfsUriStr), conf)
```
drostron: What was the offending URI? It could be useful to have a test for this case.

rabbitonweb: it came with this uncoded format `http%3e%e3%fdsfdslocalhost%ds$fd`

what kind of test u thinking about?

drostron: Something to verify that a URI of that format decodes as expected. Further, was the error properly captured before or did it fail silently? If it was silent, it could be useful to detect and fail the Task.

drostron: I should clarify, either a test or confidence that it won't fail silently is sufficient.

rabbitonweb: The problem is that if parameter is not properly decoded, Hadoop API goes silently to this mode "local file system". To tell the long story short - no exceptions, no info.

I will think of something

rabbitonweb: ok, test added

rabbitonweb:
To sum up: if you provide invalid URI when constructing HDFS file system, it will create an instance that points to local file system. 
We prevent that with an error now.

rabbitonweb:
Interesting discovery while solving this issue:

something I was not aware of but Hadoop API supports out of the box Amazons’s S3 & FTP apparently
I still need to double check this, but I thought this might be worth sharing here
that would mean that with small changes to the UI we could query S3 as well

rabbitonweb:
@drostron this is ready for review

jdegoes:
@rabbitonweb That'd be nice, we should get that worked up & spec'd out.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/hdfs/HdfsPackageSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.spark.SparkConf
+import java.net.URI
+import pathy._, Path._
+import scalaz._, concurrent.Task
+
+class HdfsPackageSpec extends quasar.Qspec {
+
+  "hdfs.generateHdfsFS" should {
+    "create a valid HDFS file system if URL is valid" in {
```
drostron: Thanks for adding a test. If I understand the bug correctly, it would be helpful to test the offending style URI as well. Further, is that why `URLDecoder.decode` was added?

rabbitonweb: well `blalbalba` is more general case I should say :) but I can add encoded URL as well, true.

rabbitonweb: No, I take that back. That test is also needed!

rabbitonweb:
@drostron rdy

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/hdfs/HdfsPackageSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+      val url = "hdfs://localhost:9000"
+      hdfsUri(url).unsafePerformSync must_== new URI(url)
+    }
+
+    "create a valid HDFS file system if URL is valid but encoded" in {
+      val url = "hdfs%3A%2F%2Flocalhost%3A9000"
+      hdfsUri(url).unsafePerformSync must_== new URI(URLDecoder.decode(url, "UTF-8"))
```
drostron: This is almost testing the implementation with some of the implementation. What do you think about renaming `url` to `urlEncoded` and pulling it and the previous test's `url` out a level for the following.
```
hdfsUri(urlEncoded).map(_.toString).unsafePerformSync must_= url
```

rabbitonweb: @drostron done + i've added `Show[Throwable]`

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/hdfs/HdfsPackageSpec.scala
```diff
@@ -19,26 +19,33 @@ package quasar.physical.sparkcore.fs.hdfs
...
 import org.apache.spark.SparkConf
-import java.net.{URI, URLDecoder}
+import java.net.URI
 import pathy._, Path._
-import scalaz._, concurrent.Task
+import scalaz._, Scalaz._, concurrent.Task
 
 class HdfsPackageSpec extends quasar.Qspec {
 
+  // TODO move to contrib/show.scala ?
```
drostron: I think the inline `getMessage` was fine. There are times one might want to show more than just the message for a `Throwable`, so I'd say just keep this here or remove it.

rabbitonweb: @drostron I've removed typeclass 
RDY

PR #2101, opened 2017-04-26
PostgreSQL version requirement

drostron:
9.5+ required for `CREATE INDEX IF NOT EXISTS ...`


drostron:
@wemrysi Correct, just adding clarity on the PostgreSQL version that MetaStore requires.

PR #2102, opened 2017-04-26
Change signature of QueryFile.Input to return Free instead of Task

rabbitonweb:
Input methods should be more generic for the spark based connectors to 'inject' different effects to it as needed. This changes is needed for `spark-cassandra` and `spark-elastic`

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/queryfile.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/queryfile.scala
@@ -24,6 +24,7 @@ import quasar.fs.FileSystemError
 import quasar.fs.FileSystemErrT
 import quasar.fs.FileSystemError._
 import quasar.fs.PathError._
+import quasar.fp.free._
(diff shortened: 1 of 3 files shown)


alissapajer:
@rabbitonweb is this waiting for anything more? If not it needs an "assignee" to be merged.

alissapajer:
@rabbitonweb sorry, I was trying to say an assignee other than yourself. We have been using "Reviewers" as the people who will code review the PR and "Assignees" as the person (different than the PR author) who is in charge of merging it after the reviewers have approved it.

I don't think we added this information to our wiki/process docs after github added the "Reviewers" field, so I'll look into that.

rabbitonweb:
@jedesah i was not aware, added

jedesah:
Oops. My bad forgot to merge after I approved.

jedesah:
@rabbitonweb For future reference, you should add a semantic version label and then I can make sure that I agree with your label before I merge. In this case it seems pretty obvious that it's a revision so I will just go ahead and add it myself before merging.

PR #2111, opened 2017-04-27
Disallow equal names in MetadataServiceSpec.

alissapajer:
Fixes #2107.

PR #2115, opened 2017-04-28
Add documentation for module mounts

jedesah:


alissapajer:
Restarted failed travis job (even though this is just a doc change). Error was the one fixed in #2111.

jedesah:
Just realized it's a little unreasonable to expect @beckyconning to merge this herself considering we use a script for merging. @sellout Can you merge this if @beckyconning signs off on it actually being clear and useful?

jedesah:
@beckyconning I know you are probably sure busy these days but if you could just give a :+1: on this PR or mention something that could be improved

README.md
```diff
@@ -294,6 +294,22 @@ For example, given the above MongoDB mount, an additional view could be defined
...
+The value of a module mount is simply the SQL string which will be parsed into a collection of SQL Statements.
+
+To create a new module one would send a json blob similar to this one to the mount endpoint:
+
+```json
+{ "module": "CREATE FUNCTION ARRAY_LENGTH(:foo) BEGIN COUNT(:foo[_]) END; CREATE FUNCTION USER_DATA(:user_id) BEGIN SELECT * FROM `/root/path/data/` WHERE user_id = :user_id END" }
+```
+
+Similar to views, modules can be mounted at any directory path. If a module's path is nested inside the path of a database mount, it will appear alongside the other directory and files in the database. A module will "shadow" any actual directory that would otherwise be mapped to the same path. Any attempt to write data to a module will result in an error.
+
```
natefaubion: Are function names like identifiers (case sensitive), or are they like builtins (case insensitive)?

jedesah: Like builtins (case insensitive). Will add a line with that information.

natefaubion: In SQL it is insensitive, but in my testing, the URL is case sensitive. It would be worth clarifying that.

sellout:
@beckyconning Just a pink so you can indicate whether this PR meets your needs.

README.md
jedesah: Addressed this issue [here](https://github.com/quasar-analytics/quasar-docs/pull/2) with a reference in this README.

jedesah:
@sellout Actually I switched over responsibility to @natefaubion for critiquing this PR which he did and I've made for updates to reflect them. So hopefully @natefaubion you can have another look at it.

I'll update the documentation on `/invoke` in the PR that add support for `offset` and `limit`.

sellout:
@natefaubion Ping.

alissapajer:
The build is now green for this. @natefaubion is this ready for merge?

jedesah:
@alissapajer I think @natefaubion already approved. I think this is just waiting for @sellout to merge

sellout:
Ah, as usual – need to add comments when you do stuff on GitHub, because very few things actually trigger a notification.

PR #2125, opened 2017-05-01
Schema discovery service

### LINKED ISSUE #1715: Statistical Structural Types: Phase A
**Statistical Structural Types**. Having a performant, responsive means for the front-end to obtain increasingly accurate statistical information about the structural characteristics of a data set will enable much improved user-experiences, and improve the fidelity of front-end experiences that currently rely on a (non-random) sample of a small number of values from a data set.

Statistical structural types describe, in a structured and statistical representation isomorphic to SQL2's pattern language, an underlying data set in a compressed way.

1. **Representation of Types**. We need a serialization format for annotated types (probably EJSON-based, unless there's a compelling reason otherwise). 
2. **Augmented Type System**. The existing type system is probably sufficient for a "v1", but may need small augmentation — for example, the ability to describe a map whose keys all have the same structural type.
3. **Compression Algorithm**. We need an ANALYZE function which goes over a data set and describes the type of all data seen, in a lossy way. There is no single way to perform compression. We should investigate different heuristics for performing the compression.

In theory, something like `Cofree` could be used over a fixed point representation of types in order to provide statistical annotations without requiring a parallel hierarchy.

**Tasks**
- [x] Fixpoint representation of LogicalPlan types (#1944)
- [x] Analysis that builds a structural type from a dataset annotated with statistical information
- [x] Lossy Compression algorithm for SST providing control over how much information is discarded (#2022)
- [x] Means of obtaining a sample of a dataset at a given path
- [x] `EJson` serialization format for types
- [x] HTTP endpoint returning SST for the dataset at the given path

### LINKED ISSUE #2022: Lossy Structural Type Compression Algorithm
In order to provide a summary of the structure of a dataset in a space-efficient way, we need to compress the information extracted.

We'd also like to be able to tune how this compression is performed via parameters to give users control over how much information is discarded, probably via specifying space bounds on the resulting structural type.

- [x] Stats annotation for SST
- [x] Entropy/information content function for SST
- [x] Long string compression (compress strings longer than a threshold to `char[]`)
- [x] Binary compression (compress binary values to `byte[]`)
- [x] Union value compression, maybe with threshold or top-k preservation
- [x] Map key compression

wemrysi:
Implements a schema discovery service, allowing users to obtain a statistical structural type (SST) for a dataset that summarizes the structure of the data, annotated with frequency, bounds and distribution information where applicable.

Both an HTTP endpoint, `/schema/fs`, and a `schema` repl command are provided and the HTTP endpoint allows for some control over the amount of information preserved in the resulting SST.

Closes #1715.
Closes #2022.

### DIFF
--- a/ejson/src/main/scala/quasar/contrib/argonaut/package.scala
+++ b/ejson/src/main/scala/quasar/contrib/argonaut/package.scala
@@ -25,40 +25,35 @@ import _root_.matryoshka._
 import _root_.scalaz._
 
 package object argonaut {
-  implicit def jsonRecursive[J[_]]
-    (implicit O: ejson.Obj :<: J, C: ejson.Common :<: J)
-      : Recursive.Aux[Json, J] =
+  private val CJ = Inject[ejson.Common, ejson.Json]
+  private val OJ = Inject[ejson.Obj,    ejson.Json]
+
+  implicit def jsonRecursive: Recursive.Aux[Json, ejson.Json] =
     new Recursive[Json] {
-      type Base[A] = J[A]
+      type Base[A] = ejson.Json[A]
 
       def project(t: Json)(implicit BF: Functor[Base]) =
         t.fold(
-          C.inj(ejson.Null()),
-          b => C.inj(ejson.Bool(b)),
-          d => C.inj(ejson.Dec(d.toBigDecimal)),
-          s => C.inj(ejson.Str(s)),
-          a => C.inj(ejson.Arr(a)),
-          o => O.inj(ejson.Obj(ListMap(o.toList: _*))))
+          CJ(ejson.Null()),
+          b => CJ(ejson.Bool(b)),
+          d => CJ(ejson.Dec(d.toBigDecimal)),
+          s => CJ(ejson.Str(s)),
+          a => CJ(ejson.Arr(a)),
+          o => OJ(ejson.Obj(ListMap(o.toList: _*))))
     }
 
-  // TODO: It would be nice to define this in terms of the components as well,
-  //       but then I don’t know how to make it total. So, for now, we use the
-  //       predefined Coproduct.
   implicit def jsonCorecursive: Corecursive.Aux[Json, ejson.Json] =
     new Corecursive[Json] {
       type Base[A] = ejson.Json[A]
 
-      val EC = Inject[ejson.Common, ejson.Json]
-      val EO = Inject[ejson.Obj,    ejson.Json]
-
       def embed(ft: ejson.Json[Json])(implicit BF: Functor[Base]) =
         ft match {
-          case EC(ejson.Null())  => jNull
-          case EC(ejson.Bool(b)) => jBool(b)
-          case EC(ejson.Dec(d))  => jNumber(d)
-          case EC(ejson.Str(s))  => jString(s)
-          case EC(ejson.Arr(a))  => jArray(a)
-          case EO(ejson.Obj(o))  => jObject(JsonObject.fromTraversableOnce(o))
+          case CJ(ejson.Null())  => jNull
+          case CJ(ejson.Bool(b)) => jBool(b)
+          case CJ(ejson.Dec(d))  => jNumber(d)
+          case CJ(ejson.Str(s))  => jString(s)
+          case CJ(ejson.Arr(a))  => jArray(a)
+          case OJ(ejson.Obj(o))  => jObject(JsonObject.fromTraversableOnce(o))
         }
     }
 }
--- a/foundation/src/test/scala/quasar/fp/numeric/SampleStatsSpec.scala
+++ b/foundation/src/test/scala/quasar/fp/numeric/SampleStatsSpec.scala
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.fp.numeric
+
+import slamdata.Predef._
+import quasar.contrib.algebra._
+
+import org.scalacheck.{Arbitrary, Gen}, Arbitrary.arbitrary
+import org.specs2.scalaz._
+import scalaz.scalacheck.{ScalazProperties => propz}
+import scalaz.{NonEmptyList, Show}
+import scalaz.std.option._
+import scalaz.syntax.foldable1._
+import scalaz.syntax.equal._
+import scalaz.syntax.std.boolean._
+import spire.laws.arb._
+import spire.math.Real
+
+final class SampleStatsSpec extends Spec with ScalazMatchers with SampleStatsArbitrary {
+  implicit val showReal: Show[Real] =
+    Show.showFromToString
+
+  /** A list of moderately sized reals, to ensure they don't grow so big that
+    * computation gets prohibitively slow.
+    */
+  case class ModerateReals(rs: NonEmptyList[Real])
+
+  object ModerateReals {
+    implicit val arbitraryModerateReals: Arbitrary[ModerateReals] =
+      Arbitrary(for {
+        h <- arbitrary[Double]
+        t <- Gen.listOf(Gen.choose(-100000000.0, 100000000.0))
+      } yield ModerateReals(NonEmptyList(Real(h), t.map(Real(_)): _*)))
+  }
+
+  def μ(xs: NonEmptyList[Real]): Real =
+    xs.foldLeft1(_ + _) / Real(xs.length)
+
+  // The k-th moment about the mean.
+  def `mₖ`(k: Int, xs: NonEmptyList[Real]): Real = {
+    val  n   = Real(xs.length)
+    val `m₁` = μ(xs)
+    xs.map(x => (x - `m₁`) ** k).foldLeft1(_ + _) / n
+  }
+
+  checkAll(propz.equal.laws[SampleStats[Real]])
+  checkAll(propz.monoid.laws[SampleStats[Real]])
+
+  "commutative monoid" >> {
+    "merge is commutative" >> prop { (x: SampleStats[Real], y: SampleStats[Real]) =>
+      (x + y) must equal(y + x)
+    }
+  }
+
+  "correctness" >> {
+    "mean" >> prop { mr: ModerateReals =>
+      val xs = mr.rs
+      val ss = SampleStats.fromFoldable(xs)
+
+      ss.mean must equal(μ(xs))
+    }
+
+    // https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Na.C3.AFve_algorithm
+    "variance" >> prop { mr: ModerateReals =>
+      val xs   = mr.rs
+      val ss   = SampleStats.fromFoldable(xs)
+      val n    = Real(xs.length)
+      val ssq  = xs.map(x => x * x).foldLeft1(_ + _)
+      val sum  = xs.foldLeft1(_ + _)
+      val `σ²` = (ssq - ((sum * sum) / n)) / n
+
+      ss.variance must beSome(equal(`σ²`))
+    }
+
+    // https://en.wikipedia.org/wiki/Skewness#Pearson.27s_moment_coefficient_of_skewness
+    "skewness" >> prop { mr: ModerateReals =>
+      val xs   = mr.rs
+      val ss   = SampleStats.fromFoldable(xs)
+      val `m₂` = `mₖ`(2, xs)
+      val `m₃` = `mₖ`(3, xs)
+      val  sk  = (`m₂` ≠ Real(0)).option(`m₃` /  `m₂`.fpow(1.5))
+
+      ss.skewness must equal(sk)
+    }
+
+    // https://en.wikipedia.org/wiki/Kurtosis#Sample_kurtosis
+    "kurtosis" >> prop { mr: ModerateReals =>
+      val  xs  = mr.rs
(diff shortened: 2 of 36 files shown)


wemrysi:
I'm writing docs for the new endpoint as well as our `Json` embedding of `EJson`, so expect another commit for that shortly.

wemrysi:
After reviewing the docs, I think I'm going to change how the SST results are serialized, and continue to use the existing `Data` codecs for now. The endpoint is currently serializing to JSON-embedded EJSON, but I realized this is going to be the only endpoint using this format.

A more user-friendly path forward would be to support the existing json modes for now and then add `mode=ejson` as an additionally supported mode and deprecate `precise`, or maybe just have `precise` _be_ EJSON? In any case, how to migrate to serialized EJSON is an api-wide issue and deserves some discussion outside the scope of this feature.

jdegoes:
ejson needs to be a separate content type, not a mode of a JSON content type.

sellout:
@wemrysi Yes, JSON-embedded EJson should be `mode=ejson`. `mode=precise` will be deprecated (as it doesn’t encode all of our data). I don’t know the fate of `mode=readable`, but I think readable is something we can modify, and maybe use actual EJson for instead of JSON, since it’s meant for human consumption, and not for re-parsing.

sellout:
@jdegoes Yeah, EJson is a separate content type, but there is also EJson encoded as valid JSON, which is `mode=ejson`. And we’ve discussed this with the frontend, as they think they’ll consume `mode=ejson` to take advantage of the browser’s optimized JSON parsing, then convert that AST to EJson.

wemrysi:
@jdegoes EJSON using its own syntax will be a separate content-type, but we also have an embedding of EJSON in JSON for the likely case where browsers and maybe other contexts (`JavaScript`?) have far more performant means of parsing JSON.

jdegoes:
@wemrysi Yes, of course, EJSON embedded inside JSON is a mode of the JSON content type and not a separate content type.

frontend/src/test/scala/quasar/sst/StructuralTypeSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  type J = Fix[EJson]
+  type S = StructuralType[J, Int]
+
+  def mkST(j: J): S = StructuralType.fromEJsonK[J](1, j)
+
+  checkAll(propz.equal.laws[S])
+  checkAll(propz.monoid.laws[S])
+  checkAll(propz.traverse1.laws[StructuralType[J, ?]])
+  // FIXME: Need Cogen
```
djspiewak: Why not define a `Cogen`?

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+    "no path given" >> {
+      service(InMemState.empty)(Request())
+        .flatMap(_.as[ApiError])
+        .map(_ must beApiErrorWithMessage(
+          BadRequest withReason "File path expected.",
+          "path" := "/"))
+        .unsafePerformSync
+    }
+
+    "directory path given" >> prop { dir: ADir =>
```
alissapajer: I tested the `schema` command out in the REPL; very cool! It returns as expected when I run `💪 $ schema zips`.

Based on this web test, I was thinking that the `schema` REPL command would fail when I provided a directory, but instead it succeeds, listing out each level of directory as `"type" : "map"` until it reaches data, and then showing the schema for the data. I don't think it's a problem to have this behavior in the REPL, but I just wanted to point out that it was different than the behavior in web, in case that was unexpected?

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+    } : _*))
+  }
+
+  "successful response" >> {
+    def shouldSucceed(file: AFile, x: Data, y: Data, cfg: analysis.CompressionSettings, req: Request) = {
+      val dataset = Vector.fill(5)(x) ++ Vector.fill(5)(y)
+
+      service(stateForFile(file, dataset))(req)
+        .flatMap(_.as[Json])
+        .map(_ must_= sstResponse(dataset, cfg))
```
alissapajer: I'm not convinced that this is testing deeply enough. The impls of `service` and `sstResponse` are very similar (both calling to `extractSchema`). So I suppose these tests are testing that all the machinery involved in `service` is working correctly, but not testing that the schema generation itself is correct?

I might have missed it in the diff, but do we additionally want a more "integration" style test for sst that will verify the correctness of the entire json-structured output?

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
sellout: So, that REPL behavior sounds correct (and awesome).

I haven’t reviewed the code, so I don’t know why `web` behaves differently, but I know there are a few places in `web` where we use things like the ReadFile algebra directly, when it should really be making a query (because queries have the unified MRA data model, whereas `ReadFile` expects to operate on a store-level file). Maybe related somehow?

ejson/src/main/scala/quasar/contrib/argonaut/package.scala
```diff
@@ -25,40 +25,35 @@ import _root_.matryoshka._
 import _root_.scalaz._
 
 package object argonaut {
-  implicit def jsonRecursive[J[_]]
-    (implicit O: ejson.Obj :<: J, C: ejson.Common :<: J)
-      : Recursive.Aux[Json, J] =
+  private val CJ = Inject[ejson.Common, ejson.Json]
+  private val OJ = Inject[ejson.Obj,    ejson.Json]
+
+  implicit def jsonRecursive: Recursive.Aux[Json, ejson.Json] =
```
wemrysi: I'm not sure, though I tend to view `contrib` things as our extensions to a library, some of which should probably be contributed back. Though, maybe @sellout has some plan for these?

foundation/src/test/scala/quasar/fp/numeric/SampleStatsSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+  checkAll(propz.equal.laws[SampleStats[Real]])
+  checkAll(propz.monoid.laws[SampleStats[Real]])
+
+  "commutative monoid" >> {
+    "merge is commutative" >> prop { (x: SampleStats[Real], y: SampleStats[Real]) =>
+      (x + y) must equal(y + x)
+    }
+  }
+
+  "correctness" >> {
```
wemrysi: Sure, I'll add them.

interface/src/main/scala/quasar/main/analysis.scala
```diff
@@ -0,0 +1,104 @@
...
+  )(implicit
+    JC: Corecursive.Aux[J, EJson],
+    JR: Recursive.Aux[J, EJson]
+  ): Process1[Data, SST[J, A]] = {
+    val structuralTrans =
+      compression.coalesceKeys[J, A](settings.mapMaxSize)      >>>
+      compression.z85EncodedBinary[J, A]                       >>>
+      compression.limitStrings[J, A](settings.stringMaxLength) >>>
+      compression.coalescePrimary[J, A]                        >>>
+      compression.narrowUnion[J, A](settings.unionMaxSize)
```
wemrysi: Yeah, I'll add it and a corresponding config option.

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
wemrysi: @alissapajer did you provide a path that looks like a directory, i.e. `/foo/bar/baz/` or `/foo/bar/baz`? This is more about the parsing of the path representation than what the virtual path maps to in the physical filesystem.

alissapajer: @wemrysi I ran `💪 $ schema quasar-test` where the subsequent path looks like `quasar-test/run_b7c55193/regression/zips`.

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
wemrysi: Yeah, the intent here wasn't to test what schema generation generates, just that the service machiner.

Adding a test of `extractSchema` itself that tests all the compression working together is a good idea, I'll add it.

ejson/src/main/scala/quasar/contrib/argonaut/package.scala
alissapajer: Right, I guess it's viewing contrib as something we definitely want to (or at least could) contribute to the library versus as an extension of the library, which we might never actually contribute.

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
wemrysi: Ok, you should be able to provide that same path via the HTTP interface and get the same result.

alissapajer: Hm, so when I try this one level higher, it doesn't work:
```
💪 $ schema quasar-test/run_c9443542 
Quasar error: None of these directories contain any files to read from: /spark_local_test/quasar-test/run_c9443542/
```
I tried it with ticks and back-ticks, but still no go. I would think this should work too?

frontend/src/main/scala/quasar/sst/StructuralType.scala
```diff
@@ -0,0 +1,269 @@
...
+
+    def mergeUnk(xu: Option[TT], yu: Option[TT]): Option[(TT, TT)] =
+      (xu, yu) match {
+        case (Some((xk, xv)), Some((yk, yv))) => some(((xk, yk), (xv, yv)))
+        case (Some((xk, xv)),           None) => some(((xk,  ⊥), (xv,  ⊥)))
+        case (          None, Some((yk, yv))) => some(((⊥ , yk), (⊥ , yv)))
+        case (          None,           None) => none
+      }
+
+    def mergable(x: T, y: T): Boolean =
```
alissapajer: I think this should be spelled ["mergeable"](https://en.wiktionary.org/wiki/mergeable).

frontend/src/test/scala/quasar/sst/StructuralTypeSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+  //checkAll(propz.comonad.laws[StructuralType[J, ?]])
+
+  "structural monoid" >> {
+    "accumulates measure for identical structure" >> prop { (ejs: J, k0: Int) =>
+      val k = scala.math.abs(k0 % 100) + 1
+      val st = mkST(ejs)
+      st.multiply(k) must equal(st as k)
+    }
+
+    "unmergable values accumulate via union" >> prop { (x: J, y: J) =>
```
alissapajer: s/unmergable/unmergeable/

frontend/src/test/scala/quasar/sst/StructuralTypeSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+    "merging known maps where some keys differ results in combination of keys" >> prop { m0: IMap[J, S] =>
+      m0.minViewWithKey map { case ((k, v), m) =>
+        val m1 = envT(1, TypeF.map[J, S](IMap.singleton(k, v), None)).embed
+        val m2 = envT(1, TypeF.map[J, S](m, None)).embed
+        val m3 = envT(2, TypeF.map[J, S](m0, None)).embed
+        (m1 |+| m2) must equal(m3)
+      } getOrElse ok
+    }
+
+    "unions merge by accumulating mergable values" >> prop { (xs: IList[S], ys: IList[S], i: BigInt, s: String, st: SimpleType) =>
```
alissapajer: s/mergable/mergeable/

web/src/test/scala/quasar/api/services/analyze/SchemaServiceSpec.scala
wemrysi: Hm, I'm not sure. Does a `select` query on that path work? The schema is generated from the resuls of `Invoke(Sample, Read(<path>), 1000)`.

alissapajer: `select * from 'quasar-test/run_c9443542'` (with back-ticks instead of regular ticks) gets the same error as above.

`select * from 'quasar-test'` returns results where the columns look like `run_29511346.regression.extraSmallZips.city`. So that seems consistent with the behavior of `schema`.

alissapajer:
@wemrysi I'm thinking about the best place to put the new sst files that are currently in the `frontend` project in `package quasar.sst`. Given that they depend on `Type` which is defined in `frontend`, this is a reasonable place.

A second option would be to create a new `sst` project. This would depend on `frontend` and would be depended on by `interface` (I think?). Assuming that's correct, this project structure would prevent `core`, `connector`, and all the backends from depending on `sst`, which is fine if they don't need it.

What do you think? Any downsides to making a new `sst` project?

If we're able to pull out `Type` in a sane way, and have access to the dependencies in `foundation`, then I do think down the road the sst project could be a candidate for having its own repo. Schema discovery with compression seems like something that could be its own library that other people might want to use!

frontend/src/test/scala/quasar/sst/StructuralTypeSpec.scala
wemrysi: We need the `Cogen` machinery from matryoshka which we can't use yet as we're awaiting a release of the ScalaCheck `1.13.x` branch that contains rickynils/scalacheck#316.

alissapajer:
@wemrysi let me know when you've pushed everything that you plan to, so I can take a look again!

wemrysi:
@alissapajer Will do! Just finishing up a couple more tests and the docs.

wemrysi:
@alissapajer Alright, everything is pushed, just waiting on Travis to build.

wemrysi:
@alissapajer Just realized I never responded to your organizational comment regarding an `sst` project. I think creating said project would be fine and yeah, if we can tease apart the quasar-specific dependencies, I think making this into its own repo would also be reasonable.

I just noticed you merged (thanks!), so maybe we revisit this as part of the polyrepo efforts?

alissapajer:
@wemrysi sounds good. I figured it was important to get this in for testing ASAP and then we can revisit organization later.

PR #2132, opened 2017-05-02
Type `Count(Const)` as `Const(1)`.

### LINKED ISSUE #2131: Compiling doesn't fail in some cases where they should
If you compile the query:

``` sql
SELECT "Hello world"
```

On the MongoDB backend, it doesn't complain about anything. If a view mount is created with that query, attempting to read from the view gives:

``` json
{
  "error": {
    "status": "Failed to execute SQL^2 query.",
    "detail": {
      "message": "Unable to insert data into MongoDB: MongoDB is only able to store documents in collections, not `Text(Hello, World!)`.",
      "data": "Text(Hello, World!)",
      "logicalplan": {
        "type": "LogicalPlan/Invoke",
        "label": "Squash",
        "children": [
          {
            "type": "LogicalPlan/Invoke",
            "label": "MakeObject",
            "children": [
              {
                "type": "LogicalPlan/Constant/Str",
                "label": "\"total\""
              },
              {
                "type": "LogicalPlan/Invoke",
                "label": "Count",
                "children": [
                  {
                    "type": "LogicalPlan/Constant/Str",
                    "label": "\"Hello, World!\""
                  }
                ]
              }
            ]
          }
        ]
      }
    }
  }
}
```

Is there any way we can detect things like this during compilation? It makes error handling in the front end somewhat problematic, as there are some queries we don't get errors for until it's "too late" to provide a good error message.

In theory we could fetch the first result of a view after mounting it or something like that to check this, but if the query is expensive this is really bad.

This kind of problem is quite rare I think, but as we currently provide `SELECT "Hello World";` as the default query, it is right in the user's face (I'll tweak this query to include an `AS value` to avoid this specific instance though).

sellout:
Fixes #2131.

### DIFF
--- a/frontend/src/main/scala/quasar/std/agg.scala
+++ b/frontend/src/main/scala/quasar/std/agg.scala
@@ -41,6 +41,7 @@ trait AggLib extends Library {
     noSimplification,
     partialTyper[nat._1] {
       case Sized(Type.Const(Data.Set(xs))) => Type.Const(Data.Int(xs.length))
+      case Sized(Type.Const(_))            => Type.Const(Data.Int(1))
       case Sized(_)                        => Type.Int
     },
     basicUntyper)
--- a/sql/src/test/scala/quasar/sql/compiler.scala
+++ b/sql/src/test/scala/quasar/sql/compiler.scala
@@ -82,6 +82,12 @@ class CompilerSpec extends quasar.Qspec with CompilerHelpers {
         lpf.constant(Data.Set(List(Data.Str("b"), Data.Str("c")))))
     }
 
+    "compile reduced constant" in {
+      testTypedLogicalPlanCompile(
+        "select count(*) as total from (select \"Hello world\") as lit",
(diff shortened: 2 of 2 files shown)


frontend/src/main/scala/quasar/std/agg.scala
```diff
@@ -41,6 +41,7 @@ trait AggLib extends Library {
     noSimplification,
     partialTyper[nat._1] {
       case Sized(Type.Const(Data.Set(xs))) => Type.Const(Data.Int(xs.length))
+      case Sized(Type.Const(_))            => Type.Const(Data.Int(1))
```
alissapajer: Confirmed that `Data.Set` is the only `Data` that might have a size larger than 1. `Data.Obj` and `Data.Arr` each have a size of 1 (thought of as a single object or array).

sellout:
@alissapajer This is ready for merge.

alissapajer:
@sellout tried to merge, it needs merge from master.

sellout:
Thanks. Updated.

PR #2134, opened 2017-05-02
Removed fallback evaluated test configuration

djspiewak:


alissapajer:
@djspiewak failing with `java.lang.RuntimeException: headers not all present` which makes sense because I haven't pushed fixes for headers yet. So are you okay with merging to mimir branch with a broken build, and we'll only worry about the build starting with the PR of mimir to quasar?

djspiewak:
Yeah I'm good with that. My expectation is that the build will work once your fixes get in, and we'll practice good PR checking hygiene after that.

PR #2140, opened 2017-05-03
fixed code coverage publication script

nicflores:
afterSuccess script fixed to publish codecov data

### DIFF
--- a/scripts/afterSuccess
+++ b/scripts/afterSuccess
@@ -6,9 +6,9 @@ IFS=$'\n\t'       # http://redsymbol.net/articles/unofficial-bash-strict-mode/
 source "$(dirname $0)/constants"
 
 # Only run for the first travis job
-if [[ "${TRAVIS_JOB_NUMBER##*.}" == "1" ]]; then
+if [[ $CONNECTOR == "mongodb_2_6" ]]; then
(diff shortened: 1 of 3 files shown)


scripts/afterSuccess
```diff
@@ -8,6 +8,7 @@ source "$(dirname $0)/constants"
 # Only run for the first travis job
 if [[ "${TRAVIS_JOB_NUMBER##*.}" == "1" ]]; then
   "$SCRIPT_DIR/publishJarIfMaster"
+elif [[ "${TRAVIS_JOB_NUMBER##*.}" == "5" ]]; then
```
drostron: Not introduced in this PR, but using `TRAVIS_JOB_NUMBER`s are a bit fragile. Perhaps we should switch to `CONNECTOR` if sufficient.

drostron:
Needs a merge from master

scripts/afterSuccess
nicflores: Good point. Should we make that switch now?

drostron: Up to you. I'll wait to merge if you want to make the change now.

drostron:
LGTM

PR #2148, opened 2017-05-04
mimir ready for quasar

alissapajer:
This PR:
- deletes `fallback`, `macros`, `benchmark` projects
- renames `com.precog` to `quasar`
- reorganizes files to have a saner structure
- adds slamdata header to all files

When this lands, mimir should be ready for a PR to quasar.


blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -1,15 +1,20 @@
 /*
- * Javolution - Java(TM) Solution for Real-Time and Embedded Systems
```
alissapajer: This looks like something we copy-pasted and as such it should keep its old header. Need to research if we can ignore a file for header generation.

blueeyes/src/main/scala/quasar/precog/common/CPath.scala
```diff
@@ -1,25 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
-package com.precog.common
 
-import blueeyes._, json._, serialization._
+package quasar.precog.common
```
alissapajer: You will notice various files in the `blueeyes` project with `quasar.precog` package. Ideally these could be moved into the `precog` project, but they depend on various blueeyes things and so they live here for now. I image we'll be continually sorting all this out.

yggdrasil/src/main/scala/quasar/yggdrasil/util/CPathUtils.scala
```diff
@@ -1,28 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
-package com.precog.yggdrasil
+
+package quasar.yggdrasil
 package util
```
alissapajer: Also, for example, we need to declare this package as `package quasar.yggdrasil.util` and then import from `quasar.yggdrasil` if necessary. In some places I cleaned this up, but not in all. I'm fine getting this into quasar first, and then cleaning this up incrementally.

yggdrasil/src/main/scala/com/precog/EvaluationContext.scala
```diff
@@ -1,34 +0,0 @@
-/*
```
alissapajer: This file (and a few others in the same path) moved over to `mimir/src/main/scala/quasar/mimir/EvaluationContext.scala` because all the call sites were in mimir. I may have forgotten to do `git mv` and just done regular `mv` with these files (I'm wondering as such because it's not showing up as a rename here but as a deletion).

blueeyes/src/main/java/quasar/precog/BitSet.java
djspiewak: Could try putting the old header after the `package` declaration.  But we might be able to do it with `headerMappings`, since we don't need any other Java files to have headers.

djspiewak:
Build changes are another problem.  I'll fix that before I PR.  LGTM.

PR #2152, opened 2017-05-04
fixed elif typo

nicflores:
Fixed typo in afterSuccess script

PR #2153, opened 2017-05-04
Disable code coverage in travis.

alissapajer:


scripts/afterSuccess
```diff
@@ -8,7 +8,4 @@ source "$(dirname $0)/constants"
 # Only run for the first travis job
 if [[ $CONNECTOR == "mongodb_2_6" ]]; then
   "$SCRIPT_DIR/publishJarIfMaster"
-if [[ $CONNECTOR == "mongodb_3_0" ]]; then
```
alissapajer: This change makes #2152 unnecessary.

scripts/build
```diff
@@ -19,17 +19,17 @@ TRAVIS_MAX_CONNECTOR_COMPILE_MINS=40
...
 
 # if we're building a pull request, DON'T clean and use the cache instead
+# coverage is disabled until we can have coverage in pull requests too
 if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
   CLEAN=clean
-  COVERAGE=coverage
-  COVERAGE_REPORT=coverageReport
+  COVERAGE=#coverage
+  COVERAGE_REPORT=#coverageReport
 else
```
nicflores: I believe lines 25 and 26 will get injected into the sbt command as:

`sbt -DisIsolatedEnv=false ++0.13 checkHeaders #coverage test exclusive:test #coverageReport`

running a little test like

` sbt #test couchbase/compile`

doesn't run the `couchbase/compile`

maybe change 25 and 26 to
`#COVERAGE=coverage`
`#COVERAGE_REPORT=coverageReport`

?

alissapajer: Good catch, just the verified behavior.

I'll make the change.

alissapajer: So, in my test example if I put a space after the var name, it treats everything after that as a comment. So like:
```
COVERAGE= #coverage
COVERAGE_REPORT= #coverageReport
```
@nicflores do you see any reason why that wouldn't work?

My test example was:
```
#!/usr/bin/env bash

FOO= #14
echo $FOO
```

drostron:
@alissapajer Needs a semantic ver label

alissapajer:
@drostron added, thanks.

PR #2162, opened 2017-05-05
Remove temporal coercion from MarkLogic relation operators

wemrysi:
Previously we were handling temporal coercion in relation operators due to some tests that were comparing dates and timestamps. They turned out to be incorrect and were corrected in #1843 and thus the behavior is no longer needed in MarkLogic.


wemrysi:
Looks like I missed a test case, will remedy shortly.

PR #2178, opened 2017-05-10
Add support for imports in Sql2 queries

### LINKED ISSUE #1792: Add SQL2 Functions — AKA Parameterized Views (DRAFT)
Currently, views may contain holes (variables), but these must be fully specified at creation time. Any attempt to update the values in the holes results in updating the view. This usage workflow is good for some purposes but wholly inadequate for interactive dashboards designed to be used by lots of users.

To address these problems, we need to introduce parameterized views, which are views that contain holes whose values are not specified until the result set is read from Quasar. This will allow front-end applications such as SlamData to store the parameterized views only, and then on a per-user basis, specify values for holes.

In addition, this opens the door to more advanced forms of caching, such as cubing.

## SQL<sup>2</sup> Syntax

A parameterized view, AKA a *process*, is a function from a one or more named values to a result set.

```sql
CREATE FUNCTION FLOOR(:num)
BEGIN
  :num - (:num % 1)
END;

CREATE FUNCTION ROUND(:num)
BEGIN
  FLOOR(:num + 0.5)
END;
```

An ordinary view emerges as a special case of an immediately applied function:

```sql
CREATE FUNCTION ANON123(:source)
BEGIN
  SELECT first_name, last_name FROM :source
END;

ANON123(`/foo/bar/baz`)
```

There are two top-level grammers we will need to support:

 * Mixing let bindings with function declarations followed by a terminal expression, which allows users to define and use their own functions inside their own queries;
 * A pure module that consists only of function declarations and possibly let bindings.

## File System: Module Nodes / Mounts

Parameterized views could piggyback off view mounts or new type of node in the file system (*module nodes*, which contain SQL<sup>2</sup> functions that accept input and return output). As with mounts, reading from module and editing their configuration are separate, until the two have been unified via metadata or other mechanism.

A module is akin to a directory of named executables (each defined by a SQL<sup>2</sup>), whereas a data node is akin to a data file, and a directory node is the same across the Quasar file system and ordinary file systems.

If the SQL<sup>2</sup> defining a module defines multiple functions, then the module node contains multiple named executables. In this sense, a module groups together a bunch of functions under a common namespace.

## Process API

Module may need a new API to retrieve results from, such as `/process/fs/<location>/<name>` (alternatively, the existing GET API can be returned but variables can be required to have valued specified in the request). A `POST` should be required to run a process inside a module, and the body of the post must be JSON / EJSON which specifies values for all named parameters. For example:

```json
{"num": 2.3}
```

## Metadata API

Retrieving metadata on a module should specify the names of all functions (in the directory listing of the mount?) and the number and types of parameters required by all the functions. This will enable front-ends built on top the API to prompt for appropriate inputs to each of the functions.

## Imports

No doubt, one of the first things users will want (after they can create functions) is to be able to import them across different queries. We should think about whether to define that here or defer to later. At least for the initial bare bones implementation, users could copy/paste definitions in each SQL file, which seems like a reasonable if painful workaround.

```sql
import `/foo/bar/baz/`;
```

## Caching

**Notes on Caching**: Not intended to be implemented in 4.2.

Caching a parameterized view is equivalent to computing all possible values for all parameters, and running the query for each value. In practice, using `GROUP BY`, the process can be collapsed into a single query.

Caching requires the determination of the possible values for all parameters (which can yield a non-empty result set). This is an algebraic problem that may or may not have a solution depending on the query.

For example, the following function may not be cached:

```sql
CREATE FUNCTION ARRAY_LENGTH(:foo)
BEGIN
  COUNT(:foo[_])
END
```

The reason the function may not be cached is because the number of solutions for `:foo` which produce a non-empty result set is infinite.

However, the following function may be cached:

```sql
CREATE FUNCTION USER_DATA(:user_id)
BEGIN
  SELECT * FROM `/root/path/data/` WHERE user_id = :user_id
END
```

In this case, the number of solutions for `:user_id` which produce a non-empty result set is finite, and can be determined by distincting the `user_id` field in the source data set.

Such a function may be cached efficiently through generation, execution, and indexed storage of the following query:

```sql
SELECT{"variables": {"user_id": user_id}, "result_set": [* ...]} 
  FROM `/root/path/data/` 
  GROUP BY user_id
```

If such a result set has an index on `variables.user_id`, or if the result set is stored as a map from variables to results, then lookups / evaluation of the function become `O(1)`.

Note that even partial solutions for a parameterized view can accelerate results. But if no part of a parameterized view can be cached, then acceleration is going to require more traditional techniques (storage in memory, indexing on relevant fields, incremental computation, etc.).

## Subtasks

- [x] Extend SQL AST and enhance compiler to apply SQL functions
- [x] Extend SQL parser to support user-defined functions and imports
- [x] Extend SQL parser to parse modules (i.e. lets+functions without terminal expressions)
- [x] Introduce module (mount?) in file system
- [x] Support file system operations (metadata, etc.) for module type
- [x] Enhance metadata API for module (mount?) & functions in module
- [x] Introduce evaluation API for functions in module (or reuse GET API but require params)
- [x] Update permission system for module (if necessary)
- [ ] Implement support for imported modules (final task)

jedesah:
Closes #1792.


jedesah:
To be clear, the limitations aren't some fundamental problem with the design. Just require a more sophisticated implementation of `resolveImports`.

A few approaches would be possible. Straight up inline everything at the call site. Another one would be to keep better track of what has been imported already to address limitation #2 and to add a prefix to the function definition names (and call sites) to address limitation #1.

Despite the fact that these are relatively serious limitations for a programming language supporting imports, I thought it was still useful to open a PR since our users are likely to start off with some relatively simple import schemes at first so at least we can support that.

I can then discuss the best approach to addressing these limitations and submit a subsequent PR to remove them.

core/src/main/scala/quasar/package.scala
```diff
@@ -78,13 +83,35 @@ package object quasar {
...
     }
 
+  // It would be nice if this were in the sql package but that is not possible right now because
+  // Mounting is defined in core
+  def resolveImports[S[_]](blob: Blob[Fix[Sql]], basePath: ADir)(implicit
+    mount: Mounting.Ops[S],
+    fsFail: Failure.Ops[FileSystemError, S]
+  ): Free[S, Block[Fix[Sql]]] = {
+    def resolvePath(path: DPath, base: ADir): ADir = refineTypeAbs(path).fold(ι, base </> _)
+    def retrieveDeclarations(module: ADir, allReadyImported: Set[ADir]): Free[S, List[FunctionDecl[Fix[Sql]]]] = {
```
alissapajer: s/allReadyImported/alreadyImported/

core/src/main/scala/quasar/package.scala
```diff
@@ -78,13 +83,35 @@ package object quasar {
       case _                           => lp.right
     }
 
+  // It would be nice if this were in the sql package but that is not possible right now because
+  // Mounting is defined in core
+  def resolveImports[S[_]](blob: Blob[Fix[Sql]], basePath: ADir)(implicit
```
alissapajer: Would it be possible to document the two limitations that you described with a couple tests? Then we know exactly what behavior is to be expected. And then when we address these limitations we should just have to update the test expectations.

core/src/main/scala/quasar/package.scala
```diff
@@ -78,13 +83,35 @@ package object quasar {
...
+    mount: Mounting.Ops[S],
+    fsFail: Failure.Ops[FileSystemError, S]
+  ): Free[S, Block[Fix[Sql]]] = {
+    def resolvePath(path: DPath, base: ADir): ADir = refineTypeAbs(path).fold(ι, base </> _)
+    def retrieveDeclarations(module: ADir, allReadyImported: Set[ADir]): Free[S, List[FunctionDecl[Fix[Sql]]]] = {
+      if (allReadyImported.contains(module)) List.empty[FunctionDecl[Fix[Sql]]].point[Free[S, ?]]
```
alissapajer: Could you also add a test that will hit this case when we've already imported `module`?

alissapajer:
@jedesah just wanted to check in on the status of this PR?

jedesah:
So I decided to revisit the implementation in order to reduce the size of the `LogicalPlan` generated after resolving all imports (basically by only including function declarations that are actually used in the expression). It turns out anyway that `invoke` is currently broken (#2281) and the new approach to imports should take of things better by inlining everything at the SQL layer so the SQL to LogicalPlan compiler no longer needs to worry about function declarations.

I'm pretty close to being done with the re-implementation and when I'm done I'll also revisit your comments to make sure they are addressed in the new implementation if they are still relevant. I know there were a few good ideas around testing

web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
```diff
@@ -224,6 +221,30 @@ class ExecuteServiceSpec extends quasar.Qspec with FileSystemFixture {
         response = (a: String) => a.trim must_= "5"
       )
     }
+    "execute a query with imported functions" >> {
```
jedesah: So as things are currently, this would not be particularly easy to add since integration tests are currently wired up to expect an `Expr`, not a `Blob`. In any case, since imports are entirely removed before anything gets to any specific connector, I don't think it's necessary to have an integration test with `import`.

web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
```diff
@@ -224,6 +221,30 @@ class ExecuteServiceSpec extends quasar.Qspec with FileSystemFixture {
...
+          |TRIVIAL("bob")
+        """.stripMargin
+      get(executeService)(
+        path = rootDir,
+        query = Some(Query(query)),
+        state = InMemState.fromFiles(Map(sampleFile -> Vector(Data.Int(5)))),
+        mounts = Map((rootDir </> dir("mymodule"): APath) -> MountConfig.moduleConfig(List(funcDec))),
+        status = Status.Ok,
+        response = (a: String) => a.trim must_= "5"
+      )
```
jedesah: 👍

jedesah:
Discard the limitations I mentioned above, they have now been addressed.

@alissapajer This is ready for another look. It depends on #2317 but you are also assigned to that one (not doing on purpose to bombard you with PRs :smile:) so that shouldn't be too much of a problem!

alissapajer:
@jedesah great I'll take a look at both PRs!

core/src/main/scala/quasar/fs/mount/MountRequestHandler.scala
```diff
@@ -59,15 +57,8 @@ final class MountRequestHandler[F[_], S[_]](
   ): Free[T, MountingError \/ Unit] = {
     val handleMount: MntErrT[Free[T, ?], Unit] =
       EitherT(req match {
-        case MountView(f, blob, vars) =>
-          queryPlan(blob, vars, fileParent(f), 0L, None).run.value
-            .leftMap(e => invalidConfig(viewConfig(blob, vars), e.map(_.shows)))
-            .void.point[Free[T, ?]]
```
jedesah: Ya, because previously views were static so it was possible to `queryPlan` and know whether or not the SQL was valid or not and inform the user in case it was not and fail mounting. Now that views (and modules) can have imports (and because SQL functions and imports are fairly "dumb"), the SQL is unknown at this point so there is not tremendous value in validating and in any case I wouldn't want to fail to mount a view that references an import that doesn't exist yet. I think once we push functions and imports down to `LogicalPlan` (which hopefully will happen sooner than later, I kind of regret not pushing for it at the beginning of this project now) it would be nice to validate the SQL here but right now it's not really easily achievable. I'll add a comment to that effect in the code since it's very pertinent question.

web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
```diff
@@ -210,19 +206,61 @@ class ExecuteServiceSpec extends quasar.Qspec with FileSystemFixture {
...
-      )
+        response = (a: String) => a.trim must_= "5")
+    }
+    "execute a query with imported functions" >> {
+      val sampleFile = rootDir </> file("foo")
+      val funcDec = FunctionDecl(CIName("Trivial"), List(CIName("from")), sqlE"select * from :from")
+      val query =
+        """
+          |import `/mymodule/`;
+          |TRIVIAL(`/foo`)
```
alissapajer: Are there any restrictions to the function names that are allowed to be imported? For example say we have `import module; SUM(foo)`. And let's say that `SUM` is imported from `module`. Since `SUM` is also a sql function, which `SUM` will be used? Or will that be ambiguous? We should test that. (And I'm not sure what the correct behavior should be either... I would think an ambiguous import would be the most clear to the user, but I'm open to other ideas.)

sql/src/test/scala/quasar/sql/compiler.scala
```diff
@@ -1751,36 +1751,6 @@ class CompilerSpec extends quasar.Qspec with CompilerHelpers {
...
-        lpf.constant(Data.Dec(10.4)),
-        lpf.constant(Data.Dec(0.5)))
-    val expected =
-      lpf.invoke2(Subtract,
-        floorArgumentLP,
-        lpf.invoke2(Modulo,
-          floorArgumentLP,
-          lpf.constant(Data.Int(1))))
-    Compiler.compile[Fix[LP]](expr, funcs).toEither must beRight(equalToPlan(expected))
-  }
```
alissapajer: Why did these two tests get deleted?

frontend/src/test/scala/quasar/sql/SqlSpec.scala
```diff
@@ -43,6 +43,17 @@ class SQLSpec extends quasar.Qspec {
...
         }
       }
+      "apply arguments to a function declaration" >> {
+        "in the general case" in {
+          val funcDef = FunctionDecl(CIName("foo"), args = List(CIName("bar")), body = sqlE"SELECT * from `/person` where :bar")
+          funcDef.applyArgs(List(sqlE"age > 10")) must_=== sqlE"SELECT * from `/person` where age > 10".right
+        }
+        "if the variable is inside of a `from`" in {
+          val funcDef = FunctionDecl(CIName("foo"), args = List(CIName("bar")), body = sqlE"SELECT * from :bar")
+          println(sqlE"`/person`")
```
alissapajer: Forgotten `println`.

frontend/src/test/scala/quasar/sql/ResolveImportSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+package quasar.sql
+
+import slamdata.Predef._
+import quasar.fp.ski.κ
+
+import pathy.Path._
+import scalaz._, Scalaz._
+
+class ResolveImportSpec extends quasar.Qspec {
+  "Import resolution" >> {
```
alissapajer: Can you add some more complex test cases here as well. Some ideas:
- [ ] use the same variable name `:foo` multiple times in the same query
- [ ] demonstrate that functions with multiple arguments work
- [ ] show that multiple `FunctionDecl` in scope with the same function name will error
- [ ] show that using the same argument name multiple times in the same function will error, so `List(CIName("foo"), CIName("foo"))`

frontend/src/test/scala/quasar/sql/ResolveImportSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+    "simple case" >> {
+      val blob = sqlB"import `/mymodule/`; TRIVIAL(`/foo`)"
```
alissapajer: If I understand this correctly, we could replace `mymodule` here with anything and this test would still pass? So, here we are not testing whether or not `TRIVIAL` is actually imported from `mymodule`?

web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
```diff
@@ -210,19 +206,61 @@ class ExecuteServiceSpec extends quasar.Qspec with FileSystemFixture {
...
       get(executeService)(
         path = rootDir,
         query = Some(Query(query)),
         state = InMemState.fromFiles(Map(sampleFile -> Vector(Data.Int(5)))),
         status = Status.Ok,
-        response = (a: String) => a.trim must_= "5"
-      )
+        response = (a: String) => a.trim must_= "5")
+    }
+    "execute a query with imported functions" >> {
```
alissapajer: Some more examples to test:

- [ ] Show that a function can be imported simultaneously from two scopes (can it?)  `import mymodule; import othermodule; mymodule.TRIVIAL("bob") + othermodule.TRIVIAL("baz")`.
- [ ] Say `FOO` imports from `module1` and `BAR` imports from `module2`. Test `import module1; import module2; FOO(1) + BAR(2)`.
- [ ] Say `FOO` and `BAR` both import from `module1`. Test `import module1; FOO(1) + BAR(2)`.
- [ ] What about unused imports: `import module1; import module2; FOO(1)` where only `module1` imports `FOO`.

alissapajer: - [ ] Must imports necessarily all be in the most outer scope? Can we write `import module1; select FOO(*) from (import module2; select * from BAR(7)) as y` Will that even compile?

core/src/main/scala/quasar/fs/mount/module/Module.scala
```diff
@@ -169,8 +171,9 @@ object Module {
             userArgs     <- EitherT(maybeAllArgs.toRightDisjunction(argumentsMissing(missingArgs)).point[Free[S, ?]])
             parsedArgs   <- EitherT(userArgs.traverse(argString => fixParser.parseExpr(Query(argString)))
                               .leftMap(parsingErr(_)).point[Free[S, ?]])
-            sqlBlob      =  Blob(invokeFunction[Fix[Sql]](CIName(name), parsedArgs).embed, List(funcDec))
-            logicalPlan  <- EitherT(quasar.precompile[Fix[LogicalPlan]](sqlBlob, Variables.empty, basePath = fileParent(file))
+            blob         =  Blob(invokeFunction[Fix[Sql]](CIName(name), parsedArgs).embed, moduleConfig.statements)
```
jedesah: The `funcDec` above is used to find out what are the arguments to the function being called so we use we can match them with the parameters supplied by `InvokeModuleFunction`. The change from `List(funcDec)` to `moduleConfig.statements` is a bugfix because it's reasonable for a function in a module to reference another function in that module.

alissapajer:
@jedesah is there a github issue that this PR should close?

jedesah:
@alissapajer Yes, it's the last sub-task of #1792

alissapajer:
@jedesah awesome! Want to add that closes annotation to the PR description?

alissapajer:
@jedesah also I just realized that some of the new tests I requested match up with the limitations that you outlined earlier :) Though I still think it would be still good to have tests for those cases so we can document what happens. Then we can update the tests later when we address the problems.

core/src/main/scala/quasar/fs/mount/module/Module.scala
jedesah: I added a test to show this works now

core/src/main/scala/quasar/fs/mount/module/Module.scala
```diff
@@ -169,8 +171,9 @@ object Module {
...
+            sql          <- EitherT(resolveImports_(blob, currenDir).leftMap(e => semErrors(e.wrapNel)).run.leftMap(fsError(_))).flattenLeft
+            logicalPlan  <- EitherT(quasar.precompile[Fix[LogicalPlan]](sql, Variables.empty, basePath = fileParent(file))
```
jedesah: Why not :smile:. Luckily our code is pure so we know it's equivalent, but I guess I missed this one.

core/src/main/scala/quasar/package.scala
```diff
@@ -78,13 +83,28 @@ package object quasar {
...
+  ): EitherT[Free[S, ?], SemanticError, Fix[Sql]] =
+    EitherT(fsFail.unattemptT(resolveImports_(blob, baseDir).run))
+
+  def resolveImports_[S[_]](blob: Blob[Fix[Sql]], baseDir: ADir)(implicit
+    mount: Mounting.Ops[S]
+  ): EitherT[FileSystemErrT[Free[S, ?], ?], SemanticError, Fix[Sql]] =
+    resolveImportsImpl(blob, baseDir, d => mount.lookupModuleConfig(d).map(_.statements).toRight(pathErr(pathNotFound(d))))
+
+  // It would be nice if this were in the sql package but that is not possible right now because
+  // Mounting is defined in core
```
jedesah: That's just a remanent of before I separated out `resolveImportsImpl` and moved it into the `sql` package.

core/src/main/scala/quasar/package.scala
```diff
@@ -78,13 +83,28 @@ package object quasar {
...
+
+
   /** Returns the `LogicalPlan` for the given SQL^2 query, or a list of
     * results, if the query was foldable to a constant.
     */
   def queryPlan(
-    blob: Blob[Fix[Sql]], vars: Variables, basePath: ADir, off: Natural, lim: Option[Positive]):
```
jedesah: Ya, sure

frontend/src/main/scala/quasar/sql/package.scala
```diff
@@ -154,6 +155,61 @@ package object sql {
...
+            decl.traverse(_.inlineInvokes(currentScope)).flattenLeft.strengthR(from)
+          }
+        }).map(_.join)
+    }
+
+    blob.expr.inlineInvokes(scopeFromHere(blob.imports, blob.defs, baseDir)).flattenLeft
+
+    // This blob has no more imports (assuming the implementation is correct)
+//    val blobInlined = blob.traverse(_.inlineInvokes(scopeFromImports(blob.imports.map(absolutizeImport(_, baseDir)))).flattenLeft)
+//    blobInlined.map(blob => Block(blob.expr, blob.defs))
```
jedesah: Yup! :smile:

frontend/src/test/scala/quasar/sql/ResolveImportSpec.scala
jedesah: I changed this to be more "real"

frontend/src/test/scala/quasar/sql/ResolveImportSpec.scala
jedesah: So that last case was totally not handled. I wrote a test for it (not that I needed the test to know it woudn't behave as expected). This really makes me annoyed that we aren't compiling functions to `LogicalPlan` yet (I mostly have myself to blame for that) because that would be the perfect opportunity to catch this kind of thing and error out early like when mounting a module. As is, you will need to call the function before you get this error :disappointed:

sql/src/test/scala/quasar/sql/compiler.scala
jedesah: Previously, the `Sql => LogicalPlan` compiler received some user defined functions but since everything is being inlined first now I changed the signature of the compiler back to accepting a single expression (`Fix[Sql]`) so these tests don't make any sense anymore. The same functionality is being tested elsewhere anyway

web/src/test/scala/quasar/api/services/InvokeServiceSpec.scala
```diff
@@ -116,36 +112,43 @@ class InvokeServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s
...
-          val mounts = Map((fileParent(functionFile):APath) -> MountConfig.moduleConfig(statements))
-          val state = InMemState.fromFiles(Map(dataFile -> sampleData))
-          val request = Request(uri = pathUri(functionFile).copy(
-            query = Query.fromPairs("bar" -> "2", "offset" -> offset.toString, "limit" -> limit.toString)))
+          val statements = List(sampleStatement(fileName(functionFile.path).value))
+          val mounts = Map((fileParent(functionFile.path):APath) -> MountConfig.moduleConfig(statements))
+          val state = InMemState.fromFiles(Map(dataFile.path -> sampleData))
+          val arg = "`" + posixCodec.printPath(dataFile.path) + "`"
+          val request = Request(uri = pathUri(functionFile.path).copy(
+            query = Query.fromPairs("bar" -> arg, "offset" -> offset.toString, "limit" -> limit.toString)))
```
jedesah: `bar` comes from `sampleStatement` above

web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
jedesah: First and last aren't currently supported but the rest I added tests. I added them in `ResolveImportSpec` though, don't think it's necessary to repeat all these scenarios at the web layer.

web/src/test/scala/quasar/api/services/query/ExecuteServiceSpec.scala
jedesah: Until we push user defined functions to `LogicalPlan` I just don't see a worthwhile way of achieving an ambiguous call/reference, so, for now, it will choose the user defined function

jedesah: I did add a test to document this behavior

jedesah:
@alissapajer The limitations in the description were removed, I mentioned it later, but went ahead and removed them from the initial description. Also, added the "closes" you suggested.

jedesah:
@alissapajer As for `Fix`, I thought I could generalize `resolveImportsImpl` to any `T[_[_]]` but if it was possible, it no longer is because I had to hardcode a couple extra functions to `Fix` in order to implement the "multiple arguments with the same name" error. `SemanticError` being hardcoded to `Fix` is a major blocker to generalizing things but then again I question whether we really want to make those generic, there is such a thing as too many type variables at one point :smile:

jedesah:
@alissapajer Thx for the thorough review! I think I have addressed everything you pointed out

project/Dependencies.scala
```diff
@@ -54,6 +54,7 @@ object Dependencies {
     "com.chuusai"                %% "shapeless"                 % shapelessVersion,
     "org.scalacheck"             %% "scalacheck"                % scalacheckVersion,
     "com.propensive"             %% "contextual"                % "1.0.1",
+    "com.softwaremill.quicklens" %% "quicklens"                 % "1.4.8",
```
alissapajer: Since we already use lenses from `monocle`, I think it would be good to not add more dependencies on other lens libraries. Could you use `monocle` instead of adding this dependency?

frontend/src/test/scala/quasar/sql/ResolveImportSpec.scala
```diff
@@ -17,17 +17,124 @@
...
+          End;
+        CREATE FUNCTION FOO(:a)
+          BEGIN
+            FLOOR(:a)
+          END;
+        FOO(1)
+      """
+      // This error is slightly bizarre because they are both defined in the same scope, but I don't think it's
+      // worth creating a special error type for this particular case
+      // Besides, this is the kind of thing that should be caugth when compiling to LogicalPlan once
```
alissapajer: s/caugth/caught/

frontend/src/main/scala/quasar/sql/package.scala
```diff
@@ -155,28 +156,30 @@ package object sql {
...
+  }
 
+  implicit class FixExprOps(q: Fix[Sql]) {
     /**
       * Inlines all function invocations with the bodies of functions in scope.
       * Leaves invocations to functions outside of scope untouched (as opposed to erroring out)
       * @param scope Returns the list of function definitions that match a given name and function arity along with a
       *              path specifying whether this function was found
       */
-    def inlineInvokes[M[_]: Monad](scope: (CIName, Int) => M[List[(FunctionDecl[T[Sql]], ADir)]]): EitherT[M, SemanticError, T[Sql]] = {
```
alissapajer: Can we keep this generalized to `T[_[_]]` instead of using `Fix`?

frontend/src/main/scala/quasar/sql/Statement.scala
```diff
@@ -65,20 +67,23 @@ object Statement {
     f(body).map(FunctionDecl(name, args, _))
   override def pprint(implicit ev: BODY <~< String) =
     s"CREATE FUNCTION ${name.shows}(${args.map(":" + _.shows).mkString(",")})\n  BEGIN\n    ${ev(body)}\n  END"
-  def applyArgs[T[_[_]]: BirecursiveT](argsProvided: List[T[Sql]])(implicit ev: BODY <~< T[Sql]): SemanticError \/ T[Sql] = {
```
alissapajer: Can we keep using `T[_[_]]` here? What necessitated the change over to `Fix`?

core/src/main/scala/quasar/fs/mount/MountRequestHandler.scala
alissapajer: @jedesah so now that we don't check to see if the SQL is valid before mounting, what will happen when the SQL is invalid and we mount anyway? Are we prepared to handle this situation? I'm concerned about introducing a user-facing issue here.

alissapajer:
@jedesah thanks for adding all those tests! A few more comments above.

frontend/src/main/scala/quasar/sql/Statement.scala
alissapajer: Ah ok, so I just caught up on your comment above about `SemanticError` having `Fix` hardcoded. Since this is new code, we're not really making things worse by using `Fix` here. We need to generalize away from `Fix` in `frontend` probably in one or two big passes anyways.

frontend/src/main/scala/quasar/sql/package.scala
alissapajer: Same, I found your comment about this above.

frontend/src/main/scala/quasar/sql/Statement.scala
sellout: RE: SemanticError needing `Fix`, I’ve avoided hardcoding in other places by doing `someError(x.convertTo[Fix[Sql]])`, where `(x: T)(implicit T: Recursive.Aux[T, Sql])`

alissapajer: Nice, thanks, @jedesah would that work for you here?

core/src/main/scala/quasar/fs/mount/MountRequestHandler.scala
sellout: @jedesah The problem is that we can’t `substVars` at this point? We should at least go as far as we can – if we can’t call `queryPlan`, we should at least be able to `parse` right? (And maybe another step or two toward LP.) And there should probably be a comment (and an issue) that talks about why we can’t get all the way to QScript (or at least LP).

alissapajer: :+1: to creating an issue and referencing it in the comment.

project/Dependencies.scala
jedesah: `Monocle` doesn't do what `quicklens` does. My view on it is that `Monocle` is for predefined composable lenses and `quicklens` is for one-off situations (it's basically a better version of `copy`). As the guy who typically upgrades stuff, I am not too worried about adding a dependency that does not itself have any dependencies as it adds little to no complications to staying current on our dependencies (i.e. the only thing I can think of is there not being a version published for a new version of Scala, but I don't foresee that being a problem with `quicklens`). We already use `quicklens` in QA and it's been no hassle at all.

frontend/src/main/scala/quasar/sql/Statement.scala
jedesah: Yes, that would most probably work! I was wondering if such a thing existed but then somehow convinced myself that probably not. I'll revert the hardcoding of `Fix` in favor of converting for errors messages.

jedesah:
@alissapajer Github won't let me comment on your comment about views not being checked when mounted so I am responding here. It means that any invalid sql syntax will be checked when it's actually used instead of when it's mounted. Same as if you were to provide an invalid query to the `execute` endpoint. Unfortunately currently, the error message might not be ideal because it will inline everything before looking to see if the final SQL expression is valid which will yield less desirable error messages then if we checked everything in isolation. But I think that will be part of pushing user functions to `LogicalPlan`. Also, in regards to the issue @drostron is looking into about mounts we serialize to the metastore that can't be parsed back, this is orthogonal as we are still parsing the Sql just not compiling it as well.

sql/src/test/scala/quasar/sql/compiler.scala
```diff
@@ -499,8 +499,9 @@ class CompilerSpec extends quasar.Qspec with CompilerHelpers {
     }
 
     "fail to compile let inside select with ambigious reference" in {
-      compile(sqlE"select foo from (bar := 12; baz) as quag") must
-        beLeftDisjunction  // AmbiguousReference(baz)
+      // TODO: Make description and query match
```
jedesah: It's talking about let expressions but the actual error message seems to have nothing to do with the let statement. I mentioned it to @sellout on Friday and we couldn't figure it out but I see now in git that you are the one who added this test so you probably have more insight into what's meant to be tested. Maybe the expression just needs to be modified slightly not to error out with a `CompiledSubtableMissing`, but a `AmbiguousReference` of some kind.

project/Dependencies.scala
sellout: I’m not opposed to a library that just gives us lenses for everything, but I agree that I don’t see much argument for not just using `@Lenses` to avoid duplicating functionality in our deps.

core/src/main/scala/quasar/fs/mount/MountRequestHandler.scala
jedesah: Apparently, I can comment here now 😕 . @sellout The problem is that we can't resolve imports at this point. We do parse because `MountView` has a `ScopeExpr` so it's been parsed already by this point. Happy to open an issue for pushing user functions and imports to `LogicalPlan` (not convinced it needs to go all the way to Qscript) that should allow us to do something better here

sql/src/test/scala/quasar/sql/compiler.scala
alissapajer: Hm, so it seems like we need to improve our error messages coming out of the compiler. There is an ambiguous reference in that query, so it would be helpful if an error about an ambiguous reference were returned to the user instead of the message about the subtable. 

Maybe update this comment (and the other one below) with a `TODO` to investigate why the query isn't producing an ambiguous reference error? My best quick guess is that we are getting this subtable error because we actually don't have a valid subtable _because_ of the ambiguous reference, so the ambiguous reference is the root cause which needs to bubble up.

alissapajer: But I think the query description is accurate (or at least accurately reflects the incorrectness in the query) and it's the error message that needs fixing.

project/Dependencies.scala
alissapajer: It looks like the only use site of `quicklens` is [here](https://github.com/quasar-analytics/quasar/pull/2178/files#diff-f8a4d5e0c8a782075c31ba0bcfbaa136R148)? I don't see the value we're getting from this when we could do the same thing with `@Lenses` from `monocle`:

```scala
@Lenses case class Foo(x: Int)

val foo = Foo(5)
val mod = Foo.x.modify(_ + 1)(foo)
```

I don't think we should have dependencies that provide duplicate functionality when we can avoid it. Our code base is more uniform and consistent when we use a single library for each task.

jedesah:
@alissapajer If you have time this evening to have a quick look at the small changes I made to fix the compile issues and address the last few things you and @sellout pointed out, then @natefaubion will have a published artifact to work with tmr morning. If not, no worries, I'm sure you'll get to it tmr. You've been super responsive about reviewing this PR, not trying to rush you, just want to get this to @natefaubion as soon as possible 😄

alissapajer:
@jedesah taking a look now!

PR #2190, opened 2017-05-11
Fix leading zeros issue in prettify test.

### LINKED ISSUE #2184: test failure in quasar.main.PrettifySpecs
```
[error] x round-trip all flat rendered values that aren't "" (787 ms)
[error]  Falsified after 96 passed tests.
[error]  > ARG_0: Id(063877087438042028614974)
[error]  > 'Some(63877087438042028614974)' is Some but '63877087438042028614974' is not equal to '063877087438042028614974' (prettify.scala:251)
[error] Actual:   []63877...
[error] Expected: [0]63877...
```

Looks like an issue with a leading `0`.

alissapajer:
Closes #2184


alissapajer:
Scala removes leading zeros when a string is successfully parsed as a `BigDecimal` or a `BigInt`. For example:
```
scala> BigDecimal("0123")
res1: scala.math.BigDecimal = 123

scala> BigDecimal("00.123")
res2: scala.math.BigDecimal = 0.123
```
This PR alters test expectations to expect the leading zeros to be removed when we call `Prettify.parse`.

interface/src/test/scala/quasar/main/prettify.scala
```diff
@@ -242,7 +242,29 @@ class PrettifySpecs extends quasar.Qspec {
       case _ => true
     }
 
-    "round-trip all flat rendered values that aren't \"\"" >> prop { (data: Data) =>
+    def parseNumericString(str: String): String =
```
jedesah: I am thinking there might be a better name for this. Perhaps `pruneNumericString` or `trimNumericString`?

interface/src/test/scala/quasar/main/prettify.scala
```diff
@@ -242,7 +242,29 @@ class PrettifySpecs extends quasar.Qspec {
...
+      str.parseBigDecimal.toOption.map(_.toString).getOrElse(str)
+
+    def removeLeadingZero(data: Data): Data = data match {
```
jedesah: Perhaps this and the above function can be defined closer to their call site so below these next two tests which don't make use of it

interface/src/test/scala/quasar/main/prettify.scala
```diff
@@ -242,7 +242,29 @@ class PrettifySpecs extends quasar.Qspec {
...
+      case Data.Str(str) => Data.Str(parseNumericString(str))
+      case Data.Id(str) => Data.Id(parseNumericString(str))
+      case data => data
+    }
+
+    "handle an integer string with a leading zero" in {
+      val data = Data.Id("012345")
```
jedesah: So why are leading zeros meaningless in identifiers? Seems to me that an identifier should be a unique string without any numeric semantics.

interface/src/test/scala/quasar/main/prettify.scala
```diff
@@ -242,7 +242,29 @@ class PrettifySpecs extends quasar.Qspec {
...
+      val r = render(data).value
+      parse(r).map(render(_).value) must beSome("12345")
+    }
+
+    "handle a decimal string with a leading zero" in {
+      val data = Data.Str("00.12345")
```
jedesah: And actually if this is meant to be treated as a `String` why do the leading zeros matter and why are they getting eliminated? Perhaps I'm missing a bit of context on this change.

interface/src/test/scala/quasar/main/prettify.scala
alissapajer: :+1:

interface/src/test/scala/quasar/main/prettify.scala
alissapajer: :+1:

interface/src/test/scala/quasar/main/prettify.scala
alissapajer: The original round-trip test is testing that when we `Prettify.parse` and then `render` we get the same result as when we just `render`. The test was failing with strings that parse as a number and start with a leading zero due to the implementation of `Prettify.parse` which calls into `parseBigDecimal` (and `parseBigInt`). So `Prettify.parse` is already removing the leading zeros when it parses, but we weren't handling that in the tests. So that's what I'm attempting to remedy.

interface/src/test/scala/quasar/main/prettify.scala
alissapajer: See answer to previous question.

interface/src/test/scala/quasar/main/prettify.scala
jedesah: I still don't understand why when we render and parse a `Data.String` anything should change. I understand that this is simply "verifying" or "documenting" the existing behavior but since we are changing the tests might as well make sure we understand them and they are correct and if not perhaps we can fix the behavior.

alissapajer: So, in the scalacheck test we begin with a `Data` and then `render` that into a string. So for example we start with `Data.Str("0123")`, which renders to `"0123"`. Then we parse that string using `Prettify.parse`, which delegates to `parseBigInt`, which will successfully parse that string as the integer `123` and wrap it in `Data.Int(123)`. Then in the test we render that and get `"123"`. That behavior of `parse` doesn't seem unreasonable to me. I think this is just a test corner case we didn't think of. If we wanted to change the behavior we'd have to parse `"0123"` as a string and not as an integer like we currently do. Is there any reason we'd want to do that?

Outside of a test we would not have a reason to `render` a `Data` only to `parse` it back into `Data`. This is just something we are doing for the sake of testing, if I understand this correctly.

alissapajer: And even if we ignore the leading zero issue, if we start with `Data.Str("123")` that renders to `"123"` and parses to `Data.Int(123)`. So rendering and then parsing isn't guaranteed to return the same `Data` type, which is why we render the result again in the test expectation.

jedesah: Ah, I see, I understand. Ya this seems more reasonable to me now. 👍

PR #2195, opened 2017-05-12
Persist RDD being used with QueryFile.evaluate

rabbitonweb:
This should boost performance.

We persist `RDD` in memory when file is evaluated. We free the memory when all data was read from it.

Resolves https://github.com/quasar-analytics/quasar/issues/1664


rabbitonweb:
@alissapajer could you have a look at this PR, it is few lines of code being changed

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
```diff
@@ -212,15 +212,20 @@ object queryfile {
               .filter(d => d._2 >= p && d._2 < (p + step))
               .map(_._1).collect.toVector
           }).into[S].liftM[FileSystemErrT]
-          rddState = if(collected.isEmpty) RddState(None, 0) else RddState(Some(rdd), p + step)
+          rddState <- lift(Task.delay {
+            if(collected.isEmpty) {
+              val unpersisted = rdd.unpersist()
```
alissapajer: What type does this return? Should we be doing something with the result?

alissapajer:
Needs a semantic version label.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/queryfile.scala
rabbitonweb: That's the funny part. `RDD.unpersist` is doing two things:
1. it frees up the memory [this is the effect that we need]
2. it returns a new `RDD` that represents

Since we are droping RDD here to be garbage collected we do not really need the 'unpersisted' `RDD`. However wart requires my to assign it to a value.

sellout: @rabbitonweb We have `quasar.fp.ignore` which is for explicitly discarding a value. You should use that instead of assigning to an unused `val`.

Hopefully we can eventually use `-Ywarn-unused` which would flag what you’re currently doing.

rabbitonweb:
@alissapajer label added + `ignore` added

alissapajer:
@rabbitonweb in the PR description you have "This solves #1664" but that isn't one of the [keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) that will actually close the issue. So I'm updating the PR description.

alissapajer:
@rabbitonweb this looks like it requires a merge from master - isn't compiling in sparkcore anymore.

Issue #1740, opened 2016-11-30
CB: Offer bucket configuration per mount and remove default bucket requirement


drostron:
Currently the Couchbase connector relies on the default bucket existing with anonymous access. CB's Java SDK requires a bucket when issuing queries. We should verify that it exists upon mount to avoid later confusion.

In its absence the repl offers the mildly useful `Quasar error: Passwords for bucket "default" do not match.` while the REST api offers the less useful 500 `{ "error": { "status": "Physical filesystem error.", "detail": { "message": "java.util.concurrent.TimeoutException" } } }`.

Alternately configuration could be offered for specifying a bucket name and credential for use with queries.

drostron:
In light of recent events and [recommendations](https://blog.couchbase.com/2017/couchbase-customer-advisory-note-security) we should offer configuration.

PR #2196, opened 2017-05-12
CB: bucket per mount mapping

nicflores:
Fixes: #1740

This PR changes the couchbase connection URI format in two ways:
a) bucket name needs to be specified after `couchbase://<host>[:<port>]/`
b) a document type needs to be specified using `docTypeKey`

In addition, a Quasar couchbase mount now maps directly to a couchbase bucket.


drostron:
@nicflores Needs a merge from master

Issue #1778, opened 2016-12-12
Add tests to StdLibSpec checking comparison between Date & Timestamp is incorrect


rabbitonweb:
Any comparison (LT, GT, EQ) between `Date` & `Timestamp` should return `Data.NA`

zeeiyerWork:
@rabbitonweb is unsure if it is done or not. But will check in the may sprint.

rabbitonweb:
It's implemented for sparkcore, waits for other connectors to fill the gap.

alissapajer:
I think we should close this ticket when #2203 lands and then open a specific ticket for each backend for which the new tests are failing.

PR #2203, opened 2017-05-16
Add tests to StdLibSpec checking comparison between Date & Timestamp …

rabbitonweb:
When comparing `Timestamp` and `Date` returned value should always be `Data.NA`

Closes https://github.com/quasar-analytics/quasar/issues/1778


rabbitonweb:
This test fails for most of the connectors. 
@sellout @drostron @wemrysi can you check this on your side?

sellout:
See, this is the kind of case where we really want universals (and type classes) in our type system. Rather than having ~ `val LT(a: Comparable, b: Comparable)`, we should have `val LT[A: Comparable](a: A, b: A)`. Which would allow the typechecker to 1. check the constraint and 2. check that the types match. But currently it just checks that each argument independently satisfies the constraint. So, ideally, a connector should never execute a case where the types don’t line up correctly (because the type check would prevent it).

rabbitonweb:
@sellout that would be nice indeed. But for now, we should fix connectors right?

sellout:
Yes, absolutely. I was just tying it back to your new book 😄

rabbitonweb:
@sellout I know :)

djspiewak:
@rabbitonweb Can you assign this PR to someone?

rabbitonweb:
@djspiewak I'm happy to assign but shouldn't I wait till every connector will fix its errors?

drostron:
I have a branch with the potential fix for CB [here](https://github.com/drostron/quasar/tree/std_comparison_CB). Some attention is still due to address a few integration test failures that cropped up due to the change. I've been pulled away by a few other issues but will try and get back to this before too long.

alissapajer:
Can we make these tests pending for the connectors the tests are failing under? Then we could get this merged.

rabbitonweb:
@alissapajer @djspiewak I can make those tests pending, np

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -1090,7 +1090,9 @@ abstract class StdLibSpec extends Qspec {
           binary(Lte(_, _).embed, Data.Str(x), Data.Str(y), Data.Bool(x <= y))
         }
 
-        // TODO: Timestamp, Interval, cross-type comparison
+        "any date & timestamp" >> prop { (d: LocalDate, i: Instant) =>
+          binary(Lte(_, _).embed, Data.Date(d), Data.Timestamp(i), Data.NA)
```
alissapajer: I think we should also have a test for `Lt` and `Eq`, just to cover all our bases.

alissapajer:
@rabbitonweb I'm working on `StdLibSpec` for mimir now, so since I'm already in this code, I'd be happy to make the failing tests pending so we can get this merged. Then as I mentioned in #1778 we can open individual tickets for each backend. It will be easier to track the remaining work that way, and then we don't have to have PRs open like this for a long time.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
rabbitonweb: those are added today

rabbitonweb:
I;ve discussed with @wemrysi possible ways to make it pending for particular connector, will have it done tomorror

rabbitonweb:
@alissapajer ready for review

alissapajer:
@rabbitonweb no idea why but the travis build didn't even run for this one. Will merge when we see a green build. I'll look into it as well.

rabbitonweb:
I'm pretty sure I've seen it green on Travis. But sure, re-run it, np. 
Im afk
I will be back in 30min.

alissapajer:
@rabbitonweb in the most recent build for this PR https://travis-ci.org/quasar-analytics/quasar/builds/253134766 ` some time-related specs are failing in `quasar.std.SimplifyStdLibSpec` for all mongo backends, so they probably just need to be make pending too.

alissapajer:
since @rabbitonweb is off on holiday starting tomorrow, I'm going to take charge of fixing the remaining compile errors so we can get this PR merged. I'll assign another reviewer before merge so I'm not merging my own work :)

frontend/src/test/scala/quasar/std/SimplifyStdLibSpec.scala
```diff
@@ -41,6 +41,12 @@ class SimplifyStdLibSpec extends StdLibSpec {
   def shortCircuit[N <: Nat](func: GenericFunc[N], args: List[Data]): Result \/ Unit = (func, args) match {
     case (relations.Between, _) => notHandled
```
rabbitonweb: wow that was in front of my eyes the whole time...

alissapajer:
Taking notes here, newly added tests are still failing in:
`quasar.physical.mongodb.MongoDbJsStdLibSpec`
`quasar.physical.mongodb.MongoDbExprStdLibSpec`
`quasar.physical.sparkcore.fs.CoreMapStdLibSpec`

PR #2204, opened 2017-05-16
Generic path card

rabbitonweb:
Part of https://github.com/quasar-analytics/quasar/issues/1716


rabbitonweb:
Was already merged in https://github.com/quasar-analytics/quasar/pull/2086

PR #2223, opened 2017-05-18
Backend module update

drostron:
- Remove the need for connectors to handle FileSystemErrors and PhaseResults
- Remove Free from BackendModule


wemrysi:
@djspiewak Yes, essentially it shifts the burden of handling `FileSystemError` and `PhaseResults` onto `BackendModule` and off of the individual connectors.

PR #2233, opened 2017-05-23
Include backport branches in publishing

drostron:


scripts/publishJarIfMaster
```diff
@@ -15,7 +15,9 @@ if [[ ! -v GITHUB_TOKEN ]] ; then
 fi
 
 # only publish on quasar-analytics/quasar#master
```
alissapajer: We should probably update the comment now that we publish backport as well.

scripts/publishJarIfMaster
```diff
@@ -15,7 +15,9 @@ if [[ ! -v GITHUB_TOKEN ]] ; then
...
-if [[ "$TRAVIS" == "true" && "$TRAVIS_BRANCH" == "master" && "$TRAVIS_REPO_SLUG" == "quasar-analytics/quasar" ]] ; then
+if [[ "$TRAVIS" == "true"                                              &&
+      ("$TRAVIS_BRANCH" == "master" || "$TRAVIS_BRANCH" == backport/*) &&
```
alissapajer: Is there a reason `"master"` is in quotes while `backport/*` is not?

scripts/publishJarIfMaster
drostron: Sure. The filename is out of date now too. Will update accordingly.

scripts/publishJarIfMaster
drostron: The prefix matching only works outside of quotes.

alissapajer: Ok cool, I figured as such, just wanted to double check that it was intentional.

scripts/publishJarIfMaster
alissapajer: Sounds good. After that commit lands I will merge!

drostron:
Realized we'll need to update sbt-slamdata's `publishAndTag` similarly. Will include that here once available.

alissapajer:
@drostron when you've pushed everything you need to, just let me know, and I'll take another look!

drostron:
@alissapajer Ready for another look

project/build.sbt
```diff
@@ -2,7 +2,7 @@ libraryDependencies += "org.kohsuke" % "github-api" % "1.59"
 
 disablePlugins(TravisCiPlugin)
 
-scalacOptions ++= commonScalacOptions_2_10
+scalacOptions ++= scalacOptions_2_10
```
alissapajer: Why did this change now?

drostron: It comes with the sbt-slamdata 0.1.0 update

PR #2236, opened 2017-05-23
Don't distinguish between exclusive and normal tests

rintcius:
MongoDbFileSystemSpec is the only spec using ExclusiveQuasarSpecification and having this trait mixed in made the spec test nothing.
Also remove ExclusiveTests from build.sbt since its only purpose seems to be for ExclusiveQuasarSpecification

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,14 +18,6 @@ import slamdata.SbtSlamData.transferPublishAndTagResources
 
 val BothScopes = "test->test;compile->compile"
 
-// Exclusive execution settings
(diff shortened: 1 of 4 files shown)


rintcius:
Ah, the test is meant to be invoked via `exclusive:test`, so it's not broken.
Nevertheless, if travis runs ok without having to distinguish between `exclusive` or not, then I think this simplification is still warranted.

rintcius:
Closing the PR - exclusive tests are more restrictive than `sequential` tests

PR #2239, opened 2017-05-23
Backport v16.1.7+ → master

drostron:


alissapajer:
@drostron needs a semver label; I guess the "highest" version semantic that was applied to these commits when we merged them into backport.

drostron:
@alissapajer Ah, had one on the PR with the strange appearance of Travis in our commit history, but forgot to re-add here. Yep, sounds right. _Done_

PR #2256, opened 2017-05-30
Map invalid mongo values (NaN and +/- infinite) to undefined

### LINKED ISSUE #1458: Cannot query/filter out `NaN` value
When `NaN` value is getting into query output, we get short quasar message:

```
SELECT number FROM `/database/collection` WHERE Timestamp = 1469578582636
Quasar error: null
```

There is also no syntax to filter our `NaN` from query output:

```
SELECT number FROM `/database/collection` WHERE number <> NaN
MongoDB
db.collection.aggregate(
  [
    {
      "$project": { "number": true, "__tmp0": { "$ne": ["$number", "$NaN"] } }
    },
    { "$match": { "__tmp0": true } },
    { "$project": { "number": true, "_id": false } }],
  { "allowDiskUse": true });

```

The only way to get `NaN` is to use `TO_STRING` function, but this is the worst case.

rintcius:
Map invalid mongo values (NaN and +/- infinite) to null

Main effort was in the test part. I wanted to reuse the IT test infrastructure
as much as possible, but the problem is that NaN does not have a representation
in both `Json` (so cannot use `RegressionTest`'s `data` directly)
and `Data` (so post-processing `Json` won't work either).

For this moment, I opted for the simplest solution that I could think of, which means it's not that general. Some ways in which this can be made more general:

* apart from db table data to be specified in `.data` files, also allow it to be specified in code
  per table for each supported backend
* add an interpretation step when loading db data
  (e.g. we could add an extra flag that would enable loading `.data` files in a backend specific way
  and making sure that e.g. `"$NaN"` would be loaded as `NaN` for mongo)

Fixes #1458


it/src/test/scala/quasar/MongoResultFileQueryRegressionSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+class MongoResultFileQueryRegressionSpec
+  extends QueryRegressionTest[FileSystemIO](
+    QueryRegressionTest.externalFS.map(_.filter(MongoResultFileQueryRegressionSpec.isSupported))
+  ) {
+
+  val read = ReadFile.Ops[FileSystemIO]
+
+  val suiteName = "MongoResultFile Queries"
+
+  def queryResults(expr: Blob[Fix[Sql]], vars: Variables, basePath: ADir) = {
```
rintcius: this method is duplicated from `ResultFileQueryRegressionSpec` - I can have a look into extracting this in a separate `trait` but maybe there's a better way to deal with this?

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -121,8 +121,10 @@ abstract class QueryRegressionTest[S[_]](
   ////
 
   /** Returns an `Example` verifying the given `RegressionTest`. */
-  def regressionExample(
+  def regressionExample[S[_]](
+    ref: BackendRef,
     loc: RFile,
+    testDir: ADir,
```
rintcius: added the `testDir` parameter here, so that it can be used in `MongoResultFileQueryRegressionSpec.loadData` to set up the data in the appropriate mongo collection. Happy to learn if this can be avoided though.

mongodb/src/main/scala/quasar/physical/mongodb/bsoncodec.scala
```diff
@@ -155,6 +155,8 @@ object BsonCodec {
     case Bson.Null             => C.inj(ejson.Null()).right
     case Bson.Bool(value)      => C.inj(ejson.Bool(value)).right
     case Bson.Text(value)      => C.inj(ejson.Str(value)).right
+    case Bson.Dec(value) if (value.isNaN || value.isInfinity)
+                               => C.inj(ejson.Null()).right
```
rintcius: Another option would be to use a single pattern match for `Bson.Dec`:
```
    case Bson.Dec(value) if (!value.isNaN && !value.isInfinity)
                               => C.inj(ejson.Dec(value)).right
```
This would make `NaN` and `+/-infinity` match on `case bson                  => bson.left` instead and runs ok with the following expected data for the added integration test:

```
    "expected": [
      { "name": "NaN"                 },
      { "name": "inf"                 },
      { "name": "-inf"                },
      { "name": "valid", "val": 42.42 },
      { "name": "null",  "val": null  }
    ]
```

I guess this may be more close to the undefined value mentioned in https://github.com/quasar-analytics/quasar/issues/1458#issuecomment-302746701 ?

it/src/test/scala/quasar/MongoResultFileQueryRegressionSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+class MongoResultFileQueryRegressionSpec
+  extends QueryRegressionTest[AnalyticalFileSystemIO](
+    QueryRegressionTest.externalFS.map(_.filter(MongoResultFileQueryRegressionSpec.isSupported))
+  ) {
+
+  val read = ReadFile.Ops[AnalyticalFileSystemIO]
+
+  val suiteName = "MongoResultFile Queries"
+
+  def queryResults(expr: Blob[Fix[Sql]], vars: Variables, basePath: ADir) = {
```
rintcius: this method is duplicated from ResultFileQueryRegressionSpec - I can have a look into extracting this in a separate trait but maybe there's a better way to deal with this?

it/src/main/resources/mongotests/NaN/selectInvalidNumbers.test
```diff
@@ -0,0 +1,15 @@
...
+    "name": "select invalid numbers",
+    "backends": {},
+    "description": "select of NaN and infinite should be treated as null",
+    "data": "NaNs.data",
+    "query": "select * from NaNs",
+    "predicate": "containsExactly",
+    "expected": [
+      { "name": "NaN",   "val": null  },
+      { "name": "inf",   "val": null  },
+      { "name": "-inf",  "val": null  },
```
sellout: So, we actually want to treat these values as _undefined_, not `null`. The correct results should look like
```json
[{ "name": "NaN" },
 { "name": "inf" },
 { "name": "-inf" },
 { "name": "valid", "val": 42.42 },
 { "name": "null",  "val": null  }]
```
as undefined values truly don’t exist. `null` in MRA is simply a poorly-named `Unit`. Any weird SQL semantics for `null` (e.g., `foo is not null`) are handled in the SQL compiler (i.e., it becomes something like `foo != null ?? false`).

mongodb/src/main/scala/quasar/physical/mongodb/bsoncodec.scala
sellout: I don’t think we want to handle this here – we’re throwing away the distinction too early, and not handling similar situations in other connectors.

Since NaN and infinity are being properly encoded as floating-point values here, we should keep them as is.

But you’re right in that the expected results here are what we want.

You need to figure out exactly where NaN is blowing up and handle it there. My guess is either when we go to Argonaut or when we serialize Argonaut.

rintcius: Problem is that `ejson.Dec` contains a `BigDecimal` so it does blow up here. 

See https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html

```
public BigDecimal(double val)
Throws:
NumberFormatException - if val is infinite or NaN.
```

sellout: Oh, well nice. So nevermind on that part then 👍🏾

sellout: Then we definitely need to ensure that the test is running on other connectors, because I bet no one is handling this properly. Yay, unchecked exceptions 🙄

rintcius: I don't think we can represent `NaN` and `+/-infinity` in `Data`. That's why I resorted to a backend-specific insert.

sellout: Oh, you’re right – it runs into the same “problem” you just brought up.

Hrmm.

Maybe we can write a test in a way that should expose NaN and infinities computationally. E.g., `select pop / 0 from zips`. That actually still has a problem in that (ideally, but maybe not currently), our optimizer should identify that and avoid a connector having to evaluate it. But if we made a new data set with an entry like `{ "val": 43, "zero": 0}` and did `select val / zero from newDataSet` that should force it. And ideally entries that also evaluate to the infinities.

Then we don’t need _any_ special test handling, just the new `.test` file and a new `.data` file.

rintcius: Hmm.. seems like that trick doesn't work. Given
```
 $ select * from Nans
MongoDB:
db.Nans.find().limit(11);


Query time: 0.0s
 val1 | val2 |
------|------|
    1 |    0 |
    1 |  0.0 |
   -1 |  0.0 |
```
both `select val1 / 0 from Nans` and `select val1 / val2 from Nans` result in a mongo server error:
```
Quasar error: Plan execution failed: MongoDB Error: Command failed with error 16608: 'exception: can't $divide by zero' on server 192.168.99.100:27018. The full response is { "errmsg" : "exception: can't $divide by zero", "code" : 16608, "ok" : 0.0 }, cause=Some(Command failed with error 16608: 'exception: can't $divide by zero' on server 192.168.99.100:27018. The full response is { "errmsg" : "exception: can't $divide by zero", "code" : 16608, "ok" : 0.0 })
```

sellout: 😱 So Mongo’s `$divide` expression is just broken, then. I suppose it can be fixed by changing
```scala
case Divide(a, b) => $divide(a, b)
```
to
```scala
case Divide(a, b) =>
  $cond($eq(b, $literal(Bson.Int32(0))), $literal(Bson.Dec(Double.NaN)), $divide(a, b))
```
I’m not sure if `Double.NaN` is actually a thing, but hopefully there’s _some_ way to represent NaN there. If not, `Bson.Undefined` would at least match our semantics, but it wouldn’t test the handling of NaN in the result.

rintcius: Given the urgency of fixing this issue and that the main problem of this PR is in the testing being backend specific, is it an idea to create a separate issue for improving the testing method (and the `divide` fix)? 

Then for this ticket I can just update the prod code so that it returns `undefined` instead of `null` & the issue can be fixed by tomorrow

rintcius: Maybe I'm too pessimistic :) Let me have a go & see how things go with the `divide` fix & only create a separate issue in case it's not working out tomorrow.

sellout: Yeah, I’m fine with making other issues that cover
1. fixing Mongo $divide handling and
2. testing NaN handling across connectors.

Also, if you do fix the `$divide` issue, we should also open an issue against Mongo suggesting that they handle NaN properly rather than exploding the query, and add a comment referencing that issue to that line of code. And then we can hopefully use the simpler case for Mongo >3.4.

rintcius:
Just disabled marklogic for now (I know they fail from local run) - let's see how the other non mongo backends go...

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -49,7 +49,14 @@ object FuncHandler {
           case Add(a1, a2)           => $add(a1, a2)
           case Multiply(a1, a2)      => $multiply(a1, a2)
           case Subtract(a1, a2)      => $subtract(a1, a2)
-          case Divide(a1, a2)        => $divide(a1, a2)
+          case Divide(a1, a2)        =>
+            $cond($eq(a2, $literal(Bson.Int32(0))),
```
sellout: So, you should make an account at https://jira.mongodb.org/secure/Dashboard.jspa and open an issue saying something like “BSON can handle NaN/infinity, but pipeline expressions can’t, so we have to write ugly stuff like this”. Then add a comment like
```scala
// TODO: remove or override this when mongodb issues SERVER-12345 is fixed.
```

sellout: And if you commit that, you can include `[skip ci]` in the commit message so that Travis doesn’t try to recompile everything just because you added a comment.

sellout:
Ok, so looking at those compilation failures …
* Couchbase – create an issue, mark it pending, not sure what’s wrong
* Spark – create an issue, mark it pending, but @rabbitonweb – the problem is that since you’re operating on `Data`, it’s using `BigDecimal`, which can’t represent these “values”, so you should identify them like the Mongo code here does, and just map them all to `Data.Undefined`.
* MongoDB – some other planner tests are expecting the simple `$divide` in the results. I would say that you should try to generate the simple case when we have a `$literal` (that isn’t `0`) for `a2`. I’m not sure if the current implementation of `FuncHandler` allows that, though, because `a2: A`, not `a2: ExprOp`. So you could add a TODO there that we should generate the simple case when possible.

mongodb/src/main/scala/quasar/physical/mongodb/bson.scala
```diff
@@ -71,6 +71,21 @@ object Bson {
     def repr = new BsonDouble(value)
 
     def toJs = Js.Num(value, true)
+
+    override def equals(that: Any): Boolean = that match {
```
rintcius: Needed to change this to make the planner tests pass (took me some time to figure this one out because the diffs in those tests didn't show anything)

rintcius:
@sellout tests are green now & I think I have addressed your comments

rabbitonweb:
@sellout you mean (in spark case) if there is a divide by zero case I should just return undefined?

sellout:
@rabbitonweb Correct.

PR #2260, opened 2017-05-30
MarkLogic BackendModule

### LINKED ISSUE #2092: Implement backend abstraction in marklogic
A stub implementation is currently provided.  Further work needs to be done to remove the phantom type-based configuration from marklogic, so that things can be isolated in `Config`.

wemrysi:
Closes #2092.


connector/src/main/scala/quasar/connector/BackendModule.scala
```diff
@@ -147,6 +150,9 @@ trait BackendModule {
     } yield PhysicalPlan(main, ISet.fromFoldable(inputs))
   }
 
+  final def config: Backend[Config] =
+    MonadReader_[Backend, Config].ask
```
djspiewak: Shouldn't this be something like…

```scala
final def config[F[_]: MonadReader_[?[_], Config]]: F[Config] = ...
```

The logic here is that this would be awfully convenient even in some of the functions which only return `Configured`.

connector/src/main/scala/quasar/connector/ManagedQueryFile.scala
```diff
@@ -0,0 +1,85 @@
...
+
+import slamdata.Predef._
+import quasar.Data
+import quasar.contrib.pathy._
+import quasar.effect._
+import quasar.fs._
+
+import scalaz._, Scalaz._
+
+trait ManagedQueryFile[C] { self: BackendModule =>
```
djspiewak: This factors out a lot of the boilerplate quite nicely.  It's still very difficult to use for any backend which has an internally safe filesystem implementation, but this is a start.

connector/src/main/scala/quasar/connector/ManagedWriteFile.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import slamdata.Predef._
+import quasar.Data
+import quasar.contrib.pathy._
+import quasar.effect._
+import quasar.fs._
+
+import scalaz._, Scalaz._
+
+trait ManagedWriteFile[C] { self: BackendModule =>
```
djspiewak: Once you start down the [dark path](https://www.youtube.com/watch?v=yLbdw06tKPQ), forever will it dominate your destiny…

marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
```diff
@@ -0,0 +1,326 @@
...
+import pathy.Path._
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+import scalaz.stream.Process
+
+final class MarkLogic(readChunkSize: Positive, writeChunkSize: Positive)
+  extends BackendModule
+  with ManagedQueryFile[XccDataStream]
+  with ManagedWriteFile[AFile]
+  with ManagedReadFile[XccDataStream] {
```
djspiewak: Minor note, but… indentation.

marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
```diff
@@ -0,0 +1,326 @@
...
+  def parseConfig(uri: ConnectionUri): DefErrT[Task, Config] =
+    MarkLogicConfig.fromUriString[EitherT[Task, ErrorMessages, ?]](uri.value)
+      .bimap(_.left[EnvironmentError], MLBackendConfig fromMarkLogicConfig _)
+
+  def compile(cfg: Config): DefErrT[Task, (M ~> Task, Task[Unit])] = {
+    val dropWritten = λ[M ~> Free[MarkLogicFs, ?]](_.value)
+    runMarkLogicFs(cfg.cfg.xccUri) map { case (f, c) => (f compose dropWritten, c) }
+  }
+
+  override def interpreter(cfg: Config): DefErrT[Task, (AnalyticalFileSystem ~> Task, Task[Unit])] = {
```
djspiewak: This seems kinda scary.  What's going on here?

marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
```diff
@@ -0,0 +1,326 @@
...
+      if (cfg.cfg.rootDir === rootDir) liftFT[AnalyticalFileSystem]
+      else chroot.analyticalFileSystem[AnalyticalFileSystem](cfg.cfg.rootDir)
+
+    super.interpreter(cfg) map {
+      case (f, c) => (foldMapNT(f) compose xformPaths, c)
+    }
+  }
+
+  def plan[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT](qs: T[QSM[T, ?]]): Backend[Repr] =
+    MonadReader_[Backend, Config].ask >>= { config =>
```
djspiewak: `s/MonadReader_[Backend, Config].ask/config/`

marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
```diff
@@ -0,0 +1,326 @@
...
+      UuidReader[FileSystemErrT[PhaseResultT[Configured, ?], ?]] asks { uuid =>
+        val fname = s"temp-$uuid"
+        refineType(path).fold(
+          d => d </> file(fname),
+          f => fileParent(f) </> file(fname))
+      }
+
+    ////
+
+    private def ifExists[A](path: APath)(thenDo: => Backend[A]): Backend[A] =
```
djspiewak: Minor sidebar… is the laziness actually all that significant here?

marklogic/src/test/scala/quasar/physical/marklogic/fs/MarkLogicConfigArbitrary.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import java.net.URI
+
+import pathy.Path._
+import pathy.scalacheck._
+import org.scalacheck._
+import scalaz.scalacheck.ScalaCheckBinding._
+import scalaz.syntax.apply._
+
+trait MarkLogicConfigArbitrary {
```
djspiewak: :+1:

connector/src/main/scala/quasar/connector/BackendModule.scala
wemrysi: Aye, will update.

marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
wemrysi: No, mostly just passing it through since `ifM` is lazy.

djspiewak: KK

marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
wemrysi: A couple backends do path rewriting to implement a sort of "chroot" feature, i.e. you can mount `/some/path/in/a/db` as `/amount` so `select /amount/foo` becomes `select /some/path/in/a/db/foo` by the time the backend sees it.

All the rewriting code is written as transforms on the algebras, and I was trying to avoid having to duplicate all the rewriting code to work on the traits defined in `BackendModule`.

We could define another mixin to `BackendModule` that does this explicitly and the connectors that need it could use it. Though, I'm not sure if there are other things that used to depend on having access to the interpreter.

djspiewak: Gotcha; cool.

Issue #2258, opened 2017-05-30
ReadFile alegbra does not work via `spark-hdfs`


rabbitonweb:
When querying files via UI everything works, when double clicking on file i get an empty result set

rabbitonweb:
Fixed

Issue #2257, opened 2017-05-30
spark-hdfs delete file does not work


rabbitonweb:
When trying to delete file via `spark-hdfs` file is not being deleted

rabbitonweb:
Fiuxed

PR #2261, opened 2017-05-30
Fix url decode bug, url should be decoded when read not when used

rabbitonweb:
Url is used in multiple places, should be decoded when read, not when used

Fixes both https://github.com/quasar-analytics/quasar/issues/2257 and https://github.com/quasar-analytics/quasar/issues/2258


sellout:
Some day you’ll remember to add a semantic version on your PRs 😄

rabbitonweb:
@sellout I'm close to writing chrome extension ... :(

Issue #2231, opened 2017-05-21
simple join query does not work


doglasfc:
Hi, 
 I'm newcomer to slamdata. just write following query against mondodb 3.4.4.

SELECT b.name as A_NAME, c.c_name as B_CLASS_NAME from `/edb/obj` as b JOIN `/edb/class_obj` AS c ON c._id=b.classobjid

Here:
1. obj and class_obj are two collections undder same database edb.
2. SELECT b.name as A_NAME from `/edb/obj` as b    -----  works, return A_NAME column data
3. SELECT c.c_name as B_CLASS_NAME from `/edb/class_obj` AS c  ---- -----  works, return B_CLASS_NAME column data

always get error as below:

{
  "error": {
    "status": "Malformed SQL^2 query.",
    "detail": {
      "message": "operator '*' expected; but found `*** error: '!' expected but _ found'"
    }
  }
}

it looks slamdata does not support "c._id" as join condition..., but look at the slamdata document, it should work!!!

sellout:
@doglasfc Sorry, I think the docs are slightly out of date with the latest release. The SQL specification requires that identifiers that don’t begin with a letter need to be quoted, so you need to do
```sql
SELECT ... ON c.`_id`=b.classobjid
```

That error message is also unfortunate. It looks like the tokenizer is failing, but then the parser treats the tokenizer’s error message as the unexpected token somehow.

doglasfc:
@sellout Thanks for quick response. actually I did that already. (someone suggested that.)  but no luck, the new error message returned as following:

"  An unknown error occurred: 500 ""  "

It even did very bad thing:  it destroyed mongodb mount info strangely and I'm no longer able to re-create mount again which I always get 
"  An unknown error ocurred: 500 ""  " ,  then I did remove slamdata software and re-install it, I still get same error... I'm using latest slamdata version 4.2.0, what version are you using? Have you tested that?

damonLL:
@doglasfc SlamData will create a `quasar-metastore.mv.db` file which stores the mount information.  You'll need to remove that file since it's likely corrupt now (which I mentioned in the forums).  It's located in the same location as the `quasar-config.json` file, see our [docs](quasar-metastore.mv.db) on where that is.

1. Stop SlamData
2. Remove metastore file
3. Restart SlamData
4. Remount MongoDB

doglasfc:
Thanks for the reply. will do that.  so the key is  following statement is still not working...

[...]

PR #2263, opened 2017-05-31
Fix sql pprint for FieldDeref with StringLiteral

drostron:
Fixes: #2231

### DIFF
--- a/frontend/src/test/scala/quasar/sql/SqlParserSpec.scala
+++ b/frontend/src/test/scala/quasar/sql/SqlParserSpec.scala
@@ -17,16 +17,16 @@
 package quasar.sql
 
 import slamdata.Predef._
-import quasar.RenderTree.ops._
 import quasar.fp._
+import quasar.RenderTree.ops._
+import quasar.specs2.QuasarMatchers._
 import quasar.sql.fixpoint._
 
 import matryoshka._
 import matryoshka.data.Fix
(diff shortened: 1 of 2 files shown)


frontend/src/test/scala/quasar/sql/SqlParserSpec.scala
```diff
@@ -670,9 +670,15 @@ class SQLParserSpec extends quasar.Qspec {
       parsed must beRightDisjOrDiff(node)
     }.set(minTestsOk = 1000) // one cannot test a parser too much
 
-    "round-trip quoted variable names through the pretty-printer" >> {
-      val q = "select * from :`A.results`"
-      (parse(q) map (pprint[Fix[Sql]] _) map (Query(_)) >>= (parse _)) must beRightDisjunction
+    "round-trip through the pretty-printer" >> {
+      def roundTrip(q: String) =
+        (parse(q) map (pprint[Fix[Sql]] _) map (Query(_)) >>= (parse _)) must beRightDisjunction
```
sellout: So, is this definition (which I know is just the old one) good enough? it seems like we want something more like
```scala
val ast = parse(q)
(ast ∘ (pprint >>> (Query(_)))) >>= parse must equal(ast)
```

drostron: A worthy update.

PR #2273, opened 2017-06-01
No longer rely on ML directories for filesystem operations

### LINKED ISSUE #2108: MarkLogic: Zero Setup
There are two serious usability issues with the MarkLogic connector that need to be rectified:

1. MarkLogic must contain one or more directories in the database before documents will be displayed.
2. MarkLogic documents must be located within a directory, which requires that users create a directory, and place their documents inside there.

Both limitations need to be eliminated so that upon configuring a new MarkLogic mount, with mount configuration accepted by Quasar, the user will see all the data in MarkLogic without any additional steps.

wemrysi:
Closes #2108.

### DIFF
--- a/marklogic/src/main/scala/quasar/physical/marklogic/fs/lib.scala
+++ b/marklogic/src/main/scala/quasar/physical/marklogic/fs/lib.scala
@@ -81,9 +81,11 @@ object lib {
     })
 
   // filesystem:directory-contents($uri as xs:string) as xs:string*
-  def directoryContents[F[_]: PrologW: Bind, T: SearchOptions]: F[FunctionDecl1] =
+  def directoryContents[F[_]: PrologW: Bind, T: SearchOptions](
+    nextUri: (XQuery, XQuery) => XQuery
+  ): F[FunctionDecl1] =
     (
-      discoverChildren[F].fn                |@|
+      discoverChildren[F](nextUri).fn       |@|
       descendantsHavingFormatExist[F, T].fn
     ).tupled flatMap { case (discKids, descsExist) =>
       fs.declare[F]("directory-contents") map (_(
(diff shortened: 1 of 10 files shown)


marklogic/src/main/scala/quasar/physical/marklogic/fs/lib.scala
```diff
@@ -175,17 +175,63 @@ object lib {
...
+  private def discoverChildren[F[_]: PrologW: Bind]: F[FunctionDecl2] =
+    (fs.name("discover-children").qn[F] |@| asChildDir[F].fn)((fname, childDir) =>
+      declare(fname)(
+        $("parent") as ST("xs:string"),
+        $("known")  as ST("xs:string*")
+      ) { (parent: XQuery, known: XQuery) =>
+        val (knownq, uri, d) = ($("knownq"), $("uri"), $("d"))
+        let_(
+          knownq := cts.notQuery(cts.orQuery(fn.map(func(d.render) { cts.directoryQuery(~d, "infinity".xs) }, known))),
+          uri    := cts.uriMatch(fn.concat(parent, "*/*".xs), IList("document".xs), some(~knownq))(1.xqy)
```
jsantos17: I think I grok the algorithm, but I'm not sure I understand the `*/*` pattern. What would be some examples of the matches we're looking for?

wemrysi: We're looking for directory children of the parent, so it means URIs of the form `$parent/*/*` i.e. a URI that begins with `$parent` and includes at least one `/`.

marklogic/src/main/scala/quasar/physical/marklogic/fs/lib.scala
```diff
@@ -175,17 +175,63 @@ object lib {
...
 
-  def subDirectories[F[_]: PrologW: Functor]: F[FunctionDecl2] =
-    fs.declare[F]("sub-directories") map (_(
-      $("uri")                as ST("xs:string"),
-      $("includeDescendants") as ST("xs:boolean")
-    ).as(ST("xs:string*")) { (uri: XQuery, includeDescendants: XQuery) =>
-      val depth = if_(includeDescendants) then_ "infinity".xs else_ "1".xs
-      fn.map(fn.ns(NCName("base-uri")) :# 1, xdmp.directoryProperties(uri, depth) `/` child(propProperties) `/` child(propDirectory))
-    })
+  def subDirectories[F[_]: PrologW: Bind]: F[FunctionDecl2] =
```
jsantos17: What do you think about including a (comment) type signature in XQuery of the `subDirectories` and `moveFile` functions? Similar to the ones in [quasar.physical.marklogic.qscript.XmlStructuralPlanner](https://github.com/quasar-analytics/quasar/blob/2dd509faaa9396bd52b759a2bc1a2ea2a441d9a3/marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala#L180). I found them very helpful when figuring out field lookup there.

wemrysi:
@jsantos17 Added a few updates that delete a bunch of now unused code as well as provide a fallback implementation of subdirectory discovery that doesn't require the URI lexicon. Feature detection is performed at runtime and the appropriate implementation chosen based on the availability of the URI lexicon.

I also disabled the URI lexicon in the `quasar_marklogic_json` docker container so we have regression coverage for this capability.

jsantos17:
@wemrysi is there something else we're missing here to merge or are we just waiting for a green build?

wemrysi:
@jsantos17 Just a green build.

Issue #1760, opened 2016-12-06
Non-static default field names result in an invalid QName error


wemrysi:
The current workaround of prefixing `_` to the default numeric field names doesn't work when the name isn't a literal number, like in the following query `select meta(*).award, count(*) from oscars group by meta(*).award`.

Ideally #1642 will be fixed first, rendering this moot.

wemrysi:
The example query given actually works now. I think this is still an issue in theory, but possibly more rare that it was before. If I run into another example during testing I'll add it here.

wemrysi:
Here's one from `groupByFlatten.test` that triggers the issue

```
select substring(parents[*].sha, 0, 1), count(*) from slamengine_commits group by substring(parents[*].sha, 0, 1)
```

jsantos17:
Closed via #2275

Issue #1948, opened 2017-03-07
MarkLogic passing all applicable query regression tests


wemrysi:
A catch-all issue to track the remaining query regression tests that are failing for MarkLogic. Only those that are expected to pass are listed here, connector-specific tests for other connectors are omitted.

Test | Notes
------|--------
caseVariations.test| Ordering after `distinct`.
caseWithGroup.test| Ordering after `distinct`.
citiesByLargestZip.test| Ordering after `distinct`.
different-flattens.test| QScript provenance issue (#1951).
flattenGroupedObject.test| Attempts to string match on an object, something isn't projecting a field properly? (#2359) 
groupByFlatten.test| ~Non-literal numeric field names (#1760)~ Some other issue related to QScript. Several connectors have it disabled. cc @alissapajer
guardedExpression.test| Runtime typechecks (#1506).
joins/3WayJoin.test|
joins/constantOnLeft.test|
joins/crossJoin.test|
joins/groupedJoin.test|
joins/innerEquiJoin.test|
joins/innerEquiJoinWithWildcard.test|
joins/joinSubSelects.test|
joins/joinWithComplexCondition.test|
joins/joinWithMultipleConditions.test|
joins/joinWithReversedCondition.test|
joins/orderJoinByAlias.test|
joins/sameFieldNameComplex.test|
joins/selfjoinConstantOnLeft.test|
joins/selfjoinWithComplexCondition.test|
LocationCount.test| Doesn't look like _any_ connectors (including legacy mongo) pass this one.
multistageFlatten.test| Runtime typechecks (#1506).
multitypeFlatten.test| Runtime typechecks (#1506).
nonJsJoinCondition.test| Due to #2110.
null/nullTestExprsWithMissing.test| Disabled for all connectors.
servlet.test| Disabled for all connectors.
shortCities.test| Ordering after `distinct`.
singleFieldOrderedDistinct.test| Ordering after `distinct`.
symmetricNonJsJoinCondition.test|  Due to #2110.
temporal/filterOnDate.test| Cast arguments to `temporalTrunc`.
temporal/timestampsAndIntervals.test| Cannot divide using `xs:duration`, need to narrow to one of the subtypes.
tripleFlatten.test| Fail to distinguish between array/object for flatten.
unifyFlattenedFieldsWithMultipleUses.test| `loc` fields not filtered properly. #2366 LogicalPlan bug.
flattenArray.test|Related to #2188. Skipped for `marklogic_xml`
[...]

PR #2275, opened 2017-06-02
Support non-QName's in XmlStructuralPlanner

### LINKED ISSUE #1642: Support map keys that aren't valid QNames in MarkLogic
The MarkLogic planner can't currently handle field names/aliases that aren't valid QNames which causes problems with the default numeric field aliases and any user-defined names that don't fall into the QName category.

The most promising idea for solving this thus far is to use an XML attribute to store the field name when it isn't a valid QName. This will require the planner for `ProjectField` to take this into account as well as serialization to/from `Data`.

jsantos17:
Fixes #1642, #1760


djspiewak:
@jsantos17 Can you assign this PR to someone?  Also it needs a `version:` label.

djspiewak:
@jsantos17 Did you rebase the commits?

jsantos17:
@djspiewak indeed. I needed to get some changes from 2286.

djspiewak:
@jsantos17 You should try to merge in that case, instead of rebasing.  Or at least, rebasing the commits that are already pushed.  Once you've pushed commits to your fork (on a non-`wip/` branch), and *especially* once you've opened a pull request, those commits are no longer something that you should consider mutable.  Someone else may have merged your branch locally to facilitate their own work, and rebasing breaks that, which is why we don't do it.

tl;dr: Don't rebase unless the affected commits *only* exist locally or on `wip/` branches.

jsantos17:
@djspiewak Cool, I'll use merge from now on. I make really sure nobody else is using my branch before I rebase/make any destructive changes. Didn't know about the `wip/` convention. I'll use it on my next issue. Thanks for telling me!

djspiewak:
> I make really sure nobody else is using my branch before I rebase/make any destructive changes

This is basically the motivation for the `wip/` convention (and more notably, for not rebasing on non-`wip/`): you really have no way of knowing for absolutely sure.  I may have grabbed the changes locally, but without telling anyone.  I wouldn't discover the issue until I hit merge conflicts trying to PR my changes to master.

jsantos17:
@djspiewak sounds like a good convention. Maybe it's worth mentioning in the "SlamData Development Process" document? The workflow in the document suggests using rebase too, so it probably needs updating to reflect the conventions we actually use.

djspiewak:
Yeah, that document is incredibly out of date.  We should definitely update it.

alissapajer:
@jsantos17 arg yeah sorry about that document :( I just updated it, modifying the sections on rebasing and merging PRs, so it should be correct now. But if anything still looks wrong, let me know (or just fix it :) )

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -187,14 +215,48 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
...
+      $("name") as ST("xs:string")
+    ).as(ST("element()?")) { (src: XQuery, name: XQuery) =>
+      val (s, n, e) = ($("s"), $("n"), $("e"))
+      fn.map(func(s.render) {
+        element { fn.nodeName(~s) } {
+          mkSeq_(
+            ~s `/` axes.attribute.node(),
+            for_ (n in (~s `/` child.element()))
+            .return_(
+              typeswitch(~n)(
```
jsantos17: I need to refactor this to extract the `typeswitch` into a function.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -321,4 +403,8 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
           .else_(~ascribed)
         })
     })
+
+
+  private def handleWithF[F[_]: Applicative](l: F[XQuery], r: F[XQuery]): F[XQuery] =
+    (l |@| r)((left, right) => try_(left).catch_($("_"))(κ(right)))
```
jsantos17: @wemrysi I wonder if there's a better way to check if a name is a `QName`. My current approach is trying to call `xs:QName` and catch the exception if it fails, but it catches any exception, not only those caused by invalid QName's. Do you think that might cause problems? Do you know about other ways to check if the string is a valid QName in XQuery?

wemrysi: Hm, I was going to suggest using `is castable` but there is an exception in the semantics of that expression that specifically state, when testing for `xs:QName`, it only works for string literals :(

I guess `try/catch` will work for now, maybe wrap it up in a function like `declare isQName($candidate as xs:string) as xs:QName?` which should help to localize the `try/catch` to just the `QName` test and reduce false-positives.

Also, I think you may need to use `fn:QName`, in general, as the docs allude to `xs:QName` only working on string literals.

jsantos17: I'll wrap the try catch. I did see the documentation on `xs:QName`, but after some testing in the Marklogic console it appears to work fine. This works despite the warning in Marklogic documentation:

```xquery
xs:QName(fn:concat("aaa", xs:string(12)))
```

[XQuery 1.0](https://www.w3.org/TR/xquery/#id-constructor-functions) seems to have the restriction on a string literal for `xs:QName`, but no such restriction appears in [XQuery 3.1](https://www.w3.org/TR/xpath-functions-31/#constructor-qname-notation). Anyway, I think I'll use `fn-QName`, since [Marklogic says that they only support XQuery 1.0](https://docs.marklogic.com/xdmp:xquery-version).

wemrysi: Oh, that's great 👍

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -165,14 +181,21 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
       $("obj2") as ST("element()?")
     ).as(ST(s"element()")) { (obj1: XQuery, obj2: XQuery) =>
       val (xs, ys, names, e, n1, n2) = ($("xs"), $("ys"), $("names"), $("e"), $("n1"), $("n2"))
+      val (elt, encodedNames) = ($("elt"), $("encodedNames"))
 
       mkObjectFn {
         let_(
           xs    := (obj2 `/` child.element()),
           names := fn.map("fn:node-name#1".xqy, ~xs),
+          encodedNames := ~xs `/` axes.attribute.attributeNamed(ejsonEncodedAttr.shows),
```
jsantos17: @wemrysi this is the only place where I check for encoded elements all the time. We could check that every key in each object is a string literal, but not sure if worth it. What do you think?

jsantos17:
This is almost done. Only missing is a test for keys not known statically. Would you mind checking again @wemrysi?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -321,4 +401,30 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
...
+
+  // ejson:guard-qname($candidate as item()*, $if-qname as item()*, $if-non-qname as item()*) as item()*
+  lazy val guardQName: F[FunctionDecl3] =
+    ejs.declare[F]("guard-qname") flatMap (_(
+      $("candidate") as ST("item()*"),
+      $("if-qname") as ST("(function() as item()*)"),
+      $("if-non-qname") as ST("(function() as item()*)")
+    ).as(ST.Top) { (candidate: XQuery, ifQName: XQuery, ifNonQName: XQuery) =>
+      isQName(candidate) map (if_(_)
+      .then_(ifQName.fnapply())
```
jsantos17: This is kind of ugly, but a lot more performant than `try-catch`

it/src/test/scala/quasar/physical/marklogic/qscript/XmlStructuralPlannerSpec.scala
```diff
@@ -80,6 +80,38 @@ final class XmlStructuralPlannerSpec
       }
     }
 
+    "objectMerge" >> {
+      "merges non-QName keys" >> prop { (x: Data, y: Data) =>
```
wemrysi: I think these specific tests here are good, but I also think we can update the general `StructuralPlannerSpec` to make fewer assumptions about key names, specifically I think we should get rid of the `asMapKey` abstract method that requires all the map keys to be `QNames`.

It might also be worth trying generating more key names instead of using fixed ones for more coverage.

wemrysi:
@jsantos17 Does the `projectFieldWithFlatten` regression test pass now (the comment in #1948) seems to indicate it should.

jsantos17:
@wemrysi projectFieldWithFlatten is working. It was fixed in my previous PR [right here](https://github.com/quasar-analytics/quasar/commit/e2f4499186ee32af7a61c456f6a93a1a3759a622). Looks like I forgot to remove it from #1948.

it/src/test/scala/quasar/physical/marklogic/qscript/XmlStructuralPlannerSpec.scala
jsantos17: Sounds like a good idea. I'll start implementing it. Shouldn't take long.

it/src/test/scala/quasar/physical/marklogic/qscript/StructuralPlannerSpec.scala
```diff
@@ -82,6 +80,39 @@ abstract class StructuralPlannerSpec[F[_]: Monad, FMT](
...
+  implicit val arbGenStrKey = Arbitrary(genStrKey)
+
+  implicit val mapEntry: Arbitrary[(String, Data)] =
+    Arbitrary((genStrKey |@| arbitrary[Data])((key, value) =>
+      (key.asString, value)))
+
+  val genNonEmptyObj: Gen[ListMap[String, Data]] =
+    arbitrary[NonEmptyList[(String, Data)]] map (nel => ListMap(nel.toList: _*))
+
+  implicit val arbNonEmptyObj: Arbitrary[ListMap[String, Data]] =
```
jsantos17: @wemrysi These implicits are a little too general, do you think it'd be best if we tag/newtype them?

wemrysi: Yep.

it/src/test/scala/quasar/physical/marklogic/qscript/StructuralPlannerSpec.scala
```diff
@@ -173,87 +201,85 @@ abstract class StructuralPlannerSpec[F[_]: Monad, FMT](
...
-        val e1 = keyed(vs1).toList
-        val e2 = keyed(vs2).toList
-        val o  = Data._obj(ListMap(e1.drop(e2.length) ::: e2 : _*))
-        evalF((lit(Data._obj(ListMap(e1: _*))) |@| lit(Data._obj(ListMap(e2: _*))))(
-          SP.objectMerge).join) must resultIn(o)
+      "right biased when keys overlap" >> prop { (testCase: TestCase, values: NonEmptyList[Data]) =>
+        val left = testCase.obj
+        val right = ListMap(testCase.obj.keys.zip(values.toList).toList: _*)
+
+        val commonKeys = left.keySet & right.keySet
```
jsantos17: This is a naive implementation of a right biased merge on Scala maps. We can either keep this implementation or depend on the fact that [`++` in ListMap](http://www.scala-lang.org/api/2.11.8/index.html#scala.collection.immutable.ListMap@++[B](that:scala.collection.GenTraversableOnce[B]):scala.collection.immutable.ListMap[B]) is right biased. However, I couldn't find documentation that states that. I'm not sure we want to depend on that if it's undocumented.

I could remove this implementation and simplify the test if we're sure `++` will always be right biased.

@wemrysi wdyt?

wemrysi: You're right, it's unfortunate that the documentation referenced is devoid of these sort of specifics (probably due to the inheritance of `++` from superclasses). I think, as you noted, the implication is toward right-bias.

Also, we're [already relying on it being right-biased](https://github.com/quasar-analytics/quasar/blob/d695c0c5d19072da9f5c8e08bfbc923db4d84e8f/frontend/src/main/scala/quasar/std/structural.scala#L89) so I think we should just use `++`.

jsantos17: Oh and in non-test code! Oh well. I have replaced the implementation with `++`.  This is ready for another look too.

jsantos17:
@wemrysi not sure if the other message sent a notification, but this is ready for another look.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -187,14 +217,48 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
       }
     })
 
-  // qscript:children-named($src as element()?, $name as xs:QName?) as item()*
-  lazy val childrenNamed: F[FunctionDecl2] =
-    ejs.declare[F]("children-named") map (_(
+  // ejson:children-named-encoded($src as element()?, $field as xs:string?) as item()*
+  lazy val childrenNamedEncoded: F[FunctionDecl2] =
```
wemrysi: I think we could convert this into an expression instead of a function.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -240,6 +304,29 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
       })
     }
 
+  // ejson:rename-or-wrap-encoded($name as xs:string, $value as item()*) as element()
+  lazy val renameOrWrapEncoded: F[FunctionDecl2] =
```
wemrysi: I don't think this will work when `$value` is itself an element with an encoded name due to conflicting attributes, have you tested this scenario?

I think we need to filter out the `ejsonEncodedName` attribute from the list of attributes. Actually, I think we need to do this always, for the encoded _and_ the non-encoded versions. If we're renaming an encoded element to something that is a valid `QName`, we don't want the `ejsonEncodedName` attribute to be present in the result.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
```diff
@@ -56,11 +55,6 @@ package object qscript {
   val EJsonTypeKey  = "_ejson.type"
   val EJsonValueKey = "_ejson.value"
 
-  /** Converts the given string to a QName if valid, failing with an error otherwise. */
-  def asQName[F[_]: MonadPlanErr: Applicative](s: String): F[QName] =
```
wemrysi: I think this means you can also get rid of the private `IntegralNumber` and `encodeForQName` values at the end of this file.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
jsantos17: Good catch! I haven't tested that case. I'll check it and get back to you.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
jsantos17: So just inline the path expression instead of defining a function and making the function call?

wemrysi: Yeah, that way we'll remain in XPath as long as possible.

jsantos17:
@wemrysi this is ready for another look.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
jsantos17: Turns out that it'd indeed a problem. It's fixed now and I added some related tests for good measure.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -217,20 +225,9 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
       }
     })
 
-  // ejson:children-named-encoded($src as element()?, $field as xs:string?) as item()*
-  lazy val childrenNamedEncoded: F[FunctionDecl2] =
```
wemrysi: This could still be a scala function, `(XQuery, XQuery) => XQuery`, to avoid the duplication.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
```diff
@@ -299,7 +296,9 @@ private[qscript] final class XmlStructuralPlanner[F[_]: Monad: MonadPlanErr: Pro
       ).as(ST(s"element()")) { (name: XQuery, value: XQuery) =>
         typeswitch(value)(
           $("e") as ST("element()") return_ (e =>
-            element { name } { mkSeq_(e `/` axes.attribute.node(), e `/` child.node()) })
+            element { name }
+              { mkSeq_(e `/` axes.attribute.node()(fn.name() =/= ejsonEncodedAttr.shows.xs),
```
wemrysi: Should factor this into a scala function to dry up the duplication.

jsantos17: @wemrysi Done

marklogic/src/main/scala/quasar/physical/marklogic/qscript/XmlStructuralPlanner.scala
jsantos17: Done

PR #2278, opened 2017-06-04
Fix improper parsing of modules statement blobs

jedesah:
It would appear, one needs to always call `normalize` after parsing SQL. This was not being done when parsing statements when loading a module mount.


jedesah:
@drostron The build is green and this should be a pretty quick review. If you merge it soon I won't have to add a commit (or rewrite history) to another PR to remove this commit 😄

PR #2279, opened 2017-06-04
Add support for `offset` and `limit` to the `invoke` endpoint

jedesah:


web/src/main/scala/quasar/api/services/invoke.scala
```diff
@@ -41,27 +41,29 @@ object invoke {
     S1: Failure[Module.Error, ?] :<: S
   ): QHttpService[S] = QHttpService {
 
-    case req @ GET -> AsPath(path) =>
-      respond_ {
+    case req @ GET -> AsPath(path) :? Offset(offsetParam) +& Limit(limitParam) =>
+      respond_ {(offsetOrInvalid(offsetParam) |@| limitOrInvalid(limitParam)){ (offset, limit) =>
```
djspiewak: I try not to leave comments like this… but!  Reformat please.  This needs to be wrapped to the next line and indented.  Similarly, the `}}` atrocity below should be split while the entire block is reindented.

web/src/main/scala/quasar/api/services/invoke.scala
```diff
@@ -41,27 +41,29 @@ object invoke {
...
   }
 
   ////
 
   private def invoke[S[_]](
     format: MessageFormat,
     filePath: AFile,
-    args: Map[String, String]
+    args: Map[String, String],
+    offest: Natural,
```
djspiewak: `offest`

core/src/main/scala/quasar/fs/mount/MountConfig.scala
```diff
@@ -85,7 +85,7 @@ object MountConfig {
     case ("view", uri) =>
       viewCfgFromUri(uri).map(i => viewConfig(i))
     case ("module", stmts) =>
-      sql.fixParser.parseWithParser(stmts, sql.fixParser.statements).bimap(
+      sql.fixParser.parseModule(stmts).bimap(
```
djspiewak: Did you mean to include this change?

web/src/main/scala/quasar/api/services/invoke.scala
```diff
@@ -41,14 +41,16 @@ object invoke {
...
-    case req @ GET -> AsPath(path) =>
+    case req @ GET -> AsPath(path) :? Offset(offsetParam) +& Limit(limitParam) =>
       respond_ {
-        refineType(path).fold(
-          dir => apiError(BadRequest withReason "Path must be a file").toResponse[S],
-          file => {
-            val requestedFormat = MessageFormat.fromAccept(req.headers.get(Accept))
-            invoke[S](requestedFormat, file, req.params)
-          })
+        (offsetOrInvalid(offsetParam) |@| limitOrInvalid(limitParam)) { (offset, limit) =>
```
djspiewak: Thanks!

core/src/main/scala/quasar/fs/mount/MountConfig.scala
jedesah: Ah 😞 . That's what happens when you juggle small commits...

djspiewak: Do you want this merged despite the double-inclusion?

jedesah: I'm just lazy and this problem will go away on its own once the other PR is merged 😄 . Then again, if it looks good to you, feel free to merge, I could have just as easily assigned you to the other PR as well.

jedesah:
So I don't know what's up with github but the commit that got accidentally added is now in `master`...

PR #2280, opened 2017-06-04
Upgrade http4s to `0.15.13a` (latest stable release)

jedesah:


project/Dependencies.scala
```diff
@@ -9,14 +9,15 @@ import slamdata.CommonDependencies
...
+  private val argonautVersion          = "6.2"
   private val disciplineVersion        = "0.5"
   private val jawnVersion              = "0.10.4"
   private val jacksonVersion           = "2.4.4"
   private val matryoshkaVersion        = "0.18.3"
   private val pathyVersion             = "0.2.9"
   private val raptureVersion           = "2.0.0-M6"
   private val scodecBitsVersion        = "1.1.0"
-  private val http4sVersion            = "0.15.9a"
+  private val http4sVersion            = "0.15.13a"
```
alissapajer: Can we update the argonaut and http4s versions in sbt-slamdata and default to those instead of including them here? Both libraries are already included in [`CommonDependencies`](https://github.com/slamdata/sbt-slamdata/blob/master/src/main/scala/slamdata/CommonDependencies.scala#L13) and it looks like the TODO comment there would be addressed by this upgrade.

jedesah: After a discussion with @dave and @sellout, my inclination would be to avoid using `sbt-slamdata` as a means to synchronize dependency versions between `quasar` and `quasar-advanced` which is why I did this change the way I did. It's still on my plate to update `sbt-slamdata` to remove all the `quasar` specific dependencies.

alissapajer: I'm a little confused. So if we're not using sbt-slamdata to synchronize dependencies, and we're removing all quasar-specific dependencies from it, then why do we have it?

jedesah: I'm not necessarily familiar with everything it does but I think the most important role is reducing duplication of the build definition in each SlamData project. There are things like publishing, organization name, etc which are the same across all our projects

alissapajer: Yeah, that makes sense. I think we just need to formalize when we include a dependency in sbt-slamdata versus when we include it in the specific repo.

drostron: In light of sbt-slamdata's value being mostly from shared build definition and publishing aspects, we'll likely remove dependency declarations from sbt-slamdata over time. It's been suggested that we could have a separate plugin for dependencies if there's sufficient benefit.

jedesah:
Looks like it was just a timeout, but not sure. Restarted and we'll see.

PR #2283, opened 2017-06-05
Travis will only output failing tests

jedesah:
This will make it easier to quickly see which test failed on Travis


djspiewak:
Big :+1: to this.  One concern though: exactly one of the builds failed with an error that seems related, and I'm not sure exactly why it was only one.  Mind taking a quick look and letting me know whether you think it's a fluke?

jedesah:
Looks like I missed an occurrence of `test`. I fail to understand why `sbt` doesn't allow one to pass arguments to the `test` command...

PR #2285, opened 2017-06-06
Vastly superior ??? operator

djspiewak:
This replaces the `???` operator with something which provides the following two features:

- Type inference based on call site
- Line information

For example:

```
scala> val x: Int = ???
scala.NotImplementedError: unimplemented value of type Int at <console>:14
  at quasar.package$.$qmark$qmark$qmark(package.scala:110)
  ... 42 elided

scala> def x[A]: List[A] = ???
x: [A]=> List[A]

scala> x[String]
scala.NotImplementedError: unimplemented value of type scala.List[A] at <console>:14
  at quasar.package$.$qmark$qmark$qmark(package.scala:110)
  at .x(<console>:14)
  ... 42 elided
```

It's much more interesting if you do it in an actual file, since you'll get the filename instead of `<console>`, but you get the idea.  Had to bring in scalactic for the `Position` macro (thanks, @bvenners!).

This should eventually go into `slamdata.Predef`, but @alissapajer and I actually need this sooner rather than later, so I thought I would drop it into core.

### DIFF
--- a/core/src/main/scala/quasar/package.scala
+++ b/core/src/main/scala/quasar/package.scala
@@ -28,6 +28,10 @@ import quasar.std.StdLib.set._
 import matryoshka._
 import matryoshka.data.Fix
 import matryoshka.implicits._
+import org.scalactic.source.Position
+import scala.NotImplementedError
+import scala.annotation.unchecked.uncheckedVariance
+import scala.reflect.runtime.universe.WeakTypeTag
 import scalaz._, Scalaz._
 
 package object quasar {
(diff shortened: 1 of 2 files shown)


sellout:
In general, code that we hope to push upstream goes in foundation/…/quasar/contrib/<library>

djspiewak:
@sellout I'd be fine with that.  This is more of a general utility though, so I thought I would just stuff it where everyone would automatically inherit it for now, and then push it into `Predef` later.

PR #2287, opened 2017-06-07
Un-static the Precog cake

### LINKED ISSUE #2290: Resolve dependent typing questions for non-static Precog
Currently in master, it is impossible to make `Precog` a class, since doing so would break `Mimir#Repr` and make it impossible to typecheck transformations, as well as several other things.

djspiewak:
This PR was originally designed to make `Precog` unloadable.  It does that, but there appears to be a strange bug with the filesystem handlers (@wemrysi ??) where the `close` effect is not being run.  More research into this is needed.

In the meantime, the cake is dynamic now, so we're in better shape even without unloading.

Fixes #2290

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/DepFn1.scala
+++ b/mimir/src/main/scala/quasar/mimir/DepFn1.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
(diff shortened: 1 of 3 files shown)


wemrysi:
@djspiewak Any idea which `close` isn't being run? (i.e. from a read, a write, a query?)

wemrysi:
nvm, just saw #2288

Issue #2201, opened 2017-05-16
Sometimes connector setup fails to work in Travis


djspiewak:
I've seen this a couple times now.  It's a new source of transient build failure.  I'm assuming it's the docker containers not quite working?

```
42.11s$ ./initContainer quasar_$CONNECTOR
Building quasar_marklogic_xml
Step 1 : FROM patrickmcelwee/marklogic-dependencies:8-latest
8-latest: Pulling from patrickmcelwee/marklogic-dependencies
Digest: sha256:b99a2e177875d21f0240e567510d625cd48fa08ccf135b698897b22e9504794e
Status: Downloaded newer image for patrickmcelwee/marklogic-dependencies:8-latest
 ---> c9b5564ce54c
Step 2 : ENV ML_VERSION "8.0-5.8"
 ---> Running in b090416a7bdf
 ---> 2a75ece88c31
Removing intermediate container b090416a7bdf
Step 3 : ENV ML_DEV_URL "https://developer.marklogic.com"
 ---> Running in 613f59f6ee30
 ---> 8bd3a64a4cf7
Removing intermediate container 613f59f6ee30
Step 4 : ENV ML_DEV_LOGIN "${ML_DEV_URL}/login"
 ---> Running in 277706213dd0
 ---> 3e2cc381251f
Removing intermediate container 277706213dd0
Step 5 : ENV ML_DEV_GET_TOKEN "${ML_DEV_URL}/get-download-url"
 ---> Running in 222aee81c50a
 ---> 747be52f3729
Removing intermediate container 222aee81c50a
Step 6 : ENV ML_DEV_COOKIES ".ml-dev-cookies"
 ---> Running in 32ae60ec6973
 ---> 0f38121c9716
Removing intermediate container 32ae60ec6973
Step 7 : ENV ML_ASSET_PATH "/download/binaries/8.0/MarkLogic-RHEL6-${ML_VERSION}.x86_64.rpm"
 ---> Running in 9a762c73cc87
 ---> 7f20d2f3f9ce
Removing intermediate container 9a762c73cc87
Step 8 : ENV ML_DOWNLOAD_EMAIL "quasarci@slamdata.com"
 ---> Running in b7dc12b8c1af
 ---> 95ddce2e53bf
Removing intermediate container b7dc12b8c1af
Step 9 : ENV ML_DOWNLOAD_PASSWORD "quasarci"
 ---> Running in d2916c8cb18e
 ---> 5b47cdc7c823
[...]

Issue #2241, opened 2017-05-24
marklogic integration tests throw Memory allocation error


alissapajer:
Sometimes when running marklogic integration tests some will error as follows. I've seen the error on multiple tests, not just this particular one. This is from `marklogic_json`. It also errors similarly on `marklogic_xml`.

The errors are sporadic. Usually these tests will pass without problem on travis, but occasionally I've seen this error. I have only seen the error on travis; I have never seen it locally.

```
[error] ! distinct * [./distinctStar.test] (1 minute 19 seconds, 521 ms)
[error]  java.lang.RuntimeException: Failed loading test data './cities.data': SVC-MEMALLOC: Memory allocation error: mmap 357564416 bytes: Cannot allocate memory on line 0
[error]  ---BEGIN QUERY---
[error]  xquery version "1.0-ml";
[error]  
[error]  declare namespace ejson = "http://quasar-analytics.org/ejson";
[error]  declare namespace filesystem = "http://quasar-analytics.org/filesystem";
[error]  
[error]  declare function ejson:json-deserialize($serialized as item()?) as node()? {
[error]    if (fn:empty($serialized)) then () else xdmp:to-json($serialized)/child::node()
[error]  };
[error]  
[error]  declare function ejson:json-encode($item as item()?) as item()* {
[error]    typeswitch($item)
[error]    case text() return $item
[error]    case boolean-node() return $item
[error]    case number-node() return $item
[error]    case array-node() return $item
[error]    case object-node() return $item
[error]    default return let $t := ejson:type-of($item) return if ($t eq "string") then $item else if ($t eq "boolean") then boolean-node{$item} else ejson:json-encode-type($item)
[error]  };
[error]  
[error]  declare function ejson:json-encode-type($item as item()?) as object-node() {
[error]    ejson:json-deserialize(let $t := ejson:type-of($item) return if (fn:empty($t)) then map:entry("_ejson.type", "na") else if ($t eq "null") then map:entry("_ejson.type", "null") else map:new((map:entry("_ejson.type", $t), map:entry("_ejson.value", $item))))
[error]  };
[error]  
[error]  declare function ejson:json-node-type($node as node()) as xs:string? {
[error]    typeswitch($node)
[error]    case text() return "string"
[error]    case null-node() return "null"
[error]    case boolean-node() return "boolean"
[error]    case number-node() return "decimal"
[error]    case array-node() return "array"
[error]    case $obj as object-node() return ejson:json-object-type($obj)
[...]

PR #2289, opened 2017-06-07
Actually unmount filesystems when exiting repl/server

### LINKED ISSUE #2246: Fix Precog shutdown
Right now, the actor networks are never shut down, meaning that things like `repl/run` will actually hang when you try to exit if you have initialized mimir (typically by cd'ing into the mount).  This is solvable simply by making `Precog` shut down, but this is a little sticky because `Precog` is an object and not a class.  We'll have to figure out some dependent typing stuff (notably relative to `Mimir#Repr`) to get lifecycle stuff.

### LINKED ISSUE #2288: FileSystem close effect isn't being run
This is a continuation of what I referenced in #2287.  It's not clear if this is a `Precog`-specific issue or something more general, though I tend to think it's more general.  Printlns indicate that the `Mimir` `close` action is not being run when I run `repl/run`, `cd mimir` followed by `exit`.  The `exit` will just hang, because the `Precog` actor system never gets shut down, and as I said, printlns indicate that the `Precog` close effect is never sequenced.

I'm guessing that the other connectors just don't exhibit such dramatic issues when they are not shut down properly.  Either that, or I'm derpy and messed up something.

wemrysi:
Fixes #2288.
Fixes #2246.

### DIFF
--- a/repl/src/main/scala/quasar/repl/Main.scala
+++ b/repl/src/main/scala/quasar/repl/Main.scala
@@ -52,42 +52,48 @@ object Main {
   type DriverEff[A]  = Coproduct[ReplFail, DriverEff0, A]
   type DriverEffM[A] = Free[DriverEff, A]
 
-  private def driver(f: Command => Free[DriverEff, Unit], e: Task[Unit]): Task[Unit] = Task.delay {
-    val console =
-      new Console(new SettingsBuilder()
-        .parseOperators(false)
-        .enableExport(false)
-        .interruptHook(new InterruptHook {
-          def handleInterrupt(console: Console, action: Action) = {
-            console.getShell.out.println("exit")
-            e.unsafePerformSync
-            console.stop
+  private def driver(f: Command => Free[DriverEff, Unit], e: Task[Unit]): Task[Unit] = {
+    def shutdownConsole(c: Console): Task[Unit] =
+      Task.delay(c.getShell.out.println("Exiting...")) *>
+      e.attempt                                        *>
+      Task.delay(c.stop)
+
+    Task delay {
+      val console =
+        new Console(new SettingsBuilder()
+          .parseOperators(false)
+          .enableExport(false)
+          .interruptHook(new InterruptHook {
+            def handleInterrupt(console: Console, action: Action) =
+              shutdownConsole(console).unsafePerformSync
+          })
+          .create())
+
+      console.setPrompt(new Prompt("💪 $ "))
+
+      val i: DriverEff ~> MainTask =
+        Failure.toError[MainTask, String]                  :+:
+        liftMT[Task, MainErrT].compose(consoleIO(console)) :+:
+        liftMT[Task, MainErrT]
+
+      console.setConsoleCallback(new AeshConsoleCallback() {
+        override def execute(input: ConsoleOperation): Int = {
+          Command.parse(input.getBuffer.trim) match {
+            case Command.Exit =>
+              shutdownConsole(console).unsafePerformSync
+
+            case command      =>
+              f(command).foldMap(i).run.unsafePerformSync.valueOr(
+                err => console.getShell.out.println("Quasar error: " + err))
           }
-        })
-        .create())
-    console.setPrompt(new Prompt("💪 $ "))
-
-    val i: DriverEff ~> MainTask =
-      Failure.toError[MainTask, String]                  :+:
-      liftMT[Task, MainErrT].compose(consoleIO(console)) :+:
-      liftMT[Task, MainErrT]
-
-    console.setConsoleCallback(new AeshConsoleCallback() {
-      override def execute(input: ConsoleOperation): Int = {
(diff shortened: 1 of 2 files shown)


repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -52,42 +52,48 @@ object Main {
...
-        .enableExport(false)
-        .interruptHook(new InterruptHook {
-          def handleInterrupt(console: Console, action: Action) = {
-            console.getShell.out.println("exit")
-            e.unsafePerformSync
-            console.stop
+  private def driver(f: Command => Free[DriverEff, Unit], e: Task[Unit]): Task[Unit] = {
+    def shutdownConsole(c: Console): Task[Unit] =
+      Task.delay(c.getShell.out.println("Exiting...")) *>
+      e.attempt                                        *>
```
djspiewak: Conceptual niggle…  Shouldn't this be `>>` instead of `*>`?  Both work, but only because `Task` happens to consider hidden evaluation semantics to be part of its substitutability.  It mightn't work with another monad, even a lawful one.

wemrysi: Hm, wouldn't that imply that `*>` and `>>` would disagree for such a monad? I haven't walked it myself, but I've heard madness lies down that path.

However, maybe `>>` conveys the intent a bit better, in that we want a monad here, not necessarily just an applicative. I'm fine with changing it.

djspiewak: It's not a big deal, but I do think it conveys the intent better.

`*>` and `>>` may disagree, modulo the definition of substitutability for a given datatype.  `Task` happens to define substitutability to include threading semantics, which is the only sane definition for `Task` since it is a monad for controlling concurrency and threading semantics, but other monads may not have such a definition.

Basically, what it comes down to is the meaning of `=` in the following law (from the `Monad` docs):

```haskell
(>>) = (*>)
```

Gödel is flipping in his grave.

djspiewak:
This is probably ready for merge, since the build passed modulo the weird marklogic setup issues.

drostron:
Have we been merging in that case? I've been waiting for the author to get a clean build in place.

djspiewak:
I guess that's probably something we should decide as a team.  I've been waiting for build #1 and a plurality of non-1 builds.  Couchbase and Marklogic are so flaky, it's really hard to wait on them for PR merging.

alissapajer:
If there is any chance that a build could fail in backend-build for a legit reason, we should definitely wait for the full travis build of all jobs. 

If we're at the stage where we're merging before the build is fully green, then maybe it's time to actually address the flakiness on travis?

drostron:
Is the CB job flakey? We should file an issue around it if so. I've seen it be fairly consistent recently, but you've been running more builds lately.

alissapajer:
@drostron I don't recall seeing any CB bulids fail recently on travis. And we have #2241 open for ML.

And agreed, if we see CB fail, we should open a ticket, because I don't see any open tickets for CB build failing sporadically.

drostron:
_Restarting ML job_

djspiewak:
@alissapajer There are two problems right now with Marklogic on Travis: the initial docker setup (which, afaik, Nic is looking into), and the fact that some of the integration tests use more memory than the build VMs have available, causing the build to hang (and get ultimately killed).  We can't really do anything about that issue, though polyrepoing will improve the situation considerably since it will be surrounded by a considerably less memory-intensive build.  At the very least, it would no longer block quasar merges.

alissapajer:
@djspiewak so if there are some integration tests enabled in marklogic_xml that are not enabled in marklogic_json, and someone made a change to qscript that broke marklogic_xml, then only the xml build would be affected, and the json build would still pass.

alissapajer:
@nicflores @djspiewak do we have an issue created for the markogic initial docker setup problem that we see on travis? We should create one if we don't.

djspiewak:
@alissapajer Are there any such tests though?  It's a good example, though.  For a qscript-changing PR, I would generally wait for the integration tests to fully pass, but other stuff (e.g. things in mimir), there doesn't seem to be a point.

Agreed that we should fix the underlying situation.  This is more a question of how do we staunch the bleeding in the interim.

alissapajer:
@djspiewak I have seen marklogic integration tests enabled in only one of json or xml.

alissapajer:
A fair rule of thumb could be: If marklogic depends on a project you touched, both marklogic builds must pass before merge.

nicflores:
@alissa, I don't think so, at least not currently. Previously #2201 was
opened to address possibly the same error. Should we just re-open that
#2201?

alissapajer:
@nicflores yeah, if we didn't actually fix #2201 then I'd say we should reopen it.

alissapajer:
@nicflores though actually, are we sure it's that same error? I've mostly (entirely?) seen the out of memory error from marklogic, so maybe we should hold off on reopening #2201 until we're sure it's actually the same problem.

PR #2291, opened 2017-06-07
Banish `PrecogUnit` to the dustbin of bad ideas

### LINKED ISSUE #2164: Remove PrecogUnit
Oh look: we have `-Ywarn-on-value-discard`!  How cool!  Why did we have `PrecogUnit` again?? :-/

djspiewak:
Fixes #2164

### DIFF
--- a/project/Precog.scala
+++ b/project/Precog.scala
@@ -6,7 +6,14 @@ import scala._, Predef._
 object Build {
   val BothScopes = "compile->compile;test->test"
 
-  def warningOpts = Seq("-g:vars", "-deprecation", "-unchecked", "-Ywarn-unused", "-Ywarn-unused-import", "-Ywarn-numeric-widen")
+  def warningOpts = Seq(
+    "-g:vars",
+    "-deprecation",
+    "-unchecked",
+    "-Ywarn-unused",
+    "-Ywarn-unused-import",
+    "-Ywarn-numeric-widen",
+    "-Ywarn-value-discard")
 
   /** Watch out Jonesy! It's the ol' double-cross!
    *  Why, you...
--- a/blueeyes/src/main/scala/quasar/precog/util/cache/Cache.scala
+++ b/blueeyes/src/main/scala/quasar/precog/util/cache/Cache.scala
@@ -66,7 +66,7 @@ object Cache {
     def apply(builder: CacheBuilder[K, V]) = builder.expireAfterWrite(timeout.length, timeout.unit)
   }
 
-  case class OnRemoval[K, V](onRemove: (K, V, RemovalCause) => PrecogUnit) extends CacheOption[K, V] {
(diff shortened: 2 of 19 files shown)


djspiewak:
Actually…  Don't merge this yet.

djspiewak:
Ok, if the build passes here, I think it's read.

alissapajer:
@djspiewak travis has figured out that `???` is ambiguous in that one file. So all the builds have failed. Do you want to just add `import slamdata.Predef{??? => _, _}` in that file? My build grabbing the new `???` from slamdata-predef is still building locally...

alissapajer:
Or alternatively you can just wait for my PR, which should be ready in not too long...

project/Precog.scala
```diff
@@ -6,7 +6,14 @@ import scala._, Predef._
...
 
-  def warningOpts = Seq("-g:vars", "-deprecation", "-unchecked", "-Ywarn-unused", "-Ywarn-unused-import", "-Ywarn-numeric-widen")
+  def warningOpts = Seq(
+    "-g:vars",
+    "-deprecation",
+    "-unchecked",
+    "-Ywarn-unused",
+    "-Ywarn-unused-import",
+    "-Ywarn-numeric-widen",
+    "-Ywarn-value-discard")
```
alissapajer: @djspiewak so, `warningOpts` is used in `withWarnings` but I don't see that used anywhere. Do you know if these warnings are actually being applied?

alissapajer: They are all inside an implicit class so maybe they are being called, but I couldn't find any documentation for this feature...

djspiewak: Oh do we just not use `withWarnings`?  I did notice that `-Ywarn-value-discard` wasn't being applied.

alissapajer: @djspiewak unless sbt is doing some magic that I have yet to uncover, I don't see anything in the block that `withWarnings` is in being used...

alissapajer: As a test I just added `"-Xfatal-warnings` to this block and nothing errored.

djspiewak: Yeah, I did the same thing.  I think it's ok to keep the change anyway (adding `warn-value-discard`), since when we reenable warnings, this will probably be where we start, and it's good to have the flag added (it wasn't there before) at least as documentation.

alissapajer: Yeah, I don't mind this being here for now, at least for this PR it's fine.

alissapajer: @djspiewak ok so here's my plan: In an unrelated PR, I'm going to apply `.withWarnings` to our FBE-related projects, but then comment out the warnings here that I'm not addressing so we don't get spammed in sbt. That way we can actually confirm that we have, for example, no more unused imports.

djspiewak: That's what I'm thinking as well.  We already have a separate issue for reenabling warnings, so it seems reasonable to address it in a separate PR.

Issue #2297, opened 2017-06-08
Add ability to mount more then one spark based connector 


EncodePanda:
Currently there can be mounted only one spark based connector (restricted by `SparkContext` limitations). This can be fixed by creating `SparkContex` only when it is needed (per query)

zeeiyerWork:
Unsure if we ourselves need to fix this. The pressure is off, since we are defaulting to Mimir and not Spark-Local. So moving it to on-hold milestone.

PR #2298, opened 2017-06-08
Enable multiple spark connectors

rabbitonweb:
Previously SparkContext was created per connection (per mount). This
had major drawback as only one spark based connector could be mounted
at given moment.
With this change SparkContext will be created only when its needed (during
query/read) and close immediately when result is returned.
This however has consequences - if two different users try to use
connection in the same time, second one will be blocked with an error.
 
Part  of https://github.com/quasar-analytics/quasar/issues/2297


jdegoes:
This is not a good long-term solution. We can expect many simultaneous queries.

rabbitonweb:
@jdegoes yeah, this is what I wanted to talked to you abgout today. Problems with this approach are in commits comments. 

I will close this PR as I have other approach which I think is lot better.

alissapajer:
@rabbitonweb can this be closed now that #2314 is open?

alissapajer:
@rabbitonweb can we close this issue?

rabbitonweb:
I physically close this issue today in the morning! Wow.. magic... closing it now!

PR #2301, opened 2017-06-08
preparation to support derived MapFuncs

rintcius:
Preparation to support derived `MapFunc`s. Main changes:

* rename `MapFunc(s)` to `MapFunc(s)Core`
* add `MapFuncDerived` - `MapFuncDerived` contains `MapFunc`s which are defined in terms of `MapFuncCore`, so they can be implemented _optionally_ in a backend. 
* add `ExpandMapFunc` - If a backend does not implement a `MapFuncDerived` then `ExpandMapFunc` takes care of expanding it in terms of `MapFuncCore`
* `FuncHandler` is now a typeclass that takes an input functor (and has 3 instances: `MapFuncCore`, `MapFuncDerived` and a `Coproduct` of `FuncHandler`s)

Note: this PR does not integrate `MapFuncDerived`, it's just a preparation at this stage


connector/src/main/scala/quasar/qscript/package.scala
```diff
@@ -58,6 +58,9 @@ import scalaz._, Scalaz._
 //     irrelevant here, and autojoin_d has been replaced with a lower-level join
 //     operation that doesn’t include the cross portion.
 package object qscript {
+
+  type MapFunc[T[_[_]], A] = Coproduct[MapFuncCore[T, ?], MapFuncDerived[T, ?], A]
```
rintcius: I'll try to get rid of this type def

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -19,28 +19,67 @@ package quasar.physical.mongodb.planner
...
+
+  def coproduct[G[_], H[_]]
+      (runG: G ~> OptionFree[F, ?], runH: H ~> OptionFree[F, ?])
+      //(implicit G: FuncHandler[T, G], H: FuncHandler[T, H])
+      : Coproduct[G, H, ?] ~> OptionFree[F, ?] =
+    λ[Coproduct[G, H, ?] ~> OptionFree[F, ?]](_.run.fold(runG(_), runH(_)))
+
+  def run[MF[_]]
+      (implicit MFC: MapFuncCore[T, ?] :<: MF, MFD: MapFuncDerived[T, ?] :<: MF):
+      MapFunc[T, ?] ~> OptionFree[F, ?] =
```
rintcius: Can remove `MapFunc` type def if I change to this type signature:
```
  def run[MF[_]]
     (implicit MFC: MapFuncCore[T, ?] :<: MF, MFD: MapFuncDerived[T, ?] :<: MF):
      Coproduct[MapFuncCore[T, ?], MapFuncDerived[T, ?], ?] ~> OptionFree[F, ?] =
    coproduct(runCore, FuncHandler.handleUnhandled(runDerived, runCore))
```
.. but not sure if this can be further improved (e.g. by specifying return type in terms of `MF` somehow)

sellout: Ah, see, this is a good example. It leads to a definition of FuncHandler as a type class.

So then we have an instance for `MapFuncCore`, an instance for `MapFuncDerived` (which depends on the instance for MapFuncCore), and an instance for `Coproduct[F, G, ?]` (which depends on instances for `F` and `G`).

So, now the question is … when do you make that change 😄

sellout: Also note that you don’t use `MF` (or its `Inject` instances) anywhere. You don’t need them because you’ve defined an explicit `coproduct` method that combines the two cases you need.

sellout: Oh right – and if you do leave it with the explicit Coproduct for now, make sure to add a TODO describing the type class approach.

rintcius: Ok! I'll give that a shot tomorrow

rintcius:
Re. PR now as opposed to later: mainly because I expect the integration part to take some time & I like to get feedback sooner rather than later. Also because this part touches quite a lot of files, so was thinking an earlier merge may avoid having to resolve merge conflicts (and since it's not integrated yet it's fairly safe to merge).

wemrysi:
Ok, are you planning on modifying `FuncHandler` as @sellout mentioned or is this ready? If you're still working on it, change the status tag to "in progress" and then ping me once you've finished.

rintcius:
@wemrysi ah yes, I'm working on that change - changed the status

rintcius:
@wemrysi FuncHandler is now a typeclass - PR ready for re-review

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -213,13 +60,293 @@ object FuncHandler {
...
+            import fp._
+
+            fa.some collect {
+              case D.Abs(a1)       => $abs(a1)
+            }
+          }
+        }
+    }
+
+  implicit def mapFuncDerivedUnhandled[T[_[_]]: CorecursiveT]
```
wemrysi: What is the benefit of having this and `mapFuncDerived` separate? It seems they could be combined into a single instance.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -213,13 +60,293 @@ object FuncHandler {
...
+    }
+
+  implicit def mapFuncCoproduct[F[_], G[_]]
+      (implicit F: FuncHandler[F], G: FuncHandler[G])
+      : FuncHandler[Coproduct[F, G, ?]] =
+    new FuncHandler[Coproduct[F, G, ?]] {
+      def handleOpsCore[EX[_]: Functor](trunc: Free[EX, ?] ~> Free[EX, ?])(implicit e26: ExprOpCoreF :<: EX)
+          : Coproduct[F, G, ?] ~> OptionFree[EX, ?] =
+        λ[Coproduct[F, G, ?] ~> OptionFree[EX, ?]](_.run.fold(
+          F.handleOpsCore(trunc)(implicitly[Functor[EX]], e26)(_),
```
wemrysi: Does switching to `F.handleOpsCore(trunc).apply _` allow you to avoid passing the implicit arguments by hand (and similarly for the other instances where this is done)?

wemrysi:
Once `MapFuncDerived` is actually being used, it may be worth attempting to generalize the optional handling implemented in `FuncHandler` to something that could be used by other connectors as I'm guessing they'll all need a similar construct.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
rintcius: They could be combined, but I separated them to make the similarity between `mapFuncDerived` and `mapFuncCore` clear. Happy to change if there's benefit to combine them.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
rintcius: ah nice trick! - 2 of them can be changed, but I also get 4 errors like this - do you know of a way to fix this without passing the implicits by hand?:
```
[error] /mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala:339: ambiguous implicit values:
[error]  both value vectorInstance in trait VectorInstances of type => scalaz.Traverse[Vector] with scalaz.MonadPlus[Vector] with scalaz.BindRec[Vector] with scalaz.Zip[Vector] with scalaz.Unzip[Vector] with scalaz.IsEmpty[Vector] with scalaz.Align[Vector]
[error]  and method eitherMonad in trait EitherInstances of type [L]=> scalaz.Traverse[[β$0$]scala.util.Either[L,β$0$]] with scalaz.MonadError[[β$1$]scala.util.Either[L,β$1$],L] with scalaz.BindRec[[β$2$]scala.util.Either[L,β$2$]] with scalaz.Cozip[[β$3$]scala.util.Either[L,β$3$]]
[error]  match expected type scalaz.Functor[EX]
[error]           G.handleOps3_2.apply _
[error]             ^
```

rintcius:
@wemrysi indeed, generalizing so that it's usable for other connectors sounds good. May also be a reason for keeping `mapFuncDerived` and `mapFuncDerivedUnhandled` separate for now

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
wemrysi: @rintcius Try applying the type argument: `G.handleOps3_2[EX].apply _`

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
wemrysi: I see what you mean, I think I found it harder to follow since there are two implementations of `FuncHandler[MapFuncDerived[T, ?]]` (though only one available implicitly). However, if you're able to generalize then you should be able to get rid of this one, so I'm fine with keeping it in hopes that it will disappear.

How about adding a `TODO` to this implementation to generalize to connectors other than MongoDB?

rintcius:
I'll try to resolve the conflicts later tonight - @wemrysi are you happy with the PR apart from that?

wemrysi:
@rintcius Yep, thanks for the updates.

PR #2304, opened 2017-06-08
Remove `CommonDependencies`

jedesah:
- Upgrade a few dependencies like `pathy`, `rapture`, `refined`, `parser-combinators`, `aesh`, `h2`, `spark` `scala-xml` and a few others. Most are revision numbers so should be safe (not that tests wouldn't alert us of some breaking change)
- Add notes on attempt to upgrade `doobie`. Conclusion: wait for `0.4.2`


jedesah:
Not the most convenient diff sorry...

project/Dependencies.scala
```diff
@@ -63,20 +74,20 @@ object Dependencies {
...
-    CommonDependencies.doobie.core,
-    CommonDependencies.doobie.hikari,
-    CommonDependencies.doobie.postgres,
-    "org.http4s"           %% "http4s-core" % http4sVersion,
-    CommonDependencies.monocle.`macro`,
-    "com.github.tototoshi" %% "scala-csv"      % "1.3.4",
-    "com.slamdata"         %% "pathy-argonaut" % pathyVersion,
-    CommonDependencies.doobie.specs2                          % Test,
-    CommonDependencies.doobie.h2                              % Test
+    "org.tpolecat"               %% "doobie-core"     % doobieVersion,
```
drostron: The other blocks align the 3rd column.

project/Dependencies.scala
```diff
@@ -96,12 +107,12 @@ object Dependencies {
   }
 
   val postgresql = Seq(
-    CommonDependencies.doobie.core     % "compile, test",
-    CommonDependencies.doobie.postgres % "compile, test"
+    "org.tpolecat" %% "doobie-core"     % doobieVersion % "compile, test",
```
drostron: Same

jedesah:
@drostron Any idea where scala `2.11.11` is coming in from?

jedesah:
(in the first build failure)

drostron:
@jedesah Strange, it even looks like `2.11.8` is being set for that stage.

jedesah:
Reverted the upgrade to Spark as I have a suspicion it might be the cause of the build failure. (I guess I could have tested this locally before pushing...)

drostron:
@jedesah Just a couple alignment nits and then this looks ready.

jedesah:
@drostron done

PR #2307, opened 2017-06-09
Publish mimir and niflheim

drostron:


build.sbt
```diff
@@ -508,6 +511,7 @@ lazy val niflheim = project.setup.noArtifacts
       "org.typelevel"      %% "spire"      % "0.14.1", // TODO use spireVersion from project/Dependencies.scala
       "org.objectweb.howl" %  "howl"       % "1.0.1-1"))
   .settings(headerSettings)
+  .settings(publishSettings)
```
alissapajer: Why do we need to publish niflheim? What is it included in that requires this?

project/Precog.scala
```diff
@@ -65,8 +65,7 @@ object Build {
 
     def setup: Project = (
       serialTests scalacPlugins (kindProjector) also(
-        organization := "quasar",
-        version := "0.1",
```
alissapajer: :+1:

build.sbt
drostron: My assumption was that `mimir` depends on `yggdrasil` and `yggdrasil` depends on `niflheim`. I'm not clear why we don't need to publish `yggdrasil`.

drostron: I mostly just followed what QAdv was telling me after publishing quasar locally.

alissapajer: Yeah, so `mimir` also depends on `blueeyes` and `precog` and those aren't getting published either, so I'm not sure why it required `niflheim` and not the others. Do you think it's good practice to just publish every project? I'm not sure what's standard practice here...

drostron: I think it's ok to only publish what QAdv needs since it's the sole intended consumer for `internal` dependencies. Once we polyrepo and version things independently then publishing all aspects is likely something we'd want to do.

drostron: @alissapajer Turns out this caused `blueeyes` and `precog` to publish in a certain fashion anyway, so we'll want to apply proper settings. I'll submit a PR to do so momentarily.

alissapajer: @drostron sounds good. What about `yggdrasil`, will we need to publish it too?

drostron: @alissapajer Correct, we will — just forgot to mention it.

PR #2311, opened 2017-06-10
Cache MarkLogic RPM in Travis

nicflores:
Keep MarkLogic.rpm cached away in Travis and load it from cache when ML container is built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ branches:
 
 cache:
   directories:
-    - $HOME/.marklogic
+    - $HOME/.cache/quasar/marklogic
     - $HOME/.coursier/cache
     - $HOME/.ivy2/cache
     - $HOME/.sbt
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ services:
 # needed by the  postgreql container
 before_install:
   - sudo /etc/init.d/postgresql stop
-
+ 
 # create all the docker containers we need for IT testing
 install:
   - cd $TRAVIS_BUILD_DIR/docker/scripts
@@ -54,6 +54,7 @@ branches:
 
 cache:
   directories:
+    - $HOME/.marklogic
     - $HOME/.coursier/cache
     - $HOME/.ivy2/cache
     - $HOME/.sbt
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,19 +14,17 @@ before_install:
  
 # create all the docker containers we need for IT testing
 install:
-  - cd $TRAVIS_BUILD_DIR/docker/scripts
-  - ./initContainer quasar_metastore
-  - ./initContainer quasar_$CONNECTOR
-  - docker-compose ps
+  - docker/scripts/initContainer quasar_metastore
+  - docker/scripts/initContainer quasar_$CONNECTOR
+  - docker ps
 
 # run db config scripts and populate the it/testing.conf file
 before_script:
-  - cd $TRAVIS_BUILD_DIR/docker/scripts
-  - ./setupContainers -c quasar_metastore
-  - ./setupContainers -c quasar_$CONNECTOR
-  - ./assembleTestingConf -c quasar_metastore
-  - ./assembleTestingConf -i quasar_$CONNECTOR
-  - cd $TRAVIS_BUILD_DIR/it && cat testing.conf
+  - docker/scripts/setupContainers -c quasar_metastore
+  - docker/scripts/setupContainers -c quasar_$CONNECTOR
+  - docker/scripts/assembleTestingConf -c quasar_metastore
+  - docker/scripts/assembleTestingConf -i quasar_$CONNECTOR
+  - cat it/testing.conf
 
 # export some needed env vars, build, then publish and tag resources
 script:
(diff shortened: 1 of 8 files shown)


docker/Dockerfiles/Marklogic/getMLRpm
```diff
@@ -0,0 +1,63 @@
...
+  curl -o $ML_RPM_PATH -m 2000 "${ML_DEV_URL}$(cat /tmp/url)"
+}
+
+if [[ ${TRAVIS:-} ]]
+then
+  echo "In Travis, checking for cached MarkLogic.rpm..."
+
+  if [[ $CONNECTOR =~ "marklogic" ]]
+  then
+    mkdir -p $HOME/.marklogic
```
wemrysi: You might consider using `$HOME/.cache/quasar` (or similar) as the cache location in all cases which would eliminate the special handling for travis.

docker/Dockerfiles/Marklogic/getMLRpm
```diff
@@ -0,0 +1,63 @@
...
+    ML_DIR="$TRAVIS_BUILD_DIR/docker/Dockerfiles/Marklogic"
+    ML_RPM_PATH="$ML_DIR/MarkLogic.rpm"
+    ML_RPM_SHA1="$ML_DIR/MarkLogic.rpm.sha1"
+
+    echo "48cfbd30d003547feb6b60e120abd4542e8b83f9  $ML_RPM_PATH" > $ML_RPM_SHA1
```
wemrysi: We should move the `SHA` into a variable near the `ML_VERSION` so we can easily update them together.

docker/docker-compose.yml
```diff
@@ -38,14 +38,14 @@ quasar_metastore:
   image: postgres:9.6
   privileged: true
   ports:
-    - "5432:5432"
```
drostron: Unintended whitespace changes?

docker/scripts/setupContainers
```diff
@@ -111,7 +111,22 @@ configure_all_live_containers() {
...
+    fi
+    docker-compose -f $(dirname $0)/../docker-compose.yml up -d $1
+  else
+    if [[ $1 =~ "marklogic" ]]
+    then
+      $(dirname $0)/../Dockerfiles/Marklogic/getMLRpm
+    fi
+    docker-compose -f $(dirname $0)/../docker-compose.yml up -d $1
+  fi
 }
```
nicflores: Looks like I get rid of the travis check in the code above as well as the extra if conditions checks and just have one.

wemrysi:
@nicflores Is this ready for another look?

nicflores:
@wemrysi, yeah it is. Take a look and let me know what you think.

.travis.yml
```diff
@@ -54,6 +52,7 @@ branches:
 
 cache:
   directories:
+    - $HOME/.marklogic
```
wemrysi: I think we should cache the RPM somewhere under `~/.cache` (the default XDG location for cache files). We should probably also name the directory something like `~/.cache/quasar-build` or put it under `~/.cache/quasar` somewhere to make it clear what was actually responsible for creating the file.

docker/Dockerfiles/Marklogic/getMLRpm
```diff
@@ -0,0 +1,42 @@
...
+
+function downloadMLRpm() {
+  curl -X POST --data "email=${ML_DOWNLOAD_EMAIL}&password=${ML_DOWNLOAD_PASSWORD}" -c $ML_DEV_COOKIES $ML_DEV_LOGIN
+  curl -X POST --data "download=$ML_ASSET_PATH" -b $ML_DEV_COOKIES $ML_DEV_GET_TOKEN > /tmp/json
+  cat /tmp/json | cut -f 3 -d: | cut -c2- | rev | cut -c3- | rev > /tmp/url
+  curl -o $ML_RPM_PATH -m 2000 "${ML_DEV_URL}$(cat /tmp/url)"
+  rm $ML_DEV_COOKIES /tmp/url /tmp/json
+}
+
+mkdir -p $HOME/.marklogic
```
wemrysi: Could swap the order of this and the `CACHE_DIR` declaration and then use `$CACHE_DIR` here to avoid repetition.

docker/Dockerfiles/Marklogic/getMLRpm
```diff
@@ -0,0 +1,42 @@
...
+CACHE_DIR="$HOME/.marklogic"
+ML_RPM_PATH="$CACHE_DIR/MarkLogic.rpm"
+ML_RPM_SHA1="$CACHE_DIR/MarkLogic.rpm.sha1"
+
+echo "$ML_SHA  $ML_RPM_PATH" > $ML_RPM_SHA1
+
+if [ -f $CACHE_DIR/MarkLogic.rpm ]
+  then
+    echo "Found MarkLogic.rpm in cache ..."
+    cp $CACHE_DIR/MarkLogic.rpm docker/Dockerfiles/Marklogic/
```
wemrysi: Can factor this out of the block and just have a single branch that downloads the RPM if it doesn't exist.

Also, replace `$CACHE_DIR/MarkLogic.rpm` with `$ML_RPM_PATH` everywhere.

.travis.yml
nicflores: Should it be `$HOME/.cache/quasar/marklogic` or just `$HOME/.cache/quasar/`? I can go either way.

wemrysi: I vote for `$HOME/.cache/quasar/marklogic/`

drostron:
@nicflores Oh, and needs a merge from master.

PR #2314, opened 2017-06-12
Multiple Spark based connectors by sharing SparkContext

rabbitonweb:
This is second attempt to provide https://github.com/quasar-analytics/quasar/issues/2297 functionality.

This time we init `SparkContext` when connector is mounted, however when second connector (based on Spark) wants to be mounted, instead of returning an error we can reuse existing `SparkContext` - as long as mount connects to the same cluster

### DIFF
--- a/interface/src/main/scala/quasar/main/package.scala
+++ b/interface/src/main/scala/quasar/main/package.scala
@@ -31,6 +31,7 @@ import quasar.physical._, couchbase.Couchbase
 
 import scala.util.control.NonFatal
 
+import org.apache.spark._
 import doobie.imports.Transactor
 import eu.timepit.refined.auto._
 import monocle.Lens
@@ -49,6 +50,8 @@ package object main {
   type MainTask[A]       = MainErrT[Task, A]
   val MainTask           = MonadError[EitherT[Task, String, ?], String]
 
+  val refSc: Task[TaskRef[Option[(Int, SparkContext)]]] = TaskRef[Option[(Int, SparkContext)]](none)
+
   /** The physical filesystems currently supported. */
   val physicalFileSystems: FileSystemDef[PhysFsEffM] = IList(
     Couchbase.definition translate injectFT[Task, PhysFsEff],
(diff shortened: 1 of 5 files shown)


rabbitonweb:
After talking to @wemrysi this approach seems problematic. The reason is that this change requires connectors to share some data (here information about`SparkContext` being used). Overall ability to share state between connectors is essential for spark based connectors and beneficial for other connectors (thread pools, jdbc connections etc). However making that change right now (the way we define connectors right now) will block the dynamic loading feature of connectors which seems to have higher priority then multiple spark connectors (at least for now) - @jdegoes please comment if I'm mistaken.

In general the idea was to provide quick and striaght-forward fix for Spark's limitations. After two different attempts there seems non to exists (at least trivial) soolution. I'll push #2297 back to backlog, focus on bringing spark connector to @djspiewak `BackendModule` and when that is done we can revisit this issue. Eventually solution to #2297 might be spawning separate JVMs - but that is a topic for another discussion.

I will keep this PR open for a day for you guys to comment and then eventually most probably will close it.

@zeeiyerWork FYI ^^

jdegoes:
Why not just create a global variable holding the spark context inside an object and use that? Shouldn't interfere with dynamic loading or implementing backend module.

djspiewak:
@jdegoes If I classloader-isolate connectors (which *is* the plan), then no state is shared at all (between connectors, including instances of the same connector) unless I know about it *a priori* and load that state in a parent classloader, making it available to the isolates.  So there is no "global state".  Everything would be isolated.

jdegoes:
Why one classloader per mount? Seems like hugely wasteful wrt RAM, etc.

djspiewak:
@jdegoes The only other option is for connectors to declare when they need isolation.  The purpose behind this is to avoid dependency overlap.  Even if they do this though, Spark will definitely be one of the connectors which has to declare itself an isolate, because of its dependency issues.

And yes, it will be somewhat wasteful in terms of metaspace.  Not as much as you would think though.  This is literally how OSGi works.

rabbitonweb:
closed, not this approach

PR #2317, opened 2017-06-12
Add support for functions and imports in views

jedesah:
- Add a statically checked string interpolator for Sql^2


alissapajer:
ping @sellout and @drostron for review (pinging only since there are no github notifications for reviewers)

djspiewak:
So `sqlE` effectively moves some tests from runtime (of the test suite) to compile time (of the test suite), which I would argue is not actually what we want.  This results in higher fragility, longer compile times, worse error messages (than test failures) and harder debugging (again, relative to debugging test failures).  It would be one thing if we were using `sqlE` in production code paths, but it doesn't seem like we would be, so I don't really see the value prop.

frontend/src/main/scala/quasar/sql/parser.scala
```diff
@@ -35,6 +37,44 @@ final case class ObjectDeref[T[_[_]]](expr: T[Sql])      extends DerefType[T]
...
 final case class DimChange[T[_[_]]](unop: UnaryOperator) extends DerefType[T]
 
+trait StaticInterpolator[A] extends Interpolator {
+  def parse(s: String): String \/ A
+
+  def contextualize(interpolation: StaticInterpolation) = {
+    interpolation.parts.foreach {
+      case lit@Literal(_, string) =>
+        parse(string) match {
+          case -\/(msg) => interpolation.abort(lit, 0, msg)
```
jedesah: Ya, it's not ideal, but it's the API it offers. The fact that it happens at compile time makes it somewhat better as any unexpected exception or what not would result in a compile error

frontend/src/main/scala/quasar/sql/parser.scala
```diff
@@ -35,6 +37,44 @@ final case class ObjectDeref[T[_[_]]](expr: T[Sql])      extends DerefType[T]
...
+          case _        => ()
+        }
+
+      case hole@Hole(_, _) =>
+        interpolation.abort(hole, "substitutions are not supported")
+    }
+
+    Nil
```
jedesah: We gain the value of not having to write a macro in order to get compile-time checked sql^2 queries. And we can add support for substitutions later, I just didn't want to put in the effort for that right now since there are only a few places in the code base where it would be useful

djspiewak: > get compile-time checked sql^2 queries

alissapajer:
@jedesah thanks for the explanation, I think you have me convinced. So, in the particular cases where we use `sqlE` we already expect to have valid sql^2 because we are testing a compiler phase after which we have created the sql AST, and so your change allows us to verify that validity at runtime. @djspiewak the value prop is that in these particular tests we intend to be test only valid sql^2.

And in the cases when it's important for us to test what happens when we write invalid sql^2, this change doesn't prevent us from still writing those tests.

jedesah:
Happy to explain my train of thought and have you poke holes when there are some! 😄

frontend/src/main/scala/quasar/sql/package.scala
```diff
@@ -60,6 +61,11 @@ package object sql {
   // TODO: Get rid of this one once we’ve parameterized everything on `T`.
   val fixParser = parser[Fix]
 
+  implicit class SqlStringContext(sc: StringContext) {
+    val sqlE = Prefix(SqlExprInterpolator, sc)
+    val sqlB = Prefix(SqlBlobInterpolator, sc)
```
alissapajer: @jedesah one more thought - do we have plans to use these outside of test code? If not, maybe they would be better in `test` directory rather than `main` directory?

frontend/src/main/scala/quasar/sql/parser.scala
```diff
@@ -35,6 +37,44 @@ final case class ObjectDeref[T[_[_]]](expr: T[Sql])      extends DerefType[T]
...
+  }
+
+  // A String that would not parse should not have made it past compile time
+  def evaluate(interpolation: RuntimeInterpolation): A =
+    parse(interpolation.parts.mkString).valueOr(errMsg =>
+      scala.sys.error(s"Something terribly wrong happened. Failed to parse something at runtime that we checked at compile time. Reason: $errMsg. This is either because the `parse` function of this `StaticInterpolator` is not pure or could (less likely) be a bug with `StaticInterpolator`"))
+}
+
+object SqlExprInterpolator extends StaticInterpolator[Fix[Sql]] {
```
alissapajer: And actually on that note, we'd like to avoid hard-coding `Fix` into main code here. Could we move all this to the `test` directory then too? Or else if we keep it here use `T[_[_]]` instead?

frontend/src/main/scala/quasar/sql/package.scala
jedesah: I think it would be nice to offer this as an API to the people who might use us as a library one day, speaking of which I think the SQL AST and parser would be a good candidate to create an external library, not sure it's mature enough yet though. Anyway, that's not right now so I don't mind moving to tests despite the fact that it's certainly not test specifc

frontend/src/main/scala/quasar/sql/parser.scala
jedesah: Unfortunately, I don't think it's possible to parameterize string interpolators, I wish it was then we wouldn't have to choose a particular fix point type

frontend/src/main/scala/quasar/sql/package.scala
alissapajer: Yeah, I agree it's not necessarily test specific, but I think for now since the only use cases are in tests, then it would make it more clear if it lives in the `test` directory. You could add a comment saying that when this code is more mature, we'd like to expose this outside of tests.

jedesah: Sounds good. Will do

frontend/src/main/scala/quasar/sql/parser.scala
alissapajer: Yeah, moving this to `test` will handle the problem of `Fix` in `main` code for now. If we expose this down the road we might just need to hard-code various fixed points.

jedesah: Nevermind what I said, I think it might be possible be make it generic, I'll explore this further after lunch, but I'll move to test in any event

alissapajer: Cool, sounds good. It would be neat to have it generic, but for the tests `Fix` is fine.

jedesah:
@alissapajer See latest commit :disappointed:

alissapajer:
LGTM, let's keep an eye on those `contextual` issues to see if we'll be able to move the `Fix` stuff over to the test directory. 

Then once this lands I'll look over #2178 again!

Issue #2320, opened 2017-06-13
Rework VFS actor flow in terms of fs2.Stream


djspiewak:
Similar to #2165, but just for the VFS stuff.

PR #2322, opened 2017-06-13
Include integration tests for mimir.

### LINKED ISSUE #2254: Enable first integration test for mimir
This involves hooking mimir into the integration test harness and adding a travis job for mimir.

alissapajer:
Closes #2254 

Includes mimir in the travis build matrix. Enables an integration test that represents a constant value (query never makes it to qscript). The enabling of queries that require loaded data is pending #2320.

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/Mimir.scala
+++ b/mimir/src/main/scala/quasar/mimir/Mimir.scala
@@ -37,7 +37,7 @@ import quasar.yggdrasil.PathMetadata
 import quasar.yggdrasil.vfs.ResourceError
 import quasar.yggdrasil.bytecode.JType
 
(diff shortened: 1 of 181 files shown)


mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -479,12 +484,17 @@ object Mimir extends BackendModule with Logging {
       } yield ()
 
       t.liftB
+      //().point[Backend]
```
djspiewak: Dead code?

alissapajer: I was just keeping that there because the first thing I have to always do when debugging any fs thing is not delete the data when the tests complete.

PR #2324, opened 2017-06-13
Increase timeout for scaladoc gen

drostron:


drostron:
Perhaps a 10 min bump will be sufficient to avoid the recent timeouts we've been seeing.

PR #2330, opened 2017-06-14
Remove `PackageAliases` from FBE

### LINKED ISSUE #2163: Remove PackageAliases
There's no clear distinction about what is and isn't in here, and there are a large number of things that don't work well under this approach.  For example, any Java type on which we need to call static members.

alissapajer:
Closes #2163


alissapajer:
Looks like travis found some unused imports in `core` which `PackageAliases` was somehow masking previously. I'll clean those up.

alissapajer:
@djspiewak ready for merge.

Issue #2333, opened 2017-06-15
CB: Create docTypeKey index when absent


drostron:
In response to #2326

Issue #2332, opened 2017-06-15
CB: Utilize a view for speedier list contents


drostron:
In response to #2326

PR #2335, opened 2017-06-15
View based list contents and docTypeKey index creation when absent

drostron:
Closes: #2332
Closes: #2333


couchbase/src/main/scala/quasar/physical/couchbase/common.scala
```diff
@@ -71,10 +107,17 @@ object common {
   def docTypeValuesFromPrefix(
     ctx: ClientContext,
     prefix: String
-  ): Task[FileSystemError \/ List[DocTypeValue]] = {
-    val qStr = s"""SELECT distinct type FROM `${ctx.bucket.name}`
-                   WHERE `${ctx.docTypeKey.v}` LIKE "${prefix}%""""
-    query(ctx.bucket, qStr).map(_.map(_.toList.map(r => DocTypeValue(r.value.getString(ctx.docTypeKey.v)))))
```
nicflores: Is this the only query we can replace with a view? Just wondering if we can do the same for `deleteHavingPrefix` and in `existsWithPrefix`?

drostron: I don't believe `deleteHavingPrefix` is possible with a view, but updating `existsWithPrefix` is a good candidate — will update accordingly.

drostron:
@nicflores Ready for another look

PR #2343, opened 2017-06-16
Change ignoreFieldOrder handling.

sellout:
We now treat `"ignoreFieldHandling"` the similarly to `"skip"` or `"pending"` in regression tests.

### DIFF
--- a/it/src/main/resources/tests/filterOnFieldComparison.test
+++ b/it/src/main/resources/tests/filterOnFieldComparison.test
@@ -8,7 +8,7 @@
     "data": "zips.data",
     "query": "select * from zips where pop < loc[1] order by `_id` limit 10",
     "predicate": "equalsExactly",
-    "ignoreFieldOrder": ["marklogic_json"],
+    "ignoreFieldOrder": ["*"],
     "expected": [
         { "_id": "01338", "city": "BUCKLAND",        "loc": [-72.764124, 42.615174], "pop": 16, "state": "MA" },
         { "_id": "02163", "city": "CAMBRIDGE",       "loc": [-71.141879, 42.364005], "pop": 0,  "state": "MA" },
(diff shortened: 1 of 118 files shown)


it/src/main/resources/tests/filterOnFieldComparison.test
```diff
@@ -2,13 +2,12 @@
...
     "backends": {
         "mimir": "skip",
-        "mongodb_read_only": "pending",
         "postgresql":        "pending"
     },
     "data": "zips.data",
     "query": "select * from zips where pop < loc[1] order by `_id` limit 10",
     "predicate": "equalsExactly",
-    "ignoreFieldOrder": ["marklogic_json"],
+    "ignoreFieldOrder": true,
```
alissapajer: @sellout Does `true` ignore the field order for all the backends? Why for some of these tests did you do it this way, rather than have `"marklogic_json": "ignoreFieldOrder"` above?

it/src/main/resources/tests/filterSortLimit.test
```diff
@@ -12,7 +12,7 @@
 
     "predicate": "equalsExactly",
     "ignoredFields": ["_id"],
-    "ignoreFieldOrder": ["marklogic_json"],
+    "ignoreFieldOrder": true,
```
alissapajer: This one too with the `true`.

it/src/main/resources/tests/filterOnFieldComparison.test
sellout: Yes – because it uses `select *` we make no guarantee as to the field order. Where `"marklogic_json": "ignoreFieldOrder"` is meant to indicate that the Marklogic connector has a deficiency.

Ideally, we’ll eventually pull this information from the query itself (as with `"data"` and `"equals…"` vs `"contains…"` predicates), but even then we may want to also be explicit about it here, so that some regression in QScript doesn’t make us suddenly start ignoring field order, but rather we get an error saying that this setting doesn’t match the query analysis).

alissapajer: Got it, makes sense. So some of the backends are preserving field order but their results would still be correct if they didn't, because the query itself doesn't guarantee that they do.

PR #2352, opened 2017-06-19
Separate result ordering from the predicate.

sellout:


sellout:
I think this is the last of these for now 😄

it/src/main/resources/tests/tripleFlatten.test
```diff
@@ -11,6 +11,6 @@
     "data": "nested.data",
     "query": "select topObj{*}{*}{*} from nested",
     "NB": "containsExactly has a bug that gets confused on duplicate results (#732).",
-    "predicate": "equalsExactly",
+    "predicate": "exactly",
```
drostron: ignoreResultOrdering?

it/src/main/resources/tests/tripleFlatten.test
```diff
@@ -11,6 +11,6 @@
     "data": "nested.data",
     "query": "select topObj{*}{*}{*} from nested",
     "NB": "containsExactly has a bug that gets confused on duplicate results (#732).",
```
drostron: Stale NB?

it/src/main/resources/tests/var-relation.test
```diff
@@ -14,7 +14,7 @@
 
     "query": "select count(*) as cnt from :table where pop < 1000",
 
-    "predicate": "equalsExactly",
+    "predicate": "exactly",
```
drostron: ignoreResultOrdering?

it/src/test/scala/quasar/regression/FieldOrder.scala
```diff
@@ -16,8 +16,8 @@
```
drostron: Stale filename

it/src/test/scala/quasar/regression/Predicate.scala
```diff
@@ -62,105 +63,126 @@ object Predicate {
...
+      case OrderPreserved =>
+        // FIXME: This case is the same as `Exactly`, but shouldn’t be.
+        val actual   = actual0.map(Some(_))
+        val expected = Process.emitAll(expected0).map(Some(_))
+
+        (actual tee expected)(tee.zipAll(None, None))
+          .flatMap {
+            case (a, e) if jsonMatches(a, e) =>
+              Process.halt
+            case (a, e) if (a == e && fieldOrder === OrderIgnored) =>
```
drostron: Oh, forgot to mention — preexisting, but looks like most of these `==` can be `===`.

it/src/main/resources/tests/different-flattens.test
```diff
@@ -17,7 +17,7 @@
                 comments[*].text       LIKE \"%Dr%\" OR
                 comments[*].replyTo[*] LIKE \"%Dr%\" OR
                 comments[*].`time`     LIKE \"%Dr%\")",
-    "predicate": "equalsExactly",
+    "predicate": "exactly",
```
sellout: Yeah, good point. I guess it works regardless since all the results are identical, but there’s no ordering in the query.

sellout:
@drostron Updated.

PR #2353, opened 2017-06-19
Check for unnested duplicate element in qscript concat.

### LINKED ISSUE #2331: Enable `unifyFlattenedFieldsWithUnflattened` for at least one qscript-based backend
This integration test query
```
select `_id` as zip, loc[*] from zips order by loc[*]
```
along with the very similar `unifyFlattenedFieldsWithMultipleUses`
```
select `_id` as zip, loc[*] from zips where loc[*] > 68 order by loc[*]
```
is creating qscript that includes a theta join. The qscript should not contain a theta join. To note, the query
```
explain select loc[*] from zips order by loc[*] 
```
does not contain a join in the qscript, so the additional field selection is causing source merging to fail. Also to note, the qscript for the query
```
explain select `_id` as zip, loc[*] from zips
```
does not contain a join but contains many unused array elements, so pruning those could help with the source merge mentioned above.

### LINKED ISSUE #2336: Remove duplicates in qscript array concat
In qscript source merging (among other places) we often call `concat` on two `FreeMap`s. This concatenates the two sides into an array, providing access to the left and right sides. 

Now we often call `concat` where one side is an array and the other side is an element of that array. For example the left side is `[A, B]` and the right side is `B`. Currently we naively call `concat` on these two, creating `[[A, B], B]`.

First off, we should not have duplicate structures like this in qscript. But furthermore, this actually prohibits further source merging as we now have two distinct ways to access `B`. Namely, down the road we fail to identify that `[[A, B], B][0][1] == [[A, B], B][1]`.

The solution is to detect this subset case and return just `[A, B]` where left side access would be the entire array and right side access would be through index 1. We then replace `concat` with this improved version.

This should remove the theta join in `select state, loc[*] from zips order by loc[*]` (#2331).

There are more complicated cases we could detect, for example `[A, B, C, D]` and `[B, C, E]` or even `[A, [B, [C, D], E], F]` and `[G, D, F]`. But at this stage it's not clear that this will actually be necessary. So this issue is just to address the simple case where one side is an element of the other.

alissapajer:
Closes #2331 
Closes #2336

### DIFF
--- a/it/src/main/resources/tests/flattenGroupedObject.test
+++ b/it/src/main/resources/tests/flattenGroupedObject.test
@@ -3,8 +3,7 @@
     "backends": {
         "mimir": "skip",
         "couchbase":         "skip",
-        "marklogic_json":    "skip",
-        "marklogic_xml":     "skip",
+        "marklogic_json":    "pending",
         "mongodb_q_3_2":     "pending",
         "postgresql":        "pending",
         "spark_local": "pending",
--- a/connector/src/main/scala/quasar/qscript/Transform.scala
+++ b/connector/src/main/scala/quasar/qscript/Transform.scala
@@ -94,14 +94,14 @@ class Transform
           val (combine, lacc, racc) = concat(lval, rval)
           (combine, lann.provenance, rann.provenance, lacc, racc)
         case (None, Some((rProvs, rBuck))) =>
-          val (combine, bacc, lacc, racc) = concat3(rBuck, lval, rval)
+          val (combine, bacc, lacc, racc) = naiveConcat3(rBuck, lval, rval)
           (combine, lann.provenance, prov.rebase(bacc, rProvs), lacc, racc)
         case (Some((lProvs, lBuck)), None) =>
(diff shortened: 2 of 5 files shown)


alissapajer:
Awesome! Build failed because an integration test is fixed now in spark! That is the best kind of build failure. @sellout thanks for getting the pending tests re-enabled on travis!

sellout:
Hah, right? It’s the kind of thing no one’s going to get upset about.

it/src/main/resources/tests/flattenGroupedObject.test
```diff
@@ -3,8 +3,7 @@
     "backends": {
         "mimir": "skip",
         "couchbase":         "skip",
-        "marklogic_json":    "skip",
-        "marklogic_xml":     "skip",
+        "marklogic_json":    "pending",
```
alissapajer: `marklogic_json` actually throws an exception out of ML, see #2359.

alissapajer:
@sellout build is green!

Issue #2359, opened 2017-06-20
`flattenGroupedObject` throws type exception in `marklogic_json`


alissapajer:
As of #2353 this query passes in `marklogic_xml` but throws the following error in `marklogic_json`.

```
[error] ! flatten a grouped object with filter [./flattenGroupedObject.test] (5 seconds, 417 ms)                                                                                                                       [164/17907]
[error]  quasar.physical.marklogic.xcc.DefaultImpl$XccXQueryException: XDMP-ARGTYPE: (err:XPTY0004) fn:matches(object-node{"name": text{"Greg Pfeil"}, "email": text{"greg@technomadic.org"...}, "^http.*$", "m") -- arg1 is not o
f type xs:string? on line 111
[error]  ---BEGIN QUERY---
[error]  xquery version "1.0-ml";
[error]  
[error]  declare namespace ejson = "http://quasar-analytics.org/ejson";
[error]  declare namespace qscript = "http://quasar-analytics.org/qscript";
[error]  
[error]  declare function ejson:json-deserialize($serialized as item()?) as node()? {
[error]    if (fn:empty($serialized)) then () else xdmp:to-json($serialized)/child::node()
[error]  };
[error]  
[error]  declare function ejson:json-encode($item as item()?) as item()* {
[error]    typeswitch($item)
[error]    case text() return $item
[error]    case boolean-node() return $item
[error]    case number-node() return $item
[error]    case array-node() return $item
[error]    case object-node() return $item
[error]    default return let $t := ejson:type-of($item) return if ($t eq "string") then $item else if ($t eq "boolean") then boolean-node{$item} else ejson:json-encode-type($item)
[error]  };
[error]  
[error]  declare function ejson:json-encode-type($item as item()?) as object-node() {
[error]    ejson:json-deserialize(let $t := ejson:type-of($item) return if (fn:empty($t)) then map:entry("_ejson.type", "na") else if ($t eq "null") then map:entry("_ejson.type", "null") else map:new((map:entry("_ejson.type", 
$t), map:entry("_ejson.value", $item))))
[error]  };
[error]  
[error]  declare function ejson:json-is-array($item as item()?) as xs:boolean {
[error]    typeswitch($item)
[error]    case array-node() return fn:true()
[error]    default return fn:false()
[error]  };
[error]  
[error]  declare function ejson:json-node-type($node as node()) as xs:string? {
[error]    typeswitch($node)
[error]    case text() return "string"
[error]    case null-node() return "null"
[...]

PR #2362, opened 2017-06-20
Enable different-flattens for marklogic_xml.

alissapajer:


it/src/main/resources/tests/different-flattens.test
```diff
@@ -2,9 +2,8 @@
     "name": "merge differently-nested flattens",
     "backends": {
         "mimir": "skip",
-        "couchbase":      "skip",
-        "marklogic_json": "skip",
-        "marklogic_xml":  "skip",
+        "couchbase": "pending",
+        "marklogic_json": "pending",
```
alissapajer: This is failing in `marklogic_json` with a similar error as in #2359 
```
quasar.physical.marklogic.xcc.DefaultImpl$XccXQueryException: XDMP-ARGTYPE: (err:XPTY0004) fn:matches(object-node{"_ejson.type": text{"integer"}, "_ejson.value": number-node{8.92...}, "^.*Dr.*$", "m") -- arg1 is not of type xs:string? on line 93
```
Though the `spark` integration test is also passing for this query, so there is a higher chance that ML has an issue. Although there is always the possibility that it's also passing `spark` by coincidence, and qscript still has a problem.

alissapajer:
@sellout build is green!

PR #2382, opened 2017-06-22
Account for null-node{} comparison oddness

### LINKED ISSUE #2119: Support `TypeOf` function in MarkLogic

### LINKED ISSUE #2369: Support null comparison/filters in marklogic_json
Split from #1948. Integration tests that include checks for nulls pass with `marklogic_xml` but fail with `marklogic_json`. Failing tests include:

- null/notNullFilter.test
- null/nullComparison.test
- null/nullFilter.test
- null/nullTestExprs.test

jsantos17:
Fixes #2369 
Fixes #2119

Problem lies in how comparison works with `null-node{}`. 

| Expression | xdmp:mapping | Result |
-------------|----------------|--------
|`(null-node{} eq 1, null-node{} ne 1)`|false|empty sequence|
|`(null-node{} = 1, null-node{} != 1)`|false|(false,false)|
|`(null-node{} eq 1, null-node{} ne 1)`|true|(false,false)|
|`(null-node{} = 1, null-node{} != 1)`|true|(false,false)|

The most fun is reflexivity:

| Expression | xdmp:mapping | Result |
-------------|----------------|--------
|`null-node{} = null-node{} `| true | false |
|`null-node{} eq null-node{} `| false | false |
|`null-node{} != null-node{}` | false |false |
|`null-node{} ne null-node{} `| true | false |

I'm not really sure if this is the correct implementation, but the workaround in `cast-if-node` brings some sanity as:

```xquery
let $n1 := ejson:cast-if-node(null-node{})
let $n2 := ejson:cast-if-node(null-node{})

return ($n1 = $n2, $n1 != $n2, $n1 = $n1)
```

yields `(true, false, true)` which seems better to me. It fixes the null tests too. Since now `cast-if-node` evaluates `null-node{}` to a string, `xdmp:mapping` shouldn't bring any surprises. But of course, that makes nulls stringly typed.

@wemrysi any suggestion on where is more appropriate to implement a workaround the comparison issues of `null-node{}`?


marklogic/src/main/scala/quasar/physical/marklogic/qscript/JsonStructuralPlanner.scala
```diff
@@ -214,6 +214,7 @@ private[qscript] final class JsonStructuralPlanner[F[_]: Monad: PrologW: QNameGe
         $("node") as ST("node()?")
       ).as(ST("item()?")) { (node: XQuery) =>
         typeswitch(node)(
+                      ST("null-node()") return_ "null".xs,
```
jsantos17: @wemrysi This is the workaround I applied for the problem with comparing nulls. Not sure if this is the correct place for the workaround. Can you recommend somewhere else? I also remember you mentioned that we won't depend on casting as much once we have an XQuery AST so we might want do defer this until then.

jsantos17:
The fun never stops:

```xquery
declare option xdmp:mapping "false";

let $n1 := null-node{}
let $n2 := null-node{}

return ($n1 = $n2, $n1 != $n2)
```

marklogic/src/main/scala/quasar/physical/marklogic/qscript/JsonStructuralPlanner.scala
wemrysi: Ideally, I _think_ we'll  want `... where foo is null` to compile to `$src[./null-node("foo")]` and `select x = null from ...` to be something like `ejson:type-of($src/node("x")) eq "null"`. Or even just use the `ejson:type-of` version in both contexts to start.

We could likely achieve the latter with a `FreeMap` transformation that rewrites `Eq(Null, x)` to `Eq("null", TypeOf(x))` and implementing `TypeOf` for MarkLogic (it is mostly implemented, modulo string names).

sellout: Just for a bit of clarification (since a lot happens between SQL and connectors), `foo is null` currently compiles to the same thing as `foo = Null`, which is `Eq(foo, Null)` and is only true when `foo` is `Null`. However, the SQL `foo is null` should actually compile to something like `Or(IfUndefined(foo, true), Eq(foo, Null))`.

So, when you say `foo is null`, you currently have to write that in SQL more like `foo ?? true or foo = null`.

sellout: As an aside … this might be another opportunity for an optional MapFunc? For connectors that have SQL-like null semantics, we could rewrite `Or(IfUndefined(foo, true), Eq(foo, Null))` to `IsNull(foo)`, and `Eq(foo, Null)`to `And(IfUndefined(foo, false), IsNull(foo))`.

wemrysi: @sellout Thanks for the clarification, I was probably muddying the waters as we're mostly concerned about the `Eq(foo, Null)`, however it began in SQL^2. In this case a transformation that converted that to `Eq(TypeOf(foo), "null")` would probably suffice, not sure how useful that is for other connectors.

jsantos17: @wemrysi can we rewrite the QScript from `Eq(foo, Null)` to `Eq(TypeOf(foo), "null")` inside the Marklogic connector?

wemrysi: Yeah, in `MarkLogic#plan` we can apply any transformations to the QScript that make sense for the connector.

jsantos17: @wemrysi Ok, I prefer that approach. I'd also like to get my hands dirty on QScript. What do you think? Or shall we merge as is?

wemrysi: @jsantos17 I'd say do the QScript approach as it seems less likely to require us to change it in the future (unless structural equality becomes a thing in MarkLogic).

jsantos17: Cool. I'll get started with that.

jsantos17:
@wemrysi this is ready for another look 😀

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
```diff
@@ -134,6 +134,37 @@ package object qscript {
   ): F[XQuery] =
     fqs.cataM(interpretM(κ(src.point[F]), QTP.plan))
 
+  def rewriteNullCheck[T[_[_]]: BirecursiveT, U, E](
+    implicit UR: Recursive.Aux[U, CoEnv[E, MapFuncCore[T, ?], ?]],
+             UC: Corecursive.Aux[U, CoEnv[E, MapFuncCore[T, ?], ?]]
+  ): CoEnv[E, MapFuncCore[T, ?], U] => CoEnv[E, MapFuncCore[T, ?], U] = {
+
+    import quasar.qscript.MapFuncsCore.{Eq, Neq, TypeOf, Constant}
```
wemrysi: These should be moved/merged with the rest of the imports at the top of the file.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
```diff
@@ -134,6 +134,37 @@ package object qscript {
...
+
+    object NullLit {
+      def unapply[T[_[_]]: RecursiveT, A](mfc: CoEnv[E, MapFuncCore[T, ?], A]): Boolean =
+        mfc.run.exists[MapFuncCore[T, A]] {
+          case Constant(ej) => EJson.isNull(ej)
+          case _            => false
+        }
+    }
+
+    def stringLit(str: String): Constant[T, U] =
```
wemrysi: Maybe just fold this into `nullString`, not sure if there is any advantage in keeping it a `def` since we only use it once.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/package.scala
```diff
@@ -134,6 +134,37 @@ package object qscript {
...
+      Constant[T, U](EJson.fromCommon(Str[T[EJson]](str)))
+
+    val nullString: U = UC.embed(CoEnv(stringLit("null").right))
+
+    fa => CoEnv(fa.run.map {
```
wemrysi: You could use `totally`, defined in `contrib.matryoshka`, to eliminate the `other` case.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/StructuralPlanner.scala
```diff
@@ -203,6 +206,23 @@ trait StructuralPlanner[F[_], FMT] { self =>
         })
     })
 
+  // ejson:type-of-q($item as item()*) as xs:string?
+  def typeOfQFn(implicit F0: Bind[F], F1: PrologW[F]): F[FunctionDecl1] =
```
wemrysi: I think this should be defined in `marklogic.qscript.lib` and just depend on an instance of `StructuralPlanner`, see `marklogic.qscript.lib.{concat, meta}` for an example.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/StructuralPlanner.scala
```diff
@@ -203,6 +206,23 @@ trait StructuralPlanner[F[_], FMT] { self =>
...
+    ejs.declare[F]("type-of-q") flatMap (_(
+      $("item") as ST.Top
+    ).as(ST("xs:string?")) { item: XQuery =>
+      val (lookup, mapped) = ($("lookup"), $("mapped"))
+      typeOf(item) map ( tpe =>
+        let_(
+          lookup := map.new_(mkSeq_(
```
wemrysi: I think nested `if` statements, though they're a bit ugly, would be better here both to avoid constructing the `map:map` every time as well as to avoid the double handling of '"na"`.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/StructuralPlanner.scala
jsantos17: Sounds good. I'll move it.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/StructuralPlanner.scala
jsantos17: Yeah. I was wondering what's the best way to do this. I wish typeswitch would support matching on strings. Apparently it does not.

```xquery
typeswitch(text {"aa"})
  case text("aa") return 1
  case text("bb") return 2
  default return 3
```

yields 3.

wemrysi: Aye, it isn't general pattern matching, it only allows you to match on `SequenceTypes` which are straightforward type tests for atomic values and slightly more powerful tests for node types, i.e. `element()` or `element("foo")` etc.

jsantos17: Oh well. Nested `if` it is! 😞

jsantos17:
@wemrysi implemented your requested changes and the build is green 😃

PR #2389, opened 2017-06-23
Don’t create an autojoin when one side is a constant.

sellout:
This avoids indirect access to constant values, which is something we can hopefully promise to connectors – that constants never have to be dereferenced.

The Mongo tests that are fixed by this is somewhat accidental. First, we can’t do `ProjectField` or `MakeMap` in Mongo expressions if the name of the field isn’t a constant string, which means we fall back to JS. And the old code _should_ work in JS, but isn’t currently supported (because it involves reaching outside our pure subset of JS, since you can only add dynamic field names with mutation 😭).

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Transform.scala
+++ b/connector/src/main/scala/quasar/qscript/Transform.scala
@@ -86,9 +86,11 @@ class Transform
 
     (left.value, right.value) match {
       case (Embed(QC(Unreferenced())), r) =>
-        AutoJoinResult(AutoJoinBase(r, rann.provenance), lann.values, rann.values)
+        val buckets = prov.joinProvenances(List(provenance.Nada()), rann.provenance)
+        AutoJoinResult(AutoJoinBase(r, buckets), lann.values, rann.values)
       case (l, Embed(QC(Unreferenced()))) =>
-        AutoJoinResult(AutoJoinBase(l, lann.provenance), lann.values, rann.values)
+        val buckets = prov.joinProvenances(lann.provenance, List(provenance.Nada()))
+        AutoJoinResult(AutoJoinBase(l, buckets), lann.values, rann.values)
       case (l, r) =>
         val SrcMerge(src, lBranch, rBranch) = merge.mergeT(l, r)
 
--- a/connector/src/main/scala/quasar/qscript/Transform.scala
+++ b/connector/src/main/scala/quasar/qscript/Transform.scala
@@ -83,35 +83,43 @@ class Transform
     val rann = right.ann
     val lval: JoinFunc = lann.values.as[JoinSide](LeftSide)
     val rval: JoinFunc = rann.values.as[JoinSide](RightSide)
-    val SrcMerge(src, lBranch, rBranch) = merge.mergeT(left.value, right.value)
 
-    val lprovs = prov.genBuckets(lann.provenance) ∘ (_ ∘ (_.as[JoinSide](LeftSide)))
-    val rprovs = prov.genBuckets(rann.provenance) ∘ (_ ∘ (_.as[JoinSide](RightSide)))
+    (left.value, right.value) match {
+      case (Embed(QC(Unreferenced())), r) =>
+        AutoJoinResult(AutoJoinBase(r, rann.provenance), lann.values, rann.values)
+      case (l, Embed(QC(Unreferenced()))) =>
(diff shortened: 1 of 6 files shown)


alissapajer:
@sellout it looks like `/demo/nestedSelectWithReduce.test` is failing in spark. And `/flattenGroupedObject.test` is failing in marklogic_xml. The couchbase build failed to load data (not a real test failure) so I restarted it.

alissapajer:
@sellout ok so `flattenGroupedObject.test` is actually passing on master by coincidence, and I'm in the middle of working on fixing that right now (see #2359). So you can pending that test in this PR.

alissapajer:
And spark is the only backend that has `demo/nestedSelectWithReduce.test` enabled, so I'm not sure immediately if this PR is introducing a real bug in qscript or not.

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -83,35 +83,43 @@ class Transform
...
-        rewrite.unifySimpleBranches[F, T[F]](src, lBranch, rBranch, combine)(rebaseT).getOrElse(
-          TJ.inj(ThetaJoin(src, lBranch, rBranch, prov.genComparisons(newLprov, newRprov), JoinType.Inner, combine))).embed,
-        prov.joinProvenances(newLprov, newRprov)),
-      lacc,
-      racc)
+    (left.value, right.value) match {
+      case (Embed(QC(Unreferenced())), r) =>
+        AutoJoinResult(AutoJoinBase(r, rann.provenance), lann.values, rann.values)
+      case (l, Embed(QC(Unreferenced()))) =>
+        AutoJoinResult(AutoJoinBase(l, lann.provenance), lann.values, rann.values)
```
alissapajer: The diff here isn't the nicest because of indentation. Is it correct that the only change you made was to add these two cases with `Unreferenced()` here and everything below this stayed the same?

sellout: Yep – remember you can add `?w=1` to the URL to avoid the whitespace changes. That makes it a lot clearer.

alissapajer: Oh yeah thanks!

alissapajer:
Though with that said about `demo/nestedSelectWithReduce.test`, if this PR were introducing a qscript bug, I don't think it's these changes that would be the bug, rather these changes would have unmasked some other bug.

sellout:
Yeah, I agree with that.

On the new Mongo branch, this uncovers a “bug”, in that this allows (more) ThetaJoin conditions to end up like `Guard(???, ???, Eq(…), ?)`, so the Theta => Equi conversion fails, and it end up generating slow cross-joins with post filtering. So it needs a change (which I’ve implemented) that pulls `Cond` and `Guard` that only reference one side of the join into pre-filters on the branch, re-exposing the `Eq` for successful Theta => Equi conversion.

So, _maybe_ the same thing is happening here? (Although I _think_ the bit that bubbles Guards to the top of an expression is only on the new Mongo branch.)

alissapajer:
Yeah, it seems reasonable that this change is preventing some other qscript transformation from being hit, and the backend was previously relying on that transformation. I'm going to take a look at the qscript for this integration test. I was already planning on working on other qscript issues off of this branch anyways :)

alissapajer:
So I've simplified the demo query and determined that `select count(*), state from zips` returns correct results against spark on master but incorrect results (empty set) against this branch. I am however not at all convinced that the qscript on master is actually correct. 

Since this is such a basic query, and this does represent a user-facing regression, I think it makes sense to keep this PR open until we resolve this issue. So I'm going to work on getting that query (and the demo integration test) working again on this branch.

alissapajer:
I am also surprised that only this one integration test failed. So I'll be adding at least that simplified query to our test suite as well.

alissapajer:
Oh interesting! So `flattenGroupedObject` also contains a similar pattern with `count(*)`, in that case `select distinct commit{*}, count(*) from ...`. So it looks like both these failing tests might be able to be revived!

alissapajer:
@sellout this change https://github.com/sellout/quasar/pull/12 enables both the failing tests locally. QScript is expecting us to preserve the provenance from the constant values.

alissapajer:
And the build is green!

PR #2391, opened 2017-06-23
Support computing SSTs for a query

### LINKED ISSUE #2344: Support SSTs on Processes
Currently the features are mutually exclusive. If we use processes, we can't use SSTs, and if we use SSTs we can't use processes.

wemrysi:
Supports generating SSTs from a query as opposed to a filesystem path. This is a breaking change as it replaces the previous behavior in the HTTP api and at the repl. The previous behavior can be easily recovered by asking for the schema of `select * from /the/path`.

Fixes #2344.


frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlanR.scala
```diff
@@ -41,6 +41,7 @@ final case class NamedConstraint[T]
 final case class ConstrainedPlan[T]
   (inferred: Type, constraints: List[NamedConstraint[T]], plan: T)
 
+// TODO: Move constraints to methods and/or pull the constructors into own class.
```
wemrysi: Not sure about the general direction, but I think it is good practice to only require that which is necessary.

I wanted to use the constructor methods in this class, which only need `Corecursive` however to construct the class I also have to provide a `Recursive` instance which forces an unnecessary constraint to propagate to call sites.

wemrysi:
@drostron The README absolutely needs an update, thanks for the reminder!

PR #2392, opened 2017-06-23
WIP / integrate MapFuncDerived

rintcius:
First integration of MapFuncDerived - `abs` is the derived function


build.sbt
```diff
@@ -41,7 +41,9 @@ lazy val buildSettings = commonBuildSettings ++ Seq(
 
   scalacOptions ++= Seq(
     "-target:jvm-1.8",
-    "-Ybackend:GenBCode"),
+    "-Ybackend:GenBCode",
+    "-Ypatmat-exhaust-depth","off"),
```
sellout: I don’t think we should commit with this option.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -325,9 +325,9 @@ class CoalesceT[T[_[_]]: BirecursiveT: EqualT: ShowT] extends TTypes[T] {
                 if nm.freeMF(struct >> redRepair) ≟ Free.point(ReduceIndex(0.some)) =>
               rightOnly(id match {
                 case IncludeId =>
-                  Free.roll(ConcatArrays[T, FreeMap](
-                    Free.roll(MakeArray(k)),
-                    Free.roll(MakeArray(elem))))
+                  Free.roll(MFC(ConcatArrays[T, FreeMap](
+                    Free.roll(MFC(MakeArray(k))),
+                    Free.roll(MFC(MakeArray(elem))))))
```
sellout: This could be simplified a bit by using `StaticArray(k, elem)`, but `StaticArray` (and other similar things) then probably need to abstract over `[MF[_](implicit MFC: MapFuncCore :<: MF)`.

mongodb/src/main/scala/quasar/physical/mongodb/planner/JsFuncHandler.scala
```diff
@@ -24,10 +24,13 @@ import quasar.qscript.MapFuncCore
 import quasar.qscript.MapFuncsCore, MapFuncsCore._
 import quasar.std.TemporalPart._
 
-import scalaz.{Free, Scalaz}, Scalaz._
+import scalaz.{Divide => _, _}, Scalaz._
 
 object JsFuncHandler {
```
sellout: As we discussed, this should be a type class now, rather than dealing with the Coproduct directly.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -97,35 +97,36 @@ object MongoDbQScriptPlanner {
...
         recovery(_).point[M],
         expression(funcHandler)))
 
   def getSelector
-    [T[_[_]]: BirecursiveT: ShowT, M[_]: Monad, EX[_]: Traverse]
-    (fm: FreeMap[T])
-    (implicit merr: MonadError_[M, FileSystemError], inj: EX :<: ExprOp)
+      [T[_[_]]: BirecursiveT: ShowT, M[_]: Monad, EX[_]: Traverse]
+      (fm: FreeMap[T])
+      (implicit merr: MonadError_[M, FileSystemError], inj: EX :<: ExprOp)
```
sellout: Trivial, but this isn’t the Scala indentation I’m familiar with.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -157,15 +158,19 @@ object MongoDbQScriptPlanner {
...
 
-    def handleCommon(mf: MapFuncCore[T, Fix[ExprOp]]): Option[Fix[ExprOp]] =
+    val MFC = quasar.qscript.MFC[T]
+    val MFD = quasar.qscript.MFD[T]
+
+    def handleCommon(mf: MapFunc[T, Fix[ExprOp]]): Option[Fix[ExprOp]] =
       funcHandler(mf).map(t => unpack(t.mapSuspension(inj)))
 
-    val handleSpecial: MapFuncCore[T, Fix[ExprOp]] => M[Fix[ExprOp]] = {
+    val handleSpecialCore: MapFuncCore[T, Fix[ExprOp]] => M[Fix[ExprOp]] = {
```
sellout: One of the next steps will be getting rid of this “special” handling, as it doesn’t allow for versioning. I _think_ it means a change to the `FuncHandler` type signature, so we can match on the tree of MapFuncs, not just the current one.

Anyway, not for this PR, so no worries.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -545,56 +571,58 @@ object MongoDbQScriptPlanner {
    * for conversion using \$where.
    */
   def selector[T[_[_]]: RecursiveT: ShowT]:
-      GAlgebra[(T[MapFuncCore[T, ?]], ?), MapFuncCore[T, ?], OutputM[PartialSelector[T]]] = { node =>
+      GAlgebra[(T[MapFunc[T, ?]], ?), MapFunc[T, ?], OutputM[PartialSelector[T]]] = { node =>
```
sellout: This is another thing that should become a type class in the vein of {JS}FuncHandler, too. But this is fine for now.

mongodb/src/main/scala/quasar/physical/mongodb/plannerQScript.scala
```diff
@@ -1142,19 +1175,25 @@ object MongoDbQScriptPlanner {
     tf => (f.lift(tf) \/> tf.project).swap
 
   def elideMoreGeneralGuards[M[_]: Applicative, T[_[_]]: RecursiveT]
-    (subType: Type)
-    (implicit merr: MonadError_[M, FileSystemError])
+      (subType: Type)
+      (implicit merr: MonadError_[M, FileSystemError])
```
sellout: Same indentation thing.

sellout:
And I _think_ this is the change that simplifies at least one case in couchbase/EquiJoinPlanner.scala:
```scala
  object KeyMetaId {
    def unapply(mf: FreeMap[T]): Boolean = mf.resume match {
      case -\/(mfs.ProjectField(Embed(CoEnv(\/-(mfs.Meta(_)))), mfs.StrLit("id"))) => true
      case _                                                                       => false
    }
  }
```

connector/src/main/scala/quasar/qscript/Coalesce.scala
rintcius: As discussed with @sellout this is only there on new mongo branch - will be changed when merging into new mongo

couchbase/src/main/scala/quasar/physical/couchbase/planner/EquiJoinPlanner.scala
```diff
@@ -74,22 +75,16 @@ final class EquiJoinPlanner[
   }
 
   object KeyMetaId {
```
rintcius: see https://github.com/quasar-analytics/quasar/pull/2392#issuecomment-310729282

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncDerivedPlanner.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import scala.Predef.implicitly
+
+import matryoshka._
+import matryoshka.data._
+import matryoshka.implicits._
+import matryoshka.patterns._
+import scalaz._
+
+final class MapFuncDerivedPlanner[T[_[_]]: BirecursiveT: ShowT, F[_]: Applicative: Monad: NameGenerator: PlannerErrorME]
```
wemrysi: This appears to be identical modulo type parameters across all the connectors, might make sense to extract the common bits into an abstract class or maybe even a function to avoid the duplication.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -100,11 +100,12 @@ object FuncHandler {
             val fp = new ExprOpCoreF.fixpoint[Free[EX, A], EX](Free.roll)
             import fp._
 
-            mfc.some collect {
-              case Undefined()           => $literal(Bson.Undefined)
-              case Add(a1, a2)           => $add(a1, a2)
-              case Multiply(a1, a2)      => $multiply(a1, a2)
-              case Subtract(a1, a2)      => $subtract(a1, a2)
+            mfc match {
```
sellout: The problem is the `Constant` case, where you end up with an `Option`. Maybe there’s a better way? But I get frustrated with the places we use `collect`, because we often then end up with a case where the `Option` is outside the pattern match and you need to touch every line in the function, like this.

My suggestion here was to use `.η[M]`, where `MonadFilter[M]` instead of `.some`, but apparently we don’t have that in Scalaz.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncDerivedPlanner.scala
sellout: We had discussed this, and my fear was that the similarities were coincidental.

E.g., Mongo has the functions split up by version, and the shape of Mongo’s algebra is going to have to change to get rid of `handleSpecial` in the planner. But I was hopeful that there was some “true” commonality to extract. Just not sure what it is.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
wemrysi: You could split it into two functions, one that handles constants and one for everything else.

couchbase/src/main/scala/quasar/physical/couchbase/planner/MapFuncDerivedPlanner.scala
wemrysi: Packaging this up into something reusable doesn't mean everything must be forced to use it. If Mongo's implementation doesn't fit the pattern, then it can go its own way, but there appears to be some commonality, even if it is just useful for new connectors to have an easy way to stub out the implementation of `MapFuncDerived`.

Anyhow, I don't think this is serious enough to hold up the merge if you disagree.

sellout: No, that’s fine. I’m just trying to make a distinction between “helper function” and “abstraction”, and this would be the former. I agree that anything that makes it easier to get a new connector up is good.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
rintcius: Yeah, trade-off between an extra function and those extra `some`s. I considered the first but opted for the latter. No strong preference for either though, so fine with changing it. 
Any opinions @sellout ?

rintcius:
Processed review comments in #2431 which also contains `trunc` as a derived mapfunc.

PR #2400, opened 2017-06-27
Add regression tests as test data, and some queries against them.

sellout:
These tests try to use `{_:}`, which is not yet well supported by
connectors (and QScript)


it/src/main/resources/tests/backendStatuses.test
```diff
@@ -0,0 +1,29 @@
...
+        "mongodb_3_2":       "pending",
+        "mongodb_3_4":       "pending",
+        "mongodb_read_only": "pending",
+        "mongodb_q_3_2":     "pending",
+        "postgresql":        "pending",
+        "spark_hdfs":        "pending",
+        "spark_local":       "pending"
+    },
+    "data": "newTests.data",
+    "query": "select backends{_:} as backend, backends{:_} as status, name from newTests",
```
alissapajer: @sellout btw great idea to use the the integration tests as data! They make for a really good example of how to use `{_:}` and `{:_}`.

PR #2405, opened 2017-06-28
Add tags to the EJson encoding for TypeStat

### LINKED ISSUE #2404: SST TypeStat measure values are hard to decode without looking deeply at the structure
From some test data like this:

``` json
{ "testString": "abc" }
{ "testString": "hello world" }
{ "testString": "qwerty" }
```

The returned structure is:

``` json

            "testString": {
                "measure": {
                    "count": 3,
                    "min": "abc",
                    "max": "qwerty",
                    "minLength": 3,
                    "maxLength": 11
                },
                "structure": {
                    "type": "array",
                    "of": {
                        "measure": {
                            "count": 3
                        },
                        "structure": {
                            "type": "character"
                        }
                    }
                }
            }
```

Which isn't incorrect, but it makes the `TypeStat` difficult to decode - we don't know it's a string, so either we have to try every type stat parse permutation and backtrack, or ideally we'd include the information somewhere in the result so that we know it's a string - this is probably related to the likes of #2354, in that stringiness might be available as metadata? 

Another option would be to include a `tag` field for each `TypeStat` constructor so we could decode unambiguously based on that alone. That would be ideal in fact, as we could solve `const` stats the same way then, as that has basically the same problem.

garyb:
Resolves #2404.

Please tell me what hiedous goofs I've made :wink:.

I'm not sure if we want to call it `tag`, `type` seemed like a decent alternative too, but I thought that might be a bit odd when we have multiple things that can be considered "type" in the response.

### DIFF
--- a/frontend/src/main/scala/quasar/sst/TypeStat.scala
+++ b/frontend/src/main/scala/quasar/sst/TypeStat.scala
@@ -176,14 +176,16 @@ sealed abstract class TypeStatInstances {
         def optEntry[B: EncodeEJson](k: String, b: Option[B]): List[(String, J)] =
           b.map(x => (k, x.asEJson[J])).toList
 
-        def minmax[B: EncodeEJson](c: A, mn: B, mx: B): J =
+        def minmax[B: EncodeEJson](kind: String, c: A, mn: B, mx: B): J =
           emap(
+            "kind"   -> kind.asEJson[J],
             "count" -> c.asEJson[J],
             "min"   -> mn.asEJson[J],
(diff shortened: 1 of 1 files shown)


wemrysi:
@garyb Yeah, I think avoiding `type` is probably good. What about `kind`, it is a bit more descriptive than `tag`, though the lack of meaning in the latter might be considered a feature?

wemrysi:
Also, thanks for the PR!

garyb:
Updated to `kind` :+1:

PR #2406, opened 2017-06-28
Add “timeout” status for regression tests.

sellout:
The “timeout” status skips tests on CI, but runs them locally.

It will also tell you if a “timeout” test should be changed to some “lesser”
failure, and vice versa.


sellout:
@drostron, @rabbitonweb, @wemrysi – I added you all as reviewers since you’re the ones who will likely run into some unexpected failures when you first try this locally.

You’ll probably get some failures with messages like `change “timeout” to “pending”`. This is to help us run as many tests as possible on Travis. If they’re not timing out, we should run them. And we should only use `"skip"` when either 1. the query is crashing the process or 2. it’s a connector-specific test (and those should hopefully go away as well).

alissapajer:
With this new status we should be able to remove all the notes in the integration tests that say things like "This test is skipped on X because it times out."

sellout:
@alissapajer Good point – and that already shows me some couchbase ones that should be made pending.

alissapajer:
@sellout also, how are you updating the `newTests.data` and `oldTests.data` files? Also, what is the difference between the two? It would be really cool if one of those could be kept up to date with our actual regressions tests via a script, and then we could write meaningful queries like "how many tests are timing out for couchbase".

sellout:
I’m not updating those data files. I considered keeping them up-to-date, but decided against it, because it doesn’t really matter for the tests and would just make them break. I think we can update them manually (`find . -name ".test" -exec tr …`) periodically if they no longer reflect the structure.

We should be able to use the spark-local connector via the REPL to write queries against our tests, no? Or importing them into whatever other data store using our import scripts.

`oldTests` is from master and `newTests` is from the new Mongo branch, since I was originally trying to figure out where things differ as I work on new Mongo.

alissapajer:
@sellout yeah, thinking about it I agree it doesn't make sense to change the test data frequently because then tests will break. And we can always query the data files locally to ask real questions. It doesn't matter if integration test questions and answers are accurate in that sense. So :+1:

it/src/test/scala/quasar/regression/QueryRegressionTest.scala
```diff
@@ -216,24 +221,42 @@ abstract class QueryRegressionTest[S[_]](
...
+          case _ =>
+            OrderPreserved
+        })
+
+    exp.backends.get(backendName) match {
+        case Some(TestDirective.Timeout) => result.map {
+          case execute.Success(_, _) =>
+            execute.Failure(s"Fixed now, you should remove the “timeout” status.")
+          case execute.Failure(m, _, _, _) =>
+            execute.Failure(s"Failed with “$m”, you should change the “timeout” status.")
```
wemrysi: ... to pending?

sellout: Not necessarily … in this case, it could be `pending`, or one of the three `ignore…` statuses, depending on what message it actually failed with.

Issue #2110, opened 2017-04-27
Handle shifting inside a join condition in qscript


alissapajer:
Example query:
```
select zips.city, smallZips.loc from zips join smallZips on zips.loc[*]{*} = smallZips[*]
```
Currently qscript detects the shift in the join condition as a "non-mapping join condition" and fails to create a join.

Fixing this should enable `joins/nonJsJoinCondition.test` and `joins/symmetricNonJsJoinCondition.test`. Additionally a test with multiple flattens on a single side (as in the example above) should be added.

alissapajer:
Additionally, to enable those integration tests, we need to match on `Constant(Arr(List(NA)))` and `Constant(Obj(Map( -> NA)))` in the LP optimizer where we move `Typecheck` outside of join conditions. With new mongo, it should be sufficient to just match on `NA`.

alissapajer:
The solution will be to extract the `LeftShift` out of the join condition into the join branch `FreeQS`. There the `LeftShift` can provide access to both the shifted and unshifted values.

alissapajer:
A duplicate (or at least an extension of) #2133. Fixing #2133 will enable this to be fixed.

alissapajer:
Also should note that this query will not execute on _any_ qscript-based backend until this issue is fixed.

PR #2407, opened 2017-06-29
Match on Data.NA array and object in typecheck join filtering.

alissapajer:


alissapajer:
@sellout this is an old commit I want to get off my wip branch. We hit this case in a query like in #2110.

PR #2409, opened 2017-06-30
Fix pretty printing of modules

jedesah:
Incorrect pretty printing of module mounts was making them invalid and unable to be mounted.

- Fix pretty printing of import statements
  - First off the `posicCodec.printPath` was not being called yielding completely bogus Sql^2
  - Secondly backticks need to be espcaped
- Generator for `Statement` was changed to produce only alpha characters in import paths for now in order for the round-trip to work
- Add instances of `Shrink` for `Fix[Sql]` and `Statement[Fix[Sql]]`
- remove unnecessary call to `makeTables(Nil)` in tests (since it's already being called in the parser)
- Add a space in pretty printing between arguments of a function
- Add tests for module round-trip and specific backtick situation
- Fix small bug in function `_qq`

### DIFF
--- a/frontend/src/main/scala/quasar/sql/ScopedExpr.scala
+++ b/frontend/src/main/scala/quasar/sql/ScopedExpr.scala
@@ -19,8 +19,10 @@ package quasar.sql
 import slamdata.Predef._
 import quasar._, RenderTree.ops._
 
+import matryoshka.Recursive
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
+import scalaz.Liskov._
 
 @Lenses final case class ScopedExpr[T](expr: T, scope: List[Statement[T]]) {
   def mapExpressionM[M[_]:Functor](f: T => M[T]): M[ScopedExpr[T]] =
@@ -29,6 +31,10 @@ import scalaz._, Scalaz._
     scope.collect { case i: Import[_] => i }
   def defs: List[FunctionDecl[T]] =
     scope.collect { case d: FunctionDecl[_] => d }
+  def pprint[T1](implicit ev: T <~< T1, T1: Recursive.Aux[T1, Sql]): String = {
+    val scopeString = if (scope.isEmpty) "" else scope.pprint + ";\n"
+    scopeString + sql.pprint(ev(expr))
+  }
 }
 
 object ScopedExpr {
--- a/frontend/src/main/scala/quasar/sql/ScopedExpr.scala
+++ b/frontend/src/main/scala/quasar/sql/ScopedExpr.scala
@@ -22,7 +22,6 @@ import quasar._, RenderTree.ops._
 import matryoshka.Recursive
 import monocle.macros.Lenses
 import scalaz._, Scalaz._
-import scalaz.Liskov._
 
 @Lenses final case class ScopedExpr[T](expr: T, scope: List[Statement[T]]) {
   def mapExpressionM[M[_]:Functor](f: T => M[T]): M[ScopedExpr[T]] =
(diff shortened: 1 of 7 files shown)


jedesah:
@natefaubion This will fix the error you ran into the other day

frontend/src/main/scala/quasar/sql/ScopedExpr.scala
```diff
@@ -29,6 +31,10 @@ import scalaz._, Scalaz._
     scope.collect { case i: Import[_] => i }
   def defs: List[FunctionDecl[T]] =
     scope.collect { case d: FunctionDecl[_] => d }
+  def pprint[T1](implicit ev: T <~< T1, T1: Recursive.Aux[T1, Sql]): String = {
```
jedesah: Yes, good point, I don't see why that wouldn't work.

frontend/src/main/scala/quasar/sql/package.scala
```diff
@@ -216,7 +220,7 @@ package object sql {
 
   private def _qq(delimiter: String, s: String): String = s match {
     case SimpleNamePattern() => s
-    case _                   => delimiter + s.replace("\\", "\\\\").replace(delimiter, "\\`") + delimiter
+    case _                   => delimiter + s.replace("\\", "\\\\").replace(delimiter, "\\" + delimiter) + delimiter
```
jedesah: It's more like future proofing this function in case it ever gets called with something else than `"`"`. But there is no test at the `pprint` level that would expose a bug with the current implementation. I don't think it's worth testing `_qq` in isolation

frontend/src/test/scala/quasar/sql/ExprArbitrary.scala
```diff
@@ -19,18 +19,33 @@ package quasar.sql
...
 import java.time.{Duration, Instant}
 import scalaz._, Scalaz._
 import scalaz.scalacheck.ScalaCheckBinding._
 
 trait ExprArbitrary {
   implicit val exprArbitrary: Arbitrary[Fix[Sql]] = Arbitrary(selectGen(4))
 
+  implicit val exprShrink: Shrink[Fix[Sql]] = Shrink (expr => (expr.project match {
+    case s @ Select(distinct, proj, relation, filter, groupBy, orderBy) =>
+      shrink(distinct).map(d => s.copy(isDistinct = d)) append
```
jedesah: Good catch. I don't think `distinct` can be shrunk, I don't know what I was thinking, I will remove it.

jedesah: Ah, I wanted to shrink `projections` which is a list. Huh `Shrink.shringAny`, what does that do?

drostron: Provides a `Shrink` that returns an empty stream.

jedesah: Oh, that's very useful...

jedesah:
@drostron Since you approved the PR and the build is green. I'll merge it now so that it's ready for @natefaubion Monday morning.

jedesah:
@drostron I guess to put a version label, hopefully you agree it's a revision

sellout:
So, I know this is already merged, but there was an existing `Shrink[Fix[Sql]]` (I know, hidden in Mongo planner tests for some unknown reason). Could you try to unify them?

Also, things like
```scala
implicit val statementArbitrary: Arbitrary[Statement[Fix[Sql]]]
```
should be
```scala
implicit val statementDelayArbitrary: Delay[Arbitrary, Statement]
// effectively: implicit def statementArbitrary[A: Arbitrary]: Arbitrary[Statement[A]],
// but avoiding problems with instance resolution.
```
And for `Shrink`, you can additionally have a lower-priority instance without the constraint on `A`:
```scala
implicit def statementArbitrary[A]: Arbitrary[Statement[A]]
```

Finally, Matryoshka has myriad ways to go from a shrinkable functor to a shrinkable fixed-point. So you should be able to define either
```scala
implicit def sqlShrink[A]: Shrink[Sql[A]]
```
or at least
```scala
implicit val sqlDelayShrink: Delay[Shrink, Sql]
```
instead of an instance on `Fix[Sql]`.

drostron:
@jedesah are you able to follow up on @sellout's [comment](https://github.com/quasar-analytics/quasar/pull/2409#issuecomment-312512098)?

jedesah:
@drostron I'll and get a new PR in by end of week to address this stuff

drostron:
@jedesah Excellent, thanks.

PR #2419, opened 2017-07-03
Initial AST for cts:query

jsantos17:


marklogic/src/main/scala/quasar/physical/marklogic/MarkLogic.scala
```diff
@@ -118,12 +124,27 @@ final class MarkLogic(readChunkSize: Positive, writeChunkSize: Positive)
...
   }
 
-  def plan[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT](qs: T[QSM[T, ?]]): Backend[Repr] =
+  def plan[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT](qs: T[QSM[T, ?]]): Backend[Repr] = {
+    def doPlan(cfg: Config): Backend[MainModule] = {
+      import cfg.{searchOptions, structuralPlannerM}
+      MainModule.fromWritten(
+        qs.cataM(cfg.planner[T].plan[Q, V])
+          .flatMap(_.fold(s =>
+            Search.plan[cfg.M, Q, V, cfg.FMT](s, κ(emptySeq))(
```
jsantos17: Instead `κ(emptySeq)` here we probably want a function that maps from the atomic types [here](https://docs.marklogic.com/javadoc/xcc/com/marklogic/xcc/types/XdmAtomic.html) (or a similar wrapper) to the corresponding cosntructor in `XQuery`

marklogic/src/main/scala/quasar/physical/marklogic/qscript/Search.scala
```diff
@@ -0,0 +1,72 @@
...
+import scalaz.syntax.monad._
+import xml.name._
+
+@Lenses
+/** Represents a cts:search expression. */
+final case class Search[Q](query: Q, idStatus: IdStatus)
+
+object Search {
+  def plan[F[_]: Monad: PrologW, Q, FMT, V, T[_[_]]: BirecursiveT]
+    (s: Search[Q], f: V => T[EJson])(
```
jsantos17: @wemrysi so this is what I ended up with after trying to use `T[EJson]` as the `V` in the `Query` AST. However, after thinking some more about this I really don't think we want `EJson` as the representation for literals. Everywhere we use `V` in the AST it's to represent an `xs:anyAtomicType`. The definition of ["atomic" in XML Schema](https://www.w3.org/TR/2009/WD-xmlschema11-2-20091203/#dt-atomic-value) seems to suggest objects are not atomic values, so using `T[EJson]` allows us to represent invalid state. We can always call `fn:data` to atomize any value, but this is little more than calling `fn:string` from the tests I have done. For example:

```xquery
fn:data(<a>
          <b>Slam</b>
          <c>Data</c>
          <c>ML</c>
        </a>)
```

yields `SlamDataML`.

Unless I'm misreading XML Schema document, it seems to me that EJson here is overkill and a simpler, non-recursive representation would suffice. What do you think?

wemrysi: I was thinking more along the lines of the previous signature that included `V => F[XQuery]` and then we'd just instantiate `V` to be `EJson` at the call-site, there is no reason it needs to show up here, is there?

And yes, `EJson` is a bit "bigger" than necessary so a first-class representation of atomic types will probably be useful in the long run, I was suggesting we defer some of that work until we actually need it.

For this PR, since we aren't actually creating anything that uses literals, do we need a representation yet? Can we stick with `Unit` and tackle this problem as part of the work for #2425?

jsantos17: My train of thought was that users of `Search.plan` would need to provide evidence their literals can be represented as `EJson` so we can plan it with `EJsonPlanner`, but it's also true that's not necessary, the caller can choose to pass `V => F[XQuery]` and use `EJsonPlanner` if they want.

So what about this: let's fix the type to `Unit` for now, revert the type to `V => F[XQuery]`, stub `f` with `κ(emptySeq)` and merge this? We can discuss the representation later.

marklogic/src/main/scala/quasar/physical/marklogic/cts/ComparisonOp.scala
```diff
@@ -0,0 +1,85 @@
...
+object ComparisonOp {
+  final object LT extends ComparisonOp
+  final object LE extends ComparisonOp
+  final object GT extends ComparisonOp
+  final object GE extends ComparisonOp
+  final object EQ extends ComparisonOp
+  final object NE extends ComparisonOp
+
+  val toXQuery: ComparisonOp => XQuery = {
+    case LT => "<".xqy
```
wemrysi: I think all of these should be `.xs` instead of `.xqy`.

marklogic/src/main/scala/quasar/physical/marklogic/cts/Query.scala
```diff
@@ -0,0 +1,429 @@
...
+    op: ComparisonOp,
+    values: IList[V]
+  ) extends Query[V, A]
+
+  final case class True[V, A]() extends Query[V, A]
+
+  // TODO: Options
+  final case class Word[V, A](words: IList[String]) extends Query[V, A]
+
+  def mkSeqF[F[_]: Foldable: Functor, A](fa: F[A])(f: A => XQuery): XQuery =
```
wemrysi: These three functions should be private to avoid polluting the namespace and so that nothing inadvertently relies on them (if we find the need for one or more of these elsewhere, they should be moved somewhere more general).

marklogic/src/main/scala/quasar/physical/marklogic/xquery/cts.scala
```diff
@@ -79,4 +130,9 @@ object cts {
     forestIds: IList[XQuery] = IList.empty
   ): XQuery =
     XQuery(s"cts:uris($start, ${mkSeq(options)}, $query, ${qualityWeight getOrElse "1.0".xqy}, ${mkSeq(forestIds)})")
+
+  val True: XQuery =
+    XQuery("cts:true-query()")
```
wemrysi: Missing a newline.

jsantos17:
Ready for another look @wemrysi

marklogic/src/main/scala/quasar/physical/marklogic/cts/Query.scala
jsantos17: Oh right. Somehow forgot to add `private`.

wemrysi:
@jsantos17 Once this is ready to merge, what do you think about squashing some of your commits and opening a new PR just to tidy up the history a bit as there was some back-and-forth during the WIP process?

jsantos17:
@wemrysi definitely. I'll close this PR and open a new one with a clean history after I finish with the review.

marklogic/src/main/scala/quasar/physical/marklogic/cts/ComparisonOp.scala
jsantos17: [Wouldn't that cause the `"<" ` to be escaped?](https://github.com/quasar-analytics/quasar/blob/master/marklogic/src/main/scala/quasar/physical/marklogic/xquery/syntax.scala#L100) As far as I understand we don't want them escaped. How about something like `XQuery.StringLit("<")`?

wemrysi: Ah, good catch, yeah that sounds good, or even `XQuery.StringLit(op match { ... })`.

jsantos17:

PR #2423, opened 2017-07-05
FBE Filter and some MapFunc

### LINKED ISSUE #1992: Enable all 1:1 mapfunc stuff
For all the sql^2 mapfuncs in quasar, if there's already a 1:1 mapping on top of `Table`, do that.  Otherwise, don't.

alissapajer:
Implements `Filter` and various `MapFunc` for FBE.

Deletes `MorphContext`.

Closes #1992

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/MapFuncPlanner.scala
+++ b/mimir/src/main/scala/quasar/mimir/MapFuncPlanner.scala
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.mimir
+
+import quasar.Data
+import quasar.qscript.{MapFuncCore, MapFuncsCore}
+
+import quasar.blueeyes.json.JValue
+import quasar.precog.common.RValue
+import quasar.yggdrasil.table.cf.math
+import quasar.yggdrasil.table.cf.util.Undefined
+
+import matryoshka.{AlgebraM, RecursiveT}
+import matryoshka.implicits._
+
+import scalaz.Applicative
+import scalaz.syntax.applicative._
+
+final class MapFuncPlanner[T[_[_]]: RecursiveT, M[_], F[_]: Applicative] {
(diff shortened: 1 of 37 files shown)


mimir/src/main/scala/quasar/mimir/MapFuncPlanner.scala
```diff
@@ -0,0 +1,154 @@
...
+    case MapFuncsCore.StartOfDay(a1) => ???
+    case MapFuncsCore.TemporalTrunc(part, a1) => ???
+    case MapFuncsCore.TimeOfDay(a1) => ???
+    case MapFuncsCore.ToTimestamp(a1) => ???
+    case MapFuncsCore.Now() => ???
+
+    case MapFuncsCore.TypeOf(a1) => ???
+
+    case MapFuncsCore.Negate(a1) =>
+      (cake.trans.Map1[cake.trans.Source1](a1, math.Negate): cake.trans.TransSpec1).point[F]
```
djspiewak: Suggestion: `import cake.trans._`

Also, are all of the type ascriptions necessary now that this is split out and no longer has to deal with `Backend` directly?

alissapajer: That is a reasonable import.

I will check on the type ascriptions.

alissapajer: @djspiewak so without the ascription:
```
[error]  found   : F[cake.trans.Map2[cake.trans.Source1]]
[error]  required: F[cake.trans.TransSpec1]
```
I haven't found a way around that... any ideas?

mimir/src/main/scala/quasar/mimir/MapFuncPlanner.scala
```diff
@@ -32,123 +32,126 @@ import scalaz.syntax.applicative._
...
-      (cake.trans.DerefObjectDynamic[cake.trans.Source1](src, field): cake.trans.TransSpec1).point[F]
-    case MapFuncsCore.DeleteField(src, field) => ???
-
-    case MapFuncsCore.Meta(a1) => ???
-
-    case MapFuncsCore.Range(from, to) => ???
-
-    // FIXME if fallback is not undefined don't throw this away
-    case MapFuncsCore.Guard(_, _, a2, _) => a2.point[F]
+    import cake.trans._
```
djspiewak: Minor formatting issue (doesn't block the merge), but IMO this would have been better in the class scope (i.e. on the line above `def plan`) precisely to avoid this situation where you have a partial function that has to be indented strangely.

alissapajer: @djspiewak I agree the formatting is not ideal, but `cake` is passed in as an argument to `plan` so this is the broadest scope in which I could place the import.

djspiewak: Oh right. I can't read.

mimir/src/main/scala/quasar/mimir/MapFuncPlanner.scala
djspiewak: Oh, it's just not lubbing through the dependent type.  That's super-annoying.

Issue #2272, opened 2017-06-01
MongoDB Arithmetic Operators - Phase A


rintcius:
Implement some of the arithmetic operators as mentioned in https://github.com/quasar-analytics/quasar/issues/1954

### Arithmetic Operators

 - [x] Support `$abs`
 - [x] Support `$ceil`
 - [x] Support `$floor`
 - [x] Support `$trunc`

zeeiyerWork:
Assigning June milestone based on chat with @rintcius

zeeiyerWork:
@rintcius, @sellout -> Take it you will create supporting tickets if needed in other epics, if the work is mostly about exposing these in SQL-2 and LP.

alissapajer:
@zeeiyerWork these math operations are low priority per the discussion here https://github.com/quasar-analytics/quasar/issues/2198#issuecomment-316174962. Would it be okay if we don't include this is the "New Mongo" epic?

rintcius:
verified

PR #2431, opened 2017-07-05
integrate derived mapfunc - add abs & trunc

rintcius:
Integrate derived mapfunc, adding `abs` and `trunc` as derived mapfuncs.


alissapajer:
@rintcius is this ready for review now?

rintcius:
@alissapajer yes, it is. Is there something I haven't marked correctly in the PR?

alissapajer:
@rintcius you marked everything correctly, I was just confirming that because you added reviewers, that meant that it was ready to be looked at!

We haven't talked about this as a group, but I'm thinking that it also makes sense to remove the "WIP" in the PR title as well when it is no longer a work in progress and is ready for review.

rintcius:
Right, that sounds good. I'll change the title.

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
```diff
@@ -51,6 +67,11 @@ sealed abstract class ExpandMapFuncInstances extends ExpandMapFuncInstancesʹ {
               Free.roll(MFC(Negate(a.point[Free[OUT, ?]]))),
               a.point[Free[OUT, ?]]
             ))
+          case D.Trunc(a) =>
+            MFC(Subtract(
+              a.point[Free[OUT, ?]],
+              Free.roll(MFC(Modulo(a.point[Free[OUT,?]], Free.roll(MFC(Constant(ejson.EJson.fromExt(ejson.int(1))))))))
+            ))
```
alissapajer: Dangling parents (here and above on line 69).

alissapajer:
@rintcius is there a ticket (or tickets) that this PR should be closing? If not then feel free to open some tickets and then close them in this PR. Maybe one for `abs`, one for `trunc`, and one for the other integration work? Or whatever makes most sense.

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
```diff
@@ -51,6 +67,11 @@ sealed abstract class ExpandMapFuncInstances extends ExpandMapFuncInstancesʹ {
               Free.roll(MFC(Negate(a.point[Free[OUT, ?]]))),
               a.point[Free[OUT, ?]]
             ))
+          case D.Trunc(a) =>
+            MFC(Subtract(
+              a.point[Free[OUT, ?]],
+              Free.roll(MFC(Modulo(a.point[Free[OUT,?]], Free.roll(MFC(Constant(ejson.EJson.fromExt(ejson.int(1))))))))
```
alissapajer: Just wanted to confirm that this is the intended implementation. Currently it's implemented as "round down to the nearest integer". Is this the intent?
```
let a = 8.5
a - (a mod 1) = a - (0.5) = 8.0

let b = -8.5
b - (b mod 1) = b - (0.5) = -9.0
```
The tests confirm this behavior, just wanted to ask about it explicitly.

connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -135,87 +141,91 @@ object MapFuncCore {
...
-        case Lower(ConstEC(ejson.Str(v1))) =>
+        case MFC(Lower(ConstEC(ejson.Str(v1)))) =>
           EC.inj(ejson.Str(v1.toLowerCase)).some
-        case Upper(ConstEC(ejson.Str(v1))) =>
+        case MFC(Upper(ConstEC(ejson.Str(v1)))) =>
           EC.inj(ejson.Str(v1.toUpperCase)).some
 
         // structural
-        case MakeArray(Embed(CoEnv(\/-(Constant(v1))))) =>
+        case MFC(MakeArray(Embed(CoEnv(\/-(MFC(Constant(v1))))))) =>
```
alissapajer: The pattern matching with `MFC` (and `CoEnv` and `Embed` and `Either`) is getting pretty ugly. I think it's fine like this for this PR, but at some point I'd like to create some extractors for these cases. Not sure exactly what the extractors will look like yet, but I think it'd be much easier to work in this part of the code base if you just had to write something like
```scala
case MakeArrayExtractor(ConstantExtractor(v1)) =>
```

connector/src/test/scala/quasar/qscript/QScript.scala
```diff
@@ -706,35 +706,35 @@ class QScriptSpec
...
                  ProjectIndexR(RightSideF, IntLit(1)),
                  Type.Obj(ScalaMap(),Some(Type.Top)),
                  ProjectIndexR(RightSideF, IntLit(1)),
-                 Free.roll(Undefined()))))),
-             Free.roll(MakeArray(
-               Free.roll(Guard(
-                 Free.roll(ProjectField(
-                   Free.roll(Guard(
+                 Free.roll(MFC(Undefined())))))))),
+             Free.roll(MFC(MakeArray(
```
alissapajer: For the readability of the tests, we should switch all these over to use `MakeArrayR`, `ProjectIndexR` etc. It's fine like this for this PR, just if you're ever in this file (or any of the other qscript test files) and feel like doing some massive search and replaces :)

And the other advantage (aside from readability) is that if we ever have to make a change like this again, we'd just have to make it where we define `MakeArrayR` (etc.) and not on every line in the test file like we did here.

alissapajer: And as I think about it, that's a good reason to have the nicer extractors for the pattern matching like I mentioned before. If we do those extractors right, then we could just have to make this kind of change in the extractors and not everywhere too.

it/src/main/resources/tests/derivedFunctionAbs.test
```diff
@@ -0,0 +1,29 @@
+{
+    "name": "derived function abs",
+    "backends": {
+        "marklogic_xml": "skip",
+        "mongodb_q_3_2": "timeout",
+        "couchbase": "pending",
+        "marklogic_json": "pending",
+        "mimir": "pending",
+        "spark_local": "pending",
+        "spark_hdfs": "pending"
```
alissapajer: Super minor, but could you put the list of backends here in alphabetical order?

alissapajer: Also, I'm surprised that this isn't passing for any qscript-based backends. Since `abs` is just implemented in terms of other basic map funcs, I think these tests should be passing. Do you know why none of these are passing? We should look at the qscript and see what we can figure out.

it/src/main/resources/tests/derivedFunctionTrunc.test
```diff
@@ -0,0 +1,29 @@
+{
+    "name": "derived function trunc",
+    "backends": {
+      "couchbase": "timeout",
+      "marklogic_json": "timeout",
+      "marklogic_xml": "timeout",
+      "mimir": "timeout",
+      "mongodb_q_3_2": "timeout",
+      "spark_local": "timeout",
+      "spark_hdfs": "timeout"
```
alissapajer: Again, I don't think it's right that all these are timing out. Some should at least be passing, I think. Let's figure this out before merge. Since the tests are passing for old mongo, my best uninformed guess is that something is wrong with the qscript.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -460,15 +460,15 @@ class PlannerSpec
...
           def right: FreeQS = Free.roll(QCT.inj(Filter(HoleQS, func("US"))))
           def key: FreeMap = ProjectFieldR(HoleF, StrLit("age"))
-          def combine: JoinFunc = Free.roll(ConcatMaps(
-            Free.roll(MakeMap(StrLit(JoinDir.Left.name), LeftSideF)),
-            Free.roll(MakeMap(StrLit(JoinDir.Right.name), RightSideF))
-          ))
+          def combine: JoinFunc = Free.roll(MFC(ConcatMaps(
+            Free.roll(MFC(MakeMap(StrLit(JoinDir.Left.name), LeftSideF))),
+            Free.roll(MFC(MakeMap(StrLit(JoinDir.Right.name), RightSideF)))
+          )))
```
alissapajer: A few cases of dangling parens. They were like this before, but might as well clean it up now :)

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
rintcius: No, the intent is that the latter case results in `-8.0` and this should also reflect in the tests.
The surprising thing is that  `-8.5 mod 1 = -0.5`

it/src/main/resources/tests/derivedFunctionAbs.test
rintcius: Yes, that's definitely something to investigate. Was at first thinking we could do it after merge, but thinking a bit more about it I agree it's better to do it before.

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
alissapajer: So, we should check the integration test cases to see what they are actually doing. I meant to look at the integration test results but actually didn't. And also we should probably add a unit test along with the scalacheck ones so we can make sure we have the behavior documented.

So, I was assuming that the `mod` of any number (positive or negative) would be positive. Since I was thinking of modulo as remainder. But I have heard of other conventions where the `mod` result can be negative. I would argue that the `mod` value should always be positive, at least from a mathematical point of view, but we need to confirm how it is implemented for each backend. There should be unit tests for this.

Total long shot, but I wonder if qscript backends have implemented `mod` one way and old mongo has it implemented another way and that's why the tests are failing?

it/src/main/resources/tests/derivedFunctionTrunc.test
rintcius: Indeed, I added a 2nd function to gather more info about these failures

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
alissapajer: fwiw, Scala has negative `mod`:
```
scala> -5.5 % 1
res2: Double = -0.5
```

rintcius:
@alissapajer there's this ticket #2272 - the intention is to mark the relevant functions as checked after merge. Does that work?

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
alissapajer: @rintcius ok so the unit tests are expecting what you expected this to do:
```
scala> BigDecimal(-5.5).setScale(0, BigDecimal.RoundingMode.DOWN)
res4: scala.math.BigDecimal = -5
```
I thought `DOWN` meant the number would get smaller, who knew :)

it/src/main/resources/tests/derivedFunctionAbs.test
alissapajer: Yeah, let's figure out what's going on. Then depending on what exactly is wrong, we can decide if we'll fix it before or after merge.

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
rintcius: not the best name :)

rintcius: Anything you like me to change regarding this?

alissapajer: Ok so I just looked at the integration test results that you added, and they expect `trunc(-5.5)` to be `-5`. So all the tests are in agreement with what you had assumed. And I had just assumed `mod` would behave differently. Additionally this result is in agreement with what appears to be the consensus on the definition of `trunc`. So :+1: to the tests expecting what you expected.

So, I think what we need to do is add a unit test that confirms that all the backends compute mod in the same way. For whatever reason, we seem to only have unit tests for `Modulo` that compute the mod of a positive number. Could you add a test for `Modulo(x, y)` where `x < 0`? The tests I'm looking at are in `frontend/src/test/scala/quasar/std/StdLibSpec.scala`. Then I'll feel confident that this implementation of `trunc` is doing what we expect (as enforced by tests) for all backends.

alissapajer: Yeah, so assuming the unit test we add for `Modulo` behaves the same for all backends, (`-5.5 mod 1 = -0.5`) then I don't think we need to change your implementation here.

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
```diff
@@ -33,6 +38,17 @@ object ExpandMapFunc extends ExpandMapFuncInstances {
...
+
+  def expand[T[_[_]]: CorecursiveT, F[_]: Monad, A]
+    (core: AlgebraM[F, MapFuncCore[T, ?], A],
+      derived: AlgebraM[(Option ∘ F)#λ, MapFuncDerived[T, ?], A])
+      : AlgebraM[F, MapFuncDerived[T, ?], A] = { f =>
+    derived(f).getOrElse(
+      Free.roll(mapFuncDerived[T, MapFuncCore[T, ?]].expand(f)).cataM(
+        interpretM(scala.Predef.implicitly[Monad[F]].point[A](_), core)
+      )
+    )
```
rintcius: Ok, but that Scala style you referred to allows dangling parens, isn't it (last paragraph)? I saw non-dangling parens in https://github.com/quasar-analytics/quasar/wiki/Scala%5BQuasar%5D though, where I personally would have used dangling ones (and also in the code), so if this is the agreed upon style I'll change it.
 
Anyone tried to encode our style using http://scalameta.org/scalafmt/ so that's automated and we don't have to spend time on manually changing it?

alissapajer: I read the Scala style guide as saying that a trailing paren is only acceptable in the situation when you are "disabling semicolon inference". And in all other situations (i.e. "parenthetical blocks wrap across lines") the paren should not be dangling. Though I think I might have even referenced the wrong section of the style guide; in this case we have function arguments being applied with parens, but I can't find any reference to that in the guide.

Not having dangling parens is something that we have generally agreed to avoid. We should add an example like this to our wiki. But with that said, style things are in general always up for discussion.

Regarding scalafmt, we've talked about that from time to time. I think the problem is that there are some style things we have not come to a consensus on. For example, some people like to line up `=>` in a pattern match while some don't. And so we let people have their own style for some situations while trying to enforce certain style in others. It's definitely not the most consistent approach, but I think it's let people feel like we're not harshly dictating style while still maintaining some semblance of consistency.

rintcius: I have some personal styling preferences, but I'll be very glad to set them aside if we can delegate them to a formatter (assuming that it's formatting decent enough).

alissapajer: Personally, I'd rather just accept your (and everyone's) personal styling preferences than delegate to a formatter. We can move this discussion to slack if you'd like, since at this point it's not really relevant to the PR. In the past we've debated the merits of different styles, and at least for me that has helped me be open to different styles.

alissapajer:
@rintcius that ticket works for me!

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
alissapajer: But yeah, we should have this conversation with everyone again, especially since we've had a lot of new people join, and I don't mean for my personal styling preference to dictate over others' :)

connector/src/main/scala/quasar/qscript/ExpandMapFunc.scala
rintcius: Ok, I dug a bit deeper in this. 

So it appears that there are 2 different interpretations of `%` depending on the language/tool. I'll call them `modulo` (always >= 0) and `remainder` (may be negative) (as in https://stackoverflow.com/questions/13683563/whats-the-difference-between-mod-and-remainder).
Java/scala, Postgres and quasar agree on that `%` stands for `remainder`, so it can be negative.

```
select -8.9 % 1, 8.9 % 1

 0     | 1    |
-------|------|
  -0.9 |  0.9 |
```

So I think this is all consistent with each other & it looks to me that this is sufficient to demand the `remainder` interpretation from connectors. 
Also `%` doesn't seem to be exposed via a named function so that's good too and even the description of the function looks consistent `Finds the remainder of one number divided by another`. 

I think we should rename the internally used term `Modulo` to `Remainder` though to make this choice more clear.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala
```diff
@@ -16,165 +16,38 @@
...
-
-    // FIXME: This isn't correct, just an interim impl to allow some queries to execute.
-    case Guard(_, _, cont, _)         => s"(: GUARD CONT :)$cont".xqy.point[F]
-  }
-
-  ////
-
-  private def asDate(x: XQuery)     = SP.castIfNode(x) >>= (lib.asDate[F] apply _)
-  private def asDateTime(x: XQuery) = SP.castIfNode(x) >>= (lib.asDateTime[F] apply _)
+abstract class MapFuncPlanner[F[_], MF[_], T[_[_]]] {
```
wemrysi: I think this should be `MapFuncPlanner[F[_], FMT, MF[_]]`.

rintcius:
@alissapajer @wemrysi ready for re-review. 

Note that I did pull the commits that were removed in the forced push to master. They shouldn't be part of this PR, but please review.

alissapajer:
@rintcius can you merge latest master in to this PR? The travis build is failing at least in part because of an unrelated failing test that was checked in by mistake.

alissapajer:
@rintcius there are some failing tests as well in travis. I didn't look at each job, but MarkLogic is failing the derived integration tests due to incorrect field order, so you can just `ignoreFieldOrder` there. Additionally it's failing in the stdlib specs as well.

Overall your changes look great, just need to get all the tests passing at this point.

rintcius:
@alissa yes, sorry about putting the PR before waiting on the test results. Reason was to get review feedback before my working day starts, because usually there's little time for me when USA workday starts & it slips easily to day after

alissapajer:
@rintcius yeah that makes perfect sense. And because our integration tests take so long to run locally, I think it's always fine if we make PRs without running all the integration tests for all the connectors. This isn't an ideal place to be, but it's just not practical to wait for the tests to run locally.

it/src/main/resources/tests/derivedFunctionAbs.test
```diff
@@ -1,6 +1,10 @@
 {
     "name": "derived function abs",
-    "backends": {},
+    "backends": {
+      "couchbase": "ignoreFieldOrder",
+      "marklogic_json": "ignoreFieldOrder",
+      "mimir": "pending"
```
alissapajer: So, conventionally we've included `"postgresql": "pending"` in all the integration tests, since currently none of the integration tests can pass on postgres because we haven't implemented query evaluation. 

But, we also don't run the integration tests for postgres on travis, so there is no way to enforce this convention. I'm fine merging this PR as is because I don't think we should be enforcing conventions that aren't verified by a failing build. 

I'll ask around to see if someone has them disabled for a particular reason, and we can address this separately.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -965,22 +965,23 @@ abstract class StdLibSpec extends Qspec {
...
-          y != 0 ==>
-            binary(Modulo(_, _).embed, Data.Int(x), Data.Dec(y), Data.Dec(BigDecimal(y).remainder(BigDecimal(x))))
-          x != 0 ==>
-            binary(Modulo(_, _).embed, Data.Dec(y), Data.Int(x), Data.Dec(BigDecimal(y).remainder(BigDecimal(x))))
-        }
+        // TODO analyze and optionally shortCircuit per connector
+        // "any doubles" >> prop { (x: Double, y: Double) =>
+        //   y != 0 ==>
+        //     binary(Modulo(_, _).embed, Data.Dec(x), Data.Dec(y), Data.Dec(BigDecimal(x).remainder(BigDecimal(y))))
+        // }
```
alissapajer: Could we make these pending instead of commenting them out? The benefits I see to that are (1) if they start passing we will be alerted and (2) there is a higher chance someone will see them pending in the test output and decide to fix them.

Could you include the `TODO ... ` message above in the pending message as well so we see in the test output what needs to be done?

alissapajer: I'm fine with it in a new PR.

alissapajer:
LGTM. Once @wemrysi approves as well then we can merge!

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncCorePlanner.scala
```diff
@@ -35,9 +35,10 @@ private[qscript] final class MapFuncCorePlanner[
   T[_[_]]: RecursiveT
 ](implicit
   SP: StructuralPlanner[F, FMT]
-) extends MapFuncPlanner[F, MapFuncCore[T, ?], T] {
+) extends MapFuncPlanner[F, FMT, MapFuncCore[T, ?]] {
   import expr.{emptySeq, if_, let_, some, try_}, XQuery.flwor
 
+  @SuppressWarnings(Array("org.wartremover.warts.Equals"))
```
wemrysi: Any idea why this was required, there shouldn't be any universal equality used here. Maybe its due to the `eq` method defined on `XQuery`, which isn't actually reference equality, it is just named as such to match the operation defined in `XQuery`.

sellout: Yeah, I bet that’s it.

And we should have comments on `@SuppressWarnings` in general.

alissapajer: :+1: to comments on `@SuppressWarnings`

rintcius: Yes, it's due to `eq` and `ne` (which are both annotated as a wart `Overloading` at their definition, so I think it's a valid wart)

Looking at the small diff for this file though, I'm puzzled why it's all of a sudden requiring this annotation.

wemrysi: It was likely due to the update to WartRemover pulled in by your latest merge from master.

alissapajer:
@rintcius can you merge master to handle the merge conflicts? Then this should be ready to go!

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncCorePlanner.scala
rintcius: @wemrysi but then it should require it on master too, isn't it?

wemrysi: Looks like [it does](https://github.com/quasar-analytics/quasar/blob/cffc57a32ed37676361aa9606bbfeabc964fda8a/marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncPlanner.scala#L39).

rintcius: Ah, right - I got fooled by [this](https://github.com/quasar-analytics/quasar/pull/2431/files/d439ce2e223a045c74fad29b4c1aaf390690837a..9937eda6f3a6ce5b71b53257ceea61e4c363525b#diff-a04b22e959934ef995bd3b9413862231) being a _subset_ of commits.

rintcius:
Argh - I think I'll need to resolve #2486 first to get the mongo builds stable again

alissapajer:
@rintcius yeah I've seen that mongo job pass on other PRs today, but for whatever reason it seems determined to fail on this PR :/

alissapajer:
@rintcius so your build finally passed! Do you know of any reason it might been failing so frequently on this branch? I'm happy to merge your PR now, just wanted to check with you first to see if you think this PR has added a change that is making the mongo specs timeout more frequently. Have you seen those timeouts on other branches?

alissapajer:
@rintcius and, there are merge conflicts again. :(

rintcius:
@alissapajer I have finally narrowed it down a bit. It's introduced in this branch, so good that it's not merged in yet.

The biggest amount of failures are caused by the `modulo` tests which were extended (see https://github.com/quasar-analytics/quasar/pull/2431/files#diff-a0ca8125eed850bd6f15d0791b5e216bL938 ) and the added `trunc` tests (`trunc` is defined in terms of `modulo`).

I also replaced `Skipped` with `Pending` in the mongo expr stdlib tests. This also causes some of the failures, although a lot less.

I propose to skip the above tests in this PR with a `FIXME` note.  Then I will narrow it down further when implementing the other math functions.

alissapajer:
@rintcius that sounds good. Can you skip the problematic tests just in mongo and not the other connectors? I think you can just add the problematic cases to `shortCircuit`.

rintcius:
@alissapajer yes, skipping them for mongo only

alissapajer:
@rintcius looks great, I'll get this merged as soon as the build is green!

alissapajer:
@rintcius noooo there are more merge conflicts :( I was just about to put a hold on the PR that made those conflicts until I merged this, but I didn't do that in time. The good news is that all the mongo builds are passing now.

Issue #2274, opened 2017-06-01
Create new API endpoint to allow clients to configure Metastore database location


jedesah:
This API will update the contents of the quasar config file to reflect the new backing store specified by the client and start using the new metadata store for any future API requests.

Couple questions:
- Should the endpoint return a failure response and not do anything if it can't connect to the new database or should it change the config file and just continue with the current metastore or should this be an option? Unless someone has an objection I will go with fail for a database configuration that does not point to a live database
- Should Quasar attempt to roll over all content of the current metastore to the new metastore if the new one is empty (or even if it is not...)? I am thinking it should not, that it should just start running with wtv it finds in the new one and initialize the new one to the starting state if it's empty. We can always create a `/migrate` endpoint for migrating the current configuration of a metastore to another. Perhpas it would be reasonable to make this an optional parameter of the request to avoid creating an entirely new endpoint. I'll go with this last option unless anyone has any objections.

jedesah:
@jdegoes If we successfully connect to the new metastore but fail when attempting to persist the new metastore location value to the config file should we abort the whole thing? I'm thinking yes.

jdegoes:
> Should the endpoint return a failure response and not do anything if it can't connect to the new database or should it change the config file and just continue with the current metastore or should this be an option? Unless someone has an objection I will go with fail for a database configuration that does not point to a live database

We should make it difficult as possible (ideally: impossible) to transition the configuration from a working state to a non-working state. If necessary, later and in response to customer demand, we can add a "force" mode.

> Should Quasar attempt to roll over all content of the current metastore to the new metastore if the new one is empty (or even if it is not...)? 

Combining them is conflation and reduces composability. Better to have a separate migration endpoint like you note. With that said, however, if the intention of the user is migration, and the new target has no metadata, then the moment that Quasar "switches over" to the new target, the user becomes unauthorized. This argues that switching should at least come with the option of migration, so a user isn't left in a stranded, non-functional state (see first point).

> If we successfully connect to the new metastore but fail when attempting to persist the new metastore location value to the config file should we abort the whole thing? I'm thinking yes.

I think so. See first point.

kRITZCREEK:
Is this fully implemented (can I start integrating this with the frontend)?

jedesah:
@kRITZCREEK Yes! Thx for asking. Wanted to notify you but forgot.

zeeiyerWork:
@jedesah - so where did we end up on the data-migration. 
[...]

PR #2435, opened 2017-07-06
Metastore can now be changed at runtime

jedesah:
- Add `MetastoreRef` to `CoreEff`
- Create data types for metastore initialization errors
- Add `onChange` to interface of `TaskRef` allowing client code to be notified of changes
- Rename `empty` `KeyValueStore` interpreter to `default`
- Remove a couple instance of the `Equals` wart
- Use kind-projector syntax for natural transformations a few places


wemrysi:
@jedesah Does this close or relate to a particular issue?

jedesah:
@wemrysi It's related to #2274. It does not implement everything described in the ticket though.

interface/src/main/scala/quasar/main/api/MetaStoreApi.scala
```diff
@@ -0,0 +1,57 @@
...
+import quasar.effect.AtomicRef
+import quasar.fp.free._
+import quasar.main.{initUpdateMigrate, MainTask}
+import quasar.main.metastore.metastoreTransactor
+import quasar.metastore.MetaStore
+
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+
+trait MetaStoreApi {
```
wemrysi: This trait suggests we're missing an abstraction, namely a way to read and modify the metastore configuration. I seem to remember you suggesting such an abstraction recently, I think it'd be a useful addition and tidy up the implementation as we could easily handle shutdown of the previous metastore in the concrete implementation and obviate the need for and observable `TaskRef`.

interface/src/main/scala/quasar/main/api/MountApi.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import slamdata.Predef.{ -> => _, _ }
+import quasar.contrib.pathy.APath
+import quasar.fs.mount.{Mounting, MountConfig, MountingFailure, PathMismatchFailure}
+
+import pathy.Path
+import pathy.Path._
+import scalaz._
+
+trait MountApi {
```
wemrysi: This doesn't seem all that different from `Mounting.Ops` so I'm having difficulty justifying its existence. It also looks like it is only used in tests and `moveMount` doesn't seem to be used anywhere. Maybe `mount` could be added as another variant on `Mounting.Ops` or even just defined locally in the test that uses it?

wemrysi:
This PR includes many changes unrelated to its core purpose, as well as a good amount of related, non-semantic changes (renames, moving code). While I appreciate the changes themselves (tidying up natural transformation defs, fixing TODOs, etc), it would be helpful in the future if these were in separate commits to make identifying/following the semantic changes easier.

core/src/main/scala/quasar/metastore/MetaStoreAccess.scala
```diff
@@ -27,7 +27,7 @@ import scalaz._, Scalaz._
 
 /** Operations that access the meta-store via doobie, all wrapped in ConnectionIO
   */
-abstract class MetaStoreAccess {
```
jedesah: I did not notice. Thx for pointing that out. I will add a comment and revert the change

example-quasar-config.json
```diff
@@ -1,3 +1,10 @@
 {
-    "server": { "port": 8080 }
+    "server": { "port": 8080 },
+    "metastore": {
+        "database": {
+            "h2": {
+                "file": "mem"
```
jedesah: :+1: Ya, I was worried about backwards compat, but your plan sounds good!

foundation/src/main/scala/quasar/contrib/scalaz/eitherT.scala
```diff
@@ -80,4 +81,13 @@ object eitherT extends EitherTInstances {
     def flattenLeft: EitherT[F, E, A] =
       a.run.flatMapF(_.point[F])
   }
+  implicit class TaskEitherT[E, A](a: EitherT[Task, E, A]) {
+    def onFinish(f: Option[Throwable] => Task[Unit]) =
```
jedesah: I'll have a look

interface/src/main/scala/quasar/main/api/MetaStoreApi.scala
jedesah: Ya, we had a discussion at the beginning of the month, and I finally convinced you that we needed a `MetaStore` algebra. But then when I sat down to implement it, I decided an `AtomicRef` would do for now. I didn't a foresee a few of the complications or I might have gone down the route of MetaSotre algebra from the beginning, but I figured it would be easy enough to change after so I plowed ahead. I think it could also decouple us somewhat from `ConnectionIO` which I think would be a good thing

interface/src/main/scala/quasar/main/api/MountApi.scala
jedesah: Ya, the idea is to slowly have a central place to find all the quasar top-level programmatic API. But what I really care about is having the functions on `QuasarFS` for ad-hoc testing in the `console`, I could be convinced these traits which combine into an `API` object is a not worth it's weight

jedesah:
@wemrysi I'll try and do better on this. Sometimes I just see things while I'm working on something and don't want to incur the overhead of opening a standalone PR. I think it's a fine line to walk but I definitely agree that this PR could have contained a couple fewer things that could have benefited from their own PRs.

interface/src/main/scala/quasar/main/api/MetaStoreApi.scala
jedesah: The more I think of it though, the more I wonder if it's really necessary to create an algebra for `MetaStore` though since it boils down to "get" and "set". There are certain things that could be improved about the current implementation, like the fact that it's the client's responsibility to make sure the config and the actual transactor stay coherent, but that could be fixed with just a smart constructor for the `MetaStore` case class. As for shutdown, I'm not sure an interpreter is going to be necessarily cleaner or easier. Whatever function supplies the interpreter is going to have to expose some kind of shutdown hook (probably using a `TaskRef` of some kind) to close the last transactor once quasar shuts down. The implementation might be ever so slightly more localized with an algebra and corresponding interpreter, but unclear that's such a huge benefit versus the overhead of a custom algebra to get/set a `Transactor`.

Ultimately, I'm not arguing that making an algebra is a terrible idea, I'm just applying the same skepticism you usually apply towards introducing new algebras to make sure it's actually a **better** idea as opposed to simply a different one.

@wemrysi What do you think? I'm not saying I don't want to change it, just want to discuss it before changing it so that it's not unnecessary work or even worse, deciding to revert later on because we didn't consider all the implications. Is there anything I didn't think of that makes a compelling case for a custom algebra. Usually having to authorize it in QA makes for a pretty obvious case for making a new algebra, but as far as I can tell, a `AtomicRef[MetaStore, ?]` is perfectly fine for authorizing changes to the metastore in QA. There is migration as well, which we may want to eventually support, but I don't think that changes the equation either.

wemrysi:
Yeah, I understand, even just having multiple commits in the same PR would be fine. Anyhow, sounds good, not a huge deal, just thought I'd mention it.

interface/src/main/scala/quasar/main/api/MetaStoreApi.scala
wemrysi: I think the abstraction is worth it as we want to be able to deal with these operations in a first-class fashion. `AtomicRef[MetaStore, ?]` is fine as an implementation detail, but doesn't abstract over reading and changing the metastore configuration, it isn't precise enough. It allows those operations to be performed, yes, but also many others, like obtaining a `Transactor` and running arbitrary database operations or swapping the existing transactor for one that points to a different metastore than is described in the corresponding configuration. These are pathological examples, but it'd be better if they weren't even possible. It also leaks the fact that the current metastore is implemented via doobie and uses `Task`.

As for the shutdown stuff, we already have to handle when the config changes and when the server is shutdown, so we aren't introducing any disadvantage there by abstracting, but we don't have to deal with the observable `TaskRef` which I'd consider an advantage.

jedesah:
@wemrysi This is ready for another look when you get the chance

interface/src/test/scala/quasar/main/api/ApiSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+
+import pathy.Path._
+import scalaz._, Scalaz._
+
+class ApiSpec extends quasar.Qspec {
+  "Api" >> {
+    "change metastore at runtime" >> {
+      val sampleMountPath = rootDir </> file("foo")
+      val sampleMount = MountConfig.viewConfig0(sqlB"select * from zips")
+      (for {
```
wemrysi: This could be written in terms of `Free[CoreEff, ?]` instead and then interpreted.

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -75,74 +76,67 @@ object Server {
...
-      port,
-      qConfig.staticContent,
-      qConfig.redirect,
-      foldMapNT(f) compose metastoreCtx.interp)
-
-    (serv, metastoreCtx.closeMnts *> metastoreCtx.metastore.shutdown)
+  /**
+    * Start the Quasar server and shutdown once the command line user presses "Enter"
+    */
+  def startServerUntilUserInput(quasarInter: CoreEff ~> QErrs_TaskM,
```
wemrysi: Looks like alignment went awry here.

interface/src/main/scala/quasar/main/MetaStoreLocation.scala
```diff
@@ -0,0 +1,69 @@
...
+      λ[MetaStoreLocation ~> Task] {
+        case Get => ref.read.map(_.connectionInfo)
+        case Set(conn, initialize) =>
+          val tryNewMetaStore = metastoreTransactor(conn).flatMap(m =>
+            if (initialize)
+              initUpdateMigrate(quasar.metastore.Schema.schema, m.transactor, None).as(m)
+            else m.point[MainTask])
+          (for {
+            m      <- tryNewMetaStore
+            _      <- ref.read.flatMap(_.shutdown).liftM[MainErrT]
```
jedesah: Yup, I think `_.shutdown.attempt` would be better despite the fact that we are potentially leaving the old running. Not sure there is anything better to do though

README.md
```diff
@@ -823,6 +835,44 @@ Takes a port number in the body, and attempts to restart the server on that port
...
+### GET /metastore
+
+Retrieve the connection information of the current metastore in use
+
+An example response:
+
+```json
+{
+    "h2": {
+        "file": "mem"
```
jedesah: Ya, good catch

foundation/src/main/scala/quasar/fp/TaskRef.scala
```diff
@@ -28,7 +28,7 @@ import scalaz.concurrent.Task
   * project: https://github.com/oncue/remotely
   *
   */
-sealed abstract class TaskRef[A] {
+sealed abstract class TaskRef[A] { self =>
```
jedesah: Indeed

foundation/src/main/scala/quasar/contrib/scalaz/MonadError_.scala
```diff
@@ -116,4 +117,10 @@ final class MonadError_Ops[F[_], E, A] private[scalaz] (self: F[A])(implicit F0:
 
   def handleWith(pf: PartialFunction[E, F[A]]): F[A] =
     F0.handleWith(self)(pf)
+
+  def unattempt[B](implicit ev: A <~< (E \/ B), M: Monad[F]): F[B] =
+    F0.unattempt(self.map(ev(_)))
+
+  def unattemptShow[B, E1: Show](implicit ev0: A <~< (E1 \/ B), ev1: E >~> Throwable, M: Monad[F]): F[B] =
```
jedesah: Ya, I agree that it's pretty specific

interface/src/main/scala/quasar/main/QuasarFS.scala
```diff
@@ -0,0 +1,78 @@
...
+import scalaz.stream.Process
+
+/**
+  * The Quasar Filesystem. Contains the `CoreEff` that can be used to interpret most
+  * operations. Also has convenience methods for executing common operation within `Task`.
+  * The `shutdown` task should be called once you are done with it.
+  * @param interp
+  * @param shutdown Trigger the underlying connector drivers to shutdown cleanly.
+  */
+final case class QuasarFS(interp: CoreEff ~> QErrs_TaskM, shutdown: Task[Unit]) {
```
jedesah: I like `Quasar`. I wasn't really happy with `QuasarFS` anyway.

interface/src/main/scala/quasar/main/QuasarFS.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val taskInter: CoreEff ~> Task =
+    foldMapNT(NaturalTransformation.refl[Task] :+: QErrs.toCatchable[Task]) compose interp
+
+  def getCurrentMetastore: Task[DbConnectionConfig] =
```
jedesah: Ya, I think it's a good idea to create something like this, perhaps call it `QuasarApi` or maybe just `Api`. I would still like `Quasar` to have all the operations as `Task`s so that I can `sbt console` in the interface project and do ad hoc testing that way instead of using `curl`. Might also be appreciated by eventual users of the programmatic API, but I think that use case is less sensitive to one extra operation so they would probably be content with the `Api` `trait`.

interface/src/main/scala/quasar/main/package.scala
```diff
@@ -362,6 +374,149 @@ package object main {
...
+          (liftMT[ConnectionIO, MainErrT] compose taskToConnectionIO) :+:
+            liftMT[ConnectionIO, MainErrT]                              :+:
+            QErrsCnxIO.qErrsToMainErrT[ConnectionIO])
+
+      Hoist[MainErrT].hoist(transactor.trans) compose f
+    }
+  }
+
+  /** Initialize the Quasar FileSystem assuming all defaults */
+  val initializeFS: MainTask[QuasarFS] = initializeFS_(None)
```
jedesah: I think that's a perfectly reasonable idea.

interface/src/test/scala/quasar/main/api/ApiSpec.scala
jedesah: It could be. But assuming we consider that `Quasar` has value in having operations as `Task` which may or may not be the case depending on the outcome of our above discussion, then is there any benefit to doing that here instead of taking advantage of the `Task` versions?

web/src/main/scala/quasar/server/Server.scala
jedesah: I'm always fighting IntelliJ on this... 😞 Thx for pointing it out.

jedesah:
@wemrysi Think I addressed everything that isn't still under discussion

interface/src/main/scala/quasar/main/QuasarFS.scala
wemrysi: If we define the above, then you'd still be able to easily experiment in the console via something like
```
val qt = Quasar.toTask[CoreEff]
qt(qt.mounting.ls(rootDir)).unsafePerformSync
```
and you wouldn't have to redefine all the methods.

wemrysi:
@jedesah Thanks, updates look good. I also responded in the `Quasar/QuasarCore` discussion.

interface/src/main/scala/quasar/main/QuasarFS.scala
jedesah: I don't think that's much more convenient than `ManageFile.ls[CoreEff](rootDir).foldMap(taskInter).unsafePerformSync`. I don't perceive the duplication as being a big problem at this level in the codebase. If you really feel strongly about not having these convenience functions, then I'll just delete them since I'd rather get this PR in sooner rather than later and I can re-open a separate PR down the road and we can discuss approaches to offering a more convenient programmatic API since it's fairly orthogonal to the objective of this PR.

wemrysi: Yeah, I wasn't asserting that it was more convenient, just on par with what you had but with the benefit of being complete without additional boilerplate.

Anyhow, I think dealing with this in a separate PR sounds good 👍

jedesah:
Hmm... I didn't really originally intend to push this last commit to this PR but since it depends on it, I guess that's okay...

wemrysi:
@jedesah I'm fine with rewriting to remove it from this branch as we were just about ready to merge.

jedesah:
Well, I need to figure out why it's triggering the Spark integration tests to fail (which was before persistence), so maybe not that close to merge. But if you'd rather not review more of my code, I can perform the rewriting :smile: so that I can create a new PR and bug someone else for a review on that one.

wemrysi:
I don't mind reviewing, I just didn't want this to be delayed unnecessarily.

jedesah:
@wemrysi The build is green now and ready for your final review.

The persistence stuff is in there, otherwise, just some stuff related to the config file format that was causing some failures as well as some tests which were not properly shutting down `Quasar` , leading to the Spark failure.

wemrysi:
@jedesah Are you still awaiting a review from @drostron for this?

jedesah:
@wemrysi @drostron's probably pretty busy with other stuff or else he probably would have reviewed it by now. I think it's good to merge. I think you did a pretty thorough review 😄

Issue #2340, opened 2017-06-16
Add metastore resilience for corrupt views


drostron:
Recently we've seen a few SQL² pretty printer bugs cause the entire metastore to become corrupt despite most mounts and views being unaffected.  Ideally we'd issue a warning and skip mounting the corrupt view rather than giving up on the entire metastore.

zeeiyerWork:
Metastore corruption related issues -> 
- https://github.com/quasar-analytics/quasar/issues/1910  
- https://github.com/quasar-analytics/quasar/issues/2294

PR #2440, opened 2017-07-07
Metastore resilience for invalid view mounts

drostron:
Closes: #2340


core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -39,13 +39,13 @@ object Mounting {
     * mounts) to be implemented more efficiently.
     */
   final case class HavingPrefix(dir: ADir)
-    extends Mounting[Map[APath, MountType]]
+    extends Mounting[Map[APath, MountingError \/ MountType]]
 
   final case class LookupType(path: APath)
-    extends Mounting[Option[MountType]]
+      extends Mounting[Option[MountingError \/ MountType]]
```
drostron: This and similar signatures are as discussed the other day. It satisfied the need but ended up being a bit unwieldy.

web/src/main/scala/quasar/api/services/metadata.scala
```diff
@@ -68,34 +62,65 @@ object metadata {
...
+
+  object InvalidMountNode {
+    implicit val invalidMountNodeOrder: Order[InvalidMountNode] =
+      Order.orderBy(n => (n.name, n.`type`, n.error))
+
+    implicit val invalidMountNodeEncodeJson: EncodeJson[InvalidMountNode] =
+      EncodeJson { case InvalidMountNode(name, tpe, error) =>
+        ("name" := name)                    ->:
+        ("type" :=? tpe)                    ->?:
+        ("mount" := Json("error" := error)) ->:
```
wemrysi: We should probably convert the error `String` into an `ApiError` so it serializes consistently with our other errors.

web/src/main/scala/quasar/api/ToApiError.scala
```diff
@@ -157,6 +157,11 @@ sealed abstract class ToApiErrorInstances extends ToApiErrorInstances0 {
           BadRequest withReason "Invalid mount configuration.",
           "reasons" := rsns)
 
+      case InvalidMount(_, e) =>
+        apiError(
+          BadRequest withReason "Invalid mount.",
```
wemrysi: I'm not sure `BadRequest` is the appropriate response here, I'd almost argue for a `5XX` code as this error only arises when reading mounts, so the requestor may not have had anything to do with constructing it. If we were validating mounts when created then this would definitely be a `500` as it would mean a bug or a change in parsing or somesuch was the culprit, but since we aren't its a bit of an unfortunate situation as the server accepted the bogus mount and only now notifies the user about it.

connector/src/main/scala/quasar/connector/EnvironmentError.scala
```diff
@@ -60,6 +60,13 @@ object EnvironmentError {
         s"Unsupported $name version: $version"
     }
 
+  implicit val equal: Equal[EnvironmentError] = Equal.equal {
+    case (ConnectionFailed(a), ConnectionFailed(b))           => a.equals(b)
```
drostron: Interested in an improved approach here.

drostron: Just discussed with @sellout and this seems reasonable. I will however stop avoiding the wart and add a corresponding comment.

web/src/main/scala/quasar/api/services/metadata.scala
```diff
@@ -70,9 +70,9 @@ object metadata {
 
     implicit val invalidMountNodeEncodeJson: EncodeJson[InvalidMountNode] =
       EncodeJson { case InvalidMountNode(name, tpe, error) =>
-        ("name" := name)                    ->:
-        ("type" :=? tpe)                    ->?:
-        ("mount" := Json("error" := error)) ->:
+        ("name" := name)                                           ->:
+        ("type" :=? tpe)                                           ->?:
+        ("mount" := ApiError.fromMsg_(InternalServerError, error)) ->:
```
drostron: @wemrysi does this follow your suggestion?

core/src/main/scala/quasar/fs/mount/module/Module.scala
```diff
@@ -162,7 +162,9 @@ object Module {
           val iArgs = args.map{ case (key, value) => (CIName(key), value)}
           val currentDir = fileParent(file)
           (for {
-            moduleConfig <- mount.lookupModuleConfig(currentDir).toRight(notFoundError)
+            moduleConfig <- EitherT(mount.lookupModuleConfig(currentDir)
+                              .leftMap(e => semErrors(SemanticError.genericError(e.shows).wrapNel))
```
drostron: Agreed and open to suggestions if you have any.

drostron:
@wemrysi suggestions appreciated — ready for another look.

web/src/main/scala/quasar/api/services/metadata.scala
wemrysi: Aye, though we probably want to keep the `Json("error" := ApiError.fromMsg_(...))` as I didn't realize that wasn't part of the JSON encoding for `ApiError`. We should rectify that in another PR.

drostron:
@wemrysi Ready

Issue #2418, opened 2017-07-03
Confusing relative import resolution within SQL^2 modules


natefaubion:
I would like to get some clarification about how relative imports should resolve in the following cases within a module mounted at `/foo`:

```sql
-- 1) Relative import
IMPORT `./bar`;

-- 2) Relative import in a function
CREATE FUNCTION bar(:n)
BEGIN
  SELECT * FROM `./baz` WHERE n < :n
END

-- 3) Relative import passed in via `invoke`
CREATE FUNCTION baz(:path, :n)
BEGIN
  SELECT * FROM :path WHERE n < :n
END
```
Where (3) is called as 
```
/invoke/fs/foo/baz?path=`./quux`&n=42
```

For (1), it currently resolves _inside_ the module, so `/foo/bar`, which is really strange since I'm also defining a function at `/foo/bar`. Should it instead resolve to `/bar`, ie. a sibling module?

For (2), I'm not actually sure what it does currently, but since a function has a path on the file system, my assumption would be that it should resolve to `/foo/baz`, and in any case, (2) and (3) should resolve identically I'd think.

Some other questions
  * Can I mount a module at the same path as a function?
  * Can I create a file at the same path as a module?
  * Can I create a file within a module?
  * Are modules and files in completely different namespaces that may or may not overlap?

natefaubion:
Or rather, are modules, functions, and files all in separate namespaces? Like could I have a function, module, or path at any given point on the filesystem?

zeeiyerWork:
@jedesah and @natefaubion -> Can this issue be closed since @jedesah is referencing a PR that seems to fix a related issue?

jedesah:
It's no longer a blocker for the frontend, but this ticket is still valid as the spec needs to be clarified regarding this issue.

jdegoes:
@jedesah In my opinion, all relative links inside a module should resolve relative to the location of the module. All relative links inside a query (including imports) resolve relative to the location of the query.

@natefaubion Thoughts?

jedesah:
> all relative links inside a module should resolve relative to the location of the module

[...]

PR #2441, opened 2017-07-07
Queries and modules now support imports with `/../`

### LINKED ISSUE #2430: Support `/../` in import paths

jedesah:
- Fixes #2430 
- Add a few more tests around importing

### DIFF
--- a/connector/src/test/scala/quasar/qscript/QScriptHelpers.scala
+++ b/connector/src/test/scala/quasar/qscript/QScriptHelpers.scala
@@ -142,7 +142,7 @@ trait QScriptHelpers extends CompilerHelpers with TTypes[Fix] {
     Free.roll(MapFuncsCore.Not(value))
 
   def lpRead(path: String): Fix[LP] =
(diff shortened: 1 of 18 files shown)


jsantos17:
@jedesah would you mind giving me some background on this? Is there an issue with more information I can check? I don't understand why we're implementing this feature.

jedesah:
@jsantos17 It's somewhat related to #2418 and the fact that `/../` is required to reference a sibling module right now although that is subject to change depending on the outcome of the discussion on that issue.

But even in isolation, I don't see a good reason to prohibit this kind of path in an import statement. It's sort of an oversight on my part to not have included it in the initial implementation.

jedesah:
@jsantos17 I think this is blocking @natefaubion. If you could have a look this evening or perhaps first thing tmr morning.

jsantos17:
Is this ready to merge @jedesah or are you planning to add more commits?

jedesah:
@jsantos17 Btw, I'm curious why you asked me if there were more commits coming? Did I miss a new step in the PR process?

PR #2442, opened 2017-07-07
More VFS Cleanup

### LINKED ISSUE #2253: Fix hacky evaluatePlan implementation
Similar to `ManageFileModule`, the stuff in `QueryFileModule` is super-broken and won't work at all for things that don't fit in memory.  Should be comparatively easy to fix.

### LINKED ISSUE #2421: Refine and polish new VFS
This includes a couple things:

- Making the VFS semantics more closely match quasar (notably w.r.t. move error cases)
- Refining some of the semantics that already exist (e.g. memory and backpressure properties)
- Clean up and a bit of hardening of the internal implementation

djspiewak:
Brings things almost entirely into alignment with the quasar FS semantics.

- Lots of new integration tests enabled
- Error reporting works now (btw: reading a non-existent file should be an error, *not* an empty dataset!!)
- Manage, read and write file is all in place
  + One test is "pendinged" still because it's really, really slow
- Executing queries (as opposed to evaluating)
- Move error semantics are now in alignment with what quasar wants, rather than what precog wants
- Other improvements that I've forgotten about

Fixes #2421 
Fixes #2253


mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -297,66 +298,152 @@ object Mimir extends BackendModule with Logging {
...
+      val t = for {
+        pager <- Task.delay(Option(map.get(h)).get)
+        check <- Task.delay(map.remove(h, pager))
+        _ <- if (check) pager.close else Task.now(())
+      } yield ()
+
+      t.liftM[MT].liftM[ConfiguredT]
+    }
+
+    def explain(repr: Repr): Backend[String] = "🤹".point[Backend]
```
alissapajer: I can only assume that is a lovely emoji.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -297,66 +298,152 @@ object Mimir extends BackendModule with Logging {
...
+        precog <- cake[Backend]
+        handle <- Task.delay(ReadHandle(file, cur.getAndIncrement())).liftM[MT].liftB
+
+        target = precog.Table.constString(Set(posixCodec.printPath(file)))
+
+        // apparently read on a non-existent file is equivalent to reading the empty file??!!
+        eitherTable <- precog.Table.load(target, JType.JUniverseT).mapT(_.toTask).run.liftM[MT].liftB
+        table = eitherTable.fold(_ => precog.Table.empty, table => table)
+
+        limited = if (offset.value == 0L && !limit.isDefined)
```
alissapajer: Can we use scalaz `===` here instead.

alissapajer: And in any other places we have `==`.

mimir/src/main/scala/quasar/mimir/Precog.scala
```diff
@@ -142,6 +147,50 @@ final class Precog private (dataDir0: File) extends VFSColumnarTableModule with
   sealed trait TableCompanion extends VFSColumnarTableCompanion
   object Table extends TableCompanion
 
+  final class TablePager private (
```
alissapajer: Does `TablePager` and its companion need to be in `Precog` or could we put it in a separate file?

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -297,66 +298,152 @@ object Mimir extends BackendModule with Logging {
...
+            if (!json.isEmpty)
+              q.enqueue1(json)
+            else
+              Task.now(())
+          } else {
+            Task.now(())
+          }
+        }
+
+        populatorWithTermination = populator >> q.enqueue1(Vector.empty)
```
alissapajer: Instead of using `Vector.empty` as the termination signal, could we use just an object `StopPop` and then have `StopPop \/ Vector[JValue]` for the values in the queue. Then maybe we could avoid the `isEmpty` checking up on line 315 and 317 and just enqueue empty vectors when we get them. Especially since we produce empty vectors naturally, one seems like not the best signal.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -297,66 +298,152 @@ object Mimir extends BackendModule with Logging {
...
       t.liftM[MT].liftB
     }
 
-    def close(h: ResultHandle): Configured[Unit] =
-      (Task delay { map.remove(h); () }).liftM[MT].liftM[ConfiguredT]
+    def close(h: ResultHandle): Configured[Unit] = {
+      val t = for {
+        pager <- Task.delay(Option(map.get(h)).get)
+        check <- Task.delay(map.remove(h, pager))
+        _ <- if (check) pager.close else Task.now(())
```
alissapajer: Wanted to do this sort of thing has always struck me as something that `ConcurrentHashMap` should provide, but they seem to have omitted any sort of callback functionality.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -297,66 +298,152 @@ object Mimir extends BackendModule with Logging {
...
+        pager <- precog.TablePager(projected).liftM[MT].liftB
+        _ <- Task.delay(map.put(handle, pager)).liftM[MT].liftB
+      } yield handle
+    }
+
+    def read(h: ReadHandle): Backend[Vector[Data]] = {
+      for {
+        maybePager <- Task.delay(Option(map.get(h))).liftM[MT].liftB
+
+        pager <- maybePager match {
```
alissapajer: I've come to strongly prefer `fold` over `match` on `Option`. What do you think?

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -297,66 +298,152 @@ object Mimir extends BackendModule with Logging {
...
+
+      t.liftM[MT].liftM[ConfiguredT]
+    }
   }
 
   object WriteFileModule extends WriteFileModule {
     import WriteFile._
 
-    private val QueueLimit = 100
+    private val QueueLimit = 1
```
alissapajer: Could you add a comment about why a size of `1`?

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -432,23 +528,90 @@ object Mimir extends BackendModule with Logging {
 
     // TODO directory moving and varying semantics
     def move(scenario: MoveScenario, semantics: MoveSemantics): Backend[Unit] = {
+      val semantics2 =
+        semantics match {
+          case MoveSemantics.Overwrite =>
+            quasar.yggdrasil.vfs.MoveSemantics.Overwrite
```
alissapajer: Why do we need our own `MoveSemantics` in yggdrasil? Why can't we just reuse the existing one?

mimir/src/main/scala/quasar/mimir/Precog.scala
```diff
@@ -142,6 +147,50 @@ final class Precog private (dataDir0: File) extends VFSColumnarTableModule with
...
+    {
+      val driver = slices foreachRec { slice =>
+        for {
+          flag <- Task.delay(running.get())
+
+          _ <- if (flag && !slice.isEmpty) {
+            val json = slice.toJsonElements.map(JValue.toData)
+
+            if (json.isEmpty)
+              Task.now(())
```
alissapajer: Similar comment to previously. We are using `Vector.empty` again to signal completion when empty vectors are being computed naturally. Seems like an easy way to prematurely (accidentally) kill the process.

mimir/src/main/scala/quasar/mimir/Precog.scala
```diff
@@ -142,6 +147,50 @@ final class Precog private (dataDir0: File) extends VFSColumnarTableModule with
...
+      queue: Queue[Task, Vector[Data]]) {
+
+    private val running = new AtomicBoolean(true)
+
+    {
+      val driver = slices foreachRec { slice =>
+        for {
+          flag <- Task.delay(running.get())
+
+          _ <- if (flag && !slice.isEmpty) {
```
alissapajer: Is it possible that we will call `close` (setting `running` to `false`) after we have entered the body of `driver` here, meaning that we would still `enqueue`. Then that object will be dequeued when we call `more`. That might be fine, but just wanted to point out that we don't have precise control over when we stop enqueueing.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/MoveSemantics.scala
```diff
@@ -0,0 +1,38 @@
...
+package quasar.yggdrasil.vfs
+
+sealed trait MoveSemantics extends Product with Serializable {
+  import MoveSemantics._
+
+  /**
+   * Given an existence check on the target, return whether or not
+   * the operation is valid.
+   */
+  final def apply(check: Boolean): Boolean = this match {
```
alissapajer: Can we just add this functionality to `ManageFile.MoveSemantics` and not have to add our own `MoveSemantics` in yggdrasil?

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -211,8 +211,15 @@ object FreeVFS {
     } yield back
   }
 
-  def exists[F[_]: Monad](path: AFile): StateT[F, VFS, Boolean] =
-    StateTContrib.get[F, VFS].map(_.paths.contains(path))
+  def exists[F[_]: Monad](path: APath): StateT[F, VFS, Boolean] = {
+    Path.refineType(path) match {
```
alissapajer: Consider `fold` on `Either` rather than using `match`?

alissapajer:
@djspiewak so I'm a bit concerned about how slowly some of the newly-enable file system tests are running. Do you have a sense of what's making them slow and how we can improve their performance?

alissapajer:
@djspiewak since this enables new file system functionality for mimir, I think this should be "feature" version instead of "revision". Thoughts?

alissapajer:
@djspiewak I'm actually confused why the mimir build got so much slower for this branch. `sbt it/testOnly *FilesSpec*` on this branch locally completes in a matter of seconds, so did the full stack integration tests get slower? We should "timeout" any that are timing out. I have them running locally to see if any fall under this category. But it seems that none of the new tests you have enabled are slow.

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak: It depends on `Table`, so it needs to be in the cake.  We could put it in a separate trait which is caked in though.

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak: If that happens, we will end up enqueueing a chunk into a queue which has already been removed from the map.  This enqueue will be successful, but the map entry is already flushed and so the object will eventually be GC'd.

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak: The reason that empty vector was chosen as the termination signal is because quasar *already* treats it as termination.  When the `More` member of the `ReadFile` algebra produces `Vector.empty`, it is assumed that the stream is terminated and no more data will be requested.  We could have also used `Option[Vector[Data]]` as the element type, but it seemed better just to unify the signals.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: Yeah, callbacks didn't become exciting until a long time after `ConcurrentHashMap` was designed, unfortunately.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: I think there's a good argument for it, but I hate the way that it's a very different style of signature from nearly every other fold (it's more like `List`'s fold than `Either`'s), and the lack of labeling on the cases sometimes makes it a little difficult to see what's going on.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: Because yggdrasil doesn't have a dependency on filesystem.  I could add that, but there are other advantages to having our own variant of this.  Notably, it was slightly easier to abstract away some of the implementation of the functionality.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/MoveSemantics.scala
djspiewak: Potentially, though it seems odd to add a function that only yggdrasil would care about.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
djspiewak: When it's multi-line, I generally shy away from `fold` on `Either`.  But in this case, I think it probably would be slightly cleaner.

djspiewak:
@alissapajer 

> since this enables new file system functionality for mimir, I think this should be "feature" version instead of "revision". Thoughts?

I'm on the fence about that.  I could see it either way.  I selected revision mostly because it's more than existing functionality is being un-errored rather than new stuff being "added" per se.  Maybe that's a meaningless distinction.  :-)

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: For those who don't have it:

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: Looks like this is the only place in the diff where we're using `==`.  We should change over other uses as we touch them, ultimately allowing us to start enabling WartRemover on these projects.

alissapajer: Yeah, definitely. :+1:

mimir/src/main/scala/quasar/mimir/Precog.scala
alissapajer: Yeah, maybe it's better to just keep it here than add more cakey bits.

alissapajer: Up to you on this one, I'm fine either way.

mimir/src/main/scala/quasar/mimir/Mimir.scala
alissapajer: [`Either#fold`](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Either.scala#L54) and [`Option#fold`](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/std/Option.scala#L182) have the same style of signature in scalaz, at least from what I can see.

The Scala standard lib fold does have a different style of signature though.

djspiewak: The `Option#fold` in scalaz is almost worthless, because you almost always need to import `scalaz.std.option._` in order to get the typeclass instances, but then if you want fold on option, the name of it is literally `fold` without any namespacing, which becomes super-hard to follow when reading the code later (`fold`? on what?).  So when I fold on `Option`, I use the stdlib one, but the stdlib one has the weird signature.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: So upon further investigation, `ManageFile` is actually in connector, not in filesystem, which is not a dependency I want to add to yggdrasil.  Perhaps at some future date we should move `MoveSemantics` into common or something like that?

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak: I split it out.

mimir/src/main/scala/quasar/mimir/Mimir.scala
alissapajer: Not sure if I follow... when you use the stdlib `fold` on `Option` what namespacing do you have then? You're still calling `.fold(...)(...)` on some `Option`.

djspiewak: Yes.  The only way to call the scalaz fold is to just call `fold` (no dot), which seems unacceptably generic for a very type-specific signature.  Basically, I wish scalaz had given it a different name.  This is almost as bad as Haskell's standard library using the name `map` for a function specifically of type `(a -> b) -> [a] -> [b]`

alissapajer:
@djspiewak regarding "revision" versus "feature" version label, if some functions are un-errored then they were effectively non-existent before. So even though the signatures were there and callable, you couldn't argue that the feature was actually implemented, only that the signatures were flushed out.

mimir/src/main/scala/quasar/mimir/Mimir.scala
alissapajer: Well, in scalaz `fold` is an alias for `cata`, though that name is arguably no better. But in any case with scalaz you can do
```scala
val foo: Option[Int] = 5.some
foo.cata(_ + 1, 10)
```
Is there something you don't like with this? Using stdlib you'd be doing the equivalent just with the different signature style
```scala
foo.fold(10)(_ + 1)
```

mimir/src/main/scala/quasar/mimir/Precog.scala
alissapajer: So, then, I think using `Vector.empty` as the termination signal in quasar isn't ideal, given how easy it could be for a backend to accidentally provide `Vector.empty`. Whereas if we used a disjunction or even `Option` like you suggest, we are forcing the backend to consciously provide the termination signal. 

I can see how it could be easier to unify the signals, as you put it, so maybe the question is, should we change the termination signal in quasar? That's maybe out of scope of this PR, but for termination signals like this I think using a different _type_ is better than using a semi-arbitrary _value_ of a type. I mean ultimately you have a `\/` or something that includes the termination type, but the termination type is still a separate type in its own right.

mimir/src/main/scala/quasar/mimir/Mimir.scala
alissapajer: Right, yeah I agree we don't want `connector` as a dependency to `yggdrasil`. I just don't like duplicating things like this because then down the road someone will ultimately be factoring it out into a commonly-accessible place. So maybe we could just put it in the common place now? I think the `common` project would be that place.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: Right, so `cata` is reasonable.  The stdlib `fold` throws me because the signature is completely wrong.  I mean, I use it whenever I found myself typing a `.map` followed by a `.getOrElse`, but I don't use it to replace matches.  And I'm not convinced that `cata` is really that much of a win over `match` when you have anything more than trivial implementations for each case.

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak: Well, here's an alternative way of looking at it.  The signature of `more` really could be the following:

```scala
def more(h: Handle): Option[NonEmptyList[Data]]
```

That has a termination signal, and it reflects the constraint that we don't want chunks to be empty (i.e. those should be explicitly filtered out by the backend, if they exist).  But `Option[NonEmptyList[A]]` is isomorphic to `List[A]`, so… You see where this is going.

The more I think about it, the more I'm actually ok with `Vector.empty` being the termination signal.  The only downside to it is that it isn't explicitly represented in the types, but the types that *could* represent it would be isomorphic to this anyway.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: Alright, will move to common.

mimir/src/main/scala/quasar/mimir/Precog.scala
alissapajer: Yeah so I agree that the alternate representation is isomorphic to how we are using `Vector.empty`. 

I'm arguing that using `Vector.empty` is a more bug-prone approach than using a unique type because we need to explicitly check for the `Vector.empty` case and do something else when we don't actually want to terminate. And so it would be really easy for someone to simply forget to check `myVector.isEmpty` and accidentally propagate an empty vector. For that reason,  `Option[NonEmptyList[Data]]` is also a better choice (even though it's isomorphic) because you are forced to handle the empty vector case and decide if it means "terminate" or not. I want the types to force me to handle the termination case (or what could be incorrectly mistaken as the termination case). `Vector.empty` does not do that.

djspiewak: I suppose that is true.

alissapajer: I'm not going to hold merge up on this since this is a semantic already in quasar, just wanted to make sure I was giving my arguments :)

djspiewak: Let's loop back to this at some point in the future.

sellout: Yeah, I seem to remember introducing a bug (in the REPL, perhaps?) where it just kept looping, reading 0 more results, because I didn’t handle Vector.empty correctly. I think separating it in `more` would be useful.

But also agree – not this PR.

PR #2443, opened 2017-07-07
Add timeout status to mimir integration tests that are timing out.

alissapajer:


alissapajer:
@djspiewak these three tests timing out explain why the mimir build was running so slowly on travis, but it doesn't make sense that these tests are timing out on _master_ and yet the master build for mimir isn't slow, but your PR build #2442 is slow... There must be something I'm missing here.

PR #2444, opened 2017-07-08
Invoke now supports constant plans in module function definitions

jedesah:


core/src/main/scala/quasar/fs/mount/module/Module.scala
```diff
@@ -139,7 +138,11 @@ object Module {
             Process.emitAll(data) ++ readUntilEmpty(h)
         }
 
-      Process.bracket(unsafe.invokeFunction(path, args, offset, limit))(closeHandle)(readUntilEmpty)
+      Process.await(unsafe.invokeFunction(path, args, offset, limit)) { dataOrHandle =>
```
djspiewak: This bracket is incorrectly nested, and may be preempted out.  There is no point to a `Process.bracket(handle.point)(...)`, since the resource may have already leaked in the scope of the `await`.

Instead of using structuring things like this, you should push the `bracket` to the outermost level, as it was originally written and adapt the finalizer to handle the case where no handle was opened.

PR #2446, opened 2017-07-08
Add test for content-disposition functionality on `invoke` endpoint

jedesah:
- Also remove `content-disposition` media-type extension from `Content-Type` response header as it has no meaning

### DIFF
--- a/web/src/test/scala/quasar/api/services/InvokeServiceSpec.scala
+++ b/web/src/test/scala/quasar/api/services/InvokeServiceSpec.scala
@@ -189,7 +189,6 @@ class InvokeServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s 
           val request = Request(
             uri = pathUri(functionFile.path).copy(query = Query.fromPairs("bar" -> arg)),
             headers = Headers(Accept(jsonReadableLine.mediaType.withExtensions(Map("disposition" -> disposition.value)))))
-          println(request.queryString)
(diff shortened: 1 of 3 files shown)


web/src/test/scala/quasar/api/services/InvokeServiceSpec.scala
```diff
@@ -176,6 +177,23 @@ class InvokeServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s
...
+                sampleData: Vector[Data]) =>
+          val disposition = `Content-Disposition`("attachement", Map("filename" -> "data.json"))
+          val statements = List(sampleStatement(fileName(functionFile.path).value))
+          val mounts = Map((fileParent(functionFile.path):APath) -> MountConfig.moduleConfig(statements))
+          val state = InMemState.fromFiles(Map(dataFile.path -> sampleData))
+          val arg = "`" + posixCodec.printPath(dataFile.path) + "`"
+          val request = Request(
+            uri = pathUri(functionFile.path).copy(query = Query.fromPairs("bar" -> arg)),
+            headers = Headers(Accept(jsonReadableLine.mediaType.withExtensions(Map("disposition" -> disposition.value)))))
+          println(request.queryString)
```
jedesah: Yup, thx for the catch. I should really use `debug` of some sort, that way the build will fail if I forget one in the codebase. I think I will do that now...

PR #2447, opened 2017-07-10
Plan single predicate filters using `cts:query`.

### LINKED ISSUE #2425: Plan single predicate filters using cts:query when index available
Planning a QScript filter consisting of a single indexable predicate should leverage the index when available.

jsantos17:
Fixes #2425


marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
```diff
@@ -185,6 +179,50 @@ private[qscript] final class QScriptCorePlanner[
       "Unreferenced".xs.right.point[F]
   }
 
+  object FilterPlanner {
```
jsantos17: Need to move all of this to a different file.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/QScriptCorePlanner.scala
```diff
@@ -138,42 +143,31 @@ private[qscript] final class QScriptCorePlanner[
...
-            order,
-            isStable,
-            ~x)
+    case Filter(\/-(src), f) =>
+      FilterPlanner.xqueryFilter(src, f) map (_.right)
 
-        case _ =>
-          for_(x in src) where_ p return_ ~x
-      }).right
+    case Filter(-\/(src), f) =>
```
jsantos17: @wemrysi This case is never with the queries I have tested so far. There are queries which I believe we can plan with `cts:query`, but aren't being planned that way because of the QScript generated. For example, the query 

```sql
select * from `/mxml/olympics` where gender = 'F'
```

is planned with a left shift and the predicate is then used as part of the repair function. This completely bypasses all changes in this PR. Just in case, [the QScript](https://gist.github.com/jsantos17/1fc3590ab35decf7556a34af7b3bcf64) I'm talking about. Suggestion on how we can optimize these cases with the QScript given to us? Any notes on the kind of queries planned directly as a `Filter`?

wemrysi: Sounds like you're testing with data uploaded via Quasar, all in a single file? You'll want to upload data one file per document and query on the parent directory in order to benefit from index usage.

I'm not concerned about optimizing the single file case at the moment as MarkLogic recommends storing data one document per file, so I'm assuming that is the common case. If it turns out that many users store lots of data in single files, we can revisit this. I'm not sure how useful indexes are in that scenario, though. You can configure document fragmentation with XML, not sure if you can with JSON and not sure how helpful it will be to how we deal with documents in Quasar.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,149 @@
...
+
+import scalaz._, Scalaz._
+
+private[qscript] final class FilterPlanner[
+  F[_]: Monad: QNameGenerator: PrologW: MonadPlanErr: Xcc,
+  FMT: SearchOptions,
+  T[_[_]]: BirecursiveT: ShowT
+](implicit SP: StructuralPlanner[F, FMT]) {
+
+  case class ProjectPath[A](src: A, path: ADir)
```
jsantos17: Need to move everything related to `ProjectPath` to a different file.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  private def interpretSearch[Q](s: Search[Q])(implicit Q: Recursive.Aux[Q, Query[T[EJson], ?]]): F[XQuery] =
+    Search.plan[F, Q, T[EJson], FMT](s, EJsonPlanner.plan[T[EJson], F, FMT])
+
+  private def planPredicate[T[_[_]]: RecursiveT, Q](fm: FreeMap[T])(
+    implicit Q:   Corecursive.Aux[Q, Query[T[EJson], ?]],
+             RTF: RenderTree[FreeMap[T]],
+             RTP: RenderTree[Free[PathMapFuncCore[T, ?], Hole]]
+  ): Option[Q] = {
+    println(fm.render.show)
```
jsantos17: The generation of the query goes here. Will delete the debug code once I figure out how to deal with the guards.

jsantos17:
@wemrysi This is working now. It plans `Filter`s of static paths using `cts:query` as long as the path range index exists. However, there's a caveat. Assume the olympics dataset, with each row split into its own file and loaded into Marklogic's path `/olympics` with `mlcp`. A sample file (there are ~2000 of them in the dataset):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<info>
  <discipline>Cross Country S</discipline>
  <city>Albertville</city>
  <gender>M</gender>
  <country>FIN</country>
  <year>1992</year>
  <sport>Skiing</sport>
  <type>Bronze</type>
  <event>4x10km relay</event>
</info>
```

When a path range index at `/info/city` exists I'd expect `SELECT * FROM /olympics WHERE city = "Innsbruck"` to be planned taking advantage of the index. However, this doesn't happen because the index is on `/info/city` and nested `ProjectField`s  are folded into `ProjectPath(/city)`. The root element `info` is not mentioned anywhere in the final QScript plan. I understand this is because of `ShiftedRead`. Anyway, `SELECT * FROM /olympics WHERE info.city = "Innsbruck"` is planned with the index and works as expected. The same query without a range index on `/info/city` returns the empty set.

Can we somehow recover the information related to the root element? Any other thoughts on this? We can request users use full paths in predicates if they want their indexes used, but this leads to some very unintuitive behavior with single segment path range indexes (i.e `/city`).

wemrysi:
@jsantos17 Hm, yeah I don't think requiring queries to reference the root element is an acceptable limitation.

Have you tried with JSON data? I'd guess it doesn't share this issue, but it'd be good to confirm.

A few thoughts/options on how we could proceed

1. Path indexes can actually be a [subset of valid `XPath` expressions](http://docs.marklogic.com/8.0/guide/admin/range_index#id_40666) so maybe we can `cts:path-range-query("/*/city", "=", "Innsbruck")`. However I'm not sure if that requires an index created on the literal path `/*/city` or if it would also use an index created on `/info/city`.

2. Attempt the `/*/city` index first and, if that doesn't exist, fallback to a `cts:element-range-query("city", "=", "Innsbruck")` and post-filter the results. So we'd look for any predicates in the filter that indexes exist for, add them to the query but keep the `Filter` node so that it can remove any false-positives from the results of the `cts:element-range-query`. This is actually what MarkLogic does for most indexed searches. We could also add all the paths referenced in the filter as predicates to leverage the universal index, so the above query, using an element index, would become `cts:search(fn:doc[./*/city], cts:element-range-query("city", "=", "Innsbruck"))` which would ensure that all the result documents contained a top-level `city` property and had some `city` property with a value equal to `Innsbruck`, we'd have to post-filter to ensure they're the _same_ `city` property.

3. If we know an index (either path or element) exists for a predicate, we _might_ be able to just add it to the search as an `XPath` expression. This would require lots of edge case testing as it could turn out to be brittle. If an `XPath` filter in a search encounters a type it doesn't expect, it throws, whereas when we reference the index directly we know this won't happen. So in this case, if you could find a path index for `/*/city` or an element index on `city` you could try `cts:search(fn:doc()[city = "Innsbruck"])` which should use either index and post-filter.

I think (2) is probably the most predictable scenario in the presence of heterogeneous values for a property and it would increase the number of scenarios where an index would be used.

Finally, before attempting any index planning, we should check the `cts.Query` to see if it contains a `Query.Document` node and fallback to the non-indexed plan if so.

jsantos17:
@wemrysi I have confirmed it works fine with JSON. The problem only happens with XML. Sadly, using `/*/city` with `cts:path-range-query` looks explicitly for the literal index, other indexes like `/info/city` don't match. I'm going to try (2) and see how it goes. 🤞

marklogic/src/test/scala/quasar/physical/marklogic/qscript/ProjectPathSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+final class ProjectPathSpec extends quasar.Qspec {
+  def projectField[T[_[_]]: BirecursiveT](src: FreeMap[T], str: String): FreeMap[T] =
+    Free.roll(MFC(ProjectField(src, StrLit(str))))
+
+  def projectPath[T[_[_]]: BirecursiveT](src: FreePathMap[T], path: ADir): FreePathMap[T] =
+    Free.roll(Inject[ProjectPath, PathMapFunc[T, ?]].inj(ProjectPath(src, path)))
+
+  def hole: FreePathMap[Fix] = Free.point[PathMapFunc[Fix, ?], Hole](SrcHole)
+
+  "foldProjectField" should {
```
jsantos17: @wemrysi I have been trouble thinking of possible edge cases to test here. Any suggestions? I already have one (not yet pushed) for unrelated nodes inside a `ProjectField` nesting. Any other suggestions?

wemrysi: How about a case for a single project?

jsantos17:
Closed in favor of #2577

Issue #2438, opened 2017-07-06
implement mimir `TimeLib`


alissapajer:
This involves porting in `TimeLibModule` from precog and then updating it to use `java.time` library.

PR #2456, opened 2017-07-11
Use `ZonedDateTime` in precog along with misc blueeyes cleanup

### LINKED ISSUE #2166: Audit time stuff
The old joda-time code was incorrectly ported.  Among other things, `org.joda.time.DateTime` is more analogous to `java.time.ZonedDateTime` than it is to `java.time.LocalDateTime` ([source](http://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html)).  It is very likely that some of the time-related code is buggy, and perhaps in ways which will only show up when we put the code on other servers.

We need to be *very* careful about this!  Attention @damonLL and others!  Just putting this on your radar for when FBE starts doing real things.  We'll loop back to this ticket later.

alissapajer:
Closes #2166 

This is the prep work for #2438.

`LocalDateTime` is still used in the security/account stuff in precog. We are ripping that out, so we don't even need to figure out what the correct date time representation is for that functionality.

### DIFF
--- a/blueeyes/src/main/scala/quasar/precog/common/Codec.scala
+++ b/blueeyes/src/main/scala/quasar/precog/common/Codec.scala
@@ -23,7 +23,11 @@ import quasar.precog.util.{ BitSetUtil, ByteBufferMonad, ByteBufferPool }
 import java.nio.{ByteBuffer, CharBuffer}
 import java.nio.charset.{ CharsetEncoder, CoderResult }
 import java.math.{ BigDecimal => BigDec }
-import java.time.LocalDateTime
+import java.time.{LocalDateTime, ZonedDateTime}
+import java.time.format.DateTimeFormatter
+
+import scala.annotation.tailrec
+import scala.specialized
 
 import scalaz._
 
@@ -34,7 +38,7 @@ import scalaz._
   * given to a `writeMore` method so it can finish the writing. It may take
   * several calls to `writeMore` before it all is finally written.
   */
-trait Codec[@spec(Boolean, Long, Double) A] { self =>
+trait Codec[@specialized(Boolean, Long, Double) A] { self =>
   type S
 
   /** Returns the exact encoded size of `a`. */
@@ -153,7 +157,7 @@ object Codec {
 
   implicit def IndexedSeqCodec[A](implicit elemCodec: Codec[A]) = new IndexedSeqCodec(elemCodec)
 
-  implicit def arrayCodec[@spec(Boolean, Long, Double) A: Codec: CTag]: Codec[Array[A]] = ArrayCodec(Codec[A])
+  implicit def arrayCodec[@specialized(Boolean, Long, Double) A: Codec: CTag]: Codec[Array[A]] = ArrayCodec(Codec[A])
 
   /**
     * A utility method for getting the encoded version of `a` as an array of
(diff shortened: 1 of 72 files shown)


alissapajer:
@djspiewak the build is failing it where I expected it to, exposing what I believe is a bug in the string codec in old precog. Let's chat about the best way around this, since I don't think tracking down that bug is a good use of time since we plan to replace all the codec stuff. This may or may not be the right time to make the codec replacement, we can chat.

blueeyes/src/main/scala/quasar/precog/common/Codec.scala
```diff
@@ -756,21 +768,21 @@ object Codec {
     }
   }
 
-  case class SparseRawBitSetCodec(size: Int) extends Codec[RawBitSet] {
+  case class SparseRawBitSetCodec(size: Int) extends Codec[Array[Int]] {
```
djspiewak: Wait… we're getting rid of `RawBitSet`?  Maybe this is explained later in the diff.  Seems a bit scary, eliminating a nominal type in the presence of a typeclass like `Codec`, not to mention the semantics which may be affected.

**EDIT:** Oh, `RawBitSet` was an alias for `Array[Int]`.  Dear god that's such a terrible concept…  We should remember to loop back and fix this to refer to a newtype sometime down the line (maybe around the time we switch to scodec).

mimir/src/main/scala/quasar/mimir/ReductionLib.scala
```diff
@@ -110,12 +109,13 @@ trait ReductionLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
         def reduce(schema: CSchema, range: Range): Result = {
           val maxs = schema.columns(JDateT) map {
             case col: DateColumn =>
-              var zmax: LocalDateTime = dateTime.minimum
+	      // FIXME `ZonedDateTime` doesn't actually have a minimum value
+              var zmin: ZonedDateTime = ZonedDateTime.of(LocalDateTime.MIN, ZoneId.of("UTC"))
```
djspiewak: Note that `UTC+12` (which, as a matter of interest, I believe corresponds to Guam) with `LocalDateTime.MIN` would be the true "minimum" `ZonedDateTime` value.  `UTC` is the mid-point of the timezone spectrum.

blueeyes/src/main/scala/quasar/precog/common/Codec.scala
alissapajer: Yes it was just an alias.

alissapajer:
@djspiewak the build was failing for legit reasons and so this was not ready to be merged. See my comment above about the string codec. I will PR a quick fix which disables the failing test so we don't have master failing. Then we'll figure out what to do about it.

mimir/src/main/scala/quasar/mimir/ReductionLib.scala
alissapajer: Ah yeah you might be right, when I try to make anything smaller I get `Invalid value for Year`.

djspiewak:
@alissapajer RATS.  Sorry.  I scanned quickly through the build failures and it seemed fine, but I didn't read all of the logs.

This is what comes of having a flaky Travis. :-(  We just stop trusting it at all.

mimir/src/main/scala/quasar/mimir/ReductionLib.scala
djspiewak: Oh interesting!  I wonder if `ZonedDateTime` just doesn't bother handling that case.

alissapajer: No sorry for the confusion, I can create a `ZonedDateTime` at `UTC+12`. I just cannot get it smaller than that, which is confirming what you originally stated.

PR #2458, opened 2017-07-12
`metadata` endpoint now supports `offset` and `limit`

### LINKED ISSUE #2315: Extend the `GET /metadata/fs/[path]` endpoint with `offset` and `limit` parameters
For slamdata/slamdata#1848, and just general sensible-web-api-ness we could use the ability to paginate the FS data for cases when there are large numbers of items present.

I'd suggest something that follows the same scheme as the `GET /data/fs/[path]` endpoint, adding optional `?offset=[offset]&limit=[limit]` query parameters.

jedesah:
Fixes #2315

### DIFF
--- a/web/src/main/scala/quasar/api/services/metadata.scala
+++ b/web/src/main/scala/quasar/api/services/metadata.scala
@@ -20,6 +20,8 @@ import slamdata.Predef.{ -> => _, _ }
 import quasar.fp.ski._
 import quasar.api._
 import quasar.contrib.pathy._
+import quasar.contrib.std._
+import quasar.fp.numeric._
 import quasar.fs._
 import quasar.fs.mount._, MountConfig.moduleConfig
 import quasar.sql.FunctionDecl
@@ -29,9 +31,7 @@ import scala.math.Ordering
 import argonaut._, Argonaut._
 import org.http4s.dsl._
 import pathy.Path._
-import scalaz._
-import scalaz.syntax.monad._
-import scalaz.syntax.std.boolean._
+import scalaz._, Scalaz._
 
 object metadata {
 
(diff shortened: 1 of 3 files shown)


web/src/main/scala/quasar/api/services/metadata.scala
```diff
@@ -92,19 +92,26 @@ object metadata {
...
       }.liftM[FileSystemErrT]
 
-    def dirMetadata(d: ADir): Free[S, QResponse[S]] = respond(
-      Q.ls(d)
-        .flatMap(mkNodes(d, _))
-        .map(nodes => Json.obj("children" := nodes.toList.sorted))
-        .run)
+    def dirMetadata(d: ADir, offset: Natural, limit: Option[Positive]): Free[S, QResponse[S]] = respond(
+      (offset.value.toIntSafe.toRightDisjunction(ApiError.fromMsg(BadRequest, "offset value is too large")) |@|
+       limit.traverse(_.value.toIntSafe.toRightDisjunction(ApiError.fromMsg(BadRequest, "limit value is too large")))) { (off, lim) =>
```
rintcius: Is it an idea to move these validations outside this method and integrate them with the ones in line 112? 

I _think_ it's possible to make refined types based off `Int` instead of `Long`.

jedesah: Yeah, it would be possible to make `Natural` and `Positive` be `Int`, but in most of the rest of the codebase, `Long` was a conscious decision. We could create new http4s param extractors that return `Int Refined Positive` but I think that's a little overkill. I considered just using `toInt` 😬  since we seem to do that in other places in the codebase, but decided I didn't want to risk having trouble sleeping last night 😆.

I think this is a bit of a special case because most of the time these values get pushed to connectors but in this case we need to use them on a `List` so I think it's reasonable to have the checks be fairly local to where the values actually need to be ints. If a similar check needs to be performed elsewhere and produce the same `ApiError` then, a function could be extracted so as not to duplicate functionality.

rintcius: I was thinking that it may make sense for all `limit` and `offset` pars to restrict it to `Int`s, but maybe there are reasons to keep them `Long`-based. 

I'm fine if you prefer to keep it as-is

jedesah: The reason for accepting a `Long` at the web layer is that the `Open` term of the `ReadFile` algebra takes a `Long`. Is it reasonable for a connector to ever have a file big enough to call for a `Long` to specify an offset? I don't know, but that is outside the scope of this PR I think.

sellout: Yeah, size is debatable – I would rather just use a Nat 😉

But remember that it’s not limited to file size – things like Joins and LeftShift can cause result sets that are much larger than any individual file.

jedesah: @sellout Good observations 😄 . I wonder if refined supports `BigInt Refined Positive`? 😆 Probably, I don't see why not, but I'm guessing we don't want to incur the performance penalty for a rather unlikely scenario... `Long` seems pretty reasonable to me. `Int.MaxValue` is not THAT big

rintcius:
@jedesah , I just restarted failing job (looks intermittent). Will merge once that passes

Issue #2477, opened 2017-07-14
Temporarily remove PostgreSQL connector


drostron:
The PostgreSQL connector only implements the filesystem algebras and the approach is almost certain to significantly change due to the original implementation targeting the minimum necessary to enable integration tests against the QScript implementation that we never reached. Removal until then eliminates the maintenance cost and revival from VCS when we return to it, at some yet to be scheduled point in the future, should be quick.

PR #2478, opened 2017-07-14
Prune PostgreSQL connector

drostron:
Closes: #2477


alissapajer:
@drostron FYI I don't think that "addresses" is a keyword that will close the issue, according to the keywords [here](https://help.github.com/articles/closing-issues-using-keywords/). And I don't _think_ it works with a colon either, but haven't tested it.

drostron:
@alissapajer Ah, good point on the keyword. I do frequently use a colon with success though.

drostron:
@alissapajer Ready

Issue #2483, opened 2017-07-15
Handle background errors by pushing them into the queue


djspiewak:
This is notably relevant to `TablePager`, which currently just swallows and logs.  We will want to change the current `Queue[Task, Vector[Data]]` to a `Queue[Task, Exception \/ Vector[Data]]]` and sequence the errors into the `Task` produced from `more`.

Issue #2481, opened 2017-07-15
Re-enable date-related mimir tests currently failing


djspiewak:
The exception:

```
java.time.format.DateTimeParseException: Text '2014-08-17T06:00:00Z' could not be parsed, unparsed text found at index 19
	at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1952)
	at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
	at quasar.precog.util.DateTimeUtil$.parseDateTime(DateTimeUtil.scala:46)
	at quasar.niflheim.Segments.detectDateTime(Segments.scala:117)
	at quasar.niflheim.Segments.addString(Segments.scala:124)
	at quasar.niflheim.Segments.initializeSegments(Segments.scala:249)
	at quasar.niflheim.Segments$$anonfun$initializeSegments$1.apply(Segments.scala:261)
	at quasar.niflheim.Segments$$anonfun$initializeSegments$1.apply(Segments.scala:259)
	at scala.collection.MapLike$MappedValues$$anonfun$foreach$3.apply(MapLike.scala:245)
	at scala.collection.MapLike$MappedValues$$anonfun$foreach$3.apply(MapLike.scala:245)
	at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
	at scala.collection.Iterator$class.foreach(Iterator.scala:893)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
	at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
	at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:732)
	at scala.collection.MapLike$MappedValues.foreach(MapLike.scala:245)
	at quasar.niflheim.Segments.initializeSegments(Segments.scala:259)
	at quasar.niflheim.Segments$$anonfun$extendWithRows$1.apply(Segments.scala:238)
	at quasar.niflheim.Segments$$anonfun$extendWithRows$1.apply(Segments.scala:237)
	at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
	at quasar.niflheim.Segments.extendWithRows(Segments.scala:237)
[...]

PR #2484, opened 2017-07-15
Implementing EquiJoin in Mimir

### LINKED ISSUE #2267: Handle equijoin
Just what the man says.

### LINKED ISSUE #2482: Detect and optimize static object/array manipulation
…in mimir.  Requires either pattern matching on `TransSpec` or using `paraM` to do the analogous match on `MapFunc`.

djspiewak:
Fixes #2267 
Fixes #2482

The big-ticket item here is joins, but a couple other things got fixed as well.  In order to make a lot of the tests work, I fixed the static/dynamic issue (without using `paraM`).  I also added error reporting to the background tasks (next on my plate is to move these errors into the paging stream where relevant: #2483).  As part of doing that, I discovered why the date/time related tests are timing out, which is to say #2481.  Hopefully that will be addressed as part of #2438.  The tests which trigger this issue will normally time out, eating up a lot of time, so I marked them as `skip` for now.

Anyway, back to joins…  One weird thing that I did is detect when qscript is secretly encoding a full cartesian, which it currently does by requesting a join on `[] = []`.  I did this partially for debugging, but I left it in because `Table#cross` is considerably more efficient than `Table#cogroup` in the event that the output is already known to be a full cartesian.  QScript really should have a better way of signaling this, so I don't have to do hacky detection stuff.


alissapajer:
@djspiewak you mentioned in the PR description that you are skipping some tests that timeout. Why not use the `"timeout"` keyword instead of `"skip"`? `"timeout"`ed tests will be skipped on travis but will run locally.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -237,7 +245,81 @@ object Mimir extends BackendModule with Logging {
...
+
+              transLeft <- combine.cataM[Backend, TransSpec1](
+                interpretM(
+                  {
+                    case qscript.LeftSide =>
+                      TransSpec1.Id.point[Backend]
+
+                    case qscript.RightSide =>
+                      tpe match {
+                        case JoinType.Inner | JoinType.RightOuter => TransSpec1.Undef.point[Backend]
```
alissapajer: Just wanted to check that the logic isn't swapped here. So we are creating `transLeft` (which means that we _just_ have an element from the left side in the `cogroup`) and the question is, what do we do when we have a `qscript.RightSide` in the `combine` and are forming a `JoinType.RightOuter`. 

For a `JoinType.RightOuter` we want to return all the values on the right side, plus the matched left values else undefined on the left. In this case we know we only have a value on the left which means it didn't have a matched value on the right, and so we return undefined. 

This all seems correct to me, just wanted to explain the logic to make sure we're in agreement.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -237,7 +245,81 @@ object Mimir extends BackendModule with Logging {
         }).liftB
     }
 
-    lazy val planEquiJoin: AlgebraM[Backend, EquiJoin[T, ?], Repr] = _ => ???
+    lazy val planEquiJoin: AlgebraM[Backend, EquiJoin[T, ?], Repr] = {
```
alissapajer: _Putting this comment in random spot we can have a thread about it._ 

1. We have two full stack integration tests that create a full outer join, `changedTestStatusInterim.test` and `changedTestStatus.test`. Can you move those both to the `joins/` directory? I'm not sure if this test (it's a single test and a workaround for it) should be passing for mimir; there might be some other underlying issue in LP or QScript.
2. Can you write new simpler tests for each of full outer, left outer, and right outer joins? We don't have any simple integration tests for any of those three. I think that mimir might be the first connector to support all these, and so I'd like to see tests for them!

it/src/main/resources/tests/complex-vars.test
```diff
@@ -2,7 +2,7 @@
     "name": "variable with a non-trivial value",
 
     "backends": {
-        "mimir":"timeout",
```
alissapajer: Is this not timing out anymore?

it/src/main/resources/tests/filterOnOid.test
```diff
@@ -2,7 +2,7 @@
     "name": "match on ObjectId",
     "NB": "This test isn’t really valid. This is neither the correct way to represent an OID (there is no generic way to do so), nor should the `_id` field even be in the data. But this is a reasonable placeholder until we actually expose the correct way to do more or less what this intends.",
     "backends": {
-        "mimir":"timeout",
+        "mimir":"skip",
```
alissapajer: Same.

it/src/main/resources/tests/joins/multipleSelectJoin.test
```diff
@@ -24,6 +23,7 @@
 
     "predicate": "atLeast",
     "ignoreResultOrder": true,
+    "ignoreFieldOrder": true,
```
alissapajer: We should be preserving the field order here for all connectors because we are prescribing the order of the fields in the select statement. So if mimir isn't preserving field order, we should instead have `"mimir": "ignoreFieldOrder"` above. And furthermore this would be a bug in mimir.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -96,6 +97,9 @@ object Mimir extends BackendModule with Logging {
     def unsafeCoerce[F[_ <: Cake]](term: F[Cake]): Task[F[P]] =
       Task.delay(term.asInstanceOf[F[P]])
 
+    def unsafeMerge(other: Repr): Task[Repr { type P = Repr.this.P.type }] =
```
alissapajer: Is there a better name for this, unless "merge" here is actually a technical term? Like `unsafeCoerceRepr` maybe?

mimir/src/main/scala/quasar/mimir/Precog.scala
```diff
@@ -96,7 +98,7 @@ final class Precog private (dataDir0: File)
     }
     val gated = vfsStr.mergeHaltBoth(vfsShutdownSignal.discrete.noneTerminate.drain)
 
-    gated.run.unsafePerformAsync(_ => ())
+    Precog.startTask(gated.run).unsafePerformSync
```
alissapajer: So here when we hit the error case in `startTask` we also have the option of calling `vfsLatch.countDown()` on the `CountDownLatch`. If we don't call `countDown()` and we do hit an exception, then we will log the exception and then hang indefinitely. I think we should avoid hanging.

alissapajer:
@djspiewak needs merge from master.

it/src/main/resources/tests/joins/multipleSelectJoin.test
djspiewak: This is actually a more general problem that we should think about.  This isn't a bug in mimir: it's by design.  And changing it would require an enormous amount of work.  Mimir just doesn't preserve field order through any operations whatsoever.

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak: The alternative is that we log the exception, release the lock and have a `Precog` object which is in an invalid state.  I'd rather release the lock while setting a flag that indicates an error, so we can raise an exception directly rather than hitting an ambiguous `NullPointerException` at some point in the future.

djspiewak:
@alissapajer Regarding `timeout` status, if we run them locally, it takes a *very* long time to complete a test run.  I just found it to be vastly, vastly too long to wait for timeouts on all of them.  (it dropped my local turnaround time from 40 minutes down to about 3).

it/src/main/resources/tests/joins/multipleSelectJoin.test
alissapajer: SQL^2 expects field order to be preserved in this situation, so we don't want to set `"ignoreFieldOrder": true` for all backends. Maybe in the future we will relax this constraint when we are focused less on SQL in the frontend.

But I do think this is a bug in mimir from _quasar's_ point of view because mimir is not preserving field order even though quasar expects it to. This isn't something urgent to fix in mimir. But this is a case where mimir is not following the specification required by quasar.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: > Can you move those both to the joins/ directory?

Willdo

> Can you write new simpler tests for each of full outer, left outer, and right outer joins?

I don't think I can.  QScript will never give rise to an `OuterObjectConcat` or `OuterArrayConcat` node, meaning that it isn't really possible to have an outer join which *doesn't* just go to undefined!  This may be special-cased somewhere, and certainly the `EquiJoin` node has enough richness to get around this problem, but I'm not sure exactly what qscript would do with it once it has it (the partial object), since such a value could never have been created "naturally".

alissapajer:
@djspiewak ok yeah, so if we're actually running all the integration tests locally, then I agree that using `"timeout"` isn't practical. 

Here is what I would really like then: `sbt it/test --skipTimeouts`. This would be a custom flag in sbt that skips all the tests marked as `"timeout"`. That way we can still mark them as timing out but can easily avoid the cost of having to wait 40 minutes.

mimir/src/main/scala/quasar/mimir/Precog.scala
alissapajer: Do you have a suggestion for how we release the lock and set this flag? Where does this flag get interpreted? As it is now, we will log and then hang forever.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: I was thinking of it in terms of "merge universes", where two different reprs come from divergent universes (meaning different cakes) and we want to bring them together.  I'm not sure I have a better name.

it/src/main/resources/tests/joins/multipleSelectJoin.test
djspiewak: ANSI SQL does indeed specify field order, though it's more an accident of how the spec is formalized (see: section **7.1** [here](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt)).  The problem is that *any* database backend which is compliant with JSON and not compliant with the "implicit extensions" (notably those which stem from JavaScript rather than JSON itself) will have problems with this.  Mimir is just a particularly obvious exemplar.

I strongly believe that, if quasar wants field order to be deterministic in some cases, it needs to be reflected in the qscript (i.e. by fields into an array before the final result is returned), rather than just expecting the backends to preserve things.  This is particularly important when passing through joins, where field order is by spec non-deterministic, and similarly when loading off disk.  Mimir doesn't care about SQL^2's semantics, it only cares about qscript and JSON, and according to the semantics of both of those elements, the current behavior is correct.

djspiewak:
@alissapajer If we're not running timing-out tests in Travis, then what's the point of `timeout` as a separate flag from `skip`?  I mean, it comes to the same thing, especially if we start just skipping them locally.  The only difference I can see is semantic, but by that standard, `skip` is actually *more* accurate in these cases, since the tests are only timing out because errors are not being re-sequenced into the output actions (#2483).

mimir/src/main/scala/quasar/mimir/Precog.scala
djspiewak:
```scala
@volatile
var isErrored = false
val latch = new CountDownLatch(1)

// ...

driver unsafePerformAsync { r =>
  r.fold(
    { err =>
      log.error("I knew you were trouble...", err)
      isErrored =true
      latch.countDown()
    },
    _ => ())
}

latch.await()

if (isErrored) {
  sys.error("...when you walked in")
}
```

it/src/main/resources/tests/joins/multipleSelectJoin.test
alissapajer: Sure, if qscript is asking too much of a backend, then we likely have a bug in qscript, not a bug in the backend.

alissapajer:
@djspiewak we have `timeout` as a separate flag from `skip` so we can know if a test has changed status from timing out to not timing out or vice versa. Meaning, if a test is labeled as `timeout` and you run it locally and that test doesn't time out, then you will get an error (in the same way that you get an error if a `pending` test passes). And additionally if a test is `pending` and it times out, you will be alerted via a test failure that you should change the flag to `timeout`. 

Though, it is true that if you run a test that is `skip`ped you won't be alerted that it is actually timing out because it is never run, and so we reserve the `skip` flag for tests that are crashing the process, or aren't supposed to pass (like non-key joins in couchbase).

So, if you have tests that are throwing exceptions, but are timing out because we are failing to handle those exceptions, then I think `skip` is fine.

I do agree that this isn't the most practical situation for a backend that has many timing out tests, which is why I suggested the `sbt` flag for when you want to run the tests quickly.

mimir/src/main/scala/quasar/mimir/Mimir.scala
alissapajer: What about the query `select * from foo left outer join bar on foo.a = bar.b`? That will be an `EquiJoin` with type `LeftOuter`. Why do you need an `OuterObjectConcat` or `OuterArrayConcat` in the qscript to have an outer join?

Are you referencing the fact that qscript just has `ConcatArrays` and doesn't distinguish between the inner and outer variants?

mimir/src/main/scala/quasar/mimir/Mimir.scala
alissapajer: Yeah, ok I think "merge" is fine.

mimir/src/main/scala/quasar/mimir/Mimir.scala
djspiewak: Hmm, I wonder what the qscript would look like in that case.  If it passes it through unmodified, then it should work out.

alissapajer: I just ran the qscript. It is 
```
EquiJoin(
  Unreferenced, 
  ShiftedRead(foo), 
  ShiftedRead(bar), 
  MakeArray(ProjectField((), _id)), 
  MakeArray(ProjectField((), _id)), 
  LeftOuter, 
  ConcatMaps(LeftSide, RightSide))
```
with a bunch of `Guard`s thown in. The `MakeArray` around the join keys aren't necessary, just a relic of how qscript is compiled.

djspiewak: Ok, that probably works then. I'll add those.

it/src/main/resources/tests/joins/leftOuterEquiJoin.test
```diff
@@ -0,0 +1,35 @@
...
+        "mongodb_q_3_4": "pending",
+        "spark_hdfs": "pending",
+        "spark_local": "pending"
+    },
+
+    "data": ["../cars.data", "../divide.data"],
+
+    "query": "select *
+              from `../divide` left outer join `../cars`
+              on cars.`_id` = divide.nr",
```
alissapajer: @djspiewak I merged this already, but I don't think this result is correct. This should return all records from `divide` and the matched records from `cars`. But `cars._id` is a string and `divide.nr` is an integer. So this condition should never be met, so we shouldn't have any results from `cars`. in the results.

Same problem with the `right outer join` tests.

alissapajer: Addressed in #2499.

PR #2491, opened 2017-07-16
Implements Union operation in Mimir

djspiewak:
Depends on #2484 

Relatively trivial change to enable the union implementation that was already there.  Can't enable any of the tests though since (it seems) all of the tests which involve union *also* involve reduce, which is not currently implemented.


djspiewak:
@alissapajer I'm pretty sure I can't.  There already *is* a very basic integration test for union… and it involves reduce.  Specifically, it involves reduce in the qscript; there is no reduction in the SQL.

alissapajer:
@djspiewak ah yeah you're right. So `select * from foo union select * from bar` compiles approximately to a logical plan of `Distinct(Union(Read(foo), Read(bar)))`. So we'd need `Distinct` implemented in mimir.

PR #2495, opened 2017-07-17
Skip distinctStar test for CB because it is flaky.

alissapajer:
This test has been failing a lot recently.

I don't like skipping tests that sometimes pass because that means the test isn't ever being run and so a bug could be introduced, but we also cannot have a couchbase build that fails so frequently.


drostron:
This seems to occur much more frequently when seeing increased timeouts on Travis.

drostron:
@alissapajer Needs a version label

alissapajer:
@drostron done.

PR #2497, opened 2017-07-17
create stdlib spec for mimir

### LINKED ISSUE #2450: Create a `StdLibSpec` for mimir
This will be a spec that extends `StdLibSpec` that probably also uses `MapFuncStdLibTestRunner`. Each backend has one of these and mimir needs one too.

alissapajer:
Closes #2450


it/src/test/scala/quasar/physical/marklogic/qscript/MarkLogicStdLibSpec.scala
```diff
@@ -63,7 +63,7 @@ abstract class MarkLogicStdLibSpec[F[_]: Monad: QNameGenerator: PrologW: MonadPl
       prg: FreeMapA[Fix, Nothing],
       expected: Data
     ): Result = {
-      val xqyPlan = planFreeMap(prg)(absurd)
+      val xqyPlan = planFreeMap[Nothing](prg)(absurd)
```
djspiewak: Side-note: our compilation times would be a *lot* lower if we did more stuff like this.  Particularly in cases where the type parameter ends up being inferred from a value that is implicitly-injected.

it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  private def check[A](fm: FreeMapA[Fix, A]): Option[Result] =
+    fm.cataM(interpretM[Result \/ ?, MapFuncCore[Fix, ?], A, Unit](
+      κ(().right),
+      shortCircuit)).swap.toOption
+
+  private def run[A](
+    freeMap: FreeMapA[Fix, A],
+    hole: A => cake.trans.TransSpec1)
```
djspiewak: We probably want `import cake.trans._` above, since we're doing a lot of this qualifying.  Or at the very least, `import cake._` and affect a broader set of qualifiers.

mimir/src/main/scala/quasar/mimir/Precog.scala
```diff
@@ -94,9 +94,14 @@ final class Precog private (dataDir0: File)
         vfsLatch.countDown()
       }
     }
+
     val gated = vfsStr.mergeHaltBoth(vfsShutdownSignal.discrete.noneTerminate.drain)
 
-    gated.run.unsafePerformAsync(_ => ())
+    gated.run.unsafePerformAsync {
+      case -\/(_) => vfsLatch.countDown()
```
djspiewak: Should be logging as well.  But we'll need to address this more once #2484 lands.

it/src/test/scala/quasar/physical/marklogic/qscript/MarkLogicStdLibSpec.scala
alissapajer: And aside from compile times, imo this makes the code a lot more readable, which is why I added this here.

it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
alissapajer: Yeah, for some reason it just doesn't bother me to have `cake.trans.XYZ` everywhere, which is probably why you have to keep commenting on my PRs about this. :)

I don't strongly object to the import though.

mimir/src/main/scala/quasar/mimir/Precog.scala
alissapajer: Oh yeah I actually meant to add a log there. But yeah we can address this more then.

djspiewak:
Needs merge with master, then good to go.

Issue #1505, opened 2016-10-04
Handle the `ThetaJoin` `JoinType` in the MarkLogic planner


wemrysi:
It is currently ignored which, I _think_, means all joins are implicitly inner.

wemrysi:
This will likely require more plan analysis and breaking joins up into multiple steps rather than the single (naive) `FLWOR` expression currently used.

sellout:
@wemrysi As always, think about ways that QScript could potentially make this easier.

In Mongo, all joins are outer joins (full outer in `mapReduce` or left outer in `$lookup`), and we post-filter cases with undefined values on one side or the other.

PR #2499, opened 2017-07-18
Fix up outer join integration tests.

alissapajer:
The previous outer join integration tests had incorrect expected results. This fixes the results as well as uses a different dataset for the queries so that we can fully tested the "outerness" of the joins.

The tests pass in old mongo and spark. They do not pass in mimir or marklogic.

### DIFF
--- a/it/src/main/resources/tests/joins/innerEquiJoinLarge.test
+++ b/it/src/main/resources/tests/joins/innerEquiJoinLarge.test
@@ -5,6 +5,7 @@
         "couchbase":         "pending",
         "marklogic_json":    "timeout",
         "marklogic_xml":     "timeout",
+        "mimir": "ignoreFieldOrder",
         "mongodb_q_3_2":     "pending"
     },
 
--- a/it/src/main/resources/tests/joins/innerEquiJoinLarge.test
+++ b/it/src/main/resources/tests/joins/innerEquiJoinLarge.test
@@ -1,0 +1,32 @@
+{
+    "name": "perform large inner equi-join",
+
+    "backends": {
+        "couchbase":         "pending",
+        "marklogic_json":    "timeout",
+        "marklogic_xml":     "timeout",
+        "mongodb_q_3_2":     "pending"
+    },
+
+    "NB": "#1587: Disabled in couchbase due to lack of general join.",
+
+    "data": ["../smallZips.data", "../zips.data"],
+
+    "query": "select smallZips.city, zips.state
+              from `../smallZips` join `../zips`
+              on smallZips.`_id` = zips.`_id`",
+
+    "predicate": "atLeast",
+    "ignoreResultOrder": true,
+
+    "expected": [{"city": "AGAWAM",       "state": "MA"},
+                 {"city": "CUSHMAN",      "state": "MA"},
+                 {"city": "BARRE",        "state": "MA"},
+                 {"city": "BELCHERTOWN",  "state": "MA"},
+                 {"city": "BLANDFORD",    "state": "MA"},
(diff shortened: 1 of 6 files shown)


alissapajer:
@wemrysi @jsantos17 do you expect outer joins to work on marklogic? This PR is adding outer join integration tests for ML and they are failing. I think they are returning empty set, but failing integration test output isn't the most clear (we should fix that at some point). I can confirm in the REPL.

wemrysi:
@alissapajer Nope, they don't work in MarkLogic. #1505 exists to track their (and other types) support.

alissapajer:
@wemrysi @jsantos17 cool thanks for the confirmation. FWIW, here are the results the query returns in the REPL. Not empty set.
```
💪 $ select cars.name as n1, cars2.name as n2 from cars2 left outer join cars on cars.`_id` = cars2.`_id` 
 
n1                 | n2             |
--------------------|----------------|
 RangeRover-Evoque  | Dodge-Caravan  |
 BMW-X5             | Jeep-Liberty   |
```
Those are the results of an inner join, so it seems that all joins are defaulting to inner.

djspiewak:
I'm pretty sure the reason this is failing on mimir is because mimir has inner semantics on `ObjectConcat` by default, and qscript doesn't request otherwise.  The `ObjectConcat` will arise from the fact that field names are being explicitly selected, as opposed to `select *`.  If this theory is accurate, then the test success in old mongo and spark is actually a bug.

As an aside though, I'm curious as to why you found the test wasn't working before.  The results did indeed demonstrate "outerness" in the rows, I just didn't bother testing outerness in the fields returned.

alissapajer:
@djspiewak so in the previous version of the test, the join condition was `on cars.`_id` = divide.val2`. One side of that is integers and the other is strings, so that join condition will never be met. So for the `left` and `right` variants of the outer joins, we should never have had any results "match" in the center, but your results expected that.

djspiewak:
@alissapajer Oh right, in retrospect that's obvious.  Curse you, untyped json.

Anyway, following up on my previous…  It looks like we're interpreting `ConcatMap` with outer semantics.  The test is failing in mimir for much more mysterious reasons.  I'm looking into it right now.  I think that should be fixed with this PR, rather than having mimir marked as `pending`.

alissapajer:
@djspiewak yeah, I agree that we have to be interpreting `ConctMap` with full outer semantics in order for this to be working in any qscript-based backend. And to be clear what I mean by that, applying the join condition will create `ConcatMaps({foo: undefined}, {bar: 5})` in the `repair` where we expect that to return `{foo: undefined, bar: 5}`. I have not completely convinced myself yet that this actually what we want in _all_ cases of concats though this is how we've been operating so far. I need to think about this more. We can continue this conversation.

Feel free to push mimir changes directly to this PR. If the changes to mimir are going to take a while though, I'd prefer to just get this merged so we don't have incorrect integration tests in master.

djspiewak:
@alissapajer I've already fixed it.  Will push a branch in a second.

djspiewak:
BTW, mimir fixes PRed in https://github.com/alissapajer/quasar/pull/10

alissapajer:

it/src/main/resources/tests/joins/innerEquiJoinLarge.test
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "perform large inner equi-join",
+
+    "backends": {
```
djspiewak: While we're at it, can you add `"mimir": "ignoreFieldOrder"` here as well?

alissapajer:
@djspiewak ready for merge.

Issue #2319, opened 2017-06-13
MarkLogic XQueryException: Expanded tree cache full


alissapajer:
Saw this integration test fail on travis marklogic_xml build.

```
Running Streaming Queries [......................................................................................................................................................................]
[error] ! distinct * [./distinctStar.test] (28 seconds, 489 ms)
[error]  quasar.physical.marklogic.xcc.DefaultImpl$XccXQueryException: XDMP-EXPNTREECACHEFULL: filesystem:directory-contents("/quasar-test/run_a69962ff/regression/") -- Expanded tree cache full on host 08009478fd2a.c.eco-emissary-99515.internal uri /quasar-test/run_a69962ff/regression/zips on line 25
[error]  ---BEGIN QUERY---
[error]  xquery version "1.0-ml";
[error]  
[error]  declare namespace filesystem = "http://quasar-analytics.org/filesystem";
[error]  
[error]  declare function filesystem:as-child-dir($parent as xs:string, $child as xs:string) as xs:string {
[error]    fn:concat($parent, fn:tokenize(fn:substring-after($child, $parent), "/")[1], "/")
[error]  };
[error]  
[error]  declare function filesystem:descendant-uri-from-lexicon($parent as xs:string, $filter as cts:query) as xs:string? {
[error]    cts:uri-match(fn:concat($parent, "*/*"), ("document"), $filter)[1]
[error]  };
[error]  
[error]  declare function filesystem:descendants-having-format-exist($uri as xs:string) as xs:boolean {
[error]    fn:exists(cts:search(fn:doc(), cts:directory-query($uri, "infinity"), ("format-xml"), 1.0, ()))
[error]  };
[error]  
[error]  declare function filesystem:directory-contents($uri as xs:string) as xs:string* {
[error]    fn:map(function($kid) { fn:substring-after($kid, $uri) }, (fn:filter(function($sdir) { filesystem:descendants-having-format-exist($sdir) }, filesystem:discover-children($uri, ())), fn:map(fn:base-uri#1, cts:search(fn:doc(), cts:directory-query($uri, "1"), ("format-xml"), 1.0, ()))))
[error]  };
[error]  
[error]  declare function filesystem:discover-children($parent as xs:string, $known as xs:string*) {
[...]

PR #2302, opened 2017-06-08
Attempt to limit the MarkLogic docker instance's memory usage.

wemrysi:

PR #2510, opened 2017-07-20
Skip distinctStar test for marklogic because it is flaky.

alissapajer:
This test is failing frequently enough in travis that I propose skipping it so that we can have a reliable build.

### DIFF
--- a/it/src/main/resources/tests/distinctStar.test
+++ b/it/src/main/resources/tests/distinctStar.test
@@ -2,6 +2,8 @@
     "name": "distinct *",
     "backends": {
         "couchbase": "skip",
+        "marklogic_json": "skip",
+        "marklogic_xml": "skip",
         "mimir":"pending",
         "mongodb_2_6":       "pending",
         "mongodb_3_0":       "pending",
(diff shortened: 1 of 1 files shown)


wemrysi:
@alissapajer In #2302 I aggressively reduced the size of the expanded tree cache to try and solve #2241, possibly causing #2319. Instead of disabling, we could also try increasing the cache a bit to see if we can find a balance between allowing the queries to run and not bumping into whatever global memory limits travis imposes.

Do you recall if the recent failures have been due to the error described in #2319? Something else?

alissapajer:
@wemrysi awesome I'm all for fixing this rather than disabling the test, if we can.

The failure I saw yesterday is still available [here](https://travis-ci.org/quasar-analytics/quasar/jobs/255463477#L926). It is the "Expanded tree cache full" error. So let's try increasing the cache a bit to see if we can find that balance. It was set to `64` in that PR, maybe double to `128`? Do you know what the default was before we imposed those manual limits?

wemrysi:
@alissapajer Excellent, the default is [determined by the amount of memory available to MarkLogic](https://help.marklogic.com/Knowledgebase/Article/View/420/0/group-level-cache-settings-based-on-ram), it appears, so I'd guess it was defaulting to at least 1GB before.

I'd say let's try doubling and see how that fares, I don't really have a good sense of how much is needed for these queries and, ideally, the requirement should decrease as the compiler/planner improve.

PR #2517, opened 2017-07-20
Implements Reduce in mimir

alissapajer:
Closes #2439

PR #2516, opened 2017-07-20
"Fixed" date parsing in mimir and added pendingIgnoreFieldOrder

djspiewak:
Fixes #2512 
Fixes #2481

This involved some considerable hackery inside of NIHDB.  Basically, `Date` works in terms of `Instant`, while `RValue` (i.e. mimir) works in terms `ZonedDateTime`.  So we're handling this right now by parsing the instant and then assigning it to the UTC timezone.  This is most definitely not correct, but it'll sort of work for now.  Long-term, we should fix the impedance mismatch between mimir and `Data`.

PR #2519, opened 2017-07-20
Implements Sort in mimir

### LINKED ISSUE #2518: Implement `Sort` in mimir
For a single bucket.  Analogous to #2439

djspiewak:
Depends on #2516 and #2517 
Fixes #2518 

Implements sorting without bucketing.  The implementation obviously depends on the `Table#sort` function being stable.  Fortunately, it is!  So we're good.  A couple tests that *should* have sprung to life… have not.  So that's annoying.  Also there's a weird bug in `pendingIgnoreFieldOrder`.  Not sure what's up with that.

Anyway, the tests that should have sprung to life seem to be sorting on the wrong thing.  The `FreeMap` is getting translated correctly, and we're passing it along to `Table#sort` correctly, so what's probably happening here is a bug in either JDBM3 or in `BlockStoreColumnarTable`.  Hard to tell which.  Will require further investigating.

### DIFF
--- a/it/src/main/resources/tests/filterOnFieldComparison.test
+++ b/it/src/main/resources/tests/filterOnFieldComparison.test
@@ -1,7 +1,7 @@
 {
     "name": "filter on field comparison",
     "backends": {
-        "mimir":"pendingIgnoreFieldOrder"
+        "mimir":"ignoreFieldOrder"
     },
     "data": "zips.data",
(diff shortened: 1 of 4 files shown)


djspiewak:
The build failure here will go away when merged with master.

alissapajer:
@djspiewak there are merge conflicts

Issue #2485, opened 2017-07-16
Mimir internal Repr should track current sort order


djspiewak:
This makes it possible to do a couple of things.  Number one, it opens the door to naive indexing (sort hints on `Read` in qscript).  Number two, it allows successive joins to be better optimized, with stable sorting carried through multiple operations.

PR #2522, opened 2017-07-20
Implement bucketed sorting and reduction in mimir

### LINKED ISSUE #1990: Fiddling with grouped reduction (groupByN/partitionMap)
`Table#groupByN` is clearly an incredibly powerful primitive, but it's possible that we don't even need it to implement the grouped reduction functionality which is in qscript.  It seemed at first glance like we might be able to implement this via `partitionMap`.  If that's the case, it would likely be vastly more efficient and long-term easier to maintain.

Either way, it has to be implemented.

### LINKED ISSUE #1991: Fiddling with grouped sorting (partitionMap?)
Similar to #1990, grouped sorting is something that isn't natively on `Table`, but might be doable with `partitionMap`.  If it's not doable with `partitionMap`, it's *certainly* doable with `groupByN`.

### LINKED ISSUE #2266: Implement the rest of qscriptcore
- [x] `Filter`
- [ ] `Reduce`
  + [x] Assume provenance order 1 first (#2439) 
  + [ ] Higher order depends on #1990 
- [ ] `Sort`
  + [x] Assume provenance order 1 first (#2518)
  + [ ] Higher order depends on #1991
- [x] `Union`
- [x] `LeftShift` (delegated to #1988)

djspiewak:
Depends on #2519, which itself depends on #2516 and #2517
Fixes #1991 
Fixes #1990 
Fixes #2266 

Bucketing turned out to be way easier than expected.  Basically, `partitionMerge` does exactly what we want.  Note for future: we should eliminate `groupByN` and simplify the implementation into `sort`, since we don't actually need its full generality (grouping in SQL^2 is considerably weaker than in Quirrel).  We have to explicitly sort on the bucket key, similar to the issue we have equijoin, but this should be made much nicer with #2485.

### DIFF
--- a/it/src/main/resources/tests/filterOnFieldComparison.test
+++ b/it/src/main/resources/tests/filterOnFieldComparison.test
@@ -1,7 +1,7 @@
 {
     "name": "filter on field comparison",
     "backends": {
-        "mimir":"pendingIgnoreFieldOrder"
+        "mimir":"ignoreFieldOrder"
     },
     "data": "zips.data",
     "query": "select * from zips where pop < loc[1] order by `_id` limit 10",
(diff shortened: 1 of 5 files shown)


alissapajer:
@djspiewak more merge conflicts

PR #2305, opened 2017-06-08
WIP / Remove old mongo

sellout:
Fixes #1940 (and others to be enumerated).

PR #2523, opened 2017-07-21
Update new mongo branch with master

wemrysi:
Merges `master` into the new mongo branch and resolves all the conflicts.


wemrysi:
There are some failing mongo planner tests and lib tests that were failing before the merge, so I don't think anything has worsened as a result of the merge.

We may see some it tests for other connectors fail due to the merge as I may have mixed up a enabling/disabling a test for a particular connector (there were conflicts on pretty much every `*.test` file).

alissapajer:
@wemrysi sounds good, let's sort out all the full stack integration tests before we merge this, but I think it's fine to merge with any mongo unit tests failing that were failing before the latest merge of master.

rintcius:
Shall we first update `next-major` with `master`? I see a lot of changes in the diff because of `next-major` not being uptodate.

alissapajer:
@rintcius I'd be fine with that. But the diff against master should be identical to the diff in #2305 which was reviewed, modulo handling merge conflicts. So I was thinking that we could merge this without extensive review because we already reviewed it over in #2305.

rintcius:
@alissapajer Ok I'm fine with merging as-is. Was just thinking it may help in looking into those failures in IT tests. We can also do that after merge as long as we get this build green ASAP.

alissapajer:
@rintcius yeah, that's a good point that we don't want to have any regressions in integration tests, and a clean diff against master would make it easy to see that we didn't skip/pending any extra tests in new mongo. 

Ok, I'm convinced, let's PR master to next-major and get that merged. Then ideally the diff here will just update to reflect that merge, but with diffs as large as this I've seen that fail in the past. So we might have to re-open this PR.

alissapajer:
The diff is cleaner over here https://github.com/quasar-analytics/quasar/pull/2524. I'll keep this open for now, but let's see if we can get https://github.com/quasar-analytics/quasar/pull/2524 merged and just close this. They are ultimately the same diff, but maybe the integration tests are in a better state over there.

rintcius:
I _think_ the diff of #2524 looks cleaner because it's not entirely uptodate with `master` (eg no diff in `.travis.yml` where there's a diff in this PR). So maybe best to merge this one first and then merge #2524

alissapajer:
I think merging the other way around might make more sense. So we merge #2524 first which gets new mongo cleanly into next-major. Then we merge master into next-major (which is what this PR does).

alissapajer:
Not sure what I was thinking before, but we are definitely going to have regressions in mongo integration tests against new mongo; we know that everything is not working yet.

rintcius:
Merged as-is

PR #2524, opened 2017-07-21
Replace old Mongo with new QScript-based Mongo

sellout:
This fixes a number of issues (those currently marked “on hold”) and completely eliminates the pre-QScript Mongo connector.


rintcius:
As discussed merged as-is

Issue #2169, opened 2017-05-08
Simplify handling of “date_part” for connectors.


sellout:
Currently there are 21 `Extract` MapFuncs that connectors need to handle, but many of these are simply SQL-isms that we shouldn’t propagate through.

Some thoughts on what we can do:
1. switch to a single function (`TemporalPart`?) with an enumeration like `TemporalTrunc`;
2. rename `IsoDayOfWeek` to `DayOfWeek` and `IsoYear` to `YearOfWeek`;
3. remove `Epoch`, making it a separate function like `ToEpoch`;
3. eliminate `Century`, `DayOfWeek`, `Decade`, `Microseconds`, `Millennium`, `Milliseconds`, and `Quarter` (defining them in terms of the remaining extractors either in the SQL² compiler or via some stdlib of parameterized views – this may require); and
4. eliminate the non-“standard” SQL calls, like `extract_timestamp_hour(date)` in favor of `date_part("timestamp_hour", date)`.

Things I’m still waffling on:
* in my proposed formulation, they all return integers _except_ `Second`. If `TemporalPart` is restricted to integers, then we probably need to keep `Milliseconds` and `Microseconds`
* can we somehow share an enumeration with `TemporalTrunc`? `-Trunc` has some components we’re getting rid of (like `Century`), and `-Part` has multiple notions with the same precision (like `DayOfYear` and `DayOfMonth`), so there’s some conflict, but maybe we could do use a type like `(TemporalPrecision, DateFormat)` for the `-Part` parameter where `DateFormat` indicates which of the ISO formats we’re using, but then `(Month, Ordinal)` doesn’t make sense, because there’s no months in the ordinal format.

zeeiyerWork:
@sellout - Does @djspiewak merge #2172 resolve this fully or partially. Based on that please add a comment and leave this open or close it. Thanks.

sellout:
@zeeiyerWork Nope – that PR is related to SQL², this issue is about QScript.

zeeiyerWork:
More of an improvement - so will be scheduled for later.

PR #2525, opened 2017-07-21
WIP / Handle complex provenance merging better.

sellout:


alissapajer:
@sellout thanks!

connector/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -757,27 +722,20 @@ object MapFuncs {
...
-  @Lenses final case class ExtractDecade[T[_[_]], A](a1: A) extends Unary[T, A]
   @Lenses final case class ExtractDayOfWeek[T[_[_]], A](a1: A) extends Unary[T, A]
   @Lenses final case class ExtractDayOfYear[T[_[_]], A](a1: A) extends Unary[T, A]
   @Lenses final case class ExtractEpoch[T[_[_]], A](a1: A) extends Unary[T, A]
   @Lenses final case class ExtractHour[T[_[_]], A](a1: A) extends Unary[T, A]
-  @Lenses final case class ExtractIsoDayOfWeek[T[_[_]], A](a1: A) extends Unary[T, A]
-  @Lenses final case class ExtractIsoYear[T[_[_]], A](a1: A) extends Unary[T, A]
-  @Lenses final case class ExtractMicroseconds[T[_[_]], A](a1: A) extends Unary[T, A]
-  @Lenses final case class ExtractMillennium[T[_[_]], A](a1: A) extends Unary[T, A]
-  @Lenses final case class ExtractMilliseconds[T[_[_]], A](a1: A) extends Unary[T, A]
```
alissapajer: @sellout was deleting all these map funcs intentional? Or even if it wasn't meant for this diff, was there some thinking that we don't need them anymore?

sellout: Ah, that got lumped into this commit accidentally. The intent is that connectors don’t need to implement all of these operations – Eg, `ExtractCentury` can be implemented as `Floor(ExtractYear / 100)`. There’s now `MapFuncDerived` which could have these cases in it, but my thinking at the time was that they would just be handled directly by the SQL² compiler. These changes cover some of the things mentioned in #2169.

alissapajer:
Closing this since I'll be rebasing the commit to exclude the `MapFunc` deletions.

PR #2526, opened 2017-07-21
Port over TimeLib from precog

### LINKED ISSUE #2438: implement mimir `TimeLib`

alissapajer:
Closes #2438 

Enables the very straightforward `StdLib` tests.


blueeyes/src/main/scala/quasar/precog/util/DateTimeUtil.scala
```diff
@@ -30,8 +30,26 @@ object DateTimeUtil {
...
+    } catch {
+      case _: Throwable => try {
+        ZonedDateTime.of(
+          LocalDate.parse(value, DateTimeFormatter.ISO_LOCAL_DATE),
+          LocalTime.of(0, 0, 0, 0),
+          utc)
+      } catch {
+        case _: Throwable =>
+          ZonedDateTime.of(
+            LocalDate.now(utc), // FIXME what is the expected default?
```
djspiewak: I think the expected default is no default at all.  We shouldn't normalize zoneless time.  That's part of where we have an impedance mismatch right now.  Precog needs a representation of `LocalDate`, and ideally would also have a representation of zoneless `Instant`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/cf/Util.scala
```diff
@@ -200,6 +203,15 @@ object util {
       }
   }
 
+  def CoerceToDate = CF1P("builtin:ct:coerceToDate") {
```
djspiewak: I'm actually surprised we didn't already have this.

blueeyes/src/main/scala/quasar/precog/util/DateTimeUtil.scala
```diff
@@ -30,8 +30,26 @@ object DateTimeUtil {
   def looksLikeIso8601(s: String): Boolean = dateTimeRegex.matcher(s).matches
```
djspiewak: So this is the function that NIHDB calls to determine whether or not it's going to coerce a string to a date during storage.  Do we need to update `dateTimeRegex`?

blueeyes/src/main/scala/quasar/precog/util/DateTimeUtil.scala
alissapajer: There are tests in `StdLibSpec` which expect a `LocalDate` to be coerced to a date time with `0` for each of the time fields. So this comment was wondering if there are any tests that expect a `LocalTime` to have some default notion of date.

But yeah, I agree that no default is the goal, and we are just doing the best we can with this mismatch for now.

yggdrasil/src/main/scala/quasar/yggdrasil/table/cf/Util.scala
alissapajer: Yeah I was too!

blueeyes/src/main/scala/quasar/precog/util/DateTimeUtil.scala
alissapajer: Yeah, so we do need to do something here. The single call site for this looks like this
```scala
  def detectDateTime(s: String): ZonedDateTime = { 
    if (!DateTimeUtil.looksLikeIso8601(s)) return null
    try {
      DateTimeUtil.parseDateTime(s)
    } catch {
      case e: IllegalArgumentException => null
    }   
  }
```
What's the value in regexing the string first to see if it has a chance at parsing if we are just going to return `null` either way. Maybe we were concerned about the overhead of parsing? I think we should just get rid of `looksLikeIso8601` altogether, unless you have a good argument for why this regex is important.

Also, `null`, seriously? Like, did we not know about `Option` at precog?

djspiewak: LOL

alissapajer: Ah yeah, ingest. I'd ilke to see the benchmarks of running the regex versus running the conversion.

Issue #2520, opened 2017-07-20
Some sort cases don't quite work


djspiewak:
As noted in #2519.  There appears to be a bug somewhere in `Table#sort`.  You can see this by enabling `joins/orderJoinByAlias.test` for mimir.  It appears that, out of two possible columns to sort on, JDBM3 is sorting on the wrong one.  We're giving `Table#sort` the appropriate projection, so it has to be after that point.

PR #2527, opened 2017-07-22
Enabled (a lot) more of the stdlib in mimir

djspiewak:
There's still quite a few things not enabled, mostly time-related.  There are a few things here that I know *how* to implement, I just haven't yet (like `IfDefined`).  Two notable things that I'm aware of being priorities right now:

- `Guard` with non-`Undefined` alternates actually happen in several join-related queries
- Speaking of stuff…  Any query with a sort in it is at risk of hitting #2520 I had to disable a couple for that reason.  We'll have to track that down.

Integration tests are now hella long.


it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
```diff
@@ -224,6 +224,13 @@ class MimirStdLibSpec extends StdLibSpec with PrecogCake {
   }
 
   tests(runner)
+
+  private object ExtractBin {
```
alissapajer: We need this out of here and in a public place in the `connector` project.

alissapajer: It will also make sense to also provide extractors that don't include `Embed`.

mimir/src/main/scala/quasar/mimir/MapFuncCorePlanner.scala
```diff
@@ -170,14 +192,28 @@ final class MapFuncCorePlanner[T[_[_]]: RecursiveT, F[_]: Applicative]
...
+
+      case MapFuncsCore.DeleteField(src, ConstLiteral(CString(key), _)) =>
+        (ObjectDelete[A](src, Set(CPathField(key))): TransSpec[A]).point[F]
+
+      // mimir doesn't have a way to implement this
       case MapFuncsCore.DeleteField(src, field) => ???
 
       case MapFuncsCore.Meta(a1) => ???
 
+      // this returns rows, rather than array; literally cannot be implemented in mimir at present
```
alissapajer: Maybe we could build an array and then flatten it. Need to check what the provenance requirements are for `Range`.

yggdrasil/src/test/scala/quasar/yggdrasil/table/ColumnarTableModuleSpec.scala
```diff
@@ -509,7 +521,7 @@ trait ColumnarTableModuleSpec[M[+_]] extends TestColumnarTableModule[M]
     }
   }
 
-  "partitionMerge" should {
+  /*"partitionMerge" should {
```
alissapajer: What's wrong with this spec?

alissapajer:
@djspiewak since we've made good strides on the correctness/implementation side of things, we should enumerate the slow parts so we can improve performance.

mimir/src/main/scala/quasar/mimir/MapFuncCorePlanner.scala
djspiewak: It's not so much the provenance requirements for `Range` but the invariants on `SliceTransform` that are preventing this.  No `TransSpec` may change the size of a slice, and the corollary of that is that no *constant* introduction may do anything other than exactly match the size of the incoming slice.  This is easily handled with a single-row constant (just repeat the same constant zero-or-more times), but multi-row constants are impossible.

If `Range` were actually implemented as an array construction syntax (that you could subsequently flatten), this would be extremely doable, though inefficient for large ranges.  As it stands, there's just no way to do it as a `TransSpec`.

yggdrasil/src/test/scala/quasar/yggdrasil/table/ColumnarTableModuleSpec.scala
djspiewak: It had the extreme misfortune of being forgotten by me when I uncommented the tests.

PR #2528, opened 2017-07-22
Hey, more mimir stdlib!

djspiewak:
Also the implementation of `Search` was completely wrong, so I've fixed that now.  A bunch more tests have sprung to life.

### DIFF
--- a/frontend/src/test/scala/quasar/std/StdLibSpec.scala
+++ b/frontend/src/test/scala/quasar/std/StdLibSpec.scala
@@ -90,27 +90,28 @@ abstract class StdLibSpec extends Qspec {
 
       "Search" >> {
         "find contents within string when case sensitive" >> {
-          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str(".*ch.*"), Data.Bool(true), Data.Bool(true)) and
-            ternary(Search(_, _, _).embed, Data.Str("China"), Data.Str("^Ch.*$"), Data.Bool(true), Data.Bool(true)) and
-            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*ch.*"), Data.Bool(true), Data.Bool(true))
+          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str(".*ch.*"), Data.Bool(false), Data.Bool(true)) and
+            ternary(Search(_, _, _).embed, Data.Str("China"), Data.Str("^Ch.*$"), Data.Bool(false), Data.Bool(true)) and
+            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*ch.*"), Data.Bool(false), Data.Bool(true))
         }
 
         "reject a non-matching string when case sensitive" >> {
-          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str("^bs.*$"), Data.Bool(true), Data.Bool(false)) and
-            ternary(Search(_, _, _).embed, Data.Str("china"), Data.Str("^bs.*$"), Data.Bool(true), Data.Bool(false)) and
-            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*bs.*"), Data.Bool(true), Data.Bool(false))
+          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str("^bs.*$"), Data.Bool(false), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("china"), Data.Str("^bs.*$"), Data.Bool(false), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*bs.*"), Data.Bool(false), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*CH.*"), Data.Bool(false), Data.Bool(false))
         }
 
         "find contents within string when case insensitive" >> {
-          ternary(Search(_, _, _).embed, Data.Str("Church"), Data.Str(".*ch.*"), Data.Bool(false), Data.Bool(true)) and
-            ternary(Search(_, _, _).embed, Data.Str("cHina"), Data.Str("^ch.*$"), Data.Bool(false), Data.Bool(true)) and
-            ternary(Search(_, _, _).embed, Data.Str("matCHing"), Data.Str(".*ch.*"), Data.Bool(false), Data.Bool(true))
+          ternary(Search(_, _, _).embed, Data.Str("Church"), Data.Str(".*ch.*"), Data.Bool(true), Data.Bool(true)) and
+            ternary(Search(_, _, _).embed, Data.Str("cHina"), Data.Str("^ch.*$"), Data.Bool(true), Data.Bool(true)) and
+            ternary(Search(_, _, _).embed, Data.Str("matCHing"), Data.Str(".*ch.*"), Data.Bool(true), Data.Bool(true))
         }
 
         "reject a non-matching string when case insensitive" >> {
-          ternary(Search(_, _, _).embed, Data.Str("Church"), Data.Str("^bs.*$"), Data.Bool(false), Data.Bool(false)) and
-            ternary(Search(_, _, _).embed, Data.Str("cHina"), Data.Str("^bs.*$"), Data.Bool(false), Data.Bool(false)) and
-            ternary(Search(_, _, _).embed, Data.Str("matCHing"), Data.Str(".*bs.*"), Data.Bool(false), Data.Bool(false))
+          ternary(Search(_, _, _).embed, Data.Str("Church"), Data.Str("^bs.*$"), Data.Bool(true), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("cHina"), Data.Str("^bs.*$"), Data.Bool(true), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("matCHing"), Data.Str(".*bs.*"), Data.Bool(true), Data.Bool(false))
         }
       }
 
--- a/frontend/src/test/scala/quasar/std/StdLibSpec.scala
+++ b/frontend/src/test/scala/quasar/std/StdLibSpec.scala
@@ -89,7 +89,29 @@ abstract class StdLibSpec extends Qspec {
       // NB: `Like` is simplified to `Search`
 
       "Search" >> {
-        todo
+        "find contents within string when case sensitive" >> {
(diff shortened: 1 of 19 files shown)


djspiewak:
Looks like literally everyone has a broken implementation of `Search` that was only revealed by adding these tests.  We're going to have to turn off their tests.

yggdrasil/src/test/scala/quasar/yggdrasil/table/ColumnarTableModuleSpec.scala
```diff
@@ -521,7 +521,7 @@ trait ColumnarTableModuleSpec[M[+_]] extends TestColumnarTableModule[M]
     }
   }
 
-  /*"partitionMerge" should {
+  "partitionMerge" should {
```
alissapajer: :+1:

mimir/src/main/scala/quasar/mimir/StringLib.scala
```diff
@@ -228,26 +228,43 @@ trait StringLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
...
+
+      case List(target: StrColumn, pattern: DateColumn, flag: BoolColumn) =>
+        searchDynamic(List(target, dateToStrCol(pattern), flag)).get
+
+      case List(target: DateColumn, pattern: DateColumn, flag: BoolColumn) =>
+        searchDynamic(List(dateToStrCol(target), dateToStrCol(pattern), flag)).get
+    }
 
-    def searchStatic(pattern: String, flag: Boolean) = {
+    // please use this one as much as possible. it is orders of magnitude faster than search
```
alissapajer: This comment is unclear because this function is also named `search` so maybe say where the "search" in the comment is coming from.

mimir/src/main/scala/quasar/mimir/ArrayLib.scala
```diff
@@ -25,8 +25,200 @@ import quasar.yggdrasil.table._
...
+        new BoolColumn {
+          def apply(row: Int) = target(row) >= lower(row) && target(row) <= upper(row)
+          def isDefinedAt(row: Int) = target.isDefinedAt(row) && lower.isDefinedAt(row) && upper.isDefinedAt(row)
+        }
+
+      case List(target: NumColumn, lower: NumColumn, upper: NumColumn) =>
+        new BoolColumn {
+          def apply(row: Int) = target(row) >= lower(row) && target(row) <= upper(row)
+          def isDefinedAt(row: Int) = target.isDefinedAt(row) && lower.isDefinedAt(row) && upper.isDefinedAt(row)
+        }
```
alissapajer: This boilerplate is horrendous. I'd really like to be able to just define this `BoolColumn` once. We'd need a way to abstract across `target`, `lower`, and `upper` so all the types line up. This is a pervasive problem across all of the mimir std libs, it's just extra noticeably bad here :/ So fine to merge as is, but I'd like to clean this up.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -89,7 +89,29 @@ abstract class StdLibSpec extends Qspec {
       // NB: `Like` is simplified to `Search`
 
       "Search" >> {
-        todo
+        "find contents within string when case sensitive" >> {
+          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str(".*ch.*"), Data.Bool(true), Data.Bool(true)) and
```
alissapajer: From [the docs](https://github.com/quasar-analytics/quasar/blob/5624a1c0a23df962646ebf9060b0edaab048757c/frontend/src/main/scala/quasar/std/string.scala#L123): "If the third argument is true, then it is a case-insensitive match."

These tests all have case sensitivity swapped, both according to the test description and their actual content.

alissapajer: Also, you're missing the tests for when it's case sensitive by fails to match due to case sensitivity (when it would otherwise match if we were testing the case-insensitive case).

mimir/src/main/scala/quasar/mimir/MapFuncCorePlanner.scala
```diff
@@ -134,12 +134,20 @@ final class MapFuncCorePlanner[T[_[_]]: RecursiveT, F[_]: Applicative]
       case MapFuncsCore.Gte(a1, a2) =>
         Infix.GtEq.spec(a1, a2).point[F]
 
-      case MapFuncsCore.IfUndefined(a1, a2) => ???
+      // this function is completely unspecified; I'm guessing on semantics here
```
alissapajer: The function is textually specified in the [relations lib](https://github.com/quasar-analytics/quasar/blob/5624a1c0a23df962646ebf9060b0edaab048757c/frontend/src/main/scala/quasar/std/relations.scala#L142). Though your implementation is correct.

it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
```diff
@@ -116,8 +116,6 @@ class MimirStdLibSpec extends StdLibSpec with PrecogCake {
     case MapFuncsCore.Now() => notImplemented.left
     case MapFuncsCore.TypeOf(_) => notImplemented.left
     case MapFuncsCore.Negate(_) => notImplemented.left // TODO this isn't passing because -Long.MinValue == Long.MinValue, so basically a limitation in ColumnarTable
-    case MapFuncsCore.IfUndefined(_, _) => notImplemented.left
-    case MapFuncsCore.Between(_, _, _) => notImplemented.left
```
alissapajer: Are there any stdlib specs that should be coming alive for `IfUndefined` and `Between`? If none exist, we should write some and include them in this PR.

mimir/src/main/scala/quasar/mimir/StringLib.scala
djspiewak: Yeah the comment dates from before I renamed the functions.

mimir/src/main/scala/quasar/mimir/ArrayLib.scala
djspiewak: I'd love to clean this up.  Honestly, boilerplate like this is part of why I was always so opposed to functions with arity >2 (back in Precog).  But unfortunately, I'm not entirely sure how to clean it up beyond some extra column builder functions (like the `BoolFrom` stuff).  But that too, incurs more boilerplate.  Remember that `apply` is concrete, and cannot be abstracted over, because the whole point is to avoid boxing.  So there is no generic implementation of all of these cases; they are merely syntactically equivalent.

mimir/src/main/scala/quasar/mimir/MapFuncCorePlanner.scala
djspiewak: Ah I grepped for `IfDefined`, not `IfUndefined`.

it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
djspiewak: Plenty for `Between`.  `IfUndefined` is untestable with the current test harness, though several integration tests touch it.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
djspiewak: Ah RIP…

mimir/src/main/scala/quasar/mimir/ArrayLib.scala
jdegoes: Well, technically switching from lazy to eager will eliminate the partial functions and take away some boilerplate. Though there will always be some due to the nature of the beast.

djspiewak: The partial functions will still be here, they will just be lifted from `Int => A` (where `A` is the component type of the column) to `Array[A]`.  So, still a metric ton of boilerplate.

Also, looking at *exactly* the way that things compose, and also getting some preliminary measurements on the effects of `materialized` (which forces `Table` into an eager mode), I think you're significantly overestimating the impact of the eager changeover.  There's no question that eliminating megamorphic calls within the hot path is in theory a huge win, but the *penalty* is an enormous amount of memory pressure on intermediate columns.  It's not as much of a slam dunk as it seems, absent the details.  But measurement will tell when we get to that point.

jdegoes: I do think all the isDefinedAt checks and delegations will go away, since it will all be hand-coded while loops that traverse the length of the span. But yeah, still plenty of boilerplate.

On performance, lunch bet again? Speaking of which I need to collect on the last one. 😉

djspiewak: No lunch bet on this one.  I'm almost positive the performance will improve with eager columns, but it's not going to jump by as much as you're hoping, and there will be cases where it will be slower just because of memory pressure.

Also the isDefinedAt checks will have to remain (though they'll all be eager in `BitSet` form) since we have heterogeneous data.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -90,27 +90,28 @@ abstract class StdLibSpec extends Qspec {
...
         }
 
         "reject a non-matching string when case sensitive" >> {
-          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str("^bs.*$"), Data.Bool(true), Data.Bool(false)) and
-            ternary(Search(_, _, _).embed, Data.Str("china"), Data.Str("^bs.*$"), Data.Bool(true), Data.Bool(false)) and
-            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*bs.*"), Data.Bool(true), Data.Bool(false))
+          ternary(Search(_, _, _).embed, Data.Str("church"), Data.Str("^bs.*$"), Data.Bool(false), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("china"), Data.Str("^bs.*$"), Data.Bool(false), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*bs.*"), Data.Bool(false), Data.Bool(false)) and
+            ternary(Search(_, _, _).embed, Data.Str("matching"), Data.Str(".*CH.*"), Data.Bool(false), Data.Bool(false))
```
alissapajer: :+1:

alissapajer:
@djspiewak once all the `Search` tests are marked as pending (or realistically, skipped) for the other connectors, then this should be ready for merge. 

Can you also open an issue for each connector that has a broken `Search`?

alissapajer:
Four more integration tests enabled, yay!

PR #2529, opened 2017-07-24
mark tests that fail on next-major branch

rintcius:


djspiewak:
This should be assigned to someone for review/merge.  Also it requires a semantic version label (probably revision).

alissapajer:
@djspiewak we are merging this manually into `next-major` branch, so no version label is required. We have previously discussed versioning `next-major` in parallel with `master` but we haven't done the work required for that.

djspiewak:
Ah right I didn't catch the target upstream.  My bad.

alissapajer:
General status update of this branch thus far.

- #2533 `MapFuncDerived` functions have regressed on this branch (queries work on master with these functions that don't work here).
- #2531 There are various non-mongo related regressions in integration tests. These are likely due to one of these causes:
  1. Changes to `QScript` were incorrectly handled in the connector (`QScript` is correct; bug is in connector)
  2. Changes to `QScript` are incorrect (bug in `QScript`)
  3. The tests were passing by coincidence in master and now that we have fixed `QScript` they fail in the connector (bug is in the connector, even on master).
- There are various mongo planner tests that are not on par with master. A full-stack integration test should be added whenever we need to confirm correctness of the plan. These are failing due to one of these causes:
  1. The mongo planner now produces a different but also correct plan, and so the test expectation needs to be updated. In this case we need to be aware of possible performance regressions.
  2. The mongo planner now produces an incorrect plan. 
- #2532 Integration tests have regressed in new mongo (they work on master).

PR #2548, opened 2017-07-25
mongo 3.4 string functions

### LINKED ISSUE #2388: MongoDB String Operation Expressions
Again are these already supported and it is more about Mapfunc work? If so, please rehome the issue under the appropriate epic/topic.
- [x] $indexOfBytes, 
- [x] $indexOfCP, 
- [x] $split, 
- [x] $strLenBytes, 
- [x] $strLenCP,  
- [x] $strcasecmp, 
- [x] $substrBytes, 
- [x] $substrCP

rintcius:
Adds support for mongo 3.4 specific string functions

Fixes #2388


mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOp3_4.scala
```diff
@@ -0,0 +1,140 @@
...
+import quasar.physical.mongodb.{Bson}
+
+import matryoshka._
+import matryoshka.data.Fix
+import scalaz._, Scalaz._
+
+/** "Pipeline" operators added in MongoDB version 3.4. */
+trait ExprOp3_4F[A]
+object ExprOp3_4F {
+  final case class $indexOfBytesF[A](string: A, substring: A, start: Option[A], end: Option[A]) extends ExprOp3_4F[A]
```
wemrysi: I think something like `bounds: Option[A \/ (A, A)]` would more accurately reflect the possible states.

mongodb/src/main/scala/quasar/physical/mongodb/expression/ExprOp3_4.scala
```diff
@@ -0,0 +1,140 @@
...
+      }
+  }
+
+  implicit def ops[F[_]: Functor](implicit I: ExprOp3_4F :<: F): ExprOpOps.Aux[ExprOp3_4F, F] = new ExprOpOps[ExprOp3_4F] {
+    type OUT[A] = F[A]
+
+    val simplify: AlgebraM[Option, ExprOp3_4F, Fix[F]] = κ(None)
+
+    def bson: Algebra[ExprOp3_4F, Bson] = {
+      case $indexOfBytesF(s, t, u, v) => Bson.Doc("$indexOfBytes" -> Bson.Arr(List(List(s, t), u.toList, v.toList).flatMap(x => x)))
```
wemrysi: `_.flatMap(x => x)` is also available as `_.join`.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -270,6 +285,23 @@ object FuncHandler {
...
+              : MapFuncCore[T, ?] ~> OptionFree[EX, ?] =
+            new (MapFuncCore[T, ?] ~> OptionFree[EX, ?]){
+              implicit def hole[D](d: D): Free[EX, D] = Free.pure(d)
+
+              def apply[A](mfc: MapFuncCore[T, A]): OptionFree[EX, A] = {
+                val fp = new ExprOp3_4F.fixpoint[Free[EX, A], EX](Free.roll)
+                import fp._
+
+                mfc.some collect {
+                  case Length(a1) => $strLenBytes(a1)
```
wemrysi: I think the `*CP` variants are what we want here, I noticed over in #2388 you mentioned some tests were failing. We may need to verify they're encoding the appropriate behavior when it comes to multibyte characters.

rintcius:
As discussed with @wemrysi I'd test a bit more before deciding whether to go for `CP` or `Bytes` variants:

* stdlib tests generally test with ascii only (`def stringDomain = gen.printableAsciiString`) with the exception of Spark which does full-fledged String tests (`def stringDomain = arbitrary[String]`)
*  `$substr` has been deprecated in favour of `$substrBytes` (quote from mongo doc: `Deprecated since version 3.4: $substr is now an alias for $substrBytes.`). We could still _choose_ to make `Substr` map to `$substrCP` instead but that will break compatibility for customers who are currently on mongo 3.4 (though it's not officially supported yet)/ it will change behaviour when upgrading to 3.4.
* basic `Substr` test `ternary(Substring(_, _, _).embed, Data.Str("cafétéria"), Data.Int(3), Data.Int(1), Data.Str("é"))` results in failure for couchbase and Mongodb (agg pipeline) < 3.4. Mimir hasn't implemented `Substr`. Mongodb mapfunc, Marklogic and Spark are successful. As reference, in Posgresql `substr('cafétéria', 4, 1) results in `é` (note index is 4 vs 3)
* basic `Length` test `unary(Length(_).embed, Data.Str("€"), Data.Int(1))` results in failure for couchbase (returns 3). Mongodb agg pipeline < 3.4 and Mimir haven't implemented `Length`. Mongodb mapfunc, Marklogic and Spark are successful. `length('€') in Posgresql results in 1.

Especially mongo mapfunc being character-based makes me think that we should use `CP` indeed (despite that it will change behaviour for `substr` on mongo 3.4 compared to older versions).

PR #2550, opened 2017-07-25
Update a few MongoDB regression test expectations to reflect reality.

wemrysi:
Adjusts a few tests that were too pessimistic for all the MongoDB versions.


rintcius:
👍  will merge once Travis is happy

PR #2552, opened 2017-07-25
Add a 'compile' command to the repl

wemrysi:
Differs from `explain` in that `compile` does not require a backend, useful for debugging LP -> QScript issues.

### DIFF
--- a/repl/src/main/scala/quasar/repl/Repl.scala
+++ b/repl/src/main/scala/quasar/repl/Repl.scala
@@ -18,18 +18,23 @@ package quasar.repl
 
 import slamdata.Predef._
 
-import quasar.{Data, DataCodec, Variables, resolveImports}
+import quasar.{Data, DataCodec, Variables, resolveImports, queryPlan}
 import quasar.common.PhaseResults
+import quasar.connector.CompileM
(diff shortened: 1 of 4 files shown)


repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -233,6 +255,24 @@ object Repl {
...
+      case Compile(q) =>
+        for {
+          state <- RS.get
+          expr  <- DF.unattempt_(sql.fixParser.parse(q).leftMap(_.message))
+          vars  =  Variables.fromMap(state.variables)
+          block <- DF.unattemptT(resolveImports(expr, state.cwd).leftMap(_.message))
+          (log, result) = compileQuery(block, vars, state.cwd)
+          // TODO: This is brittle, relying on particular phase positions in the
+          //       log, we should add more structure to be able to project out
+          //       particular sections directly.
```
alissapajer: Each `PhaseResult` has a `def name: String` so maybe it would be slightly less brittle if we search for the particular name `"Logical Plan"` and then drop everything before that. That way if we add more SQL compiler phases at least this would still do what we expect.

PR #2554, opened 2017-07-26
add split function

### LINKED ISSUE #2455: SPLIT functionality request
Requesting a `split` feature be added to SQL² which allows string conversion to an array based on a pattern (delimiter).

A short example from @jdegoes of what it might look like:
`WHERE foo IN TRIM(SPLIT(:bar, ",")[*])` or something

For reference sake, PostgreSQL uses `split_part` found [here](https://www.postgresql.org/docs/9.1/static/functions-string.html)  `split_part(string text, delimiter text, field int)`

Origin of this request is from a customer who would like to take a String from SlamData created in Markdown and use the individual elements in the String. For example, Markdown may be:

`interests = ____`

The variable `interests` might then contain a String value of `soccer, hockey, tennis`

Ideally we query against that similar to:

```
SELECT ...
FROM ...
WHERE userInterest IN TRIM(SPLIT(:interests,',',[*])
```

### LINKED ISSUE #2618: Split function for mimir
Connector implementation of #2455

rintcius:
Fixes #2455 
Fixes #2618


djspiewak:
This is relatively easy to implement in mimir (there's already a backing-function for it!), so that should probably be done in this PR.

rintcius:
added mimir implementation @djspiewak

frontend/src/main/scala/quasar/std/string.scala
```diff
@@ -241,6 +242,21 @@ trait StringLib extends Library {
     },
     basicUntyper)
 
+  val Split = BinaryFunc(
+    Mapping,
+    "Splits a string into an array of substrings based on a delimiter.",
+    Type.Str,
```
drostron: Should the codomain be a `FlexArr`?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/MapFuncCorePlanner.scala
```diff
@@ -121,6 +121,7 @@ private[qscript] final class MapFuncCorePlanner[
     case ToString(x)                  => SP.asString(x)
     case QSearch(in, ptn, ci)         => fn.matches(in, ptn, Some(if_ (ci) then_ "im".xs else_ "m".xs)).point[F]
     case Substring(s, loc, len)       => lib.safeSubstring[F] apply (s, loc + 1.xqy, len)
+    case Split(s, d)                  => SP.mkArray(fn.tokenize(s, d))
```
drostron: Is array preferred over tokenize's return of seq for MarkLogic?

drostron: Ah nm, but looks like you can enable `marklogic_xml` against `split.test` if you switch to`SP.seqToArray(fn.tokenize(s, d))`.

PR #2556, opened 2017-07-26
Update next-major with recent changes from master

wemrysi:


wemrysi:
@djspiewak @alissapajer This isn't compiling yet due to changes needed in `mimir` to deal with the new bucket representation and indexing in `next-major` for `Reduce` and `Sort` nodes.

Would one of you mind taking a look at this when you have time?

djspiewak:
@wemrysi Can do.  Probably won't get to it before Friday though.  Does that work out?

wemrysi:
@djspiewak Yeah, I just put this up here since I resolved all the other merged conflicts and didn't want that to go to waste, but the `mimir` bit isn't in the way otherwise.

I made a cursory attempt at adding support, but quickly found myself in strange new lands and ran away in fear of becoming lost for too long.

djspiewak:
@wemrysi We can pair on it a bit if you would like (on Friday), so there is some knowledge sharing.  I think it'd be good to get a few more people comfortable with at least the general scope of mimir's internals.

wemrysi:
@djspiewak Sure, that'd be fine.

djspiewak:
@wemrysi We have to explicitly recognize the new distinct pattern, since the way that it is encoded in qscript is now different than it was.  I think the current code for recognizing that pattern (in qscript) is commented as such; it just needs to be updated to the new pattern.

wemrysi:
@djspiewak Ok, I think everything's fixed now, hopefully travis agrees.

rintcius:
@wemrysi Unless I have missed someone fixing this, I don't think any `timeout` on mongo can be removed yet (also not for mongo > 3.0)

There are at least a couple of them removed from `fullOuterEquiJoin` and `selfJoinWithComplexCondition`. Problem is that they only `timeout` every now and then.

wemrysi:
@rintcius Ok, I'll disable them. Hopefully they can be enabled when #2573 is merged.

Issue #2542, opened 2017-07-25
Enable remaining integration tests for mimir


alissapajer:
These are the tests that are _not_ failing due to a known `QScript` (or pre-`QScript`) bug.

Test | Status
--- | ---
caseWithGroup.test | qscript is using arrays to carry context information, rather than objects (#2681)
complex-vars.test | time-related issues
filterByJsWithOnlyReducing.test | probably also due to mongo's polymorphic `min`, `max`, and `concat`. 
flattenNestedObject.test | this actually works, but the time mismatch makes it look like a failure
generatedFieldNames.test | this appears to be an issue with `Guard` when applied to a type which is a large union
locationCount.test | another victim of #2681, combined with an over-complicated query. see #2559 for more details
multilineLike.test | immediately returns count `5` (correct result is `13`)
multipleSelectJoin.test | qscript is using arrays to carry context information, rather than objects (#2681)
projectIndexWithObjectFlatten.test | this actually works, but the time mismatch makes it look like a failure
rightOuterJoin.test | qscript is using arrays to carry context information, rather than objects (#2681)
selectConcat.test | requires `TIMESTAMP` implementation (#2564)
servlet.test |
statesByShortestFirstCity.test | This is due to mongo's polymorphic `min`, `max`, and `concat`. A simplified query shows that `min` returns nothing: `select min(city) from zips`. 
typeOf.test |
undefinedJsInExpr.test | possibly a problem with the test itself, see `FIXME` therein
unsupportedTests.test |
demo/timeSeriesAggregation.test |
joins/changedTestStatusInterim.test |
joins/changedTestStatus.test |
joins/crossJoinWithOneSidedConditions.test | takes a very long time (~100s) and returns a lot of duplicated rows with incorrect columns
joins/groupedJoin.test | see #2121
null/nullTestExprsWithMissing.test | `missing parent` and `missing` are… uh, missing
temporal/convertTimestamp.test |
temporal/datePartsConvertedSimple.test |
temporal/datePartsConverted.test |
temporal/days.test |
temporal/filterOnDate.test |
temporal/timestampsAndIntervals.test |
temporal/timestampRead.test |
temporal/time.test |
temporal/toFromString.test |

alissapajer:
This is ongoing and doesn't have a milestone. Individual issues with milestones will be created to fix the individual queries listed in this issue.

alissapajer:
Closing, as these tests are all tracked by individual issues.

PR #2557, opened 2017-07-26
Mimir fixes and progression

### LINKED ISSUE #2520: Some sort cases don't quite work

djspiewak:
Lots of stuff being addressed here.  Lots and lots of stuff.  Affects #2542.  Fixes #2520.  Makes it easier to uncover some of the other issues.

We're getting closer!


alissapajer:
So many integration tests enabled, awesome!

PR #2560, opened 2017-07-26
Reimplement and replace Table#takeRange

### LINKED ISSUE #2506: Drop in mimir is slow
`select * from zips offset 10 ` returns in ~5 minutes.

`select * from zips` returns in < 1 second.

djspiewak:
Fixes #2506
Depends on #2557 
Affects #2542

Now we have a nice and sane `drop` and `take`, both of which run quite quickly and have the properties you would want.

### DIFF
--- a/it/src/main/resources/tests/basicAnalyticQuery.test
+++ b/it/src/main/resources/tests/basicAnalyticQuery.test
@@ -3,7 +3,7 @@
     "backends": {
         "couchbase": "ignoreFieldOrder",
         "marklogic_json": "ignoreFieldOrder",
-        "mimir":          "skip",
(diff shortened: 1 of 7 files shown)


alissapajer:
@djspiewak the build for this failed here https://travis-ci.org/quasar-analytics/quasar/jobs/258622386#L875 in `quasar.yggdrasil.table.ColumnarTableModuleSpec`, can you take a look?

djspiewak:
@alissapajer I think I have an idea of what's going on.  `drop`/`take` aren't quite working correctly, so these are legit failures.  Will fix.

alissapajer:
Will merge when green.

Issue #2121, opened 2017-04-28
Enable grouped join test for at least two connectors


alissapajer:
The test `joins/groupedJoin.test` is currently disabled for all qscript-based connectors.
```
      "query": "SELECT p.author.login, COUNT(*) as count
                FROM `../slamengine_commits` as p INNER JOIN `../slamengine_commits_dup` as c
                ON p.sha = c.sha
                GROUP BY p.author.login",
```
This is likely due to a yet-undeciphered qscript bug, which this ticket will track. This ticket it to ensure the correctness of the qscript by enabling this test for ideally at least two connectors.

PR #2562, opened 2017-07-27
Reworking of unifySimpleBranches.

alissapajer:


connector/src/main/scala/quasar/qscript/package.scala
```diff
@@ -245,13 +244,19 @@ package object qscript {
     }
   }
 
-  // FIXME naive - use `concat`
-  def naiveConcat4[T[_[_]]: BirecursiveT: EqualT: ShowT, A: Equal: Show](
+  def concat4[T[_[_]]: BirecursiveT: EqualT: ShowT, A: Equal: Show](
```
alissapajer: @wemrysi this PR doesn't seem to make the qscript for the query over in #2121 any better, but maybe there is a different equality case that is being hit in `concat3` or `concat4` that would help.

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -129,127 +129,206 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
     case x                                 => FtoG(QC.inj(x))
   }
 
-  def unifySimpleBranches[F[_], A]
-    (src: A, l: FreeQS, r: FreeQS, combine: JoinFunc)
+  type Remap[A] = JoinFunc => Option[FreeMapA[A]]
+  type Combine[F[_], A, B] = Option[FreeMapA[A]] => Option[F[B]]
```
wemrysi: It seems like this could be `FreeMapA[A] => Option[F[B]]`, is there a reason the domain needs to be `Option[FreeMapA[A]]`? I tend to find functions that accept optional inputs a bit harder to reason about as one has to consider how they behave when given `None`.

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -129,127 +129,206 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
...
+    (implicit
+      QC: QScriptCore :<: F,
+      FI: Injectable.Aux[F, QScriptTotal])
+      : Option[F[A]] = {
+    val branchHole: BranchUnification[F, Hole, A] =
+      unifySimpleBranchesHole(src, left, right)(rebase)(QC, FI)
+    val branchSide: BranchUnification[F, JoinSide, A] =
+      unifySimpleBranchesJoinSide(src, left, right)(rebase)(QC, FI)
+
+    branchHole.combine(branchHole.remap(func)).cata(
```
wemrysi: `branchHole.combine(branchHole.remap(func)) orElse branchSide.combine(branchSide.remap(func))` is a bit more straightforward.

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -129,127 +129,206 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
...
-                  status,
-                  combine >>= {
-                    case LeftSide  => mf1 >> LeftSideF  // references `src`
-                    case RightSide => repair >>= {
-                      case LeftSide  => mf2 >> LeftSideF
-                      case RightSide => RightSideF
-                    }
-                  })).some
-              case _ => None
+                def remap(jf: JoinFunc): Option[JoinFunc] = (jf >>= {
```
wemrysi: Adding a constructor function like `apply[F[_], A, B](remap: Remap[A])(combine: Combine[F, A, B])` would tidy these blocks up a bit as then you could
```scala
BranchUnification {
  case ...
} (func => ...)
```
and avoid the intermediate `def`s.

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -99,28 +99,85 @@ class Transform
...
+
+        def theta(combine: JoinFunc, newLprov: List[prov.Provenance], newRprov: List[prov.Provenance]): F[T[F]] =
+          TJ.inj(ThetaJoin(src, lBranch, rBranch, prov.genComparisons(newLprov, newRprov), JoinType.Inner, combine))
+
+        val (res, newLprov, newRprov, lacc, racc) =
           (lprovs, rprovs) match {
             case (None, None) =>
-              val (combine, lacc, racc) = concat(lval, rval)
+              val (combine, lacc, racc) =
+                c2(uniHole).getOrElse(c2(uniSide).getOrElse {
```
wemrysi: Similarly here, I think `c2(uniHole) orElse c2(uniSide) getOrElse { ... }` is a bit more straightforward.

connector/src/main/scala/quasar/qscript/package.scala
wemrysi: Bummer, yeah, once this is merged I'll see if there is another pattern to identify.

connector/src/main/scala/quasar/qscript/Rewrite.scala
alissapajer: Hm, at some point there was a reason for this :) But looking at all the implementations it appears that it could indeed be `FreeMapA[A] => Option[F[B]]`, so I'll give that a go.

alissapajer:

alissapajer:
@wemrysi just ran all the mongo query integration tests locally, and three more came to life! So I'm going to push that change out now, and then hopefully we can get travis's approval.

alissapajer:
`/joins/joinWithComplexCondition.test` and `/joins/selfjoinWithComplexCondition.test` each took 2 minutes and 28 seconds to pass locally, so hopefully they won't slow down too much on travis. They were most likely much faster on old mongo, and if so, are a good performance improvement target.

 `/joins/fullOuterEquiJoin.test` took only 1.5 seconds.

alissapajer:
Closing this PR and reopening as a new PR to see if the travis mongo build issues (running out of disk space) are resolved.

alissapajer:
Reopened as #2573

PR #2576, opened 2017-07-31
Require callback to initMetaStoreOrStart to handle shutdown

wemrysi:
Fixes an issue where the repl would immediately close all mounts after initialization.

### DIFF
--- a/interface/src/main/scala/quasar/main/package.scala
+++ b/interface/src/main/scala/quasar/main/package.scala
@@ -388,7 +388,7 @@ package object main {
     */
   def initMetaStoreOrStart[C: argonaut.DecodeJson](
     config: CmdLineConfig,
-    start: (C, CoreEff ~> QErrs_TaskM) => MainTask[Unit],
+    start: (C, CoreEff ~> QErrs_TaskM, Task[Unit]) => MainTask[Unit],
     persist: DbConnectionConfig => MainTask[Unit]
   )(implicit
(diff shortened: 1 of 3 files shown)


jedesah:
I can see that this change hands over the responsibility of calling the Quasar shutdown `Task` to the client code. That's less desirable of course, but I imagine there is a good reason, but a glance at the diff does not immediately inform me as to why that's necessary. @wemrysi Would you mind just briefly walking me through what's happening to cause the error @djspiewak  mentioned? Based on the error message, I imagine that shutdown was being called too early. Ah! I think I just figured it out, I guess the "Start the Repl" `Task` is async so we call shutdown before the repl is actually done. I guess it would be preferable to make the start Repl task synchronous but perhaps that is tricky so maybe this is a reasonable compromise to get things working

wemrysi:
Yeah, the previous impl assumed the `Task` returned by the callback to `initMetaStoreOrStart` blocked, which isn't always true.

jedesah:
I feel like we should use `Task.async` to build the Repl `Task` so that it is blocking. That way we wouldn't have to delegate shutdown to the client code. Also, I feel like there are lot's of places in the codebase where we assume that `onFinish` behaves appropriately. I think the fact that `ensuring` is not working as expected in this case is because we are not properly building the `Task`. @wemrysi What do you think?

wemrysi:
Closing in favor of #2578.

PR #2577, opened 2017-08-01
Use indexes in Marklogic single predicate filters whenever possible

### LINKED ISSUE #2425: Plan single predicate filters using cts:query when index available

jsantos17:
Fixes #2425 

Mostly done, but still TODO:

- [x] Unit/property tests for `FilterPlanner.{StarIndexPlanner, PathIndexPlanner, ElementIndexPlanner}`
- [x] Post-filter on `ElementIndexPlanner`?
- [x] Remove calls to pathy's `prettyprint`
- [x] Integration tests


jsantos17:
@wemrysi Hi! Can you please start reviewing this? I don't foresee many new changes other than tests and the post-filter (complete TODO in PR description).

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,202 @@
...
+    private def planElementIndex[Q](src: Search[Q], fm: FreeMap[T])(
+      implicit Q: Corecursive.Aux[Q, Query[T[EJson], ?]]
+    ): Option[Search[Q]] = rewrite(fm) match {
+      case PathProjection(op, QNamePath(qname), const) => {
+        val q = Query.ElementRange[T[EJson], Q](IList(qname), op, IList(const)).embed
+        val src0 = Search.query.modify((qr: Q) => Q.embed(Query.And(IList(qr, q))))(src)
+
+        Search.pred.modify {
+          case Some(pred) => None
+          case None       => (child.* `/` child.elementNamed(QName.string(qname))).some
```
jsantos17: @wemrysi I'm not sure if this is the correct unabbreviated XQuery. This renders as `*/child::element(someQName)`, which indeeds finds elements named `someQName`  nested within the root, but not those with a `someQName` that is not a direct child of the root element. However this is, this is different from `./*/child::element(someQName)`, which was your suggestion and seems to have the same behavior as my impl. Is there are a difference?

wemrysi: They are equivalent, the `.` just refers to the context node explicitly, but it isn't required in this case.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FormatFilterPlanner.scala
```diff
@@ -0,0 +1,85 @@
...
+      implicit Q: Birecursive.Aux[Q, Query[T[EJson], ?]],
+               SP: StructuralPlanner[F, FMT]
+    ): F[Option[Search[Q]]] = {
+      val planner = new FilterPlanner[F, FMT, T]
+
+      lazy val pathQuery    = planner.validSearch(planner.PathIndexPlanner(src, f))
+      lazy val starQuery    = planner.validSearch(planner.StarIndexPlanner(src, f))
+      lazy val elementQuery = planner.validSearch(planner.ElementIndexPlanner(src, f))
+
+      (pathQuery |@| starQuery |@| elementQuery) {
```
wemrysi: If `validSearch` returned `OptionT[F, Search[Q]]` this block could be replaced with
```scala
(pathQuery ||| starQuery ||| elementQuery).run
```

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FormatFilterPlanner.scala
```diff
@@ -0,0 +1,85 @@
...
+      T[_[_]]: BirecursiveT: ShowT, Q](src: Search[Q], f: FreeMap[T])(
+      implicit Q: Birecursive.Aux[Q, Query[T[EJson], ?]],
+               SP: StructuralPlanner[F, FMT]
+    ): F[Option[Search[Q]]] = {
+      val planner = new FilterPlanner[F, FMT, T]
+
+      lazy val pathQuery    = planner.validSearch(planner.PathIndexPlanner(src, f))
+      lazy val elementQuery = planner.validSearch(planner.ElementIndexPlanner(src, f))
+
+      (pathQuery |@| elementQuery) {
```
wemrysi: Same here.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,202 @@
...
+    def unapply[T[_[_]]](fpm: FreePathMap[T]): Option[(ComparisonOp, ADir, T[EJson])] = fpm match {
+      case Embed(CoEnv(\/-(MFPath(MFComp(op, Embed(CoEnv(\/-(PathProject(pp)))), Embed(CoEnv(\/-(MFPath(MFCore.Constant(v)))))))))) =>
+        (op, pp.path, v).some
+      case _ => none
+    }
+  }
+
+  private object QNamePath {
+    def unapply(path: ADir): Option[QName] = {
+      if(depth(path) === 1) {
```
wemrysi: `Pathy.dirName` will get you the name of the last segment, if one exists.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  /* Discards nested projection guards. The existence of a path range index a/b/c
+   * guarantees that the nested projection a/b/c is valid. */
+  private def rewrite(fm: FreeMap[T]): FreePathMap[T] =
+    ProjectPath.elideGuards(ProjectPath.foldProjectField(fm))
+
+  object StarIndexPlanner {
+    private def planPathStarIndex[Q](src: Search[Q], fm: FreeMap[T])(
```
jsantos17: There was a good reason sometime, but refactoring eliminated the contents of the `def apply` apparently. Thanks for catching this, fixing.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,202 @@
...
+
+    private def planElementIndex[Q](src: Search[Q], fm: FreeMap[T])(
+      implicit Q: Corecursive.Aux[Q, Query[T[EJson], ?]]
+    ): Option[Search[Q]] = rewrite(fm) match {
+      case PathProjection(op, QNamePath(qname), const) => {
+        val q = Query.ElementRange[T[EJson], Q](IList(qname), op, IList(const)).embed
+        val src0 = Search.query.modify((qr: Q) => Q.embed(Query.And(IList(qr, q))))(src)
+
+        Search.pred.modify {
+          case Some(pred) => None
```
jsantos17: Oh I didn't know that! Will remove this restriction now! Thanks.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FormatFilterPlanner.scala
```diff
@@ -0,0 +1,85 @@
...
+import simulacrum.typeclass
+import matryoshka._
+import scalaz._, Scalaz._
+
+@typeclass
+trait FormatFilterPlanner[A] {
+  def plan[
+    F[_]: Monad: QNameGenerator: PrologW: MonadPlanErr: Xcc,
+    FMT: SearchOptions,
+    T[_[_]]: BirecursiveT: ShowT, Q](src: Search[Q], f: FreeMap[T])(
```
jsantos17: Left-over debugging code. Will delete.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,202 @@
...
+import xml.name._
+
+import scalaz._, Scalaz._
+
+private[qscript] final class FilterPlanner[
+  F[_]: Monad: QNameGenerator: PrologW: MonadPlanErr: Xcc,
+  FMT: SearchOptions,
+  T[_[_]]: BirecursiveT: ShowT
+](implicit SP: StructuralPlanner[F, FMT]) {
+  def plan[Q](src0: Search[Q] \/ XQuery, f: FreeMap[T])(
```
jsantos17: Indeed! Definitely seems like a better organization. Thanks.

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FilterPlanner.scala
```diff
@@ -0,0 +1,202 @@
...
+    private def planElementIndex[Q](src: Search[Q], fm: FreeMap[T])(
+      implicit Q: Corecursive.Aux[Q, Query[T[EJson], ?]]
+    ): Option[Search[Q]] = rewrite(fm) match {
+      case PathProjection(op, QNamePath(qname), const) => {
+        val q = Query.ElementRange[T[EJson], Q](IList(qname), op, IList(const)).embed
+        val src0 = Search.query.modify((qr: Q) => Q.embed(Query.And(IList(qr, q))))(src)
+
+        Search.pred.modify {
+          case Some(pred) => None
+          case None       => (child.* `/` child.elementNamed(QName.string(qname))).some
```
jsantos17: The latter. We definitely want it to be different for JSON. Good catch!

jsantos17:
@wemrysi this is ready for another look. I'm working to remove `FormatFilterPlanner` and implement it inside the companion object of `FilterPlanner` to remove the cyclic dependency. Any other observations?

marklogic/src/main/scala/quasar/physical/marklogic/qscript/FormatFilterPlanner.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  implicit val xmlFilterPlanner: FormatFilterPlanner[DocType.Xml] = new FormatFilterPlanner[DocType.Xml] {
+    def plan[F[_]: Monad: QNameGenerator: PrologW: MonadPlanErr: Xcc,
+      FMT: SearchOptions: StructuralPlanner[F, ?],
+      T[_[_]]: BirecursiveT, Q](src: Search[Q], f: FreeMap[T])(
+      implicit Q: Birecursive.Aux[Q, Query[T[EJson], ?]]
+    ): F[Option[Search[Q]]] = {
+      val planner = new FilterPlanner[T]
+
+      lazy val starQuery    = validSearch[T, F, FMT, Q](planner.StarIndexPlanner(src, f))
```
jsantos17: Removed `pathQuery` from the XML filter planner. It leads to some really unintuitive behavior if enabled with reasonable path indexes. Assume a directory `/someDir` with files of the following form:

```xml
<rooted>
  <name>Some name</name>
<rooted>
```

XML planning of `pathQuery` succeeds for a query like `select * from /someDir where name = "whatever"` when a path range index exists on `/name`. If I were a user, I'd expect that query to filter on `/rooted/name` elements, but it does not work that way. The index exists on `/name` and not on `/rooted/named`. The result is that planning succeeds, the query runs successfully and an empty result set is returned.

jsantos17:
@wemrysi Oh the post-filter! Quick question on that, the post filter should be as simple as wrapping the resulting `Search` in the old XQuery filter impl, right? Or is there anything else I should know?

wemrysi:
@jsantos17 Yep, I think that should do it.

jsantos17:
@wemrysi ready for another look. `FormatFilterPlanner` has been removed and we now post-filter plans with element range queries :)

marklogic/src/main/scala/quasar/physical/marklogic/qscript/EJsonPlanner.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import quasar.Data
+import quasar.ejson._
+import quasar.physical.marklogic.xquery._
+
+import matryoshka._
+import matryoshka.implicits._
+import scalaz._
+
+object EJsonPlanner {
```
wemrysi: This should be used to plan `MapFuncCore.Constant` as well.

jsantos17:
@wemrysi sounds reasonable. I'll add two small integration tests like you mentioned, one for each format.

it/src/test/scala/quasar/physical/marklogic/qscript/FilterPlannerSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+  type M[A] = MarkLogicPlanErrT[PrologT[StateT[Free[XccEvalEff, ?], Long, ?], ?], A]
+  type G[A] = WriterT[Id, Prologs, A]
+
+  xccSpec(backendName => s"Filter Planner for ${backendName.name}") { (evalPlan, evalXQuery) =>
+    "does not plan with indexes if not available" >> {
+      evalPlan(filterExpr("doesNotExist")) must beSome((_: Search[U] \/ XQuery).isRight)
+    }
+
+    "uses an element range index when it exists" >> {
+      val idxName = "someIndex"
```
jsantos17: @wemrysi any ideas on how we can make this function safer? I wanted to use transactions to clean up after the test, but it looks like [transactions work differently with `admin` functions](https://docs.marklogic.com/guide/admin-api/usingAPI#id_71075). The same docs suggest you make your admin functions idempotent and I'm working to do that, but we still need to ensure the test path index is deleted after the test.

jsantos17:
@wemrysi Ready for another look. Any suggestions on how to improve [this test](https://github.com/quasar-analytics/quasar/pull/2577/files#diff-6cd1910630e48182c8fa97289a3a2ee8R61) are very welcome. It should be cleaning reliably now, so not sure if we need to make the index creation/removal functions idempotent anymore.

it/src/test/scala/quasar/physical/marklogic/qscript/FilterPlannerSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+
+      val test = (createIndex >> planFilter).onFinish(_ => cleanup)
+
+      test.unsafePerformSync must beSome.which(includesPathRange(idxName, _))
+    }
+  }
+
+  private def includesPathRange(idxName: String, q: Search[U] \/ XQuery): Boolean = {
+    val alg: AlgebraM[Option, Query[Fix[EJson], ?], U] = {
+      case node @ Query.PathRange(_, _, _) => None
```
wemrysi: Could we match on the index name here?

it/src/test/scala/quasar/physical/marklogic/qscript/FilterPlannerSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+        reflNT[Task] :+: monoSeq :+: Read.constant[Task, Session](sess) :+: Read.constant[Task, ContentSource](cs)
+      val eval: Free[XccEvalEff, ?] ~> Task = foldMapNT(xccToTask)
+
+      eval(f)
+    }
+
+  private def planQs(cs: ContentSource, qs: Fix[QSR]): Task[Option[Search[U] \/ XQuery]] = {
+    def planQ[F[_]: Monad: QNameGenerator: PrologW: MonadPlanErr: Xcc](qs: Fix[QSR]
+    ): F[Search[U] \/ XQuery] =
+      qs.cataM(Planner[F, DocType.Json, QSR, Fix[EJson]].plan[U])
```
wemrysi: We'd like to be able to test both XML and JSON, especially since they use different implementations.

jsantos17:
@wemrysi ready for another look. I test for the codepath that yields filters that use path range indexes. Unfortunately, I'm not sure how to test the other two codepaths (i.e element range query with post-filter), since the post-filter squashes the `Search` into `XQuery`.

Also, yes, the clean up works reliably. Even with exceptions and whatnot I have never had left over test indexes.

wemrysi:
@jsantos17 One last thing, this needs a semantic version label in order to merge.

wemrysi:
@jsantos17 I think this is actually a `revision` as it doesn't grow any public facing APIs (SQL^2 or the HTTP interface).

jsantos17:
@wemrysi oh right. This doesn't change any APIs. Changed it now. Thanks :)

PR #2578, opened 2017-08-01
Fix Repl from crashing immediately

jedesah:
The repl `Task` was not being created properly causing violated assumptions making a recent change break the Repl as the Quasar FileSystem was being shutdown right after the Repl was created. Now properly using `Task.async` constructor designed specifically for callback apis such as `AeshConsole`.

### DIFF
--- a/repl/src/main/scala/quasar/repl/Main.scala
+++ b/repl/src/main/scala/quasar/repl/Main.scala
@@ -50,18 +50,14 @@ object Main {
   type DriverEffM[A] = Free[DriverEff, A]
 
   private def driver(f: Command => Free[DriverEff, Unit]): Task[Unit] = {
-    def shutdownConsole(c: Console): Task[Unit] =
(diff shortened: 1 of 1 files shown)


jedesah:
This strikes me as a more desirable fix to the Repl problem as I commented on #2576 and since I think @wemrysi is out for the night, wanted to open this PR now in case @djspiewak you want to review and merge this before then since sounds like you are eager to get the Repl working again.

The `flatMap(shutdown)` part kind of bugs me because I am not completely convinced that that captures all failure scenarios, but I can't think of a better way to do it. I definitely think that using the `async` is the way to go though, perhaps @djspiewak you can way in on whether this does indeed capture all failure scenarios or whether there is an alternative way of doing that.

djspiewak:
The caveat with everything `onFinish` is that preemption isn't captured, but it's basically ok here.

Issue #2579, opened 2017-08-01
Improvements to spark-elastic


rabbitonweb:
- [x]  configuration allowing to connect to any host:port
- [x]  configuration allowing to specify spark's cluster parameters (like in spark-hdfs)
~~- [ ] Deal with illegal chars in ElasticSearch (white spaces, leading _ etc.)~~ (moved to https://github.com/quasar-analytics/quasar/issues/2719)
~~- [ ] Integration tests 100% passing~~ (moved to https://github.com/quasar-analytics/quasar/issues/2717)

PR #2580, opened 2017-08-01
Spark elastic improvements part 1

rabbitonweb:
Part of #2579


rabbitonweb:
@drostron This is ready for review

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/package.scala
```diff
@@ -93,8 +93,16 @@ package object elastic {
...
-    // TODO_ES missng ES nodes and port
-    sparkConfOrErr.map(sparkConf => (sparkConf, SparkFSConf(sparkConf)))
+
+    def fetchParameter(name: String): DefinitionError \/ String = uriOrErr.flatMap(uri =>
+      uri.params.get(name).fold(liftErr(s"'$name' parameter not provided").left[String])(_.right[DefinitionError])
+    )
+
+    for {
+      sparkConf   <- sparkConfOrErr
+      elasticHost <- fetchParameter("elasticHost")
```
drostron: Could use `ValidationNel` along with its `Applicative` to offer an inclusive error when both are absent rather than an error for just the first.

sparkcore/src/test/scala/quasar/physical/sparkcore/fs/elastic/ElasticCallSpec.scala
```diff
@@ -43,27 +43,27 @@ class ElasticCallSpec extends quasar.Qspec
...
       val checkProgram = for {
-        bar2Exists <- elastic.typeExists(IndexType(toIndex, "bar2"))
-        bazExists <- elastic.typeExists(IndexType(toIndex, "baz"))
+        bar2Exists <- elastic.typeExists(IndexType("foo2", "bar2"))
+        bazExists <- elastic.typeExists(IndexType("foo2", "baz"))
       } yield (bar2Exists, bazExists)
 
-      program.foldMap(ElasticCall.interpreter).unsafePerformSync
+      execute(program)
       java.lang.Thread.sleep(2000) // test needs to wait for the views to be updated
```
drostron: Preexisting, but this seems brittle. Was there discussion of a poll with a timeout along with an associated error?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/package.scala
rabbitonweb: Ok, good point!

rabbitonweb: No, wait. I must disagree :) The `\/` is implied by how it is being used. And that is actually what `BackendDef` requires. 

`BackendDef` wraps over `FsCfg => Option[DefErrT[F, DefinitionResult[F]]]` where `DefErrT` is `EitherT[F, DefinitionError, A]`. 

`DefinitionError` is a `NonEmptyList[String] \/ EnvironmentError` so maybe if we could provide `Semigroup` for `DefinitionError` then this would become a trival task. 

Regardless I  believe this is out of scope of this PR (imporoving ES connector). Aggregating configuration errors as `ValidationNel` rather then `\/` is common to all Spark based connectors and should be treated separately I think. Maybe as part of https://github.com/quasar-analytics/quasar/issues/2584 or a separate issue of its own. What do you think @drostron ?

drostron: If it seems like a good idea more generally across Spark, sure and I agree that shouldn't take place here. For the near term I was just suggesting something like the following.

```scala
for {
  sparkConf    <- sparkConfOrErr
  hostPort     <- (fetchParameter("elasticHost") ⊛ fetchParameter("elasticPort")).tupled
  (host, port) =  hostPort
} yield (sparkConf, SparkFSConf(sparkConf, host, port.toInt))
```

rabbitonweb: oh.... i thought u meant the whole for-comprehension. ok, i will update accordingly

rabbitonweb:
I've fixed what @drostron asked for but he is OOO for the whole week thus changing reviewer to @jsantos17

drostron:
@rabbitonweb @jsantos17 I have time to take a look.

drostron:
Seems suspect that there's a `org.elasticsearch.client.ResponseException` on the `mongo_2_6` job.

rabbitonweb:
@drostron this was hopefully fixed in https://github.com/quasar-analytics/quasar/pull/2600 thus chances are we will not see it again

drostron:
@rabbitonweb Ah, 👌🏽

rabbitonweb:
@drostron this is fixed with merge

Issue #1634, opened 2016-10-24
Replace "input" functions in spark-core with algebras so connectors can write interpreters for them


rabbitonweb:
For example `readfile` in spark-core defines `Input` as 

```
  final case class Input[S[_]](
    rddFrom: (AFile, Offset, Limit)  => Free[S, RDD[String]],
    fileExists: AFile => Free[S, Boolean],
    readChunkSize: () => Int
  )
```

`readfile` uses that `Input` while connectors implement methods of this class.

Better approach would be to define algebra that will represent `rddFrom`, `fileExists` & `readChunkSize` and let connectors write interpreter for that algebra.

rabbitonweb:
@zeeiyerWork u pushed it to September2017 but I'm brining it back to August2017 because I will not do `BackendModule` without this being fixed

PR #2585, opened 2017-08-02
Spark details algebra - queryfile.Input / def fileExists

rabbitonweb:
This is first PR to https://github.com/quasar-analytics/quasar/issues/1634

Here we try to eliminate `queryfile.Input` / `fileExists` method and replace it with algebra (SparkConnectorDetails) to which connectors have to provide interpreter.

Before I continue with any other functions from `queryfile.Input` I want to have this reviewed - confirmation that my approach is the right one.

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/queryfile.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/queryfile.scala
@@ -25,6 +25,7 @@ import quasar.fs.FileSystemErrT
 import quasar.fp.free._
 import quasar.fp.ski._
 import quasar.contrib.pathy._
+import quasar.physical.sparkcore.fs.{SparkConnectorDetails, FileExists}
 
 import org.elasticsearch.spark._
 import org.apache.spark._
@@ -89,5 +90,15 @@ object queryfile {
     s0: Task :<: S,
     elastic: ElasticCall :<: S
   ): Input[S] =
-    Input[S](fromFile _, store[S] _, fileExists[S] _, listContents[S] _, readChunkSize _)
+    Input[S](fromFile _, store[S] _, (f: AFile) => false.point[Free[S, ?]], listContents[S] _, readChunkSize _)
+
+  def detailsInterpreter[S[_], G[_] : Monad](int: S ~> G)(implicit
+    E: ElasticCall.Ops[S]
+  ): SparkConnectorDetails ~> G =
+    new (SparkConnectorDetails ~> G) {
+      def apply[A](from: SparkConnectorDetails[A]) = from match {
+        case FileExists(f) => E.typeExists(file2ES(f)).foldMap(int)
+      }
+    }
+
 }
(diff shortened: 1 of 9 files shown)


rabbitonweb:
My main concern are things like https://github.com/quasar-analytics/quasar/pull/2585/files#diff-dfcc19cada2ced73e9ee1a179cd4d023R95

Originally I had a NT `SparkConnectorDetails ~> Free[S, ?]` but then in `package.scala` I had to go from `Eff ~> Task` , that's why I changed signature to `SparkConnectorDetails ~> G` where we provide `int: S ~> G`. If there are better ways to do it, I'm all ears

wemrysi:
@rabbitonweb I would use the most specific effect possible in your interpreters, if you need to lift them into some other effect you can always compose them with a natural transformation that does so.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/queryfile.scala
```diff
@@ -89,5 +90,15 @@ object queryfile {
     s0: Task :<: S,
     elastic: ElasticCall :<: S
   ): Input[S] =
-    Input[S](fromFile _, store[S] _, fileExists[S] _, listContents[S] _, readChunkSize _)
+    Input[S](fromFile _, store[S] _, (f: AFile) => false.point[Free[S, ?]], listContents[S] _, readChunkSize _)
+
+  def detailsInterpreter[S[_], G[_] : Monad](int: S ~> G)(implicit
```
wemrysi: Here you could interpret into `Free[S, ?] where ElasticCall :<: S`. When you choose how to interpret `ElasticCall`, you can then compose that interpreter with this one.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/hdfs/queryfile.scala
```diff
@@ -107,8 +108,22 @@ class queryfile(fileSystem: Task[FileSystem]) {
 
   def input[S[_]](implicit s0: Task :<: S): Input[S] =
     Input[S](fromFile _, store[S] _, fileExists[S] _, listContents[S] _, readChunkSize _)
+
+
 }
 
 object queryfile {
+
+  def detailsInterpreter[S[_], G[_] : Monad](fileSystem: Task[FileSystem], int: S ~> G)(implicit
```
wemrysi: Here it looks like you could interpret into any `F[_]: Capture` (from the `effect` project) since you're not doing anything Task specific.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/queryfile.scala
```diff
@@ -79,4 +80,14 @@ object queryfile {
   def input[S[_]](implicit
     S: Task :<: S
   ): Input[S] = Input[S](fromFile _, store[S] _, fileExists[S] _, listContents[S] _, readChunkSize _)
+
+  def detailsInterpreter[S[_], G[_] : Monad](int: S ~> G)(implicit
```
wemrysi: Another candidate for `F[_]: Capture`.

PR #2588, opened 2017-08-02
WIP 2502 mongo now

rintcius:
A not so ideal implementation of `now()` in agg. pipeline.

Just putting this up for discussion. 

There _is_ a `bson` representation that is a backend managed `now()` - `Bson.JavaScript(Js.New(Js.Call(Js.Ident("Date"), List())))` - but its usage is quite limited. E.g. `select fieldA, now() from tableA` does not work, nor does `select ts - now(), now() - ts from days` 

I also tried using `$let` tricks such as below, but this has the same limitations.

```
val jsNewDate = Bson.JavaScript(quasar.javascript.Js.New(quasar.javascript.Js.Call(quasar.javascript.Js.Ident("Date"), List())))
Bson.Doc("$let" -> Bson.Doc(
       "vars" -> Bson.Doc("now" -> jsNewDate),
       "in"   -> Bson.Text("$$now"))
)
```

I see another backend-managed `now` solution that _may_ work, but it's a long shot. Not sure if it's worth trying... In a nutshell: we can try **writing** (ugh) `now()` via https://docs.mongodb.com/manual/reference/operator/update/currentDate/ and then looking up that value.

Alternatively, we can calculate `now` from quasar and then pass that to mongo, but this has the problem of clock skew. Anyway, this PR does that in a quick and dirty way (with some additional problems that 1) it's not referential transparent 2) may be different for each occurrence of `now()`). 
I am hoping someone sees a way that doesn't involve too much refactoring to provide this value for `now` once, so that we can solve issue 1) and 2) (adding some kind of evaluation context maybe?).

### DIFF
--- a/it/src/main/resources/tests/temporal/nowComparison.test
+++ b/it/src/main/resources/tests/temporal/nowComparison.test
@@ -1,0 +1,24 @@
+{
+  "name": "use now function (comparison)",
+
+  "backends": {
+  },
+
+  "data": "../days.data",
+
+  "query": "select now() > ts as past, ts > now() as future from `../days`",
+
+  "predicate": "exactly",
+
+  "expected": [
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false },
+    { "past": true, "future": false }
+  ]
+}
--- a/it/src/main/resources/tests/temporal/nowSubtract.test
+++ b/it/src/main/resources/tests/temporal/nowSubtract.test
@@ -1,0 +1,24 @@
+{
+  "name": "use now function (subtract)",
+
+  "backends": {
+  },
+
+  "data": "../days.data",
+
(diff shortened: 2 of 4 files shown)


jdegoes:
I don't think Javascript is the right approach. Won't that cause a query that uses `NOW()` to be run as map/reduce? Javascript can't run in the agg pipeline. Or at least couldn't in older versions. Also it's important to have the guarantee that `NOW()` in one part of a query matches `NOW()` in another part — i.e. referentially transparency. Better to spend a little extra time developing machinery to just propagate the "query start time" date/time into the QScript or whatever is required to solve this without relying on MongoDB. Backends that provide good support for `NOW()` can use a different approach.

rintcius:
@jdegoes - thanks! Somehow my mind was glued to solving it in mongo connector, but I didn't think of solving this in QScript. Now, my mind is unglued somewhat, I'll try moving `Now` from `MapFuncCore` to `MapFuncDerived` and give it a default implementation in  QScript.

BTW the Javascript I was talking about above **does** avoid map/reduce - it is a part of `bson` that I stumbled upon, see type number 13 here: https://docs.mongodb.com/manual/reference/bson-types so it generates an agg pipeline query, but yeah runtime it results usually in an error (though not always).

PR #2591, opened 2017-08-03
Implement ToString for mongo aggregation pipeline

### LINKED ISSUE #2453: Leverage $substr to convert from numbers to strings, and avoid map/reduce
Use this hack -> 
$substr:["$foo", 0, -1 ] will convert a foo number to a string in older versions of MongoDB. This is an undocumented hack.

rintcius:
This implementation takes advantage of an undocumented hack $substr(x, 0, -1) to convert x to a string

Fixes #2453


mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
```diff
@@ -318,6 +329,18 @@ object FuncHandler {
                         $substrCP(a1, a2, a3)
                       )
                     )
+                  case ToString(a1) =>
+                    // NB: Identical to the one in Core, but uses $substrBytes
```
rintcius: Note: it should be possible to deduplicate the common code here, but I refrained from doing that because that'll become ugly too (see how `trunc` is incorporated in `FuncHandler` trait - for `trunc` the extra complexity is warranted because `trunc` is used multiple times). 

Instead I went for the same solution as the code [here](https://github.com/rintcius/quasar/blob/c764bd703863339cb84779c50f2c321061f59073/mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala#L288), which is more similar to the `ToString` use-case

jdegoes:
In later version of MongoDB, this functionality natively exists, and we should use the native version rather than the undocumented hack.

rintcius:
I checked with @jdegoes - there's no `tostring` functionality in mongo agg. pipeline yet, so need to use the undocumented hack for all versions.

mongodb/src/main/scala/quasar/physical/mongodb/planner/FuncHandler.scala
alissapajer: Is it not possible to create this function:
```scala
// makeToString(a1, $substrBytes)
// makeToString(a1, $substr)
def makeToString(a1: A, func: ???) = {
  $cond(
    $eq(a1, $literal(Bson.Null)),
    $literal(Bson.Text("null")),
    $cond(
      $eq(a1, $literal(Bson.Bool(false))),
      $literal(Bson.Text("false")),
      $cond(
        $eq(a1, $literal(Bson.Bool(true))),
        $literal(Bson.Text("true")),
        func(a1, $literal(Bson.Int32(0)), $literal(Bson.Int32(-1))))))
}
```
I don't know what all the types are, but maybe we could do that?

rintcius: Ooh, it _**is**_ possible, even quite nicely so... I was just not looking in the right place before. It's now placed inside that magical `ExprOpCore.fixpoint` and all types work out great! :)

alissapajer:
@rintcius is there any similar deduplication that can be done around the `trunc` implementation?

rintcius:
@alissapajer I definitely have in mind to try to cleanup `trunc` too, but I prefer to do that after `next-major` and `master` are back together.

Issue #2598, opened 2017-08-04
test failure in ElasticCallSpec


alissapajer:
See in travis on the `mongodb_2_6` build.

Failure in `quasar.physical.sparkcore.fs.elastic.ElasticCallSpec`.

```
[error] ! delete existing type from existsing index (3 seconds, 334 ms)
[error]  org.elasticsearch.client.ResponseException: GET http://localhost:9200/_cat/indices?v&format=json&bytes=b: HTTP/1.1 503 Service Unavailable
[error]  {"error":{"root_cause":[{"type":"cluster_block_exception","reason":"blocked by: [SERVICE_UNAVAILABLE/1/state not recovered / initialized];"}],"type":"cluster_block_exception","reason":"blocked by: [SERVICE_UNAVAILABLE/1/state not recovered / initialized];"},"status":503} (RestClient.java:354)
[error] org.elasticsearch.client.RestClient$1.completed(RestClient.java:354)
[error] org.elasticsearch.client.RestClient$1.completed(RestClient.java:343)
[error] org.apache.http.concurrent.BasicFuture.completed(BasicFuture.java:119)
[error] org.apache.http.impl.nio.client.DefaultClientExchangeHandlerImpl.responseCompleted(DefaultClientExchangeHandlerImpl.java:177)
[error] org.apache.http.nio.protocol.HttpAsyncRequestExecutor.processResponse(HttpAsyncRequestExecutor.java:436)
[error] org.apache.http.nio.protocol.HttpAsyncRequestExecutor.inputReady(HttpAsyncRequestExecutor.java:326)
[error] org.apache.http.impl.nio.DefaultNHttpClientConnection.consumeInput(DefaultNHttpClientConnection.java:265)
[error] org.apache.http.impl.nio.client.InternalIODispatch.onInputReady(InternalIODispatch.java:81)
[error] org.apache.http.impl.nio.client.InternalIODispatch.onInputReady(InternalIODispatch.java:39)
[error] org.apache.http.impl.nio.reactor.AbstractIODispatch.inputReady(AbstractIODispatch.java:114)
[error] org.apache.http.impl.nio.reactor.BaseIOReactor.readable(BaseIOReactor.java:162)
[error] org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvent(AbstractIOReactor.java:337)
[error] org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvents(AbstractIOReactor.java:315)
[...]

PR #2600, opened 2017-08-07
Run healthchecks few times before running test

rabbitonweb:
Approach to resolve https://github.com/quasar-analytics/quasar/issues/2598

@alissapajer : the issue (that I tried to reproduce locally and failed multiple times) seems to happen when tests are running while embedded Elastic Search is not yet properly instantiated. This PR gives it a "5 times retry", each separated by 200 mills sleep. Lets try to merge that and see if it resolves the issue on CI. What do you think?


alissapajer:
@rabbitonweb sounds good, let's give this a go.

PR #2602, opened 2017-08-07
A handful of QScript fixes/improvements

wemrysi:
Improves or fixes the following in QScript
* Avoids the use of universal equality in `MapFuncCore.extractGuards`.
* When guards are extracted from a `ThetaJoin` condition into a filter, they are also elided from the combiner.
* Fixed duplicate detection in `concat`.
* Improved equality for `Provenance.Both` by eliding `Nada` and comparing a tree of `Both`s as sets instead of structurally.


alissapajer:
@wemrysi looks like `different-flattens.test` is working now for a few more connectors, but `union.test` started failing for at least mimir.

Issue #2216, opened 2017-05-18
Move SparkCore to BackendModule


rabbitonweb:
- [x] spark-local
- [x] spark-hdfs
~~- [ ] spark-elastic~~ (moved to https://github.com/quasar-analytics/quasar/issues/2715)
~~- [ ]  spark-cassandra~~ (moved to https://github.com/quasar-analytics/quasar/issues/2716)

zeeiyerWork:
Closing this because it is a duplicate of #2096

Issue #2105, opened 2017-04-26
Implement dynamic loading of backends


djspiewak:
**interface** doesn't really need a static dependency on each of the connectors.  Once each connector is abstracted behind `BackendModule`, it becomes very easy to just runtime load connectors based on configuration state.  This also opens the door to cool things like classloader isolation to allow individual connectors to have different dependency versions (or even different Scala versions!) without affecting quasar as a whole.

- [ ] Mimir
- [ ] Couchbase
- [ ] Marklogic
  - Depends on #2092 
- [ ] Spark
  - Depends on #2096
- [ ] Postgres
  - Depends on #2095
- [ ] (New) Mongo
  - Depends on #2094

Issue #2378, opened 2017-06-22
Implement Multi-Version Plumbing


zeeiyerWork:
The idea is to universally implement for the mongo connector - a cascading version implementation. i.e. fall down from latest Mongo version to older Mongo versions to base implementation for a given operator.
It presumably could be applied to other connectors if we are convinced that it works satisfactorily for Mongo.

sellout:
See #2375 and #2376 for the concrete Mongo issues. I’d say this ticket should basically cover the discussion as to whether this is the right approach for all connectors, or if there’s something else we should be doing.

jdegoes:
I was originally going to handle this (way back in 2014) as some sort of explicit lazy data structure, e.g.:

```scala
case class Versioned[A](value: Map[MongoVersion. A])
```

Idea being you can produce in parallel (possibly lazily) different plans in all the junctions that require them. "Selective versioning" since many nodes don't require that variation.

rintcius:
ping @zeeiyerWork - I removed this from `epic: Connector: New Mongo - Initial Production Rollout`

zeeiyerWork:
@rintcius  - why? I thought it was a pre-requisite to the way you were implementing some of the other issues?

rintcius:
@zeeiyerWork Based on @sellout's comment above regarding scope of this ticket.

It's not a pre-req really, I implemented mongo 3.4. plumbing according to current mechanism (which is already a cascading version implementation) and this works fine for now. We _can_ look into improving the mechanism, but I'd say this is lower prio.

zeeiyerWork:
ok sounds good.

PR #2607, opened 2017-08-08
upgrade mongo driver

### LINKED ISSUE #2167: Can't use decimal types in MongoDB
Can't use a MongoDB collection that has decimals in it. Using MongoDB 3.4.4 and Quasar 16.1.6.

Example:

Import the following CSV file:
```
"car.total_value.decimal()","car.name.string()"
"1.50","Example"
```

Imported with:
`mongoimport -h 127.0.0.1  --db dectest --collection col --type csv --headerline --columnsHaveTypes --file dectest.csv`

The following is returned in Mongo Shell:
```
> db.col.findOne()
{
	"_id" : ObjectId("59105c6c476de35edb9f452c"),
	"car" : {
		"total_value" : NumberDecimal("1.50"),
		"name" : "Example"
	}
}
```

From Quasar REPL, I get the following exception:

```
💪 $ select * from col
💪 $ Exception in thread "AeshProcess: 12" java.lang.RuntimeException: Detected unknown BSON type "\x13" for fieldname "total_value". Are you using the latest driver version?
	at quasar.effect.Failure$.quasar$effect$Failure$$$anonfun$16(Failure.scala:125)
	at quasar.effect.Failure$$anon$5.apply(Failure.scala:105)
	at quasar.effect.Failure$$anon$5.apply(Failure.scala:103)
	at scalaz.NaturalTransformation$$anon$2.apply(NaturalTransformation.scala:19)
	at scalaz.NaturalTransformations$$anonfun$natToFunction$1.apply(NaturalTransformation.scala:41)
	at scalaz.$bslash$div.fold(Either.scala:57)
	at quasar.fp.free.package$EnrichNT$$anon$2.apply(package.scala:26)
	at quasar.fp.free.package$EnrichNT$$anon$2.apply(package.scala:26)
	at scalaz.NaturalTransformation$$anon$2.apply(NaturalTransformation.scala:19)
	at scalaz.NaturalTransformations$$anonfun$natToFunction$1.apply(NaturalTransformation.scala:41)
	at scalaz.$bslash$div.fold(Either.scala:58)
	at quasar.fp.free.package$EnrichNT$$anon$2.apply(package.scala:26)
	at quasar.fp.free.package$EnrichNT$$anon$2.apply(package.scala:26)
	at scalaz.NaturalTransformations$$anonfun$natToFunction$1.apply(NaturalTransformation.scala:41)
	at scalaz.$bslash$div.fold(Either.scala:58)
	at quasar.fp.free.package$EnrichNT$$anon$2.apply(package.scala:26)
	at quasar.fp.free.package$EnrichNT$$anon$2.apply(package.scala:26)
	at scalaz.Free.foldMap(Free.scala:217)
	at scalaz.Free.foldMap(Free.scala:219)
	at scalaz.Free$$anonfun$foldMap$2.apply(Free.scala:219)
	at scalaz.EitherT$$anonfun$flatMap$1$$anonfun$apply$4.apply(EitherT.scala:93)
	at scalaz.$bslash$div.fold(Either.scala:58)
	at scalaz.EitherT$$anonfun$flatMap$1.apply(EitherT.scala:93)
	at scalaz.EitherT$$anonfun$flatMap$1.apply(EitherT.scala:93)
	at scalaz.Free$$anonfun$step$1.apply(Free.scala:200)
	at scalaz.Free$$anonfun$step$1.apply(Free.scala:200)
	at scalaz.Free.step(Free.scala:202)
	at scalaz.Free.foldMap(Free.scala:215)
	at scalaz.Free$$anonfun$foldMap$2.apply(Free.scala:219)
	at scalaz.Kleisli$$anonfun$flatMap$1$$anonfun$apply$1.apply(Kleisli.scala:51)
	at scalaz.concurrent.Task$$anonfun$flatMap$1$$anonfun$1.apply(Task.scala:35)
	at scalaz.concurrent.Task$$anonfun$flatMap$1$$anonfun$1.apply(Task.scala:35)
	at scalaz.concurrent.Task$.Try(Task.scala:457)
	at scalaz.concurrent.Task$$anonfun$flatMap$1.apply(Task.scala:35)
	at scalaz.concurrent.Task$$anonfun$flatMap$1.apply(Task.scala:33)
	at scala.Function1$$anonfun$andThen$1.apply(Function1.scala:52)
	at scalaz.concurrent.Future$$anonfun$flatMap$1.apply(Future.scala:58)
	at scalaz.concurrent.Future$$anonfun$flatMap$1.apply(Future.scala:58)
	at scalaz.concurrent.Future.step(Future.scala:118)
	at scalaz.concurrent.Future.unsafePerformListen(Future.scala:75)
	at scalaz.concurrent.Future.unsafePerformAsync(Future.scala:159)
	at scalaz.concurrent.Future.unsafePerformSync(Future.scala:184)
	at scalaz.concurrent.Task.unsafePerformSync(Task.scala:99)
	at quasar.repl.Main$$anon$1.execute(Main.scala:81)
	at org.jboss.aesh.console.AeshProcess.run(AeshProcess.java:53)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
```

rintcius:
* upgrade mongo driver to version 3.5.0
* add support for decimal128
* handle `BSONException`

Note: 
Mongo driver requires Netty 4.1 which conflicts with Spark's Netty (4.0).
For now just excluding Spark's Netty, and thus breaking Spark.
This should be resolved once both Spark and new Mongo are on BackendModule (#2216 resp #2094) and we have dynamic backend loading in place (#2105).

Fixes #2167


mongodb/src/main/scala/quasar/physical/mongodb/bson.scala
```diff
@@ -39,32 +40,45 @@ sealed abstract class Bson extends Product with Serializable {
   //       BsonF[BsonValue] => BsonValue
   def repr: BsonValue
   def toJs: Js.Expr
+  def isSupported(v: MongoQueryModel): Boolean = true
```
rintcius: @wemrysi @alissapajer Before I go further with this PR, I'd like to have your thoughts on this implementation of versioning for `Bson` and `BsonValue`. 

I have added an `isSupported` method for `BsonValue` (line 48). 
The only usage is in line 58.  `fromRepr` is currently hard-coded to `3.4` but that's just to get it compiling without having to change code depending on it yet. The intention is to change `fromRepr`s signature to `MongoQueryModel => BsonValue => Bson` or something similar.

Similarly, I have added an `isSupported` method for `Bson` (and overrode it in `Dec128` below). This method isn't used yet.

wemrysi: @rintcius I think it is preferable to keep what versions of mongo support which bson values out of the definition of the datatype itself. We should be able to just add the new case for `Dec128` to `Bson` and then control when it is produced in the planner.

Can you give an example of why we'd need versioning support here?

rintcius: @wemrysi Ok, yeah I'm happy to move `isSupported` out of `Bson` and implement it like `isSupported` of `BsonValue`. Does that address your question or are you wondering whether we need version support for `Bson` at all? 

Main point why I have put this code up is what you think of this way of modelling versioning  (as opposed to e.g. different types a la `ExprOp<version>`), so in a way this is a possible solution for #2378

wemrysi: @rintcius Yeah, I'm not sure I see the need to version `Bson` right now, unless the `Decimal128` type was added in a new version. I found the [spec for Bson 1.1](http://bsonspec.org/) which includes `Decimal128`, but I can't seem to find a spec for Bson 1.0 so I'm not sure if it existed there or not. Even then, we'd want to version `Bson` itself and then have a mapping between Mongo version and Bson version where appropriate.

As for versioning at the value level, I think it is a useful approach, I don't think we can say yet whether it is how all things should be versioned (or if we even need to have a single way to version everything).

rintcius: Right, `Decimal128` was added in [mongo 3.4 (see type 19)](https://docs.mongodb.com/v3.4/reference/bson-types/) (also see [java doc](https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/BsonDecimal128.java#L26)), though it's apparently part of [bson 1.0 spec](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst).

wemrysi: Looks like [they rev'd the BSON spec](https://github.com/mongodb/bsonspec.org/commit/449ebfa65f7fcb04ffb6fa375a4dac5203e7bd43) when `Decimal128` was added.

rintcius: Ok, let's take that source :) Then I'll add a version for `Bson` specifically and use that for `isSupported` instead.

rintcius:
@wemrysi this is ready for review now. FYI, I ended up doing the bson version in `bsoncodec` instead of `bson`

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -191,7 +186,7 @@ object MongoDbPlanner {
       funcHandler(mf).map(t => unpack(t.mapSuspension(inj)))
 
     val handleSpecialCore: MapFuncCore[T, Fix[ExprOp]] => M[Fix[ExprOp]] = {
-      case Constant(v1) => ejsonToExpression[M, T[EJson]](v1)
+      case Constant(v1) => unimplemented[M, Fix[ExprOp]]("Constant expression")
```
rintcius: The translation of `Constant` is done in `handleCore`, so it never uses this code. I changed it to `unimplemented` otherwise I needed to parameterize this too

it/src/main/resources/tests/convertToFromString.test
```diff
@@ -3,8 +3,13 @@
     "backends": {
         "couchbase": "ignoreFieldOrder",
         "marklogic_json": "ignoreFieldOrder",
-        "mimir":"ignoreFieldOrder"
+        "mimir":"ignoreFieldOrder",
+        "spark_hdfs": "pending",
+        "spark_local": "pending"
     },
+    "NB": "Spark broken because of Netty conflict of Spark with new Mongo.
```
rintcius: Surprisingly this was the only actual failure (see the green build) but I did see lots of stacktraces

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1177,30 +1178,34 @@ object MongoDbPlanner {
       case RightSide => a2
     } ∘ (JsFn(JsFn.defaultName, _))
 
-  def exprOrJs[M[_]: Applicative, A]
-    (a: A)
-    (exf: A => M[Fix[ExprOp]], jsf: A => M[JsFn])
+  def exprOrJs[M[_]: Applicative, A, B]
```
rintcius: Looks like that can be reverted indeed, thanks - well spotted!

rintcius:
@wemrysi regarding `EJsonHandler`: that wouldn't lead to less parameter threading though? As I see it, everywhere `BsonVersion` is used now, `EJsonHandler` would be passed instead. Or am I missing something?

wemrysi:
@rintcius Yes, that's correct, but it would separate the need for different implementations of `EJson` handling from how they are obtained. Similar to how we have different implementations of `MapFunc` planning based on the server version, but we determine the implementation to use up front and the rest of the code base only needs to deal with the resulting `MapFunc[T, ?] ~> OptionFree[EX, ?]`, remaining closed to how the implementation was chosen.

In any case, as `BsonVersion` is the only factor affecting this decision we could wait and refactor if we end up needing to consider another dimension in order to determine how to generate the appropriate `Bson`.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -160,10 +161,10 @@ object MongoDbPlanner {
...
   def ejsonToExpression[M[_]: Applicative, EJ]
-    (ej: EJ)
+    (v: BsonVersion)(ej: EJ)
     (implicit merr: MonadError_[M, FileSystemError], EJ: Recursive.Aux[EJ, EJson])
       : M[Fix[ExprOp]] =
-    ej.cataM(BsonCodec.fromEJson).fold(pe => merr.raiseError(qscriptPlanningFailed(pe)), $literal(_).point[M])
+    ej.cataM(BsonCodec.fromEJson(v)).fold(pe => merr.raiseError(qscriptPlanningFailed(pe)), $literal(_).point[M])
 
   // TODO: Use `JsonCodec.encode` and avoid failing.
   def ejsonToJs[M[_]: Applicative, EJ: Show]
```
wemrysi: It looks like we should also be using `NumberDecimal` values here, when supported. We currently transform `EJson -> Data -> Js`, could we instead use `EJson -> Bson -> Js` and benefit from the new `Bson` handling?

rintcius: I don't think we can use `EJson -> Bson -> Js` and should not use `NumberDecimal` here. 

E.g. compare:
```
  final case class Int32(value: Int) extends Bson {
    def repr = new BsonInt32(value)
    def toJs = Js.Call(Js.Ident("NumberInt"), List(Js.Str(value.shows)))
  }
```
with:
```
  final case class Int(value: BigInt) extends Number {
    def dataType = Type.Int
    def toJs = jscore.Literal(Js.Num(value.doubleValue, false)).some
  }
```

Also the types of both `toJs` functions are different (but that's not a reason on its own, they could still have been "sufficiently" isomorphic)

See https://gist.github.com/rintcius/3750a9cd45650942fc7f8d9aa97b4b75 for a grep on `Data.toJs` vs `Bson.toJs`

wemrysi: I didn't realize it but they also generate values in two _different_ Javascript representations (not sure why we have two...).

However `Data` does generate MongoDB specific Javascript as well (it shouldn't actually have the `toJs` method at all, but that's another issue), it creates `BinData`, `ISODate` and `ObjectId` objects, which aren't generic Javascript, which is why I thought we might also want to use `NumberDecimal` as well. Unfortunately, it looks like `NumberDecimal` may [not even be supported](https://stackoverflow.com/questions/45040685/mongodb-3-4-map-reduce-and-decimal128) in map-reduce, so this is a moot point.

mongodb/src/main/scala/quasar/physical/mongodb/MongoQueryModel.scala
```diff
@@ -16,25 +16,50 @@
...
+
+  def toBsonVersion(v: MongoQueryModel): BsonVersion =
+    if (v lt `3.4`) BsonVersion.`1.0`
+    else BsonVersion.`1.1`
+
+  implicit val showMongoQueryModel: Show[MongoQueryModel] = new Show[MongoQueryModel] {
+    override def shows(v: MongoQueryModel) = v.s
+  }
+
+  implicit val equalMongoQueryModel: Equal[MongoQueryModel] = Equal.equalA
```
wemrysi: Missed this before, `Order <: Equal` so you don't need to define both. Doing so will actually cause implicit resolution for `Equal[MongoQueryModel]` to fail due to ambiguity, so we can get rid of this instance.

rintcius:
@wemrysi PR updated

BTW I left `EJsonHandler` out for now, but I'll keep it in mind when there's more time to refactor `MongoDbPlanner`

alissapajer:
We are going to wait to merge this until we have dynamic loading for connectors completed. If we merged it now, we would no longer be able to test against spark. This would let us unwittingly introduce bugs into spark. Additionally, spark is an important test case for ongoing `QScript` work.

PR #2608, opened 2017-08-08
Improve MetaStore programmatic API

jedesah:


core/src/main/scala/quasar/db/package.scala
```diff
@@ -32,11 +32,20 @@ package object db {
...
 
-  def poolingTransactor(cxn: ConnectionInfo, config: HikariConfig => Task[Unit]): Task[StatefulTransactor] =
-    for {
+  /** Transactor that makes use of a connection pool for performance. Requires cleanup. */
+  def poolingTransactor(cxn: ConnectionInfo, config: HikariConfig => Task[Unit]): Task[Throwable \/ StatefulTransactor] =
+    (for {
       xa <- HikariTransactor[Task](cxn.driverClassName, cxn.url, cxn.userName, cxn.password)
       _  <- xa.configure(config)
-    } yield StatefulTransactor(xa, xa.shutdown)
+    } yield StatefulTransactor(xa, xa.shutdown)).attempt
```
wemrysi: I think it'd be better to just return the `Task` rather than `attempt` it here as, unless _I_ attempted a `Task[A]`, the fact that I am in possession of a `Task[Throwable \/ A]` doesn't mean the `Task` isn't going to fail.

core/src/main/scala/quasar/metastore/MetaStore.scala
```diff
@@ -17,19 +17,86 @@
...
+    *               needs to store additional information in the MetaStore
+    * @return A `MetaStore` object containing the `transactor` to use to perform operation on the MetaStore as
+    *         well as it's location and expected Schema
+    */
+  def connect(dbConfig: DbConnectionConfig, initializeOrUpdate: Boolean, schemas: List[Schema[Int]]): EitherT[Task, MetastoreFailure, MetaStore] = {
+    for {
+      tx <- EitherT(poolingTransactor(DbConnectionConfig.connectionInfo(dbConfig), DefaultConfig))
+              .leftMap(t => UnknownError(t, "While connecting to MetaStore"):MetastoreFailure)
+      _  <-
+        if (initializeOrUpdate)
```
wemrysi: Alternatively, `initializeOrUpdate.whenM(schemas.traverse(...))`

core/src/main/scala/quasar/db/package.scala
jedesah: My thought process was that although we consider the possibility of a `Task` failing at the edge of the world, in the rest of the codebase we generally assume that if the `Task` has a known failure scenario that it will be reflected in the type of the result, i.e. `EitherT`. Since we know that this call will fail in certain scenarios, I wanted to make that explicit so that the client code doesn't have to remember to call attempt lest the `Task` failure bubble up to a `InternalServerError` in the case where it's not handled properly.

core/src/main/scala/quasar/metastore/MetaStore.scala
jedesah: 👍

core/src/main/scala/quasar/db/package.scala
wemrysi: Is there a more specific exception we can catch then? If the lhs of the disjunction was something more specific than `Throwable` I think it'd be fine, its just awkward as-is because `Task` already implies the computation may fail with an exception.

core/src/main/scala/quasar/metastore/MetaStore.scala
```diff
@@ -17,19 +17,86 @@
...
+
+  def initializeOrUpdate[A](
+    schema: Schema[A], transactor: Transactor[Task], jCfg: Option[Json]
+  ): EitherT[Task, MetastoreFailure, Option[Json]] = {
+    val mntsFieldName = "mountings"
+
+    val mountingsConfigDecodeJson: DecodeJson[Option[MountingsConfig]] =
+      DecodeJson(cur => (cur --\ mntsFieldName).as[Option[MountingsConfig]])
+
+    val mountingsConfig: ConnectionIO[Option[MountingsConfig]] =
```
nicflores: Are we still keeping this functionality? I know you thought about removing it, just thought I'd ask.

jedesah: Ya, I wish I could remove it, but I think it's best to leave it there in case people have still not upgraded to a version of Quasar with the MetaStore. Perhaps one day we can remove it in a yak shaving exercise

core/src/main/scala/quasar/db/package.scala
jedesah: The documentation of `doobie` does not list any specific errors, but I changed `Throwable` to `metastore.UnknownError` which I think is better anyway.

PR #2610, opened 2017-08-08
Correct provenance used in autojoin condition

### LINKED ISSUE #2121: Enable grouped join test for at least two connectors

### LINKED ISSUE #2331: Enable `unifyFlattenedFieldsWithUnflattened` for at least one qscript-based backend

### LINKED ISSUE #2543: Basic union query results in a bad plan on QScript-based MongoDB connector
The query from the `union.test` regression test

`select _id as zip from zips union select city, state from zips`

produces a MongoDB workflow that errors at runtime (from 3.2, but also tested and failed on 3.0 and 2.6):

```
Quasar error: Plan execution failed: MongoDB Error: Command failed with error 139: 'TypeError: __val[0][3][0] is undefined :
_funcs1/</<@:10:44
_funcs1/<@:7:10
_funcs1@:4:1
' on server localhost:27021. The full response is { "ok" : 0.0, "errmsg" : "TypeError: __val[0][3][0] is undefined :\n_funcs1/</<@:10:44\n_funcs1/<@:7:10\n_funcs1@:4:1\n", "code" : 139 }, cause=Some(Command failed with error 139: 'TypeError: __val[0][3][0] is undefined :
_funcs1/</<@:10:44
_funcs1/<@:7:10
_funcs1@:4:1
' on server localhost:27021. The full response is { "ok" : 0.0, "errmsg" : "TypeError: __val[0][3][0] is undefined :\n_funcs1/</<@:10:44\n_funcs1/<@:7:10\n_funcs1@:4:1\n", "code" : 139 })
```

wemrysi:
When `autojoin` needed to actually generate a theta-join it was building the join condition using the provenance that resulted from the join rather than the provenance of the inputs to it, this fixes it to correctly use the input provenance.

Fixes #2121.
Fixes #2331.
Fixes #2543.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/PruneArrays.scala
+++ b/connector/src/main/scala/quasar/qscript/PruneArrays.scala
@@ -172,7 +172,7 @@ class PAHelpers[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
     } map (_.merge)
 
   /** Prune the provided `array` keeping only the indices in `indicesToKeep`,
-    * eliding it alltogether if only a single element is retained.
+    * eliding it altogether if only a single element is retained.
     */
   def rewriteRepair[A](repair: FreeMapA[A], seen: SeenIndices): Option[FreeMapA[A]] =
     repair.project.some collect {
--- a/connector/src/main/scala/quasar/qscript/PruneArrays.scala
+++ b/connector/src/main/scala/quasar/qscript/PruneArrays.scala
@@ -162,7 +162,7 @@ class PAHelpers[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
       case co => cbf.leftMap(co)(_.right[JoinSide])
     } map (_.merge)
 
-  def remapIndicesInLeftShift[A](struct: FreeMap, repair: JoinFunc, mapping: IndexMapping): JoinFunc =
+  def remapIndicesInLeftShift[A](repair: JoinFunc, mapping: IndexMapping): JoinFunc =
     repair.transCata[EitherMap[JoinSide]] {
       case CoEnv(\/-(MFC(ProjectIndex(Embed(CoEnv(-\/(\/-(LeftSide)))), IntLit(idx))))) =>
         remapResult[JoinSide](LeftSide, mapping, idx)
@@ -223,7 +223,8 @@ object PruneArrays {
       : PruneArrays[IN] =
     new PruneArrays[IN] {
       def find[M[_], A](in: IN[A])(implicit M: MonadState[M, RewriteState]) =
-        M.put(Ignore).as(Ignore)
+        haltRemap(Ignore)
+
       def remap[M[_], A](env: RewriteState, in: IN[A])(implicit M: MonadState[M, RewriteState]) =
         haltRemap(in)
     }
@@ -434,7 +435,7 @@ object PruneArrays {
             }
             M.modify(bucketState |+| orderState |+| _).as(Ignore)
 
-          case Unreferenced() => M.modify(ι).as(Ignore)
+          case Unreferenced() => (Ignore: RewriteState).point[M]
         }
 
       @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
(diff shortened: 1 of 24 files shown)


it/src/main/resources/tests/unifyFlattenedFieldsWithUnflattened.test
```diff
@@ -2,18 +2,13 @@
...
         "mongodb_3_0": "pending",
         "mongodb_3_2": "pending",
         "mongodb_3_4": "pending",
-        "mongodb_read_only": "pending",
-        "spark_hdfs": "timeout",
-        "spark_local": "timeout"
+        "mongodb_read_only": "pending"
     },
-    "NB": "mimir, marklogic_json, marklogic_xml, spark_hdfs and spark_local not on par with master",
+    "NB": "Fails in mongodb due to the results becoming unordered by MR.",
```
alissapajer: Can we us the `ignoreResultOrder` flag and enable this test for mongo?

it/src/main/resources/tests/unifyFlattenedFieldsWithMultipleUses.test
```diff
@@ -1,17 +1,17 @@
...
+        "marklogic_json": "skip",
+        "marklogic_xml": "skip",
+        "mimir":"skip",
+        "mongodb_2_6": "skip",
+        "mongodb_3_0": "skip",
+        "mongodb_3_2": "skip",
+        "mongodb_3_4": "skip",
+        "mongodb_read_only": "skip",
+        "spark_hdfs": "skip",
+        "spark_local": "skip"
```
alissapajer: Why skip all these now?

connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -162,12 +162,12 @@ class PAHelpers[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
...
     } map (_.merge)
 
-  def remapIndicesInLeftShift[A](struct: FreeMap, repair: JoinFunc, mapping: IndexMapping): JoinFunc =
+  def remapIndicesInLeftShift[A](repair: JoinFunc, mapping: IndexMapping): JoinFunc =
     repair.transCata[EitherMap[JoinSide]] {
       case CoEnv(\/-(MFC(ProjectIndex(Embed(CoEnv(-\/(\/-(LeftSide)))), IntLit(idx))))) =>
         remapResult[JoinSide](LeftSide, mapping, idx)
-      case CoEnv(\/-(MFC(ProjectIndex(Embed(CoEnv(-\/(\/-(RightSide)))), IntLit(idx))))) if struct ≟ HoleF =>
-        remapResult[JoinSide](RightSide, mapping, idx)
+      case CoEnv(\/-(MFC(ProjectIndex(value @ Embed(CoEnv(\/-(MFC(ProjectIndex(Embed(CoEnv(-\/(\/-(RightSide)))), IntLit(ridx)))))), IntLit(idx))))) if ridx ≟ 1 =>
```
alissapajer: Can we not just match on `IntLit(1)` rather than constrain with `if ridx ≟ 1`?

connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -861,6 +861,38 @@ class PruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScriptHelp
       initial.pruneArraysF must equal(expected)
     }
 
+    "rewrite left shift when struct references a pruned index" in {
+      val initial: Fix[QST] =
+        QCT.inj(Map(
+          QCT.inj(Filter(
+            QCT.inj(LeftShift(
+              ReadRT(rootDir </> file("data")).embed,
+              ProjectIndexR(ProjectIndexR(HoleF, IntLit(2)), IntLit(1)),
```
alissapajer: The test description doesn't seem to match what's being tested here. The `struct` is dereferencing from the `Read`. And the array that's being pruned contains only constant values, and doesn't referenced `LeftSide` or `RightSide`. Is this what was intended?

connector/src/test/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -861,6 +861,38 @@ class PruneArraysSpec extends quasar.Qspec with CompilerHelpers with QScriptHelp
       initial.pruneArraysF must equal(expected)
     }
```
alissapajer: Since we were previously incorrectly rewriting the `ProjectIndex(RightSide, 1)` pattern, is there another negative unit test you could add here ensuring we are no longer doing that?

it/src/main/resources/tests/unifyFlattenedFieldsWithMultipleUses.test
wemrysi: The QScript has problems and, for MongoDB at least, the resulting behavior is to fill up local disk with gigabytes of data (probably due to a runaway cross or similar). It's a bit coarse-grained, I admit, but it allows running the rest of the suite in ~5minutes without having to wait for this (and other pathological cases like it) to timeout.

connector/src/main/scala/quasar/qscript/PruneArrays.scala
wemrysi: I tried that, but `IntLit` results in a `BigInteger` which you can't match against `1`.

connector/src/test/scala/quasar/qscript/PruneArrays.scala
wemrysi: I'll figure out a better wording, by "references a pruned index" I mean projects an index that is in the set to be pruned, the problem only manifests itself when we're pruning an index in the `repair` (in this case, `2`) that also happens to appear in the `struct`.

alissapajer:
Overall LGTM. Once those minor changes land, I'll merge.

PR #2611, opened 2017-08-09
Handle shifting on both sides of a join condition.

alissapajer:
Work toward #2110.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/PruneArrays.scala
+++ b/connector/src/main/scala/quasar/qscript/PruneArrays.scala
@@ -172,7 +172,7 @@ class PAHelpers[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
     } map (_.merge)
 
   /** Prune the provided `array` keeping only the indices in `indicesToKeep`,
-    * eliding it alltogether if only a single element is retained.
+    * eliding it altogether if only a single element is retained.
     */
   def rewriteRepair[A](repair: FreeMapA[A], seen: SeenIndices): Option[FreeMapA[A]] =
     repair.project.some collect {
--- a/connector/src/main/scala/quasar/qscript/ReplaceMapFunc.scala
+++ b/connector/src/main/scala/quasar/qscript/ReplaceMapFunc.scala
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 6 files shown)


connector/src/main/scala/quasar/qscript/PruneArrays.scala
```diff
@@ -172,7 +172,7 @@ class PAHelpers[T[_[_]]: BirecursiveT: EqualT] extends TTypes[T] {
     } map (_.merge)
 
   /** Prune the provided `array` keeping only the indices in `indicesToKeep`,
-    * eliding it alltogether if only a single element is retained.
+    * eliding it altogether if only a single element is retained.
```
wemrysi: TIL 😄

connector/src/main/scala/quasar/qscript/ReplaceMapFunc.scala
```diff
@@ -0,0 +1,107 @@
...
+import quasar.fp._
+
+import matryoshka._
+import matryoshka.patterns._
+import matryoshka.data._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+trait ReplaceMapFunc[T[_[_]], F[_]] {
+  def replace(name: Symbol, repl: FreeMap[T]): NTComp[F, State[Boolean, ?]]
```
wemrysi: Could also use `Writer[Boolean @@ Disjunction, ?]` since we don't need to modify the state.

connector/src/main/scala/quasar/qscript/Transform.scala
```diff
@@ -628,26 +628,82 @@ class Transform
...
+          } yield (l, r)
+
+          val replaceSwap: Option[(FreeQS, FreeQS)] = for {
+            l <- replaceLeft(rTheta)
+            r <- replaceRight(lTheta)
+          } yield (l, r)
+
+          val result: PlannerError \/ (FreeQS, FreeQS) =
+            replace.orElse(replaceSwap).toRightDisjunction(
+            InternalError.fromMsg(s"Non-supported join condition found: branches don't each contain a single join side."))
```
wemrysi: Any idea if this case is possible "naturally" or would it indicate a bug in the compiler somewhere?

alissapajer: Hm, yeah I can't think of a query that would naturally hit this case, but I'm not completely sure right now that there isn't one.

connector/src/main/scala/quasar/qscript/ReplaceMapFunc.scala
alissapajer: Good idea, we can make that change.

Issue #2471, opened 2017-07-14
Integration tests for Spark-family of connectors. 


zeeiyerWork:

nicflores:
Some work on this front has already begun in this branch:

https://github.com/nicflores/quasar/tree/wip/docker-spark-connectors

Currently this branch contains docker-compose definitions for containerized Small Spark Cluster, FTP, ElasticSearch, Cassandra, and Travis matrix definitions for each. What's missing is Hadoop, a little more config in order to get travis to spin up each of these up along side a Spark, and maybe a little more config to make Spark aware of each of these storage container.

nicflores:
Currently the branch above contains the following working containers:
```
quasar_spark_cluster
quasar_ftp
quasar_elasticsearch
quasar_cassandra
quasar_hadoop
```
Still to do
- [ ] travis matrix builds to bring up `quasar_spark_cluster` and associated connector
- [ ] put spark and associated connector containers on an IP that allows for container to container communication and host to container communication.

nicflores:
@rabbitonweb, I have a zip branch building in travis right now:

https://travis-ci.org/quasar-analytics/quasar/jobs/260787413#L612

There may be some work that still needs to be done to pin down the correct connectionURI's for the new connectors but for the most part I think the difficult stuff is out of the way. Check the link above and let me know what you think. We can discuss what needs to change at a later day.

Also, @zeeiyerWork mentioned that you may need an S3 connector as well. I'd like to get some clarification from you as how you'd like to proceed on this front.

zeeiyerWork:
Yeah @damonLL, @nicflores, @rabbitonweb -> wrt testing Spark against JSON on S3
1. Feel this is important - since we might get more @bats in the field.
2. Options are:
- Keep a test S3 mount available with pre-existing JSON content which we refer to in tests. This might be cheaper, since on S3 Storage Costs = 1/10th of Upload Costs. 
- However to be consistent with how we run other tests, we may want to create a S3 test-folder on the fly and upload JSON to it, run integration tests and tear it down. Maybe costs more given repeated S3 ingests, but perhaps negligible if it is just a handful of JSONs uploaded on each test execution.

Thanks

nicflores:
@rabbitonweb, can you please post the full connection URI's for each of the new spark connectors just as we need them to show up in`it/testing.conf`. My guess is that they will look something like this:
```
spark_cassandra=http://${CONNECTORIPADDR}:9042
[...]

PR #2632, opened 2017-08-15
spark-hdfs back to normal

rabbitonweb:
This is part of #2471 
It upgrades spark to latest `2.2.0` + brings spark-hdfs back to normal

### DIFF
--- a/frontend/src/main/scala/quasar/std/date.scala
+++ b/frontend/src/main/scala/quasar/std/date.scala
@@ -168,69 +168,55 @@ trait DateLib extends Library with Serializable {
       constTyper[nat._1](Type.Numeric),
       basicUntyper)
 
-  /** 
-    * Please note that all vals are lazy in this companion object. There is a reason for that, yet not completely 
-    * understood yet - it's like quantum theory.
-    * 
-    * It is still not completely clear for me why this is needed. In essence we had an error, that
-    * manifasted itself as ClassNotFoundError on spark's executors. This normally means that sparkcore.jar (uberjar) is
-    * not send to the cluster or the given class (DateLib$) is not in that jar. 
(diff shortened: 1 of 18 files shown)


frontend/src/main/scala/quasar/std/date.scala
```diff
@@ -168,55 +168,55 @@ trait DateLib extends Library with Serializable {
       constTyper[nat._1](Type.Numeric),
       basicUntyper)
 
-  val ExtractCentury      = extract(
+  lazy val ExtractCentury      = extract(
```
rintcius: can you add a note (+ link to SO question) why this is needed?

rintcius:
Oh, can you also add `version : ...` label?

frontend/src/main/scala/quasar/std/date.scala
rabbitonweb: It is still not completely clear for me why this is needed. In essence we had error manifasting itself as `ClassNotFoundError` on spark's executors. This normally means that sparkcore.jar (uberjar) is not send to the cluster or the given class (`DateLib`) is not in that jar. After verifying that is not the case, I kept on googling and I found this https://stackoverflow.com/questions/7262799/why-does-scala-fail-to-instantiate-a-companion-object Seemed odd but it was worth a try. And it solved the issue. This is something worth investigating further and I'm actually planning to do so (recreate the issue on a sandboxed project), but for now I need working spark connectors and this thing solves it (even though it runs on magic at the moment)

rabbitonweb: Please note that
1. Any sufficiently advanced technology is indistinguishable from magic
2. spark does a lot of dynamic loading using classloaders and we might have fallen to one of the (not yet discovered) bugs

PR #2633, opened 2017-08-15
[Fix #2313] Fix video playback in Safari

### LINKED ISSUE #2313: MP4 H.264 Videos don't play in Safari
They do when hosted from other web servers but not when hosted from Quasar.

jedesah:


web/src/main/scala/quasar/api/package.scala
```diff
@@ -231,11 +231,47 @@ package object api {
     from.parsePath(to.unsafePrintPath, to.unsafePrintPath, to.unsafePrintPath, to.unsafePrintPath)
 
   def staticFileService(basePath: String): HttpService = {
```
drostron: Would something like the following do the trick?
```scala
  def staticFileService(basePath: String): HttpService = {
    val fsConfig = FileService.Config(systemPath = basePath)

    fileService(fsConfig.copy(
      pathCollector = (file, config, req) =>
        if (file.isDirectory) Task.delay(StaticFile.fromFile(new jFile(file, "index.html"), Some(req)))
        else fsConfig.pathCollector(file, config, req)))
  }
```

jedesah: Yes! Thx for the suggestion. I also opened a PR against http4s changing the default behavior which got promptly merged and released in `0.16.0a-RC1`. Unfortunately, I tried upgrading, and the change from `Response` to `MaybeResponse` is kind of a pain in the ass so I opened another PR which if accepted should help us migrate to `0.16.0a` without major changes to our tests. But for now, this will fix the issue until we can upgrade to `0.16.0a`.

PR #2646, opened 2017-08-17
Change planner signature

rabbitonweb:
Finally closes https://github.com/quasar-analytics/quasar/issues/1634

It happily:

1. removes QueryFile.Input completely 
2. ~~makes sparkcore Planner more generic - it can be used with any `M` as long as it is `Monad` and `Capture`~~

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/QScriptCorePlanner.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/QScriptCorePlanner.scala
@@ -22,6 +22,7 @@ import quasar.common.SortDir
 import quasar.contrib.pathy.AFile
 import quasar.fp._, ski._
 import quasar.qscript._, ReduceFuncs._, SortDir._
+import quasar.effect.Capture
 
 import scala.math.{Ordering => SOrdering}, SOrdering.Implicits._
 
@@ -34,8 +35,8 @@ import matryoshka.patterns._
 import scalaz._, Scalaz._
 import scalaz.concurrent.Task
 
-class QScriptCorePlanner[T[_[_]]: BirecursiveT: ShowT]
-    extends Planner[QScriptCore[T, ?]] {
+class QScriptCorePlanner[T[_[_]]: BirecursiveT: ShowT, M[_]:Capture]
+    extends Planner[QScriptCore[T, ?], M] {
 
   import Planner.{SparkState, SparkStateT}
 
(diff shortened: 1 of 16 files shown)


rabbitonweb:
@djspiewak this is ready for review

rabbitonweb:
I also have an idea about removing `StateT` from `SparkState` but I need to talk to @alissapajer about that so I might address that in other PR

rabbitonweb:
@djspiewak  take it back, it is not ready -  it has a failing test.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/package.scala
```diff
@@ -172,7 +172,7 @@ package object elastic {
 
     type FreeEff[A]  = Free[Eff, A]
```
djspiewak: Wait… what?  I think I missed this when looking through the Cassandra PR.  Why are we taking `Free` of `Eff`?  `Eff` is already a free monad, unless this is a different `Eff`?  Also which `Eff` implementation is this?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/QScriptCorePlanner.scala
```diff
@@ -233,7 +233,7 @@ class QScriptCorePlanner[T[_[_]]: BirecursiveT: ShowT]
           val main = sortBys.head._2
           val keys = (bucket, main) :: sortBys.toList
           src.sortBy(d => KO(keys.map(p => p.leftMap(_(d))), main), main === SortDir.Ascending)
```
djspiewak: Trying to understand where things are being serialized and sent to Spark and where they're not…  How is the above not requiring `scalaz.Equal` to be `Serializable`?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/package.scala
edmundnoble: This `Eff` is a type alias on line 51. They're in most of the package objects of the connectors, as a common effect type for the connector.

rabbitonweb:
I've updated code. `M[_]:Capture:Monad` had to be removed from `Planner` as `Planner` is being serialized (with all evidences that it contains). Thus we are almost back to square one. The nice difference is that we removed `QueryFile.Input` completely from the code base which was the origin idea behind this code base anyway. Shame we had to make our effects concrete but you know, Spark ¯\_(ツ)_/¯

rabbitonweb:
@djspiewak waiting the tests to pass but this is pretty much ready for review

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/QScriptCorePlanner.scala
rabbitonweb: this can work as this is being invoked on the cluster, the problem is that the whole instance is serialized as part of the clojure

rabbitonweb:
Will reopen as original meaning of this PR was lost

PR #2650, opened 2017-08-17
[Fix #2485] Keep track of sort order in Mimir Repr

### LINKED ISSUE #2485: Mimir internal Repr should track current sort order

edmundnoble:
Keeps track of the current sort order in Mimir to avoid sorting the same data the same way repeatedly.

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/Mimir.scala
+++ b/mimir/src/main/scala/quasar/mimir/Mimir.scala
@@ -61,6 +61,10 @@ import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicLong
 
+import quasar.yggdrasil.TableModule.{DesiredSortOrder, SortAscending}
+
+import scalaz.Leibniz.===
+
 object Mimir extends BackendModule with Logging {
   import FileSystemError._
   import PathError._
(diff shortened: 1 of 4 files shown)


djspiewak:
Needs a version tag and an assignee.

yggdrasil/src/main/scala/quasar/yggdrasil/TransSpecModule.scala
```diff
@@ -254,6 +255,274 @@ trait TransSpecModule extends FNModule {
...
+          case InnerArrayConcat(ls@_*) =>
+            Some(ls.toList.flatMap(a => flattenInnerArrayConcats(a).getOrElse(a :: Nil)))
+          case _ => None
+        }
+        def flattenInnerObjectConcats[A <: SourceType](proj: TransSpec[A]): Option[List[TransSpec[A]]] = proj match {
+          case InnerObjectConcat(ls@_*) =>
+            Some(ls.toList.flatMap(a => flattenInnerObjectConcats(a).getOrElse(a :: Nil)))
+          case _ => None
+        }
+        @inline def flattenedOuterArrayConcatNormalized: Option[TransSpec[A]] =
```
djspiewak: Random FYI: `@inline` doesn't really work.

edmundnoble: Yeah I guess my brain was stuck in 2.12-mode :P

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -77,48 +81,76 @@ object Mimir extends BackendModule with Logging {
...
   private type Cake = Precog with Singleton
 
+  // EquiJoin results are sorted by both keys at the same time, so we need to keep track of both
+  final case class SortOrdering[TS1](sortKeys: Set[TS1], sortOrder: DesiredSortOrder, unique: Boolean) {
+    def sort(src: Cake)(table: src.Table)(implicit ev: TS1 === src.trans.TransSpec1): Future[src.Table] =
+      if (sortKeys.isEmpty) Future.successful(table)
+      else table.sort(ev(sortKeys.head), sortOrder, unique)
+  }
+
+  final case class SortRepr[TS1](bucket: Option[TS1], orderings: List[SortOrdering[TS1]])
```
edmundnoble: Agreed.

edmundnoble:
Any examples of inconsistent formatting, so I know what to look out for?

yggdrasil/src/main/scala/quasar/yggdrasil/TransSpecModule.scala
djspiewak: :-)  It's also worth noting that, even if `@inline` did work, there are a lot of cases where you don't *want* it to, due to HotSpot's byte limit on certain analyses.

djspiewak:
> Any examples of inconsistent formatting, so I know what to look out for?

- Sometimes you use spaces to separate operators, sometimes you don't (note: `.traverse{`)
- Sometimes you use dots in function calls, sometimes not even when things fit on a single line (dot vs space dispatch syntax)
- Sometimes you use curly braces for multi-line things, sometimes not
- Sometimes you newline and indent constructs like `if`, `else` and `yield` (even when things could fit on a single line), and sometimes you put things on a single line
- Use of blank spacing lines is inconsistent (e.g. almost none in `normalize`, but more in `rephrase`)
- Some things are vertically aligned (you explicitly vertically aligned the `MapN` case even with no other changes on that line), and some things are not
  + Even worse, some things look like they were *intended* to be vertically aligned, but don't really line up with anything (e.g. the parameters to `rephrase`)
- Dangling paren usage is inconsistent (sometimes you dangle closing parens, sometimes you don't; this is related to the inconsistent use of curly braces)
- Multi-line invocations sometimes have the first actual on the first line, sometimes have multiple actuals on the first line, sometimes none at all

I didn't want to call out each thing individually, because that feels like super nit-picking.  Just wanted to generically whine about it.  :-)  Personal style disagreements aside (e.g. whether or not , *inconsistent* style makes things very very difficult to read, at least for me.

Issue #1994, opened 2017-03-21
MongoDB: Timestamp values changes after Flattening


sriksrin:
Hi,

 I created a sample document in MongoDB for knowing how flattening works. 

This is the document i created

 {
   "dvcid":"1_!",
    "ts":  ISODate("2017-01-01T00:00:00Z"),
    "temp": {
           "0": 20,
           "1": 30,
           "2": 24,
           "3": 23
        }
}

when, queried without flattening i.e. 
  SELECT dvcid,
ts,
temp
FROM `/test/test/test1`

the ts values shows the proper value. i.e whats there in the mongodb.

But, when queired with flattening 
  SELECT dvcid,
ts,
temp{*}
FROM `/test/test/test1`

the "ts" values are not what is stored in mongodb. The current time stamp/ system timestamp is considered and shown in the output.

Why is this happening? 
How to enforce "ts" values to show from mongodb and not take system time?

Regards,
Srikanth

zeeiyerWork:
Sounds like this is a correctness issue - and hence important to fix. Let's see if this behavior goes away with the new Mongo connector. !

jsantos17:
This is still happening with next-major:

I added a `ts` column to `smallZips.data` to try this.

Without flattening:  `SELECT loc[1], loc[2], ts  FROM 'mngo/test/smallZips'` yields

```
 0          | ts                    |
------------|-----------------------|
  42.070206 |  2017-01-05T00:00:00Z |
  42.377017 |  2017-01-05T00:00:00Z |
  42.409698 |  2017-01-05T00:00:00Z |
  42.275103 |  2017-01-05T00:00:00Z |
  42.182949 |  2017-01-05T00:00:00Z |
  42.116543 |  2017-01-05T00:00:00Z |
  42.279421 |  2017-01-05T00:00:00Z |
   42.38167 |  2017-01-05T00:00:00Z |
  42.162046 |  2017-01-05T00:00:00Z |
  42.176443 |  2017-01-05T00:00:00Z |
```

which includes the correct date stored in Mongo.

With flattening: `SELECT loc[*], ts  FROM '/mngo/test/smallZips'` yields

```
 loc         | ts                        |
-------------|---------------------------|
  -72.622739 |  2017-08-09T03:21:46.928Z |
   42.070206 |  2017-08-09T03:21:46.928Z |
   -72.51565 |  2017-08-09T03:21:46.929Z |
   42.377017 |  2017-08-09T03:21:46.929Z |
[...]

PR #2651, opened 2017-08-17
Date changes after flattening.

jsantos17:
Addresses #1994

So basically the bug was in the `clone` function we use in MapReduce plans.

With the previous `clone` function

```javascript
> clone(ISODate("1990-05-16"))
ISODate("2017-08-17T23:37:37.209Z")
```

With the patched version:

```javascript
> clone(ISODate("1990-05-16"))
ISODate("1990-05-16T00:00:00Z")
```

So turns out, not strictly related to flattening. More like anywhere we use `clone`.


alissapajer:
@jsantos17 can you add an integration test for the problematic query?

PR #2655, opened 2017-08-18
Mimir equijoin fix

nicflores:
Fix tests selectCount and nestedSelectWithReduce in Mimir.

### DIFF
--- a/it/src/main/resources/tests/demo/nestedSelectWithReduce.test
+++ b/it/src/main/resources/tests/demo/nestedSelectWithReduce.test
@@ -1,7 +1,6 @@
 {
     "name": "select reduction from nested select",
     "backends": {
-        "mimir":"pending",
         "couchbase": "pending",
         "marklogic_json": "pending",
         "marklogic_xml": "pending",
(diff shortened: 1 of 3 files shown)


it/src/main/resources/tests/selectCount.test
```diff
@@ -2,7 +2,6 @@
     "name": "select count and another field",
     "backends": {
         "couchbase": "pending",
-        "mimir":"pendingIgnoreFieldOrder",
```
djspiewak: Shouldn't this be `"mimir": "ignoreFieldOrder"`?

it/src/main/resources/tests/demo/nestedSelectWithReduce.test
```diff
@@ -1,7 +1,6 @@
 {
     "name": "select reduction from nested select",
     "backends": {
-        "mimir":"pendingIgnoreFieldOrder",
```
djspiewak: Shouldn't this be `"mimir": "ignoreFieldOrder"`?

djspiewak:
@nicflores Did you update #2542 with the results of this PR?

it/src/main/resources/tests/demo/nestedSelectWithReduce.test
nicflores: test are passing with and without using `"mimir": "ignoreFieldOrder"`. Is it possible that I'm getting lucky and getting a passing test without using `"mimir": "ignoreFieldOrder"`?

nicflores:
#2542 is now updated. I was gonna wait until this was merged.

it/src/main/resources/tests/demo/nestedSelectWithReduce.test
djspiewak: Very possible.  If the test doesn't have `"ignoreFieldOrder": true` *and* the expected output has multiple fields, then the only way it would be passing now is getting lucky on `ColumnRef` hash order.

djspiewak:
@nicflores I like to update the issue by crossing out the ones I've fixed as soon as I fix them, so that others know they're fixed on some branch *somewhere*.  Then, once the PR lands, I just remove those rows.

On another note, merge conflicts!

it/src/main/resources/tests/demo/nestedSelectWithReduce.test
nicflores: ok, I'll add `"mimir": "ignoreFieldOrder"` to each of those tests.

PR #2660, opened 2017-08-22
Change planner signature

### LINKED ISSUE #1634: Replace "input" functions in spark-core with algebras so connectors can write interpreters for them

rabbitonweb:
Finally closes #1634 - it removes QueryFile.Input completely. Also in `Planner` the `Task` is replace with `Free[S,?`]

### DIFF
--- a/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/package.scala
+++ b/sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/package.scala
@@ -102,7 +102,7 @@ package object local {
 
   def fsInterpret: SparkFSConf => (FileSystem ~> Free[Eff, ?]) =
     (fsConf: SparkFSConf) => interpretFileSystem(
-      corequeryfile.chrooted[Eff](queryfile.input, FsType, fsConf.prefix),
+      corequeryfile.chrooted[Eff](FsType, fsConf.prefix),
       corereadfile.chrooted(fsConf.prefix),
       writefile.chrooted[Eff](fsConf.prefix),
       managefile.chrooted[Eff](fsConf.prefix))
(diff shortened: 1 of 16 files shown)


sparkcore/src/test/scala/quasar/physical/sparkcore/fs/PlannerSpec.scala
```diff
@@ -592,21 +595,32 @@ class PlannerSpec
...
-      sc.parallelize(List())
-    }
+  private def emptyFF[S[_]](sc: SparkContext)(implicit
+    S: Task :<: S
+  ): AFile => Free[S, RDD[Data]] =
+    (file: AFile) => lift(Task.delay {
+      sc.parallelize(List.empty[Data])
+    }).into[S]
+
+  private def first[S[_]](implicit
```
jsantos17: This function is a duplicate of [`first` in the implementation](https://github.com/quasar-analytics/quasar/pull/2660/files#diff-5fe27a8c6c445b3d97511877e2258a56R97)? Can we use the function from the implementation instead of duplicating it here?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/local/package.scala
```diff
@@ -102,7 +102,7 @@ package object local {
 
   def fsInterpret: SparkFSConf => (FileSystem ~> Free[Eff, ?]) =
     (fsConf: SparkFSConf) => interpretFileSystem(
-      corequeryfile.chrooted[Eff](queryfile.input, FsType, fsConf.prefix),
+      corequeryfile.chrooted[Eff](FsType, fsConf.prefix),
```
jsantos17: :+1:

PR #2666, opened 2017-08-23
Support left shifting on one side of a join condition.

### LINKED ISSUE #2110: Handle shifting inside a join condition in qscript

alissapajer:
Fixes #2110 

Handles the case when a join condition compiles into a `LeftShift`.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/ReplaceMapFunc.scala
+++ b/connector/src/main/scala/quasar/qscript/ReplaceMapFunc.scala
@@ -30,19 +30,19 @@ trait ReplaceMapFunc[T[_[_]], F[_]] {
 }
 
 trait ReplaceMapFuncInstances {
-  type CoMapFuncR[T[_[_]]] = CoEnv[Hole, MapFunc[T, ?], FreeMap[T]]
+  type CoMapFuncR[T[_[_]], A] = CoEnv[A, MapFunc[T, ?], FreeMapA[T, A]]
 
-  private def replaceJoinSide[T[_[_]]](name: Symbol, repl: FreeMap[T])
-      : CoMapFuncR[T] => State[Boolean, CoMapFuncR[T]] =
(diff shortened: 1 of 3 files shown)


djspiewak:
@alissapajer Could you perhaps explain a bit how this works?  I'm having a hard time parsing the code itself, and it's not clear to me (conceptually) how this case could be addressed at all.

alissapajer:
@djspiewak sure, so an example query that this PR fixes is
```sql
select zips1.city as city1, zips2.city as city2 from zips1 join zips2 on zips1.loc[*] = zips2.pop
```
When this is compiled from `LogicalPlan` to `QScript` the join condition `z1.loc[*] = z2.pop` will compile to a `LeftShift`. So to clarify, when we hit the `lp.Join` case in the `QScript` compiler, the join condition in that `LogicalPlan` is represented as a `LeftShift`. 

This PR extracts the relevant bits of information from that `LeftShift`, creating a single `ThetaJoin` using information from the `LeftShift` in its join condition. We move the shifting itself into either the left or right branch of the `ThetaJoin`, depending on which side in the `JoinCondition` was being shifted. Additionally, we expose access to the `LeftSide` and the `RightSide` of the `LeftShift` that is now a join branch (exposing identities, the shifted values, and the source), so that we can use the shifted values in the actual join condition, as well as have access to the original values.

Half of the diff is boilerplate-ish because we don't know ahead of time if the the left side or right side values are the ones being shifted, so we have to try the second case if the first one fails.

Issue #2502, opened 2017-07-18
Support NOW() function in mongo


alissapajer:
Per discussion here https://github.com/quasar-analytics/quasar/issues/2198#issuecomment-316169783 adding this issue to be completed in time for the new mongo release.

alissapajer:
The `NOW()` function exists as a `MapFunc` and so it only needs to be implemented in mongo.

rintcius:
New mongo supports `NOW()` already, but only on mapReduce, not on aggregation pipeline.
I suspect that it's not on the aggregation pipeline because it is represented by [`new Date()` in Bson](https://stackoverflow.com/questions/16455528/mongodb-aggregation-framework-date-now), but the java driver only supports constructing [these with an argument](http://mongodb.github.io/mongo-java-driver/3.4/javadoc/org/bson/BsonDateTime.html), not without an argument.

I _think_ it's possible to add support for this in agg pipeline as well by registering a [codec](http://mongodb.github.io/mongo-java-driver/3.4/bson/codecs/) but before starting implementation I want to double check if we want to go this road.

rintcius:
Added IT [here](https://github.com/rintcius/quasar/tree/2502-mongo-now) on top of @sellout's new mongo branch

jdegoes:
In theory, in batch mode, `NOW()` can be replaced with the current date / time before the query is handed off to the backend.

rintcius:
@jdegoes Yes, indeed I was thinking about that as well. I was thinking that it may even be better, e.g. in case we are doing cross-server queries (which lead me to think that it would be nice to have query hints for these kind of things). 

BTW there are currently different aliases for `now`. We also choose one of them to have that meaning instead:

```
  CIName("clock_timestamp")         -> date.Now, // Postgres (instantaneous)
  CIName("current_timestamp")       -> date.Now, // *, SQL92
  CIName("getdate")                 -> date.Now, // SQL Server
  CIName("localtimestamp")          -> date.Now, // MySQL, Postgres (trans start)
  CIName("now")                     -> date.Now, // MySQL, Postgres (trans start)
  CIName("statement_timestamp")     -> date.Now, // Postgres (statement start)
  CIName("transaction_timestamp")   -> date.Now, // Postgres (trans start)
```

zeeiyerWork:
@rintcius - reassigned this to @jsantos17 -> since he has a little bit more time in August.

rintcius:
[...]

PR #2668, opened 2017-08-24
Implement Mongo NOW()

jsantos17:
Addresses #2502


jsantos17:
@rintcius is something like this what you had in mind?

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -98,21 +102,21 @@ object MongoDbPlanner {
...
     fm.cataM(
       interpretM[M, MapFunc[T, ?], A, Fix[ExprOp]](
         recovery(_).point[M],
         expression(funcHandler)))
 
   def getSelector
-    [T[_[_]]: BirecursiveT: ShowT, M[_]: Monad, EX[_]: Traverse]
+    [T[_[_]]: BirecursiveT: ShowT, M[_]: Monad: FileSystemErr, EX[_]: Traverse]
     (fm: FreeMap[T])
-    (implicit merr: MonadError_[M, FileSystemError], inj: EX :<: ExprOp)
```
jsantos17: Did some refactoring making these implicits view bounds instead. I'm itching to make the same refactor in the rest of this file. Aye or nay?

jsantos17: Thanks. I always mix up the names.

mongodb/src/main/scala/quasar/physical/mongodb/planner/JsFuncHandler.scala
```diff
@@ -347,8 +347,6 @@ object JsFuncHandler {
               case TemporalTrunc(Year, date) =>
                 dateZ(year(date), litNum(1), litNum(1), litNum(0), litNum(0), litNum(0), litNum(0))
 
-              case Now() => Call(ident("ISODate"), Nil)
```
jsantos17: This was preventing the two `handleSpecialCore` code paths being hit. I deleted it because it looked like a stub impl. Is that reasonable @rintcius? Not familiar with the wider connector arch just yet.

jsantos17:
This is ready for another look @rintcius

mongodb/src/main/scala/quasar/physical/mongodb/planner/JsFuncHandler.scala
rintcius: This should be deleted indeed, although I don't think it's a stub impl. 

The reason why it should be deleted is that otherwise this would generate a different timestamp compared to agg. pipeline (because for JS  it would then be generated on mongo server instead of JVM/quasar)

jsantos17: Yes, that's the behavior I was seeing. thanks.

rintcius:
@jsantos17 I'll have a proper look tomorrow, but I think it'd be good to add a `now` test to `StdLibSpec`. Also an `.test` IT maybe?

jsantos17:
@rintcius StdLibSpec! That's definitely important! I'll see if I can come up with a useful `.test` too.

rintcius:
@jsantos17 Feel free to come up with something better, but I had something like this as `.test`: `"select now() > ts as past, ts - now() > 0 as future from `../days`"`

djspiewak:
Needs an assignee and a version label.

it/src/main/resources/tests/temporal/nowTransparent.test
```diff
@@ -0,0 +1,16 @@
...
+    "marklogic_json": "pending"
+  },
+
+  "data": "../zips.data",
+
+  "query": "select sum(when) from ((select date_part(\"milliseconds\", now()) as when, pop from `../zips` order by pop desc limit 1) union (select -date_part(\"milliseconds\", now()) as when, pop from `../zips` order by pop asc limit 1)) as times",
+
+  "predicate": "exactly",
+
+  "expected": [0.0]
```
jsantos17: i.e the `now()` result in the left hand side of the `UNION` is the same as the `now()` from the right hand side.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -70,7 +71,18 @@ object MongoDbPlanner {
 
   type Partial[T[_[_]], In, Out] = (PartialFunction[List[In], Out], List[InputFinder[T]])
 
-  type OutputM[A] = PlannerError \/ A
+  type OutputM[A]          = PlannerError \/ A
+  type PhaseW[F[_]]        = MonadTell_[F, PhaseResults]
+  type FileSystemErr[F[_]] = MonadError_[F, FileSystemError]
```
rintcius: similar here, you can reuse `MonadFsErr` from `quasar.common.fs`

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -159,40 +169,46 @@ object MongoDbPlanner {
...
+    def execTime(implicit ev: ExecTimeR[M]): M[Bson.Date] =
+      OptionT[M, Bson.Date](ev.ask.map(Bson.Date.fromInstant(_)))
+        .getOrElseF(raiseErr(
+          qscriptPlanningFailed(InternalError.fromMsg("Could not get the current timestamp"))))
+
+
     val handleSpecialCore: MapFuncCore[T, Fix[ExprOp]] => M[Fix[ExprOp]] = {
       case Constant(v1) => ejsonToExpression[M, T[EJson]](v1)
-      case Now() => unimplemented[M, Fix[ExprOp]]("Now expression")
+      case Now() => execTime map (bsn => Fix($literalF[ExprOp, Fix[ExprOp]](bsn)))
```
rintcius: it should be possible to express this with `$literal` instead of `$literalF` (see eg line 236)

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
jsantos17: Didn't see those! thanks!

jsantos17:
@rintcius I attempted to write a `StdLibSpec` test for `NOW()` but couldn't figure out how to write it without coupling the entire `StdLibSpec` to the Mongo connector. For connectors that have `NOW()` implemented in Quasar and passed as a literal to the backend (i.e Mongo after this PR), it's possible to know what `NOW()` will return and hence assert on it.

Some other connectors have correct `NOW()` implementation such as Marklogic XML/JSON

rintcius:
Yeah, a `stdlib` test for `now` is less useful compared to tests for pure functions and indeed I don't think we should do anything specific for Mongo in the `stdlib` tests or changing the framework for it.

Maybe a last thing to try is to see if something like this works out? pseudocode: `nullary(now() - now(), 0)`

jsantos17:
@rintcius I attempted the `now() - now()` approach. It does work and the result is indeed zero. However, it turns out that the Mongo connector returns timestamps as `Data.Dec`. Marklogic returns the timestamps as `Data.Interval`. Marklogic's type is correct. I tried to fix the Mongo connector, but it looks like BSON is not rich enough to represent intervals. When 
Mongo returns an interval, it does not include enough information for us to recognize as an interval. Hence the tests are red.

rintcius:
@jsantos17 Ah, good to know! I wouldn't try fixing that as part of this PR. I'm happy if you just mark mongo as pending with a comment that it returns 0 instead of an interval.

BTW: "returns timestamps as Data.Dec" should be "returns difference of 2 timestamps as Data.Dec", I think?

jsantos17:
@rintcius sounds good. I'll disable the test for Mongo to get this green.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -656,6 +656,14 @@ abstract class StdLibSpec extends Qspec {
         }
       }
 
+      "Now" >> {
+        import MathLib.Subtract
+
+        "now" >> prop { (_: Int) =>
+          nullary(Fix(Subtract(Fix(Now()), Fix(Now()))), Data.Interval(Duration.ZERO))
```
jsantos17: @rintcius Any idea why the implicit conversion to add `.embed` here doesn't work? I had to use `Fix` directly. I'd like to write this as `Subtract(Now().embed, Now().embed).embed`, but the implicit conversion is not triggering.

wemrysi: @jsantos17 Try changing the calls to `Now()` to `Now[Fix[LogicalPlan]]()`, without it they're likely inferring as `LogicalPlan[Nothing]` causing the implicit resolution to fail.

jsantos17: @wemrysi  That worked! Thanks!

jsantos17:
@rintcius will close this and re-open a new PR with a cleaner history once #2700 is merged.

jsantos17:
Closing in favor of #2704

PR #2671, opened 2017-08-24
Enable distinct test with correct expectation.

alissapajer:
I don't know why this integration test expectation got changed. This is the commit it got changed in https://github.com/quasar-analytics/quasar/commit/518dc93b61b337da23bbbd35237d60ec7578325e. SQL semantics along with the other integration tests confirm that the new expectation (which is also the expectation on master) is the one we want.

### DIFF
--- a/it/src/main/resources/tests/citiesByLargestZip.test
+++ b/it/src/main/resources/tests/citiesByLargestZip.test
@@ -2,23 +2,12 @@
     "name": "cities with largest individual zip codes",
     "backends": {
         "couchbase":      "pending",
-        "marklogic_json": "pending",
-        "marklogic_xml":  "pending",
-        "mimir":          "pending",
         "mongodb_2_6":    "pending",
         "mongodb_3_0":    "pending",
         "mongodb_3_2":    "pending",
         "mongodb_3_4":    "pending",
-        "mongodb_read_only": "pending",
-        "spark_hdfs":     "pending",
-        "spark_local":    "pending"
+        "mongodb_read_only": "pending"
     },
-    "NB": "Skipped for all connectors due to reduce/sort ordering bug.
-           Bug is fixed on @sellout's new mongo branch, but the fix breaks old mongo.
-
-           This is tricky because the sort key does not appear in the result.
-           The result can be verified by running a similar query with
-          `select city, pop`.",
 
     "data": "zips.data",
 
@@ -26,9 +15,9 @@
 
     "predicate": "exactly",
 
-    "expected": ["CHICAGO",
-                 "BROOKLYN",
-                 "NEW YORK",
-                 "BELL GARDENS",
-                 "LOS ANGELES"]
+    "expected": [{ "city": "CHICAGO" },
+                 { "city": "BROOKLYN" },
+                 { "city": "NEW YORK" },
+                 { "city": "BELL GARDENS" },
+                 { "city": "LOS ANGELES" }]
 }


alissapajer:
@rintcius yeah you're right. I really don't like this implicit `select value` though because it's not consistent with a query like `select foo, bar from baz` which will return an object with two fields. I think the user should have to explicitly `select value` in their query if they want to return the unwrapped values.

alissapajer:
The reason this particular query is returning the wrapped values is because of implementation details wrt the `order by`. So the correct thing to do for now is to fix this test to be consistent with a version of the query without `order by`. But I would like to revisit our previous decision on implicit `select value` at some point.

alissapajer:
To note, we have other tests on both master and next-major https://github.com/quasar-analytics/quasar/blob/next-major/it/src/main/resources/tests/shortCities.test which inconsistently expect results to be wrapped in a field name. So we need to audit all these tests and expect a consistent result.

alissapajer:
@rintcius so I'm thinking about reopening this. On master we have an inconsistency where tests like `shortCities.test` and `citiesByLargestZip.test` returned object-wrapped results, while tests like the ones you pointed out return just the values. `citiesByLargestZip.test` is returning correct results for many connectors, modulo the object-wrapping. So rather than change the test expectation away from the one on master now, what do you think about keeping the test expectation the same as it is in maste and then later (after next-major is merged into master) we have a discussion with @jdegoes to determine a sane behavior that we will enforce consistently.

rintcius:
@alissapajer Yeah fixing this consistently after next-major has been merged into master & discussing with John sounds good.
With respect to what to do right now, maybe just open an issue & spend minimum dev time now?

alissapajer:
@rintcius that sounds good to me. I'll reopen this since on master this test expects the values to be wrapped.

jdegoes:
The argument for implicit `SELECT VALUE` is roughly as follows:

Ordinary SQL `SELECT` had no ability to return anything other than a table, so if you issued some query like `SELECT COUNT(*) FROM foo`, you always got the result boxed in an ordinal column:

```
-----
| 0 |
-----
| 3 |
-----
```

SQL2 has no such limitations, because it can return single values, sets of values, or traditional "records" (string-valued maps).

More than that, however, there are two obvious cases in which auto-boxing combined with SQL2 semantics produces extremely confusing results.

The first case is when users generate records directly. For example:

```
SELECT {"foo": foo, "bar": bar, "type": "baz"} FROM table
```

In this case, the user is explicitly constructing a record, and yet the entire result set is boxed in "0".

The other, far more common use case is in _any_ sub select. For example, a sub select with `IN`:

```
... WHERE foo IN (SELECT bar FROM baz)
```

This class of queries, which work in ANSI SQL, fail in SQL2 with auto-boxing, because the semantics of SQL2 allow `IN` to work with records, so the fact that the sub select results in a dataset of records is not sufficient reason to auto-unbox (whereas, in SQL, all values are primitives so it can magically auto-infer this semantic).

A similar situation results in many other sub selects.

One can either require `SELECT VALUE` in such cases as these, which breaks compatibility with SQL in a way that causes simple SQL queries to not work at all in SQL2, or one can do implicit `SELECT VALUE`, which causes all sub select queries to work, but at the cost of sometimes unexpectedly returning data that is unboxed with the ordinal `0`.

The reason I argue for implicit `SELECT VALUE` is that the user can easily restore the SQL semantic by using `AS`, e.g. ``SELECT foo AS `0` FROM bar``, or by using JSON notation.

In addition, I'd further make the case that interpreting `AS` and `,` as row-builder expressions means that `a, b` or `a AS b` has a different semantic than `a`. i.e. the former _should_ result in a record, because a keyword or column concatenation operator is being used, whereas the latter should _not_ result in a record.

Those are just my initial thoughts though, I'm open to what others think on this issue.

rintcius:
A part of me likes the explicit `SELECT VALUE` (assuming it would be `SELECT VALUE foo FROM bar`), because that's a more principled solution when considering SQL2 in isolation (but yeah the disadvantage of breaking compat. with SQL is quite big).

Sounds unfortunate that we can't auto-unbox when doing sub-selects, since it looks like that could have opened the door to a more satisfactory solution (not breaking compatibility with SQL for sub selects & not unboxing when doing a select of a single column).

Makes me curious, what's the semantics of `IN` for records (and why)? Is it an essential part of how SQL2 operates?

FWIW I did a bit of experimentation:
```
// SETS
// works as expected
select foo from bar where foo in (1, 2)
// excerpt from generated mongo query: 
//  { "$match": { "foo": { "$in": [NumberInt("1"), NumberInt("2")] } } }

// ARRAYS
// I expected this to work in the same way, but it didn't give any results. 
select foo from bar where foo in [1, 2]
// excerpt shows that's it's wrapping the array in another `[..]`:
//  { "$match": { "foo": { "$in": [[NumberInt("1"), NumberInt("2")]] } } },

// RECORDS
// Was expecting this to not compile (if you wouldn't have said otherwise ;) )
select foo from bar where foo in {"baz": 42}
// also wrapping in `[...]` for records:
//  { "$match": { "foo": { "$in": [{ "baz": NumberInt("42") }] } } }
```

PR #2673, opened 2017-08-24
Implement Within function in Mimir

### LINKED ISSUE #2669: Implement Within for Mimir
This is holding up a number of integration tests, and it's a relatively common SQL operator.  We should absolutely have this for 4.3.  It's also not difficult to do.

We should add tests to `StdLibSpec` for this when we add it.

djspiewak:
Fixes #2669 

I had to add `Within` as a first-class `TransSpec`, since it is effectively performing `Equals` repeatedly over each array index.  Since it just delegates to `equalsImpl`, there really isn't a lot of code here.  The trickiest bit is just figuring out which things to materialize and what they look like after you're done.


alissapajer:
@nicflores can you take a look at this?

nicflores:
Sure thing Alissa.

PR #2684, opened 2017-08-28
Initial PostgreSQL connector

kciesielski:
This is the first part of PostgreSQL connector code. It's also my first take on a bigger contribution to Quasar :)
Scope: read file + write file fully implemented, manage file + query file only minimally so that integration tests for read/write can run ("delete" and "ls").
Some remarks:
- I am using JSON field type. Another option would be to map `Data` structure to flat tables, which is a quite complex task. After discussing options with @jdegoes I think it's better to stick with the simpler approach now.
- The RDBMS backend includes `ConnectionIO` as a part of it's `Eff`ect. I'm not sure whether this is correct, because `ConnectionIO` is a `Free[]`, so this means our `M[A]` is now a free monad with a "nested" free monad as a part of its algebra. I thought of using `ConnectionOp` instead, but it seems that there's no easy way to build an interpreter for this. Doobie offers Transactors, which can execute a `ConnectionIO`. Internally, this manages connections, transactions, etc. If we wanted to return `ConnectionOp` as a part of `Eff`, we would need to rewrite that.
- Postgres docker container for tests is exposed on port 35432 in order to avoid conflicts with the metastore container.
- The "read" implementation may be slow with this impl. On `open()` it just reads all the data and stores it, then returns stored records in `read()`. This may be improved by reading in batches.


djspiewak:
I don't want this to stall out completely on the shores of our release crunch.  Going to assign it to myself.  Any of the other team members should feel free to take it if they feel so motivated.

@kciesielski When you get a chance, could you merge in master to resolve conflicts?  Related to that is #2756, which will make some changes that will affect every backend, including this one, so you may want to hold off until it lands.

I'll carve out some time to give this a proper review some time next week!

kciesielski:
@djspiewak I rebased with master (will be merging from now on). I can see a failed test for `mongodb_3_0`, but this may just be another flakey false alarm. I don't have permissions to trigger rebuild though.

djspiewak:
@kciesielski I'll trigger a rebuild.  As a process note for future, unless the PR (ideally and the branch) is marked as WIP in some way, you shouldn't rebase.  Rebasing is fine for local and WIP work, but anything that is non-WIP and in a PR should be merged, otherwise it becomes hazardous to use other people's work.  It's definitely not going to cause a problem on this one, just letting you know for future.  :-)

kciesielski:
@djspiewak Thanks, I got told the same by @rabbitonweb. I just rebased exceptionally, knowing that there was no review in progress (ok, "hoping that..." ;) )  I'll be merging after removing `[WIP]` from now on.

build.sbt
```diff
@@ -420,10 +420,20 @@ lazy val skeleton = project
...
+  .settings(githubReleaseSettings)
+  .settings(libraryDependencies ++= Dependencies.rdbmscore)
+  .settings(isolatedBackendSettings("quasar.physical.rdbms.fs.postgres.Postgres$"))
+  .enablePlugins(AutomateHeaderPlugin)
+
 /** Implementation of the Spark connector.
   */
 lazy val sparkcore = project
-  .settings(name := "quasar-sparkcore-internal")
+  .settings(name := "quasar-4-internal")
```
djspiewak: 4?

interface/src/main/scala/quasar/main/package.scala
```diff
@@ -74,23 +74,23 @@ package object main extends Logging {
...
-   * It won't misbehave, but it will waste resources if you run it multiple
-   * times.  Thus, all uses of the value from this Task should be handled
-   * by `Read[BackendDef[PhysFsEffM], ?]` (or an analogous `Kleisli`).
-   */
+    * The physical filesystems currently supported.  Please note that it
+    * is really best if you only sequence this task ''once'' per runtime.
+    * It won't misbehave, but it will waste resources if you run it multiple
+    * times.  Thus, all uses of the value from this Task should be handled
+    * by `Read[BackendDef[PhysFsEffM], ?]` (or an analogous `Kleisli`).
+    */
```
djspiewak: It looks like there's a lot of reformatting in this file, all of it whitespace, and most of it not actually in accordance with the current-in-progress scalafmt style.  I generally don't see the need to include these whitespace changes, since as far as I can tell, you're not otherwise touching this file?  Would you mind reverting this file please?

it/src/test/scala/quasar/TestConfig.scala
```diff
@@ -21,11 +21,12 @@ import quasar.contrib.scalaz._
...
 import quasar.fs._
 import quasar.fs.mount.{BackendDef, ConnectionUri, MountConfig}
-import quasar.main.{ClassName, ClassPath, BackendConfig}
-
+import quasar.main.{BackendConfig, ClassName, ClassPath}
 import pathy.Path._
-import knobs.{Required, Optional, FileResource}
-import scalaz._, Scalaz._
+import knobs.{FileResource, Optional, Required}
+
```
djspiewak: I hate it, but our convention is to *not* have a newline here.  Import formatting guidelines are on the wiki, I believe.

it/src/test/scala/quasar/TestConfig.scala
```diff
@@ -42,12 +43,12 @@ object TestConfig {
...
-   * This is an artifact of the fact that we haven't inverted the dependency between
-   * `it` and the connectors.  Hence, the redundant hard-coding of constants.  We
-   * should get rid of this abomination as soon as possible.
-   */
+    * External Backends.
+    *
+    * This is an artifact of the fact that we haven't inverted the dependency between
+    * `it` and the connectors.  Hence, the redundant hard-coding of constants.  We
+    * should get rid of this abomination as soon as possible.
+    */
```
djspiewak: Whitespace change here.

it/src/test/scala/quasar/TestConfig.scala
```diff
@@ -92,12 +94,12 @@ object TestConfig {
     * to select an interpreter for a given config.
     */
   def externalFileSystems[S[_]](
-    pf: PartialFunction[BackendDef.FsCfg, Task[(S ~> Task, Task[Unit])]]
-  ): Task[IList[SupportedFs[S]]] = {
+                                 pf: PartialFunction[BackendDef.FsCfg, Task[(S ~> Task, Task[Unit])]]
+                               ): Task[IList[SupportedFs[S]]] = {
```
djspiewak: More whitespace

it/src/test/scala/quasar/TestConfig.scala
```diff
@@ -92,12 +94,12 @@ object TestConfig {
...
     def fs(
-      envName: String,
-      typ: FileSystemType
-    ): OptionT[Task, Task[(S ~> Task, Task[Unit])]] =
+            envName: String,
+            typ: FileSystemType
+          ): OptionT[Task, Task[(S ~> Task, Task[Unit])]] =
```
djspiewak: Actually, there's just a lot of whitespace changes here.  They obfuscate the real changes (which only touched two lines in this file), and should be reverted.

it/src/test/scala/quasar/fs/FileSystemTest.scala
```diff
@@ -50,8 +50,8 @@ import scalaz.stream.Process
   *       filesystems would run concurrently.
   */
 abstract class FileSystemTest[S[_]](
-  val fileSystems: Task[IList[SupportedFs[S]]]
-) extends quasar.Qspec {
+                                     val fileSystems: Task[IList[SupportedFs[S]]]
+                                   ) extends quasar.Qspec {
```
djspiewak: As far as I can tell, nothing in this file is meaningfully changed by this PR, it's just whitespace reformatting.  Please revert.

rdbms/src/main/scala/quasar/physical/rdbms/Interpreter.scala
```diff
@@ -0,0 +1,52 @@
...
+import quasar.physical.rdbms.fs.SqlReadCursor
+import slamdata.Predef.Map
+
+import doobie.imports.Transactor
+import scalaz.concurrent.Task
+import scalaz.syntax.applicative._
+import scalaz.~>
+
+trait Interpreter {
+  this: Rdbms =>
```
djspiewak: Is there value in separating this from `Rdbms`?  I mean, the dependency is cyclic in either case.

rdbms/src/main/scala/quasar/physical/rdbms/Rdbms.scala
```diff
@@ -0,0 +1,104 @@
...
+  type Eff[A] = (
+      ConnectionIO :\:
+      MonotonicSeq :\:
+      GenUUID :\:
+      KeyValueStore[ReadHandle, SqlReadCursor, ?] :/:
+      KeyValueStore[WriteHandle, TablePath, ?]
+  )#M[A]
+
+  type QS[T[_[_]]] = QScriptCore[T, ?] :\: EquiJoin[T, ?] :/: Const[ShiftedRead[AFile], ?]
+  type Repr        = String // TODO define best Repr for a SQL query (Doobie Fragment?)
```
djspiewak: *trips*

🤔

Ok, so I guess that makes sense.  Though I think Doobie now has a composable SQL datum (it's like the `tsql` macro or something like that)?  My knowledge of Doobie is relatively limited here.

rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
```diff
@@ -0,0 +1,119 @@
...
+  val ReadFileModule: ReadFileModule = new ReadFileModule {
+
+    private def readAll(
+        dbPath: TablePath,
+        offset: Int,
+        limit: Option[Int]
+    ): ConnectionIO[Vector[Data]] = {
+      val streamWithOffset = (fr"select * from" ++ Fragment.const(dbPath.shows))
+        .query[Data]
+        .process
```
djspiewak: This brings to mind the interesting problems with using frameworks like Doobie together with `Free`: it becomes very difficult to represent streaming.  I wonder if you're going to have to make `Eff` something concrete down the line.  I don't see an immediate need, but it's a thought to keep in mind.

rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
```diff
@@ -0,0 +1,119 @@
...
+
+      val streamWithLimit = limit match {
+        case Some(l) => streamWithOffset.take(l)
+        case None    => streamWithOffset
+      }
+      streamWithLimit
+        .vector
+    }
+
+    private def toInt(long: Long, varName: String): Backend[\/[FileSystemError, Int]] =
```
djspiewak: Curly braces are your friend.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/Postgres.scala
```diff
@@ -0,0 +1,65 @@
...
+  val JdbcUriOuter = """(jdbc:postgresql://[^\?]*)(.*)$""".r
+  val UserPasswordParams = """^\?user=(.*)&password=([^&]*)(.*)$""".r
+  val OnlyUserParams = """^\?user=([^&]*)(.*)$""".r
+
+  private def parsingErr(uri: ConnectionUri) =
+    -\/(NonEmptyList(
+      s"Cannot extract credentials from URI [${uri.value}]. Expected format: $formatHint"))
+
+  override def parseConnectionUri(
+      uri: ConnectionUri): \/[DefinitionError, JdbcConnectionInfo] = {
```
djspiewak: I'm pretty sure you can do all of this parsing with `URI`, and there's no need to munge around with regular expressions.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/Postgres.scala
```diff
@@ -0,0 +1,65 @@
...
+import scalaz.syntax.either._
+
+object Postgres extends Rdbms with PostgresInsert with PostgresDescribeTable with PostgresCreateJsonTable {
+
+  override val Type = FsType
+
+  val driverClass = "org.postgresql.Driver"
+
+  val formatHint =
+    "jdbc:postgres://host:port/db_name?user=username(&password=pw)"
```
djspiewak: I wonder if there's a better place to put the password.  I don't have any great solutions here, but it just makes me a bit uncomfortable floating database passwords around in plain text (and storing them as such!).

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresCreateJsonTable.scala
```diff
@@ -0,0 +1,45 @@
...
+    val createSchema: ConnectionIO[Unit] = tablePath.schema match {
+      case DefaultSchema => ().point[ConnectionIO]
+      case CustomSchema(name) => (fr"CREATE SCHEMA IF NOT EXISTS" ++ Fragment.const(name)).update.run.map(_ => ())
+    }
+
+    createSchema
+      .flatMap(_ =>
+        (fr"CREATE TABLE IF NOT EXISTS"
+          ++ Fragment.const(tablePath.shows)
+          ++ fr"(data json NOT NULL)").update.run)
```
djspiewak: 💪

I'm still not convinced that this is the best approach.  I agree it's the right road to go down right now, but it completely gives up on using SlamData on any meaningful pre-existing data in an RDBMs, which seems like an awfully large sacrifice.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresCreateJsonTable.scala
```diff
@@ -0,0 +1,45 @@
...
+      .map(_ => ())
```
djspiewak: `.void`

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/package.scala
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package quasar.physical.rdbms.fs
+
+import quasar.fs.FileSystemType
+
+package object postgres {
+
+  val FsType = FileSystemType("postgres")
```
djspiewak: Shouldn't this be inside the `Postgres` object?

rdbms/src/main/scala/quasar/physical/rdbms/Rdbms.scala
kciesielski: This is definitely not resolved yet. I left `String` here just to make `Rdbms` compile, I don't use `Repr` yet for read/write/manage file purposes. I'll revisit this when I start working on queries, maybe it will be Doobie `Fragment`, `Query0`, not sure yet.

rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
kciesielski: Doobie stream is a `scalaz.stream.Process[ConnectionIO, A]`. Calling `.vector` makes it a `ConnectionIO[Vector[A]]`. If I understand your point, the question is how to represent streams if we wanted to make the ReadFile API explicitly stream-driven? Which is "What type should be returned by `read`?"

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/Postgres.scala
kciesielski: I had exactly the same concern. I noticed that all backends follow this pattern. How about encrypting passwords in metastore and reading encryption key from an environment variable?

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresCreateJsonTable.scala
kciesielski: I always ended up writing my own `.toUnit`, and look at this, there's `.void`! Thanks!

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresCreateJsonTable.scala
kciesielski: Another approach would be to go for a similar thing to  [ToroDB internals](https://www.slideshare.net/8kdata/torodb-internals-how-to-create-a-nosql-database-on-top-of-sql-55275036). This means quite a sophisticated machinery for building intermediate tables/indices for representing nested documents and arrays. When I started to investigate this, it looked almost like "writing your own hibernate" ;) 
I also think that performance-wise using JSON fields won't take us far in BigData world, so we'll probably end up writing this mentioned complex mapper anyway.

build.sbt
kciesielski: No idea how this ended up here (some typo), fixing.

it/src/test/scala/quasar/TestConfig.scala
kciesielski: Right, fixing.

it/src/test/scala/quasar/TestConfig.scala
kciesielski: Reverting this whole file and leaving only two meaningful changed lines. This all is probably an effect of my experiments with scalafmt, I'll watch out more for those.

rdbms/src/main/scala/quasar/physical/rdbms/Interpreter.scala
kciesielski: Only for having less stuff to read in `Rdbms`. I think in the beginning I suspected that `interp` would be quite long.

rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
kciesielski: Sure, fixing.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/package.scala
kciesielski: Yes. It's probably a leftover from earlier situation when there might have been some risk of cyclic dependency. Moving to `Postgres` now.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/Postgres.scala
kciesielski: How about using `org.http4s.Uri.fromString()`? It's more powerful than plain `URI`, for example allows extracting params (user, password) as a `Map[String, Option[String]`, while `URI` gives only a String. However, I'm not sure if using http4s here is OK here?

kciesielski: Well, scrap that. I got deceived by IntelliJ, seems that http4s is not available as a dependency here, and adding it for parsing URIs doesn't seem like a good idea.

kciesielski:
@djspiewak I posted updates according to your review. It seems that there's a compilation error, which I don't get locally, even after a `clean`... I'll try to do a fresh clone to double-check this later today. If you have the branch on your machine, can you see if you're also getting this error?

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/Postgres.scala
djspiewak: The key management still has to go *somewhere*.  I think we would need to consider a more holistic solution if we wanted to fix this, since as you said, all the backends have this issue.

interface/src/main/scala/quasar/main/package.scala
djspiewak: This file still needs a revert.

djspiewak:
@kciesielski I think the build error is being caused by a stale cache.  I'll look into flushing those out this afternoon.  I similarly was not able to reproduce it locally.

kciesielski:
@djspiewak more fixes pushed, together with ManageFile implementation (move dir/file). Merged with master, build passes locally but again, something unknown happens on CI.

rintcius:
@kciesielski Looks like you are also hitting #2830 - I cleared the cache on Travis and restarted the build. Hopefully that resolves the CI issue

djspiewak:
@kciesielski Cool, merging into master!  Once more of the connector is in place w.r.t. integration tests, we can look into enabling the Travis build job.

PR #2686, opened 2017-08-28
Upgrade to Doobie 0.4.4

kciesielski:
Tests pass, but problably there's more to check to ensure that these changes don't break anything. 
I need this update to keep working on #2684

### DIFF
--- a/core/src/main/scala/quasar/db/package.scala
+++ b/core/src/main/scala/quasar/db/package.scala
@@ -17,10 +17,12 @@
 package quasar
 
 import slamdata.Predef._
-import com.zaxxer.hikari.HikariConfig
-import doobie.imports._
-import doobie.hikari.hikaritransactor.HikariTransactor
 
+import com.zaxxer.hikari.HikariConfig
+import doobie.free.connection.ConnectionIO
+import doobie.hikari.hikaritransactor.HikariTransactor
+import doobie.imports.DriverManagerTransactor
+import doobie.util.transactor.Transactor
 import scalaz._
 import scalaz.concurrent.Task
 
--- a/core/src/main/scala/quasar/db/package.scala
+++ b/core/src/main/scala/quasar/db/package.scala
@@ -17,10 +17,10 @@
 package quasar
 
 import slamdata.Predef._
-
 import com.zaxxer.hikari.HikariConfig
 import doobie.imports._
-import doobie.contrib.hikari.hikaritransactor._
+import doobie.hikari.hikaritransactor.HikariTransactor
+
 import scalaz._
 import scalaz.concurrent.Task
 
@@ -37,7 +37,7 @@ package object db {
     EitherT((for {
       xa <- HikariTransactor[Task](cxn.driverClassName, cxn.url, cxn.userName, cxn.password)
       _  <- xa.configure(config)
-    } yield StatefulTransactor(xa, xa.shutdown)).attempt.map(_.leftMap(e => metastore.UnknownError(e, "While connecting to MetaStore"))))
+    } yield StatefulTransactor(xa, xa.configure(_.close()))).attempt.map(_.leftMap(e => metastore.UnknownError(e, "While connecting to MetaStore"))))
 
   /** Transactor that does not use a connection pool, so doesn't require any cleanup. */
   def simpleTransactor(cxn: ConnectionInfo): Transactor[Task] =
--- a/interface/src/main/scala/quasar/main/Quasar.scala
+++ b/interface/src/main/scala/quasar/main/Quasar.scala
@@ -59,11 +59,11 @@ object Quasar {
     def qErrsToMainErrT[F[_]: Catchable: Monad]: QErrs ~> MainErrT[F, ?] =
       liftMT[F, MainErrT].compose(QErrs.toCatchable[F])
 
-    def toMainTask(transactor: Transactor[Task]): QErrsCnxIOM ~> MainTask = {
+    def toMainTask(transactor: Transactor[Task])(implicit tm: Monad[Task]): QErrsCnxIOM ~> MainTask = {
       val f: QErrsCnxIOM ~> MainErrT[ConnectionIO, ?] =
         foldMapNT(liftMT[ConnectionIO, MainErrT] :+: qErrsToMainErrT[ConnectionIO])
 
-      Hoist[MainErrT].hoist(transactor.trans) compose f
+      Hoist[MainErrT].hoist(transactor.trans(tm)) compose f
     }
   }
 
@@ -127,6 +127,7 @@ object Quasar {
       _          <- failedMnts.toList.traverse_(logFailedMount).liftM[MainErrT]
 
       runCore    <- CoreEff.runFs[QEffIO](hfsRef).liftM[MainErrT]
+
     } yield {
       val f: QEffIO ~> QErrs_CnxIO_Task_MetaStoreLocM =
         injectFT[Task, QErrs_CnxIO_Task_MetaStoreLoc]               :+:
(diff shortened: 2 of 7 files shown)


interface/src/main/scala/quasar/main/Quasar.scala
```diff
@@ -59,11 +59,11 @@ object Quasar {
     def qErrsToMainErrT[F[_]: Catchable: Monad]: QErrs ~> MainErrT[F, ?] =
       liftMT[F, MainErrT].compose(QErrs.toCatchable[F])
 
-    def toMainTask(transactor: Transactor[Task]): QErrsCnxIOM ~> MainTask = {
+    def toMainTask(transactor: Transactor[Task])(implicit tm: Monad[Task]): QErrsCnxIOM ~> MainTask = {
```
jedesah: I don't think this is desirable. Best to resolve the `Monad` instance for `Task` within this scope

interface/src/main/scala/quasar/main/Quasar.scala
```diff
@@ -127,6 +127,7 @@ object Quasar {
       _          <- failedMnts.toList.traverse_(logFailedMount).liftM[MainErrT]
 
       runCore    <- CoreEff.runFs[QEffIO](hfsRef).liftM[MainErrT]
+
```
jedesah: Odd newline

jedesah:
Looks like all the builds are failing though, so something is up...

jedesah:
@drostron I think it would be good for you to have a quick look at this to make sure I didn't miss anything

core/src/main/scala/quasar/db/package.scala
```diff
@@ -17,10 +17,10 @@
 package quasar
 
 import slamdata.Predef._
-
 import com.zaxxer.hikari.HikariConfig
 import doobie.imports._
-import doobie.contrib.hikari.hikaritransactor._
+import doobie.hikari.hikaritransactor.HikariTransactor
+
```
drostron: See [Scala[Quasar] Imports](https://github.com/quasar-analytics/quasar/wiki/Scala%5BQuasar%5D#imports) for these and others.

djspiewak: This is an issue that @alissapajer and have been meaning to re-raise with the team.  I was planning on discussing it after the scalafmt thing was behind us, but it's topical here…

Basically, we use `_` imports a *lot*, and I really don't think that's a good idea.  This is for three general reasons.  First, it's actually impacting compilation times.  Second, it becomes quite difficult for non-IDE/ENSIME users (so, most of us) to figure out where types are coming from.  Finally, Scala's relative imports mean that very large numbers of underscore imports are basically asking for issues, especially when we encourage relative nested imports (e.g. `import scalaz._, Scalaz._`).  That last point isn't a hypothetical either: this has been a real issue that has cost real time on a number of files.

I would much much rather be forced to be explicit with my imports (anything other than implicits, such as `import scalaz.syntax.monad._`).  It's a bit painful when first typing things out, but not as painful as it seems, and it makes files *much* more maintainable in the long term, it improves compilation time, and all-but eliminates issues caused by default-relative imports.

So in other words, I disagree with this review comment: I think the change should stand as it is.

kciesielski: Personally I prefer avoiding `_` imports, especially with Scalaz/cats due to compilation time impact. Good point with non-IDE users too. I'll review my work so far in order to keep such scheme.

jedesah: I think @drostron was just complaining about the newline on line 23 and not about the removal of the wildcard import 😄

drostron: Yep, thanks for clarifying @jedesah.

kciesielski: Thanks, I'll look into organizing imports too.

djspiewak: LOL! There goes my beautiful essay… :-P

kciesielski:
I managed to fix some dependency issues, and now tests seem to pass. However, I can't figure out why one of the build jobs fails here: https://travis-ci.org/quasar-analytics/quasar/jobs/269468912#L1830
Looks like the Quasar app exits prematurely?

djspiewak:
@kciesielski I bounced the job. Sometimes travis does weird things.  We'll see if it persists.

kciesielski:
@djspiewak thanks, but looks like the problem remains https://travis-ci.org/quasar-analytics/quasar/jobs/269468912#L1812

By the way - I just checked and Doobie 0.5.x and future versions will be cats-only...

jedesah:
@kciesielski Looks like the metastore might be causing Quasar to quit right after starting up. Have you tried running `web/run` in sbt to see if the server starts up properly?

kciesielski:
@jedesah I didn't check that. I'll verify, however I'm wondering why it happens only when `CONNECTOR=mongodb_2_6`

jedesah:
@kciesielski That specific travis job has extra responsibility such as making sure the app starts properly. Just a historical artifact of Mongo 2.6 being the first connector we supported

kciesielski:
@jedesah Looks like it starts and runs until I stop it. 
I initialized the metastore on PostgreSQL with these steps (it's my first time running the actual app):
- I configured `~/.config/quasar/quasar/quasar-config.json` to point to my local Postgres: docker container initialized with `docker/scripts/setupContainers`
- I initialized metastore using `java -jar quasar-repl-assembly-20.14.13.jar initUpdateMetaStore`
- I started `web/run`. Output:
```
[info] Running quasar.server.Server 
Read web config from /home/kc/.config/quasar/quasar-config.json
Using metastore: jdbc:postgresql://localhost:5432/metastore
Server started listening on port 8080
Press Enter to stop.
```
And I can call webservices locally, etc.

drostron:
@kciesielski I pulled down your branch and see the following error when using the default config which uses H2.

```
root:PR-2686> repl/run initUpdateMetaStore
[info] Compiling 1 Scala source to /Users/dave/code/quasar/.targets/core/scala-2.11/classes...
[info] Running quasar.repl.Main initUpdateMetaStore
Read core config from /Users/dave/Library/Application Support/quasar/quasar-config.json
Error: While initializing or updating: unexpected schema version; found: 1; latest known: 0. Is the metastore database running?
```

It is opaque, somewhat misleading, and I haven't tracked down the cause, but it is reproducible locally. You'll see the default config used when you remove your `quasar-config.json`. Like you, I don't see the error when running against PostgreSQL.

kciesielski:
@drostron I don't get such error with default settings. My output:
```
> repl/run initUpdateMetaStore
[info] Running quasar.repl.Main initUpdateMetaStore
Configuration file '/home/kc/.config/quasar/quasar-config.json' not found, using default configuration.
```
and then `~/.config/quasar/quasar-metastore.db.mv.db` gets silently created. Calling `web/run` then starts the app:
```
[info] Running quasar.server.Server 
Configuration file '/home/kc/.config/quasar/quasar-config.json' not found, using default configuration.
Using metastore: jdbc:h2:/home/kc/.config/quasar/quasar-metastore.db
Server started listening on port 20223
Press Enter to stop.
```

drostron:
@kciesielski NM and sorry for the confusion, I swore I had removed my H2 database, but indeed one from slamdata-backend was still around (so a reasonable error message after all). I now see the same behavior you're seeing locally. As for Travis, still unclear.

drostron:
I'm going to give our newly enabled Travis debug mode a whirl to see if it yields increased visibility.

djspiewak:
@kciesielski Regarding your note from here:

> By the way - I just checked and Doobie 0.5.x and future versions will be cats-only...

The same is also true of http4s, monocle, fs2, and other dependencies.  We'll be shimming back to scalaz at boundaries, with the goal of making it as transparent as possible internally.  If I can ever crack the implicit resolution issues, my [shims](https://github.com/djspiewak/shims) project will probably be the way to go there.  If not, we'll have some special-cased utilities within quasar which will at least be functional.

drostron:
The failure is a recent introduction and unrelated to this change, #2693.

drostron:
@kciesielski I added the `version: revision` label. For future PRs a version label is required by our merge script and is applied by the PR author, see [Versioning](https://github.com/quasar-analytics/quasar/wiki/Versioning).

PR #2687, opened 2017-08-28
[WIP] Scalafmt

djspiewak:
Alright everyone, bikeshed away!  The configuration is in `.scalafmt.conf`.  Poke around with your favorite files and see how they've been violated.

The exact infrastructure here is being provided by neo-scalafmt, but it is compatible with the various scalafmt incarnations (including the command line tool).  Format on compile is disabled, for reasons discussed in Slack.  Instead, we have a *test* that things are correctly formatted included in the CI build script, so we will fail-fast on the build if the formatting is not applied.  To apply formatting, ~~use the `scalafmt` or the `test:scalafmt` task within SBT.~~ just run the `format` alias within SBT.

If there is something that really needs to be formatted in a certain way (the parser is a good example) in violation of the rules, then you can do this by using `// format: off` and `// format: on` *around* the code block which you want to manually format.  Just be sure that the indentation of these two comments is the same, otherwise scalafmt will get confused.

Another thing disabled is formatting of `.sbt` files.  I don't really find the feature to be all that useful, but we could turn it on if desired.

This branch isn't going to land until *after* `next-major` does, so I haven't assigned a reviewer.  Enjoy!


djspiewak:
As a side note, it looks like there are some bugs in scalafmt (or possibly neo-scalafmt) regarding the formatting tests.  We may have to fiddle with the files in question a bit to get around these issues.

blueeyes/src/main/scala/quasar/blueeyes/json/serialization/DefaultDecomposers.scala
```diff
@@ -35,94 +35,152 @@ trait DefaultDecomposers {
...
   }
 
-  implicit def OptionDecomposer[T](implicit decomposer: Decomposer[T]): Decomposer[Option[T]] = new Decomposer[Option[T]] {
-    def decompose(tvalue: Option[T]): JValue = tvalue match {
-      case None    => JNull
-      case Some(v) => decomposer.decompose(v)
+  implicit def OptionDecomposer[T](
+      implicit decomposer: Decomposer[T]): Decomposer[Option[T]] =
+    new Decomposer[Option[T]] {
+
```
alissapajer: I'm personally not a fan of this extra new line here.

djspiewak: Neither am I.  It happens because of the forcing newline rewrite prior to top-level definitions.  The reason I added *that* is because there's plenty of code where that is really really required, and plenty of examples of that code where newlines were not inserted.  Basically, we have to pick our poison here.  :-(

blueeyes/src/main/scala/quasar/blueeyes/json/serialization/IsoSerialization.scala
```diff
@@ -74,71 +75,121 @@ object IsoSerialization {
...
+        dt: DecomposerAux[FT, T]) =
       new DecomposerAux[RichField :: FT, Option[H] :: T] {
+
         def decompose(fields: RichField :: FT, values: Option[H] :: T) = {
           val tail = dt.decompose(fields.tail, values.tail)
           // No point propagating decompose validation to the top level, we'd need to throw there anyway
-          values.head.map(h => tail.insert(fields.head.alts.head, dh.decompose(h)).fold(throw _, identity)).getOrElse(tail)
+          values.head
+            .map(
+              h =>
```
alissapajer: It looks like when it decides that the operations performed on `tail` are too long and need their own lines, it also gives a new line to `h =>`. Whereas above on line 123 we have `.map(h =>`. I prefer the formatting on line 123; is it possible to keep `h =>` on the same line as `.map(` in both situations?

blueeyes/src/main/scala/quasar/blueeyes/json/serialization/DefaultDecomposers.scala
alissapajer: What's an example of where the forcing newline rewrite prior to top-level definitions is really really required?

blueeyes/src/main/scala/quasar/blueeyes/json/serialization/IsoSerialization.scala
djspiewak: Yeah that seems a bit weird.  Perhaps a bug.  The fact that `map` is using a `(` rather than a `{` is probably the issue here.

blueeyes/src/main/scala/quasar/blueeyes/json/serialization/DefaultDecomposers.scala
djspiewak: The definition of `TransSpec.normalize` comes to mind.

alissapajer: What newline are you thinking of here? https://github.com/quasar-analytics/quasar/pull/2687/files#diff-95e17c270a1e733d0fd5bd12d5288c64L261

djspiewak: Oh wow, it didn't even work.  I'll just remove the rewrite, because it didn't address the situation I was thinking of.

djspiewak:
Alright, I removed the "newline before top level definitions" bit and reapplied the formatting (also rebased on top of master).  Additionally, I've added an alias command to sbt, `format`, which formats both test and main files.

connector/src/main/scala/quasar/connector/BackendModule.scala
```diff
@@ -157,18 +174,26 @@ trait BackendModule {
 
   type QS[T[_[_]]] <: CoM
 
-  implicit def qScriptToQScriptTotal[T[_[_]]]: Injectable.Aux[QSM[T, ?], QScriptTotal[T, ?]]
+  implicit def qScriptToQScriptTotal[T[_[_]]]: Injectable.Aux[
```
alissapajer: I think it's strange to start a new line (like in 178 and 179 below) for each type that is required to specify this type as a parameter. If this line is too long I would prefer to see a new line either before or after the `:`, so:
```scala
implicit def qScriptToQScriptTotal[T[_[_]]]:
    Injectable.Aux[QSM[T, ?], QScriptTotal[T ?]]
```

djspiewak: I generally agree. It might be possible. By default, scalafmt allows configuring the newline to be *before* the `:`, but I'm not sure if it allows configuring it after.  I disabled newlines before the `:`, which forces it into the brackets.

.scalafmt.conf
```diff
@@ -0,0 +1,32 @@
+# this roughly corresponds to half the screen of a MacBook Pro at default resolution with editor chrome
+maxColumn = 80
```
wemrysi: Anyone else interested in increasing this 25 or 50%?

alissapajer: My normal vim setup has 109 columns, so I'd be happy with anything up to 105.

connector/src/main/scala/quasar/connector/BackendModule.scala
alissapajer: If we configure it to have the newline before the `:` could it look like this?
```scala
implicit def qScriptToQScriptTotal[T[_[_]]]
    : Injectable.Aux[QSM[T, ?], QScriptTotal[T ?]]
```
I would be fine with that as well.

djspiewak: Ehhhhhh.  I've tried that, and it just seems incredibly ugly in most cases.

connector/src/main/scala/quasar/qscript/Coalesce.scala
```diff
@@ -143,191 +137,233 @@ trait CoalesceInstances {
...
+      implicit QC: QScriptCore :<: G): Coalesce.Aux[T, QScriptCore, G] =
     new Coalesce[QScriptCore] {
       type IT[F[_]] = T[F]
       type OUT[A] = G[A]
 
       // TODO: I feel like this must be some standard fold.
-      def sequenceReduce(rf: ReduceFunc[(IdStatus, JoinFunc)])
-          : Option[(IdStatus, ReduceFunc[JoinFunc])] =
+      def sequenceReduce(rf: ReduceFunc[(IdStatus, JoinFunc)]): Option[
+        (IdStatus, ReduceFunc[JoinFunc])] =
```
alissapajer: Here's another example. I fine this very had to read.

.scalafmt.conf
djspiewak: My editor has a column with of 96 for files shorter than 10,000 lines.  I'd be ok with larger than that, but wouldn't be as happy as I would be with 96 (or less).

connector/src/main/scala/quasar/qscript/Coalesce.scala
djspiewak: Agreed.  I'd rather see the type wrapped after the colon.

blueeyes/src/main/scala/quasar/blueeyes/HttpHeader.scala
```diff
@@ -34,27 +34,31 @@ sealed trait HttpHeaderField[T <: HttpHeader] extends ProductPrefixUnmangler {
...
 
-  val ByName: Map[String, HttpHeaderField[_]] = All.map(v => (v.name.toLowerCase -> v)).toMap
+  val ByName: Map[String, HttpHeaderField[_]] =
+    All.map(v => (v.name.toLowerCase -> v)).toMap
 
-  def parseAll(inString: String, parseType: String): List[HttpHeaderField[_]] = {
-    val fields = inString.toLowerCase.split("""\s*,\s*""").toList.flatMap(HttpHeaderField.ByName.get)
+  def parseAll(
+      inString: String,
+      parseType: String): List[HttpHeaderField[_]] = {
```
wemrysi: Curious if anyone has used this argument breaking style in practice? Did you ever feel like you were hunting for the return type?

I've preferred
```scala
def parseAll(
  inString: String,
  parseType: String
): List[HttpHeaderField[_]] = {
  ...
}
```
in the past, to distinguish the arguments from the return type. However, I've never used the proposed style, so it may work just as well once you get used to it.

https://github.com/quasar-analytics/quasar/blob/204ef34bbd84cdfdccb9c0d2a0c5b5298582ff11/frontend/src/main/scala/quasar/tpe/normalization.scala is an example of a file where the aforementioned style is used pervasively.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
```diff
@@ -76,11 +75,11 @@ object managefile {
 
   /** Run [[MongoManage]] with the given `MongoClient`. */
   def run[S[_]](
-    client: MongoClient
-  )(implicit
-    S0: Task :<: S,
-    S1: PhysErr :<: S
-  ): Task[MongoManage ~> Free[S, ?]] =
+      client: MongoClient
+  )(
```
alissapajer: I'd prefer something other than the `)(` on a line by themselves here. Something like
```scala
def run[S[_]](
    client: MongoClient)(
    implicit
    S0: Task :<: S,
    S1: PhysErr :<: S): Task[MongoManage ~> Free[S, ?]] =
```

blueeyes/src/main/scala/quasar/blueeyes/HttpHeader.scala
djspiewak: I have a couple objections to that proposed style.

- I hate dangling parentheses in all forms
- I like double-indenting arguments in order to visually distinguish them from the function body.  This is especially important with class parameters.  But double-indentation looks very strange if you're wrapping in newlined parentheses
- This style doesn't scale particularly intuitively to long type parameter blocks

I do like how your proposed style leaves more space for the return type, but at least to my eyes, that's the only advantage to it, and it comes at the cost of the common case (which is a smaller function signature with a simple return type which *just* manages to overflow the line limit in the parameter block).

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
djspiewak: Agreed. I much prefer this style.  Unfortunately, scalafmt is ignoring the configuration which is supposed to prevent the "`)(` on their own line" problem.

wemrysi: Maybe?
```scala
def run[S[_]](
  client: MongoClient
)(implicit
  S0: Task :<: S,
  S1: PhysErr :<: S
): Task[MongoManage ~> Free[S, ?]] =
```

connector/src/main/scala/quasar/qscript/Coalesce.scala
djspiewak: It looks like our only option is [before the colon](http://scalameta.org/scalafmt/#newlines.sometimesBeforeColonInMethodReturnType), rather than after.  It's probably preferable to breaking within the type, so I'll turn that on.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
djspiewak: Concrete example of why I dislike the single indent style:

```scala
// original
def foo[F[_]: Functor](fa: F[(String, Int)]): List[(String, Int, Boolean)] = fa.map(t => (t_1, t_2, false))

// double-indent:
def foo[F[_]: Functor](
    fa: F[(String, Int)]): List[(String, Int, Boolean)] =
  fa.map(t => (t_1, t_2, false))

// single-indent
def foo[F[_]: Functor](
  fa: F[(String, Int)]
): List[(String, Int, Boolean)] =
  fa.map(t => (t_1, t_2, false))
```

djspiewak: Another example, using class parameters:

```scala
// original
final case class Foo[F[_], A](fa1: F[A], fa2: F[(A, A)], fa3: F[(A, A, A)]) {
  def attempt = (fa1, fa2, fa3)
}

// double-indent
final case class Foo[F[_], A](
    fa1: F[A],
    fa2: F[(A, A)],
    fa3: F[(A, A, A)]) {
  def attempt = (fa1, fa2, fa3)
}

// single-indent
final case class Foo[F[_], A](
  fa1: F[A],
  fa2: F[(A, A)],
  fa3: F[(A, A, A)]
) {
  def attempt = (fa1, fa2, fa3)
}
```

frontend/src/main/scala/quasar/tpe/normalization.scala
```diff
@@ -33,62 +33,66 @@ object normalization {
...
   def liftConst[J, T](
-    implicit
-    TR: Recursive.Aux[T, TypeF[J, ?]],
-    JC: Corecursive.Aux[J, EJson],
-    JR: Recursive.Aux[J, EJson]
+      implicit
+      TR: Recursive.Aux[T, TypeF[J, ?]],
+      JC: Corecursive.Aux[J, EJson],
+      JR: Recursive.Aux[J, EJson]
   ): TypeF[J, T] => TypeF[J, T] = totally {
```
alissapajer: Here it appears to have left the type ascription on a new line, which is different than what it did here https://github.com/quasar-analytics/quasar/pull/2687#discussion-diff-136414921R45. So when the line is too long to fit the type ascription, it moves it to a new line. Otherwise it keeps it on the previous line.

Since it's possible to hit this case, where we newline the type ascription like this, then I think it would be more consistent if we _always_ newline the type ascription. Otherwise we will have to sometimes look on the previous and sometimes on the next line to see the new type, and that's not consistent.

blueeyes/src/main/scala/quasar/blueeyes/HttpHeader.scala
wemrysi: I'm not against double-indenting (wasn't an intentional omission) but I haven't ever tried it with the proposed style so it may be less compelling given the extra structure the function name and return type provide.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
alissapajer: @djspiewak in your first example block, what happens if the line is too long to support the type on a single line and it instead gets pushed to the next line (or we choose to always push it to the next line):
```scala
def foo[F[_]: Functor](
  fa: F[(String, Int)])
    : List[(String, Int, Boolean)] =
  fa.map(t => (t_1, t_2, false))
```

alissapajer: That way there are no dangling parens and the type is on a new line.

wemrysi: @alissapajer Don't you need to double-indent the argument in that case as well?

frontend/src/main/scala/quasar/tpe/normalization.scala
alissapajer: I would be fine avoiding the dangling paren and instead having 
```scala
    JR: Recursive.Aux[J, EJson])
  : TypeF[J, T] => TypeF[J, T] = totally {
```
or
```scala
  JR: Recursive.Aux[J, EJson])
    : TypeF[J, T] => TypeF[J, T] = totally {
```
I do like the second one better because then the type ascription is indented from the start of the function body that starts on the next line.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
alissapajer: @wemrysi I prefer the function arguments indented differently than the type ascription, when they are both on new lines. But I'd be fine if they were both double indented if that was the general preference.

djspiewak: @alissapajer Restoring the "sometimes before colon" setting to true would indeed yield a situation like what you describe, except with the parameters double-indented.

djspiewak: @alissapajer Restoring the "sometimes before colon" setting to true would indeed yield a situation like what you describe, except with the parameters double-indented.

frontend/src/main/scala/quasar/tpe/normalization.scala
djspiewak: @alissapajer So it's important to understand that scalafmt will sometimes break the rules (unless you use the settings which prevent this) in order to avoid extremely severe line penalties.  The line you cite is one of those situations, where scalafmt *wanted* to put the type and the parenthesis on the previous line, but the line width prevented that.  Since I formatted this hard-forbidding newlines before colons, it pushed the parenthesis to the subsequent line.  If I re-ran the formatter without that setting, the type ascription would be on the new line, indented, with the parenthesis no longer dangling.

blueeyes/src/main/scala/quasar/blueeyes/HttpHeader.scala
djspiewak: If you don't have a return type though…  It doesn't mesh well with type inference, because then you end up with a line which is simply `) =`

djspiewak:
Ok, I reapplied the formatting with `newlines.sometimesBeforeColonInMethodReturnType = true` and max line width set to 96.

frontend/src/main/scala/quasar/tpe/normalization.scala
alissapajer: That's exactly my point: that line width will sometimes push the type ascription to the next line. My suggestion was that we force type ascriptions to _always_ be on a new line so that the user can always look in the same place to find the type ascription.

djspiewak: Scalafmt doesn't have a setting to do that.

connector/src/main/scala/quasar/connector/BackendModule.scala
```diff
@@ -163,12 +169,14 @@ trait BackendModule {
   type M[A]
 
   def FunctorQSM[T[_[_]]]: Functor[QSM[T, ?]]
-  def DelayRenderTreeQSM[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT]: Delay[RenderTree, QSM[T, ?]]
+  def DelayRenderTreeQSM[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT]
+    : Delay[RenderTree, QSM[T, ?]]
```
alissapajer: This is better than breaking up the type on multiple new lines. :+1:

rintcius:
I'm guessing scalafmt is just subtly trying to encourage us to rewrite this code:
```
  def javascript[T[_[_]]: BirecursiveT: ShowT, M[_]: Applicative](
      implicit merr: MonadError_[M, FileSystemError]): AlgebraM[M, MapFunc[T, ?], JsCore] = {
    import jscore.{
      Add => _,
      In => _,
      Lt => _,
      Lte => _,
      Gt => _,
      Gte => _,
      Eq => _,
      Neq => _,
      And => _,
      Or => _,
      Not => _,
      _
    }
```

And there's this oddity of just `)(` on a line that @alissapajer already pointed out:

```
  def runF[S[_]](
      c: MongoClient
  )(
      implicit
      S0: Task :<: S,
      S1: PhysErr :<: S): Free[S, A] = {
    val mongoErr = Failure.Ops[PhysicalError, S]
    mongoErr.unattempt(free.lift(attemptMongo.run.run(c)).into[S])
  }
```

Generally looks fine to me.

djspiewak:
Currently blocked on finding a workaround for https://github.com/scalameta/scalafmt/issues/1055

djspiewak:
Ok, this is getting closer to reality.  Don't want to do anything until after we deliver 4.3, since there will be some disruption, but I think we're ready for a final-pass review and 👍 / 👎 

For the record, when this gets applied, basically all outstanding local work is going to get conflicted.  The solution to this will be to cherry-pick the configuration commit and then double-apply scalafmt to the entire local state.  Then, merge with master.  There really isn't a way to make it work with rebasing.

alissapajer:
I give :+1:.

alissapajer:
@beckyconning this PR was the initial discussion about having automatic code formatting. We had put this off while working on 5.0, but moving forward I'd like to see this prioritized. @djspiewak did the initial work and so could say the effort level to finish this.

djspiewak:
The biggest issue I see with landing it is how slow scalafmt is. :-( I keep thinking through how the workflow is going to be, and I can’t come up with good answers. I guess formatting before each commit? I want the standardized auto formatting as well, it just seems like the cost is going to be pretty high.

alissapajer:
@djspiewak for quasar, how long does it take to run the formatting once manually locally? We could run it not on every commit but just once for each PR. Personally, given how I rebase before PRing, requiring each commit formatted would be really annoying.

I do agree overall though that if we agree the overhead is too high, we shouldn't land this.

djspiewak:
@alissapajer Reformatting all of quasar takes about 3 minutes, if memory serves, starting from master state.  Running the formatter on all of quasar in the case that no files *need* to be reformatted (and the JVM is already warm) takes 17 seconds.  Some files are still structured such that the reformatting is O(k^n), a problem which can be fixed by not abusing Scala in some specific ways (breaking out subexpressions into `val`s fixes it almost every time; the issue is usually generated by huge blobs of multi-line function invocations with nested function calls and expressions).

alissapajer:
Hm, that's right on the edge. I'd be happy to try landing this, and if the overhead turns out to be too high, disable it. 

Thoughts from others?

alissapajer:
Closing as we'll open a new PR if we decide to do this. Discussion can continue on the issue qz-3340.

Issue #2681, opened 2017-08-26
QScript should use objects rather than arrays to carry context


djspiewak:
QScript currently builds a lot of arrays, which it then peels apart later in order to carry along information that it needs (often duplication, one side of which will be transformed while the other must be preserved).  The problem is when this introduces undefinedness, connectors are being implicitly expected to maintain sparse arrays throughout the computation pipeline, which is not always an assumption that backends can satisfy.  Objects do not have this flaw, as objects are sparse by definition, and thus qscript would not be making any assumption to use objects rather than arrays as its context structure.

So rather than something that builds an array `[.foo, .foo]` and later derefs `[0]` or `[1]`, it would build an object `{"0": .foo, "1": .foo}` and later deref `.0` or `.1`.

alissapajer:
Completed with the switch to QSU.

PR #2688, opened 2017-08-28
Merging master again more

djspiewak:
This brings in the latest from master into next-major, including some pretty significant mimir changes.  Hopefully this will be the last time we have to do a major merge in this direction.

I had to turn off several tests, which is a shame.  This is mostly because qscript is now even more aggressive about using arrays, and often building structures that have undefineds in those arrays (which mimir will pack away, breaking the query).  We already have a ticket out to track this (#2681).


djspiewak:
@alissapajer Looks like new Mongo doesn't have a `DeleteField`.  That seems like a pretty substantial limitation, given how often qscript generates those nodes on next-major.  Should we just merge despite the failing build?  Or would you rather pending out that test and make a strong TODO to fix it?

it/src/test/scala/quasar/physical/sparkcore/SparkStdLibSpec.scala
```diff
@@ -49,6 +49,7 @@ class SparkStdLibSpec extends StdLibSpec {
     case MFD(Trunc(_))           => TODO
     case MFC(Power(_, _))        => Pending("TODO: handle large value").left
     case MFC(ConcatArrays(_, _)) => Pending("TODO: handle mixed string/array").left
+    case MFC(DeleteField(_, _))  => Pending("TODO: deleting a non-existent field should be the original object, NOT undefined").left
```
alissapajer: This is a poor test framework in that we cannot disable individual tests for a specific `MapFunc`; we have to disable the entire `MapFunc` even though most of the tests for it are passing.

alissapajer:
@djspiewak new mongo does have `DeleteField` implemented here https://github.com/quasar-analytics/quasar/blob/next-major/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala#L562. It just doesn't have it implemented specifically for the aggregation pipeline (`ExprOp`) here https://github.com/quasar-analytics/quasar/blob/next-major/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala#L224 because, afaik, the aggregation pipeline doesn't support field deletion. It's possible that field deletion will be supported in mongo 3.4 or greater, but I haven't looked into it, maybe @rintcius would know. 

The test you had to disable for mongo is just the test for `ExprOp`. The map-reduce test for `DeleteField` is passing. @djspiewak why didn't you have to disable the `DeleteField` test for spark and mongo on master? There is the possibility that something changed for mongo on next-major, but nothing has changed for spark that I know of.

Also, you're correct that there are many `DeleteField` nodes are created initially in `QScript`, but backends are generally able to rewrite those away due to `QScript`'s "common source" design. Whereas the "linear" design of `LogicalPlan` forced us to have those `DeleteField`s in the first place (because we have to create a temporary field to sort by and then delete it). So few `DeleteField`s are actually making it as far as evaluation.

djspiewak:
The DeleteField tests were added in this merge; they don’t exist on master. I would guess that old mongo (and spark) has the same problem on master that it does here.

alissapajer:
@djspiewak ah, I didn't realize that any new commits were added after you merged in master, aside from the couple you pushed after I approved. Can you make note of that explicitly next time?

Mongo absolutely has the same problem on master, and I would assume that spark does too.

djspiewak:
Sorry I should have noted the changes more explicitly. There are others as well (I had to rewrite a lot of the sorting stuff in mimir from both master and next-major, for example). Next-major is now far enough away from master that pure merges don’t exist anymore, so basically any merge is going to give rise to new stuff just as a consequence of making things work again.

alissapajer:
Yeah, absolutely, it makes perfect sense you had to add more commits. And I do think it's best to have the  merge commit, and then add separate commits after that like you did, rather than stick all the additional changed into the merge commit; that way we see more clearly what these changes are.

alissapajer:
@djspiewak we are merged! Thanks for all your work on this; maintaining two parallel branches like this is not the most fun thing.

djspiewak:
@alissapajer I generally agree on that as well, but in this case it conflicted with my other maxim, which is that a working state merged into a working state should produce a working state in the merge commit.  The `DeleteField` tests (and the rewrite of the mimir sort stuff) were added to try to figure out why things *weren't* working in the merge commit.  I guess that perhaps I could have chunked up the commit after the fact, staging and committing things one logical unit at a time, but I'm not sure.  It's definitely getting quite tricky.

alissapajer:
@djspiewak given that we're maintaining two continually diverging branches, I don't think that in this particular case "a working state merged into a working state should produce a working state in the merge commit", although in general I think we should absolutely hold to that. This is unfortunately a special situation.

rintcius:
@alissapajer @djspiewak I'm not aware of doing a delete field on 3.4 on aggregation pipeline, however on 3.6 there's: https://jira.mongodb.org/browse/SERVER-27614

PR #2694, opened 2017-08-30
Moved blocking console read off of the main(!!) thread pool

### LINKED ISSUE #2693: `scripts/testJar` failing on mongodb_2_6 Travis jobs
Travis log
```
Quasar Version: 20.14.15
Initializing Quasar MetaStore...
log4j:WARN No appenders could be found for logger (com.couchbase.client.core.logging.CouchbaseLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Read web config from /home/travis/.config/quasar/quasar-config.json
Launching Quasar JAR...
Quasar PID: 12946
log4j:WARN No appenders could be found for logger (com.couchbase.client.core.logging.CouchbaseLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Read web config from /home/travis/.config/quasar/quasar-config.json
Using metastore: jdbc:h2:~/quasar-metastore.db
Server started listening on port 8080
Press Enter to stop.
Stopped last server listening on port 8080
Killing Quasar JAR...
/home/travis/build/quasar-analytics/quasar/scripts/testJar: line 38: kill: (12946) - No such process
The command "./scripts/build" exited with 1.
```

Running `scripts/testJar` succeeds locally, yet fails on Travis.

djspiewak:
Fixes #2693

Like I said, we need to be very very careful of this sort of stuff.  The `fork` here was doing exactly nothing, since it was keeping the action sequenced onto the main pool, resulting in eating a thread.  The fact that this is only failing with Mongo 2.6 is also suspicious, and suggests that the mongo driver is using blocking IO and we're not thread shifting it.  We need to take a hard look at how all of this stuff is working right now, because it's tremendously easy for us to paint ourselves into terrible corners without realizing it except in specific circumstances.

### DIFF
--- a/web/src/main/scala/quasar/server/PortChangingServer.scala
+++ b/web/src/main/scala/quasar/server/PortChangingServer.scala
@@ -43,8 +43,9 @@ final case class PortChangingServer(servers: Process[Task, (http4s.server.Server
   def shutdownOnUserInput: Task[Unit] = for {
     _ <- stdout("Press Enter to stop.")
     // TODO this ignores errors all over the place
-    _ <- Task.delay(Task.fork(waitForInput)(PortChangingServer.AwaitingPool).unsafePerformAsync(_ => shutdownImpl.unsafePerformAsync(_ => ())))
     _ <- servers.run
+    _ <- Task.fork(waitForInput)(PortChangingServer.AwaitingPool)
+    _ <- shutdownImpl
   } yield ()
 
   def shutdown: Task[Unit] =
--- a/web/src/main/scala/quasar/server/PortChangingServer.scala
+++ b/web/src/main/scala/quasar/server/PortChangingServer.scala
@@ -23,6 +23,8 @@ import quasar.server.Http4sUtils._
 
 import scala.concurrent.duration._
 
+import java.lang.{Runnable, Thread}
+import java.util.concurrent.{Executors, ThreadFactory}
 import org.http4s
 import org.http4s.HttpService
 import scalaz._, Scalaz._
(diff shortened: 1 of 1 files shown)


web/src/main/scala/quasar/server/PortChangingServer.scala
```diff
@@ -36,16 +38,30 @@ import scalaz.stream.{async, Process}
...
+  } yield ()
+
   def shutdown: Task[Unit] =
-    shutdownImpl *> servers.run // We need to run the servers in order to make sure everything is cleaned up properly
+    shutdownImpl >> servers.run // We need to run the servers in order to make sure everything is cleaned up properly
 }
 
 object PortChangingServer {
+
+  private val AwaitingPool = Executors.newFixedThreadPool(1, new ThreadFactory {
```
jedesah: We construct a `ExecutorService` this way in SB:

```
Executors.newSingleThreadExecutor(
      NamedDaemonThreadFactory("slamdata.backend-poll-shutdown"))
```

Ironically, because we don't use `PortChangingServer` in SB, we might be able to get rid of that thread pool over there but it does seem like a more concise way of constructing a `ExecutorService` though and seems like it would not require importing java stuff 😄

jedesah: I don't know if there is a way to specify thread priority this way, what made you decide low btw?

djspiewak: The priority doesn't make that much of a difference since it's just blocking.  The *non*-daemon nature of the thread *does* make a difference though, and so that alone prevents us from using `NamedDaemonThreadFactory`.

jedesah: Right, I would have thought `Thread.Max_PRIORITY`. Could we end up with an overloaded server that we can't shutdown because it's too busy on its other threads? But it is indeed a fairly academic question since like you say it's sole purpose is to block on user input

djspiewak: Even min priority threads are going to get scheduled eventually.  It's not a hard lockout.

web/src/main/scala/quasar/server/PortChangingServer.scala
```diff
@@ -36,16 +38,31 @@ import scalaz.stream.{async, Process}
...
+
   /** Wait for the command line user to press enter before shutting down the server */
-  def shutdownOnUserInput: Task[Unit] =
-    stdout("Press Enter to stop.") <*
-    Task.delay(Task.fork(waitForInput).unsafePerformAsync(_ => shutdownImpl.unsafePerformSync)) <*
-     servers.run
+  def shutdownOnUserInput: Task[Unit] = for {
+    _ <- stdout("Press Enter to stop.")
+    // TODO this ignores errors all over the place
+    _ <- servers.run
```
jedesah: Does this really work? I thought `servers.run` was blocking? I guess we'll see on travis but I'm a little dubious. Did you give it a whirl?

djspiewak: `servers.run` is not blocking

jedesah: Okay, Travis is happy so I'll merge but I don't understand why `servers.run` is non-blocking. How can an infinite process on which one calls `run` be non-blocking?

PR #2695, opened 2017-08-30
Increase request-timeout to deal with timeouts in Marklogic tests.

jsantos17:
This *should* prevent the timeouts we're seeing. 🤞


jsantos17:
This only moved the timeout from Marklogic to the test suite. The failing tests here are simply slow and timeout in five minutes. I guess I was just lucky when it fixed the problem in my branch.

PR #2698, opened 2017-08-31
[Fix #2679] Rephrase ObjectDelete in projection

### LINKED ISSUE #2679: TransSpec rephrase should support ObjectDelete
Title

This is made a lot more relevant by the fact that qscript in next-major generates *a lot* more `DeleteField` nodes.

edmundnoble:


djspiewak:
Added a feature label.  Also please remember to add some text describing the change, since that's what will end up in the changelog.

PR #2699, opened 2017-08-31
Mongo backendmodule

rintcius:


it/src/test/scala/quasar/physical/mongodb/fs/MongoDbFileSystemSpec.scala
```diff
@@ -261,13 +261,13 @@ class MongoDbFileSystemSpec
           shouldFailWithPathNotFound { path =>
             s"""SELECT name FROM `$path` WHERE LENGTH(name) > 10"""
           }
-        }
+        }.pendingUntilFixed("""error does not include file: -\/(QScriptPlanningFailed(NoFilesFound(List())))""")
```
rintcius: I am not sure if the error should be improved (and if so where that check happens now)

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
```diff
@@ -50,48 +53,20 @@ package object fs {
...
-                      .leftMap(_.right[NonEmptyList[String]])
-                      .run
-                  ).into[S])
-        close  =  free.lift(Task.delay(client.close()).attempt.void).into[S]
-      } yield BackendDef.DefinitionResult[M](fs, close)
-  }
+  def compile(cfg: MongoConfig): BackendDef.DefErrT[Task, (M ~> Task, Task[Unit])] =
+    (effToTask(cfg) map (i => (
+      foldMapNT[Eff, Task](i),
+      Task.delay(cfg.client.close).void))).liftM[DefErrT]
```
djspiewak: Lost an `attempt` here.  Was that intentional?

connector/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -38,7 +38,7 @@ object FileSystemError {
   import WriteFile.WriteHandle
 
   final case class ExecutionFailed private (
-    lp: Fix[LogicalPlan],
+    lp: Option[Fix[LogicalPlan]],
```
wemrysi: The intent of this error is to communicate why executing the specified `LogicalPlan` failed, so it shouldn't ever be absent.

Due to how things are structured in `BackendModule`, we could introduce a new error type local to `BackendModule` that implementations can use to communicate these sorts of errors that doesn't involve an `LP` and then `BackendModule` can convert these to `ExecutionFailed` by combining the local error type with its knowledge of the `LP`.

rintcius: @wemrysi  I see a couple of ways to tackle this:

1) add another transformer `BackendFsErrT` to `BackendT` where
   `type BackendFsErrT[F[_], A] = EitherT[F, BackendFileSystemErr, A]` and
   `BackendFileSystemErr` would contain a variant of `ExecutionFailed` without the `LP` parameter
2) change `BackendT` to `FsErrT[PhaseResultT[ConfiguredT[F, ?], ?], A]` where
   `type FsErrT[F[_], A] = EitherT[F, BackendFileSystemErr \/ FileSystemErr, A]`

These options above complicate the `Backend` stack though

3) change `BackendT` to `BackendFsErrT[PhaseResultT[ConfiguredT[F, ?], ?], A]`
   where `type BackendFsErrT[F[_], A] = EitherT[F, BackendFileSystemErr, A]` and
   `BackendFileSystemErr` duplicates all `FileSystemError` but without `LP`
   (or create a `BackendFileSystemErr(fs: FileSystemError)` instead to remove some of the duplication)
4) change `BackendT` to `BackendFsErrT[PhaseResultT[ConfiguredT[F, ?], ?], A]` where
   `type BackendFsErrT[F[_], A] = EitherT[F, Fix[LogicalPlan] => FileSystemError, A]`
5) add a new `FileSystemError` analogous to `ExecutionFailed`
   but with a `Repr` instead of a `Fix[LP]`, e.g.:
```
final case class BackendExecutionFailed[Repr: RenderTree] private (
  repr: Repr,
  reason: String,
  detail: JsonObject,
  cause: Option[PhysicalError]
) extends FileSystemError
```
(but not sure how/if prisms work with type parameters)

I personally like number 4 or number 5 (if that works), but maybe we can discuss in a call?

mongodb/src/main/scala/quasar/physical/mongodb/fs/package.scala
rintcius: Hmm.. the difference is propagating a Task failure (in the current code) vs swallowing it (when adding `attempt`), right? I _think_ it's ok to just propagate a failure here?

rintcius:
@wemrysi I removed some `MongoDb` cyclic deps. The ones that are still there (`fs.queryfile`, `fs.managefile`, `fs.writefile`, `fs.readfile`) are less evil I think, although if you prefer I could move them inside `MongoDb.scala`.

connector/src/main/scala/quasar/fs/FileSystemError.scala
wemrysi: I think 1) might be the best option in the near-term as it still allows us to perform the path-rewriting necessary to ensure the paths in the errors are meaningful to users (as an aside, I just noticed we aren't rewriting paths of `ExecutionFailed` in `quasar.fs.transformPaths`, though we should be).

Eventually, all the `Fix[LP]` will probably be replaced with a lub variant of QScript which should then obviate the need for this new error type in `BackendModule`. We should be able to make this change when time permits now that the legacy mongo connector is gone.

djspiewak: If we're adding a new layer to the `Backend` stack, then we really need to shift call-sites of that type to more of an MTL-style.  It's already pretty unwieldy (especially in Mimir, which adds another transformer layer), but this would cross the line, IMO.  :-)

rintcius: @djspiewak : After discussing with @wemrysi , I'm now going to leave the `Backend` stack intact, reverse the change in `ExecutionFailed` and from the Mongo code use `ExecutionFailed` with an empty LP (with a FIXME note).

rintcius: +1 for more MTL-style though :)

PR #2702, opened 2017-08-31
Fixes spark `DeleteField` bug

### LINKED ISSUE #2691: Broken DeleteField semantics on Spark
Deleting a nonexistent field currently returns `NA`, where it should return the original object.  As an example:

```
DeleteField({"a":1, "b":2}, .c)
```

Spark returns `Data.NA` from the above.  The result should be `{"a":1, "b":2}`

alissapajer:
Fixes #2691 

Fixes deleting a nonexistent field in spark. Previously if a field was nonexistent the result would be undefined. Now if a field is nonexistent, the original object is returned.


rabbitonweb:
missing semantic label

alissapajer:
@rabbitonweb this is merging into `next-major` so it doesn't need a version label. You can just use the merge button on github (we don't have the merge script set up for `next-major` either).

PR #2709, opened 2017-09-01
Fix server not stopping when the user presses `Enter`

jedesah:
- Refactors the way `PortChangingServer` operates and how things are setup to simplify and optimize the code around launching the server

### DIFF
--- a/foundation/src/main/scala/quasar/contrib/scalaz/package.scala
+++ b/foundation/src/main/scala/quasar/contrib/scalaz/package.scala
@@ -16,7 +16,10 @@
 
 package quasar.contrib
 
+import slamdata.Predef._
+
 import _root_.scalaz._, \&/._
+import _root_.scalaz.concurrent.Task
 
 package object scalaz {
   def -\&/[A, B](a: A): These[A, B] = This(a)
@@ -27,4 +30,14 @@ package object scalaz {
 
   implicit def toMonadError_Ops[F[_], E, A](fa: F[A])(implicit F: MonadError_[F, E]): MonadError_Ops[F, E, A] =
     new MonadError_Ops[F, E, A](fa)
+
+  implicit class NewThreadTask(t: Task[Unit]) {
+    def runAsyncOnNewThread(threadName: String, daemon: Boolean): Task[Unit] = Task.delay {
+      val thread = new java.lang.Thread(new java.lang.Runnable {
+        override def run(): Unit = t.unsafePerformSync
+      }, threadName)
+      thread.setDaemon(daemon)
+      thread.start()
+    }
+  }
 }
--- a/foundation/src/main/scala/quasar/contrib/scalaz/package.scala
+++ b/foundation/src/main/scala/quasar/contrib/scalaz/package.scala
@@ -16,10 +16,7 @@
 
 package quasar.contrib
 
-import slamdata.Predef._
-
 import _root_.scalaz._, \&/._
-import _root_.scalaz.concurrent.Task
 
 package object scalaz {
   def -\&/[A, B](a: A): These[A, B] = This(a)
@@ -30,14 +27,4 @@ package object scalaz {
 
   implicit def toMonadError_Ops[F[_], E, A](fa: F[A])(implicit F: MonadError_[F, E]): MonadError_Ops[F, E, A] =
     new MonadError_Ops[F, E, A](fa)
-
-  implicit class NewThreadTask(t: Task[Unit]) {
-    def runAsyncOnNewThread(threadName: String, daemon: Boolean): Task[Unit] = Task.delay {
-      val thread = new java.lang.Thread(new java.lang.Runnable {
-        override def run(): Unit = t.unsafePerformSync
-      }, threadName)
-      thread.setDaemon(daemon)
-      thread.start()
-    }
-  }
 }
--- a/it/src/test/scala/quasar/server/ServiceSpec.scala
+++ b/it/src/test/scala/quasar/server/ServiceSpec.scala
@@ -62,11 +62,7 @@ class ServiceSpec extends quasar.Qspec {
       metaRef    <- TaskRef(metastore).liftM[MainErrT]
       quasarFs   <- Quasar.initWithMeta(metaRef, _ => ().point[MainTask])
       shutdown   <- Server.startServer(quasarFs.interp, port, Nil, None, _ => ().point[MainTask]).liftM[MainErrT]
-      // Give the server a chance to startup
-      _          <- Task.delay(java.lang.Thread.sleep(500)).liftM[MainErrT]
-      r          <- f(uri)
-                      .onFinish(κ(shutdown.onFinish(κ(quasarFs.shutdown))))
-                      .liftM[MainErrT]
+      r          <- f(uri).onFinish(κ(shutdown.onFinish(κ(quasarFs.shutdown)))).liftM[MainErrT]
     } yield r).run.unsafePerformSync
   }
 
--- a/it/src/test/scala/quasar/server/ServiceSpec.scala
+++ b/it/src/test/scala/quasar/server/ServiceSpec.scala
@@ -45,7 +45,8 @@ class ServiceSpec extends quasar.Qspec {
 
   val client = org.http4s.client.blaze.defaultClient
 
-  sequential
+  sequential // These tests spin up a server and there is potential for port conflicts if
+             // they don't run sequentially
 
   def withServer[A]
     (port: Int = 8888, metastoreInit: ConnectionIO[Unit] = ().η[ConnectionIO])
@@ -60,7 +61,9 @@ class ServiceSpec extends quasar.Qspec {
       _          <- metastoreInit.transact(transactor).liftM[MainErrT]
       metaRef    <- TaskRef(metastore).liftM[MainErrT]
       quasarFs   <- Quasar.initWithMeta(metaRef, _ => ().point[MainTask])
-      shutdown   <- Server.startServer(quasarFs.interp, port, Nil, None, _ => ().point[Task]).liftM[MainErrT]
+      shutdown   <- Server.startServer(quasarFs.interp, port, Nil, None, _ => ().point[MainTask]).liftM[MainErrT]
+      // Give the server a chance to startup
+      _          <- Task.delay(java.lang.Thread.sleep(500)).liftM[MainErrT]
       r          <- f(uri)
                       .onFinish(κ(shutdown.onFinish(κ(quasarFs.shutdown))))
                       .liftM[MainErrT]
--- a/web/src/main/scala/quasar/server/Http4sUtils.scala
+++ b/web/src/main/scala/quasar/server/Http4sUtils.scala
@@ -41,7 +41,7 @@ object Http4sUtils {
     * indicate that this application was launched from a script, then this Task will
     * immediately complete with a return value of `false`
     * @return `true` if we waited on user input before returning, false if returning
-    *        immediately for lack of any possibility or receiving user input
+    *        immediately for lack of any possibility of receiving user input
     */
   def waitForUserEnter: Task[Boolean] = Task.delay {
     Option(scala.io.StdIn.readLine).isDefined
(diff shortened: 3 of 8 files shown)


jedesah:
Okay, I might have got a little distracted by this change but on the bright side this vastly simplifies the code around this "restarting server" of ours, achieves the desired behavior without an infinite loop waking up periodically and now I am much more familiar with Java's threading model as well as how this work in `scalaz.concurrent._`

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
```diff
@@ -27,4 +30,14 @@ package object scalaz {
 
   implicit def toMonadError_Ops[F[_], E, A](fa: F[A])(implicit F: MonadError_[F, E]): MonadError_Ops[F, E, A] =
     new MonadError_Ops[F, E, A](fa)
+
+  implicit class NewThreadTask(t: Task[Unit]) {
+    def runAsyncOnNewThread(threadName: String, daemon: Boolean): Task[Unit] = Task.delay {
+      val thread = new java.lang.Thread(new java.lang.Runnable {
+        override def run(): Unit = t.unsafePerformSync
```
djspiewak: You're better off running this in the following way:

```scala
t.unsafePerformAsync(_.fold(throw _, _ => ()))
```

There's no reason to hold onto the thread longer than necessary.

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
```diff
@@ -27,4 +30,14 @@ package object scalaz {
 
   implicit def toMonadError_Ops[F[_], E, A](fa: F[A])(implicit F: MonadError_[F, E]): MonadError_Ops[F, E, A] =
     new MonadError_Ops[F, E, A](fa)
+
+  implicit class NewThreadTask(t: Task[Unit]) {
+    def runAsyncOnNewThread(threadName: String, daemon: Boolean): Task[Unit] = Task.delay {
```
djspiewak: This should take an `ExecutorService` rather than manually creating threads and such.

it/src/test/scala/quasar/server/ServiceSpec.scala
```diff
@@ -60,7 +61,9 @@ class ServiceSpec extends quasar.Qspec {
       _          <- metastoreInit.transact(transactor).liftM[MainErrT]
       metaRef    <- TaskRef(metastore).liftM[MainErrT]
       quasarFs   <- Quasar.initWithMeta(metaRef, _ => ().point[MainTask])
-      shutdown   <- Server.startServer(quasarFs.interp, port, Nil, None, _ => ().point[Task]).liftM[MainErrT]
+      shutdown   <- Server.startServer(quasarFs.interp, port, Nil, None, _ => ().point[MainTask]).liftM[MainErrT]
+      // Give the server a chance to startup
+      _          <- Task.delay(java.lang.Thread.sleep(500)).liftM[MainErrT]
```
djspiewak: Uhhhhhhhhhhh…  What?

web/src/main/scala/quasar/server/Http4sUtils.scala
```diff
@@ -34,7 +34,18 @@ object Http4sUtils {
...
 
-  def waitForInput: Task[Unit] = Task.delay(scala.io.StdIn.readLine).void
+  /**
+    * Wait for the user to press enter before returning. This is a blocking operation,
+    * so the thread will be suspended until it is notified of the user pressing Enter following which
+    * the `Task` will complete. If the standard input stream is ended, which would usually
+    * indicate that this application was launched from a script, then this Task will
+    * immediately complete with a return value of `false`
+    * @return `true` if we waited on user input before returning, false if returning
+    *        immediately for lack of any possibility or receiving user input
```
djspiewak: Oh, that's cool.

web/src/main/scala/quasar/server/PortChangingServer.scala
```diff
@@ -33,61 +32,32 @@ import scalaz.stream.{async, Process}
...
+          val conf = ServerBlueprint(port, idleTimeout = Duration.Inf, produceService(startNew))
+          configQ.enqueueOne(conf).map(_.right)
+      }
     }
-    startNew(initialPort) >> (servers(configQ.dequeue, false).unconsOption.map {
-      case None => throw new java.lang.AssertionError("should never happen")
-      case Some((head, rest)) => PortChangingServer(Process.emit(head) ++ rest, configQ.close)
-    })
+    startNew(initialPort) >>
+    servers(configQ.dequeue, false).run.runAsyncOnNewThread("port-changing-thread", daemon = false).as(configQ.close)
```
djspiewak: Is there a reason to do the thread shifting in `Task` rather than directly on `Process`, given that we're actually working with scalaz-stream here?

web/src/test/scala/quasar/server/ControlServiceSpec.scala
```diff
@@ -18,38 +18,30 @@ package quasar.server
...
-        _ <- causeRestart(uri)
-        unconsResult2 <- others.unconsOption
-        (_, others2) = unconsResult2.get
-        b <- afterRestart
-        _ <- server.shutdown
-        _ <- others2.run
-      } yield b).timed(timeoutMillis)(DefaultTimeoutScheduler).onFinish(_ => server.shutdown)
+      shutdown <- PortChangingServer.start(initialPort, reload => control.service(defaultPort, reload) orElse info.service)
+      // Give the server a little time to startup before issuing a request
+      _        <- Task.delay(java.lang.Thread.sleep(500))
```
djspiewak: Ditto on the 😱 here.

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
jedesah: 👍

web/src/main/scala/quasar/server/PortChangingServer.scala
jedesah: No particular reason aside from me being unfamiliar with how to do that, but I'll have a look

web/src/test/scala/quasar/server/ControlServiceSpec.scala
jedesah: Hey! The last version of `master` that's behaving as expected in terms of this stuff has a `sleep` in the "production" code whereas now they have been relegated to test code. Granted this `sleep` is probably a whole lot more brittle than the infinite loop polling for input, but still!

The problem is that the startup is happening in another thread, but perhaps it's possible to only let the `Task` return once the server is running. I feel pretty confident this is achievable for the first one, but the sleeps will still most likely be necessary in these tests between the restart request and the subsequent request. But since it's pretty similar to how one would go about testing this functionality manually. I can't say that I think these sleeps are the end of the world.

Happy to take suggestions!

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -161,9 +146,20 @@ object Server {
...
+               val persistPort = persistPortChange(webCmdLineCfg.configPath)
+               (for {
+                 shutdown <- startServer(quasarInter, port, webCmdLineCfg.staticContent, webCmdLineCfg.redirect, persistPort)
+                 _        <- openBrowser(port).whenM(webCmdLineCfg.openClient)
+                 _        <- stdout("Press Enter to stop.")
+                 // If user pressed enter (after this main thread has been blocked on it),
+                 // then we shutdown, otherwise we just run indefinitely until the JVM is killed
+                 // If we don't call shutdown and this main thread completes, the application will
+                 // continue to run indefinitely as `startServer` uses a non-daemon `ExecutorService`
+                 // TODO: Figure out why it's necessary to `sleep` to keep the main thread from completing
```
jedesah: @djspiewak Any ideas what could be causing this?

jedesah:
@djspiewak Ready for another look

web/src/main/scala/quasar/server/Http4sUtils.scala
```diff
@@ -79,13 +79,32 @@ object Http4sUtils {
...
+        new java.util.concurrent.LinkedBlockingQueue[java.lang.Runnable],
+        org.http4s.util.threads.threadFactory(i => s"${name}-$i", daemon = false))
+      exec.allowCoreThreadTimeOut(timeout)
+      exec
+    }
+    BlazeBuilder
+      .withIdleTimeout(blueprint.idleTimeout)
+      .bindHttp(blueprint.port, "0.0.0.0")
+      .mountService(blueprint.svc)
+      .withServiceExecutor(newPool("http4s-pool", 4, 3.0, false))
```
djspiewak: Does http4s really run with a CPU factor of 3.0?  That's really really strange…

web/src/main/scala/quasar/server/Server.scala
djspiewak: Not sure, but it's certainly possible that something later on downstream is explicitly calling `System.exit`.  That would override the non-daemon status.

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -154,8 +154,11 @@ object Server {
                  // If user pressed enter (after this main thread has been blocked on it),
                  // then we shutdown, otherwise we just run indefinitely until the JVM is killed
                  // If we don't call shutdown and this main thread completes, the application will
-                 // continue to run indefinitely as `startServer` spanned a new non-daemon thread
-                 _        <- waitForUserEnter.ifM(shutdown, Task.now(()))
+                 // continue to run indefinitely as `startServer` uses a non-daemon `ExecutorService`
+                 // TODO: Figure out why it's necessary to `sleep` to keep the main thread from completing
+                 // instead of simply relying on the fact that the server is using a pool of non-daemon threads
+                 // to ensure the application doesn't shutdown
+                 _        <- waitForUserEnter.ifM(shutdown, Task.delay(java.lang.Thread.sleep(Long.MaxValue)))
```
djspiewak: If your intention is to await forever, then a *vastly* superior method would be the following:

```scala
Task.async(_ => ())
```

This solves two problems.  First, a `sleep` is never guaranteed to actually sleep for that amount of time.  Interruptions can happen which short circuit things.  Second, you're completely eating a thread here.  Given that your intention is just to indefinitely block the continuation, you're better off doing it without a synchronous suspension and the resource leak it implies.

web/src/main/scala/quasar/server/Http4sUtils.scala
jedesah: It would seem so

web/src/main/scala/quasar/server/Server.scala
jedesah: You will be eating up the main Thread either way and that's the whole point because, for reasons I don't understand, if the main thread completes, the application shuts down. Now, I'm not sure who/what would "awake" the main Thread but I could be convinced by that argument as well as the annoying `Long.MaxValue` which is unimaginably long but still a finite amount of time.

web/src/main/scala/quasar/server/Server.scala
jedesah: There is nothing downstream. It's not trivially obvious because this is not the `main` function so you never know but it's pretty easy to follow the few calls from `main` and convince oneself that this is pretty much the last statement

jedesah:
@djspiewak The previous implementation of `PortChangingServer` was using an `async.BoundedQueue` which meant that adding a "restart" to the queue through the `Int => Task[Unit]` function would complete immediately, independently from when the thread reading the queue at the other end would actually trigger the restart and the server actually be up and running. With the new implementation that ditches `Process` in favor of `TaskRef`, the `Int => Task[Unit]` function that the endpoint uses to restart the server only completes once the new server has been started. The one drawback to this is where before the server would probably fail to serve a response to the request asking to restart the server about 90% of times, now it deterministically fails to send a response back to the client (because the server will always have been killed already). But I added this to the doc for the endpoint and I think the implementation is much cleaner and easier to test. It would perhaps be possible to delay the closing of the "original" server to give the endpoint the opportunity to return a successful response, but I'm not sure how to accomplish that without using an `unsafe[something]` that's not at the very edge of the world and so I decide to leave it like this for now.

PR #2710, opened 2017-09-01
Add an additional test to metastore API

jedesah:


interface/src/test/scala/quasar/main/api/ApiSpec.scala
```diff
@@ -36,7 +36,7 @@ class ApiSpec extends quasar.Qspec {
       val mount = Mounting.Ops[CoreEff]
       (for {
         firstMetaConf <- MetaStoreFixture.createNewTestMetaStoreConfig
-        quasarFS <- Quasar.initWithDbConfig(firstMetaConf, _ => ().point[MainTask]).run.map(a => a.leftMap(e => new scala.Exception(e.shows))).unattempt
+        quasarFS <- Quasar.initWithDbConfig(firstMetaConf, _ => ().point[MainTask]).leftMap(e => new scala.Exception(e.shows)).run.unattempt
```
nicflores: Does this mean that `task1.run.task2` is the same as `task1.task2.run`, probably nonsense notation, but just wondering about `run`'s factoring properties.

nicflores: I guess you can always delay the `run` until later if you have a stack of `tasks`.

jedesah: Yup. In this case, it's just making use of the function on `EitherT` to make the code ever so slightly more concise

jedesah: We need to get rid of `EitherT` as some point but might as well make use of `leftMap` before doing so

PR #2714, opened 2017-09-01
Avoid synthetic sort key for flattened field appearing in projection

### LINKED ISSUE #2712: Avoid creating synthetic sort key when ordering by projected flattened value
A query such as
```sql
select state, loc[*] from zips order by loc[*]
```
results in the creation of a synthetic sort key for the flattened `loc` values, even though they're already included in the projection. We should recognize this and avoid the superfluous synthetic key.

wemrysi:
This fixes an issue in the `projectSortKeys` analysis where projecting and sorting by the same flattened value would unnecessarily result in a synthetic sort key. 

Fixes #2712.


alissapajer:
@wemrysi after I've merged this, it occurs to me that we should add a sql compiler test for this particular case to ensure it doesn't regress. Could you add a test and open a new PR for that?

wemrysi:
@alissapajer Will do!

PR #2723, opened 2017-09-05
Add a test to ensure #2712 doesn't regress

wemrysi:
SQL^2 compiler test ensuring ordering by a projected flattened field doesn't use a synthetic sort key.

Should have been included in #2714, but I forgot. Thanks @alissapajer for pointing this out!


wemrysi:
@alissapajer I also verified that this test fails on the commit prior to the merge of #2714.

PR #2731, opened 2017-09-06
Metastore endpoint now obscures postgres password on GET request

jedesah:
Fixes an issue that was reported in Slamdata Backend

### DIFF
--- a/core/src/main/scala/quasar/db/DbConnectionConfig.scala
+++ b/core/src/main/scala/quasar/db/DbConnectionConfig.scala
@@ -46,16 +46,22 @@ object DbConnectionConfig {
   }
 
   implicit val encodeJson: EncodeJson[DbConnectionConfig] =
+    encodeJsonImpl(false)
+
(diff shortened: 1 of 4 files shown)


core/src/main/scala/quasar/db/DbConnectionConfig.scala
```diff
@@ -46,16 +46,22 @@ object DbConnectionConfig {
   }
 
   implicit val encodeJson: EncodeJson[DbConnectionConfig] =
+    encodeJsonImpl(false)
+
+  val secureEncodeJson: EncodeJson[DbConnectionConfig] =
```
nicflores: should this be _`implicit`_ `val secureEncodeJson: ...`?

jedesah: There should only be one implicit val or else we will run into ambiguous implicits. I think it's reasonable to have to be explicit in order to get this "special" encoding

PR #2738, opened 2017-09-07
Implemented string || string in mimir

### LINKED ISSUE #2692: Mimir should handle `ConcatArrays` of two strings or a string and an array
SQL^2 uses `||` for both array and string concat (and heterogeneous concat of a string and an array). For example,`"foo" || "bar"` evaluates to `"foobar"` and `"foo" || [1,2,3]` evaluates to `["f","o","o",1,2,3]`. Mimir does not currently handle the string case or the heterogeneous case.

It's a separate discussion whether these are the semantics for `||` that we'd like to uphold for SQL^2.

djspiewak:
Fixes #2692 (sort of).  Doesn't address the `str || array` or `array || str` case, and we're not likely to get to those for a while.  Also I still think those application of the operator are quite bad (my canonical example: `"🤷‍♂️"`).  But that's a discussion for another day.


mimir/src/main/scala/quasar/mimir/MapFuncCorePlanner.scala
```diff
@@ -187,8 +192,13 @@ final class MapFuncCorePlanner[T[_[_]]: RecursiveT, F[_]: Applicative]
         (WrapObject[A](value, key): TransSpec[A]).point[F]
       case MapFuncsCore.MakeMap(key, value) =>
         (WrapObjectDynamic[A](key, value): TransSpec[A]).point[F]
+
       case MapFuncsCore.ConcatArrays(a1, a2) =>
-        (OuterArrayConcat[A](a1, a2): TransSpec[A]).point[F]
+        (Cond(
+          Infix.And.spec[A](IsType(a1, StrAndDateT), IsType(a2, StrAndDateT)),
```
jsantos17: `StrAndDateT` for concat? What is the expected result of concatenating two dates?

djspiewak: Why, the `toString` of both dates glued together, of course!  😜

But seriously, dates don't exist in json.  They're actually strings, and we just detect them and internally manipulate them as dates for efficiency reasons.  So anywhere that you have a date, you *actually* have a string (and potentially vice versa).  It's kind of terrible, but the auto-coercion allows us to remain a json database while still providing some of this rich functionality without killing performance.

jsantos17: Makes sense in this context

PR #2739, opened 2017-09-07
Use MapFunc extractor for FreeMap.

alissapajer:


connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -262,7 +262,7 @@ object MapFuncCore {
       case func =>
         val writer =
           func.traverse[Writer[List[(FreeMapA[T, A], Type)], ?], FreeMapA[T, A]] {
-            case Embed(CoEnv(\/-(MFC(Guard(e, t, s, Embed(CoEnv(\/-(MFC(Undefined()))))))))) =>
+            case Embed(CoEnv(\/-(MFC(Guard(e, t, s, ExtractFunc(Undefined())))))) =>
```
alissapajer: Due to type erasure we cannot write `case ExtractFunc(Guard(e, t, s, ExtractFunc(Undefined()))) =>` and I don't know a way around this. Any ideas?

djspiewak: I think it's actually a buggy interaction with the existential types.  The type of `t`, `s` and `ExtractFunc(Undefined())` in your example will be `a forSome { type a }`, and I think patmat is just bailing on that.  There's probably no way to get around it.

PR #2744, opened 2017-09-10
Fix resource leak in `MetaStore`

jedesah:


core/src/main/scala/quasar/metastore/MetaStore.scala
```diff
@@ -53,13 +53,22 @@ object MetaStore {
   def connect(dbConfig: DbConnectionConfig, initializeOrUpdate: Boolean, schemas: List[Schema[Int]]): EitherT[Task, MetastoreFailure, MetaStore] = {
     for {
       tx <- poolingTransactor(DbConnectionConfig.connectionInfo(dbConfig), DefaultConfig).leftMap(f => f:MetastoreFailure)
-      _  <- initializeOrUpdate.whenM(schemas.traverse(this.initializeOrUpdate(_, tx.transactor, None)))
-      _  <- schemas.traverse(verifySchema(_, tx.transactor))
-      _  <- stdout(s"Using metastore: ${DbConnectionConfig.connectionInfo(dbConfig).url}").liftM[EitherT[?[_], MetastoreFailure, ?]]
+      _  <- onFailOrLeft(initializeOrUpdate.whenM(schemas.traverse(this.initializeOrUpdate(_, tx.transactor, None))) >>
+            schemas.traverse(verifySchema(_, tx.transactor)) >>
+            stdout(s"Using metastore: ${DbConnectionConfig.connectionInfo(dbConfig).url}").liftM[EitherT[?[_], MetastoreFailure, ?]])(tx.shutdown)
```
djspiewak: Just a style thing, but why bother using `>>` here when you're already inside `for`-comprehension.

jedesah: It's so I can call `onFailOrLeft` only once

djspiewak: Oh, that makes sense.

Issue #2541, opened 2017-07-25
date_part should return undefined when applied to a non-temporal argument in MongoDB


wemrysi:
The `invalidDateFilter.test` query is failing on QScript-based MongoDB connectors due to `date_part` being applied to a numeric argument, instead it should return `undefined`.

alissapajer:
@drostron are you in progress on this? If so, :+1:. If not, would it be okay if I unassign you so someone else (or yourself in the future) can grab it?

drostron:
@alissapajer I started looking at it the other day and am back to it today now that the initial VCache work is submitted.

jsantos17:
This a manifestation of a wider problem. There are other queries that exhibit similar behavior, yet don't error on execution. Any query with a filter that requires typechecks in the predicate is also affected by this issue.

jsantos17:
After some research I have found that it's not only typechecks that have this problem. Pretty much every `Guard` in a `Filter`'s `FreeMap` have similar problems. The implementation pulls every `Guard`  from `Filter` and puts them after `Project`. Consider the following query and its plan for Mongo 3.4:

```sql
select city from zips where length(city) < 4
```

On master, this plans as:

QScript:

```
QScript (Mongo-specific):
Map
├─ Filter
│  ├─ ShiftedRead(/jps/zips, ExcludeId)
│  ╰─ Guard
│     ├─ ProjectField
│     │  ├─ ○
│     │  ╰─ Constant(Str(city))
│     ├─ Type(Str)
│     ├─ Lt
│     │  ├─ Length
│     │  │  ╰─ ProjectField
│     │  │     ├─ ○
│     │  │     ╰─ Constant(Str(city))
│     │  ╰─ Constant(Int(4))
│     ╰─ Undefined
╰─ ProjectField
   ├─ ○
   ╰─ Constant(Str(city))
```

Workflow:
```
Workflow (raw):
Chain
├─ $ReadF(jps; zips)
├─ $ProjectF
│  ├─ Name("0" -> "$city")
│  ├─ Name("1" -> { "$strLenCP": "$city" })
│  ├─ Name("src" -> "$$ROOT")
│  ╰─ IgnoreId
├─ $MatchF
│  ╰─ And
│     ├─ Doc
│     │  ╰─ Expr($0 -> Type(Text))
│     ╰─ Doc
│        ╰─ Expr($1 -> Lt(Int32(4)))
╰─ $ProjectF
[...]

PR #2753, opened 2017-09-13
Fix some length-related tests

### LINKED ISSUE #2733: Broken length/array_length in new Mongo
Several of the tests mentioned in #2667 are broken because of a broken `length`/`array_length`. This is broken in the current master build.

Example query:

```sql
select array_length(loc, 1), city from /mngo/jps/zips
```

returns an error due to incorrect usage of `strLenCP`

```
Plan execution failed: MongoDB Error: Command failed with error 34471: '$strLenCP requires a string argument, found: array' on server localhost:27017. The full response is { "ok" : 0.0, "errmsg" : "$strLenCP requires a string argument, found: array", "code" : 34471, "codeName" : "Location34471" }
```

Will create issues stemming from #2667 as we discover them. This one seems fairly serious so priority is high.

jsantos17:
Fixes #2733


alissapajer:
@jsantos17 is this ready for review? If so, can you assign it to someone.

jsantos17:
@alissapajer not ready for review just yet! I still have a few more changes I'll be pushing soon

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -710,51 +732,47 @@ class PlannerSpec extends
...
-        $match(
-          Selector.Doc(
-            BsonField.Name("__tmp2") -> Selector.Eq(Bson.Bool(true)))),
+        $match(Selector.Doc(BsonField.Name("0") -> Selector.Eq(Bson.Bool(true)))),
         $project(
-          reshape("value" -> $field("__tmp3")),
+          reshape("value" -> $field("src")),
           ExcludeId)))
-    }.pendingUntilFixed(notOnPar)
+    }.pendingUntilFixed("#2541")
```
jsantos17: When #2541 is fixed, this test will also be fixed with the fixed expectations above.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -2597,25 +2617,42 @@ class PlannerSpec extends
...
     "plan select length() and simple field" in {
       plan(sqlE"select city, length(city) from zips") must
       beWorkflow(chain[Workflow](
         $read(collection("db", "zips")),
+        $project(
+          reshape(
+            "city" -> $field("city"),
+            "1" ->
+              $cond($and(
+                $lte($literal(Bson.Text("")), $field("city")),
```
jsantos17: In here, we effectively check twice if `city` is a string before attempting to obtain its length. Not sure how to improve this. I guess we could elide guards before `Length` applications and rely solely on the `Length` impl to provide the appropriate guards.

wemrysi: @jsantos17 We could add simplifications to flatten nested `$conds` and eliminate duplicate predicates.

jsantos17:
@wemrysi So I tried to do a simplification to remove the double typechecks, but ran into trouble pretty quickly when I attempted to write `simplify` for `ExprOp3_4F`. I changed the instances type from `ExprOpOps.Aux[ExprOp3_4F, F]` to `ExprOpOps.Uni[F]` (with `ExprOpCore` and `ExprOp3_4F` injectable in F). The latter type is what I need for the rewrite, but changing the type annotation requires the `bson` method to handle non 3.4 operations too. I could simply delegate that to `ExprOpCore` by calling it directly, but afaik the intent of the `ExprOpOps` typeclass is to handle operations in different versions of Mongo in separate instances. Then the coproduct instance ties them all together. You can look at the code for my attempt [here](https://github.com/jsantos17/quasar/compare/2733-broken-length...jsantos17:2733-broken-length-simplify). Put simply: my rewrite requires operations from core and from 3.4. However, the structure of the `ExprOpOps` typeclass makes that very hard.

My next attempt is exploiting your observation  that the distinct typechecks for `array_length` and `length` call sites survive the LP to QScript translation. So I would generate the correct expression in the first place, rather than fixing it later.

Any other ideas are very welcome too. I also had to refactor the `Check` class a bit to make it usable outside `MongoDbPlanner`.

mongodb/src/main/scala/quasar/physical/mongodb/expression/Check.scala
```diff
@@ -31,14 +31,13 @@ import scalaz._, Scalaz._
   * seen in the order of declarations in this class.
   */
 final class Check[T, EX[_]]
+  (embed: EX[T] => T)
   (implicit
-    TR: Recursive.Aux[T, EX],
-    TC: Corecursive.Aux[T, EX],
```
jsantos17: The `Recursive` instance is not necessary for this class so I removed it.

wemrysi:
@jsantos17 The nested conditional normalization isn't specific to `ExprOp3_4F`, so you could write it for any `F[_]` such that `ExprOpCore :<: F` as a standalone transformation. That said, I agree that if we can use the typecheck information in QScript to just plan the proper operation in the first place, that'd be better.

As for `Check`, in this case I think it'd be better to just define `[F[_], A] => Corecursive.[Free[F, A], F]` and use that instance with the unmodified `Check` (we should add that as a low-priority instance in matryoshka).

As an aside, we also use `PrismNT` for this sort of thing elsewhere, like
```scala
def someTransform[T, F[_], G[_]](GtoF: PrismNT[G, F])(implicit T: Birecursive.Aux[T, G], C: ExprOpCoreF :<: F, E34: ExprOp3_4F :<: F): F[T] => G[T] // or whatever return type you need
```
and then you'd pass in the `id` prismNT for shapes like `T[F]` and the `coEnv` prismNT for `Free[F, A]`, for example.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -237,6 +240,10 @@ object MongoDbPlanner {
       case ProjectIndex(a1, a2)  => unimplemented[M, Fix[ExprOp]]("ProjectIndex expression")
       case DeleteField(a1, a2)  => unimplemented[M, Fix[ExprOp]]("DeleteField expression")
 
+      case Guard(expr, Type.Str, cont @ $strLenCP(_), fallback) =>
```
jsantos17: I managed to eliminate the duplicate typechecks with this. It's more of a workaround, but I don't know how we can do better unless we introduce `array_length`. 

Also, this workaround doesn't work with `Subset`. The guards are more complicated in that case. Behavior in that case is unchanged from what we have in master. But it should easy enough to fix by adding `Subset` to `assumeReadType` in a subsequent PR.

wemrysi: Mind adding a comment here explaining why these exist?

jsantos17:
@wemrysi ready for another look. Added a test and included the comments.

PR #2754, opened 2017-09-13
Move spark-cassandra to backend module

rabbitonweb:
Fixes https://github.com/quasar-analytics/quasar/issues/2716


alissapajer:
@drostron it would probably be best if we can land this today so that @djspiewak can then merge it into the dynamic loading PR #2756. Otherwise we'd have to deal with the conflicts going the other way.

djspiewak:
@alissapajer Conflicts in either direction going to be minor, though they are semantically more validly conflicts generated by #2756.  I agree that this should land before my PR, just in the interest of good git hygiene.

drostron:
@alissapajer @djspiewak Sure, I was most of the way through yesterday and will switch to finishing the review momentarily.

PR #2735, opened 2017-09-07
upgrade mongo driver

rintcius:
Same as #2607 - just against `master` instead of `next-major`

* upgrade mongo driver to version 3.5.0
* add support for decimal128
* handle `BSONException`

~~Note: 
Mongo driver requires Netty 4.1 which conflicts with Spark's Netty (4.0).
For now just excluding Spark's Netty, and thus breaking Spark.
This should be resolved once both Spark and new Mongo are on BackendModule (#2216 resp #2094) and we have dynamic backend loading in place (#2105).~~

Fixes #2167

PR #2756, opened 2017-09-13
Dynamic backend loading

### LINKED ISSUE #2105: Implement dynamic loading of backends

djspiewak:
Fixes #2105 
Unblocks #2735 

This is probably not ready for merge in its precise current form, but this should get the ball rolling.  I expect build failures.  Also spark is not quite working (need help from @rabbitonweb on that one).  But the core of it is there, and it's at least ready for review.  At the very least, worst comes to worst, someone could take this and finish it if I dropped off the face of the earth.

A couple things that still need to be done:

- [x] Testing against marklogic and couchbase, as well as non-3.4 mongo versions (I'm using Travis as a way of short-circuiting that work)
- [ ] Updating the installer infrastructure to fetch the connector jars and build them into a directory (there's no issue for it, but @nicflores and I have discussed at length)
- [ ] Make Spark not explode
  + [ ] Investigate removing one of the two major Spark hacks to get around their broken classloader usage
  + [x] Maybe propagate those Spark hacks further 
  + [ ] Weep, weep, for the world is lost

I'm provisionally assigning this to @alissapajer, mostly because no one else expressed extreme enthusiasm about reviewing it.  @wemrysi paired with me extensively here, and he deserves by far the bulk of the credit for making this work.  I'm adding @rabbitonweb as a reviewer, both because I want his thoughts, and because I need his help on the spark stuff.  Also adding @jedesah as a reviewer, since he touches the surface-level interface stuff more than most of us.

---

After this PR lands, if you're doing work on the integration tests (by which I mean, running them), you *must* either use `it/test` at least once (not `testOnly`), or run the `it/sideEffectTestFSConfig` task.  Once you've run either task at least once in an SBT session, everything else should behave normally.  This is a hack to get around the fact that I really didn't want to monkey-patch my way into the `testOnly` and `testQuick` tasks.


djspiewak:
Looks like it's just Spark failing now.

build.sbt
```diff
@@ -392,10 +448,6 @@ lazy val interface = project
   .settings(name := "quasar-interface-internal")
   .dependsOn(
     core % BothScopes,
-    couchbase,
-    marklogic % BothScopes,
-    mongodb,
-    sparkcore,
```
alissapajer: :tada:

interface/src/main/scala/quasar/main/Quasar.scala
```diff
@@ -93,34 +98,38 @@ object Quasar {
...
     } yield quasarFS
 
-  def initWithMeta(metaRef: TaskRef[MetaStore], persist: DbConnectionConfig => MainTask[Unit]): MainTask[Quasar] =
+  def initWithMeta(loadConfig: FsLoadCfg, metaRef: TaskRef[MetaStore], persist: DbConnectionConfig => MainTask[Unit]): MainTask[Quasar] =
     (for {
       metastore  <- metaRef.read.liftM[MainErrT]
       hfsRef     <- TaskRef(Empty.backendEffect[HierarchicalFsEffM]).liftM[MainErrT]
       mntdRef    <- TaskRef(Mounts.empty[DefinitionResult[PhysFsEffM]]).liftM[MainErrT]
 
+      // TODO
```
alissapajer: Can you elaborate on this `TODO`?

alissapajer:
@djspiewak have we tested that jars built from this branch (i.e. `sbt web/assembly` and `sbt repl/assembly`) work?

interface/src/main/scala/quasar/main/Quasar.scala
djspiewak: It's stale. I'll remove it.

djspiewak:
@alissapajer `repl/assembly` most definitely will not work.  Wouldn't be that hard to make it work, but I wasn't under the impression that was a necessary use-case, and it was extra effort.  `web/assembly` should definitely work, and testing it is on my docket for today.  The loading follows the same code path as the local case, save for the fact that it reads the module information from the manifest file.  Testing that is definitely worth doing, and longer-term, I'd like to get an automated test for it, but we don't have the infrastructure for that right now.

interface/src/main/scala/quasar/main/package.scala
```diff
@@ -40,37 +41,200 @@ import scala.util.control.NonFatal
...
+              _ <- if (result.isEmpty)
+                Task.delay(log.warn(s"unable to load any backends from $jar; perhaps the 'Backend-Module' attribute is not defined"))
+              else
+                Task.now(())
+            } yield result
+          }
+        } yield loaded.flatMap(r => IList.fromList(r.toList)).fold
+
+      case ExplodedDirs(backends) =>
+        val maybeDefinitionsM: Task[IList[Option[BackendDef[Task]]]] = backends traverse {
```
alissapajer: If we "fail nicely" when loading a backend (i.e. the `Task` we're traversing hasn't errored), then we'll get a result of `None` for that particular backend. But if a `Task` were to error, then this result would be a single erroring `Task` with no backends loaded at all, even if only one backend errored. This isn't ideal, unless the exception really were fatal.

alissapajer:
@djspiewak we release a repl assembly jar as a part of continuous delivery, so if we don't get it working for this PR we will no longer be delivering it. It is used in QA testing by @dugdun as well as possibly by the frontend for testing (to rule out SlamData-caused bugs). I don't how important the repl jar is for customers.

interface/src/main/scala/quasar/main/package.scala
djspiewak: The meaningful exceptions are caught, logged and go to `None`.  There may be cases I missed, but I think I got them all.  You're correct that an uncaught error would short-circuit the whole process and actually crash quasar startup.

djspiewak:
@alissapajer Thinking more deeply about it…  It would be possible to keep `repl/assembly` working by using the `--backend` option manually.  Basically, any invocation of it would need to include something like `--backend:quasar.physical.mongodb.MongoDb\$=<path-to-mongodb-jar>`.  That's kind of a pain, but it's workable.  Making it so the JARs themselves are parsed in the way that `web/assembly` does is *possible*, but not trivial enough that I could do it today without delaying other things.

alissapajer:
@djspiewak at least for the use cases that I know of, I think using `--backend` manually is an acceptable workaround, as long as we document it (can you update [the README](https://github.com/quasar-analytics/quasar#repl-jar) in this PR).

djspiewak:
Yeah, README updates for both web and repl will be needed here.

alissapajer:
@djspiewak what else is left to do before we can merge this?
- [ ] Update the README (repl, web, lack of spark if applicable)
- [ ] Update the installer infrastructure to fetch the connector jars - we could presumably do this after the merge, but then no one could build an installer and we want to know that there are no surprises here

@jedesah @rabbitonweb we are still pending review from you as well.

alissapajer:
We can hold off a couple days to see if we can revive spark as well.

djspiewak:
@alissapajer Updating the README is definitely on my immediate TODO list.  Updating the infrastructure is on @nicflores's plate, but it's much easier to do that once the PR has been merged (since then the publications will actually be happening).  I think it's ok to just have broken installer builds for a little bit.

I'm not sure whether or not we should hold this based on Spark.  I mean, nominally, the whole idea is to enable the new Mongo driver *without* disabling Spark, but this is already a good step in that direction.  Spark is still there.  It's still building.  You can still work on it locally.  It's just that the integration tests don't run.  Turning all of that back on again is much simpler (`git revert c57fcdc003`) than it would be with the new mongo driver and *without* dynamic loading.

I'm going to keep chasing things down as much as I can, but realistically I don't know much about how Spark wires this stuff together.  It actually surprises me a lot that Netty is at all involved in spark_local.  The best thing to do might be to land the PR (after the README changes) so that the new mongo driver can land and the infrastructure can be updated, and then address re-enabling Spark in a new PR.

alissapajer:
@djspiewak @nicflores is there an issue for updating the installer infrastructure?

djspiewak:
Not yet.  I'll create one so we can have a more formal continuation, if you're :+1: on the remainder of the plan I outlined.

alissapajer:
@djspiewak my one concern is that we may uncover something large and unexpected regarding the installer infrastructure. I don't know everything that's involved in that, so could you or @nicflores provide an outline of what work that will require, and maybe highlight any possible tricky bits that you foresee? If you want to just write that up in the issue, that would be fine. Once I understand what the installer changes entail, I'll feel better (or at least more knowledgeable :) ) about it.

djspiewak:
@alissapajer See my ninja-edit.

alissapajer:
@djspiewak that _was_ a ninja edit! Ok, the installer infrastructure issue looks good. I am :+1: to merging this before we land slamdata/installers#25. So your above plan works for me. It's easy to enable Spark after this and the mongo driver update land.

djspiewak:
@alissapajer KK, I'll update the README then in my next push (no guarantees that will happen today).  Assuming the build is green then, it seems like we'll be 👍 for merge unless @jedesah or @rabbitonweb have overriding objections.

alissapajer:
@djspiewak perfect. I still want one more day to let my mind churn over this anyway, just in case I think of anything else to discuss.

web/src/main/scala/quasar/server/CliOptions.scala
```diff
@@ -20,24 +20,32 @@ import slamdata.Predef._
...
 @Lenses
 final case class CliOptions(
-  cmd: Cmd,
-  config: Option[String],
-  contentLoc: Option[String],
-  contentPath: Option[String],
-  contentPathRelative: Boolean,
-  openClient: Boolean,
-  port: Option[Int])
+    cmd: Cmd,
```
jedesah: Why the 4 spaces?

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -39,28 +40,49 @@ import scalaz.concurrent.Task
...
-    openClient: Boolean) {
-    def toCmdLineConfig: CmdLineConfig = CmdLineConfig(configPath, cmd)
+      cmd: Cmd,
+      staticContent: List[StaticContent],
+      redirect: Option[String],
+      port: Option[Int],
+      configPath: Option[FsFile],
+      loadConfig: FsLoadCfg,
+      openClient: Boolean) {
+
```
jedesah: I guess it's just habit, but I don't see the benefit of this extra line

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -39,28 +40,49 @@ import scalaz.concurrent.Task
...
   object WebCmdLineConfig {
     def fromArgs(args: Vector[String]): MainTask[WebCmdLineConfig] =
       CliOptions.parser.safeParse(args, CliOptions.default)
         .flatMap(fromCliOptions(_))
 
-    def fromCliOptions(opts: CliOptions): MainTask[WebCmdLineConfig] =
+    def fromCliOptions(opts: CliOptions): MainTask[WebCmdLineConfig] = {
+      import scala.sys
+
+      val loadConfig = opts.loadConfig.fold(
```
jedesah: This is very similar to the code in the `Main` of the REPL. Any way we can reduce the duplication here by moving this code to `interface`?

foundation/src/main/scala/quasar/contrib/pathy/package.scala
```diff
@@ -49,6 +50,25 @@ package object pathy {
   def pathName(p: APath): Option[PathSegment] =
     refineType(p).fold(x => dirName(x) map liftDirName, x => some(fileName(x)))
 
+  object ADir {
+
+    // performs some effects; I'm probably going to regret this...
+    def fromFile(file: JFile): Option[ADir] = {
```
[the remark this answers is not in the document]
djspiewak: I don't really mind fixing it. It's just a pain because the usage sites don't necessarily have a good way of forcing a `Task` suspension.  I'll see what I can do.

interface/src/main/scala/quasar/main/package.scala
```diff
@@ -40,37 +41,189 @@ import scala.util.control.NonFatal
...
-  val physicalFileSystems: BackendDef[PhysFsEffM] = IList(
-    Couchbase.definition translate injectFT[Task, PhysFsEff],
-    marklogic.MarkLogic(
-      readChunkSize  = 10000L,
-      writeChunkSize = 10000L
-    ).definition translate injectFT[Task, PhysFsEff],
+  final case class ClassName(value: String) extends AnyVal
+  final case class ClassPath(value: IList[APath]) extends AnyVal
+
+  sealed trait FsLoadCfg extends Product with Serializable
```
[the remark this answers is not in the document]
djspiewak: `BackendConfig` is fine.  It does annoy me that our internal terminology (in code) is pretty inconsistent on this front.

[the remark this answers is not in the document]
djspiewak: It lives here mostly so that it can be imported.  I can probably back that off now that it's only used in like three places, but I don't share your objection.  The `quasar.main` package object already includes loads of other big chunks of tightly coupled code; this is no different.

interface/src/main/scala/quasar/main/package.scala
```diff
@@ -40,37 +41,189 @@ import scala.util.control.NonFatal
...
+      NullPointerException
+    }
+    import java.net.URLClassLoader
+
+    import FsLoadCfg._
+
+    // this is a side-effect in the same way that `new` is a side-effect
+    val ParentCL = this.getClass.getClassLoader
+
+    def loadBackend(cn: String, cl: ClassLoader): OptionT[Task, BackendDef[Task]] = {
```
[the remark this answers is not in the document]
djspiewak: `loadBackend` is an internal function only called from two places.  Both of those places get the `cn` value by unpacking a newtype.  I'm not sure there's really a problem with using the bare value here.

interface/src/main/scala/quasar/main/Quasar.scala
```diff
@@ -93,34 +98,37 @@ object Quasar {
     *
     * Not used in the codebase, but useful for manual testing at the console.
     */
-  val init: MainTask[Quasar] = initFromMetaConfig(None, _ => ().point[MainTask])
+  def init(loadConfig: FsLoadCfg): MainTask[Quasar] =
```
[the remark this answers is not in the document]
djspiewak: There is literally no sensible default other than `FsLoadCfg.Empty`, which is very probably not what you want under any circumstance.  There certainly is no way to usefully specify "any `BackendDef` on the classpath" simply because none of the backends (other than mimir) will ever be on the classpath!

it/src/test/scala/quasar/TestConfig.scala
```diff
@@ -139,7 +143,6 @@ object TestConfig {
 
   def confValue(name: String): OptionT[Task, String] = {
     val config = knobs.loadImmutable(
-      Optional(SysPropsResource(Prefix("")))                    ::
```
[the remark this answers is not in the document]
djspiewak: We weren't using it, and Knobs is unable to parse the system property I added to pass in classpath configuration to the backend, which made it a non-starter even if we *were* using it.

it/src/test/scala/quasar/TestConfig.scala
```diff
@@ -180,6 +183,28 @@ object TestConfig {
         fail[ADir](s"Test data dir must be an absolute dir, got: $s").liftM[OptionT])
     } getOrElse DefaultTestPrefix
 
+  val testFsLoadCfg: Task[FsLoadCfg] = {
+    val confStrM =
+      Task.delay(java.lang.System.getProperty("slamdata.internal.fs-load-cfg", ""))
```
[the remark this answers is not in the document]
djspiewak: We have to get the classpath information for the individual backends into the test configuration so that they can be loaded.  This is a temporary state of affairs, and will go away completely once we invert the dependency between `it` and the connectors, but for right now it's a necessity.  Unlike with `Quasar` itself, we can't just use a `main` argument since it's a test.  System properties are the only way of getting this data into the test suite.

it/src/test/scala/quasar/server/ServiceSpec.scala
```diff
@@ -176,7 +184,7 @@ class ServiceSpec extends quasar.Qspec {
       }
 
       r ==== finalCfg.asJson.right
-    }
+    }.flakyTest   // FIXME this test is actually non-deterministic depending on server scheduling
```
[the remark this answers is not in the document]
djspiewak: We don't appear to be correctly propagating backpressure from the database response through to the http service response.  This is a minor issue from a load standpoint, since I don't expect us to get DDoSed on our `/mount` endpoint, but it does mean that tests like this are effectively asequential.

repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -141,12 +142,25 @@ object Main {
   def safeMain(args: Vector[String]): Task[Unit] =
     logErrors(for {
       opts    <- CliOptions.parser.safeParse(args, CliOptions.default)
+
```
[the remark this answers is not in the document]
djspiewak: Whitespace helps readability.  We don't use enough of it.

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/cassandra/SparkCassandra.scala
```diff
@@ -197,6 +197,7 @@ object SparkCassandra extends SparkCore with ManagedWriteFile[AFile] with Chroot
   }
 
   private def initSC: Config => DefErrT[Task, SparkContext] = (config: Config) => EitherT(Task.delay {
+    java.lang.Thread.currentThread().setContextClassLoader(getClass.getClassLoader)
```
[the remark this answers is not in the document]
djspiewak: I'd rather leave a single explanation in a common location, to avoid needing to be redundant in each of the places where this hack has to be implemented.  Suggestions on a locus?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/SparkElastic.scala
```diff
@@ -208,7 +208,11 @@ object SparkElastic extends SparkCore with ManagedWriteFile[AFile] with Chrooted
     OptionT(jar).toRight(NonEmptyList("Could not fetch sparkcore.jar").left[EnvironmentError])
   }
 
+  @SuppressWarnings(Array("org.wartremover.warts.Null"))
   private def initSC: Config => DefErrT[Task, SparkContext] = (config: Config) => EitherT(Task.delay {
+    // look, I didn't make Spark the way it is...
```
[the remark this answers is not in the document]
djspiewak: The `null` suppression is vestigial and can be removed.

web/src/main/scala/quasar/server/CliOptions.scala
```diff
@@ -20,24 +20,32 @@ import slamdata.Predef._
 import quasar.build.BuildInfo
 import quasar.cli.Cmd, Cmd._
 
+import java.io.File
```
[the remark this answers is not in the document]
djspiewak: Scalafmt won't be able to reformat imports particularly well.  All it can do is alphabetize `{`...`}` blocks.

web/src/main/scala/quasar/server/CliOptions.scala
djspiewak: Readability.  The two space convention is broken and shouldn't be used.

web/src/main/scala/quasar/server/Server.scala
djspiewak: Again, readability.  We don't use whitespace enough in quasar, and it causes a lot of problems.  Not to stereotype, but I've observed this *pervasively* from IDE users, not just on Quasar but on other Scala codebases.

web/src/main/scala/quasar/server/Server.scala
djspiewak: We could, but I don't like to do so because of how unsafe it is.  Scopt code is neatly confined to *here* and only here.  Also given how scopt is using mutability to achieve its goals, I'm not confident in my ability to safely factor it out.

interface/src/main/scala/quasar/main/package.scala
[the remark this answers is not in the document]
jedesah: I'm not saying it should have a different type. Just that it be called `className` instead of `cn`

interface/src/main/scala/quasar/main/Quasar.scala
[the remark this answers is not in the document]
jedesah: Fair point. Shows that I haven't fully wrapped my head yet around how the dynamic loading is happening

it/src/test/scala/quasar/TestConfig.scala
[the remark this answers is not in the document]
jedesah: Is this a bug in `knobs`?

it/src/test/scala/quasar/server/ServiceSpec.scala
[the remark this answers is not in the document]
jedesah: What makes you say that we aren't correctly propagating the backpressure from the database?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/elastic/SparkElastic.scala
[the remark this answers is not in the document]
jedesah: My comment was assuming the `null` suppression was not vestigial. If there are no `null`s then great! and let's remove the `@SuppressWarnings` 😄

web/src/main/scala/quasar/server/Server.scala
jedesah: What problems?

sparkcore/src/main/scala/quasar/physical/sparkcore/fs/cassandra/SparkCassandra.scala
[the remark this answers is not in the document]
jedesah: That's fair. Any way we can avoid the code duplication at the same time? Considering the code is duplicated, I'd be okay with a duplicated explanation as well, but I'm okay with some reasonable central place. Don't have any particularly amazing suggestion though

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -39,28 +40,49 @@ import scalaz.concurrent.Task
...
+        plugins => FsLoadCfg.JarDirectory(ADir.fromFile(plugins).getOrElse(sys.error("plugin directory does not exist (or is a file)"))),
```
jedesah: Can we get this error message into the failure case of `MainTask`?

web/src/main/scala/quasar/server/Server.scala
jedesah: `CliOptions` is a case class defined in Quasar that appears pure to me from a quick glance. I understand `scopt` uses mutability internally which is unfortunate, but I'm not aware that it exposes it other than by exposing effectful top-level functions that are not wrapped in any effect type. The duplicated code here seems entirely pure to me (upon a quick glance). In fact looking at the diff in Github, I'm not aware that `scopt` is being used in any of this new code you've added. It was probably involved in creating `CliOptions` but like I said that's a normal case class defined in Quasar.

it/src/test/scala/quasar/TestConfig.scala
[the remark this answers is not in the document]
djspiewak: No. I just used characters in the property *value* which are HOCON-reserved symbols, and so knobs' parser bailed.

it/src/test/scala/quasar/server/ServiceSpec.scala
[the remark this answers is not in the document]
djspiewak: Primarily because the test fails intermittently with a result which indicates that the ordering of the two actions is non-deterministic.  The only sensible explanation there is that either the server of the client is resulting in the database commit being run in parallel, without the continuation being dependent on the completion of the prior action.  I haven't dug into it any further, but it's definitely something we should fix if true.

web/src/main/scala/quasar/server/Server.scala
djspiewak: Reading, navigating and transforming large blocks of entirely contiguous code, unrelated code that uses multiple constructs (e.g. interleaved `def`s, `class`es, etc) is extremely challenging.  It becomes easier if you have a structure-based editor (i.e. and IDE), which I suspect is part of why IDE users seem to have a greater tendency to create such undelimited blocks.

Visual breaks are very helpful in general for indicating related, and by extension, *unrelated* expressions.  I see no reason to avoid them other than a desire for concision, which is not a goal unto itself in my opinion.

web/src/main/scala/quasar/server/Server.scala
djspiewak: You are correct.  I wish Github showed more context around review comments; I thought this comment was elsewhere.  I'll see about factoring out the duplication.

interface/src/main/scala/quasar/main/Quasar.scala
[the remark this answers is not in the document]
djspiewak: One thing we could do which might help is add a synthesized `BackendConfig` value to the console initial lines, where the synthesized `BackendConfig` is automatically generated by SBT on each run.  This would have the unfortunate effect of making `console` and `consoleQuick` as slow as `repl/run` and `web/run` though.

web/src/main/scala/quasar/server/Server.scala
djspiewak: We could, but there's really a reason to do that.  We're not going to handle the error; in fact, we very much want it to be unrecoverable.  We don't need any special logging for it.  With this approach, quasar is going to crash and print to stderr, which is exactly what we want.

I have locally refactored this somewhat to thread the `Task` through from `fromFile`.  It's certainly possible to move the error into `Failure`, but that further complicates an already annoyingly-verbose bit of lifting and wrapping.

web/src/main/scala/quasar/server/Server.scala
jedesah: Okay, I'll buy that it makes reading a little easier if used judiciously. Just had to call you out on "lots of problems" 😄 . In this particular case I don't think the newline actually helps, just like I said initially it's just a question of habit and what a given person is used to parsing. Won't bikeshed this any further.

web/src/main/scala/quasar/server/Server.scala
jedesah: > annoyingly-verbose bit of lifting and wrapping

Plenty of that to spread around in Quasar. It sucks and I have some vague ideas on macros and techniques that might help it, but it's the trade-off for very precise types I think. Granted it's less valuable at the edge of the world.

interface/src/main/scala/quasar/main/Quasar.scala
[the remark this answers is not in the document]
jedesah: Ya, that could be cool. Don't know how slow `repl/run` and `web/run` are now so coudn't comment on the validity of the trade-off. Perhaps `console` might become more valuable if it's faster to startup for ad-hoc testing

it/src/test/scala/quasar/server/ServiceSpec.scala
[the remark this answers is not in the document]
jedesah: Ah, just wasn't sure if you had actually seen it flake. Then there is something to figure out because it would seem to me at least that the operations should be deterministic. `flakyTest` has an overloaded version which takes a `String` as an explanation. I think it would be better if the little blurb you wrote was a parameter to `flakyTest` so that it prints the explanation for why we think it's flay

it/src/test/scala/quasar/TestConfig.scala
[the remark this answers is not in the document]
jedesah: Ah, I see.

foundation/src/main/scala/quasar/contrib/pathy/package.scala
```diff
@@ -52,20 +53,26 @@ package object pathy {
 
   object ADir {
 
-    // performs some effects; I'm probably going to regret this...
-    def fromFile(file: JFile): Option[ADir] = {
-      val check = file.exists() && file.isDirectory()
-      // trailing '/' is significant!  yay, pathy...
-      posixCodec.parseAbsDir(file.getAbsolutePath + "/").map(unsafeSandboxAbs).filter(_ => check)
+    def fromFile(file: JFile): OptionT[Task, ADir] = {
```
jedesah: I still think that this is conflating two concerns. I understand that this what the single usage site wants and if this were defined in that file as a utility function, I wouldn't say anything. But since this is defined in `contrib`, which is intended for more general-purpose utilities that we may want to contribute upstream, I think the two concerns should be separated.

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
```diff
@@ -40,6 +40,11 @@ package object scalaz {
     }
   }
 
+  implicit final class OptionTOps[F[_], A](val self: OptionT[F, A]) extends AnyVal {
```
jedesah: I think this could be defined over any `Functor`

it/src/test/scala/quasar/server/ServiceSpec.scala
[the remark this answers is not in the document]
djspiewak: I wouldn't just add `flakyTest` arbitrarily.  It fails, and it fails frequently.  I don't really see a need to elaborate in this case because the test is straightforward and the only thing I'd be able to write could have been inferred just as quickly by any reader.

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
djspiewak: Probably.  We can generalize it later.

foundation/src/main/scala/quasar/contrib/pathy/package.scala
djspiewak: I disagree.  It is meaningless to convert a `File` into any sort of `Path` value *without* checking whether or not it is a directory or a file.  I think it's valid to complain about the existence check, but the problem is that file/directory tests are meaningless without existence.

README.md
```diff
@@ -169,6 +169,63 @@ java -jar [<path to jar>] [-c <config file>]
...
+
+This technique is designed for local development use, where the backend implementation is changing frequently.  Under certain circumstances though, it may be useful for the pre-built JAR case.
+
+As with the plugins directory approach, you will need to run the `assembly` task for each backend that you want to use.  But instead of copying the JAR files into a directory, you will be referencing each JAR file individually using the `--backend` switch on the web or REPL JAR invocation:
+
+```bash
+java -jar [<path to jar>] [-c <config file>] --backend:quasar.physical.mongodb.MongoDb\\$=.targets/mongodb/scala-2.11/quasar-mongodb-internal-assembly-23.1.5.jar
+```
+
+Replace the JAR file in the above with the path to the backend who's `assembly` you ran.  The `--backend` switch may be repeated as many times as necessary: once for each backend you wish to add.  The value to the left of the `=` is the `BackendModule` object *class name* which defines the backend in question.  Note that we need to escape the `$` character which will be present in each class name, solely because of bash syntax.
```
alissapajer: s/who's/whose/

README.md
```diff
@@ -169,6 +169,63 @@ java -jar [<path to jar>] [-c <config file>]
...
+
+This technique is designed for local development use, where the backend implementation is changing frequently.  Under certain circumstances though, it may be useful for the pre-built JAR case.
+
+As with the plugins directory approach, you will need to run the `assembly` task for each backend that you want to use.  But instead of copying the JAR files into a directory, you will be referencing each JAR file individually using the `--backend` switch on the web or REPL JAR invocation:
+
+```bash
+java -jar [<path to jar>] [-c <config file>] --backend:quasar.physical.mongodb.MongoDb\$=.targets/mongodb/scala-2.11/quasar-mongodb-internal-assembly-23.1.5.jar
+```
+
+Replace the JAR file in the above with the path to the backend who's `assembly` you ran.  The `--backend` switch may be repeated as many times as necessary: once for each backend you wish to add.  The value to the left of the `=` is the `BackendModule` object *class name* which defines the backend in question.  Note that we need to escape the `$` character which will be present in each class name, solely because of bash syntax.
```
alissapajer: s/who's/whose/

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
jedesah: R.I.P. parametricity 😞

djspiewak: I just want to get this in. There’s so much stuff going on right now. It’s not that I think generalizing it is bad or even trivial, I just don’t want to hold things up longer than we have to.

it/src/test/scala/quasar/server/ServiceSpec.scala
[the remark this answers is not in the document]
jedesah: I'm not saying you need to expand on what you already wrote as a comment. Just that you should convert the comment into a `String` parameter to the `flakyTest` method. That way we see a brief explanation for the flakiness in the test output

foundation/src/main/scala/quasar/contrib/pathy/package.scala
jedesah: I see your point. I didn't grasp until now that `file.isDirectory` always returns `false` when the path does not exist. Might as well get rid of `file.exists` then in that expression as it's redundant right? (although perhaps replaced by a comment saying that `isDirectory` also checks for existence)

djspiewak:
@jedesah @alissapajer I think this is ready for merge now?

PR #2757, opened 2017-09-13
Fix mount fs issue

### LINKED ISSUE #2758: initial request to`/mount/fs` fails
In master we regressed the ability to use the `/mount/fs` endpoint when there's on the first initial mount. Folks are reporting the following error:

```
 [53] → curl -v -X PUT http://localhost:8080/mount/fs/mongodb34/ -d '{ "mongodb": { "connectionUri": "mongodb://localhost:27022" } }'
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> PUT /mount/fs/mongodb34/ HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.51.0
> Accept: */*
> Content-Length: 63
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 63 out of 63 bytes
< HTTP/1.1 500 Connection to backend failed.
< Content-Type: application/json; charset=UTF-8
< Date: Fri, 08 Sep 2017 18:09:23 GMT
< Content-Length: 199
<
* Curl_http_done: called premature == 0
* Connection #0 to host localhost left intact
{ "error": { "status": "Connection to backend failed.", "detail": { "message": "Connection failed: java.util.concurrent.TimeoutException: Timed out attempting to connect to: [localhost:27022]." } } }
```

nicflores:
Fixes the `/mount/fs` regression.

Fixes #2758


web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -25,6 +25,7 @@ import quasar.fs.cache.{VCache, ViewCache}
 import quasar.fs.mount._
 import quasar.fs.ManageFile
 
+
```
jedesah: Extra newline

alissapajer: Since we're going to have scalafmt in the not too distant future, can we let styling things slide for now?

alissapajer:
@nicflores did you know if this fixes the failing `./scipts/uploadTestData` for mongo?

web/src/main/scala/quasar/api/services/mount.scala
jedesah: I'm okay with letting a few styling things slip, but I'd rather not add random newlines that were probably unintentional for no particularly good reason

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -144,36 +145,49 @@ object mount {
...
-                    κ(vcache.modify(p, _.copy(
-                      query = nvc.query,
-                      maxAgeSeconds = nvc.maxAgeSeconds,
-                      refreshAfter = nvc.refreshAfter,
-                      status = nvc.status,
-                      dataFile = nvc.dataFile))),
-                    vcache.put(p, nvc )).join
-                }.orZero.liftM[ApiErrT]
+      _      <- (refineType(path).toOption ⊛ MountConfig.viewConfig.getOption(bConf).map(MountConfig.ViewConfig.tupled) ⊛ maxAge)(createNewViewCache[S])
+                  .cata(ι,().point[EitherT[Free[S, ?], ApiError, ?]])
```
jedesah: Can we use `getOrElse` here instead?

jedesah: Also, inference with `getOrElse` might be better (or not) and we might be able to get rid of the type parameter on `point` which in this case I think would be nice (sometimes it's nice to have explicit types, but I don't think here it's doing much for us if it can be avoided)

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -144,36 +145,49 @@ object mount {
...
-                  vcache.get(p).fold(
-                    κ(vcache.modify(p, _.copy(
-                      query = nvc.query,
-                      maxAgeSeconds = nvc.maxAgeSeconds,
-                      refreshAfter = nvc.refreshAfter,
-                      status = nvc.status,
-                      dataFile = nvc.dataFile))),
-                    vcache.put(p, nvc )).join
-                }.orZero.liftM[ApiErrT]
+      _      <- (refineType(path).toOption ⊛ MountConfig.viewConfig.getOption(bConf).map(MountConfig.ViewConfig.tupled) ⊛ maxAge)(createNewViewCache[S])
```
jedesah: The one thing that bugs me a little bit about this is that if we fail to setup the view cache properly we will return an error to the user but we will still have mounted the view. We could handle a failure in setting up the view cache and unmount the view but then the unmounting might fail...

I guess what we would really need here is transactionality and I think that's beyond the scope of this PR but I thought I would mention it nonetheless.

jedesah:
Just the newline and the `getOrElse` but otherwise looks good. Would like to see those changes but I won't hold up this PR on it as it would be nice to get this merged sooner rather than later I suppose

djspiewak:
Is there a ticket for this?

nicflores:
@alissapajer, it is possible that the `uploadTestData` issue you were seeing is/was unrelated. I was able to run the same mount command you showed me initially and then run `uploadTestData` immediately afterwards without any problems.

nicflores:
@djspiewak, there was no ticket for this issue. @alissapajer and I first started talking about it then @jedesah and I started on a solution and then @drostron and I finished it today.

alissapajer:
@nicflores today @djspiewak was also unable to run `uploadTestData` to mongo on master. So it sounds like perhaps this PR fixed it. I'll test when this lands.

alissapajer:
@nicflores also, I think it might be worth creating an issue for this which gets closed with this PR. That way we can search for it more easily in the future, and it also tracks the work we're doing.

nicflores:
Created related issue: #2758

alissapajer:
@nicflores are there any unit tests we can add to ensure (1) this fixes the problem and (2) the problem won't surface again?

nicflores:
@alissapajer, new integration test had been added to check that `/mount/fs` works.

PR #2759, opened 2017-09-14
Remove old-spark leftovers

rabbitonweb:
After moving all spark based connectors on backend module, this removes some old code plus unifies any duplications


rabbitonweb:
@alissapajer this PR cleans up some left overs after introducing backend module to spark. Also some renaming applied for readability. Would be also cool to have it in master ASAP (before dynamic loading) so if review possible today that would be awesome

alissapajer:
@djspiewak FYI this will cause minor conflicts in #2756 when it lands.

PR #2767, opened 2017-09-15
Fix format for h2 config in the README

garyb:


garyb:
@kRITZCREEK tells me that the `location` version of this works for the endpoints, but this is definitely right for the config. I guess that might be worth opening an issue for?

alissapajer:
@garyb yeah, can you open issue for that? I'm not sure if that discrepancy would be expected or not, but we can discuss there.

PR #2768, opened 2017-09-15
sort single value reduction

### LINKED ISSUE #2750: `select count(*) from x order by y` yields broken queries in new Mongo
For example, 

`SELECT COUNT(*) AS cnt FROM /mngo/jps/zips ORDER BY cnt DESC` results in Mongo throwing an exception.

whereas 

`SELECT COUNT(*) as cnt FROM /mngo/jps/zips` yields the correct `db.zips.count()` plan

Discovered as part of #2667.

rintcius:
Fixes #2750

### DIFF
--- a/connector/src/main/scala/quasar/qscript/Transform.scala
+++ b/connector/src/main/scala/quasar/qscript/Transform.scala
@@ -371,7 +371,7 @@ class Transform
                 Free.point(ReduceIndex(0.right)))))).embed)
         case None =>
           Target(
-            Ann(join.base.buckets, HoleF),
+            Ann(Nil, HoleF),
             QC.inj(Reduce[T, T[F]](
               values(0).value,
               Nil,
--- a/connector/src/main/scala/quasar/qscript/Transform.scala
+++ b/connector/src/main/scala/quasar/qscript/Transform.scala
@@ -337,7 +337,7 @@ class Transform
                 Free.point(ReduceIndex(0.right)))))).embed)
         case None =>
           Target(
-            Ann(provs, HoleF),
+            Ann(Nil, HoleF),
             QC.inj(Reduce[T, T[F]](
               values(0).value,
               Nil,
--- a/connector/src/test/scala/quasar/qscript/QScript.scala
+++ b/connector/src/test/scala/quasar/qscript/QScript.scala
@@ -730,6 +730,52 @@ class QScriptSpec
         implicitly, Corecursive[Fix[QS], QS])))
     }
 
+    "convert a sort of a single value unary reduction" in {
+      val lp = fullCompileExp(sqlE"select count(*) as cnt from zips order by cnt")
+      val qs = convert(lc.some, lp)
+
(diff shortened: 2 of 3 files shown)


rintcius:
Note: there's also a similar piece of code for binary reductions, which may also have this problem: https://github.com/quasar-analytics/quasar/blob/master/connector/src/main/scala/quasar/qscript/Transform.scala#L372-L379

However the only binary reduction is `UnshiftMap` of which there seems to be no usage and I don't know how to test this. If someone knows an example I can add a test.

alissapajer:
@rintcius can you add a unit test for this query over in https://github.com/quasar-analytics/quasar/blob/master/connector/src/test/scala/quasar/qscript/QScript.scala? Even if the qscript isn't ideal yet, it would be good to track it.

alissapajer:
@rintcius you can generate `UnshiftMap` directly with `select {state : pop ...} from zips` (and `UnshiftArray` as `select [pop ...] from zips`). Mongo 3.4 doesn't have `UnshiftMap` implemented, and I'm actually not sure if any connectors do. (As a side note, you can work backwards from [the parser](https://github.com/quasar-analytics/quasar/blob/master/frontend/src/main/scala/quasar/sql/parser.scala#L334) to figure out how to manually generate certain nodes.)

Additionally we create `UnshiftMap` over in [`ShiftRead`](https://github.com/quasar-analytics/quasar/blob/master/connector/src/main/scala/quasar/qscript/ShiftRead.scala#L73) but then elide the `UnshiftMap` over in [`Coalesce`](https://github.com/quasar-analytics/quasar/blob/master/connector/src/main/scala/quasar/qscript/Coalesce.scala#L359) when we can. So any query where the `UnshiftMap` wasn't elided in that way would also contain one.

If no connectors have `UnshiftMap` implemented, you could still write a qscript unit test for it and verify the provenance by manually verifying the qscript.

alissapajer:
Also, I'm actually a bit surprised that no other integration tests came to life with this fix.

jsantos17:
Might be worth checking other queries with `count(*)` in the [unit tests](https://github.com/quasar-analytics/quasar/blob/master/mongodb/src/test/scala/quasar/physical/mongodb/planner.scala#L1462). Some tests might be fixed but still marked as pending because of different generated names or something else irrelevant.

alissapajer:
@jsantos17 :+1: good idea, any non-bucketed reduction would fall into this case.

rintcius:
@alissapajer Updated binary reduction as well. And added unit tests for both unary and binary reduction.

The problem I had regarding `UnshiftMap` was more semantically than syntactically. Anyway, with the unit test it's now at least tracked.

@jsantos17 @alissapajer Regarding the tests in `mongodb/planner`, I went through those queries already - and went through them again just now - but none of them seem to be affected by this fix.  However let me know if I missed one.

jsantos17:
@rintcius sad to hear that's the case. I'll give it sometime later today.

PR #2782, opened 2017-09-19
Handle errors in linking a bit more gracefully

### LINKED ISSUE #2765: Mimir doesn't work on windows
Steps to reproduce using SD master (2017.09.15) 
Quasar 23.1.2 and 20.14.13

+ Mount mimir somehow
+ Try to upload any file 
+ Open same uri in other page

Expected result 
+ File appears in fs ui
+ Fs ui in other page is loaded

Actual result
+ PUT requests doesn't return anything. Loading spins forever
+ Fs ui is blocked. Loading spins forever

Same behaviour for workspaces 
+ Mount mimir
+ Drilldown to mimir folder
+ Create workspace
+ Insert query card
+ Put any valid query (e.g. 
```sql
SELECT COUNT(*) AS ct, country FROM `/local/local/olympics` GROUP BY country
```
+ Press `Run query`
+ Slide to the next card

Expected result
+ Next card action list appears

Actual result 
+ Spinner. `PUT` doesn't return anything. 

If you go to mimir folder in fs ui, it doesn't work.

djspiewak:
Fixes #2765 (I believe…)

Windows was failing because creating symlinks on windows… fails.  So now we handle those errors and return `false`, which the VFS in turn ignores because it doesn't really care whether or not the symlink is *actually* created.

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
@@ -33,7 +33,7 @@ import scalaz.syntax.traverse._
 
 import scodec.bits.ByteVector
 
-import java.io.File
+import java.io.{File, IOException}
 import java.nio.file.{Files, StandardCopyOption}
 import java.util.UUID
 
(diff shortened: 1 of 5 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
```diff
@@ -90,19 +90,27 @@ object RealPOSIX {
...
 
           Task delay {
-            Files.createSymbolicLink(ptarget, psrc)
+            try {
+              Files.createSymbolicLink(ptarget, psrc)
 
-            ()
+              true
+            } catch {
+              case _: IOException => false
```
nicflores: will ignoring the possibility of symbolic links not being created affect the way mimir works on windows?

djspiewak: Nope.  The symlinks are solely for human consumption.  Mimir itself ignores them.

Issue #2751, opened 2017-09-11
Temporary cached data sets are not purged


natefaubion:
Since I upgraded to v21 and enabled a cached view, my directory listing is now littered with thousands of unique `__quasar.tmp` files. They all appear to point to data snapshots for my cached query.

<img width="668" alt="screen shot 2017-09-11 at 11 32 11 am" src="https://user-images.githubusercontent.com/144058/30290630-e9bd4720-96e4-11e7-90ff-681d2fe22708.png">

A couple of things:

* These should definitely be cleaned up
* We should make these dotfiles so that the FS view does not show them by default.

zeeiyerWork:
@drostron -> It seems that 
a> older copies of cached content should not stick around (perhaps we leave 1-2 versions around just in case).
b> If a user deletes a deck (with a cached card) or just a cached card within a deck, I hope we are plugged in to those events to tear-down cache appropriately.

natefaubion:
> If a user deletes a deck (with a cached card) or just a cached card within a deck, I hope we are plugged in to those events to tear-down cache appropriately.

In my PR, I issue a DELETE to card mounts. There aren't any events for the backend to respond to other than that, but it should cleanup tmp files if the original view is deleted.

drostron:
We can and should cleanup temp files.

Issue #2740, opened 2017-09-08
test failure in `quasar.fs.cache.VCacheH2Spec`


alissapajer:
```
[error] ! move a key (1 second, 17 ms)
[error]  org.h2.jdbc.JdbcSQLException: Exception raised on property evaluation.
[error]  > ARG_0: FileIn(Root,FileName(y))
[error]  > ARG_1: ViewCache(ConnectionUri(:),None,None,-709275496,None,None,9223372036854775807,1988-08-19T04:38:48.729Z,Successful,None,FileIn(Root,FileName(9)),None)
[error]  > ARG_2: FileIn(Root,FileName("))
[error]  > Exception: org.h2.jdbc.JdbcSQLException: Unique index or primary key violation: "PRIMARY_KEY_F ON PUBLIC.VIEW_CACHE(PATH) VALUES ('/""', 6)"; SQL statement:
[error]  INSERT INTO view_cache values (
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?,
[error]              ?, 
[error]              ?,
[error]              ?) [23505-196] (DbException.java:345)
[error] org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
[error] org.h2.message.DbException.get(DbException.java:179)
[error] org.h2.message.DbException.get(DbException.java:155)
[error] org.h2.index.BaseIndex.getDuplicateKeyException(BaseIndex.java:103)
[error] org.h2.mvstore.db.MVSecondaryIndex.checkUnique(MVSecondaryIndex.java:231)
[error] org.h2.mvstore.db.MVSecondaryIndex.add(MVSecondaryIndex.java:190)
[error] org.h2.mvstore.db.MVTable.addRow(MVTable.java:705)
[error] org.h2.command.dml.Insert.insertRows(Insert.java:156)
[error] org.h2.command.dml.Insert.update(Insert.java:114)
[error] org.h2.command.CommandContainer.update(CommandContainer.java:101)
[error] org.h2.command.Command.executeUpdate(Command.java:260)
[error] org.h2.jdbc.JdbcPreparedStatement.executeUpdateInternal(JdbcPreparedStatement.java:164)
[error] org.h2.jdbc.JdbcPreparedStatement.executeUpdate(JdbcPreparedStatement.java:150)
[...]

PR #2783, opened 2017-09-19
Clean vcache temp files

drostron:
Fixes: #2751 
Fixes: #2740


drostron:
@jedesah Assigning you since you're familiar with the original vcache PR and we paired on the circular interpreter dependency untanglement.

core/src/main/scala/quasar/metastore/MetaStoreAccess.scala
```diff
@@ -56,9 +56,9 @@ trait MetaStoreAccess {
   def lookupViewCache(path: AFile): ConnectionIO[Option[ViewCache]] =
     (Queries.lookupViewCache(path) ∘ (_.vc)).option
 
-  def updateInsertViewCache(path: AFile, viewCache: ViewCache): ConnectionIO[Unit] =
-    runOneRowUpdateOpt(Queries.updateViewCache(path, viewCache)).isEmpty >>= (_.whenM(
-      runOneRowUpdate(Queries.insertViewCache(path, viewCache))))
+  def insertUpdateViewCache(path: AFile, viewCache: ViewCache): ConnectionIO[Unit] =
```
jedesah: Since we are changing the name, can we go with `insertOrUpdateViewCache`?

Also, is this change what fixes the flaky test?

drostron: A reasonable name update and yes, this is to address the flaky test.

drostron: The change is in a separate commit for future convenience.

drostron:
@jedesah Ready for another look

core/src/main/scala/quasar/fs/mount/cache/VCache.scala
```diff
@@ -75,8 +75,13 @@ object VCache {
     case CompareAndPut(k, expect, v) =>
       injectFT[ConnectionIO, S].apply(lookupViewCache(k)) >>= (vc =>
         (vc ≟ expect).fold(
-          (vc.traverse(deleteVCacheFiles[S]) >>
-            injectFT[ConnectionIO, S].apply(insertOrUpdateViewCache(k, v))).as(true),
+          {
+            vc.foldMap(c =>
+              deleteFiles(
+                ((c.dataFile ≠ v.dataFile) ?? List(c.dataFile)) :::
```
jedesah: I would add a comment here saying something like: "if the data files are the same in the new view cache as the old one, let's not wipe them away and invalidate the cache". Perhaps I'm on a literate programming fad :smile:  but I'm becoming a huge fan of writing a natural language accompaniment when it helps grok the code faster than just looking at the implementation

jedesah: And it's especially easy to spot places where it took me longer to understand the code than a sentence would have taken me when reviewing someone else's code so I take the opportunity to point them out! :smile:

interface/src/main/scala/quasar/main/ViewCacheRefresh.scala
```diff
@@ -100,18 +102,20 @@ object ViewCacheRefresh {
...
+  ): Q.transforms.CompExecM[Unit] = {
     val fsQ = new FilesystemQueries[S]
 
     for {
       q       <- EitherT(EitherT(
                    (quasar.resolveImports_[S](view.query, basePath).leftMap(nels(_)).run.run ∘ (_.sequence))
                      .liftM[PhaseResultT]))
-      r       <- fsQ.executeQuery(q, view.vars, basePath, tmpDataPath)
-    } yield r
+      r       <- fsQ.queryResults(q, view.vars, basePath, 0L, none)
```
jedesah: What's the reason for this change?

core/src/main/scala/quasar/fs/mount/cache/VCache.scala
drostron: Happy to oblige

core/src/test/scala/quasar/fs/mount/ViewFileSystemSpec.scala
```diff
@@ -442,7 +442,7 @@ class ViewFileSystemSpec extends quasar.Qspec with TreeMatchers {
       val f = manage.delete(p).run
 
       viewInterpTrace(vc, f) must_=== ViewInterpResultTrace(
-        Vector.empty,
+        traceInterp(f, Map())._1,
```
jedesah: What's this change for?

interface/src/main/scala/quasar/main/ViewCacheRefresh.scala
drostron: `executeQuery` uses the given output file _if possible_. This keeps us from having to detect and handle that case separately.

jedesah: Interesting. Not quite certain what _if possible_ means. Like it won't overwrite an existing file? Anyway, I now understand the change

drostron: The example I ran into was a simple query like `select * from zips` returns `.../zips` instead of the provided output file. Since the query result if equivalent to `.../zips` we don't need to create a separate file for those results. When view caching is involved we need to store the result regardless.

jedesah: Ah, I see. Ya, I think it's cool and all that `executeQuery` tries to be smart that way but it also makes it less useful. Here we are going to be streaming all these cached view results in and out of Quasar for no particularly good reason because of this little bit of "smartness" that is probably not used all that often. We might want to revisit that because this will have a serious performance impact on the common case for view caches

core/src/test/scala/quasar/fs/mount/ViewFileSystemSpec.scala
drostron: Per [this change](https://github.com/quasar-analytics/quasar/pull/2783/files/ff6bec3ae5c0e0fd6cf052c84c48d90b7f24108b#diff-9ba21f4161d21d904bf64452dd57b37aR165)

interface/src/main/scala/quasar/main/ViewCacheRefresh.scala
jedesah: If we think it's useful, we could add a flag to turn that on and off. I think it would definitely be useful to have a "strict" version of `execute` which does exactly what it's told for cases like this

drostron: A fair concern. I'll look at what's involved in making that change to `executeQuery`.

wemrysi: @drostron Would it be possible to set the view cache's data file to be the path returned from `executeQuery` if it differs from the argument path? Would we then need a way to signal that it shouldn't be automatically deleted?

drostron: @wemrysi For sure, and I originally went that direction but switched due to the absence of a copy term that can be pushed down to a connector. Signaling is possible but doesn't take care of the caching aspect. In this case the cache would be more of a plain view. In fact before this change we were falling back to the associated view anyway albiet with a continually failing vcache refresh. Also, @jedesah and I were discussing the possibility of just removing this special behavior from `execute`. Perhaps you have a sense for why we might preserve it? It seems if you don't want results written you'd just use `evaluate`.

wemrysi: I wasn't sure if what I suggested would be easier than updating `ExecutePlan`, so I just wanted to mention it.

Getting rid of the `ExecutePlan` maybe-it'll-write-maybe-it-won't behavior is fine with me.

drostron: Sounds good, I'll roll with updating `ExecutePlan` then.

drostron:
@jedesah Ready for another look

Issue #2532, opened 2017-07-24
Eliminate Query Regressions in QScript-based MongoDB Connector


wemrysi:
This issue exists to track the progress of eliminating the regressions, individual regressions will have their own issues and will be associated with the query they're affecting above.

The following queries have regressed in the QScript-based MongoDB connector:

- [x] caseWithGroup.test (#1755, #2847)
- [x] different-flattens.test (#1951, #1929)
- [x] flattenGroupedObject.test (#2359)
- [x] groupByFlatten.test (#2537)
- [x] joins/fullOuterEquiJoin.test (fixed in #2562)
- [x] joins/groupedJoin.test (#2121)
- [x] joins/joinWithComplexCondition.test (fixed in #2562)
- [x] joins/nonJsJoinCondition.test (#2110)
- [x] joins/selfjoinWithComplexCondition.test (fixed in #2562)
- [x] joins/symmetricNonJsJoinCondition.test (#2110)
- [ ] mongodb/fakeObjectId.test (`OID` function not available in QScript)
- [ ] multistageFlatten.test (#2540)
- [ ] multitypeFlatten.test (#2133)
- [ ] temporal/invalidDateFilter.test (#2541)
- [x] undefinedJsInExpr.test (possibly a problem with the test itself, see `FIXME` therein)
- [ ] unifyFlattenedFieldsWithMultipleUses.test (#2366)
- [x] union.test (#2543)

The following queries have regressed (or are not passing) in mongo 2.6 and 3.0, but are passing in mongo 3.2 and 3.4:

- [ ] basicAnalyticQuery.test (pre-3.2, #2535)
- [ ] caseVariations.test (#2535, #2785) (pre-3.2 passing with result order ignored)
- [ ] citiesByLargestZip.test (#2535)
- [ ] citiesByTotalPopulation.test (pre-3.2, #2535)
- [ ] concatArrayWithLiteral.test (pre-3.2, #2536)
- [ ] doubleFlatten.test (pre-3.2, #2544)
- [ ] doubleFlattenWithIntervening.test (pre-3.2, #2544)
- [ ] filterSortLimit.test (pre-3.2 passing with result order ignored)
- [ ] flattenArray.test (pre-3.2, #2545)
- [ ] flattenNestedObject.test (pre-3.2, #2545)
- [ ] flattenObject.test (pre-3.2, #2545)
- [ ] joins/orderJoinByAlias.test (#2725 #2785 pre 3.2 passing with result order ignored)
[...]

PR #2785, opened 2017-09-20
Preserve sorting after a MR-requiring selection

### LINKED ISSUE #2725: Preserve sorting after a MapReduce-requiring selection in MongoDB
We need to detect the case in the mongo planner where the sort key is both a selection and requires a map reduce, and rewrite the plan to sort after the select. An example query:
```sql
select length(city) from extraSmallZips order by length(city)
```

Note that if the selection requires a map reduce but is not a sort key, then the ordering will not be lost.

This is part of #2535.

alissapajer:
Closes #2725


alissapajer:
I plan to add an integration test for
```sql
select city ?? "foo" as name from smallZips order by city ?? "foo"
```
That query is returning expected results in the REPL, but the integration test predicate isn't being respected, so I just need to figure that out.

alissapajer:
Once we have a green build I'll update #2532.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1289,6 +1290,31 @@ object MongoDbPlanner {
...
+      : QScriptCore[T, T[G]] => F[T[G]] = {
+    case qs @ Map(Embed(src), fm) =>
+      GtoF.get(src) >>= QC.prj match {
+        case Some(Sort(innerSrc, bucket, order)) =>
+          val innerMap = GtoF.reverseGet(QC.inj(Map(innerSrc,
+	    // qscript <3 arrays
+            Free.roll(MFC(MF.ConcatArrays(
+              Free.roll(MFC(MF.MakeArray(fm))),
+              Free.roll(MFC(MF.MakeArray(HoleF[T]))))))))).embed
+          QC.inj(Map(
```
wemrysi: It's unfortunate that we still end up with a `Map` after the `Sort` here. I wonder if we could detect whether the sort keys are still available in the result of applying the `Map` expression to the source of the `Sort` and then just update the key expressions and avoid the outer `Map`?

connector/src/main/scala/quasar/qscript/package.scala
```diff
@@ -323,6 +325,59 @@ package object qscript {
...
+
+    val partial: F[Free[F, Hole]] => CoEnvFree[F, Hole] = fa => {
+      liftAlgebra[T, F, CoEnv[Hole, F, ?]](alg, coenvPrism[F, Hole])
+        .apply(fa ∘ bij.toK.run) ∘ bij.fromK.run
+    }
+
+    liftCo[T, F, Hole, Free[F, Hole]](partial)
+  }
+
+  trait Alg[T[_[_]]] {
```
wemrysi: What do you think of naming this `Trans` instead, since it's more of a transformation than an algebra? Just to be consistent with matryoshka.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
alissapajer: Agreed. I'll address this.

alissapajer: Consider the following query followed by its qscript before the `mapBeforeSort` rewrite. In this case we don't have access to the sort keys after applying the `Map`, and so I'm not sure how we can elide the outer `Map` (which agrees with your comment, if I understood correctly).
```sql
select city ?? "foo" as name from `smallZips.json` order by state ?? "bar"
```
```
Subset(Take)
├─ Unreferenced
├─ From
│  ╰─ Map
│     ├─ Sort
│     │  ├─ ShiftedRead(/test/smallZips.json, ExcludeId)
│     │  ├─ Bucket
│     │  │  ╰─ List
│     │  ╰─ Order
│     │     ╰─ Ascending
│     │        ╰─ By
│     │           ╰─ IfUndefined
│     │              ├─ Guard
│     │              │  ├─ ○
│     │              │  ├─ Type(Obj(Map[], Some(Top)))
│     │              │  ├─ ProjectField
│     │              │  │  ├─ ○
│     │              │  │  ╰─ Constant(Str(state))
│     │              │  ╰─ Undefined
│     │              ╰─ Constant(Str(bar))
│     ╰─ MakeMap
│        ├─ Constant(Str(name))
│        ╰─ IfUndefined
│           ├─ Guard
│           │  ├─ ○
│           │  ├─ Type(Obj(Map[], Some(Top)))
│           │  ├─ ProjectField
│           │  │  ├─ ○
│           │  │  ╰─ Constant(Str(city))
│           │  ╰─ Undefined
│           ╰─ Constant(Str(foo))
╰─ Count
   ╰─ Map
      ├─ Unreferenced
      ╰─ Constant(Int(11))
```
If we instead consider the query 
```sql
 select city ?? "foo" as name from `smallZips.json` order by city ?? "foo"
```
then we could rewrite the qscript to be like
```
Sort(
  Map(
    ShiftedRead(smallZips),
    MakeMap("name", Guard))
  List(),
  ProjectField("name", Hole))
```
The source of the `Sort` has to be equivalent to what we return after running the outer `Map` in the original query, so that has us constrained. Furthermore we may have to compute an "inverse" mapping operation, for example in this case the sort key is `ProjectField` because that's the "inverse" of `MakeMap`. And if there were buckets then we'd need to retain access to them as well in the source of the `Sort`.

So I think in certain particular cases, where there are no buckets and where the sort key is very similar to the selection, we could remove the outer `Map` entirely. This solution only solves the problem for a small subset of cases though, and I'm not convinced these cases are important enough to warrant the extra work?

@wemrysi let me know what you think, or if you see another way around this.

wemrysi: Hm, yeah I see what you're saying. I brought it up as the original ticket described only attempting this when we knew the sort key was available in the selection.

However, after thinking about it some more, your solution here is more general (at the cost of an extra `Map`) as, for 3.2+, we're trading an arbitrary expression after the sort, which always has the chance of requiring map-reduce, for a known expression (array projection) that we know can be done via aggregation.

PR #2790, opened 2017-09-21
Made web/run and repl/run fast again, because I'm awesome like that

djspiewak:
Fixes @alissapajer's sadness


alissapajer:
Confirmed `repl/run` now takes ~10 seconds, which is on par with before we landed dynamic loading. @djspiewak thank you thank you!

PR #2792, opened 2017-09-21
Added githubReleaseAll task

djspiewak:
This should fix connector JAR pushing, which isn't currently working.  Have to see it run in Travis (in master) to be sure.


nicflores:
Looks good. Lets see it run, will approve and merge. Thanks @djspiewak.

PR #2801, opened 2017-09-22
Track pending mongodb planner tests

rintcius:
This adds a specs2 trait `PendingWithActualTracking` which enables tracking of the actual result.

This helps in detecting changes in behaviour concerning tests that are pending (notably in `mongodb.planner`).

Also added a trait `PendingWithActualTrackingWriteMode` that helps in generating new actuals.


rintcius:
@alissapajer @jsantos17 @wemrysi 
I'd like to make the `mongodb.planner` tests that are `pending` more strict, because now we are basically in limboland with all these pending tests. Can I get your feedback on this mechanism? Once it's ok with you, I'll change all applicable `pending` tests in `mongodb.planner` in the same way and add to this PR.

jsantos17:
I really, really like this idea. Not sure I like to do this at the string-level tho. Could we do something similar but at the workflow level? Specify the last known failing plan just as we do with expectations. We can get better diffs that way too!

I can see this becoming annoying to maintain after a while, but we shouldn't have many pending tests for long. 

Other than that: 💯

rintcius:
@jsantos17  Yes, I thought about specifying the actual objects but opted for strings so that it is less time consuming to create and maintain. Maybe we can move towards a non-string based implementation after most of these `pending`s have been resolved so that it becomes less cumbersome to maintain?

foundation/src/test/scala/quasar/contrib/specs2/PendingWithActualTracking.scala
```diff
@@ -0,0 +1,50 @@
...
+import slamdata.Predef._
+
+import org.specs2.execute._
+
+/**
+  * Specs2 trait that is like "pendingUntilFixed" but additionally tracks
+  * actual results. It fails a test in case the expected actual is not equal to
+  * the (actual) actual.
+  */
+trait PendingWithActualTracking {
```
djspiewak: This trait should really have a different name, since you're currently shadowing it with an inner-type, which seems like a recipe for disaster (or at the very least, confusion).  `Capability`, `Module`, `Component` and such are pretty solid generic "container trait" suffixes, if there isn't something better.

foundation/src/test/scala/quasar/contrib/specs2/PendingWithActualTracking.scala
```diff
@@ -0,0 +1,50 @@
...
+    : PendingWithActualTracking[T] = new PendingWithActualTracking(t)
+
+  private def c(msg: String, extra: String) = if (msg == "") msg else msg + extra
+
+  class PendingWithActualTracking[T : AsResult](t: =>T) {
+    def pendingWithActual(m: String, expectedActual: String): Result = ResultExecution.execute(AsResult(t)) match {
+      case s @ Success(_,_) =>
+        Failure(m + " Fixed now, you should remove the 'pendingWithActual' marker")
+      case f @ Failure(msg, e, stackTrace, FailureDetails(actual, expected)) =>
+        if (actual != expectedActual)
```
djspiewak: This should definitely be done in a non-stringish manner.  Though, it only changes the failure/pending status, so it's less problematic than it could be.

jsantos17:
As this is intended to be a short term aid I'd be okay with keeping the strings. We'd also need an issue to remove this so we don't forget.

I still think we should evaluate how unrelated changes could lead to fictional, annoying breakage. An immediate thought is that changes to `RenderTree` would force you to rewrite the string.

foundation/src/test/scala/quasar/contrib/specs2/PendingWithActualTracking.scala
rintcius: I hope it's not a recipe for disaster, because I (think I) just followed the template that specs2 uses, see https://github.com/etorreborre/specs2/blob/5d14719c4438ea01cf0c7cc7c4a52c1d8fd6c29b/common/shared/src/main/scala/org/specs2/execute/PendingUntilFixed.scala

I can't say I particularly like this style of programming, but I think it's not a bad practise to adopt the style of the lib we are integrating with for `contrib` code like this. 

But to address the confusion part, are you ok with renaming the inner class instead, because the trait is what will be mixed in the test classes, whereas the inner class is just internal.

djspiewak: Ah didn't realize you were following an existing pattern.  That's pretty horrible.  :-)  Agreed based on what you're describing that renaming the inner class is better.

alissapajer:
:+1: to the idea from me.

rintcius:
@jsantos17 @djspiewak I reworked the `PendingWithActualTracking` - it now loads the contents from a resource file instead. Also added `PendingWithActualTrackingWriteMode` that is designed to help when the actuals change (I think that should also address @jsantos17 concern regarding annoying breakage)

wemrysi:
👍 from me as well.

I'd like to explore not having to compare entire trees in tests and focus on verifying particular properties about them instead, e.g doesn't use map-reduce, all conditions end up as selectors, no map-reduce after sort, etc. Hopefully that will allow us to verify what is being tested while being more resilient to benign/unrelated changes.

rintcius:
@wemrysi @jsantos17 created #2829

PR #2835, opened 2017-09-26
Fix object concatenation (and by extension, IfUndefined) in Mimir

### LINKED ISSUE #2776: Mimir - IfUndefined behaving incorrectly
With this data:

``` json
[
{ "c": [], "a": 1, "b": { "bc": 0.5, "ba": { "innermost": true }, "bb": [ "x", false, { "foo": "bar" } ] } },
{ "a": 2, "b": { "bc": 60, "ba": { "innermost": false }, "bb": [ "y", true, { "foo": "baz" } ] }, "c": [ true, false, "maybe" ] },
{ "a": 3, "c": null, "b": { "ba": true, "bb": [ "z", { "foo": "quux" } ] } }
]
```

And a query like this:
``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "a"), {"test": OBJECT_CONCAT(t0.test, {"a": OBJECT_CONCAT(t0.test.a, t0.a) ?? t0.a}) ?? {"a": t0.a}}) FROM `/mimir/structure_test_deep` AS t0
```

However, if an `??` is included a the top level, the `a` remains:
``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "a"), {"test": OBJECT_CONCAT(t0.test, {"a": OBJECT_CONCAT(t0.test.a, t0.a) ?? t0.a}) ?? {"a": t0.a}}) ?? t0 FROM `/mimir/structure_test_deep` AS t0
```

djspiewak:
Fixes #2776 

Apparently object concat has just been borked.  For like, forever.  I vaguely remember finding this at Precog and being worried about it, but we never fixed it.  In fact, we literally wrote tests which assume the broken behavior!  No more!

In more prosaic news, this PR also adds integration tests for `IfUndefined` and `ObjectProject`.

### DIFF
--- a/blueeyes/src/main/scala/quasar/precog/common/CPath.scala
+++ b/blueeyes/src/main/scala/quasar/precog/common/CPath.scala
@@ -55,6 +55,9 @@ sealed trait CPath { self =>
   def \:(that: String): CPath = CPath(CPathField(that) +: self.nodes)
   def \:(that: Int): CPath    = CPath(CPathIndex(that) +: self.nodes)
 
+  def hasPrefixComponent(p: CPathNode): Boolean = nodes.startsWith(p :: Nil)
+  def hasSuffixComponent(p: CPathNode): Boolean = nodes.endsWith(p :: Nil)
+
   def hasPrefix(p: CPath): Boolean = nodes.startsWith(p.nodes)
   def hasSuffix(p: CPath): Boolean = nodes.endsWith(p.nodes)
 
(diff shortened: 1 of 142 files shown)


alissapajer:
@djspiewak there are what appear to be legitimate failures in a couple builds. Can you take a look?

djspiewak:
Yep, those failures are indeed legitimate.

PR #2842, opened 2017-09-27
Revert change to example h2 config

jedesah:


jedesah:
@garyb Sorry I didn't catch this earlier. We still support `file` for backwards compatibility but will use `location` moving forward since we now support any h2 url (apparently you can connect to remote h2 installations)

alissapajer:
I was under the impression that `location` doesn't currently work, while `file` does. Should `location` currently work?

garyb:
@alissapajer I've just re-tested with the latest Quasar and `location` is acceptable in the config now. I guess I must have caught it in between when the PR was merged for the change and when a release was made with it working!

Sorry @jedesah :smile:

garyb:
I also was able to use `mem:` for my H2 database location, which is what I wanted to do in the first place when I "discovered" this :)

jedesah:
@garyb No worries! 😄

Issue #2771, opened 2017-09-15
QScript MongoDB planner resorts to map-reduce when creating new object literals


wemrysi:
The MongoDB planner appears to resort to map-reduce any time a new object literal needs to be created as part of a plan, this is unnecessary and a regression from the legacy MongoDB planner.

To illustrate, consider the query
```sql
select { "foo" : pop } as bar from zips
```
Legacy Planner (MongoDB 3.2)
```
db.zips.aggregate(
  [{ "$project": { "bar": { "foo": "$pop" } } }],
  { "allowDiskUse": true });
```
QScript Planner (MongoDB 3.2)
```
db.zips.mapReduce(
  function () {
    emit.apply(
      null,
      (function (key, value) { return [key, { "bar": { "foo": value.pop } }] })(
        this._id,
        this))
  },
  function (key, values) { return values[0] },
  { "out": { "replace": "tmp.gen_0", "db": "testdata" } });
db.tmp.gen_0.aggregate(
  [{ "$project": { "bar": "$value.bar" } }],
  { "allowDiskUse": true });
```

PR #2850, opened 2017-09-28
Support mongo array literals in agg.pipeline for mongo 3.2 and higher

### LINKED ISSUE #2770: Simple aggregations have regressed performance-wise in new Mongo
Most aggregations have regressed performance-wise. It's useful to consider the equivalent queries:

```sql
select city from `/mngo/jps/zips` group by city
```

and 

```sql
select distinct city from `/mngo/jps/zips`
```

The QScript for the former:

```
QScript (Mongo-specific):
Reduce
├─ ShiftedRead(/jps/zips, ExcludeId)
├─ Bucket
│  ╰─ List
│     ╰─ MakeArray
│        ╰─ ProjectField
│           ├─ ○
│           ╰─ Constant(Str(city))
├─ Reducers
│  ╰─ Arbitrary
│     ╰─ ProjectField
│        ├─ ○
│        ╰─ Constant(Str(city))
╰─ Repair
   ╰─ ReduceIndex(\/-(0))
```

The QScript for the latter:

```
QScript (Mongo-specific):
Reduce
├─ ShiftedRead(/jps/zips, ExcludeId)
├─ Bucket
│  ╰─ List
│     ╰─ ProjectField
│        ├─ ○
│        ╰─ Constant(Str(city))
├─ Reducers
│  ╰─ Arbitrary
│     ╰─ ProjectField
│        ├─ ○
│        ╰─ Constant(Str(city))
╰─ Repair
   ╰─ ReduceIndex(\/-(0))
```

Performance for `distinct by` is at least an order of magnitude better. Note that the only difference between the plans is the lack of `MakeArray` for the second query/QScript.

Branched off #2667. The associated unit tests should be fixed as well when plans perform equivalently.

Quoting @wemrysi: "If there are any literal arrays (`MakeArray` and friends) in the plan, that will require MR until we update the mongo planner"

### LINKED ISSUE #2772: Leverage ability to create array literals in $project for MongoDB 3.2+
As of MongoDB 3.2, [array literals can be created](https://docs.mongodb.com/v3.2/reference/operator/aggregation/project/#project-new-array-fields) in `$project` stages of the aggregation pipeline.

Utilizing this capability in the MongoDB planner would alleviate a large class of expressions from having to resort to map-reduce, especially given how prevalent array literals currently are in QScript.

rintcius:
Fixes #2770 #2772


rintcius:
Note: many queries now use less map reduce but are still not on par with old mongo. I will go through complete list as part of #2771 (which should bring more queries on par)

mongodb/src/main/scala/quasar/physical/mongodb/accumulator/package.scala
```diff
@@ -31,15 +31,30 @@ package object accumulator {
...
+  private def wrapInLet(bson: Bson): Bson =
+    Bson.Doc("$let" -> Bson.Doc(
+      "vars" -> Bson.Doc("a" -> bson),
+      "in" -> Bson.Text("$$a")))
+
+  // We can't use arrays directly inside accumulators because of
+  // https://jira.mongodb.org/browse/SERVER-23839
+  // Ideally this wrapping would happen before we have translated to Bson
+  // but the trade-off is that in that case this wrapping needs to occur at all
+  // call-sites which would be prone to errors.
```
rintcius: Happy to hear if there's a single place where I can rewrite this instead.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -912,7 +914,7 @@ object MongoDbPlanner {
         def plan
           [M[_]: Monad: ExecTimeR: MonadFsErr, WF[_]: Functor: Coalesce: Crush: Crystallize, EX[_]: Traverse]
           (joinHandler: JoinHandler[WF, WBM],
-            funcHandler: MapFunc[T, ?] ~> OptionFree[EX, ?],
+            cfg: PlannerConfig[T, EX],
             v: BsonVersion)
```
rintcius: I'm leaning towards also adding `joinHandler` and `bsonVersion` to `PlannerConfig` in the PR for #2771. This makes the `Planner` interface more concise, but on the downside `PlannerConfig` can also be used in less places compared to with just the 2 fields (without having redundant fields).
Any opinions on this?

mongodb/src/main/scala/quasar/physical/mongodb/planner/StaticHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  object v3_2 {
+    def apply(): StaticHandler =
+      new StaticHandler {
+
+        def handle[T[_[_]]: BirecursiveT, M[_]: Monad, A]
+          (alg: AlgebraM[M, CoEnvMapA[T, A, ?], Fix[ExprOp]])
+          (fm: FreeMapA[T, A])
+            : Option[M[Fix[ExprOp]]] =
+          handleOps3_2[T, M, Expr3_2, A](alg)(fm)
```
wemrysi: This should fallback to the `v2_6` implementation, no?

mongodb/src/main/scala/quasar/physical/mongodb/planner/StaticHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+import slamdata.Predef._
+import quasar.physical.mongodb.expression._
+import quasar.qscript._
+
+import matryoshka._
+import matryoshka.data._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+trait StaticHandler {
```
wemrysi: Mind adding a comment briefly describing the intent of this type as well as a `TODO` to remove it once we make the changes to `MakeMap` and `MakeArray` to accept multiple arguments?

mongodb/src/main/scala/quasar/physical/mongodb/planner/StaticHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+      : Option[M[Fix[ExprOp]]] = {
+    val fp32 = new ExprOp3_2F.fixpoint[Fix[ExprOp], ExprOp](Fix(_))
+    fm.project.some collect {
+      case MapFuncCore.StaticArray(a) =>
+        val cs: M[List[Fix[ExprOp]]] = a.map(_.cataM(alg)).sequence
+        cs.map(fp32.$arrayLit)
+    }
+  }
+
+  object v2_6 {
```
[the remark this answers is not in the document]
rintcius: made them `def`s because of the generalization we discussed in slack

mongodb/src/main/scala/quasar/physical/mongodb/planner/StaticHandler.scala
rintcius: That was meant to be added in the #2771 PR, but changed it now as part of the `StaticHandler` refactoring.

mongodb/src/main/scala/quasar/physical/mongodb/planner/StaticHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+import quasar.physical.mongodb.expression._
+import quasar.qscript._
+
+import matryoshka._
+import matryoshka.data._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+trait StaticHandler {
+  def handle[T[_[_]]: BirecursiveT, M[_]: Monad, A]
```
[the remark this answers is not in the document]
rintcius: Great suggestion, this works out a lot nicer.

mongodb/src/main/scala/quasar/physical/mongodb/expression/transform.scala
```diff
@@ -0,0 +1,35 @@
...
+package quasar.physical.mongodb.expression
+
+import slamdata.Predef._
+
+import matryoshka._
+import matryoshka.implicits._
+import scalaz._
+
+object transform {
+  def wrapArrayInLet[T[_[_]]: CorecursiveT, EX[_]: Functor]
```
wemrysi: I'd like to suggest a single transform per file encoded via an `object` having a single public `apply` method implementing the transform. This would make it easier to find particular transforms and should also make incremental compilation more useful as we'll move away from huge files with lots of dependencies.

mongodb/src/main/scala/quasar/physical/mongodb/planner/PlannerConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import quasar.qscript.MapFunc
+import quasar.physical.mongodb.BsonVersion
+import quasar.physical.mongodb.workflow.{Coalesce, Crush, Crystallize}
+
+import matryoshka._
+import scalaz._
+
+final case class PlannerConfig[
+  T[_[_]]: BirecursiveT: ShowT,
```
wemrysi: It looks like all these constraints can removed as they aren't used anywhere.

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
```diff
@@ -585,6 +586,32 @@ package object workflow {
...
 
+
+  private def wrapArrayLit[EX[_]: Functor]
+    (accum: AccumOp[Fix[EX]])
+    (implicit ev: ExprOpCoreF :<: EX, ev32: ExprOp3_2F :<: EX)
+      : AccumOp[Fix[EX]] = {
+
+    def wrap(expr: Fix[EX]): Fix[EX] = (wrapArrayInLet[Fix, EX](expr.unFix)).embed
+
+    accum match {
```
wemrysi: `AccumOp` has a `Functor` instance, so you can just `accum map wrap`.

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
```diff
@@ -585,6 +586,32 @@ package object workflow {
...
+      case $push(value)     => $push(wrap(value))
+      case $sum(value)      => $sum(wrap(value))
+    }
+  }
+
+  private def wrapArrayLitExprInLet[EX[_]: Functor]
+    (grouped: Grouped[EX])
+    (implicit ev: ExprOpCoreF :<: EX, ev32: ExprOp3_2F :<: EX, ev2: ExprOpOps.Uni[ExprOp])
+      : Grouped[EX] =
+    Grouped(grouped.value.map(t => t._1 -> wrapArrayLit[EX](t._2)))
```
wemrysi: :golf: `grouped.value.mapValues(wrapArrayLit[EX])`

mongodb/src/main/scala/quasar/physical/mongodb/expression/transform.scala
rintcius: OK, let's try that

mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala
rintcius: rewrote it using `mapValues` but it's returning a `Map` so not as nice as `grouped.value.mapValues(wrapArrayLit[EX])`

wemrysi: Oh, I didn't realize that, unfortunate :disappointed:

Issue #2213, opened 2017-05-18
Provide Analyze for BackendModule


rabbitonweb:
With default implementation that will be available for all connectors to use

PR #2851, opened 2017-09-28
Backend module qanalysis

rabbitonweb:
This fixes https://github.com/quasar-analytics/quasar/issues/2213

In general we provide proper implementation for `  private final def analyzeInterpreter: Analyze ~> Configured` in `BackendModule`.

Connectors should provide two methods:
```
  def CardinalityQSM: Cardinality[QSM[Fix, ?]]
  def CostQSM: Cost[QSM[Fix, ?]]
```
For both `Cost` and `Cardinality` there are "generic", default implementations available (under `Cost._` and `Cardinality._`) that connectors can used - and this is what they actually do right now. But at some point they should provide their own implementation (especially for the `Cost` typeclass as it can take more connecotr specific information into consideration)

There is also a method called `pathCard` that provides default implementation to calculate `APath` cardinality (it basically runs `select count(*) from apath`) but connectors can (and should if they can) provide their own, more efficient version of this method.

Regarding the code, I tried to make this

```
  def CardinalityQSM: Cardinality[QSM[Fix, ?]] = Cardinality[QSM[Fix, ?]]
  def CostQSM: Cost[QSM[Fix, ?]] = Cost[QSM[Fix, ?]]
```

a bit more polymorphic (not fixing on the `Fix` - pun not intended), but I failed. Compiler was just not happy about it. But maybe since this method will only be used with `Fix` it is not such a big of a problem?

### DIFF
--- a/connector/src/main/scala/quasar/connector/BackendModule.scala
+++ b/connector/src/main/scala/quasar/connector/BackendModule.scala
@@ -18,6 +18,8 @@ package quasar
 package connector
 
 import slamdata.Predef._
+import quasar.Data
+import quasar.Data._int
 import quasar.common._
 import quasar.contrib.pathy._
 import quasar.contrib.matryoshka._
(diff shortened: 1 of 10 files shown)


rabbitonweb:
Please also note that `FunctorQSM` was changed to `TraverseQSM` since `zygoM` needs instance of traverse for `QSM`

alissapajer:
@rabbitonweb looking at travis, I think this isn't compiling.

rabbitonweb:
thx @alissapajer , fixing right now

djspiewak:
@rabbitonweb Still not compiling

rabbitonweb:
@alissapajer @wemrysi this is now ready for review

connector/src/main/scala/quasar/connector/BackendModule.scala
```diff
@@ -76,8 +84,39 @@ trait BackendModule {
         (runM compose runCfg compose runFs, close)
     }
 
-  private final def analyzeInterpreter: Analyze ~> Configured =
-    Empty.analyze[Configured]
+  def pathCard: APath => Backend[Int] = (apath: APath) => {
```
wemrysi: In order to keep `BackendModule` as abstract as possible, how about defining
```scala
trait AnalyzeModule {
  def queryCost(qs: Fix[QSM[Fix, ?]]): Backend[Int]
}
```
and implement `analyzeInterpreter` in terms of it. This particular implementation could then become a mixin that implements `AnalyzeModule` and connectors could include it as needed.

rabbitonweb: Sure, and we could then extract the "default" implementation as a separate thing, correct?

PR #2856, opened 2017-09-28
Experimental implementation of build stages

### LINKED ISSUE #2810: Rework the Travis configuration using stages
See [this blog post](https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions) for some details.  This has the potential to drop our build times appreciably, since we'll only be compiling *once*.  It also disentangles our publication steps from job specifics, making it easier for us to do things like lop off the Mongo 2.6 build.  The downside is that it will likely make our publication a bit more sensitive to unreliability in integration tests, since all prior stages have to be green for subsequent stages to run.

### LINKED ISSUE #2854: many ITs not running on Travis anymore
It looks like many ITs from connector specific sbt modules such as `mongoIt`, `marklogicIt` are not running in Travis anymore. 

I have checked these 
* not running `mongoit` tests: mongo 3.2 and 3.4
* running `mongoit` tests: mongo 2.6 and 3.0
Would be good to check the other connectors too

Note that I'm just assuming that the `mongoit` are running for 2.6 and 3.0 - the logs are not very clear on was has run exactly, see this excerpt for a 3.0 build:

```
[info] TaskResourceSpec
[info] CouchbaseStdLibSpec
[info] SparkStdLibSpec
[info] FilterPlannerSpec
[info] XmlStructuralPlannerSpec
[info] DirectoryQueriesSpec
[info] OperationsSpec
[info] FormatAwareFileSystemSpec
[info] JsonMarkLogicStdLibSpec
[info] XmlMarkLogicStdLibSpec
[info] JsonStructuralPlannerSpec
[info] MongoDbJsStdLibSpec
[info] MongoDbExprStdLibSpec
[info] MongoDbIOSpec
[info] MongoDbFileSystemSpec
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] No tests to run for rdbmsIt/exclusive:testOnly
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] Passed: Total 25, Failed 0, Errors 0, Passed 12, Skipped 13
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[success] Total time: 37 s, completed Sep 28, 2017 11:39:37 AM
```

djspiewak:
Fixes #2854
Fixes #2810

Reworks the build using Travis' new build stages feature.  Also moves a lot of the build tasks into `.travis.yml`, where they arguably should have been all along.  The build should now fail reliably when the github release step fails, and it should be more obvious *where* failures occur when they occur.  Additionally, the publication steps are now considerably less hacky, so we can pretty easily drop the mongo 2.6 and 3.0 builds if we would like; they are no longer privileged.

One thing I didn't really port from the old script was test coverage.  We need to reenable this eventually, but since we're doing PR target caching, there's almost no point to it right now.  Once things get polyrepo'd, we should reevaluate build times and see if we can reasonably disable target caching and reenable coverage.


djspiewak:
@nicflores I believe this is now ready for merge.  There may be issues in master, but the only way to see is to merge it.

Note that this change does disable automatically building non-master/backport branches on quasar.  We almost never use that feature anyway, so it's more or less ok, but it's still worth noting.  This disabling is because of a bug in the build stage conditional evaluator.

nicflores:
Thanks @djspiewak, I'll take a look.

.travis.yml
```diff
@@ -7,35 +7,113 @@ sudo: required
...
+
+    - stage: publish
+      env:
+      script:
+        - set -e
+        - './sbt ++$TRAVIS_SCALA_VERSION doc web/assembly'
+        - scripts/testJar
+
+        # release to sonatype
+        - scripts/quasarPublishAndTag
```
nicflores: I think we can replace this line with:
```
- './sbt transferPublishAndTagResources'
- scripts/publishAndTag 'quasar-analytics/quasar'
```
What do you think? If we can, then we can delete the rest of `scripts/quasarPublishAndTag`.

nicflores: Oh, but I don't know what that means for `scripts/constants`.

djspiewak: Yeah I was leery of making that change because of that.  I do want to push that script into `.travis.yml` though.  I would vote for doing that in a new PR.

nicflores: Sounds good. I'll merge shortly.

PR #2857, opened 2017-09-28
Fix bug when moving mount to a new path contained within it's current path

### LINKED ISSUE #2852: "succeed with module mount" is flaky
error: https://gist.github.com/djspiewak/65358a5c4296a1f1a43ad3f7f96dc26a

jedesah:
Fixes #2852


wemrysi:
@jedesah Oh, actually, would you mind adding a test of the bug this fixes.

jedesah:
@wemrysi There is a property test that will fail occasionally when this scenario happens. I guess I could copy the test and make it not a property with a set of hardcoded values that would trigger this bug. I wish scalacheck (or another such library) offered a way to specify values that must be tested for the property to be validation.

djspiewak:
@jedesah Can you copy the property test into an adjacent unit test with the failing generated values hard-coded?  I like to do that as a general principle when property tests have rare failures, just so it becomes a deterministic failure (at the cost of some redundancy).

PR #2860, opened 2017-09-29
Response body Process effect workaround transplant

drostron:


web/src/main/scala/quasar/api/services/metadata.scala
```diff
@@ -87,7 +89,7 @@ object metadata {
       InvalidMountNode.invalidMountNodeEncodeJson(_),
       FsNode.fsNodeEncodeJson(_)))
 
-  def service[S[_]](implicit Q: QueryFile.Ops[S], M: Mounting.Ops[S]): QHttpService[S] = {
+  def service[S[_]](implicit S0: Task :<: S, Q: QueryFile.Ops[S], M: Mounting.Ops[S]): QHttpService[S] = {
```
wemrysi: We could be a bit more precise here (and elsewhere) and instead require `Catchable[Free[S, ?]]` and leave it up to the caller to figure out how to make it so.

drostron: Dig it

drostron:
@wemrysi Ready

jedesah:
So what was the motivation for this change? As `QHttpService.apply` is used in some places to transform an existing `QHttpService`, I'm not convinced putting this logic in this constructor is ideal as we will end with this behavior repeated multiple times.

I think perhaps adding it to `QResponse.streaming` would be more appropriate.

Anyway, I'll be curious to hear the reason for the original change. Hopefully, we can figure this is out early Monday morning as it's sort of blocking the update to latest Quasar in SB

PR #2863, opened 2017-10-01
Create a top-level in memory interpreter

jedesah:
This interpreter can be re-used across any ServiceSpec and in other third party applications such as SlamData Backend


web/src/test/scala/quasar/api/services/Fixture.scala
```diff
@@ -93,4 +101,72 @@ object Fixture {
...
+      (vc andThen injectFT[Task, QErrs_Task])                                             :+:
+      (Timing.toTask andThen injectFT[Task, QErrs_Task])                                  :+:
+      injectFT[Module.Failure, QErrs_Task]                                                :+:
+      injectFT[PathMismatchFailure, QErrs_Task]                                           :+:
+      injectFT[MountingFailure, QErrs_Task]                                               :+:
+      injectFT[FileSystemFailure, QErrs_Task]
+    }
+  }
+
+  def inMemFS_(
```
drostron: Does the `_` suffix represent something here?

connector/src/main/scala/quasar/fs/InMemory.scala
```diff
@@ -261,6 +264,9 @@ object InMemory {
   def runFs(initial: InMemState): Task[FileSystem ~> Task] =
     runStatefully(initial).map(_ compose fileSystem)
 
+  def runFs0(initial: InMemState): Task[BackendEffect ~> Task] =
```
drostron: What do you think of a more evocative naming?

PR #2867, opened 2017-10-02
Refine mongo planner tests

jsantos17:
These are some changes I had sitting in my #2541 branch. They fix a few tests.


mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1344,6 +1353,7 @@ object MongoDbPlanner {
         case QC(Filter(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _))
            | QC(Sort(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _, _))
            | QC(Sort(Embed(QC(Filter(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _))), _ , _))
+           | QC(Subset(_, FreeShiftedRead(ShiftedRead(_, ExcludeId)), _ , _))
```
jsantos17: Should we do the same with `Union`?

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -837,21 +837,18 @@ class PlannerSpec extends
...
-                 Selector.Type(BsonType.Text)),
-               Selector.Doc(BsonField.Name("bar") ->
-                 Selector.Regex("^A.*$", false, true, false, false))),
-             Selector.And(
-               Selector.Doc(BsonField.Name("bar") ->
-                 Selector.Type(BsonType.Text)),
-               Selector.Doc(BsonField.Name("bar") ->
-                 Selector.Regex("^Z.*$", false, true, false, false)))))))
-    }.pendingWithActual(notOnPar, testFile("plan filter with LIKE and OR"))
+           Selector.And(
```
jsantos17: These conditions turned out to be the same. Old Mongo plans in DNF, while New Mongo uses CNF.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -946,36 +943,27 @@ class PlannerSpec extends
...
-                BsonField.Name("__tmp8") -> Selector.Eq(Bson.Bool(true)))),
-            Selector.And(
-              Selector.Doc(
-                BsonField.Name("__tmp10") -> Selector.Type(BsonType.Text)),
-              Selector.And(
-                Selector.Doc(
-                  BsonField.Name("__tmp11") -> Selector.Type(BsonType.Text)),
-                Selector.Doc(
-                  BsonField.Name("__tmp12") -> Selector.Eq(Bson.Bool(true))))))),
+          Selector.And(
```
jsantos17: This is similar to `plan filter with LIKE and OR`, but new Mongo is a little more restrictive. Here are the truth tables for both conditions:

Truth table for old Mongo: https://www.wolframalpha.com/input/?i=(A+AND+B)+OR+(A+AND+C+AND+D)+truth+table

Truth table for new Mongo: https://www.wolframalpha.com/input/?i=A+AND+C+AND+(B+OR+D)+truth+table

A, B, C, and D correspond sequentially to the conditions in the old Mongo plan.

jsantos17: Here are my notes I used to notice these conditions are mostly equivalent:

https://gist.github.com/jsantos17/7477c99f5157b29e626c1dbc9a8cc855

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1404,21 +1417,28 @@ object MongoDbPlanner {
...
+        .transCata[T[MQS]](liftFF[QScriptCore[T, ?], MQS, T[MQS]](
+          repeatedly(O.subsetBeforeMap[MQS, MQS](reflNT[MQS])))).point[M]
+      mongoQsOpt <- optimized.transCataM(liftFGM(assumeReadType[M, T, MQS](Type.AnyObject)))
+
+      prefPrj = PreferProjection.preferProjection[fs.MongoQScript[T, ?]](mongoQsOpt)
+      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript (Prefer Projection; Mongo-specific)", prefPrj))
+
+      mongoQs <- prefPrj.transCata[T[MQS]](R.normalize0[MQS]).point[M]
+      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript (Normalized; Mongo-specific)", mongoQs))
       // TODO: Once field deletion is implemented for 3.4, this could be selectively applied, if necessary.
```
rintcius: looks like this TODO needs to move up

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1381,6 +1393,7 @@ object MongoDbPlanner {
         case QC(Filter(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _))
            | QC(Sort(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _, _))
            | QC(Sort(Embed(QC(Filter(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _))), _ , _))
+           | QC(Subset(_, FreeShiftedRead(ShiftedRead(_, ExcludeId)), _ , _))
            | SR(Const(ShiftedRead(_, ExcludeId))) =>
```
rintcius: Not something you introduced but I'd be nice if we can get rid of these duplicated code blocks somehow

```
case QC(Filter(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _))
             | QC(Sort(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _, _))
             | QC(Sort(Embed(QC(Filter(Embed(SR(Const(ShiftedRead(_, ExcludeId)))), _))), _ , _))
             | QC(Subset(_, FreeShiftedRead(ShiftedRead(_, ExcludeId)), _ , _))
             | SR(Const(ShiftedRead(_, ExcludeId)))
```

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
jsantos17: :+1:

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
jsantos17: Can't we write a extractor for this? Probably not in this PR but I do agree reducing this duplication would be good!

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1404,21 +1417,26 @@ object MongoDbPlanner {
...
-          liftFF[QScriptCore[T, ?], fs.MongoQScript[T, ?], T[fs.MongoQScript[T, ?]]](
-            repeatedly(O.subsetBeforeMap[fs.MongoQScript[T, ?], fs.MongoQScript[T, ?]](reflNT[fs.MongoQScript[T, ?]]))))
-        .point[M]
-      mongoQs <- optimized.transCataM(liftFGM(assumeReadType[M, T, fs.MongoQScript[T, ?]](Type.AnyObject)))
-      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript (Mongo-specific)", mongoQs))
+        .transCata[T[MQS]](liftFF[QScriptCore[T, ?], MQS, T[MQS]](
+          repeatedly(O.subsetBeforeMap[MQS, MQS](reflNT[MQS])))).point[M]
+      mongoQsOpt <- optimized.transCataM(liftFGM(assumeReadType[M, T, MQS](Type.AnyObject)))
+
+      prefPrj = PreferProjection.preferProjection[fs.MongoQScript[T, ?]](mongoQsOpt)
```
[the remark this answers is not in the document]
jsantos17: Most definitely! Somehow didn't replace this one :D

connector/src/main/scala/quasar/qscript/Rewrite.scala
```diff
@@ -513,6 +513,25 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] extends TTypes[
...
+      F[T[G]] => G[T[G]] =
+    (repeatedly(Normalizable[F].normalizeF(_: F[T[G]])) _) ⋙
+      liftFF(repeatedly(compactQC(_: QScriptCore[T[G]]))) ⋙
+      liftFG(injectRepeatedly(compactLeftShift[F, G](prism).apply(_: QScriptCore[T[G]]))) ⋙
+      liftFF(repeatedly(applyTransforms(
+        uniqueBuckets(_: QScriptCore[T[G]]),
+        compactReductions(_: QScriptCore[T[G]])))) ⋙
+      repeatedly(C.coalesceQCNormalize[G](prism)) ⋙
+      liftFG(injectRepeatedly(C.coalesceEJNormalize[G](prism.get))) ⋙
+      (fa => QC.prj(fa).fold(prism.reverseGet(fa))(elideNopQC[F, G](prism.reverseGet)))
```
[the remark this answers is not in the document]
jsantos17: Yeah, the history is dirty since I introduced it and reverted it in this same PR...

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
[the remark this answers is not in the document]
jsantos17: They fixed one unit and allowed for easier debugging of Q#2541. So I thought we should keep it. I believe @wemrysi will refactor very soon too to apply the rest of normalizations before handing the QScript to Mongo.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
jsantos17: @rintcius the planned condition in new Mongo is slightly more restrictive than the one in old Mongo. Any idea if this might cause problems? The old Mongo plan allowed the lhs of `("foo" ~ pattern) OR (target ~ pattern)` to return true even if `target` is not a string.

cc @wemrysi

rintcius: Hmm, yeah not sure. I don't know how this should behave. For reference (tested on latest master), this query is in line with old mongo:

```
select val, true or val, val or true, false or val, val or false from `null/nulls`
Query time: 0.4s
{ "val": null, "1": true, "2": true, "3": null, "4": null }
{ "val": {  }, "1": true, "2": true, "3": {  }, "4": {  } }
{ "val": 1, "1": true, "2": true, "3": 1, "4": 1 }
```

Would be good to get the specs for this.

PR #2874, opened 2017-10-03
Support mongo object literals in agg.pipeline

### LINKED ISSUE #2771: QScript MongoDB planner resorts to map-reduce when creating new object literals

rintcius:
Fixes #2771


mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -479,6 +479,15 @@ class PlannerSpec extends
              ListMap())))
     }
 
+    "plan select map" in {
```
rintcius: the example from #2771 does not use `objectLit` but added the test anyway (& not hitting map-reduce anymore)

rintcius:
After the many planner test changes for `arrayLit`s, this 1 test change is a bit disappointing. That may change though after #2681 gets addressed.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
wemrysi: I think a minor revision to `select {"p": pop}, state from zips` should involve `objectLit`.

mongodb/src/main/scala/quasar/physical/mongodb/planner/StaticHandler.scala
```diff
@@ -42,8 +45,18 @@ object StaticHandler {
...
+
+      def toBsonFieldName(ej: T[EJson]): Option[BsonField.Name] =
+        (CommonEJson.prj(ej.project) >>= (str.getOption(_))).map(BsonField.Name(_))
+
       def handle[A](fm: FreeMapA[T, A]): Option[EX[FreeMapA[T, A]]] =
-        None
+        fm.project match {
+          case MapFuncCore.StaticMap(m) =>
+            val x: Option[List[(BsonField.Name, FreeMapA[T, A])]] =
+              m.map(t => toBsonFieldName(t._1) map ((_, t._2))).sequence
```
wemrysi: :golf: `m.traverse(t => ...)`

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
rintcius: 👍  added that test

Issue #2799, opened 2017-09-22
Set up continuous integration on windows


djspiewak:
We need to *at least* be doing this for Mimir, since people will use that locally on windows quite often.  This probably means using AppVeyor, which is basically Travis but for windows builds.

PR #2880, opened 2017-10-04
Working windows build

### LINKED ISSUE #2815: sbt-header has issues on Windows
Specifically, two things were observed:

- [ ] We are forced to override the JVM character encoding to be `UTF-8` using an envar, otherwise bad things happen
- [ ] For whatever reason, sbt-header seems to *always* touch every file every time it runs, even though it doesn't have to change headers.  This could be line-ending-related

### LINKED ISSUE #2819: Need an SBT script for Windows
This is related to getting AppVeyor set up (#2799), but it also affects people who just want to build quasar locally on Windows.  It's a pain to have to use the Lightbend SBT installer, and that pain is not one which is shared with Linux/Mac users.  This is tricky because sbt-extras has no analogous thing for Windows, but we might be able to get around it ourselves by using [coursier](https://get-coursier.io) to explicitly fetch the SBT launcher from Maven and bootstrap things from there.  Worth looking into, anyway.

djspiewak:
Fixes #2815 
Fixes #2819 

I can now checkout and build quasar on windows and the only dependency is the JDK.  The `sbt.bat` script will bootstrap and run SBT, similar to how the `sbt` script does.  We should be able to use this when we setup AppVeyor (to address #2799).

### DIFF
--- a/sbt.bat
+++ b/sbt.bat
@@ -1,0 +1,29 @@
+@REM https://github.com/xerial/sbt-pack/blob/master/src/main/templates/launch-bat.mustache
+@REM would be worth getting more inspiration from
+
+@echo off
+
+SET ERROR_CODE=0
+
+SET LAUNCHER_PATH=%localappdata%/coursier
+
+IF NOT EXIST "%LAUNCHER_PATH%" (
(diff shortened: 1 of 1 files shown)


nicflores:
@djspiewak, will you be pushing another change to kickstart another appveyor build? The page says the job was stopped, just wondering if it will restart.

djspiewak:
@nicflores No, AppVeyor wasn't intended to build this PR.

PR #2886, opened 2017-10-05
AppVeyor Configuration

### LINKED ISSUE #2799: Set up continuous integration on windows

### LINKED ISSUE #2899: RawHandlerSpecs fails on Windows
```
[error]   x recover from errors
[error]    the value is not equal to 'true' (RawHandlerSpecs.scala:177)
```

This failure is consistent, and appears to be legitimate.

### LINKED ISSUE #2900: ServiceSpec fails on Windows
```
[error] x mount filesystem (2 seconds, 234 ms)
[error]  '\/-(false)' is not \/- with value'true' (ServiceSpec.scala:116)
```

This error appears to be stable and reproducible.  I'm pretty sure it's an actual failure.

djspiewak:
Fixes #2799 
Fixes #2900
Fixes #2899 

Fixes a few issues with the Windows `sbt.bat` script and also fixes a couple random bugs that Windows quirks uncovered.  This PR primarily configures AppVeyor to compile, unit test and run the Mimir integration tests.  Caching is enabled along similar rules as Travis.  Both master and backport/ branches will be built, in addition to PRs.

Right now, it seems like fully-cached builds take around 28 minutes, while entirely uncached builds take just over an hour.  No build should be *entirely* uncached, since coursier stuff should be kept around, but it's worth keeping it in mind.  AppVeyor's build limit is 60 minutes, after which the job will be killed.  We're discussing options with them now.

Additionally, and this is the more constraining bit, AppVeyor has a concurrency limit of 1 for OSS projects.  We're looking into getting a paid account which should give us access to higher concurrency, but for the time being, there may be a bit of a queue building up if a lot of PRs are opened at once.

----

As a note regarding Windows stuff in general…  All of the random issues that I fixed to make the build work on Windows were *legitimate* issues.  Most of them were assumptions about the non-existence of backslashes in certain inputs, and one of which was actually a file handle leak that was only caught by file lock bugs.  In other words, if Travis blesses your build but AppVeyor doesn't, chances are very good you have a bug that you didn't realize was there!

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,31 @@
+version: 1.0.{build}
+
+pull_requests:
+  do_not_increment_build_number: true
+
+branches:
+  only:
+  - master
+  - /^backport\/v.*/
+
+skip_tags: true
+
+# TODO add ivy/coursier artifact cache
+cache:
+  - .targets
+  - '%localappdata%/coursier'
+  - ''
+
+environment:
+  JAVA_HOME: 'C:\Program Files\Java\jdk1.8.0'
+  PATH: '%JAVA_HOME%\bin;%PATH%'
+
+# TODO remove target cache here
+# before_build:
+  # - cmd:
+
+build_script:
+  - cmd: .\sbt ++"2.11.8" test:compile
+
+test_script:
+  - cmd: .\sbt ++"2.11.8" test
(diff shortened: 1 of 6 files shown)


djspiewak:
Currently blocked on #2899 and #2900

djspiewak:
@nicflores I believe this is now ready for review.  I've assigned it to you.

Issue #2226, opened 2017-05-18
Error in query


rabbitonweb:
works:
```
SELECT
  quarter AS Q,
  value AS Value,
  region AS Region
FROM `/loc/revenue.ldjson
```

does not work:

```
SELECT DISTINCT Value AS value, Q AS dimension, Region AS series, null AS parallel FROM `/loc/Untitled Workspace 9.slam/.tmp/out2bcf11b8-ccba-445b-9c96-a2beda9f7a4b` ORDER BY Region ASC, Q ASC
```

rabbitonweb:
Should be working once https://github.com/quasar-analytics/quasar/issues/1882 is fixed

zeeiyerWork:
#1882 is fixed in new-mongo. So this can be tested against the next-major branch or against master once next major merges to quasar master (before Aug31).

PR #2887, opened 2017-10-05
Add regresion test for bug #2226

rabbitonweb:
This fixes https://github.com/quasar-analytics/quasar/issues/2226

It ensures there will never be a regression for this bug


it/src/main/resources/tests/revenue.test
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "bug-2226",
+    "data": "revenue.data",
```
jedesah: Where is `revenue.data`?

rabbitonweb: https://github.com/quasar-analytics/quasar/pull/2887/files#diff-da639025bfa39d0ec7c821abc50fe778

jedesah: That link just redirects to this PR

rabbitonweb: That file was added in this PR. That link was suppose to point to it

jedesah: I don't see any `revenue.data` file

rabbitonweb: i'm stupid,  renamed the wrong file..

rabbitonweb: will fix that tomorrow

it/src/main/resources/tests/bug-2226.test
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "bug-2226",
+  "data": "revenue.data",
+  "backends": {
+    "couchbase":         "ignoreFieldOrder",
+    "marklogic_json":    "ignoreFieldOrder"
```
djspiewak: You should add `mimir` here as well, even though it is passing, since mimir does not guarantee field order in any meaningful way.

jedesah:
LGTM once you add `mimir` to the list as @djspiewak suggested

rabbitonweb:
@jedesah ready

PR #2890, opened 2017-10-05
Add Copy operation to Filesystem

### LINKED ISSUE #2393: Add copy/duplicate to filesystem
Backend support is needed for https://github.com/slamdata/slamdata/issues/1796.

jedesah:
Fixes #2393.

This PR itself does not actually add any useful user-facing functionality because none of the connectors support copying and nor does the views and modules interpreters just yet (There is an implementation for `InMemoryFS` thought so that the web layer can be tested! 😄 ). That said I wanted to open a PR now so that my work doesn't drift as I focus on other things as well as to allow the parallel work of any connector that wants to support copy (i.e. mimir as well as views and modules seem like the important candidates).

I do need to add a section to the README documenting the new HTTP verb before this gets merged though (mostly a reminder for myself) and I am sure there will be a few revisions on the implementation after you guys review it.

### DIFF
--- a/web/src/main/scala/quasar/api/ToApiError.scala
+++ b/web/src/main/scala/quasar/api/ToApiError.scala
@@ -97,7 +97,9 @@ sealed abstract class ToApiErrorInstances extends ToApiErrorInstances0 {
       case QScriptPlanningFailed(e) =>
         e.toApiError
       case UnsupportedOperation(reason) =>
-        apiError(BadRequest, "reason" := reason)
+        fromMsg(
+          BadRequest withReason "Unsupported Operation",
+          reason)
       case UnknownReadHandle(ReadHandle(path, id)) =>
         apiError(
           InternalServerError withReason "Unknown read handle.",
--- a/web/src/main/scala/quasar/api/ToApiError.scala
+++ b/web/src/main/scala/quasar/api/ToApiError.scala
@@ -96,6 +96,8 @@ sealed abstract class ToApiErrorInstances extends ToApiErrorInstances0 {
         e.toApiError :+ ("logicalplan" := lp.render)
       case QScriptPlanningFailed(e) =>
         e.toApiError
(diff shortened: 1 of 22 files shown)


djspiewak:
It all looks very good to me.  Do you think we should consider a default implementation which reads and re-writes the data?  I tend to think that such an implementation would be worse than useless, since there's no way it will ever be usefully performant, but I thought I would bring it up.

jedesah:
We certainly could and it might be rather useful for small amounts of data but also like you say terribly unhelpful in other cases... not sure.

djspiewak:
@wemrysi Thoughts?  :+1: from me.

web/src/main/scala/quasar/api/ToApiError.scala
```diff
@@ -96,6 +96,8 @@ sealed abstract class ToApiErrorInstances extends ToApiErrorInstances0 {
         e.toApiError :+ ("logicalplan" := lp.render)
       case QScriptPlanningFailed(e) =>
         e.toApiError
+      case UnsupportedOperation(reason) =>
+        apiError(BadRequest, "reason" := reason)
```
wemrysi: Mind using the `fromMsg_` constructor and update the reason on the `BadRequest` status to "Unsupported Operation" (or equivalent), for consistency with the rest of the errors.

wemrysi:
@jedesah Also, maybe this should be breaking since you changed existing response codes? The category of the codes didn't change, so maybe not, but worth considering.

jedesah:
@wemrysi I think you make a good point about being breaking.

web/src/main/scala/quasar/api/ToApiError.scala
jedesah: Will do

djspiewak:
@jedesah Test failures look legit.

PR #2892, opened 2017-10-06
Eliminate MongoDB value wrapping

### LINKED ISSUE #2762: `value` wrapping still occurs in the new Mongo connector
This is a bit of a blocker for the structure editor, since the frontend doesn't know when it will need to modify the projection paths to account for the `value`, and that would need special casing per connector too which has its own difficulties... but anyway, the specifics:

Using this example data set:
``` json
{ "a": 1, "b": { "ba": true }, "c": [] }
{ "a": 2, "b": { "ba": false }, "c": [true, false, "maybe"] }
{ "a": 3, "b": { "ba": true }, "c": null }
```

In the structure editor, choosing to "relabel key" `c` as `ccc` will result in a query like this:

``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "c"), {"ccc": OBJECT_CONCAT(t0.ccc, t0.c) ?? t0.c}) ?? t0 FROM `/mongo/test/structure_test_simple.json` AS t0
```

Which results in this:

``` json
{ "value": { "_id": { "$oid": "59ba92c22272dc2d3c07ba58" }, "a": 1.0, "b": { "ba": true } } },
{ "value": { "_id": { "$oid": "59ba92c22272dc2d3c07ba59" }, "a": 2.0, "b": { "ba": false } } },
{ "value": { "_id": { "$oid": "59ba92c22272dc2d3c07ba5a" }, "a": 3.0, "b": { "ba": true } } }
```

1. The results are now under a `value` key
2. An `_id` field is now appearing with each "row" under the `value`

`c` and `ccc` are also missing now, but that's probably because the delete succeeded, but re-inserting it failed due to the `value`-wrap.

Issue 1 is not new - old mongo behaved this way too. Issue 2 is kinda new - with old mongo, often a `__tmpN` key would appear under the results with copies of the old values in it, but that may be entirely unrelated to the `_id` appearence - it's just similar in the sense that "new data" is appearing after the query.

For reference, here's the expected result, as returned by `spark-local`:

``` json
{ "a": 1, "b": { "ba": true }, "ccc": [] },
{ "a": 2, "b": { "ba": false }, "ccc": [ true, false, "maybe" ] },
{ "a": 3, "b": { "ba": true }, "ccc": null }
```

### LINKED ISSUE #2805: Elide the wrapper object when streaming map-reduce results
MongoDB wraps the results of a map-reduce operation in objects that look like
```
{
  "_id" : <reduce key>,
  "value": <reduced value>
}
```
and we want to just emit `<reduced value>`, when streaming results out from Quasar.

Part of #2762.

### LINKED ISSUE #2826: Elide sigil during post-processing of aggregation pipeline results
Certain scenarios require nesting a result under _some_ field when yielding values from the aggregation pipeline. This could be because the version of MongoDB doesn't support `$replaceRoot` or because we can't determine statically that the value is a document (in which case, `$replaceRoot` wouldn't help).

In these scenarios, we want to nest the result under a well-known but rare key that we can look for in results and project away as a post-processing step.

The currently proposed sigil is: `__quasar_mongodb_wrapper`

Part of #2762.

wemrysi:
Eliminates the "value-wrapping" from MongoDB query results using the strategies described in #2762.

Additionally:
* Adds `EquiJoin` variants of normalization alongside the `ThetaJoin` versions. I didn't have time to explore how to unify these, but I think we should eventually.
* Adds support for multiple test directives per backend in regression test. This was useful during development, even though it didn't end up being necessary at the end.

Fixes #2762.
Fixes #2805.
Fixes #2826.

### DIFF
--- a/connector/src/main/scala/quasar/fs/QueryFile.scala
+++ b/connector/src/main/scala/quasar/fs/QueryFile.scala
@@ -95,7 +95,7 @@ object QueryFile {
     val rewrite = new Rewrite[T]
 
     val normUntilFixpoint: T[QS] => Option[T[QS]] = tqs => {
-      val next = tqs.transAna[T[QS]](rewrite.normalize)
+      val next = tqs.transAna[T[QS]](rewrite.normalizeTJ)
       (next =/= tqs) option next
     }
 
@@ -137,7 +137,7 @@ object QueryFile {
     val rewrite = new Rewrite[T]
 
     val qs =
-      convertAndNormalize[T, QScriptInternal[T, ?]](lp)(rewrite.normalize)
+      convertAndNormalize[T, QScriptInternal[T, ?]](lp)(rewrite.normalizeTJ)
         .leftMap(FileSystemError.planningFailed(lp.convertTo[Fix[LogicalPlan]], _)) ∘
         simplifyAndNormalize[T, QScriptInternal[T, ?], QS]
 
@@ -179,7 +179,7 @@ object QueryFile {
             Injectable.coproduct(Injectable.inject[Const[Read[ADir], ?], QScriptTotal[T, ?]],
               Injectable.inject[Const[Read[AFile], ?], QScriptTotal[T, ?]]))))
 
-    convertAndNormalize[T, QScriptInternal[T, ?]](lp)(rewrite.normalize)
+    convertAndNormalize[T, QScriptInternal[T, ?]](lp)(rewrite.normalizeTJ)
       .fold(
         perr => merr.raiseError(FileSystemError.planningFailed(lp.convertTo[Fix[LogicalPlan]], perr)),
         _.point[M])
--- a/connector/src/main/scala/quasar/fs/QueryFile.scala
+++ b/connector/src/main/scala/quasar/fs/QueryFile.scala
@@ -61,7 +61,7 @@ object QueryFile {
       FI: Injectable.Aux[QS, QScriptTotal[T, ?]],
       mergeable: Mergeable.Aux[T, QS],
       render: Delay[RenderTree, QS],
-      eq: Delay[Equal, QS],
+      eql: Delay[Equal, QS],
       show: Delay[Show, QS])
       : PlannerError \/ T[QS] = {
     val transform = new Transform[T, QS]
@@ -89,16 +89,18 @@ object QueryFile {
       QC: QScriptCore[T, ?] :<: QS,
       TJ:   ThetaJoin[T, ?] :<: QS,
       render: Delay[RenderTree, QS],
+      equal: Delay[Equal, QS],
       FI: Injectable.Aux[QS, QScriptTotal[T, ?]])
       : T[IQS] => T[QS] = {
     val rewrite = new Rewrite[T]
 
-    // TODO: This would be `transHylo` if there were such a thing.
-    _.transAna[T[QS]](SP.simplifyProjection)
-      // TODO: Rather than explicitly applying multiple times, we should apply
-      //       repeatedly until unchanged.
-      .transAna[T[QS]](rewrite.normalize)
-      .transAna[T[QS]](rewrite.normalize)
+    val normUntilFixpoint: T[QS] => Option[T[QS]] = tqs => {
+      val next = tqs.transAna[T[QS]](rewrite.normalize)
+      (next =/= tqs) option next
+    }
+
+    iqs => repeatedly(normUntilFixpoint)(
+      iqs.transAna[T[QS]](SP.simplifyProjection))
       .pruneArraysF
   }
 
@@ -126,6 +128,7 @@ object QueryFile {
       QC:  QScriptCore[T, ?] :<: QS,
       TJ:    ThetaJoin[T, ?] :<: QS,
       FI: Injectable.Aux[QS, QScriptTotal[T, ?]],
+      eql: Delay[Equal, QS],
       show: Delay[Show, QS],
       renderI: Delay[RenderTree, QScriptInternal[T, ?]],
       render: Delay[RenderTree, QS])
@@ -157,6 +160,7 @@ object QueryFile {
       CQ: Coalesce.Aux[T, QS, QS],
       PA: PruneArrays[QS],
       FI: Injectable.Aux[QS, QScriptTotal[T, ?]],
+      eql: Delay[Equal, QS],
       show: Delay[Show, QS],
       renderI: Delay[RenderTree, QScriptInternal[T, ?]],
       render: Delay[RenderTree, QS])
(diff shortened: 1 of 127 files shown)


alissapajer:
I'll take a look at this before Monday so we can get it merged ASAP.

rintcius:
`mongoIt` needs to be updated I think, looking at the build failures.

wemrysi:
Aye, I forgot to run the filesystem tests locally, will push fixes soon.

jdegoes:
A day of great rejoicing! 🎉 This bug has plagued us for 3 years.

alissapajer:
I gave this a once-over and overall looks good, but I didn't have a whole lot of time to give it. @wemrysi maybe Monday morning we can do a quick call and to talk through the main changes?

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -2329,10 +2355,10 @@ class PlannerSpec extends
             $read(collection("db", "zips")),
             $limit(5),
             $project(
-              reshape("value" -> $field("city")),
+              reshape(sigil.Quasar -> $field("city")),
               ExcludeId))
         }
-    }
+    }.pendingWithActual(notOnPar, testFile("plan simple single field selection and limit"))
```
rintcius: isn't this on par?

mongodb/src/test/resources/planner/plan_simple_sort_with_wildcard.txt
```diff
@@ -1,16 +1,20 @@
...
+│  │         { "$literal": undefined }]
+│  │     })
+│  ├─ Name("src" -> "$$ROOT")
+│  ╰─ ExcludeId
 ├─ $SortF
 │  ╰─ SortKey(0 -> Ascending)
 ╰─ $ProjectF
-   ├─ Name("value" -> { "$arrayElemAt": ["$src", { "$literal": NumberInt("0") }] })
+   ├─ Name("__quasar_mongodb_sigil" -> "$src")
    ╰─ ExcludeId
```
rintcius: onPar now?

it/src/test/scala/quasar/regression/Predicate.scala
```diff
@@ -155,6 +155,15 @@ object Predicate {
               failure
           }
     }
+
+    // Removes the element at `idx` from `as`.
+    private def deleteAt[A](idx: Int, as: Vector[A]): Vector[A] =
+      if (idx < 0)
+        as
+      else {
+        val (l, r) = as splitAt (idx + 1)
```
rintcius: `as.patch(idx, Nil, 1)` ?

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -89,16 +89,18 @@ object QueryFile {
...
       render: Delay[RenderTree, QS],
+      equal: Delay[Equal, QS],
       FI: Injectable.Aux[QS, QScriptTotal[T, ?]])
       : T[IQS] => T[QS] = {
     val rewrite = new Rewrite[T]
 
-    // TODO: This would be `transHylo` if there were such a thing.
-    _.transAna[T[QS]](SP.simplifyProjection)
-      // TODO: Rather than explicitly applying multiple times, we should apply
-      //       repeatedly until unchanged.
```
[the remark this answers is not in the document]
wemrysi: I think it would have worked before, we just weren't doing it.

it/src/main/resources/tests/typeOf.test
```diff
@@ -1,23 +1,32 @@
...
-        "spark_hdfs":     "pending",
-        "spark_local":    "pending",
-	"spark_cassandra": "pending"
+        "couchbase":         "pending",
+        "mimir":             "pendingIgnoreFieldOrder",
+        "mongodb_2_6":       "pending",
+        "mongodb_3_0":       "pending",
+        "mongodb_3_2":       "pending",
+        "mongodb_3_4":       "pending",
+        "mongodb_read_only": "pending",
```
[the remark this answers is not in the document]
wemrysi: It wasn't _really_ correct before. With the changes eliminating the auto-promotion of values from the regression tests, this started failing as it was producing the empty objects referenced in the `FIXME`, but the matcher wasn't recognizing them.

To avoid the empty objects I added the `IfUndefined` handling which exposed that mongo returns "number" for `type_of(42)` instead of the correct `integer`. So it is only partially correct. With these expectations, the only connectors that pass are `spark_local` and MarkLogic.

mongoIt/src/test/scala/quasar/physical/mongodb/MongoDbStdLibSpec.scala
```diff
@@ -98,17 +98,8 @@ abstract class MongoDbStdLibSpec extends StdLibSpec {
...
-        case Data.Obj(m) :: Nil =>
-          m.toList match {
-            case (_, v) :: Nil => v.some
-            case _             => None
-          }
-        case _ => None
-      },
-      check)
+    final case class SingleResultCheckedMatcher(check: ValueCheck[Data])
+        extends OptionLikeCheckedMatcher[List, Data, Data]("a single result", _.headOption, check)
```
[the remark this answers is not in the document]
wemrysi: Aye, you're right, I don't think it needs to be, I'll update to ensure there is only a single result.

mongodb/src/main/scala/quasar/physical/mongodb/sigil.scala
```diff
@@ -0,0 +1,103 @@
...
+import scalaz.syntax.std.option._
+
+/** MongoDB Document keys posessing special meaning. */
+object sigil {
+  import MapFuncsCore._
+
+  /** Quasar-specific sigil, used to construct a singleton result object under
+    * the following conditions:
+    *
+    * 1. Emitting a document value in aggregation when `$replaceRoot` is not available.
```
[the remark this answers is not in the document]
wemrysi: Or maybe we reword to "when `$replaceRoot` is not applicable."? I know we're not actually planning with `$replaceRoot` yet, but the concept still holds, I think.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
wemrysi: No, there is a typecheck that isn't being elided because `assumeReadType` doesn't traverse `Subset`, this should go back to passing once that is done. I had to reorder the `QScript` transformations applied in the MongoDB planner a bit, which resulted in some fallout. We'll need to spend some time to adjust how they compose and eventually run them to fixpoint, as-is the plans generated are very sensitive to the order in which they're applied.

mongodb/src/test/resources/planner/plan_simple_sort_with_wildcard.txt
wemrysi: Not quite yet, still has the typecheck (same as the other that is no longer on par).

wemrysi:
@rintcius Ready for another look.

PR #2905, opened 2017-10-09
ShiftedRead for PostgreSQL connector

kciesielski:
The first approach to QScript translation for RDBMS.
- `Repr` is `Fix[SqlExpr]` where `SqlExpr[T]` is a recursive structure, similar to `N1QL` in Couchbase connector.
- First implemented planner is `ShiftedRead`. There is also an example test in `PlannerSpec` which builds `Repr` from a SQL^2 query and compares it to expected structure. 
- There's a rendered for query representation: `SqlExprRenderTree`, used in mentioned tests.
- There's also a renderer that produces a String with final database query: trait `RenderQuery` with initial implementation `PostgresFlatRenderQuery`. The idea is to have different renderers selected by their BackendModules. Postgres renderer will come in two variants: one for "flat db schema", where data is stored in classic columns, another for JSON schema, where data is stored in a single JSON field.


kciesielski:
How to write a SQL^2 query to verify different `idStatus` semantics in ShiftedRead? (IdOnly, IncludeId, ExcludeId). I would like to add such cases to my tests.

kciesielski:
There are some compilation errors in the `MongoDB` module. I can't see how these may be related to my changes? 
https://travis-ci.org/quasar-analytics/quasar/jobs/286520256

wemrysi:
@kciesielski There are some issues with travis caching and/or sbt not detecting dependency changes when using the cache. I've cleared them out and restarted your build.

As for testing the planner, I would construct the input QScript by hand and make assertions about the output of the planner given QScript input, not SQL^2.

djspiewak:
@kciesielski Unless you're working on a WIP pull request or responding to review, you shouldn't be adding commits to a pre-existing PR.  Makes it very difficult to review.  Instead, cut a new local branch *from* your PR branch and work on that.  Since we don't rebase PRs, it's always safe to derive from a PR branch.

kciesielski:
@djspiewak Right, sorry for messing up any review that was ongoing. I'm now pushing to a separate branch, forked from last commit here.

kciesielski:
@wemrysi Thanks for the review. I'm closing this PR and I'll reopen a new one with changes only to the FileSystem part (which is using built-in OFFSET/LIMIT and using streaming/chunks for ReadFile).

PR #2908, opened 2017-10-10
Fix streaming of responses

jedesah:
Unbeknownst to us streaming in Quasar has been broken for quite a while. The implementation of `stepUntil` was to blame\.

Also reworks the "guarding" of streams to be more pinpointed and easier to work with

### DIFF
--- a/web/src/main/scala/quasar/api/QResponse.scala
+++ b/web/src/main/scala/quasar/api/QResponse.scala
@@ -100,9 +100,21 @@ object QResponse {
       Process.await(E.toEntity(a))(_.body).translate[Free[S, ?]](free.injectFT))
 
   /**
-    * This QResponse constructor guards against `Processes` that include "expected"
-    * errors as part of the head of the `Process` so that these get transformed into
-    * useful Http responses to the user of endpoints who make use of such streams.
+    * This QResponse constructor attempts to eliminate all effects in this
+    * stream that aren't actually part of the response stream by eagerly evaluating
+    * the "first few" effects of the stream. Examples of such effects are
+    * opening a quasar filesystem file, initiating a query to a connector, etc. Doing so
+    * allows us to provide a proper error response to clients instead of sending a
+    * Http OK and then failing to stream the response body because actually it
+    * turned out the file didn't exist after all.
+    *
+    * The better fix for this problem would probably be to avoid constructing
+    * such processes (at least for use in the web layer) and instead make a
+    * distinction between the effects that create a stream and the effects
+    * that "run" the stream.
+    *
+    * In the meantime, this constructor should mostly do the right thing if
+    * provided with such processes.
     */
   @SuppressWarnings(Array("org.wartremover.warts.Overloading"))
   def streaming[S[_], A]
--- a/web/src/main/scala/quasar/api/QResponse.scala
+++ b/web/src/main/scala/quasar/api/QResponse.scala
@@ -17,6 +17,7 @@
 package quasar.api
 
 import slamdata.Predef._
+import quasar.contrib.scalaz.catchable._
 import quasar.contrib.scalaz.eitherT._
 import quasar.effect.Failure
 import quasar.fp._
@@ -27,7 +28,8 @@ import org.http4s._
 import org.http4s.headers._
 import org.http4s.dsl._
 import scalaz.{Optional => _, _}
-import scalaz.syntax.monad._
+import scalaz.std.iterable._
+import scalaz.Scalaz._
 import scalaz.concurrent.Task
 import scalaz.stream.Process
 import scodec.bits.ByteVector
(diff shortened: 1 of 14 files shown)


web/src/test/scala/quasar/api/QHttpServiceSpec.scala
```diff
@@ -86,65 +82,62 @@ class QHttpServiceSpec extends quasar.Qspec {
...
+        QHttpService { case _ => QResponse.streaming(failStream) }
           .toHttpService(evalStr("one"))(request)
           .flatMap(_.orNotFound.as[String])
           .unsafePerformSync must_=== "FAIL"
       }
 
       "responds with alternate response when a small amount of data before first effect" >> {
-        val pad = Process.emit(ByteVector.low(max(0L, PROCESS_EFFECT_THRESHOLD_BYTES - 1)))
-        val padStream = failStream.copy(body = pad.append[StrIOM, ByteVector](failStream.body))
+        val pad = Process.emit("small amount")
```
wemrysi: It seems like testing with the actual threshold gives us better confidence about edge conditions, any reason not to continue doing so?

web/src/main/scala/quasar/api/QResponse.scala
```diff
@@ -97,21 +99,51 @@ object QResponse {
       E.headers,
       Process.await(E.toEntity(a))(_.body).translate[Free[S, ?]](free.injectFT))
 
+  /**
+    * This QResponse constructor guards against `Processes` that include "expected"
```
wemrysi: I'd say this is a bit more general that guarding against errors as it ensures (or attempts to) only the effects required to stream the response _body_ are present in the body process.

This is really masking over how we use `Process` in the filesystem as there are some effects required to initiate streaming of data (open a file, run a query, etc) and then there are the effects involved in streaming itself (fetching chunks) and we lose the distinction between the two.

jedesah: Just to be clear, you're suggesting I improve the doc right? I'm happy to attempt to improve it based on what you said. Just want to make sure you aren't bringing up some other issue with this function

web/src/test/scala/quasar/api/QHttpServiceSpec.scala
jedesah: I don't really like this test anyway (even though it served instrumental in figuring out that streaming was only working under specific circumstances) as it's not really a functional requirement, the functional requirement is that streaming still works in a streaming fashion despite the inspection of the head. I don't think it would be too hard to reformulate it as such. I'll give it a try.

web/src/main/scala/quasar/api/QResponse.scala
wemrysi: Yeah, just commenting on the docs 👍

web/src/test/scala/quasar/api/QHttpServiceSpec.scala
jedesah: Oops, when I said I didn't like this test, I meant the last one, not this one

jedesah: But even if I rewrite that test, the question of including the threshold in these tests is still relevant. My thoughts are that since it's a somewhat arbitrary amount and it doesn't really matter where it starts streaming at the 99th byte of the 101th byte, I would say it's not necessarily desirable to have the threshold in the test. It's also just somewhat convenient because of the change in type of `failedStream` since `pad` is now a stream of strings instead of bytes.

jedesah:
@wemrysi I think I improved the doc on that method, let me know what you think.

So turns out the new implementation of `stepUntil` wasn't quite right and was leading to an infinite loop if the stream was completely pure. Fixed that.

Added an additional test that attempts to validate that we aren't loading entire streams into memory.

web/src/test/scala/quasar/api/QHttpServiceSpec.scala
```diff
@@ -135,13 +135,26 @@ class QHttpServiceSpec extends quasar.Qspec {
...
+
+      "supports infinite streams without loading everything into memory" >> {
+        val continuous: Process[StrIOM, String] = Process.constant("filler")
+
+        val body = QHttpService { case _ => QResponse.streaming(continuous) }
+          .toHttpService(evalStr("two")).orNotFound(request)
+          .unsafePerformSync.body
+
+        // If we were no longer properly streaming and instead loading everything
+        // into memory, this line would never complete (I think)
```
wemrysi: Can you confirm this is the case? And, if so, how about adding a timeout (a minute or so) so that if it does fail, we know about it (rather than just having the tests hang)?

jedesah:
Good suggestion, will do!

On Oct 10, 2017 1:50 PM, "Emrys Ingersoll" <notifications@github.com> wrote:

*@wemrysi* commented on this pull request.
------------------------------

In web/src/test/scala/quasar/api/QHttpServiceSpec.scala
<https://github.com/quasar-analytics/quasar/pull/2908#discussion_r143833615>
:

>
         QHttpService { case _ =>
QResponse.streaming(pad.append[StrIOM, String](failStream)) }
           .toHttpService(evalStr("two"))(request)
           .flatMap(_.orNotFound.as[String])
           .unsafePerformSync must throwA[HttpResponseStreamFailureException]
       }
+
+      "supports infinite streams without loading everything into memory" >> {
+        val continuous: Process[StrIOM, String] = Process.constant("filler")
+
+        val body = QHttpService { case _ => QResponse.streaming(continuous) }
+          .toHttpService(evalStr("two")).orNotFound(request)
+          .unsafePerformSync.body
+
+        // If we were no longer properly streaming and instead
loading everything
+        // into memory, this line would never complete (I think)

Can you confirm this is the case? And, if so, how about adding a timeout (a
minute or so) so that if it does fail, we know about it (rather than just
having the tests hang)?

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/quasar-analytics/quasar/pull/2908#pullrequestreview-68404551>,
or mute the thread

jedesah:
@wemrysi Confirmed and adjusted. Thx for making me confirm it since I did indeed have a false assumption as it was the previous `Task` that was hanging which makes total sense

Issue #2923, opened 2017-10-12
MimirStdLibSpec failing for particular input


kciesielski:
During one of my builds when writing PostgreSQL connector I stumbled upon such failing test:
```
[error] x any doubles (745 ms)
[error]  Falsified after 29 passed tests.
[error]  > ARG_0: 2.5013896297685519E18
[error]  > ARG_1: -7.722061907008521E15
[error]  > List(Dec(2493667567861543379)) doesn't have size 0 but size 1 (MimirStdLibSpec.scala:195)
```
https://travis-ci.org/quasar-analytics/quasar/jobs/286989508

djspiewak:
I'm pretty sure this is a failure of `beCloseTo`.  Check it out:

```
@ 2.5013896297685519E18 - 7.722061907008521E15 
res1: Double = 2.4936675678615434E18
```

Mimir would have done this as `BigDecimal` addition due to the fact that it is receiving these values through `Data`.  Meaning that it would have gotten a more precise answer:

```
@ BigDecimal(2.5013896297685519E18) - BigDecimal(7.722061907008521E15) 
res2: BigDecimal = 2493667567861543379
```

The expectation is done in terms of `Double` though, meaning that the IEEE rounding takes effect.  So the answers are off by exactly **21**.

We can probably solve this problem by ensuring that all of the property tests within `StdLibSpec` use `BigDecimal` for their expectations.

alissapajer:
When these (or any) tests fail in the master build, we don't publish the current release. So in order to have stable continuous delivery, we need to fix flaky tests like this.

PR #2924, opened 2017-10-12
Include failtrace for specs2 console output in travis.

alissapajer:
Without `failtrace` enabled, the console sometimes prints too little information about a failed test, since only the test name is printed along with a single line number reference.

Usually this is enough information, but in the case of `StdLibSpec`, each connector delegates to that specification, and the line number in the test failure references the individual connector spec, not `StdLibSpec`, which is not enough information to know which test failed if there are multiple tests with the same name. (For example, we have a case for "any doubles" for many different mathematical operations.)

See https://github.com/quasar-analytics/quasar/issues/2923 for an example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,9 @@ script:
     SPECIFIC_DELEGATE=
 
     case $CONNECTOR in
-      couchbase) SPECIFIC_DELEGATE="couchbaseIt/testOnly -- xonly" ;;
-      marklogic_*) SPECIFIC_DELEGATE="marklogicIt/testOnly -- xonly" ;;
-      mongodb_*) SPECIFIC_DELEGATE="mongoIt/testOnly -- xonly" ;;
+      couchbase) SPECIFIC_DELEGATE="couchbaseIt/testOnly -- xonly failtrace" ;;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -55,9 +55,9 @@ script:
     SPECIFIC_DELEGATE=
 
     case $CONNECTOR in
-      couchbase) SPECIFIC_DELEGATE="couchbaseIt/testOnly -- xonly" ;;
-      marklogic_*) SPECIFIC_DELEGATE="marklogicIt/testOnly -- xonly" ;;
-      mongodb_*) SPECIFIC_DELEGATE="mongoIt/testOnly -- xonly" ;;
+      couchbase) SPECIFIC_DELEGATE="couchbaseIt/testOnly -- xonly failtrace" ;;
```
alissapajer: We also have the option of including `fullstacktrace`.

I checked locally and `failtrace` prints enough to trace through `StdLibSpec`, but maybe the entire trace would just be better?

nicflores: @alissapajer, maybe we roll with `failtrace` and if folks think we should bump the output we can move to `fullstacktrace`.

nicflores:
Looks good @alissapajer. I'll merge once travis builds.

PR #2928, opened 2017-10-13
Refactor of top-level interpreter

jedesah:
This achieves better re-use between production and testing interpreters as well as being cleaner and more modular across the board

This will allow third party clients to more easily re-use parts of Quasar


jedesah:
Taking suggestions on a better name for `FSThing`...

jedesah:
Looks like @djspiewak might be offline this morning. Switching to @wemrysi as this is pretty high priority

djspiewak:
@wemrysi is a much better reviewer for this in any case.  :-)

core/src/main/scala/quasar/fs/mount/package.scala
```diff
@@ -100,4 +100,43 @@ package object mount {
     ): ViewFileSystem ~> F =
       mounting :+: mismatchFailure :+: mountingFailure :+: viewState :+: vcache :+: monotonicSeq :+: fileSystem
   }
+
+  def overlayModulesViews[S[_], T[_]](
+    f: BackendEffect ~> Free[T, ?]
```
wemrysi: `f` and `T[_]` don't seem to be necessary, could just require `BackendEffect :<: S` instead.

interface/src/main/scala/quasar/main/FSThing.scala
```diff
@@ -0,0 +1,29 @@
...
+package quasar.main
+
+import slamdata.Predef.Unit
+import quasar.fs.BackendEffect
+import quasar.fs.mount.Mounting
+
+import scalaz.~>
+import scalaz.concurrent.Task
+
+final case class FSThing(
```
wemrysi: How about `CompositeFS`?

interface/src/main/scala/quasar/main/Quasar.scala
```diff
@@ -47,60 +37,14 @@ import scalaz.concurrent.Task
   */
 final case class Quasar(interp: CoreEff ~> QErrs_TaskM, shutdown: Task[Unit]) {
   val taskInter: CoreEff ~> Task =
-    foldMapNT(NaturalTransformation.refl[Task] :+: QErrs.toCatchable[Task]) compose interp
+    Quasar.toTask compose interp
+
+  def extendShutdown(step: Task[Unit]): Quasar =
+    copy(shutdown = shutdown >> step)
```
wemrysi: I think we'll want to `attemptNonFatal.void` the existing `shutdown` so a failure there doesn't prevent `step` from executing.

web/src/test/scala/quasar/api/services/Fixture.scala
```diff
@@ -102,71 +100,29 @@ object Fixture {
...
-        injectFT[MountingFailure, QErrs_Task]      :+:
-        injectFT[PathMismatchFailure, QErrs_Task]
-      toV andThen foldMapNT(vToTask)
-    }
-  }
-
   def inMemFS(
     state: InMemState = InMemState.empty,
     mounts: MountingsConfig = MountingsConfig.empty,
+    metaRef: TaskRef[MetaStore] = MetaStoreFixture.createNewTestMetastore.flatMap(TaskRef(_)).unsafePerformSync,
```
wemrysi: I'd like to request that we not perform side effects in default arguments. Better to just create another function that supplies the default values.

web/src/test/scala/quasar/api/services/Fixture.scala
```diff
@@ -102,71 +100,29 @@ object Fixture {
...
+                  mount andThen injectFT[Task, QErrs_Task],
+                  noShutdown)
+      eval    <- CoreEff.defaultImpl(fsThing, metaRef, persist)
+    } yield injectFT[Task, QErrs_Task] :+: eval
   }
 
   def inMemFSWeb(
     state: InMemState = InMemState.empty,
     mounts: MountingsConfig = MountingsConfig.empty,
+    metaRef: TaskRef[MetaStore] = MetaStoreFixture.createNewTestMetastore.flatMap(TaskRef(_)).unsafePerformSync,
```
wemrysi: Same.

web/src/test/scala/quasar/api/services/Fixture.scala
djspiewak: > I'd like to request that we not perform side effects in default arguments

😱

interface/src/main/scala/quasar/main/Quasar.scala
jedesah: That's not unreasonable

interface/src/main/scala/quasar/main/FSThing.scala
jedesah: `CompositeFS` seems is a little too similar to `CompositeFilesystem`. Doesn't convey that it can be either `CompositeFilesytem`, `InMemory` or anything else that satisfies that general shape. Really, I wanted to call it `FileSystem` but that seems too much of an overloaded term already

core/src/main/scala/quasar/fs/mount/package.scala
jedesah: But then it wouldn't be as convenient to use or I am missing something?

wemrysi: Not sure I understand how one is more convenient than the other, if you don't want to interpret `BackendEffect` when adding modules and views, then you'd have to unnecessarily supply an interpreter lifting into `Free` here, which doesn't seem convenient.

Anyhow, this isn't something worth holding up the merge for.

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
```diff
@@ -45,6 +45,10 @@ package object scalaz {
       OptionT(self.run.map(opt => opt: Option[B]))
   }
 
+  implicit final class MoreBindOps[F[_], A](val self: F[A])(implicit val F: Bind[F]) extends _root_.scalaz.syntax.Ops[F[A]] {
```
wemrysi: Mind moving this to `quasar.contrib.scalaz.bind`, similar to the additions for `foldable`, `zipper`, etc. Makes imports more informative and helps to avoid large package objects.

interface/src/main/scala/quasar/main/FSThing.scala
wemrysi: I don't think that's a problem, but if you're not a fan, please suggest an alternative. We need something better than `FSThing`.

core/src/main/scala/quasar/fs/mount/package.scala
jedesah: I guess I'm just not 100% sure what the call site looks like. Let's fix it in a subsequent PR. I'm certainly open to a more generic signature, this is just what was already there (granted I'm the one who added this function with @dave not that long ago)

interface/src/main/scala/quasar/main/FSThing.scala
jedesah: Haha, of course, I purposefully left it like this to spur a discussion. How about `FS` (at least for now)? We can also change this later

nicflores:
I'm gonna go ahead and merge this since we need this for the release.

PR #2931, opened 2017-10-13
Removed some dead code in the Quirrel evaluator

djspiewak:
The evaluator itself still needs to be removed.  The only reason it's still present (along with DAG, bytecode, and some other things) is because it is what powers the test harnesses for nearly all of mimir's function implementations.  One thing at a time though.


alissapajer:
When the diff looks like this
diff
and travis passes, I pretty much don't have to review it :stuck_out_tongue:

PR #2934, opened 2017-10-14
Revert "Bump sbt-slamdata to 0.5.5"

drostron:
This reverts commit 8f9f4fc19c777c36e39310f4117e16dbe520176a.

sbt-slamdata's handling of `publishTo` is applied to root as desired yet not other projects. This leads to faulty publishing. Reinstate once addressed.


djspiewak:
Oh, that makes a ton of sense!  Good call on the subprojects.  Will merge as soon as green.

PR #2936, opened 2017-10-16
Upgrade `http4s` to `0.16.5a` (latest on `0.16.x` series)

jedesah:
There were a few important fixes released in these past few releases that we probably want to have in the release.


jedesah:
@nicflores When you have a chance can merge this since we'll have to wait on a release cycle anyway, might as well get this into the release

Issue #2228, opened 2017-05-19
Multitable Joins in Spark returns no data


dugdun:
When building a workspace in Spark, it was noticed that a four table join query returned no results, although the same query in MongoDB with similar data works just fine.

To use the following query, you need to first open the `/loc/customers.data` file and then add a dropdown from the **Setup Form** card and display it. Mirror this deck and add a **Query** card with the following SQL Query.

```sql
SELECT
  o.order_key AS OrderKey,
  o.purchase_date AS PurchaseDate,
  c.last_name || ", " || c.first_name AS Name,
  c.city AS City,
  c.state AS State,
  i.clothing_size AS Size,
  i.clothing_type AS Type,
  i.clothing_color AS Color,
  i.price AS UnitPrice,
  oi.qty AS Quantity,
  i.price * oi.qty AS PricePaid,
  SUM(i.price * oi.qty) AS TotalPaid,
  i.clothing_brand AS Brand
FROM
  :results AS c
JOIN `/loc/orders.data` AS o
ON c.customer_key = o.customer_key
JOIN `/loc/ordered_items.data` AS oi
ON oi.order_key = o.order_key
JOIN `/loc/inventory_items.data` AS i
ON i.item_key = oi.item_key
GROUP BY
  i.price * oi.qty
```

If you limit this query to join on just the `:results` of `/loc/customer.data` and `/loc/orders.data` you will get the expected results.

zeeiyerWork:
A 2-way join between Orders and Customers did return data - so hopefully this turns out to be low effort. Hence scheduled for May, since it would be good to be able to demo this functionality at Spark Summit 2017.

alissapajer:
This may be a qscript issue. There is a know qscript bug (@wemrysi is working on it now for #1929) where we fail to align join provenance with other related provenance. Given how we use the results of joins in this query, we may be hitting that bug.

rabbitonweb:
This issue is solved on backend side. There are separate issues on the frontend that effect this functionality (and some other demo workspaces), but issue for it is already opened on slamdata repo https://github.com/slamdata/slamdata/issues/2142

PR #2937, opened 2017-10-16
Add regression tests for bug #2228

rabbitonweb:
Fixes https://github.com/quasar-analytics/quasar/issues/2228


it/testing.conf.example
```diff
@@ -4,7 +4,7 @@
 #mongodb_3_2       ="mongodb://<host>:<port>"
 #mongodb_read_only ="mongodb://<host>:<port>"
 #spark_local       ="<path_to_folder>"
-#spark_hdfs        ="spark://<spark_host>:<spark_port>?hdfsUrl=hdfs://<hdfs_host>:<hdfs_port>&rootPath=<root_path>"
+# spark_hdfs        = "spark://hideout:7077?hdfsUrl=hdfs://serv1.demo.slamdata.com:9000&rootPath=/"
```
djspiewak: Spurious whitespace

it/src/main/resources/tests/bug2228.test
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "bug2228",
+  "data": ["customers.data", "orders.data", "ordered_items.data", "inventory_items.data"],
+  "backends": {
+    "couchbase":         "ignoreFieldOrder",
+    "marklogic_json":    "ignoreFieldOrder",
+    "mimir":             "pending"
```
djspiewak: Can you add this to #2542 with a quick note as to what happens?

djspiewak:
Needs version label.  Also a `skip` for Couchbase.

it/testing.conf.example
alissapajer: Do we want to be checking this line in, showing demo server access as an example?

it/src/main/resources/tests/bug2228.test
alissapajer: `mimir` should be `pendingIgnoreFieldOrder`.

it/testing.conf.example
djspiewak: Uh… good call. @rabbitonweb you need to amend this commit asap to remove this data.  Don't add a commit; `amend` and `push -f`.

djspiewak:
@rabbitonweb @damonLL We've scrubbed the HDFS cluster information from the commits.  However, the damage is already done.  That cluster needs to be authenticated (it should have been authenticated all along, though that wouldn't have saved us here since the credentials would have been committed).  Any data in that cluster should now be considered public, and the cluster itself can be safely considered open to the public until it is secured.

rabbitonweb:
@djspiewak I don't quite follow your comment. Do you mean the data used for tests is private? Or do you mean we should have secure HDFS?

rabbitonweb:
version label added, ready for merge

djspiewak:
@rabbitonweb I mean we shouldn't be exposing services to the open internet without security.  Whether or not the data contained within HDFS is private, HDFS *itself* is not intended to be exposed in that way.  The service is designed to run behind a firewall, exposed only to application servers (and in some environments, only *certain* application servers).  The only "security" we have on our HDFS cluster is the fact that no one knew we had it.  That has now been stripped away.

The right thing to do is firewall off HDFS from everything except the demo server.  Remote access (for dev or demo purposes) can be conveniently and securely achieved via SSH tunneling.

alissapajer:
@rabbitonweb there are conflicts to resolve.

alissapajer:
Huh, this now appears to be merged...

PR #2941, opened 2017-10-16
Fixed spark jar publication (both sparkless and plugin jars)

djspiewak:
The `sparkcore.jar` publication is a bit of a hack.  We should do better in future when we have more time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ env:
     - CONNECTOR=spark_local_test
     - CONNECTOR=spark_hdfs
   global:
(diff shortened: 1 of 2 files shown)


nicflores:
@djspiewak, should the `sparkcore.jar` file be generated in the spark_local job? I don't see it being generated.
https://travis-ci.org/quasar-analytics/quasar/jobs/288801072#L569
Maybe I'm missing something.

nicflores:
sounds like it doesn't have to be generated in spark local job. I'll merge.

rabbitonweb:
sparkcore.jar should be generated in different folder, will cover this in other PR

PR #2943, opened 2017-10-17
views and modules can be copied

### LINKED ISSUE #2942: Support `COPY` in views and modules interpreters
This will allow views and modules to be moved in the filesystem.

I'm assuming higher priority on this than #2741 as it will solve most of the same use cases and seems to me at this point like a lower hanging fruit.

jedesah:
Resolves #2942


djspiewak:
Reassigning to @drostron since his review was requested.  👍 from me for merge.

PR #2944, opened 2017-10-17
Make use of `InMemFS` instead of custom interpreters in `DataServiceSpec`

jedesah:
View cache requires some extra special interpreter so leaving that there for now (although I have a hunch those tests could be re-written in another way to avoid the custom interpreter)

### DIFF
--- a/connector/src/main/scala/quasar/fs/InMemory.scala
+++ b/connector/src/main/scala/quasar/fs/InMemory.scala
@@ -270,6 +270,9 @@ object InMemory {
   def runBackend(initial: InMemState): Task[BackendEffect ~> Task] =
     runStatefully(initial).map(_ compose fileSystem0)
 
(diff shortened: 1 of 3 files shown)


jedesah:
Just realized this one depends on #2942.  Not to compile but so that some tests don't fail in `DataServiceSpec`.

jedesah:
Which only goes to prove why this change is more important than it might seem at first because it revealed the fact that until copy for views and modules is implemented, no other connector has the chance to implement it themselves without the views and modules interpreters jumping in and saying that they don't support that before the request can even make it to any connector. So @djspiewak keep in mind that #2942 needs to be merged before Mimir copy will ever work at the REST API level.

drostron:
@jedesah should this be `status: in progress` until #2942 is in place?

jedesah:
Superseded by #2943

PR #2946, opened 2017-10-17
Generalize branches as traversal

rintcius:
* Generalize `Branches` as `Traversal`
* Write `mapBeforeSort`  in terms of it
* Add `normalize` across branches

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
@@ -1311,20 +1311,17 @@ object MongoDbPlanner {
       _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo", mongoQS3))
 
       // NB: Normalizing after these appears to revert the effects of `mapBeforeSort`.
-      mongoQS4 =  mongoQS3.transCata[T[MQS]](liftId[T, MQS](mapBeforeSort[T].trans(idPrism[MQS])))
-      mongoQS5 =  BR.branches.modify(
-        _.transCata[FreeQS[T]](liftCoEnv[T, QST](mapBeforeSort[T].trans(coenvPrism[QST, Hole])))
-      )(mongoQS4.project).embed
-      mongoQS6 =  mongoQS5.transCata[T[MQS]](
+      mongoQS4 =  Trans(mapBeforeSort[T], mongoQS3)
+      mongoQS5 =  mongoQS4.transCata[T[MQS]](
                     liftFF[QScriptCore[T, ?], MQS, T[MQS]](
                       repeatedly(O.subsetBeforeMap[MQS, MQS](
                         reflNT[MQS]))))
-      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo (Shuffle Maps)", mongoQS6))
+      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo (Shuffle Maps)", mongoQS5))
 
       // TODO: Once field deletion is implemented for 3.4, this could be selectively applied, if necessary.
-      mongoQS7 =  PreferProjection.preferProjection[MQS](mongoQS6)
-      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo (Prefer Projection)", mongoQS7))
-    } yield mongoQS7
+      mongoQS6 =  PreferProjection.preferProjection[MQS](mongoQS5)
+      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo (Prefer Projection)", mongoQS6))
+    } yield mongoQS6
   }
 
   def plan0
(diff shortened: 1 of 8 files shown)


mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1431,23 +1302,29 @@ object MongoDbPlanner {
...
-      mongoQS3 =  applyTrans(mongoQS2)(mapBeforeSort[T])
-      mongoQS4 =  mongoQS3.transCata[T[fs.MongoQScript[T, ?]]](
-                    liftFF[QScriptCore[T, ?], fs.MongoQScript[T, ?], T[fs.MongoQScript[T, ?]]](
-                      repeatedly(O.subsetBeforeMap[fs.MongoQScript[T, ?], fs.MongoQScript[T, ?]](
-                        reflNT[fs.MongoQScript[T, ?]]))))
-      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo (Shuffle Maps)", mongoQS4))
+      mongoQS4 =  mongoQS3.transCata[T[MQS]](liftId[T, MQS](mapBeforeSort[T].trans(idPrism[MQS])))
+      mongoQS5 =  BR.branches.modify(
+        _.transCata[FreeQS[T]](liftCoEnv[T, QST](mapBeforeSort[T].trans(coenvPrism[QST, Hole])))
+      )(mongoQS4.project).embed
```
alissapajer: The pattern here is very similar to `applyTrans`. Perhaps we can keep `applyTrans` around and modify it if necessary? There is a lot of boilerplate here with the `.project`, `.embed`, and prism, and so it would be nice to have a function that incorporates all that. The overhead to add a new qscript rewrite should be low, so any way we can contain the boilerplate helps with that.

It looks like `mongoQS3` above shares a similar pattern, so maybe that one can be rewritten to call the same function as this one.

rintcius: There is similarity indeed, but also some differences. There are a couple of reasons why I stopped using `applyTrans` (for now):

* I was thinking it makes sense to partition all transformations based on whether it's across branches or not (as opposed to per "conceptual rewrite", which is the strategy that follows from using `applyTrans`). That's not implemented yet, but now is probably a good time to decide on the strategy.
* The signatures of `mapBeforeSort` and `normalize` are different (the ones across branches in particular). E.g. one relies on `liftCo`, the other `liftCoEnv`. I _think_ we can do with just one of these methods and I am leaning towards `liftCo` variant (which is not the one used in `applyTrans`), but good to discuss so we can standardize on the signature of the transformation.
* I also had trouble generalizing `applyTrans` because of the extra constraints that `normalize` uses (that's how I bumped into that comment that `Branches` could be generalized using `Traversal`s), but that can totally be my limitation. But anyhow, it will depend a bit on the decisions of the former points how `applyTrans` will look like exactly. Maybe good to reintroduce it after all mongo transformations have been implemented to work across branches? I'm also open to reintroduce it now of course, but then I'd need some guidance from @alissapajer or @wemrysi

rintcius: That looks great @wemrysi - I have started to integrate that in this PR.
1 compile error left in `Trans` which I don't know how to resolve in a proper way:
`implicit val ev: Branches[T, QScriptTotal[T, ?]] = ???`

wemrysi: Is it due to the `RenderTreeT` constraint on the `equiJoin` `Branches` instance? Can that constraint be removed?

alissapajer: The `RenderTreeT` constraints are generally there for debugging, and I'd prefer to keep them if possible, because otherwise it is a lot of work adding all the constraints back in each time we need to print a tree.

wemrysi: @alissapajer I'm sympathetic to this where we've introduced it elsewhere, however in this case it appears to _only_ be on the `EquiJoin` instance and none of the others and may be causing this particular instance to not be found during implicit search because of it.

alissapajer: Gotcha, if removing it from this instance helps than :+1: And if we end up needing it for this typeclass (as a whole) we can address that at a separate time.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1431,23 +1302,26 @@ object MongoDbPlanner {
...
-      mongoQS1 <- mongoQS0.transCataM(elideQuasarSigil[T, fs.MongoQScript[T, ?], M](anyDoc))
-      mongoQS2 =  mongoQS1.transCata[T[fs.MongoQScript[T, ?]]](R.normalizeEJ[fs.MongoQScript[T, ?]])
-      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo", mongoQS2))
+      mongoQS0 <- qs.transCataM(liftFGM(assumeReadType[M, T, MQS](Type.AnyObject)))
+      mongoQS1 <- mongoQS0.transCataM(elideQuasarSigil[T, MQS, M](anyDoc))
+      mongoQS2 =  mongoQS1.transCata[T[MQS]](R.normalizeEJ[MQS])
+      mongoQS3 =  BR.branches.modify(
+        _.transCata[FreeQS[T]](liftCo(R.normalizeEJCoEnv[QScriptTotal[T, ?]]))
+        )(mongoQS2.project).embed
+      _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo", mongoQS3))
```
rintcius: I have left `normalizeEJ` like this for now. It's a slight improvement compared to the old impl, but not in terms of `Trans` yet.
Locally I have an implementation in terms of `Trans` - but it doesn't include `coalesce`, because I couldn't fit that in the new scheme. I'd like to address this separately, if that's alright.

rintcius:
@alissapajer @wemrysi ready for another look

PR #2948, opened 2017-10-18
Add VCache headers to `/query` and `/invoke`

### LINKED ISSUE #2661: Consider view cache staleness for `/query` and `/invoke`
A warning header should additionally be issued for `/query` and `/invoke` when a stale view cache is involved.

drostron:
Closes #2661


effect/src/main/scala/quasar/effect/Writer.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+
+sealed abstract class Writer[W, A]
+
+object Writer {
+  final case class Tell[W](w: W) extends Writer[W, Unit]
+
+  final case class Listen[W, A](f: W => A) extends Writer[W, A]
```
wemrysi: We might want to remove this constructor in favor of just using `Read`. Having a "sink" effect could be useful for cases like logging where we don't really need to be able to read back the written value.

web/src/main/scala/quasar/api/QHttpService.scala
```diff
@@ -40,11 +41,11 @@ final case class QHttpService[S[_]] private (val f: PartialFunction[Request, Fre
     new QHttpService(f orElse other.f)
 
   def toHttpService(i: S ~> ResponseOr): HttpService =
-    toHttpServiceF(free.foldMapNT(i))
+    toHttpServiceF(free.foldMapNT(i).η[Task])
 
-  def toHttpServiceF(i: Free[S, ?] ~> ResponseOr): HttpService = {
-    def mkResponse(prg: Free[S, QResponse[S]]) =
-      i(prg).map(_.toHttpResponseF(i)).merge
+  def toHttpServiceF(i: Task[Free[S, ?] ~> ResponseOr]): HttpService = {
```
wemrysi: I _think_ we can avoid this change, it shouldn't be necessary.

web/src/main/scala/quasar/server/Server.scala
```diff
@@ -96,39 +98,45 @@ object Server {
...
+            liftMT[Task, ResponseT]                               :+:
+            qErrsToResponseT[Task]
+        ) compose
+          (injectFT[VCacheExpW, QErrs_CW_Task] :+: injectFT[Task, QErrs_CW_Task] :+: eval))
+
     (reload: Int => Task[String \/ Unit]) =>
-      finalizeServices(
-        toHttpServices(liftMT[Task, ResponseT] :+: webInter(eval), coreServices[CoreEffIO]) ++
+    finalizeServices(
+      toHttpServices(interp, coreServices[CoreEffIOW]) ++
```
wemrysi: If you use the previous version of `toHttpServicesF` here instead, you should be able to do something like
```scala
(fa: Free[S, A]) => interp.liftM[ResponseT] flatMap (fa foldMap _)
```
and not have to modify the signature of `QHttpService.toHttpServiceF`.

drostron:
Agreed, definitely want to move caching towards an orthogonal application over time.

core/src/main/scala/quasar/fs/mount/cache/VCache.scala
```diff
@@ -18,20 +18,36 @@ package quasar.fs.mount.cache
...
+  type VCacheKVS[A] = KeyValueStore[AFile, ViewCache, A]
+
+  object VCacheKVS {
+    type Ops[S[_]] = KeyValueStore.Ops[AFile, ViewCache, S]
 
-  implicit def Ops[S[_]](implicit S0: VCache :<: S): Ops[S] = KeyValueStore.Ops[AFile, ViewCache, S]
+    def Ops[S[_]](implicit S0: VCacheKVS :<: S): Ops[S] = KeyValueStore.Ops[AFile, ViewCache, S]
+  }
+
+  type VCacheExpW[A] = Writer[List[Expiration], A]
```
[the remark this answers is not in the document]
drostron: I like that idea, although we might require a fuller set of expirations at some point. Of course in that case we'd likely want something richer than just expiration anyway which brings us back to `MinOption`.

effect/src/main/scala/quasar/effect/Writer.scala
drostron: To clarify, would it be up to the interpreters to tie the `Sink` and `Read` effects together or are you thinking of something else here?

wemrysi: Aye, that's what I was thinking.

drostron:
@wemrysi Ready for another look

PR #2951, opened 2017-10-18
Stop MimirStdLibSpec executing tests twice

edmundnoble:
Just noticed this on the way through. `MimirStdLibSpec` always executed test queries twice, changing this repeatably shaves around 600 ms off test suite run time, which I suppose means setup time and specs2 are taking a good fraction.


djspiewak:
@edmundnoble ☝️☝️☝️☝️

PR #2952, opened 2017-10-18
Remove last case of dead code in untyper

edmundnoble:
Alissa showed me an extra example of dead code in the untyper of `MakeObject` after merging the last dead code removal, looks to be the last.

### DIFF
--- a/frontend/src/main/scala/quasar/std/structural.scala
+++ b/frontend/src/main/scala/quasar/std/structural.scala
@@ -42,14 +42,7 @@ trait StructuralLib extends Library {
       case Sized(Const(Data.Str(name)), valueType)   => Obj(Map(name -> valueType), None)
       case Sized(_, valueType)                       => Obj(Map(), Some(valueType))
     },
-    partialUntyperV[nat._2] {
-      case Obj(map, uk) => map.headOption.fold(
-        uk.fold[Func.VDomain[nat._2]](
-          failure(NonEmptyList(GenericError("MAKE_OBJECT can’t result in an empty object"))))(
-          t => success(Func.Input2(Str, t)))) {
-        case (key, value) => success(Func.Input2(Const(Data.Str(key)), value))
-      }
-    })
+    basicUntyper)
 
   val MakeArray = UnaryFunc(
     Mapping,


frontend/src/main/scala/quasar/std/structural.scala
```diff
@@ -42,14 +42,7 @@ trait StructuralLib extends Library {
       case Sized(Const(Data.Str(name)), valueType)   => Obj(Map(name -> valueType), None)
       case Sized(_, valueType)                       => Obj(Map(), Some(valueType))
     },
-    partialUntyperV[nat._2] {
-      case Obj(map, uk) => map.headOption.fold(
-        uk.fold[Func.VDomain[nat._2]](
```
alissapajer: I think the case when `map.headOption` returns `None` may not be dead code. I don't think we can know in the untyper any actual `String` value of a key in the map, meaning that `map` would always be an empty `Map[String, Type]`. But could `uk` have a value in it, meaning we should keep around this case? If `uk` is populated, then that `Type` is the type of the values contained in the object we are returning from `MakeObject`. Is there an example of how we can reach this case, or an argument for why we can never reach it?

The untyper below for `MakeArray` is an analogous situation.

djspiewak:
Needs a version label, btw.

frontend/src/main/scala/quasar/std/structural.scala
edmundnoble: I don't see any functions with input types that are specializations of `Obj`; all of them just take in `AnyObject`, which means there's no place for the information to propagate from. So it doesn't seem possible for us to reach this case.

alissapajer: Can the analogous logic be applied to `case Arr(List(elemType)) =>` in `MakeArray` below?

alissapajer:
@edmundnoble we had discussed possibly closing this. What do you think?

edmundnoble:
@alissapajer I'm trying now to add precise types to the untyper for ObjectProject, and delete the `map.headOption.fold` case because as you say it's impossible to know any keys in particular. If that doesn't work I'll close this.

alissapajer:
This is minor in the grand scheme of getting 5.0.0 out the door, so if it's not a quick fix I think we should revisit it later.

Also, eventually I'd like to move typechecking into qscript, and then we can revisit how we want to structure the whole algorithm.

edmundnoble:
:+1: for revisiting it later, right now it looks very much like it must be dead code, now that I know I was wrong about what `Type.Obj.unknowns` means and there is no way to constrain `unknowns` to anything other than `None` or `Some(Type.Top)` in the codebase.

PR #2955, opened 2017-10-18
Avoid extracting guards from OR expressions.

wemrysi:
Pulling guards out of `or` expressions has the effect of requiring all the guards in the expression to pass before the `or` is evaluated whereas we'd rather evaluate each operand to the `or` independently, guards and all.

This is necessary for the search card functionality in SlamData to work (see slamdata/slamdata#2367) as well as other queries involving guards in filter disjunctions.


alissapajer:
@wemrysi could you add an integration test with a minimal example of guards in filter disjunctions?

PR #2966, opened 2017-10-22
express $$a differently so that it doesn't get rewritten

### LINKED ISSUE #2933: Queries with a field and a reduction return undefined
Consider the query

```
select state, sum(pop) from smallZips
```

The expected results are objects with the shape:

```
{"state": "MA", "1": 12093}
{"state": "MA", "1": 12093}
{"state": "MA", "1": 12093}
```

where `1` is the sum of the populations in `smallZips`. One per entry in `smallZips`. Instead, we get an object with a single key. 

```
{"1": 12093}
```

This issue is only present in the MongoDB connector. Using a different reduction like `count` showcases the same behavior. Mimir and Marklogic both return correct results.

The query `select city, state, sum(pop) from smallZips` works correctly as well, even in the Mongo connector.

rintcius:
Fixes #2933


rintcius:
This issue was that in its old form `$$a` gets rewritten into `$$ROOT`, in the new form that's avoided. 

Still feel it's a bit of a workaround though because the old form looks like the better representation.  I think the real issue is the [prefixBase](https://github.com/quasar-analytics/quasar/blob/834fab03747853c87eea6850ae91112e9881935a/mongodb/src/main/scala/quasar/physical/mongodb/workflow/package.scala#L216-L217) function using [`\\`](https://github.com/quasar-analytics/quasar/blob/834fab03747853c87eea6850ae91112e9881935a/mongodb/src/main/scala/quasar/physical/mongodb/expression/docvar.scala#L48) and not doing an actual prefix in all cases, but I don't see how to fix that without risk. Glad to change though if someone does see it.

PR #2971, opened 2017-10-24
add archive metadata hidden file to single zip download

### LINKED ISSUE #2652: Export -> Import .zip file fail
### Description

Downloading a file in `.zip` format, when uploading that file you get given an error.

### Reproduction guide

on Mongo: download `smallZips` selecting `.zip` -> upload that zip `smallZips.zips` -> Error message

### Results
Here are screen shots of the error and network headers etc.

screen shot 2017-08-18 at 12 07 21
screen shot 2017-08-18 at 12 13 13

nicflores:
Fix inability to upload downloaded single zipped file. 

Fixes #2652.

### DIFF
--- a/web/src/main/scala/quasar/api/services/package.scala
+++ b/web/src/main/scala/quasar/api/services/package.scala
@@ -84,15 +84,22 @@ package object services {
     S0: Failure[E, ?] :<: S,
     S1: Task :<: S
   ): Free[S, QResponse[S]] = {
-    val headers: List[Header] = `Content-Type`(MediaType.`application/zip`) :: 
-      (format.disposition.toList: List[Header])
-    val p = format.encode(data).map(str => ByteVector.view(str.getBytes(StandardCharsets.UTF_8)))
+    val headers: List[Header] = `Content-Type`(MediaType.`application/zip`) :: (format.disposition.toList: List[Header])
+
     val suffix = format match {
           case JsonContentType(_, _, _) => "json"
           case Csv(_, _) => "csv"
         }
-    val f = currentDir[Sandboxed] </> file1[Sandboxed](fileName(filePath).changeExtension(κ(suffix))) 
-    val z = Zip.zipFiles(Map(f -> p))
+
+    val f = currentDir[Sandboxed] </> file1[Sandboxed](fileName(filePath).changeExtension(κ(suffix)))
+    val qFilesAndContent = List(f).toList.map { file => (file, format.encode(data)) }
+
+    val metadata = ArchiveMetadata(List(f).strengthR(FileMetadata(`Content-Type`(format.mediaType))).toMap)
+    val metaFileAndContent = (ArchiveMetadata.HiddenFile, Process.emit(metadata.asJson.spaces2))
+
+    val allFiles = metaFileAndContent :: qFilesAndContent
+    val z = Zip.zipFiles(allFiles.toMap.mapValues(strContent => strContent.map(str => ByteVector.view(str.getBytes(StandardCharsets.UTF_8)))))
+
     QResponse.streaming(z).map(QResponse.headers.modify(_ ++ headers))
   }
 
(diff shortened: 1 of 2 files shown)


web/src/main/scala/quasar/api/services/package.scala
```diff
@@ -84,15 +84,22 @@ package object services {
...
           case JsonContentType(_, _, _) => "json"
           case Csv(_, _) => "csv"
         }
-    val f = currentDir[Sandboxed] </> file1[Sandboxed](fileName(filePath).changeExtension(κ(suffix))) 
-    val z = Zip.zipFiles(Map(f -> p))
+
+    val f = currentDir[Sandboxed] </> file1[Sandboxed](fileName(filePath).changeExtension(κ(suffix)))
+    val qFilesAndContent = List(f).toList.map { file => (file, format.encode(data)) }
+
+    val metadata = ArchiveMetadata(List(f).strengthR(FileMetadata(`Content-Type`(format.mediaType))).toMap)
```
jedesah: This is a roundabout way of creating a single element `Map`

web/src/main/scala/quasar/api/services/package.scala
```diff
@@ -84,15 +84,22 @@ package object services {
...
+
     val suffix = format match {
           case JsonContentType(_, _, _) => "json"
           case Csv(_, _) => "csv"
         }
-    val f = currentDir[Sandboxed] </> file1[Sandboxed](fileName(filePath).changeExtension(κ(suffix))) 
-    val z = Zip.zipFiles(Map(f -> p))
+
+    val f = currentDir[Sandboxed] </> file1[Sandboxed](fileName(filePath).changeExtension(κ(suffix)))
+    val qFilesAndContent = List(f).toList.map { file => (file, format.encode(data)) }
```
jedesah: `List.toList` is a no-op (IntelliJ would have told you this kind of thing! 😄 ) and there is no need to `map` since we are building a list of a single element right before

web/src/main/scala/quasar/api/services/package.scala
```diff
@@ -84,15 +84,22 @@ package object services {
...
+
+    val metadata = ArchiveMetadata(List(f).strengthR(FileMetadata(`Content-Type`(format.mediaType))).toMap)
+    val metaFileAndContent = (ArchiveMetadata.HiddenFile, Process.emit(metadata.asJson.spaces2))
+
+    val allFiles = metaFileAndContent :: qFilesAndContent
```
jedesah: How about `fileAndMeta`? `allFiles` is a little misleading I think (although technically correct)

web/src/main/scala/quasar/api/services/package.scala
jedesah: Also, I'd create `Map`s above and concatenate them together here and then avoid the `toMap` lower

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -340,6 +340,28 @@ class DataServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s {
...
+          "download single zipped file and then re-upload zipped file" >> { 
+            val baseDir = rootDir[Sandboxed]
+            val sampleFile = baseDir </> file("foo.json")
+            val data = Vector(Data.Obj("a" -> Data.Str("foo"), "b" -> Data.Bool(true)))
+            val initialContent = Map(sampleFile -> data)
+
+            val downloadRequest = Request(
+                  uri = pathUri(sampleFile),
+                  headers = Headers(Header("Accept", "application/zip,application/json;disposition=\"attachment;filename*=UTF-8''foo.zip\"")))
+            val (originalService, _) = serviceRef(InMemState.fromFiles(initialContent))
```
jedesah: If we don't need the "Ref", I'm pretty sure there is a `service` alternative

PR #2977, opened 2017-10-25
Use ManagedReadFile for RDBMS fs/ReadFile

kciesielski:
This implementation uses streaming and incremental loading from the DB instead of putting all in the memory.


rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
```diff
@@ -20,99 +20,82 @@ import slamdata.Predef._
...
+    def getTableCursor(
+        dbPath: TablePath,
+        cfg: Config,
+        offset: Natural,
+        limit: Option[Positive]
+    ): Backend[DbDataStream] = {
+
+      transactor(cfg).map { xa =>
+        DbDataStream(selectAllQuery(dbPath, offset, limit)
+                       .query[Data]
```
kciesielski: This can throw an exception. How can I handle it to make it a stream of `FileSystemError \/ Data`?

kciesielski:
@wemrysi Can you take a look? Also, if you could bump the build please.

djspiewak:
Assigning @wemrysi for the time being.  This probably isn't something we're going to be able to look at in the next week or two though, because of other priorities.

rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
drostron: You might be looking for `Process.attempt`

rdbms/src/main/scala/quasar/physical/rdbms/Interpreter.scala
```diff
@@ -36,14 +36,14 @@ trait Interpreter {
...
+      TaskRef(Map.empty[ReadHandle, DbDataStream]) |@|
+        TaskRef(Map.empty[WriteHandle, TablePath]) |@|
         xa.map(_.trans) |@|
         TaskRef(0L) |@|
         GenUUID.type1[Task]
-    )(
+      )(
       (kvR, kvW, x, i, genUUID) =>
-
+        reflNT[Task]                        :+:
```
drostron: Alignment nit here and above — we usually left align these. Also, might as well keep `:+:` spaced in a similar fashion to its siblings.

drostron:
Would be great to see tests running against postgres in Travis. I realize setting that up with integration tests partly pending isn't supported easily out of the box, so 👌🏾.

rdbms/src/main/scala/quasar/physical/rdbms/Interpreter.scala
kciesielski: Sure, I'll realign this.

rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsReadFile.scala
kciesielski: Exactly, thanks!

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresScanTable.scala
```diff
@@ -0,0 +1,48 @@
...
+   }
+
+  def offsetFr(offset: Natural): Fragment = {
+    if (offset > 0)
+      fr"OFFSET" ++ Fragment.const(offset.shows)
+    else Fragment.empty
+  }
+
+  override def selectAllQuery(tablePath: TablePath, offset: Natural, limit: Option[Positive]): Fragment = {
+    fr"select * from " ++ Fragment.const(tablePath.shows) ++ limitFr(limit) ++ offsetFr(offset)
```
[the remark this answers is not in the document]
kciesielski: Right, I'll fix this.

kciesielski:
@drostron Thanks for the review, all the remarks are now addressed and the PR is merged with master.

PR #2979, opened 2017-10-25
Fix for bug2958 - wrong typechecks for Invoke(IfUndefined)

### LINKED ISSUE #2958: Various structure editor queries result in missing data
A couple of Quasar versions back, I had all 4 of the currently supported structure edit operations working on the Spark local connector, but trying on both mimir and mongo in Quasar 25.0.16 they're not giving the results I'd expect.

Using this dataset:
``` json
{ "a": 1, "b": 2 }
{ "a": 2, "b": 4 }
{ "a": 3, "b": 6 }
```
And this query:
``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "b"), {"c": OBJECT_CONCAT(t0.c, t0.b) ?? t0.b}) FROM `/mimir/test-data` AS t0
```
I get this result:
``` json
{ "a": 1 }
{ "a": 2 }
{ "a": 3 }
```
But was expecting:
``` json
{ "a": 1, "c": 2 }
{ "a": 2, "c": 4 }
{ "a": 3, "c": 6 }
```

I can include a few other example queries, but they're of similar form, so I'm pretty sure the problem is arising due to the `OBJECT_CONCAT` and `??` on the inside, as a query like this:

``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "b"), {"c": t0.b}) FROM `/mimir/test-data` AS t0
```

Gives the expected result. So maybe it's `IfUndefined` issues again? Although last time was mimir-specific, I think. `OBJECT_CONCAT(t0.c, t0.b)` should definitely be undefined in this case (but we need that in the query for the intended behaviour in cases where `t0.c` does also already exist), so it seems like the `?? t0.b` fallback should be working there.

rabbitonweb:
Fixes #2958


rabbitonweb:
@drostron this PR is now in CI but should be green in matter of minutes. Is ready for review

rabbitonweb:
@drostron solution here required adding special  case for `Invoke(IfUndefined)` by applying constraints immediately (not pushing them up). 
I've added two tests: integration one (regression test) and two unit tests. One of those tests was crafted simply in order to understand how `inferTypes` function works. This test should be extended with more examples in the future and I've added an issue for that https://github.com/quasar-analytics/quasar/issues/3061

alissapajer:
@rabbitonweb in the PR description you need to use one of the github keywords to actually close #2958.

rabbitonweb:
@alissapajer what do you mean by that?

rintcius:
@rabbitonweb for example: 

> Fixes #2958

See https://help.github.com/articles/closing-issues-using-keywords/

rabbitonweb:
thx @rintcius

rabbitonweb:
@drostron ping :)

drostron:
@rabbitonweb Planning on taking a look today 😄

it/src/main/resources/tests/bug2958.test
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "bug2958",
+  "data": "bug2958.data",
+  "backends": {
+    "couchbase":         "pending",
+    "marklogic_xml":     "pending"
+  },
+  "NB": "couchbase is pending becasue the value c contains null.  marklogic_xml  is pending because the valeu c is {}. Both issues seem connector specific and are not related to bug2958",
+  "query": "SELECT MAP_CONCAT(DELETE_KEY(t0, \"b\"), {\"c\": MAP_CONCAT(t0.c, t0.b) ?? t0.b}) FROM bug2958 AS t0",
```
drostron: Does this test surface the issue? I ask since it is introduced with "Regression test for bug2958 (which is apparently solved)" and hasn't been adjusted as such.

rabbitonweb: This test fails without the fix done in `LogicalPlanR`. It introduced as "regresiion tests for bug that is fixed" but that was a lie - bug was not manifesting itself on spark but was failing on other connectors

drostron: 👌🏾

drostron:
@rabbitonweb Needs a semver label

PR #2980, opened 2017-10-25
2541 order typechecks

### LINKED ISSUE #2541: date_part should return undefined when applied to a non-temporal argument in MongoDB

jsantos17:
Fixes #2541


jsantos17:
There are still two filter tests I haven't been able to fix:

1. plan filter with both index and field projections
2. plan filter on date

For these two tests, the `typeSelector` impl in this PR is unable to extract inputs and I'm not really sure why.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1020,22 +1036,42 @@ object MongoDbPlanner {
...
+          case Filter(src, cond0) => {
+            // TODO: Apply elideMoreGeneralGuards to all FreeMap's in the plan, not only here
+            cond0.transCataM(assumeReadType.elideMoreGeneralGuards[M, T](Type.AnyObject)) >>= { cond =>
+              val selectors = getSelector[T, M, EX](cond, selector[T](cfg.bsonVersion)).toOption
+              val typeSelectors = getSelector[T, M, EX](cond, typeSelector[T]).toOption
+
+              (selectors, typeSelectors) match {
+                case (Some((_, Nil)), Some((typeSel, typeInputs))) =>
+                  typeInputs.traverse(f => handleFreeMap[T, M, EX](cfg.funcHandler, cfg.staticHandler, f(cond)))
+                    .map(WB.filter(src, _, typeSel))
```
rintcius: maybe worth creating a function for this, since it's being used several times

jsantos17:
@rintcius ready for another look.

djspiewak:
Merging on account of @rintcius's approval.  This never got an assignee, so I'll take it and just trust his review.  :-D

PR #2990, opened 2017-10-27
QScript/EJson construction DSL

### LINKED ISSUE #2989: QScript DSL
QScript could use a DSL to make it easier to write, especially for testing the mongo planner for 5.0. Ideally the way it would work is something like `val dsl = makeDSL[QScriptDialect]; import dsl._` and everything is magically nice to use.

Would also be beautiful to have something like a `Show` instance which prints out QScript as this DSL. It's literally not possible to do this right now because QScript construction is so explicit.

edmundnoble:
Designed for use in app code and in tests. Instantiate a `Dsl` and a `Func`, or (more likely, for tests) just use `makeDefaults` to get a `Dsl` for your `FreeQS` analogue and your `Fix[QST]` analogue as well as a `Func` for your `T[_[_]]`.

Edit: See `ExprOpCoreF.fixpoint` for a *very* similar DSL for Mongo plans.

Fixes #2989

### DIFF
--- a/connector/src/main/scala/quasar/qscript/construction.scala
+++ b/connector/src/main/scala/quasar/qscript/construction.scala
@@ -1,0 +1,302 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript
+
+import slamdata.Predef._
+import matryoshka._
+import matryoshka.data.Fix
+
+import scalaz._
+import scalaz.Leibniz.===
+import quasar.qscript
+import quasar.common.{JoinType, SortDir}
+import quasar.std.TemporalPart
+import quasar.ejson.EJson
+
+object construction {
+  final case class Func[T[_[_]]]()(implicit corec: CorecursiveT[T]) {
+      private def rollCore[A](in: MapFuncCore[T, FreeMapA[T, A]]): FreeMapA[T, A] = Free.roll(MFC(in))
+      private def rollDerived[A](in: MapFuncDerived[T, FreeMapA[T, A]]): FreeMapA[T, A] = Free.roll(MFD(in))
+      def Constant[A](in: T[EJson]): FreeMapA[T, A] =
+        rollCore(MapFuncsCore.Constant(in))
+      def Undefined[A]: FreeMapA[T, A] =
+        rollCore(MapFuncsCore.Undefined())
+      def Now[A]: FreeMapA[T, A] =
+        rollCore(MapFuncsCore.Now())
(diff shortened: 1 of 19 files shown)


djspiewak:
@edmundnoble version label!

edmundnoble:
Aaaaaargh the shame :cry:

jsantos17:
@edmundnoble do we need the `Injectable` constraints? Or is `Inject` enough? Are the `Injectable` instances easily available? 

I ask because I'd prefer not to have to declare the instances before building a QScript tree for a test.

Otherwise looks good 👍

edmundnoble:
@jsantos17 from what I understand `Injectable` instances must be available for you to construct QScript even without this DSL, because there's an immense compile-time overhead in using `Inject` for the same thing. They're not too terrible; they're something like:
```scala
  implicit val QS: Injectable.Aux[QS, QST] =
    ::\::[QScriptCore](
      ::\::[ThetaJoin](
        ::\::[Const[Read[ADir], ?]](
          ::/::[Fix, Const[Read[AFile], ?], Const[DeadEnd, ?]])))
```
And you only need *one* per QScript dialect, so you should only need one for the entire mongo connector.

djspiewak:
> And you only need one per QScript dialect, so you should only need one for the entire mongo connector.

In fact, one is already available.  `MongoDB.qScriptToQScriptTotal`

edmundnoble:
Actually my bad. You need multiple inject instances, one for your dialect to QST, one for QSC to your dialect, and one for every extra qscript node type in your dialect in addition to QSC.

edmundnoble:
@jsantos17 I've added these instances for every dialect of QScript into package objects so they can be easily accessed.

connector/src/main/scala/quasar/qscript/construction.scala
```diff
@@ -0,0 +1,302 @@
...
+import quasar.qscript
+import quasar.common.{JoinType, SortDir}
+import quasar.std.TemporalPart
+import quasar.ejson.EJson
+
+object construction {
+  final case class Func[T[_[_]]]()(implicit corec: CorecursiveT[T]) {
+      private def rollCore[A](in: MapFuncCore[T, FreeMapA[T, A]]): FreeMapA[T, A] = Free.roll(MFC(in))
+      private def rollDerived[A](in: MapFuncDerived[T, FreeMapA[T, A]]): FreeMapA[T, A] = Free.roll(MFD(in))
+      def Constant[A](in: T[EJson]): FreeMapA[T, A] =
```
alissapajer: There are already a handful of such constructors over in [`QScriptHelpers`](https://github.com/quasar-analytics/quasar/blob/master/connector/src/test/scala/quasar/qscript/QScriptHelpers.scala#L84) so we can remove those in favor of these.

Though also, we could add an `apply` to parallel this [`unapply`](https://github.com/quasar-analytics/quasar/blob/master/connector/src/main/scala/quasar/qscript/package.scala#L145) to avoid listing all these cases, but then construction of the actual qscript wouldn't be quite as clean.

edmundnoble: I'm not hugely bothered by the case listing myself, I can maintain it fairly nicely with macros and I wanted it to be maximally clean to construct.

I'll rip out QScriptHelpers in a further PR, and I'll take a look at the helpers in ExpandMapFunc as well.

alissapajer:
@edmundnoble should this PR be closing #2989?

edmundnoble:
Yes, it should. I'll open separate issues for the pretty-printer and replacing QScriptHelpers.

edmundnoble:
I'll address the build failure in an hour when I get home.

djspiewak:
I'll merge this as soon as the build is clean, since both assignees have approved.

rintcius:
Merged PR since build is green now

PR #2993, opened 2017-10-30
Shifted Read planner for RDBMS

kciesielski:


djspiewak:
Assigning @wemrysi for the time being.  This probably isn't something we're going to be able to look at in the next week or two though, because of other priorities.

rabbitonweb:
Assigning this to myself

alissapajer:
@rabbitonweb build is green.

PR #2999, opened 2017-10-31
Move towards the desired map and key nomenclature

### LINKED ISSUE #2808: Map vs Object naming in SQL² functions
Maybe we could sort this out a bit? Either by adding both varieties of the functions, or settling on a consistent choice :smile:

Currently there's:
- `OBJECT_CONCAT`
- `MAKE_OBJECT`
- `FLATTEN_MAP`
- `SHIFT_MAP`

Also, we refer to maps having "keys" everywhere, but then the function for deleting them is called `DELETE_FIELD`.

### LINKED ISSUE #2859: Consistent naming of logical plan and qscript functions
We need to make the function names across logical plan and qscript consistent (and the same where possible). These should also be consistent with the names of the SQL² functions described in https://github.com/quasar-analytics/quasar/issues/2808.

For example, in logical plan we have `ObjectProject` and in qscript we have `ProjectField`. See https://github.com/quasar-analytics/quasar/issues/2808#issuecomment-332976168 for more description.

drostron:
Closes #2808 
Closes #2859


djspiewak:
Did this get push -f'd?

drostron:
Yes, quick fix for a build compilation failure due to an updated master being merged between PR submission and build. Will merge to avoid confusion in the future. No one was basing work on it that I knew of and it didn't have a clean build yet.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -243,11 +243,11 @@ final class Compiler[M[_], T: Equal]
...
       CIName("null")                    -> string.Null,
       CIName("to_string")               -> string.ToString,
-      CIName("make_object")             -> structural.MakeObject,
+      CIName("make_map")                -> structural.MakeMap,
       CIName("make_array")              -> structural.MakeArray,
-      CIName("object_concat")           -> structural.ObjectConcat,
+      CIName("map_concat")              -> structural.MapConcat,
       CIName("array_concat")            -> structural.ArrayConcat,
-      CIName("delete_field")            -> structural.DeleteField,
+      CIName("delete_key")              -> structural.DeleteKey,
```
alissapajer: @garyb these three SQL^2 function names have been changed per #2808 , and so the frontend will have to update its function generation when upgrading to this (currently unreleased) version of quasar. Is there any further coordination we need to do for this?

garyb: I don't think so - the changes on our end will be pretty trivial, we'll just have to communicate amongst ourselves that our local Quasar / SDBEs need updating when this change goes in to the FE.

Will this result in a major bump of Quasar?

garyb: & thanks for checking / pinging!

alissapajer: Yes, we have the `version: breaking` label, which means that it will be a major version bump!!

PR #3003, opened 2017-10-31
Initial work on ReadLP

djspiewak:
First steps toward LP => QSU.  Hopefully the stub-ish areas are obvious.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
@@ -16,11 +16,26 @@
 
 package quasar.qscript.qsu
 
-import slamdata.Predef.{Map => SMap, Symbol}
+import quasar.fp._
+import slamdata.Predef.{Map => SMap, _}
+
+import scalaz.syntax.equal._
 
 final case class QSUGraph[T[_[_]]](
     root: Symbol,
     vertices: SMap[Symbol, QScriptUniform[T, Symbol]]) {
+
+  def rename(from: Symbol, to: Symbol): QSUGraph[T] = {
+    val root2 = if (from === root)
+      from
+    else
+      root
+
+    val vertices2 =
+      vertices.get(from).map(node => vertices - from + (to -> node)).getOrElse(vertices)
+
+    QSUGraph(root2, vertices2)
+  }
 
   /**
    * Uniquely merge the graphs, retaining the root from the right.
--- a/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
@@ -20,4 +20,15 @@ import slamdata.Predef.{Map => SMap, Symbol}
 
 final case class QSUGraph[T[_[_]]](
     root: Symbol,
-    vertices: SMap[Symbol, QScriptUniform[T, Symbol]])
+    vertices: SMap[Symbol, QScriptUniform[T, Symbol]]) {
+
+  /**
+   * Uniquely merge the graphs, retaining the root from the right.
+   */
+  def ++:(left: QSUGraph[T]): QSUGraph[T] = QSUGraph(root, vertices ++ left.vertices)
+
+  /**
+   * Uniquely merge the graphs, retaining the root from the left.
+   */
+  def :++(right: QSUGraph[T]): QSUGraph[T] = right ++: this
+}
--- a/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
@@ -16,26 +16,11 @@
 
 package quasar.qscript.qsu
 
-import quasar.fp._
 import slamdata.Predef.{Map => SMap, _}
-
-import scalaz.syntax.equal._
 
 final case class QSUGraph[T[_[_]]](
     root: Symbol,
     vertices: SMap[Symbol, QScriptUniform[T, Symbol]]) {
-
-  def rename(from: Symbol, to: Symbol): QSUGraph[T] = {
-    val root2 = if (from === root)
-      from
-    else
-      root
-
-    val vertices2 =
-      vertices.get(from).map(node => vertices - from + (to -> node)).getOrElse(vertices)
-
-    QSUGraph(root2, vertices2)
-  }
 
   /**
    * Uniquely merge the graphs, retaining the root from the right.
--- a/connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
@@ -21,11 +21,10 @@ import quasar.{RenderTree, RenderTreeT}
 import quasar.common.{JoinType, SortDir}
 import quasar.contrib.pathy.AFile
 import quasar.ejson.EJson
-import quasar.fp.numeric.Natural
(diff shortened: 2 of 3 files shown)


djspiewak:
Current status: 

- I'm pretty sure `Read` is just wrong
- There's some boilerplate that could probably be factored out a bit
- `Range` is unhandled, mostly because the current handling in qscript is completely wrong, and then the handling in connectors is more differently wrong, and basically nothing works.  We should probably just be compatible with qscript in QSU, with very clear TODO markers for when we go make it right
- I left the fallthrough case `case lp.InvokeUnapply(func, values) => ???`, even though I'm pretty sure everything is now handled
  + Except `UnshiftObject`, which isn't handled

connector/src/main/scala/quasar/qscript/qsu/ReadLP.scala
```diff
@@ -0,0 +1,243 @@
...
+          Free.pure[MapFunc, Hole](SrcHole))
+
+      val node = QSU.Map[T, Symbol](a.root, Free.roll[MapFunc, Hole](translated))
+      withName(node).map(_ :++ a)
+
+    case lp.InvokeUnapply(func: BinaryFunc, Sized(a, b)) if func.effect === Mapping =>
+      val translated =
+        MapFunc.translateBinaryMapping[T, MapFunc, Int].apply(func)(0, 1)
+
+      val node = QSU.AutoJoin[T, Symbol](NEL(a.root, b.root), translated)
```
wemrysi: A smart function for constructing `AutoJoin` that ensures the symbols and indices align properly might be a worthwhile addition, we only create a handful of these but it would be easy to transpose something and break things.

connector/src/main/scala/quasar/qscript/qsu/ReadLP.scala
```diff
@@ -0,0 +1,243 @@
...
+
+sealed abstract class ReadLP[
+    T[_[_]]: BirecursiveT,
+    F[_]: Monad: MonadError_[?[_], PlannerError]: NameGenerator]
+    extends TTypes[T] {
+
+  private val IC = Inject[MapFuncCore, MapFunc]
+  private val ID = Inject[MapFuncDerived, MapFunc]
+
+  private val IdIndex = 0
```
wemrysi: :+1:

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -42,7 +42,7 @@ object QScriptUniform {
 
   final case class AutoJoin[T[_[_]], A](
       sources: NEL[A],
-      combiner: MapFuncCore[T, Natural]) extends QScriptUniform[T, A]
+      combiner: MapFunc[T, Int]) extends QScriptUniform[T, A]
```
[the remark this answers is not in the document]
djspiewak: `Natural` sort of worked, but it required some more implicits just to get it off the ground, and it wasn't worth the effort.  This is particularly true since `Natural` is actually a `Long`, so there's coercion here regardless.  We're already asserting that the numbers in question line up with valid indices of the `NEL`; generalizing that to saying "well, there are no valid negative indices" isn't that much of a stretch.  So tl;dr, `Natural` isn't really a benefit and it imposes some extra headache just to get it working.

connector/src/main/scala/quasar/qscript/qsu/ReadLP.scala
```diff
@@ -0,0 +1,243 @@
...
+          case d =>
+            NonRepresentableData(d).left
+        },
+        { x: T[EJson] => QSU.Constant[T, Symbol](x).right })
+
+      MonadError_[F, PlannerError].unattempt(back.point[F]).flatMap(withName)
+
+    case lp.InvokeUnapply(StructuralLib.FlattenMap, Sized(a)) =>
+      val transpose =
+        withName(QSU.Transpose[T, Symbol](a.root, QSU.Rotation.FlattenMap)).map(_ :++ a)
```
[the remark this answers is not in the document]
djspiewak: I would need a different version for this arity, but I agree doing so would be better than what is here now.

connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -16,8 +16,34 @@
...
+import scalaz.syntax.equal._
 
 final case class QSUGraph[T[_[_]]](
     root: Symbol,
-    vertices: SMap[Symbol, QScriptUniform[T, Symbol]])
+    vertices: SMap[Symbol, QScriptUniform[T, Symbol]]) {
+
+  def rename(from: Symbol, to: Symbol): QSUGraph[T] = {
+    val root2 = if (from === root)
+      from
```
wemrysi: I think this should be `to`, no?

connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -16,8 +16,34 @@
...
+import slamdata.Predef.{Map => SMap, _}
+
+import scalaz.syntax.equal._
 
 final case class QSUGraph[T[_[_]]](
     root: Symbol,
-    vertices: SMap[Symbol, QScriptUniform[T, Symbol]])
+    vertices: SMap[Symbol, QScriptUniform[T, Symbol]]) {
+
+  def rename(from: Symbol, to: Symbol): QSUGraph[T] = {
```
wemrysi: This appears to only be safe when you know, a priori, that nothing references `from`. It might make more sense to make this private to `ReadLP` since we know in the `let` case this is true. Alternatively we could have it search through the graph and update any references to `from` with `to`.

djspiewak: You are correct about the precondition.  It would be better to confine this to just `Let`.  I don't want to iterate through the graph since we know it's unnecessary in the only place that uses it, and it's much more efficient to just change the names.

connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
djspiewak: Yep.

connector/src/main/scala/quasar/qscript/qsu/ReadLP.scala
```diff
@@ -0,0 +1,292 @@
...
+
+  private def extend3(parent1: QSUGraph[T], parent2: QSUGraph[T], parent3: QSUGraph[T])(
+      constr: (Symbol, Symbol, Symbol) => QSU[T, Symbol]): G[QSUGraph[T]] =
+    withName(constr(parent1.root, parent2.root, parent3.root)).map(_ :++ parent1 :++ parent2 :++ parent3)
+
+  private def withName(node: QSU[T, Symbol]): G[QSUGraph[T]] = {
+    for {
+      reverse <- MS.get
+
+      back <- reverse.get(node) match {
```
alissapajer: We want to use the `scalaz.Equal` instance for `QSU` here.

djspiewak: We can't.  At least, not without defining an `Order` and using `IMap`.  We're sort of stuck with intrinsic equality whenever we use something as a map key.

alissapajer: Yeah, I was suggesting we use `IMap`.

djspiewak: We would need to define an Order though, which is probably possible, but it’s less clear how it would work.

PR #3012, opened 2017-11-01
Correct to_string implementation for ObjectId.

### LINKED ISSUE #2969: SQL^2 OID() function throws an error
Tested against Quasar: 25.0.24

**Sample Query**: 

```sql
select
        oid(test)
from
        `/mongodb-3.4/test/oidtest`
```

This query returns a Server 500 error.

The test data set has just a couple of documents that have a field called **test** with a string representation of a MongoDB ObjectID value.

**Sample Data**:

```
db.oidtest.insert({"test" : "59ef6d6a6c11c02f6e4922c8"})
db.oidtest.insert({"test" : "5852f1e6153eb29eb1f32978"})
```

This same query works in Quasar 20.14.7

### LINKED ISSUE #2974: Error when converting MongoDB ObjectId to String
Quasar Version: 25.0.24

In newer versions of Quasar if you run the following query:

```sql
select to_string(`_id`) from `/mongodb-3.4/data/patients`
```

you will get the following error:

screen shot 2017-10-24 at 15 40 29

In Quasar 20.14.7 (shipped with SlamData 4.2.6) this worked as expected:

screen shot 2017-10-24 at 15 42 30

jsantos17:
Fixes #2974, Fixes #2969 

Not sure if including `Id` in `Syntaxed` makes sense. Opening this for discussion and Travis. `Syntaxed` is used only for `ToString`, so I don't think it'll lead to issues in the future or in other connectors.


frontend/src/main/scala/quasar/types.scala
```diff
@@ -531,7 +531,7 @@ object Type extends TypeInstances {
   val Numeric = Int ⨿ Dec
   val Temporal = Timestamp ⨿ Date ⨿ Time
   val Comparable = Numeric ⨿ Interval ⨿ Str ⨿ Temporal ⨿ Bool
-  val Syntaxed = Type.Null ⨿ Type.Comparable
+  val Syntaxed = Type.Null ⨿ Type.Comparable ⨿ Type.Id
```
jsantos17: This is necessary for `to_string(oid(_id))`. Not sure if that's something we want to allow. It's also necessary for `objectIdToString.test` since AFAIK there's no way to load an `ObjectId` into Mongo for `it` tests.

jsantos17: I guess converting an `Id` to `String` is something that might be useful in general, for any connector.

build.sbt
```diff
@@ -38,6 +38,7 @@ lazy val buildSettings = commonBuildSettings ++ Seq(
 
   scalacOptions ++= Seq(
     "-target:jvm-1.8",
+    "-Ypatmat-exhaust-depth", "40",
```
jsantos17: `MongoDbPlanner` does not compile without this. Not sure why.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -336,6 +344,7 @@ object MongoDbPlanner {
 
     val handleSpecialCore: MapFuncCore[T, JsCore] => M[JsCore] = {
       case Constant(v1) => ejsonToJs[M, T[EJson]](v1)
+      case ToId(a1) => New(Name("ObjectId"), List(a1)).point[M]
```
rintcius: move to `JsFuncHandler` instead?

frontend/src/main/scala/quasar/types.scala
wemrysi: The `Id` type is dubious and seems very mongo-centric, but as it already exists and exposed in SQL<sup>2</sup> I'll give a reluctant 👍. It will go away once we switch to the new type representation in favor of metadata.

@jsantos17 Since you're changing `Syntaxed` you'll need to update the typer/untyper for `StringLib.ToString` to handle `Id`.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -212,7 +212,15 @@ object MongoDbPlanner {
         unimplemented[M, Fix[ExprOp]]("ExtractIsoYear expression")
       case Integer(a1) => unimplemented[M, Fix[ExprOp]]("Integer expression")
       case Decimal(a1) => unimplemented[M, Fix[ExprOp]]("Decimal expression")
-      case ToString(a1) => unimplemented[M, Fix[ExprOp]]("ToString expression")
+      // NB: Using $substr or $substrBytes to convert to string does not work for ObjectId, so we force
+      //     this to be planned using JS
+      case ToString($var(DocVar(DocVar.ROOT, Some(BsonField.Name("_id"))))) =>
```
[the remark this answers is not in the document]
wemrysi: How about moving this to a new function that is run before the `funcHandler`? This would allow this special case to be intercepted and failed while preserving the versioned variants of `ToString`.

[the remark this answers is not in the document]
rintcius: @wemrysi You mean with new function something to use instead of `ToString` (like `ObjectIdToString`), right?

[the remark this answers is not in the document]
wemrysi: No, I mean a new scala function that is called before `funcHandler`. Right now we have `funcHandler` -> `handleSpecial`, I'm proposing `newFunction` -> `funcHandler` -> `handleSpecial`.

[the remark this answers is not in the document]
rintcius: But how do you see this being intercepted just for ObjectId's then?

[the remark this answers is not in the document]
wemrysi: We'd match on the case above and fail, but let everything else through, to be handled by the versioned `funcHandler`.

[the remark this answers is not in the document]
rintcius: Ok, that addresses just the versioning issue then? 
A more serious issue is that the code does not properly recognize ObjectId's. We'd still have that issue, if I understand you correctly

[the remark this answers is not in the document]
wemrysi: Yes I understand the overall approach fails in many cases, I'm just suggesting an improvement if this is the approach that ends up being taken.

build.sbt
djspiewak: Note that most of the pattern matching on qscript/lp is already not checked for exhaustiveness because of a bug in scalac.  You can thus get around exhaustiveness depth warnings by using `@unchecked` on the match target without actually changing anything about the safety of the patmat.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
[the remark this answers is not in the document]
jsantos17: Yes, I don't really see a way to keep the `ToString`  aggregation impl while supporting `ToString` on `ObjectId`s. 

There's nothing to tell us we're converting an `ObjectId`, so we rely on the `_id` string here. Not ideal, but I see no way around it. Hopefully someone else can chime in with other ideas.

jsantos17:
@rintcius ready for another look, except for the versioned `ToString` since I'm still not sure what approach we're going to go with.

build.sbt
jsantos17: What bug is this? I'd like to add a comment above `@unchecked` linking it.

frontend/src/main/scala/quasar/std/string.scala
```diff
@@ -342,6 +342,7 @@ trait StringLib extends Library {
         case Data.Date(d)      => success(d.toString)
         case Data.Time(t)      => success(t.format(DataCodec.timeFormatter))
         case Data.Interval(i)  => success(i.toString)
+        case Data.Id(i)        => success(i.toString)
```
rintcius: I think it's worth adding a test for this if it's not too difficult (I think using `oid(<some literal hex>)` would hit this case)

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -601,22 +485,12 @@ object MongoDbPlanner {
...
-      case Range(start, end)        =>
-        Call(
-          Select(
-            Call(Select(ident("Array"), "apply"), List(
-              Literal(Js.Null),
-              Call(ident("Array"), List(BinOp(Sub, end, start))))),
-            "map"),
-          List(
-            Fun(List(Name("element"), Name("index")),
-              BinOp(jscore.Add, ident("index"), start)))).point[M]
```
rintcius: Needs an extra `case _ => unimplemented[M, JsCore]("JS function")` - would be nice if we could add the name of the specific mapFunc (like is done eg in `handleSpecialDerived`), but I think this is not implemented yet, so that could be added as a TODO

build.sbt
rintcius: Any problems with adding `@unchecked` and removing `"-Ypatmat-exhaust-depth", "40"`?

frontend/src/main/scala/quasar/std/string.scala
jsantos17: Yes, I was unable to use unchecked. I tried `(func: @unchecked) match [...]`, refactoring into a function to tag it with unchecked, but the same error still comes up. What's the correct way to do it?

jsantos17: Added a test for this.

jsantos17:
@rintcius I added `version: revision`. This PR doesn't change any APIs since the `oid` function was already exposed. I just implemented it in new Mongo.

PR #3013, opened 2017-11-01
Make typers and untypers coherent (Fixes #2755)

### LINKED ISSUE #2755: Missing typechecks for results of `Divide`
Whenever we perform a `Division` with an unknown type and a known integer, the typer assumes the result of the division will always be an integer and guards are not inserted before the result is used an integer.

Consider:
```sql
select substring("abcdefg", 0, pop / 10000) from zips
```
yields the logical plan

```
Typechecked:
Invoke(Squash)
╰─ Let('__checku0)
   ├─ Read(/zips)
   ╰─ Typecheck(Obj(Map[], Some(Top)))
      ├─ Free('__checku0)
      ├─ Let('__checku1)
      │  ├─ ObjectProject(Free('__checku0){"pop"})
      │  ╰─ Typecheck(Int)
      │     ├─ Free('__checku1)
      │     ├─ Invoke(Substring)
      │     │  ├─ Str("abcdefg")
      │     │  ├─ Constant(Int(0))
      │     │  ╰─ Invoke(Divide)
      │     │     ├─ Free('__checku1)
      │     │     ╰─ Constant(Int(10000))
      │     ╰─ Constant(NA)
      ╰─ Constant(NA)
```

Note `Invoke(Substring)` is invoked with the unguarded result of the division, even when we don't know the result will be an integer or a decimal/floating point

```
      │     ├─ Invoke(Substring)
      │     │  ├─ Str("abcdefg")
      │     │  ├─ Constant(Int(0))
      │     │  ╰─ Invoke(Divide)
      │     │     ├─ Free('__checku1)
      │     │     ╰─ Constant(Int(10000))
```

The typechecker works properly with literals in the query:

```
select substring("abcdefg", 0, 10 / 10000) from zips
```

results in 

```
Quasar error: [You provided a constant value Dec(0.001) where we expected a Int...]
``` 

(ellipsis mine)

This is affecting Mongo since version 3.4. 3.4 introduced `$substrCP` and deprecated `$substr`. The latter would automatically truncate a decimal before performing obtaining the substring. `$substrCP` is not as tolerant and simply throws an error.

This isn't blocking new Mongo since there's a principled workaround for queries that no longer work due to a lack of magical truncation:

```sql
select substring("abcdefg", 0, trunc(pop / 10000)) from zips
```

Branched off #2667.

edmundnoble:
## Type coherence

To put it bluntly, typers and untypers have been lying to each other. This happens when the typer of a function returns an output type that, when fed through the untyper, doesn't return subtypes of the argument types fed into the typer originally. This is equally the case if you start with the untyper, feed it back through the typer, and get a type which isn't a subtype of the original output type. The Divide bug was one example; here's a list of everything fixed that falls into this category:
- Firstly, `Divide`'s untyper (and codomain) pretended that `Divide` was able to return an `Int` at all. And that this would mean the arguments should be typed as `Int`s. This was the cause of #2755.
- Firstly, most every mathematical operation uses `numericWidening` as a fall-through case for its typer. `numericWidening` was broken, because it operated even on non-numeric types, so coproduct types would cause nonsensical behavior.
- Number of missing cases in `Add`'s typer, which would make adding an interval to a constant date/time/timestamp fail to typecheck.
- Almost every mathematical function has an untyper (`biReflexiveUntyper`) which is incorrect, which assumes that if they output a `Dec` they must have had two `Dec`s as inputs (despite numeric widening mostly making them output a `Dec` even if they have only *one* `Dec` as an input)
- `Multiply` in the typer lets you do `Interval` * `Int` and `Int` * `Interval`, but the `domain` of the function suggests that `Interval` is only allowed on the left
- `Multiply` and `Modulo`'s untypers were `biReflexiveUntyper` which was even worse with `Interval` and `Int`, because despite outputting an `Interval` their arguments are an `Int` and an `Interval` (and in the latter case an `Int` or `Dec` and an `Interval`), *not* two `Interval`s.
- `Subtract` would let you subtract the constant `0` from a coproduct type including an interval and a numeric, and get back the input, even when that doesn't type check.
- `Divide` would return a compile-time error if you divided by `0`, instead of returning `undefined`. This resulted in a test failure after making sure the typer was actually being called, and I don't think it's useful behavior. May be wrong on this.
- Number of functions, including `Divide`, would fail to type check if given constant types despite those constant types conforming to the domain.
- `Negate`, `Abs`, `Ceil`, `Trunc`, `Round`'s untypers handled constant types, despite it not being possible for them to be passed constant types.
- `Multiply`'s untyper says if you get an `Int` out you had two `Int` inputs, however if one input is a constant zero the output will be an `Int` *regardless* of whether the other input is. Also, multiplying an `Interval` by constant zero returns a constant zero, which is wrong, because it should give you the empty `Interval`.

In order to ensure nothing like this ever happens again, `StdLibTypeCoherenceSpec` has been added. It does two things for each function in the stdlib.

1. It generates types which are subtypes of a function's domain, runs them through the typer, then runs that result through the untyper. If the typer failed, the test succeeds; if the untyper failed, the test fails. The test then asserts that the input types that the untyper returns are subtypes of the input types generated at the beginning.
2. It generates types which are subtypes of a function's codomain, runs them through the untyper, then runs that result through the typer. If the untyper failed, the test succeeds; if the typer failed, the test fails. The test then asserts that the output type that the typer returns is a subtype of the output type generated at the beginning.

## Typechecking fixes required
Firstly, `LogicalPlanR`'s `unifyOrCheck` doesn't generate enough Typecheck nodes. If I have a function taking in an `A or B`, and you have a value of type `A or C`, we *should* be able to insert a `Typecheck(A)` node in between, so you can feed your value into my function. `unifyOrCheck` makes a strict subtyping check here; you can feed your `A or C` in *only* if my function takes either an `A` or `C`. Instead, I am now checking for intersections between the two types if they're both coproduct types, so that we can insert those `Typecheck` nodes and not give up.

Secondly, typers and untypers need to be able to essentially "not return an answer" when called. This is because neither the typers nor untypers are exhaustive, nor can they be, because function inputs often have `Coproduct` types, and we'd need to enumerate every possible coproduct which is a subtype of the input. I added the ability for a typer or untyper to "give up" and return `None`, by adding `Option` to their return types. This means that if the inputs aren't matched on by a typer, the inputs are typechecked against the function's domain as a fallback and the function's codomain is returned; conversely, if an output isn't matched on by an untyper, the output is typechecked against the function's codomain as a fallback and the function's domain is returned.

*Dragons*: it could be that this means we provide *less type errors* than previously.


alissapajer:
Needs version label.

djspiewak:
The build failure here appears legitimate.

alissapajer:
Yeah, the spark tests for `divide0.test` stopped working, while the ML ones are now fixed.

edmundnoble:
I think simplifiers actually cannot return `Data.NA` without corrupting the data in some cases, because undefined in arrays and object values is not folded out in the simplifier. That's why the fix here is to remove the line in the simplifier.

rintcius:
@edmundnoble Could you add that as a comment in the code at the places where they'd normally be expected? Or maybe it's even possible to include that in these new coherence tests?

edmundnoble:
@rintcius It's in the tests now. I *think* this doesn't make the tests too multi-purpose, but I'd like @alissapajer's input on that.

alissapajer:
@edmundnoble can you add a unit test similar to [this one](https://github.com/quasar-analytics/quasar/blob/master/sql/src/test/scala/quasar/sql/compiler.scala#L111) that is testing that the query in https://github.com/quasar-analytics/quasar/issues/2755 produces the expected `LogicalPlan`?

edmundnoble:
@alissapajer Review should be addressed, added the SQL^2 query test as well as adding back div-by-zero tests, after adding which I found that the simplifier was throwing `ArithmeticException` rather than divide by zero. Fixed that too.

@djspiewak Any idea why appveyor is refusing to build?

alissapajer:
@edmundnoble looks like there are merge conflicts to resolve as well.

alissapajer:
@edmundnoble also, could you add an integration test for the query in #2755 as well? The initial bug report was against mongo, so we should verify this query works against mongo (and other connectors as well).

edmundnoble:
Ah I'll resolve those conflicts @alissapajer, that's probably why it won't build. I'm wondering if adding an integration test is worth it here though, because the bug in the report is that the original query compiled and it shouldn't.

alissapajer:
@edmundnoble shouldn't the original query compile? A `substring` will only exist when the result of the division returns an integer, but that's a possibility.

edmundnoble:
So we don't have an auto-coercion semantic, from what I can tell, that lets you pass a `Dec` to a function expecting an `Int` and get a new `Typecheck` node (or `trunc` call) inserted. That's just not a feature we have. I talked to @jdegoes about it, seeing as this is how Mongo works and not how Postgres works for example, and he said to do whichever is simpler but Postgres semantics are good as a general rule; so I was thinking we should keep the Postgres semantics (i.e. the user needs to make their own `trunc` call here). Even if we do change our minds here, this is currently consistent with how every other function works.

alissapajer:
@edmundnoble :+1: makes sense. Should we add a test showing `select substring("abcdefg", 0, trunc(pop / 10000)) from zips` failing to compile then? This should fail to compile in quasar, and not error in monogo, like it was doing before.

edmundnoble:
@alissapajer Done, I made it test specifically that it doesn't type check even.

PR #3018, opened 2017-11-03
QSU Graduation

alissapajer:
Completes implementation of `graduate`, except for `Nullary` and `Distinct`.

In particular implements
```scala
def mergeSources(left: QSUGraph, right: QSUGraph): (QSUGraph, FreeQS, FreeQS)
```

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
@@ -57,7 +57,7 @@ sealed abstract class Graduate[T[_[_]]: CorecursiveT: ShowT] extends QSUTTypes[T
   private type QSU[A] = QScriptUniform[A]
 
   private def mergeSources(left: QSUGraph, right: QSUGraph): SrcMerge[QSUGraph, FreeQS] = {
-    val source: QSUGraph = MergeSources.merge(left, right)
+    val source: QSUGraph = MergeSources.merge(left.vertices, right.vertices)
     SrcMerge(source, graduateCoEnv(source.root, left), graduateCoEnv(source.root, right))
   }
 
@@ -95,9 +95,10 @@ sealed abstract class Graduate[T[_[_]]: CorecursiveT: ShowT] extends QSUTTypes[T
       }
     }
 
-    def merge(left: QSUGraph, right: QSUGraph): QSUGraph = {
-      val edges: ISet[Edge] = findEdges(left.vertices) intersection findEdges(right.vertices)
-      QSUGraph[T](edgesToRoot(edges), left.vertices)
+    def merge(left: SMap[Symbol, QSU[Symbol]], right: SMap[Symbol, QSU[Symbol]])
+        : QSUGraph = {
+      val edges: ISet[Edge] = findEdges(left) intersection findEdges(right)
+      QSUGraph[T](edgesToRoot(edges), left)
     }
   }
 
--- a/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
@@ -63,9 +63,9 @@ sealed abstract class Graduate[T[_[_]]: CorecursiveT: ShowT] extends QSUTTypes[T
 
   object MergeSources {
 
-    case class Edge(from: Symbol, to: Symbol)
+    private case class Edge(from: Symbol, to: Symbol)
 
-    object Edge {
+    private object Edge {
       implicit val order: Order[Edge] = Order.order {
         case (e1, e2) =>
           Order[(Symbol, Symbol)].order((e1.from, e1.to), (e2.from, e2.to))
@@ -73,7 +73,7 @@ sealed abstract class Graduate[T[_[_]]: CorecursiveT: ShowT] extends QSUTTypes[T
     }
 
     // (QSUGraph, QSUPattern[T, ISet[Edge]]) => ISet[Edge]
-    val findEdgesƒ: ElgotAlgebra[(QSUGraph, ?), QSUPattern[T, ?], ISet[Edge]] = {
+    private val findEdgesƒ: ElgotAlgebra[(QSUGraph, ?), QSUPattern[T, ?], ISet[Edge]] = {
       case (QSUGraph(root, vertices), QSUPattern(_, qsu)) =>
         val previousEdges: ISet[Edge] = ISet.unions(qsu.toList)
         val newEdges: ISet[Edge] = ISet.fromList(vertices(root).toList.map(Edge(root, _)))
--- a/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
@@ -54,14 +54,14 @@ sealed abstract class Graduate[T[_[_]]: CorecursiveT: ShowT] extends QSUTTypes[T
   import QSUPattern._
 
   type QSE[A] = QScriptEducated[A]
-  type QSU[A] = QScriptUniform[A]
+  private type QSU[A] = QScriptUniform[A]
 
   private def mergeSources(left: QSUGraph, right: QSUGraph): SrcMerge[QSUGraph, FreeQS] = {
     val source: QSUGraph = MergeSources.merge(left, right)
     SrcMerge(source, graduateCoEnv(source.root, left), graduateCoEnv(source.root, right))
   }
 
-  object MergeSources {
+  private object MergeSources {
 
     private case class Edge(from: Symbol, to: Symbol)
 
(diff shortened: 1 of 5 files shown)


connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -40,62 +44,148 @@ import quasar.qscript.qsu.{QScriptUniform => QSU}
...
-    Recursive[QSUGraph, QSUPattern[T, ?]].project(_) match {
-      case QSUPattern(root, qsu) => qsu match {
+  private def mergeSources(left: QSUGraph, right: QSUGraph): SrcMerge[QSUGraph, FreeQS] = {
+    val source: QSUGraph = MergeSources.merge(left, right)
+    SrcMerge(source, graduateCoEnv(source.root, left), graduateCoEnv(source.root, right))
+  }
 
-        case QSU.Read(path) =>
-          Inject[Const[Read[AFile], ?], QScriptEducated].inj(Const(Read(path)))
+  private object MergeSources {
```
alissapajer: I'd like to be able to unit test the individual functions in here, so we'll have to refactor this a bit.

Some of this algorithm is just set manipulation (sets of vertices and edges), so we can test it separately in the context of a graph without having to consider anything qscript-related.

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -40,62 +44,148 @@ import quasar.qscript.qsu.{QScriptUniform => QSU}
...
+      implicit val order: Order[Edge] = Order.order {
+        case (e1, e2) =>
+          Order[(Symbol, Symbol)].order((e1.from, e1.to), (e2.from, e2.to))
+      }
+    }
 
-        case QSU.QSReduce(source, buckets, reducers, repair) =>
-          QCE(Reduce[T, QSUGraph](source, buckets, reducers, repair))
+    // (QSUGraph, QSUPattern[T, ISet[Edge]]) => ISet[Edge]
+    private val findEdgesƒ: ElgotAlgebra[(QSUGraph, ?), QSUPattern[T, ?], ISet[Edge]] = {
```
alissapajer: I have the intuition that maybe there is a simpler fold we could use here, because we don't care about the shape of the `QScriptUniform` at all and only care about the `ISet[Edge]`s contained in its recursive positions.

foundation/src/main/scala/quasar/fp/package.scala
```diff
@@ -246,6 +246,7 @@ package object fp
 
   implicit val symbolEqual: Equal[Symbol] = Equal.equalA
   implicit val symbolShow: Show[Symbol] = Show.showFromToString
+  implicit val symbolOrder: Order[Symbol] = Order.order((s1, s2) => Order[String].order(s1.name, s2.name))
```
alissapajer: I haven't actually confirmed that this is valid.

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
wemrysi: How about `vertices.toList foldMap { case (name, node) => node foldMap (ISet(Edge(name, _)))) }`?

alissapajer: That is... much simpler. Thanks. The `ElgotAlgebra` was fun though! :stuck_out_tongue:

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -41,61 +45,142 @@ import quasar.qscript.qsu.QSUGraph.QSUPattern
...
+    private object Edge {
+      implicit val order: Order[Edge] = Order.order {
+        case (e1, e2) =>
+          Order[(Symbol, Symbol)].order((e1.from, e1.to), (e2.from, e2.to))
+      }
+    }
 
-        case QSU.QSReduce(source, buckets, reducers, repair) =>
-          QCE(Reduce[T, QSUGraph](source, buckets, reducers, repair))
+    private def findEdges(vertices: SMap[Symbol, QSU[Symbol]])
```
alissapajer: We can just require `[F[_]: Foldable](SMap[Symbol, F[Symbol]])` here now!

alissapajer:
Ok awesome. I just pushed a change so that all of `MergeSources` is written in terms of `Symbol`. So we can test this independently of anything qscript-related. It makes sense to take this out of `Graduate`.

connector/src/main/scala/quasar/qscript/qsu/IntersectGraphs.scala
```diff
@@ -0,0 +1,63 @@
...
+
+package quasar.qscript.qsu
+
+import slamdata.Predef.{Map => SMap, _}
+
+import quasar.fp._
+
+import scalaz._, Scalaz._
+
+// TODO move this to a different package, maybe foundation
```
alissapajer: I just didn't want to wait for scalac, so it's here for now.

connector/src/main/scala/quasar/qscript/qsu/IntersectGraphs.scala
```diff
@@ -0,0 +1,63 @@
...
+    rootCandidates.toIList match {
+      case ICons(head, INil()) => head
+      // FIXME real error handling
+      case _ => scala.sys.error(s"Source merging failed. Candidates: ${Show[ISet[Symbol]].shows(rootCandidates)}")
+    }
+  }
+
+  def intersect[F[_]: Foldable](left: SMap[Symbol, F[Symbol]], right: SMap[Symbol, F[Symbol]])
+      : Symbol =
+    edgesToRoot(findEdges[F](left) intersection findEdges[F](right))
```
alissapajer: If we need to performance-tune this algorithm in the future, we'll be doing so in an isolated context. This was the easiest algorithm to implement quickly, but at some point we can do some research into the best ways to find the greatest common subgraph. Also, as I think about it now, I'm not convinced there is a unique answer to this question, so that will be interesting to consider as well, especially in the context of the actual problem we're solving in qscript.

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -41,61 +45,104 @@ import quasar.qscript.qsu.QSUGraph.QSUPattern
...
-  private def mergeSources(left: QSUGraph, right: QSUGraph)
-      : SrcMerge[QSUGraph, FreeQS] =
-    slamdata.Predef.???
+  type QSE[A] = QScriptEducated[A]
+  private type QSU[A] = QScriptUniform[A]
 
-  val graduateƒ: Coalgebra[QScriptEducated, QSUGraph] =
-    Recursive[QSUGraph, QSUPattern[T, ?]].project(_) match {
-      case QSUPattern(root, qsu) => qsu match {
+  private def mergeSources(left: QSUGraph, right: QSUGraph): SrcMerge[QSUGraph, FreeQS] = {
```
alissapajer: Merging two graphs is reduced to the graph problem of intersecting graphs, and then graduating each of the sides in the context of their common source `Symbol`.

connector/src/main/scala/quasar/qscript/qsu/IntersectGraphs.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import scalaz._, Scalaz._
+
+// TODO move this to a different package, maybe foundation
+object IntersectGraphs {
+
+  private final case class Edge(from: Symbol, to: Symbol)
+
+  private object Edge {
+    implicit val order: Order[Edge] = Order.order {
```
wemrysi: ⛳️ `Order.orderBy(e => (e.from, e.to))`

connector/src/main/scala/quasar/qscript/qsu/IntersectGraphs.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private def findEdges[F[_]: Foldable](vertices: SMap[Symbol, F[Symbol]])
+      : ISet[Edge] =
+    vertices.toList foldMap {
+      case (from, node) => node.foldMap(to => ISet.singleton(Edge(from, to)))
+    }
+
+  private def edgesToRoot(edges: ISet[Edge]): Symbol = {
+    val (froms, tos): (ISet[Symbol], ISet[Symbol]) =
```
wemrysi: ⛳️ `edges foldMap (e => (ISet.singleton(e.from), ISet.singleton(e.to)))`

foundation/src/main/scala/quasar/fp/package.scala
wemrysi: I don't see why it wouldn't be 👍 Also, you'll want to get rid of the `Equal` instance to avoid ambiguous implicits.

Lastly, ⛳️ `Order.orderBy(_.name)`.

alissapajer:
@wemrysi PR updated.

PR #3019, opened 2017-11-03
RDBMS table model

kciesielski:
This PR brings a notion of `Table Model`. We now don't store everything in single-json-column tables, but we try to determine the best column types to represent data. `RdbmsWriteFile` scans the incoming `Vector[Data]` and incrementally creates columns. For simple elements like `Data.Int`, `Data.String` etc. we create appropriate DB fields. For nested `Data.Obj`, arrays and such, we use a `jsonb` column.
More details available here: https://docs.google.com/document/d/1loSdaInrdMaFg08EHqaLa2oxWP6wZziN2-uzLenJCuA/edit?usp=sharing

More column type <-> quasar data type mappings will be added later.

### DIFF
--- a/rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresInsert.scala
+++ b/rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresInsert.scala
@@ -21,27 +21,58 @@ import quasar.Data
 import quasar.fs.FileSystemError
 import quasar.physical.rdbms.common.TablePath
 import quasar.physical.rdbms.fs.RdbmsInsert
+import quasar.physical.rdbms.model.{ColumnDesc, ColumnarTable, JsonTable, TableModel}
 
-import doobie.free.connection.ConnectionIO
-import doobie.imports.{Meta, Update}
-import doobie.syntax.string._
-import doobie.util.fragment.Fragment
-import scalaz.std.vector._
-import scalaz.syntax.show._
+import doobie.imports._
+import scalaz._
+import Scalaz._
 
 trait PostgresInsert extends RdbmsInsert {
 
   implicit def dataMeta: Meta[Data]
 
+  def toColValues(cols: Set[ColumnDesc])(row: Data): \/[FileSystemError, List[String]] = {
+    row match {
+      case Data.Obj(fields) =>
+        (fields.toList.filter(f => cols.exists(_.name === f._1)).sortBy(_._1).map {
+          case (_, v) =>
+            v match {
+              case Data.Obj(innerJs) => "{TODO decode this to JS string}"// TODO
+              case Data.Int(num) => s"$num"
+              case Data.Str(txt) => s"'$txt'"
+              case _ => s"TODO" // TODO
+            }
+        }).right
+      case _ => FileSystemError.unsupportedOperation(s"Unexpected data row $row not matching model $cols").left
+    }
+  }
+
   def batchInsert(
       dbPath: TablePath,
-      chunk: Vector[Data]
+      chunk: Vector[Data],
+      model: TableModel
   ): ConnectionIO[Vector[FileSystemError]] = {
 
-    val fQuery = fr"insert into " ++ Fragment.const(dbPath.shows) ++ fr"(data) values(?::JSON)"
+    val insertIntoTable = fr"insert into " ++ Fragment.const(dbPath.shows)
 
-    Update[Data](fQuery.update.sql)
-      .updateMany(chunk)
-      .map(_ => Vector.empty[FileSystemError])
+    model match {
+      case JsonTable =>
+        val fQuery = fr"(data) values(?::JSON)"
+        Update[Data](fQuery.update.sql)
+          .updateMany(chunk)
+          .map(_ => Vector.empty[FileSystemError])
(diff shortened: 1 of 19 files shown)


kciesielski:
@drostron @rabbitonweb The PR is ready for review. There was a failing test with couchbase, probably requiring restarting the build.

drostron:
@kciesielski 👌🏾, review commencing.

drostron:
I've been pulled away by a few other issues. I'll be able to offer a review today or tomorrow.

rdbms/src/main/scala/quasar/physical/rdbms/model/TableModel.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  implicit val columnTypeEq: Equal[ColumnType] = Equal.equalA
+
+  /**
+    * When updating current table model with new model, the resulting model
+    * can be either extended by new columns, or just a JsonTable. The latter
+    * means that current and new models cannot produce a new definition which
+    * reconciles data types for both (data is heterogenous).
+    */
+  implicit val tableModelMonoid: Monoid[TableModel] = new Monoid[TableModel] {
```
drostron: Would be nice to see laws checking for `Monoid[TableModel]` similar to https://github.com/quasar-analytics/quasar/blob/67ac837200e47e4554f97d610525db2d06c4bc36/foundation/src/test/scala/quasar/fp/numeric/SampleStatsSpec.scala#L61

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/PostgresInsert.scala
```diff
@@ -21,27 +21,79 @@ import quasar.Data
...
-import scalaz.syntax.show._
+import scalaz._
+import Scalaz._
 
 trait PostgresInsert extends RdbmsInsert {
 
   implicit def dataMeta: Meta[Data]
 
+  def toColValues(cols: Set[ColumnDesc])(
+      row: Data)(implicit formatter: DataFormatter): \/[FileSystemError, Map[String, String]] = {
```
drostron: `\/` is typically infix. Here and a few others.

rdbms/src/main/scala/quasar/physical/rdbms/model/TableModel.scala
```diff
@@ -0,0 +1,166 @@
...
+case object IntCol extends ColumnType
+case object NullCol extends ColumnType
+
+final case class ColumnDesc(name: String, tpe: ColumnType)
+
+sealed trait TableModel
+
+case object JsonTable extends TableModel
+
+final case class ColumnarTable(columns: Set[ColumnDesc])
```
drostron: Why not preserve ordering on columns?

drostron: nm

rdbms/src/main/scala/quasar/physical/rdbms/model/TableModel.scala
kciesielski: Thanks, I didn't know about such utils. I'm adding law checking for TableModel.

kciesielski:
@drostron PR updated according to your remarks. Failed builds seem to be caused by some non-related issues, please bump the build.

rdbms/src/main/scala/quasar/physical/rdbms/model/TableModel.scala
```diff
@@ -76,11 +76,25 @@ object TableModel {
...
   }
 
+  implicit val columnDescEq: Equal[ColumnDesc] = Equal.equalBy(_.name)
+
   implicit val columnDescOrdering: Ordering[ColumnDesc] =
     Ordering.fromLessThan[ColumnDesc](_.name < _.name)
 
   implicit val columnTypeEq: Equal[ColumnType] = Equal.equalA
 
+  implicit val tableModelEq: Equal[TableModel] = new Equal[TableModel] {
```
drostron: There's also the shorter
```scala
 Equal.equal {
    case (JsonTable, JsonTable) => true
    case (ColumnarTable(c1), ColumnarTable(c2)) => c1.toList === c2.toList
    case _ => false
  }
```

drostron: Would be nice to also have `checkAll(propz.equal.laws[TableModel])` now that there's an `Equal` instance

kciesielski:
@drostron Equality updated and tests added.

rdbms/src/test/scala/quasar/physical/rdbms/model/TableModelTest.scala
```diff
@@ -63,6 +63,7 @@ class TableModelTest extends Spec  with ScalazMatchers {
 
   "Table Model Monoid" >> {
     checkAll(propz.monoid.laws[TableModel])
+    checkAll(propz.equal.laws[TableModel])
```
drostron: I'd pull this out of "Table Model Monoid" since it's not related to `Monoid`. The `checkAll(...)` can be on its own outside of a spec block.

kciesielski:
@drostron I pulled all checks to the top-level in tests, also added "must" literal where needed to make test output correct.

PR #3023, opened 2017-11-03
ApplyProvenance and MappableRegion for QSU

wemrysi:
Adds a function to compute provenance for `QSUGraph` and remove `DimEdit` nodes in the process. Also provides `MappableRegion` which attempts to express as much of a graph as possible in terms of `MapFunc`.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/ApplyProvenance.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ApplyProvenance.scala
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript.qsu
+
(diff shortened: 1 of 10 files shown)


connector/src/main/scala/quasar/qscript/qsu/ApplyProvenance.scala
```diff
@@ -0,0 +1,144 @@
...
+      case DimEdit(src, DTrans.Squash()) => dims.squash(src).point[F]
+
+      case DimEdit(src, DTrans.Group(k)) =>
+        dims.swap(0, 1, dims.lshift(k as gpf.root, src)).point[F]
+
+      case Distinct(src) => src.point[F]
+
+      case GroupBy(_, _) => unexpectedError("GroupBy", gpf.root)
+
+      case JoinSideRef(_) => dims.empty.point[F]
```
alissapajer: I think this one should be an `unexpectedError` because all `JoinSideRef` should be removed when we transform `LPJoin` to `ThetaJoin`, which happens before we compute provenance.

wemrysi:
@alissapajer Ready for another look.

foundation/src/main/scala/quasar/contrib/scalaz/MonadState_.scala
```diff
@@ -18,19 +18,21 @@ package quasar.contrib.scalaz
...
-
-  def bind[A, B](fa: F[A])(f: (A) ⇒ F[B]): F[B]
+trait MonadState_[F[_], S] {
   def get: F[S]
   def init: F[S]
-  def point[A](a: ⇒ A): F[A]
   def put(s: S): F[Unit]
+
+  def gets[A](f: S => A)(implicit F: Functor[F]): F[A] =
+    F.map(init)(f)
```
djspiewak: What is `init` and why isn't it `get`?

wemrysi: An artifact copied from scalaz, removed.

PR #3024, opened 2017-11-03
eliminate nested SpliceObjects

### LINKED ISSUE #2841: Normalize JsCore SpliceObjects
Consider this query, its workflow and QScript:

Query:
```sql
select *, city as city2, pop as pop2 from zips
```

QScript:
```
QScript (Mongo-specific):
Map
├─ ShiftedRead(/jps/zips, ExcludeId)
╰─ ConcatMaps
   ├─ ConcatMaps
   │  ├─ ○
   │  ╰─ MakeMap
   │     ├─ Constant(Str(city2))
   │     ╰─ ProjectField
   │        ├─ ○
   │        ╰─ Constant(Str(city))
   ╰─ MakeMap
      ├─ Constant(Str(pop2))
      ╰─ ProjectField
         ├─ ○
         ╰─ Constant(Str(pop))
```

Workflow:

```
Chain
├─ $ReadF(jps; zips)
╰─ $SimpleMapF
   ├─ Map
   │  ╰─ SpliceObjects
   │     ├─ SpliceObjects
   │     │  ├─ Ident(_)
   │     │  ╰─ Obj
   │     │     ╰─ Key(city2: _.city)
   │     ╰─ Obj
   │        ╰─ Key(pop2: _.pop)
   ╰─ Scope(Map())
```

As more aliases are added to the `select`, the more nested `SpliceObjects` we get. This leads to quadratic (or worse) slow down for each field alias added 

As an example, I haven't seen the following query return at all:

```sql
select *, city as city2, city as city3, city as city4, pop as pop2, pop as pop3, pop as pop4, pop as pop5 from zips
```

*EDIT*: query returned after 34 minutes (!)

The workflow above can be rewritten into a single call to `SpliceObjects` to avoid this.

```
Workflow (raw):
Chain
├─ $ReadF(jps; zips)
╰─ $SimpleMapF
   ├─ Map
   │  ╰─ SpliceObjects
   │     ├─ Ident(_)
   │     ├─ Obj
   │     │  ╰─ Key(city2: _.city)
   │     ╰─ Obj
   │        ╰─ Key(pop2: _.pop)
   ╰─ Scope(Map())
```

Something similar to this should the trick:

```scala
{
  case SpliceObjectsF(objs) =>
    objs flatMap {
      case Embed(SpliceObjectsF(xs)) => xs
      case other => List(other)
    }
}
```

(snippet from @wemrysi)

nicflores:
Improve query times containing multiple aliases.

Fixes #2841


nicflores:
@rintcius, good to know. I'll trash those files and push the changes.

js/src/main/scala/quasar/jscore/package.scala
```diff
@@ -98,6 +99,14 @@ package object jscore {
       case (ObjF(values), LiteralF(Js.Str(name))) => values.get(Name(name)).map(_.project)
       case (_,            _)                      => None
     }
+    case SpliceObjectsF(l) => {
```
jsantos17: Would you mind adding a test for this? Maybe check some edge cases like `SpliceObjects(SomeMap, SpliceObjects(..))` and `SpliceObjects(SpliceObjects(..), SomeMap)`. It's fine if you don't handle anything but the exact case we want to handle. `JsCoreSpecs` has other tests for this. Otherwise looks good.

That is, SpliceObjects(SomeMap, SpliceObjects(..)) does not need to be simplified as `SpliceObjects` is the second argument, and not the first.

jsantos17: You don't need the brackets as this is a single expression.

mongodb/src/test/scala/quasar/physical/mongodb/planner.scala
```diff
@@ -1272,29 +1272,29 @@ class PlannerSpec extends
...
+              obj(sigil.Quasar -> SpliceObjects(List(
                 ident("x"),
                 obj(
                   "1" -> jscore.Literal(Js.Str("1"))),
                 obj(
-                  "2" -> jscore.Literal(Js.Str("2")))))))),
+                  "2" -> jscore.Literal(Js.Str("2"))))))))),
             ListMap())))
-    }.pendingWithActual("#2841", testFile("plan select with wildcard and two constants"))
+    }
```
jsantos17: 🎉

nicflores:
@jsantos17, ready for another look. Let me know what you think of that nested splice object. I tried modeling it after your example query in #2841.

jsantos17:
@nicflores needs merge from master. Any idea why Travis didn't run the build for this?

js/src/test/scala/quasar/jscore/jscore.scala
```diff
@@ -122,6 +122,36 @@ class JsCoreSpecs extends quasar.Qspec with TreeMatchers with ScalazMatchers {
 
       x.simplify must_=== Arr(List(ident("y"), ident("y"), If(ident("r"), Select(ident("r"), "foo"), ident("bar"))))
     }
+
+    "reduce multiple splice obj to single splice obj" in {
+      val anif = If(
+                   BinOp(And,
+                    Call(ident("isObject"),List(ident("value"))),
+                    UnOp(Not,Call(Access(ident("Array"),Literal(Js.Str("isArray"))),List(ident("value"))))
```
jsantos17: Can we merge the dangling parentheses with the previous line?

rintcius:
Happy once Travis is green.

@jsantos17 I think Travis doesn't run when it has conflicting files at the moment the PR is created

PR #3028, opened 2017-11-06
qscript-based mongo planner spec

### LINKED ISSUE #3015: Add join tests for Mongo using the QScript DSL
The largest amount of disabled tests for #2667 is related to join planning. There is some rather straightforward QScript that still results in unnecessarily complex physical plans, which makes me believe we have Mongo-specific issues planning joins. For example, consider the simple join in this query:

```sql
select zips2.city from zips join zips2 on zips.`_id` = zips2.`_id`
```

The QScript for the query:

```
EquiJoin
├─ Unreferenced
├─ Filter
│  ├─ ShiftedRead(/jps/zips, IncludeId)
│  ╰─ Guard
│     ├─ ProjectIndex
│     │  ├─ ○
│     │  ╰─ Constant(Int(1))
│     ├─ Type(Obj(Map[], Some(Top)))
│     ├─ Constant(Bool(true))
│     ╰─ Constant(Bool(false))
├─ Filter
│  ├─ ShiftedRead(/jps/zips2, IncludeId)
│  ╰─ Guard
│     ├─ ProjectIndex
│     │  ├─ ○
│     │  ╰─ Constant(Int(1))
│     ├─ Type(Obj(Map[], Some(Top)))
│     ├─ Constant(Bool(true))
│     ╰─ Constant(Bool(false))
├─ List
│  ╰─ tuple
│     ├─ ProjectKey
│     │  ├─ ProjectIndex
│     │  │  ├─ ○
│     │  │  ╰─ Constant(Int(1))
│     │  ╰─ Constant(Str(_id))
│     ╰─ ProjectKey
│        ├─ ProjectIndex
│        │  ├─ ○
│        │  ╰─ Constant(Int(1))
│        ╰─ Constant(Str(_id))
├─ Inner
╰─ ProjectKey
   ├─ ProjectIndex
   │  ├─ RightSide
   │  ╰─ Constant(Int(1))
   ╰─ Constant(Str(city))
```

I don't see any way we can simplify this QScript further, but new Mongo's plan still needs mapReduce, unlike old Mongo's plan that can plan this entirely in aggregation. This is related to #2878

rintcius:
* rename `planner` -> `PlannerSpec`
* move mongo planner helpers into file separate from `PlannerSpec`
* add `PlannerQScriptSpec` that allows mongo planner testing based on qscript
* example qscript-based spec

Fixes #3015


mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+        $lookup(
+          CollectionName("smallZips"),
+          JoinHandler.LeftName \ BsonField.Name("_id"),
+          BsonField.Name("_id"),
+          JoinHandler.RightName),
+        $unwind(DocField(JoinHandler.RightName)),
+        $project(
+          reshape(sigil.Quasar -> $field(JoinDir.Right.name, "city")),
+          ExcludeId)))
+    }
```
rintcius: Note that this is `onPar` now!

mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  import fixExprOp._
+  import PlannerHelpers._
+  import CollectionUtil._
+  import Reshape.reshape
+
+  val constr =
+    quasar.qscript.construction.mkDefaults[Fix, fs.MongoQScript[Fix, ?]]
+
+  import constr._1._, constr._2.{Hole => _, _}, constr.{_3 => fix}
```
rintcius: @edmundnoble maybe nice if we can import `_1`, `_2` and `_3` using a name instead (eg resp. `func`, `free` and `fix`)?

mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  //TODO make this independent of MongoQScript and move to a place where all
+  //     connector tests can refer to it
+  val simpleJoin: Fix[fs.MongoQScript[Fix, ?]] =
+    fix.EquiJoin(
+      fix.Unreferenced,
+      Filter(
+        ShiftedRead[AFile](rootDir </> dir("db") </> file("zips"), qscript.ExcludeId),
+        Guard(Hole, Type.Obj(ScalaMap(),Some(Type.Top)), Constant(bool[Fix](true)), Constant(bool[Fix](false)))),
```
jsantos17: I think you can replace `Type.Obj(ScalaMap(),Some(Type.Top))` with `Type.AnyObject`.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
edmundnoble: You shouldn't be able to import all of the members of these without a bunch of conflicts: it's intended to be used with the prefixes, `func` `free` and `fix`. `free` and `fix` have a bunch of methods with the same names, they just put `QScript` into `Free` or `Fix`. The way I set them up is:
`val (func, free, fix) = quasar.qscript.construction.mkDefaults[FIx, fs.MongoQScript[Fix, ?]]`

rintcius:
@jsantos17 updated

djspiewak:
Version label.  :-P

PR #3031, opened 2017-11-06
Initial work on provenance reification for autojoin, transpose, reduce.

alissapajer:
`AutoJoin` of size 2 defaults to `ThetaJoin`. `AutoJoin` of greater sizes not yet supported.

`LPReduce` requires an adjusted impl of [`reifiedIdArray`](https://github.com/wemrysi/quasar/blob/wip/proper-prov/connector/src/main/scala/quasar/qscript/QProv.scala#L68) to compute its `buckets`.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript.qsu
+
+import slamdata.Predef._
+
+import quasar.common.JoinType
+import quasar.contrib.scalaz.MonadError_
+import quasar.fp.ski.κ
+import quasar.qscript.{HoleF, IncludeId, LeftSideF, MFC, ReduceIndexF, RightSideF}
+import quasar.qscript.ReduceFunc._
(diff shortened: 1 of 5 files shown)


connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
```diff
@@ -0,0 +1,95 @@
...
+    case QSU.LPReduce(source, reduce) =>
+      val bucket: FreeMap = slamdata.Predef.??? // TODO computed from provenance
+      val qsu: QSU[Symbol] = QSU.QSReduce[T, Symbol](source, List(bucket), List(reduce.map(κ(HoleF))), ReduceIndexF(\/-(0)))
+
+      qsu.point[F]
+
+    case qsu => qsu.point[F]
+  }
+
+  def reifyProvenance[F[_]: Monad: ErrorM](qsuF: F[AuthenticatedQSU[T]])
```
wemrysi: This could just accept an `AuthenticatedQSU`, no? Any reason it needs to be in `F[_]`?

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
```diff
@@ -0,0 +1,95 @@
...
+  type QSU[A] = QScriptUniform[A]
+
+  type ErrorM[F[_]] = MonadError_[F, String]
+  def ErrorM[F[_]](implicit ev: ErrorM[F]): ErrorM[F] = ev
+
+  val prov = new QProv[T]
+
+  private def toQScript[F[_]: Applicative: ErrorM](dims: QSUDims[T])
+      : QSU[Symbol] => F[QSU[Symbol]] = {
+    case QSU.AutoJoin(sources, combiner0) =>
```
wemrysi: I think having `AutoJoin` and `AutoJoin3` would be nice to avoid the partiality here. If we want to add that to the TODO list, that's fine.

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
alissapajer: I came to the same conclusion. That's already changed in my PR over here https://github.com/quasar-analytics/quasar/pull/3033/files#diff-df1739caf23238d032bd02033b296d13R80

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
alissapajer: Let's add that to the TODO list.

PR #3035, opened 2017-11-07
Fix `data` endpoint to return 404 Notfound if file does not exist

### LINKED ISSUE #3038: Data endpoint returns empty response instead of `404` if file does exist

jedesah:
Just wanted to document this somewhat surprising (IMHO) behavior and perhaps call attention to it if we want to change it.

Personally, I would go with a 404 `NotFound` response in that case but perhaps this was designed this way initially for a particular reason.

At the very least we should document it I think, but if enough people think we should change it to 404 `NotFound`, then I can go ahead and do that too

Fixes #3038

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,6 +647,8 @@ Retrieves data from the specified path in the [format](#data-formats) specified 
 
 If the supplied path represents a directory (ends with a slash), this request produces a `zip` archive containing the contents of the named directory, database, etc. Each file in the archive is formatted as specified in the request query and/or `Accept` header.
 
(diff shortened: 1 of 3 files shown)


wemrysi:
@jedesah @rintcius Yeah, it should be a `404`. It was originally because a `Read` of a nonexistent file would error with `PathNotFound` but this semantic was changed due to not all connectors being able to differentiate between an empty file and a nonexistent one and so I think the `200` is a consequence of that change.

We'll have to explicitly check if the path exists rather than relying on the error response of read.

jedesah:
Fixed!

PR #3037, opened 2017-11-07
Implemented rewrite for recognizing and specializing Distinct

djspiewak:
Depends on #3036 

I actually didn't know that the `Distinct` encoding was done in the compiler (before LP!).  That's somewhat concerning, because it means `graduate` is going to have to re-encode `Distinct` back into the reduce form in such a way that connectors like mongo and mimir can get it back out again.  This would be much easier if we had a `Distinct` node in `QScriptTotal`… :-(

Anyway, it's done, and more importantly this demonstrates how to add trivial little rewrites to the graph, which we expect to be doing a lot of.


connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -62,13 +62,89 @@ final case class QSUGraph[T[_[_]]](
...
+   * g into f(g).  When this happens, you must internally generate a
+   * new symbol for g and rewrite your subgraph to point to that new
+   * g, then the composite graph f(g) must be given the original symbol.
+   * As a safety check, because this case is just entirely unsound, the
+   * symbol you return at the root is entirely ignored in favor of the
+   * original root at that locus.
+   */
+  def rewriteM[F[_]: Monad](pf: PartialFunction[QSUGraph[T], F[QSUGraph[T]]]): F[QSUGraph[T]] = {
+    type G[A] = StateT[F, Set[Symbol], A]
+    val MS = MonadState[G, Set[Symbol]]
```
alissapajer: Is there a reason you are using `scalaz.MonadState` rather than our own `MonadState_`?

djspiewak: Because in this case, there’s no difference and scalaz’s is one fewer character. :-)

connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -62,6 +62,61 @@ final case class QSUGraph[T[_[_]]](
       this
   }
 
+  def overwriteAtRoot(qsu: QScriptUniform[T, Symbol]): QSUGraph[T] =
```
wemrysi: NBD, but how about `replaceRoot`?

djspiewak: I decided against any variant of `replace` because we already have a function named `replace`, and it does something very different from this one.

PR #3042, opened 2017-11-07
DSL for constructing QScriptUniform trees

wemrysi:


connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -232,4 +236,226 @@ object QScriptUniform {
...
+      iso composePrism O.distinct
+
+    val groupBy: Prism[QSU, (QSU, QSU)] =
+      iso composePrism O.groupBy
+
+    val joinSideRef: Prism[QSU, Symbol] =
+      iso composePrism O.joinSideRef
+
+    val leftShift: Prism[QSU, (QSU, FreeMap, IdStatus, JoinFunc)] =
+      iso composePrism O.leftShift
```
djspiewak: I'd like to see some convenience alternative functions wherever we have `FreeMap` to allow us to more conveniently construct them in single- or zero-function cases, but these are easy to add as we see fit.

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -232,4 +236,226 @@ object QScriptUniform {
...
+      iso composePrism O.thetaJoin
+
+    val transpose: Prism[QSU, (QSU, Rotation)] =
+      iso composePrism O.transpose
+
+    def tread(file: AFile): QSU =
+      transpose(read(file), Rotation.ShiftMap)
+
+    def tread1(name: String): QSU =
+      tread(Path.rootDir </> Path.file(name))
```
djspiewak: 😍

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
wemrysi: Yeah, I figured doing that made more sense than trying to anticipate all of the useful cases ahead of time 👍

PR #3043, opened 2017-11-07
Fix a couple tests that were failing (but not detected because marked as flaky)

jedesah:


jedesah:
@alissapajer I have no reason to suspect they are no longer flaky. Unfortunately, pretty much any property that limits the amount of examples is at a much higher likelihood of failing due to failed generation unless the generator is really "tight" (doesn't use very many filters).

PR #3049, opened 2017-11-08
Nullary and constant handling for QSU

### LINKED ISSUE #3041: Graduate `QSU.Nullary`
The `QSU.Nullary` node needs to `graduate` to qscript.

alissapajer:
Closes #3041


connector/src/test/scala/quasar/qscript/qsu/ReadLPSpec.scala
```diff
@@ -319,7 +317,8 @@ object ReadLPSpec extends Qspec with CompilerHelpers with DataArbitrary with QSU
 
   object DataConstant {
```
djspiewak: We really should move some of these extractors into a common location at some point, since they are pretty useful.  Either that or replicate them in the optics.

PR #3056, opened 2017-11-09
QScript DSL pretty-printer, with REPL integration (Fixes #3017)

### LINKED ISSUE #3017: QScript DSL Pretty-Printer
Name says it all: in go QScript trees, out comes construction code for those trees in the QScript construction DSL. Also, need a REPL option to print this out.

edmundnoble:
`set phaseFormat = code`, and everything which can be printed as code, is printed as code. Everything else is printed as an ordinary tree.

Maybe we want to suppress printing anything that's not code, but that would require more `phaseFormat` changes from the user while using the REPL. Maybe I'm also abusing `PhaseResult` here.

N.B. I had to remove the EJson DSL from the EJson object and put it in `construction`. There's no way to know the `T[_[_]]` at runtime, so there's no way to pass it as a type parameter back to the DSL constructors.

### DIFF
--- a/common/src/main/scala/quasar/common/PhaseResult.scala
+++ b/common/src/main/scala/quasar/common/PhaseResult.scala
@@ -17,38 +17,45 @@
 package quasar.common
 
 import slamdata.Predef._
-import quasar.{NonTerminal, RenderedTree, RenderTree, Terminal}, RenderTree.ops._
+import quasar.{DSLTree, NonTerminal, RenderedTree, RenderDSL, RenderTree, Terminal}
+import RenderTree.ops._, RenderDSL.ops._
 
 import argonaut._, Argonaut._
-import scalaz.Show
 import scalaz.syntax.show._
 
 sealed abstract class PhaseResult {
+  import PhaseResult._
   def name: String
+  def showTree: String = this match {
+    case Tree(name, value, _) => name + ":\n" + value.shows
+    case Detail(name, value) => name + ":\n" + value
+  }
+  def showCode: String = this match {
+    case Tree(name, _, Some(code)) => name + ":\n" + code.shows
+    case Tree(name, value, None) => name + ":\n" + value.shows
+    case Detail(name, value) => name + ":\n" + value
+  }
 }
 
 object PhaseResult {
-  final case class Tree(name: String, value: RenderedTree) extends PhaseResult
+  final case class Tree(name: String, value: RenderedTree, code: Option[DSLTree]) extends PhaseResult
   final case class Detail(name: String, value: String)     extends PhaseResult
 
   def tree[A: RenderTree](name: String, value: A): PhaseResult =
-    Tree(name, value.render)
+    Tree(name, value.render, None)
+  def treeAndCode[A: RenderTree: RenderDSL](name: String, value: A): PhaseResult =
+    Tree(name, value.render, Some(value.toDsl))
   def detail(name: String, value: String): PhaseResult = Detail(name, value)
-
-  implicit def show: Show[PhaseResult] = Show.shows {
-    case Tree(name, value)   => name + ":\n" + value.shows
-    case Detail(name, value) => name + ":\n" + value
-  }
 
   implicit def renderTree: RenderTree[PhaseResult] = new RenderTree[PhaseResult] {
     def render(v: PhaseResult) = v match {
-      case Tree(name, value)   => NonTerminal(List("PhaseResult"), Some(name), List(value))
-      case Detail(name, value) => NonTerminal(List("PhaseResult"), Some(name), List(Terminal(List("Detail"), Some(value))))
+      case Tree(name, value, _) => NonTerminal(List("PhaseResult"), Some(name), List(value))
+      case Detail(name, value)  => NonTerminal(List("PhaseResult"), Some(name), List(Terminal(List("Detail"), Some(value))))
     }
   }
 
   implicit def phaseResultEncodeJson: EncodeJson[PhaseResult] = EncodeJson {
-    case Tree(name, value)   => Json.obj("name" := name, "tree" := value)
-    case Detail(name, value) => Json.obj("name" := name, "detail" := value)
+    case Tree(name, value, _) => Json.obj("name" := name, "tree" := value)
+    case Detail(name, value)  => Json.obj("name" := name, "detail" := value)
   }
 }
--- a/connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
+++ b/connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
@@ -1,0 +1,453 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript
+
+import matryoshka.data.Fix
+import slamdata.Predef.{Eq => _, _}
+import matryoshka.{Hole => _, _}
+import matryoshka.implicits._
+import quasar.{Data, DSLTree, RenderDSL, Type, ejson}
+import quasar.contrib.pathy.{ADir, AFile}
+import quasar.ejson.EJson
+import quasar.fp._
+import quasar.fp.ski._
+import slamdata.Predef
+
+import scalaz.{Const, Coproduct, Free, Functor}
+import scalaz.syntax.bifunctor._
+import scalaz.syntax.either._
+import scalaz.syntax.show._
+import scalaz.syntax.std.tuple._
+import scalaz.std.anyVal._
+import scalaz.std.option._
+import scalaz.std.tuple._
+
+@SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+object RenderQScriptDSL {
+  type RenderQScriptDSL[A] = (String, A) => DSLTree
+  implicit def qscriptInstance[T[_[_]]: RecursiveT, F[_]: Functor]
+  (implicit I: Injectable.Aux[F, QScriptTotal[T, ?]]): RenderDSL[T[F]] =
+    new RenderDSL[T[F]] {
+      // hard-coded here to fix.
+      def toDsl(a: T[F]) = fixQSRender.apply("fix", a.transCata[Fix[QScriptTotal[T, ?]]](I.inject(_)))
+    }
+
+  def delayRenderQScriptDSL[T[_[_]]: RecursiveT, F[_]: Functor]
+  (D: Delay[RenderQScriptDSL, F]): RenderQScriptDSL[T[F]] = {
+    def toDsl(base: String, a: T[F]): DSLTree = D[T[F]](toDsl)("fix", a.project)
+    toDsl
+  }
+
+  def freeDelayRenderQScriptDSL[T[_[_]]: RecursiveT, F[_]: Functor, A]
+  (D: Delay[RenderQScriptDSL, F], A: RenderQScriptDSL[A]): RenderQScriptDSL[Free[F, A]] = {
+    def toDsl(base: String, a: Free[F, A]): DSLTree =
+      a.resume.fold[DSLTree](D[Free[F, A]](toDsl)("free", _), A("free", _))
+    toDsl
+  }
+
+  def delayRenderQScriptDSLFreeDelay[F[_]]
+  (implicit D: Delay[RenderQScriptDSL, F], F: Functor[F]): Delay[RenderQScriptDSL, Free[F, ?]] =
+    new Delay[RenderQScriptDSL, Free[F, ?]] {
+      def apply[A](A: RenderQScriptDSL[A]) = {
+        def toDsl(base: String, a: Free[F, A]): DSLTree =
+          a.resume.fold(D[Free[F, A]](toDsl)("free", _), A("free", _))
+        toDsl
+      }
+    }
+
+  def coproduct[F[_], G[_]]
+  (delF: Delay[RenderQScriptDSL, F], delG: Delay[RenderQScriptDSL, G]): Delay[RenderQScriptDSL, Coproduct[F, G, ?]] =
+    new Delay[RenderQScriptDSL, Coproduct[F, G, ?]] {
+      def apply[A](rec: RenderQScriptDSL[A]) = {
+        (base: Predef.String, a: Coproduct[F, G, A]) =>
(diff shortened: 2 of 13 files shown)


connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+import scalaz.std.option._
+import scalaz.std.tuple._
+
+@SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+object RenderQScriptDSL {
+  type RenderQScriptDSL[A] = (String, A) => DSLTree
+  implicit def qscriptInstance[T[_[_]]: RecursiveT, F[_]: Functor]
+  (implicit I: Injectable.Aux[F, QScriptTotal[T, ?]]): RenderDSL[T[F]] =
+    new RenderDSL[T[F]] {
+      // hard-coded here to fix.
```
alissapajer: Why are you hard-coding this to `Fix` here?

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+                  coproduct(readFileRenderDelay,
+                    deadEndRenderDelay))))))))
+
+  def freeQSRender[T[_[_]]: RecursiveT]: RenderQScriptDSL[FreeQS[T]] =
+    freeDelayRenderQScriptDSL[T, QScriptTotal[T, ?], Hole](qscriptTotalRenderDelay[T], holeRender("free"))
+
+  def fixQSRender[T[_[_]]: RecursiveT]: RenderQScriptDSL[Fix[QScriptTotal[T, ?]]] =
+    delayRenderQScriptDSL(qscriptTotalRenderDelay[T])
+
+  def qscriptCoreRenderDelay[T[_[_]]: RecursiveT]: Delay[RenderQScriptDSL, QScriptCore[T, ?]] =
```
alissapajer: This should be in the file QScriptCore.scala, and the analogous for all the other qscript component instances.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  def freeDelayRenderQScriptDSL[T[_[_]]: RecursiveT, F[_]: Functor, A]
+  (D: Delay[RenderQScriptDSL, F], A: RenderQScriptDSL[A]): RenderQScriptDSL[Free[F, A]] = {
+    def toDsl(base: String, a: Free[F, A]): DSLTree =
+      a.resume.fold[DSLTree](D[Free[F, A]](toDsl)("free", _), A("free", _))
+    toDsl
+  }
+
+  def delayRenderQScriptDSLFreeDelay[F[_]]
```
alissapajer: All of the non-qscript related instances can go in a `sealed abstract class RenderDSLInstances` in the file foundation/src/main/scala/quasar/RenderDSL.scala. See `RenderTree` for an example of this kind of organization.

common/src/main/scala/quasar/common/PhaseResult.scala
```diff
@@ -17,38 +17,45 @@
...
 import scalaz.syntax.show._
 
 sealed abstract class PhaseResult {
+  import PhaseResult._
   def name: String
+  def showTree: String = this match {
+    case Tree(name, value, _) => name + ":\n" + value.shows
+    case Detail(name, value) => name + ":\n" + value
+  }
+  def showCode: String = this match {
```
alissapajer: I'm uncomfortable with how this implementation forces effectively two different `Show` instances, but I don't have a good solution yet. I'll loop back to this shortly.

edmundnoble: I don't think there are effectively two different `Show` instances; there are zero `Show` instances here. What might be more extensible is adding a type parameter to `PhaseResult` for the tree or code contained inside, but that precludes printing anything that isn't code when `phaseFormat = code`.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
edmundnoble: There's no way for me to predict what the user will call their DSL objects except for `free` and `fix`.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
edmundnoble: I'll add in the `Delay`-derived and `coproduct` instances, not sure if `EJson` is downstream but I'll try for those too. Any others you had in mind?

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+          case ejson.Dec(value)  => ("dec", value.toString.left :: Nil)
+        })
+        DSLTree(base, label, children)
+    }
+  }
+
+  def eJsonRenderQScriptDSL[T[_[_]]](implicit T: RecursiveT[T]): RenderQScriptDSL[T[EJson]] =
+    delayRenderQScriptDSL(ejsonRenderQScriptDSLDelay)
+
+  def showType(ty: Type): String = ty match {
```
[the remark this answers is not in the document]
edmundnoble: That instance doesn't print types with `Type.` prefixing them.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+      "Type.FlexArr(" + min.shows + ", " + max.shows + ", "  + showType(mbrs) + ")"
+    case Type.Obj(assocs, unkns) =>
+      assocs.map { case (k, v) => ("\"" + k + "\"", showType(v)) }.mkString("Type.Obj(Map(", ", ", "), " +
+        unkns.fold("None")(t => "Some(" + showType(t) + ")") + ")")
+    case Type.Coproduct(l, r) =>
+      "Type.Coproduct(" + showType(l) + ", " + showType(r) + ")"
+    case x => "Type." + x.shows
+  }
+
+  def showData(data: Data): String = data match {
```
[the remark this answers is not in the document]
edmundnoble: Ditto for this one.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  def reduceIndexRender: RenderQScriptDSL[ReduceIndex] = {
+    (base: String, a: ReduceIndex) =>
+      val label = a.idx.fold(κ("Left"), κ("Right"))
+      val idx = a.idx.merge
+      DSLTree(base, label, idx.toString.left :: Nil)
+  }
+
+  def qscriptTotalRenderDelay[T[_[_]]: RecursiveT]: Delay[RenderQScriptDSL, QScriptTotal[T, ?]] =
```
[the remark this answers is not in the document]
edmundnoble: The reason I do this is because constructing an inductive `RenderDSL` for a QScript dialect implicitly will cause a (major, if `Injectable` is to be believed) compiler slowdown. If that isn't the case, I can change that.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+              DSLTree("", "List", key.map { case (k, v) => DSLTree("", "", freeMap(base, k).right :: freeMap(base, v).right :: Nil).right }).right ::
+              DSLTree("JoinType", f.shows, Nil).right ::
+              joinFunc(base, combine).right ::
+              Nil
+            DSLTree(base, "EquiJoin", args)
+        }
+      }
+    }
+
+  def readFileRenderDelay: Delay[RenderQScriptDSL, Const[Read[AFile], ?]] =
```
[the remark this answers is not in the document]
edmundnoble: Gotcha.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
edmundnoble: True, I'll do that.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,454 @@
...
+    case Data.NA => "Data.NA"
+    case Data.Null => "Data.Null"
+    case Data.Obj(o) => o.mapValues(showData).mkString("Data.Obj(", ", ", ")")
+    case Data.Set(s) => s.mkString("Data.Set(List(", ", ", "))")
+    case Data.Str(s) => "Data.Str(\"" + s + "\")"
+    case Data.Time(t) => "Data.Time(java.time.LocalTime.parse(\"" + t.toString + "\"))"
+    case Data.Timestamp(ts) => "Data.Timestamp(java.time.Instant.parse(\"" + ts.toString + "\"))"
+  }
+
+  def mapFuncRenderQScriptDSLDelay[T[_[_]]: RecursiveT]: Delay[RenderQScriptDSL, MapFunc[T, ?]] =
```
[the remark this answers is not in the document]
edmundnoble: Gotcha.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
edmundnoble: Actually, I'm not 100% sure on this. There's only one type class instance in this entire file, the rest of them aren't instances; does it make sense to put that functionality into the same file as the QScript components, when no users of those files want to use this functionality and it's only used in this file?

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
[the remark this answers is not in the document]
alissapajer: Can you call the `Show[Type]` instance and then prefix that result with `"Type."`?

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
alissapajer: Hm, I see your point. The various component instances for `RenderTree` are used in various places for debugging and error message printing. This dsl is used only for pretty printing in the REPL.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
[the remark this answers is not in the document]
edmundnoble: I can't, because a) `Type` values nest in `Type.Coproduct`, and b) I also need a `Data.` prefix for `Type.Const`.

alissapajer:
@edmundnoble I think it would be good to have some unit tests for this, since engineers want to be able to rely on its correctness. And once there's a unit test framework in place, that makes it easy to add more tests down the road if we find bugs. Tests also serve as documentation of what it should be doing, and if there were ever a change to our dsl, it would be nice if the unit tests catch that change immediately, rather than a user discovering it accidentally much later.

edmundnoble:
@alissapajer I'm wondering what form these tests would take. We could try to compile the code that comes out of the printer at runtime as a form of property test; we could just make sure that the printer always at least successfully runs on all QScript trees; we could come up with some QScript trees which are "representative" to test against.

I think runtime compilation is probably the most all-encompassing and useful test here, especially so we know it typechecks. I'm not sure how slow a test like that would be, but I'm willing to give it a shot today.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,440 @@
...
+            case Gte(a1, a2)          => ("Gte", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case IfUndefined(a1, a2)  => ("IfUndefined", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case And(a1, a2)          => ("fand", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case Or(a1, a2)           => ("Or", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case Within(a1, a2)       => ("Within", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case MakeMap(a1, a2)      => ("MakeMap", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case ConcatMaps(a1, a2)   => ("ConcatMaps", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case ProjectIndex(a1, a2) => ("ProjectIndex", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case ProjectKey(a1, a2)   => ("ProjectField", fa(base, a1).right :: fa(base, a2).right :: Nil)
+            case DeleteKey(a1, a2)    => ("DeleteField", fa(base, a1).right :: fa(base, a2).right :: Nil)
```
rintcius: `ProjectKey` and `DeleteKey`  have been renamed

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
```diff
@@ -0,0 +1,440 @@
...
+            case ConcatArrays(a1, a2) => ("Concatfarrays", fa(base, a1).right :: fa(base, a2).right :: Nil)
```
rintcius: typo

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
rintcius: What do you think of using https://github.com/lihaoyi/sourcecode#embedding-domain-specific-languages for this so that we don't have to maintain this mapping by hand? (and possibly https://github.com/lihaoyi/PPrint too for better output)

BTW it would also be great to have the name available separately for cases like this:
https://github.com/quasar-analytics/quasar/blob/master/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala#L493-L494

edmundnoble:
So I've written some tests, and caught a few bugs and a missing case in the DSL. There's runtime reflection to invoke the compiler, which is kind of scary, and that compiler doesn't have any of our compiler plugins, but it looks to work nicely.

edmundnoble:
I think this build failure isn't legitimate @djspiewak, can you confirm? There may also be a resource exhaustion (code cache) problem with calling the runtime compiler.

connector/src/main/scala/quasar/qscript/RenderQScriptDSL.scala
edmundnoble: I think having the name for the class available could definitely be useful, but it looks like here it would only be useful with the runtime class, so you could probably use `getClass.getName` for that there.

The usefulness of having the classname available wouldn't be too big here, I don't think, as firstly the most testable logic in here is the rendering of arguments to the constructors, not the constructor names themselves, and secondly the assumption is that all cases in `RenderQScriptDSL` are covered by `RenderQScriptDSLSpec`, so there wouldn't be any gained coverage.

alissapajer:
@edmundnoble looks great!

Needs merge from master.

Also, if you haven't, could you locally run the REPL against the final version of this? Just as a final check that the REPL is working as expected.

edmundnoble:
@alissapajer Both done now.

PR #3057, opened 2017-11-09
Alternative implementation of mergeSources

### LINKED ISSUE #3055: QSU  source merging must only consider reachable subgraph
We allow for orphaned nodes in a `QSUGraph`, which are defined as nodes contained in the `vertices` but not reachable from the `root`. When source merging computes the graph intersection of two `QSUGraph`s, it must do so against only the vertices reachable from the `root`.

Technically we never need to preserve orphan nodes in a `QSUGraph`, meaning all nodes in `vertices` could be definitionally reachable from the `root`. This isn't something we're expressing at the type level, so the current solution is to prune graphs only when algorithmically necessary. But we do have the option of enforcing this universally at the value level (or encoding it at the type level, though that will probably require more work than we have time for now).

djspiewak:
Fixes #3055 

This is mostly just an idea, though I'm pretty confident it works as written.  This should find the *closest* common source between two subgraphs, ignoring orphans, in at most linear time (proportional to the size of the larger subgraph).  In the event that there are multiple "closest" common sources, one is chosen effectively arbitrarily (there is space to be more principled about this in future).


connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -56,14 +56,42 @@ final class Graduate[T[_[_]]: CorecursiveT] extends QSUTTypes[T] {
...
+        rights: Set[Symbol],
+        visited: Set[Symbol]): Option[Symbol] = {
+
+      val lnodes = lefts.map(lvert)
+      val rnodes = rights.map(rvert)
+
+      val lexp = lnodes.flatMap(_.foldLeft(Set[Symbol]())(_ + _))
+      val rexp = rnodes.flatMap(_.foldLeft(Set[Symbol]())(_ + _))
+
+      val check = (lexp intersect visited) union (rexp intersect visited)
```
alissapajer: This implementation will fail to find the common source if the common source is the same number of steps away from each side. In that case, `lexp` and `rexp` will have a nonempty intersection, but each of their intersections with `visited` will be empty, and so `check` will be empty. So we need to also include `lext intersect rexp` here.

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -56,14 +56,42 @@ final class Graduate[T[_[_]]: CorecursiveT] extends QSUTTypes[T] {
...
+    else
+      lub(Set(left.root), Set(right.root), Set(left.root, right.root))
+
+    maybeSource map { root =>
+      SrcMerge(
+        // we merge the vertices, just in case the graphs are additively applied to a common root
+        QSUGraph[T](root, left.vertices ++ right.vertices),
+        graduateCoEnv[F](root, left),
+        graduateCoEnv[F](root, right))
+    } getOrElse scala.sys.error(s"Source merging failed from ${left.root} and ${right.root} in ${lvert ++ rvert}")
```
alissapajer: If we fail to find the common source, it's not actually error but is `Unreferenced`. In that case we need to generate a new name and add it to the graph, and this becomes the resulting `root`. In that case we call `graduateCoEnv(None, left)` and `graduateCoEnv(None, right)` since we desire no `Hole`s in the resulting `FreeQS`.

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
djspiewak: Ah that’s very correct! When I originally sketched the algorithm, I thought of the comparison in two stages with the visited set changing in between, but obviously that doesn’t happen now. Will fix

connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
alissapajer: Also, this wasn't in the original implementation, it's just something we need to do, so we can merge this PR without this change as well.

And another thought I had for testing: We should have a merge sources that returns `Set[Symbol]` for testing, where the set can have any cardinality. Then we'll have a separate function that chooses an arbitrary element from that set to be the actual common source we use.

alissapajer:
@wemrysi I pushed some changes. Could you take a look if @djspiewak doesn't get to it first? Thanks!

PR #3058, opened 2017-11-09
selectively apply mapBeforeSort

### LINKED ISSUE #3029: Selectively apply mongo's mapBeforeSort
The `mapBeforeSort` rewrite was added because a `map` on map-reduce that happens after a `sort`, unsorts the results. However, currently `mapBeforeSort` is _also_ applied when a query can be executed completely on agg.pipeline, so in these cases it is not necessary to apply this rewrite. In fact doing this rewrite unnecessarily often is the root-cause that the query needs to run on map-reduce instead of agg, so we should only apply this rewrite when it is necessary.

Part of #2884

rintcius:
Fixes #3029


mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1218,37 +1221,40 @@ object MongoDbPlanner {
       _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo", mongoQS3))
 
       // NB: Normalizing after these appears to revert the effects of `mapBeforeSort`.
-      mongoQS4 <- Trans(mapBeforeSort[T, M], mongoQS3)
+      mongoQS4 <- if (applyMapBeforeSort) Trans(mapBeforeSort[T, M], mongoQS3)
```
wemrysi: Can you add a `TODO` here, and an issue as well (if we don't have one already), mentioning that we'd like to move this transformation to `WorkflowBuilder` or `Workflow`.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1258,6 +1264,43 @@ object MongoDbPlanner {
...
+      cfg: PlannerConfig[T, EX, WF])
+    (qs: T[fs.MongoQScript[T, ?]])
+    (implicit
+      ev0: WorkflowOpCoreF :<: WF,
+      ev1: WorkflowBuilder.Ops[WF],
+      ev2: EX :<: ExprOp,
+      ev3: RenderTree[Fix[WF]])
+      : M[Crystallized[WF]] = {
+
+    def liftDoc[F[_]: Monad: ExecTimeR]
```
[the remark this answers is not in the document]
rintcius: Hmm, maybe I'm missing something, but I think in that case the type of `(anyDoc: Collection => OptionT[M, BsonDocument]` doesn't match up with `M[_]` anymore?

rintcius:
@wemrysi Ready for another look

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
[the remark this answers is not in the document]
wemrysi: Ah, ok 👍

mongodb/src/main/scala/quasar/physical/mongodb/workflow/needsMapBeforeSort.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import slamdata.Predef._
+
+import matryoshka._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+object needsMapBeforeSort {
+
+  def apply[T[_[_]]: BirecursiveT, F[_]: Traverse](wf: T[F])
```
wemrysi: As an aside, prefer `Recursive`/`Corecursive` to the `...T` variants. The latter unnecessarily constrains the shape of the recursive type. No need to change it for this PR, just wanted to mention it.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1218,37 +1219,45 @@ object MongoDbPlanner {
       _ <- BackendModule.logPhase[M](PhaseResult.tree("QScript Mongo", mongoQS3))
 
       // NB: Normalizing after these appears to revert the effects of `mapBeforeSort`.
-      mongoQS4 <- Trans(mapBeforeSort[T, M], mongoQS3)
+      // TODO look into adding mapBeforeSort to WorkflowBuilder or Workflow stage
+      // instead, so that we can avoid having to rerun the transformations.
+      // See #3063
+      mongoQS4 <- if (applyMapBeforeSort)
```
wemrysi: @rintcius I wonder if we could move the application of `mapBeforeSort` into it's own function so that we only have to apply it once? The idea being we apply all the other qscript transformations, save that value and then apply `mapBeforeSort` to it. If it is unnecessary we just use the previous value.

I'm a little concerned about the double transformation having an adverse affect on compile times since we know they can already be a bit long in some cases.

rintcius:
@wemrysi I implemented the optimizations that we discussed

jsantos17:
I would like to see this merged soon as some of the queries in #3078 are affected by this problem. 

@wemrysi: I can take over the review if there's too much in your plate at the moment.

wemrysi:
@rintcius @jsantos17 Sorry about the delay, this had just slipped my mind.

PR #3059, opened 2017-11-09
OID() on join func

### LINKED ISSUE #3052: MongoDB JOIN fails if you use OID() function if the field is missing
Tested in Quasar 26.0.5.

While the issues around `OID()` were fixed in #2969, a new bug has surfaced using the `OID()` function as part of a `JOIN` statement.

When you do this you will get an error as follows:

```
Failed to execute SQL^2 query.: MongoDB Error: Command failed with error 10448: 'Error: invalid object id: length : @:8:78 @:4:1 ' on server calibermind-sb-shard-00-00-aweli.mongodb.net:27017. The full response is { "ok" : 0.0, "errmsg" : "Error: invalid object id: length :\n@:8:78\n@:4:1\n", "code" : 10448, "codeName" : "Location10448" }
```

**Steps to reproduce**:

Run the following `insert`s in the MongoDB shell in the `test` database:

```
db.oidtestparent.insert({"_id": ObjectId("5a035d84a1180dcd84162e72"), "a": 1})
db.oidtestchild.insert({"parentid": "5a035d84a1180dcd84162e72", "b": 2})
db.oidtestchild.insert({"b": 3})
```
Run the following command in the Quasar REPL:

```
select * from `/mongodb-3.4/test/oidtestparent` as p inner join `/mongodb-3.4/test/oidtestchild` as c on p.`_id` = oid(c.parentid)
```

Doing this you will get the same error mentioned above.

**NOTE**: Running the following in the REPL works without error:

```
select oid(parent) from `/mongodb-3.4/test/oidtestchild`
```

so it seems that there might be missing checks during the `JOIN` phase that are present if you're using the `OID()` function call to just project out the field.

jsantos17:
Fixes #3052


rintcius:
@jsantos17 I'll review this first thing tomorrow. 

FYI the appveyor build failed - doesn't look like it's related to your PR, but can you create an issue for that failing test if there isn't already? 

@djspiewak can you restart appveyor (or can we do it ourselves already)?

djspiewak:
@rintcius I've restarted it.  We have a real appveyor account now, but we haven't migrated the build yet.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -535,15 +535,37 @@ object MongoDbPlanner {
...
+      //     and `Or` case above, the selector choice can be tweaked depending on how
+      //     strict we want to be with extracted typechecks. See #2883
+      case MFC(Eq(a, b))  => invoke2Rel(a._2, b._2)(Selector.And(_, _))
+      case MFC(Neq(a, b)) => invoke2Rel(a._2, b._2)(Selector.And(_, _))
+      case MFC(Lt(a, b))  => invoke2Rel(a._2, b._2)(Selector.And(_, _))
+      case MFC(Lte(a, b)) => invoke2Rel(a._2, b._2)(Selector.And(_, _))
+      case MFC(Gt(a, b))  => invoke2Rel(a._2, b._2)(Selector.And(_, _))
+      case MFC(Gte(a, b)) => invoke2Rel(a._2, b._2)(Selector.And(_, _))
+
+      case MFC(Guard((Embed(MFC(ProjectKey(Embed(MFC(Undefined())), _))), _), typ, cont, _)) =>
```
rintcius: I don't understand the `Undefined` here. Can you explain?

jsantos17: That's `Hole` in disguise. As part of the fold that calculates the extracted typechecks we move the `Filter`'s predicate from a `Free` fixed point into an arbitrary `RecursiveT`. Since we don't have a way to represent leaves directly (i.e Hole) anymore we replace it with `Undefined`. It's done [here](https://github.com/quasar-analytics/quasar/blob/master/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala#L125).

jsantos17: Good idea on the test. I'll add it.

rintcius: Ah ok, that makes sense. Maybe good to add a comment in the code as well?

jsantos17:
@rintcius ready for another look. I have added the comment and the integration test.

rintcius:
@jsantos17 looks good - feel free to merge as soon as it's green (I guess I'll be offline by then)

PR #3060, opened 2017-11-09
Organize and extend optics for EJson

wemrysi:
Improves the optics story for EJson by

1. Moving types and type-specific optics out of the package object and into their own respective files, making them easier to find and easier to import a la carte.

2. Adding fixpoint prisms, based on component prisms, allowing for convenient, compositional construction and pattern matching for values having `EJson` as their pattern-functor.

### DIFF
--- a/ejson/src/main/scala/quasar/ejson/package.scala
+++ b/ejson/src/main/scala/quasar/ejson/package.scala
@@ -17,7 +17,6 @@
 package quasar
 
 import slamdata.Predef._
-import quasar.contrib.matryoshka._
 import quasar.fp.ski._
 
 import java.lang.String
@@ -26,10 +25,8 @@ import scala.Unit
 import scala.collection.immutable.{List, ListMap}
 import scala.math.{BigDecimal, BigInt}
 
-import matryoshka._
-import matryoshka.implicits._
 import monocle.{Iso, Prism}
-import scalaz._, Scalaz._
+import scalaz._
 
 package object ejson {
   def arr[A] =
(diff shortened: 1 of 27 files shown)


alissapajer:
As a heads up, this will have conflicts with #3056

edmundnoble:
In fact, I think I should remove what I added in #3056 for EJson in favor of this. I'll try to get this merged first to that end, does that make sense to you @alissapajer?

alissapajer:
@edmundnoble that sounds good to me. I'll publish my review of #3056 with that in mind.

ejson/src/main/scala/quasar/ejson/package.scala
```diff
@@ -16,193 +16,17 @@
...
-import scala.math.{BigDecimal, BigInt}
-
-import matryoshka._
-import matryoshka.implicits._
-import monocle.{Iso, Prism}
-import scalaz._, Scalaz._
+import scalaz.{Coproduct, Inject}
 
 package object ejson {
-  def arr[A] =
```
edmundnoble: :+1: :+1: :+1: :+1:

foundation/src/main/scala/quasar/contrib/matryoshka/package.scala
```diff
@@ -35,6 +35,11 @@ package object matryoshka {
       first)(
       (prev, next) => x => prev(x).fold(next(x))(orOriginal(next)(_).some))
 
+  def birecursiveIso[T, F[_]: Functor]
```
edmundnoble: I'll be making use of this, thanks :+1:

frontend/src/test/scala/quasar/sst/CompressionSpec.scala
```diff
@@ -53,19 +53,21 @@ final class CompressionSpec extends quasar.Qspec
...
     Arbitrary(Gen.oneOf(
-      Gen.const(C(ejson.nul[J]()).embed),
-      arbitrary[Boolean] map (b => C(ejson.bool[J](b)).embed),
-      arbitrary[String] map (s => C(ejson.str[J](s)).embed),
-      arbitrary[BigDecimal] map (d => C(ejson.dec[J](d)).embed),
-      arbitrary[Byte] map (b => E(ejson.byte[J](b)).embed),
-      arbitrary[Char] map (c => E(ejson.char[J](c)).embed),
-      arbitrary[BigInt] map (i => E(ejson.int[J](i)).embed)
+      Gen.const(J.nul()),
+      arbitrary[Boolean] map (b => J.bool(b)),
```
edmundnoble: Does `map(J.bool)` not eta-expand?

wemrysi: Nope 😞

wemrysi: Though, looks like `map(J.bool(_))` works. I had tunnel-vision with replacing the old stuff, so didn't think to use that syntax.

PR #3064, opened 2017-11-10
Add some tests in `ApiSpec` to validate how querying views behaves

jedesah:
- Add `QuasarAPIImpl` with the intention of eventually aggregating the entire "top-level" Quasar programmatic API
- Make all Quasar errors a subclass of `QuasarError`
- Add `interpWithErrs` to Quasar to provide a better programmatic API
- Add `initMimirOnly` to have a no argument constructor for `Quasar` for the simplest case

### DIFF
--- a/connector/src/main/scala/quasar/fs/FileSystemError.scala
+++ b/connector/src/main/scala/quasar/fs/FileSystemError.scala
@@ -17,7 +17,7 @@
 package quasar.fs
 
 import slamdata.Predef._
-import quasar.Data
+import quasar.{Data, QuasarError}
 import quasar.Planner.PlannerError
 import quasar.fp._
 import quasar.frontend.logicalplan.LogicalPlan
@@ -30,7 +30,7 @@ import pathy.Path.posixCodec
 import scalaz._
 import scalaz.syntax.show._
 
-sealed abstract class FileSystemError
+sealed abstract class FileSystemError extends QuasarError
 
 object FileSystemError {
   import QueryFile.ResultHandle
--- a/interface/src/main/scala/quasar/main/QuasarAPIImpl.scala
+++ b/interface/src/main/scala/quasar/main/QuasarAPIImpl.scala
@@ -22,7 +22,7 @@ import quasar.contrib.pathy._
 import quasar.sql._
 import quasar.fp.free.foldMapNT
 import quasar.fs.{FileSystemError, QueryFile, ReadFile}
-import quasar.fs.mount.{Mounting, MountConfig}
+import quasar.fs.mount.{Mounting, MountConfig, MountingError}
 import quasar.fs.mount.module.Module
 
 import matryoshka.data.Fix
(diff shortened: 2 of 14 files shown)


rintcius:
@jedesah You assigned me and added @wemrysi as reviewer. How do I interpret this in terms of what is expected from me vs Emrys?

wemrysi:
@rintcius Assignment is a request to review and be responsible for the final merging of the PR while a reviewer is just the review.

connector/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -30,7 +30,7 @@ import pathy.Path.posixCodec
 import scalaz._
 import scalaz.syntax.show._
 
-sealed abstract class FileSystemError
+sealed abstract class FileSystemError extends QuasarError
```
rintcius: I am wondering if these nested sum types have any negative implications w.r.t. https://github.com/quasar-analytics/quasar/wiki/Faking-Sum-Types and if there's a better encoding. Maybe worth trying the Optics variant mentioned on the wiki page?

rintcius: I can also be a nice opportunity to start using https://github.com/frees-io/iota (though I had an [issue](https://github.com/frees-io/iota/issues/89) with it when I tried it out in the weekend)

jedesah: Since the types we want to deal with most of the places are more concrete like `FileSystemError`, `SemanticError`, I don't think we want prisms that construct them as `QuasarError`. `QuasarError` is just so that they can be combined at the top-level api layer to make things simpler. We could potentially use `iota` for this, but I think the current approach appears to work well. If there is some concrete reason why we might run into issues in the future, then we should discuss otherwise I say let's try this and revisit if it proves problematic.

interface/src/main/scala/quasar/main/QuasarAPIImpl.scala
```diff
@@ -0,0 +1,65 @@
...
+  private val mount = Mounting.Ops[CoreEff]
+  private val fsQ = new FilesystemQueries[CoreEff]
+
+  def createView(path: AFile, sql: ScopedExpr[Fix[Sql]]): F[Unit] =
+    mount.mount(path, MountConfig.viewConfig(sql, Variables.empty)) foldMap inter
+
+  def createModule(path: ADir, statements: List[Statement[Fix[Sql]]]) =
+    mount.mount(path, MountConfig.moduleConfig(statements)) foldMap inter
+
+  def getMount(path: APath) =
```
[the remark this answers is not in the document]
jedesah: sure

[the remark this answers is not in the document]
jedesah: Done! (sorry for the delay, didn't want to change branch in midst of something else! 😁 )

connector/src/main/scala/quasar/fs/FileSystemError.scala
andyscott: Iota 0.3.3 is released and should work now for your use case. One caveat is that you must pass the full type parameters when using iota to create `NaturalTransformation`s for coproducts. This is due to the interplay between variances and macro type tags. /cc @jdegoes

PR #3065, opened 2017-11-10
QSU MappableRegion tests

wemrysi:
Adds tests for `MappableRegion` and integrates the `EJson` prisms into the `QSU` dsl to make working with constants more convenient.


wemrysi:
@alissapajer I just found a bug in the `MappableRegion` impl, so expect another commit here with the fix and another test.

alissapajer:
@wemrysi :+1: I'll hold off on merge until that's added.

PR #3068, opened 2017-11-11
AutoJoin3 handling in ReifyProvenance

### LINKED ISSUE #3066: Transform a 3-way autojoin into nested theta joins
If an `AutoJoin3` reaches `ReifyProvenance`, it must be transformed into two nested `ThetaJoin`s. A previous stage is responsible for transforming autojoins into non-join structures, when possible. So anything that reaches `ReifyProvenance` must be transformed directly into its qscript-ish counterpart.

alissapajer:
Depends on #3067 

Closes #3066 

Handles the case where we need to turn an `AutoJoin3` into nested `ThetaJoin`s because we have failed to eliminate any of the joins.

Also, this needs tests.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
@@ -143,13 +143,14 @@ final class ReifyProvenance[T[_[_]]: BirecursiveT: EqualT] extends QSUTTypes[T] 
   def apply[F[_]: Monad: PlannerErrorME: NameGenerator](qsu: AuthenticatedQSU[T])
       : F[AuthenticatedQSU[T]] = {
 
-    val pair: F[(List[NewVertex], SMap[Symbol, QSU[Symbol]])] =
-      qsu.graph.vertices.traverse[X[F, ?], QSU[Symbol]](toQScript[F](qsu.dims))(appX[F])
+    val pair: F[(List[NewVertex], QSUNodes[T])] =
+      qsu.graph.vertices.traverse[X[F, ?], QSU[Symbol]](
+        toQScript[F](qsu.dims))(appX[F])
 
     pair.map {
       case (nw, oldVertices) =>
         val (newV, newD) =
-          nw.foldLeft((SMap[Symbol, QSU[Symbol]](), SMap[Symbol, Dimensions[prov.P]]())) {
+          nw.foldLeft[(QSUNodes[T], QSUDims[T])](SMap() -> SMap()) {
             case ((vAcc, dAcc), NewVertex(name, value, dims)) =>
               (vAcc + (name -> value), dAcc + (name -> dims))
           }
--- a/connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
@@ -18,7 +18,7 @@ package quasar.qscript.qsu
 
 import slamdata.Predef._
 
-import quasar.Planner.{InternalError, PlannerErrorME}
+import quasar.Planner.PlannerErrorME
 import quasar.common.JoinType
 import quasar.fp.ski.κ
 import quasar.qscript.{HoleF, IncludeId, LeftSideF, MFC, ReduceIndexF, RightSideF}
@@ -28,7 +28,7 @@ import quasar.qscript.qsu.{QScriptUniform => QSU}
 import quasar.qscript.qsu.ApplyProvenance.AuthenticatedQSU
 
 import matryoshka.{BirecursiveT, EqualT}
-import scalaz.{\/-, Applicative, Free, ICons, INil, Monad}
+import scalaz.{\/-, Applicative, Free, Monad}
 import scalaz.Scalaz._
 
 final class ReifyProvenance[T[_[_]]: BirecursiveT: EqualT] extends QSUTTypes[T] {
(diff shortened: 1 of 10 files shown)


connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
```diff
@@ -16,74 +16,147 @@
...
-  def apply[F[_]: Monad: PlannerErrorME](qsu: AuthenticatedQSU[T])
-      : F[AuthenticatedQSU[T]] =
-    for {
-      vertices <- qsu.graph.vertices.traverse[F, QSU[Symbol]](toQScript[F](qsu.dims))
-    } yield {
-      AuthenticatedQSU[T](QSUGraph(qsu.graph.root, vertices), qsu.dims)
+  def apply[F[_]: Monad: PlannerErrorME: NameGenerator](qsu: AuthenticatedQSU[T])
+      : F[AuthenticatedQSU[T]] = {
+
+    val pair: F[(List[NewVertex], QSUNodes[T])] =
```
alissapajer: There is probably some code :golf: to be played in these parts.

connector/src/main/scala/quasar/qscript/qsu/package.scala
```diff
@@ -21,4 +21,5 @@ import quasar.qscript.provenance.Dimensions
 
 package object qsu {
   type QSUDims[T[_[_]]] = SMap[Symbol, Dimensions[QProv.P[T]]]
+  type QSUNodes[T[_[_]]] = SMap[Symbol, QScriptUniform[T, Symbol]]
```
alissapajer: We are using "nodes" and "vertices" interchangeably, and it would be good to choose one. Though actually, to be fair, I think this is the introduction of "nodes" :)

"vertices"  and "edges" are the most common terms used when describing a graph (from what I've seen), and we do have a real graph. But I've seen "nodes" used in the literature as well. Do you all have a preference? I wish "vertices" wasn't so many syllables... maybe we can use "verts"?

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
```diff
@@ -16,74 +16,147 @@
...
 import quasar.qscript.ReduceFunc._
-import quasar.qscript.MapFuncsCore.ConcatArrays
+import quasar.qscript.MapFuncsCore.IntLit
 import quasar.qscript.qsu.{QScriptUniform => QSU}
 import quasar.qscript.qsu.ApplyProvenance.AuthenticatedQSU
 
-import matryoshka.{BirecursiveT, EqualT}
-import scalaz.{\/-, Applicative, Free, ICons, INil, Monad}
+import matryoshka._
+import matryoshka.data._
```
alissapajer: I didn't want to import the world from these two, but they are bringing in some required implicits that I didn't take the time to track down.

connector/src/main/scala/quasar/qscript/qsu/package.scala
djspiewak: Vertices and "verts" (when we don't feel like typing) is my preference.  Nodes is a good term in and of itself, but unlike vertices it doesn't have a conventional pairing to describe edges.

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
```diff
@@ -16,61 +16,146 @@
...
+      val qsu: QSU[Symbol] =
+        QSU.ThetaJoin(left, right, condition, JoinType.Inner, combiner)
+
+      appX[F].point(qsu)
+
+    case QSU.AutoJoin3(left, center, right, combiner3) =>
+      val _condition: JoinFunc =
+        prov.autojoinCondition(dims(left), dims(center))(κ(HoleF))
+
+      val _combiner: JoinFunc = // <3 [[]]
```
wemrysi: Should we do a `{left: ..., center: ...}` map instead?

connector/src/main/scala/quasar/qscript/qsu/package.scala
wemrysi: Agreed, maybe `QSUVertices` or `QSUVerts` here?

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
```diff
@@ -16,74 +16,147 @@
...
-          val condition: JoinFunc =
-            prov.autojoinCondition(dims(left), dims(right))(κ(HoleF))
+  case class NewVertex(name: Symbol, value: QSU[Symbol], dims: Dimensions[prov.P])
 
-          val combiner: JoinFunc = Free.roll(combiner0.map {
-            case 0 => LeftSideF
-            case 1 => RightSideF
-            case n => scala.sys.error(s"AutoJoin combiner has wrong size $n. Expected size 2.")
-          })
+  type X[F[_], A] = F[(List[NewVertex], A)]
```
wemrysi: I think this is just `WriterT[F, List[NewVertex], A]`.

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
alissapajer: Yeah, actually what I think we should do is generate new names for these object keys, now that we have name generation so nicely accessible.

connector/src/main/scala/quasar/qscript/qsu/ReifyProvenance.scala
alissapajer: It is, and it provides the `Applicative` I need. :+1:

alissapajer:

PR #3070, opened 2017-11-13
Fixes for listContent and create dir (RDBMS)

kciesielski:
This PR fixes 3 issues in the RDBMS connector:
1. There is no more `DefaultSchema`/`CustomSchema` differentiation. This was based on an erroneous assumption that `DefaultSchema` maps to the root directory. After we resign from this assumption, we don't need a special case for `DefaultSchema`,
2. QueryFile.listContent now returns child dirs and files correctly, without listing recursive subdirs,
3. Creating a DB schema in Postgres now creates all intermediary schemas if schema represents a nested directory.


kciesielski:
@rabbitonweb These are important fixes, we need them for the other branches. Hoping for a quick review and merge :)

PR #3076, opened 2017-11-13
Graduate distinct

### LINKED ISSUE #3040: Graduate `QSU.Distinct`
The `QSU.Distinct` node needs to `graduate` to a `Reduce`.

alissapajer:
Fixes #3040

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
@@ -26,11 +26,14 @@ import quasar.qscript.{
   educatedToTotal,
   Filter,
   Hole,
+  HoleF,
   LeftShift,
   Map,
   QCE,
   Read,
   Reduce,
+  ReduceFuncs,
+  ReduceIndexF,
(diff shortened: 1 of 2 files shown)


connector/src/main/scala/quasar/qscript/qsu/Graduate.scala
```diff
@@ -153,7 +155,13 @@ final class Graduate[T[_[_]]: CorecursiveT] extends QSUTTypes[T] {
             QCE(Subset[T, QSUGraph](source, fromBranch, op, countBranch))
         }
 
-      case QSU.Distinct(source) => slamdata.Predef.??? // TODO
+      // TODO distinct should be its own node in qscript proper
+      case QSU.Distinct(source) =>
```
wemrysi: I think this needs to happen in `ReifyProvenance` in order to obtain the correct buckets.

wemrysi: Actually, nevermind, not sure what I was thinking, this should be fine.

PR #3082, opened 2017-11-14
Enable HTTP/2

### LINKED ISSUE #3081: Support HTTP/2
I think it should be as trivial as just adding a line [`.enableHttp2(true)`](http://http4s.org/v0.18/api/org/http4s/server/blaze/BlazeBuilder.html#enableHttp2(enabled:Boolean):BlazeBuilder.this.Self) here:

https://github.com/quasar-analytics/quasar/blob/2efde00a0bc69f942acdb41a14ca77d3053ab28e/web/src/main/scala/quasar/server/Http4sUtils.scala#L103-L108

safareli:
fix #3081

### DIFF
--- a/web/src/main/scala/quasar/server/Http4sUtils.scala
+++ b/web/src/main/scala/quasar/server/Http4sUtils.scala
@@ -101,6 +101,7 @@ object Http4sUtils {
       exec
     }
     BlazeBuilder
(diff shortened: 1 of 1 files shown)


safareli:
Not sure if it actually fixes the issue tho, just made the pr as it was too easy

PR #3084, opened 2017-11-14
Fixes to reading and rewriting

djspiewak:
Collapse wasn't working, so I had to add a `Unary` node which is rewritten away immediately.  Also `LPtoQS` wasn't working at all due to `\circ` precedence.  All that is fixed.  We're very close to graduating.  Just have to get that thesis banged out…

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
@@ -91,7 +91,7 @@ final case class QSUGraph[T[_[_]]](
         visited <- MS.get
 
         back <- if (visited(g.root)) {
-          this.point[G]
+          g.point[G]
         } else {
           for {
             _ <- MS.modify(_ + g.root)
--- a/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
@@ -90,17 +90,41 @@ final case class QSUGraph[T[_[_]]](
       for {
         visited <- MS.get
 
-        back <- if (visited(root)) {
+        back <- if (visited(g.root)) {
           this.point[G]
         } else {
           for {
-            _ <- MS.modify(_ + root)
-            recursive <- g.unfold.traverse(inner)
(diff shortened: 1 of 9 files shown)


alissapajer:
I'd like to see exactly what goes wrong if we use `Map` in place of `Unary` using the new version of `rewriteM`. Any problem with `FreeMap` equality will crop up again, so we should understand that more deeply now before masking over it with `Unary`.

djspiewak:
@alissapajer It's not just about `Equal[FreeMap]`, though that does seem to be problematic.  The bigger issue is that there is no sane `Order[FreeMap]`, much less `Hash[FreeMap]`, which is what I *really* need to do this.  The compaction algorithm is fundamentally dependent on an `SMap[QSU, Symbol]`, which naturally requires equality and hash checks on `QSU`.  Any other data structure there destroys the asymptotic complexity of the algorithm (e.g. `EqMap` would be O(n^2)).

alissapajer:
@djspiewak can you add some scaladoc about why have `Unary`then? A sentence saying we use `SMap[QSU, Symbol]` so we need sane comparison checks on `QSU` is all we need. We currently have one giant blob of `QSU` types, so it's impossible to tell just by looking at them which are used in which contexts. `Unary` appears unnecessary unless you know it's being used in a comparison situation.

alissapajer:
Down the road when we're figuring out how to break apart these various `QSU` types, hashable-ness will be a useful distinction.

djspiewak:
@alissapajer Willdo.  I meant to document it, but never got back round to it.

connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -90,17 +90,41 @@ final case class QSUGraph[T[_[_]]](
       for {
         visited <- MS.get
 
-        back <- if (visited(root)) {
+        back <- if (visited(g.root)) {
           this.point[G]
```
wemrysi: This should be `g.point[G]`, no?

djspiewak: 👍 🤦‍♂️

PR #3099, opened 2017-11-15
Basic autojoin elimination

djspiewak:
Depends on, like, all my other PRs.

Doesn't yet coalesce reductions, because it's happening in the wrong place.  I need `ReifyProvenance` to be split up a bit, so that I can do this rewrite *after* we get `QSReduce` nodes, but before we lose our `AutoJoin` nodes.

But even with that limitation, this is still a pretty big step forward due to the way it cleans up queries.


connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -144,9 +144,41 @@ final case class QSUGraph[T[_[_]]](
...
+  def generateRevIndex: QSUGraph.RevIdx[T] =
+    vertices map { case (key, value) => value -> key }
 }
 
 object QSUGraph extends QSUGraphInstances {
+  type RevIdx[T[_[_]]] = SMap[QScriptUniform[T, Symbol], Symbol]
+
+  def withName[T[_[_]], F[_]: Monad: NameGenerator](
+      node: QScriptUniform[T, Symbol])(
+      implicit MS: MonadState_[F, SMap[QScriptUniform[T, Symbol], Symbol]]): F[QSUGraph[T]] = {
```
wemrysi: An opportunity for `RevIdx`.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -0,0 +1,122 @@
...
+          (mask, acc)
+        else
+          (mask + g.root, g :: acc)
+    }
+
+    minimized
+  }
+
+  private def extractSources[A](fma: FreeMapA[A]): List[A] =
+    fma.foldMap(λ[MapFunc ~> List](_.toList))
```
wemrysi: ⛳️ `Foldable[F]` implies `Foldable[Free[F, ?]]` so should be able to just call `toList`.

djspiewak: Oh nice.  I didn't realize that.

PR #3109, opened 2017-11-15
Fix "Removing folders when .trash does not exist stopped working on spark-hdfs"

### LINKED ISSUE #2973: Removing folders when .trash does not exist stopped working on spark-hdfs

rabbitonweb:
Fixes #2973


rabbitonweb:
@wemrysi this is green on CI now

PR #3116, opened 2017-11-16
Rewrite tests to use QScript DSL

edmundnoble:
Also reduced the use of tuples in `mkDefaults` because it makes the pretty printer more fragile.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/construction.scala
+++ b/connector/src/main/scala/quasar/qscript/construction.scala
@@ -29,7 +29,7 @@ import quasar.ejson.EJson
 
 object construction {
   final case class Func[T[_[_]]]()(implicit birec: BirecursiveT[T]) {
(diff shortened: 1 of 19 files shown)


edmundnoble:
@wemrysi It looks like that file's using the DSL too, it was written after the DSL was introduced - maybe the imports are throwing you off. I caught `RewriteNullSpec` in the marklogic connector though, after another pass.

wemrysi:
@edmundnoble Hm, did you see the compile errors for `PlannerQScriptSpec`, that's why I pointed it out.

edmundnoble:
Ahhh no I didn't, thanks.

edmundnoble:
@djspiewak Could you restart appveyor please?

alissapajer:
@edmundnoble ditto what @wemrysi said, thanks for doing all this, our tests are much better for it!

PR #3117, opened 2017-11-16
Improve the way we handle constant plans

jedesah:
This both:

- improves the top-level api by not forcing clients to handle the constant plan case separately
- improves the internal handling by replacing a bunch of ad-hoc places where we were calling `refineConstantPlan` by a single interpreter in addition to views and modules which intercepts constant plans. This also achieves much better separation of concerns

Other changes necessary along the way:

- add new `deleteOrIgnore` to `ManageFile.Ops`
- add `listContents` smart constructor for `ListContents` algebra term (surprised this wasn't found to be useful earlier...)
- Put `FileSystemError` from `save` and `create` into `M` instead of the result type
  - Having the errors be values makes it easy to ignore the errors
  - It's also less convenient to work with
  - I did have to just choose one error though instead of all the errors but I don't think that's a huge deal
- Use new test interpreters for `CompileServiceSpec` and `ExecuteServiceSpec`
- Remove `Module.ResultHandle` as `Module` can just re-use `QueryFile.ResultHandle` since it's delegating to `QueryFile` anyway
- Fix some alignment issues as well as remove some unnecessary constraints on some functions
- Other miscellaneous improvements


core/src/main/scala/quasar/fs/constantPlans.scala
```diff
@@ -0,0 +1,98 @@
...
+import quasar.frontend.logicalplan.{Constant, LogicalPlan}
+
+import matryoshka.data.Fix
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+object constantPlans {
+
+  /** Identify plans which reduce to a (set of) constant value(s). */
+  def isConstant(lp: Fix[LogicalPlan]): Option[List[Data]] =
```
edmundnoble: Maybe this could have a different name, seeing as it doesn't just tell you which plans are constant, it actually gives you back the constant values.

jedesah: @edmundnoble I agree, but fell short coming up with something better, do you have suggestions?

core/src/test/scala/quasar/fs/mount/ViewFileSystemSpec.scala
```diff
@@ -760,8 +742,7 @@ class ViewFileSystemSpec extends quasar.Qspec with TreeMatchers {
 
       val qlp =
         quasar.queryPlan(q, Variables.empty, rootDir, 0L, None)
-          .run.value.toOption.get
-          .valueOr(_ => scala.sys.error("Expected a non-constant plan but received a constant plan"))
+          .run.value.valueOr(e => scala.sys.error("Unexpected error compiling sql query:" + e.shows))
```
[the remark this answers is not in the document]
jedesah: Will do!

core/src/main/scala/quasar/fs/mount/module/Module.scala
```diff
@@ -37,14 +37,6 @@ sealed abstract class Module[A]
 
 object Module {
 
-  final case class ResultHandle(run: Long) extends scala.AnyVal
```
wemrysi: I'm 👎 on this, I think https://github.com/quasar-analytics/quasar/pull/2043#discussion_r111639052 still applies.

it/src/main/resources/tests/literalReductionWithNegative.test
```diff
@@ -1,7 +0,0 @@
-{
```
wemrysi: Can we keep this around? Seems useful to ensure we can handle constant plans at this level.

connector/src/main/scala/quasar/fs/QueryFile.scala
```diff
@@ -345,6 +345,9 @@ object QueryFile {
       lsR(currentDir).foldLeft(Set.empty[RFile])(_ + _)
     }
 
+    def listContents(dir: ADir): M[Set[PathSegment]] =
```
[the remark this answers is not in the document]
jedesah: I'm happy to remove it, I just didn't think of `ls` at the time since almost all our "smart constructors" usually have the exact same name

connector/src/main/scala/quasar/fs/WriteFile.scala
```diff
@@ -134,111 +134,116 @@ object WriteFile {
...
+                            : Process[M, Unit] = {
 
       def cleanupTmp(tmp: AFile)(t: Throwable): Process[M, Nothing] =
-        Process.eval_(ensureAbsent(tmp)).causedBy(Cause.Error(t))
+        Process.eval_(MF.deleteOrIgnore(tmp).liftM[FileSystemErrT]).causedBy(Cause.Error(t))
 
       MF.tempFile(dst).liftM[Process] flatMap { tmp =>
         appendChunked(tmp, src)
           .map(some).append(Process.emit(none))
+          // TODO: Explain why this is correct because `take(1)` seems all kind of
```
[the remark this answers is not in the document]
jedesah: Ah, yes that makes sense, maybe I'll a comment since it's a little confusing at first

core/src/main/scala/quasar/fs/constantPlans.scala
jedesah: I like `asConstant`, I'll make the change unless @edmundnoble has an objection

it/src/main/resources/tests/literalReductionWithNegative.test
jedesah: I don't think it's the responsibility of connectors to know how to handle constant plans. They translate the QScript they receive.

Besides, I removed it because it was failing and it seemed to me like a lot of effort to add the constant plan interpreter to the test fixture code in `it` (but maybe that's because I'm not terribly familiar with that part of the codebase). But after wondering what to do about this for a little bit, I was like: "wait a minute why is this here if it needs something defined in `core` and which should not be the responsibility of anything that requires a stateful test to assert".

core/src/main/scala/quasar/fs/constantPlans.scala
edmundnoble: That sounds fine to me.

it/src/main/resources/tests/literalReductionWithNegative.test
wemrysi: @jedesah Ok, fair enough, maybe we can add a constant query to our top-level query service spec or something, just to show we can deal with them.

connector/src/main/scala/quasar/fs/QueryFile.scala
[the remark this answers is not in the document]
wemrysi: Yeah, I'm not opposed to having both, if we do I'd just prefer that one be an alias of the other. We can probably make `listContents` the canonical one for the reason you mention above.

it/src/main/resources/tests/literalReductionWithNegative.test
jedesah: @wemrysi Did you notice that I migrated this test to `ConstantPlansSpec` in another form? But even though I think that's where we should probably add additional tricky cases we want to make sure work, I think it still makes sense to have one test at the `ExecuteServiceSpec` level to make sure the service does actually call `asConstant` somewhere in the call stack. `CompileServiceSpec` already has a test for a constant plan, I'll add the one to `ExecuteServiceSpec` though

core/src/main/scala/quasar/fs/constantPlans.scala
```diff
@@ -0,0 +1,98 @@
...
+      for {
+        h <- seq.next.map(ResultHandle(_))
+        _ <- state.put(h, data.toVector)
+      } yield h
+
+    λ[QueryFile ~> Free[S, ?]] {
+      case ExecutePlan(lp, out) =>
+        isConstant(lp).fold(
+          query.execute(lp, out).run.run)(
+          // I believe it is safe to call void here because we generated the data
```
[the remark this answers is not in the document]
jedesah: Actually, I ended up fixing this problem by putting the error into the `FileSystemErrT` which also cleaned up some code in other places as well. The only drawback is that you can only see one of possibly many errors. I don't think that's a huge problem and makes this api much safer to use. But thanks for pointing out this comment is no longer relevant and that void is redundant as the return type is already `Unit`

alissapajer:
@jedesah is this ready for @wemrysi to review again?

jedesah:
@alissapajer @wemrysi It is. Not sure what the protocol is here, perhaps should have mentionned it was earlier

PR #3119, opened 2017-11-17
Add property to QSU.Transpose indicating what to retain

### LINKED ISSUE #3112: Add a flag to Transpose to allow non-identity-including shifts
This will allow us to emit `ShiftedRead` without `IncludeIds`, as well as cut down on some of the extraneous projection going on.  We really only need the id inclusion during reads of things like `FlattenMapKeys`, so it's not really that much of a burden.

wemrysi:
Also moves the conversion of `Transpose` -> `LeftShift` to `ApplyProvenance`.

Fixes #3112.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
@@ -69,8 +69,8 @@ object QScriptUniform {
 
       case Read(path) => (Read(path): QScriptUniform[T, B]).point[G]
 
-      case Transpose(source, rotations) =>
-        f(source).map(Transpose(_, rotations))
+      case Transpose(source, retain, rotations) =>
+        f(source).map(Transpose(_, retain, rotations))
 
       case LeftShift(source, struct, idStatus, repair) =>
         f(source).map(LeftShift(_, struct, idStatus, repair))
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -187,8 +187,21 @@ object QScriptUniform {
   // LPish
   final case class Transpose[T[_[_]], A](
       source: A,
+      retain: Retain,
       rotations: Rotation) extends QScriptUniform[T, A]
 
+  sealed trait Retain extends Product with Serializable {
+    def fold[A](ids: => A, vals: => A): A = this match {
```
djspiewak: When we have time to catch our breath, we should probably take a pass over all of our idiosyncratic encodings of things to make sure that we're relatively uniform.  :-)  Like I think defining `fold` is a good idea here, but our other algebras don't have it, so we should probably fix that at some point.

connector/src/test/scala/quasar/qscript/qsu/ReadLPSpec.scala
```diff
@@ -93,73 +84,49 @@ object ReadLPSpec extends Qspec with CompilerHelpers with DataArbitrary with QSU
...
       }
     }
 
     "convert FlattenArray" in {
       lpf.invoke1(StructuralLib.FlattenArray, read("foo")) must readQsuAs {
-        case AutoJoin2C(
-          Transpose(TRead(_), QSU.Rotation.FlattenArray),
-          DataConstant(Data.Int(i)),
-          MapFuncsCore.ProjectIndex(LeftSide, RightSide)) => i mustEqual 1
+        case Transpose(TRead(_), QSU.Retain.Values, QSU.Rotation.FlattenArray) => ok
```
alissapajer: Such a nice declarative pattern!! :tada:

Issue #3090, opened 2017-11-14
Fix (or remove) overly-fragile planner specs


djspiewak:
The poster child here is mongo's `PlannerSpec`, which has completely broken under qsu.  Some of that is probably just because qsu is producing terrible qscript right now, but we're very rapidly going to get to a point where these tests are just going to be overly-specific.  They should probably be rephrased in terms of @edmundnoble's DSL, with only a couple very general tests remaining.

Couchbase also has some tests like this, btw.

This is a blocker for landing in master, since the unit tests won't run so long as this is broken.

rintcius:
This relates to #2802 

I agree that `PlannerSpec` should only have a couple of basic tests (I'm right now porting some of the queries to the qscript DSL and it's a pain for larger queries). For the more complex queries we can start _tracking_ the plans of the ITs (the `.test`). Compared to `PlannerSpec` this is both more useful and less work to maintain.

rintcius:
I'm picking this up in favour of #3080 as this is more urgent and useful. 

@djspiewak Shall we discuss once you are online (and had your coffee)?

rintcius:
@djspiewak So I'll work on making the tests less fragile. E.g. assert that there's no Map Reduce, assert max. number of pipeline stages, check for a certain stage etc.

Apart from this, I think we should also verify whether the sql^2 that's tested in the `PlannerSpec` is being tested in IT & if not add it there, otherwise it's going to be difficult to ensure their correctness after QSU has landed. Agree or do you see a better way?

djspiewak:
Agreed.

PR #3120, opened 2017-11-17
Refactor mongo planner spec

rintcius:
First PR related to #3090 

This PR separates `PlannerSpec` in multiple files based on what they are testing. 

Both `PlannerSpec` and `PlannerSql2ExactSpec` test plans given sql^2 but there's a different intention:

* the plans that are tested in `PlannerSql2ExactSpec` contain exact workflow expectations and are considered to be simple enough to be maintained by hand. These can later be refactored into proper unit tests by refactoring the sql^2 inputs into qscript inputs.
* the plans in `PlannerSpec` are more complex. The expectations of these tests won't be exact but will just verify whether the plans have certain properties (eg not using map-reduce).
For these tests we need an associated IT need to be created, to make sure they give correct results.


mongodb/src/test/scala/quasar/physical/mongodb/PlannerLPSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+  import StdLib.{set => s, _}
+  import structural._
+  import Grouped.grouped
+  import Reshape.reshape
+  import CollectionUtil._
+
+  import fixExprOp._
+  import PlannerHelpers._
+
+  "plan from LogicalPlan" should {
```
jsantos17: I wonder if these still make sense for the new Mongo connector. What do you think?

rintcius: Yeah, I was thinking we could translate these to be based on QScript instead, but wanted to do that later. I'll add a TODO if you agree

PR #3121, opened 2017-11-17
Automatically run `sideEffectTestFSConfig` before `testOnly`

jedesah:


djspiewak:
I'm 👎  on this.  The performance hit is going to be appreciable (though not the end of the world).  And ultimately, all of this is just a hack until we invert the dependency anyway.

jedesah:
@djspiewak What performance hit?

djspiewak:
@jedesah You have to recompute the dependency set (which is a non-trivial bit of work) on each run of `testOnly`.  It doesn't get cached.  This is less relevant for extremely infrequent things like `repl/run` and `http/run`, but still noticeable even there.  It would be killer on workflows that run `it/testOnly` frequently.

jedesah:
@djspiewak My quick tests in the `sbt` repl told me that this work was not noticeably slow but still would the last commit I just pushed solve the problem you mention?

djspiewak:
@jedesah The `.value` macro isn't doing what you think it is.  The performance of your change is identical to the original performance.

As to the exact hit, it depends a lot on the system (especially the filesystem itself; e.g. BTRFS, EXT4, APFS, etc), and also the state of your kernel cache.

jsantos17:
I defer to @djspiewak since he implemented this.

djspiewak:
If everyone really really wants this, I'll stand aside, but I don't think it's a good idea.

alissapajer:
`it/sideEffectTestFSConfig` requires a compilation of everything depended on by `it`. If I am running `testOnly` on a project earlier in the dependency chain, I don't want to have to wait for everything upstream of me to compile to run my tests.

alissapajer:
Or actually, it requires compilation of the connectors, not actually `it`, from what I'm seeing now.

djspiewak:
@alissapajer I believe this only affects `it/testOnly`

alissapajer:
@djspiewak ah, you're right. The change is just in the `it` portion of the `build.sbt`.

rintcius:
just tested running it a couple of times - it's pretty slow for me the first runs:

```
> it/sideEffectTestFSConfig
[success] Total time: 96 s, completed Nov 20, 2017 9:58:29 PM
> it/sideEffectTestFSConfig
[success] Total time: 21 s, completed Nov 20, 2017 9:59:19 PM
> it/sideEffectTestFSConfig
[success] Total time: 5 s, completed Nov 20, 2017 9:59:26 PM
> it/sideEffectTestFSConfig
[success] Total time: 1 s, completed Nov 20, 2017 9:59:30 PM
```

djspiewak:
@rintcius What you're seeing there is the cache warming up.  I would guess that if you did this test with a bunch of other things interleaved (other tasks on your system, or within SBT), you would see it pretty consistently take closer somewhere between the first and second run.

jedesah:
@rintcius Wow, that's pretty slow. The slowest I saw was 6 seconds which seemed like peanuts compared to our build times.

@djspiewak I'd like to better understand why you say the second commit doesn't basically make this task instant if the java property is already set. I mean, if this is true, there is something I'm missing here and I'd like to understand what it is.

@rintcius Did you test against the PR branch (after second commit) or `master`? Because if it's the PR branch (after the second commit) then looks like what @djspiewak is saying is true and I'd really want to understand why that's how it behaves. If only for my own curiosity even if there is nothing we can do about it.

rintcius:
@jedesah that was on `master`

jedesah:
@rintcius Glad to hear! Could you give the PR branch a quick whirl since your system appears slower than mine for some reason! Since my second run is 1 second even before the second PR it's hard for me to tell whether the change is behaving as expected or it's just fast on my computer regardless once the cache is warm. My expectation on the PR branch would be for it to take 1-2 seconds on your computer after the first run

rintcius:
@jedesah yes, correct. This is run from your branch:

```
> it/sideEffectTestFSConfig
[success] Total time: 41 s, completed Nov 21, 2017 9:43:23 AM
> it/sideEffectTestFSConfig
[success] Total time: 1 s, completed Nov 21, 2017 9:43:27 AM
> it/sideEffectTestFSConfig
[success] Total time: 1 s, completed Nov 21, 2017 9:43:30 AM
```

rintcius:
@djspiewak @jedesah 
When trying to simulate interleaved usage on this branch, the command takes somewhere between 5 and 20 secs. Not sure if that's because of `sbt` or this command specifically, but overhead looks quite substantial.

alissapajer:
I'm going to close this since there is a fair bit of concern about overhead, especially with interleaved usage. If there is still more discussion to be had, feel free to reopen or open an issue.

PR #3127, opened 2017-11-20
mongo join tests

### LINKED ISSUE #3080: Add unit tests for mongo join plans
Unit tests give clarity whether issues are in mongo connector or in an earlier stage

rintcius:
* add structure `PlanSpec` for mongo planner tests that allows to specify `sql2`, `qscript` and `workflow` among others
* add a function that tests `sql2` -> `qscript`, `qscript` -> `workflow` and `sql2` -> `workflow` from this structure. In case of `notOnPar` this gives insight into whether the problem is in mongo connector or in a stage before that
* rewrite 2 basic join tests using this structure
* rewrite some `sql2` in `PlannerSpec` so that it works using IT data without changes
* add a couple of more complex join tests in `PlannerQScriptSpec` (some of these are probably too fragile and will be refactored further while addressing #3090)

Fixes #3080


rintcius:
```
Error while downloading https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.11/1.0.0-RC13/coursier-cli_2.11-1.0.0-RC13-standalone.jar: C:\Users\appveyor\AppData\Local\Coursier\cache\v1\.structure.lock (The system cannot find the path specified), ignoring it
```

@djspiewak Can you restart Appveyor please? Hopefully this error is just incidental

djspiewak:
@rintcius You should be able to restart it now, once you log in.  Can you check?

rintcius:
@djspiewak I can login, but don't see any rebuild button. Bit similar similar to https://github.com/appveyor/ci/issues/1643 but I did choose `slamdata`. Does it work ok for you?

djspiewak:
Oh interesting.  I don't see a rebuild button either.  @zeeiyerWork can you look at permissions stuff?

zeeiyerWork:
Hmm- this is what i see? 
image

@rintcius, @djspiewak -> y'all don't see the buttons on the top right?

rintcius:
@zeeiyerWork I see only `LOG` there

zeeiyerWork:
@rintcius -> Please retry and lmk.

rintcius:
For reference: it works after @zeeiyerWork created new user for me

zeeiyerWork:
yeah i'll followup with appveyor - as per docs, should be no need to add specific users as administrators, when the entire github team is already added in. could be a bug on their end.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSql2ExactSpec.scala
```diff
@@ -51,9 +55,113 @@ class PlannerSql2ExactSpec extends
...
+            $cond(
+              $and(
+                $lte($literal(Bson.Doc()), $field(JoinDir.Right.name)),
+                $lt($field(JoinDir.Right.name), $literal(Bson.Arr(Nil)))),
+              $field(JoinDir.Right.name, "year"),
+              $literal(Bson.Undefined))),
+          ExcludeId)))
+  )
+
+  for (s <- specs) {
```
jsantos17: Non-review comment: Would something like `specs.traverse(testPlanSpec(_))` work?

rintcius: It works e.g. if we'd change the signature of `testPlanSpec` to `PlanSpec => Id[Fragment]` to fit the `f` in: `def traverse[G[_]:Applicative,A,B](fa: F[A])(f: A => G[B]): G[F[B]]`

rintcius:
@jsantos17 I also merged those 2 `PendingWithActualTracking` traits into 1. This avoids having to change the code in 2 places in case of using it in write mode.

foundation/src/test/scala/quasar/contrib/specs2/PendingWithActualTracking.scala
```diff
@@ -51,7 +76,14 @@ trait PendingWithActualTracking {
       case other =>
         Failure(m + " Behaviour has changed. Please review the test and set new expectation. Test result is: " + other)
     }
+
+    def pendingWithActualWriteMode(m: String, file: JFile): Result = ResultExecution.execute(AsResult(t)) match {
+      case f @ Failure(msg, e, stackTrace, FailureDetails(actual, expected)) =>
```
jsantos17: Do we need to bind `f` here? Or `msg`, `stackTrace`, and `expected`.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -40,6 +40,9 @@ class PlannerSpec extends
     PlannerWorkflowHelpers with
     PendingWithActualTracking {
 
+  //to write the new actuals:
+  //override val mode = WriteMode
```
jsantos17: :+1:

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSql2ExactSpec.scala
jsantos17: Oh so it's just `spec.map(testPlanSpec(_))`? `for` expressions in Scala confuse me...

jsantos17:
@zeeiyerWork any news on the AppVeyor? Still can't restart the rebuild. Would you mind restarting the build too?

rintcius:
@jsantos17 I got the same error a couple of times, I guess something wrong in this env, so a restart will probably not help in this case.

jsantos17:
@rintcius :(


You got the same `Error while downloading: https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.11/1.0.0-RC13/coursier-cli_2.11-1.0.0-RC13-standalone.jar`? That seem very transient given the file is there now. I have ran into issues with Sonatype mirroring taking some time to update before. I wonder what's going on.

At least access to AppVeyor would help us troubleshoot the build.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSql2ExactSpec.scala
rintcius: yes, maybe there are some subtle differences in general, but here `spec.map(testPlanSpec(_))` should work too

PR #3130, opened 2017-11-20
Reduction Coalescence

### LINKED ISSUE #3091: Implement autojoin elimination by freemap regions
Should find `AutoJoin2`, `AutoJoin3` and `Map` nodes and run the `MappableRegion` function on the holes.  If the regions are a shared source, then the result is trivially a `Map`.  If the regions are not a shared source, then heuristics should be applied to see if the sources can be merged.

- Reductions with the same provenance
- etc...?

If cannot be merged, then still create a `Map` node above the `AutoJoin` and leave it there for later rewrites (this avoids having to re-compute the mappable regions)

### LINKED ISSUE #3129: Extract mappable region from reduce source into reducer
`QSReduce` nodes expect to apply buckets and reducers to a common source, which means we need to extract the mappable region from the source of an `LPReduce` and push it into the reducer, otherwise the buckets are applied to the value to be reduced and not its source.

djspiewak:
Depends on #3128
Fixes #3129 (for anything being coalesced anyway)
Fixes #3091

There are some caveats to the current implementation.  In particular, it's *very* aggressive about pulling source `FreeMap`s into reduce functions, which is not necessarily optimal (depending on connector implementation).  It should already do better than what we have in master though, because of the `Filter` => `Cond` rewriting.


alissapajer:
@djspiewak merge conflict now. Also, merging in master should remove the commits from the PR this depends on.

PR #3137, opened 2017-11-21
All my things

djspiewak:
Merge this one, not the others.  :-D

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/LPtoQS.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/LPtoQS.scala
@@ -62,15 +62,21 @@ final class LPtoQS[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes[T] {
   @SuppressWarnings(Array("org.wartremover.warts.ToString"))
   private def debugG[F[_]: Applicative](prefix: String): K[F, QSUGraph, QSUGraph] =
     K { g =>
-      println("\n\n" + prefix + g.shows)    // uh... yeah do better
+      maybePrint("\n\n" + prefix + g.shows)    // uh... yeah do better
       g.point[F]
     }
 
   private def debugAG[F[_]: Applicative](prefix: String): K[F, AuthenticatedQSU[T], AuthenticatedQSU[T]] =
     K { aqsu =>
-      println("\n\n" + prefix + aqsu.graph.shows + "\n" + aqsu.dims.shows)
+      maybePrint("\n\n" + prefix + aqsu.graph.shows + "\n" + aqsu.dims.shows)
       aqsu.point[F]
     }
+
+  private def maybePrint(str: String): Unit = {
+    // println(str)
+
+    ()
+  }
 }
 
 object LPtoQS {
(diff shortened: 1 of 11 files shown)


connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -16,103 +16,272 @@
...
+  private def updateGraph[
+      G[_]: Monad: NameGenerator: MonadState_[?[_], RevIdx]: MonadState_[?[_], QSUDims[T]]: PlannerErrorME](
+      pat: QScriptUniform[Symbol]): G[QSUGraph] = {
+
+    for {
+      qgraph <- QSUGraph.withName[T, G](pat)
 
-    minimized
+      dims <- MonadState_[G, QSUDims[T]].get
+      computed <- AP.computeProvenanceƒ[G].apply(QSUGraph.QSUPattern(qgraph.root, pat.map(dims)))
```
wemrysi: This is necessary, but I don't think it is sufficient. Other vertices downstream from one of the merged reductions will have dimensions referencing the bucket(s) of that reducer. Those dimensions either need to be remapped to refer to the buckets of the merged reducer or just recomputed, which will have the same effect.

connector/src/main/scala/quasar/qscript/qsu/LPtoQS.scala
```diff
@@ -63,15 +63,21 @@ final class LPtoQS[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes[T] {
...
 
   private def debugAG[F[_]: Applicative](prefix: String): K[F, AuthenticatedQSU[T], AuthenticatedQSU[T]] =
     K { aqsu =>
-      println("\n\n" + prefix + aqsu.graph.shows + "\n" + aqsu.dims.shows)
+      maybePrint("\n\n" + prefix + aqsu.graph.shows + "\n" + aqsu.dims.shows)
       aqsu.point[F]
     }
+
+  private def maybePrint(str: String): Unit = {
+    // println(str)
```
wemrysi: Thanks for cleaning up the mess I left :cold_sweat:

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -16,103 +16,272 @@
...
-      case minimized =>
-        // TODO implement coalesce heuristics here
+    case candidates =>
+      val reducerAttempt = candidates collect {
+        case g @ QSReduce(source, buckets, reducers, repair) =>
+          (source, buckets, reducers, repair)
+      }
+
+      // candidates.forall(_ ~= QSReduce)
+      if (reducerAttempt.lengthCompare(candidates.length) === 0) {
```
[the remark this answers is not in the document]
djspiewak: I thought about it, but we actually have the guarantee not only that the list is non-empty, but it's actually length >= 2.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
djspiewak: The buckets of the merged reducer will be equivalent to the buckets of the unmerged reducer, by definition.

wemrysi: True, but for two reducer nodes, `'a` and `'b`, prior to the merge each would have dimensions like `[Value(Bucket('a, 0)), Proj(foo)]` and `[Value(Bucket('b, 0)), Proj(foo)]` so if `'b` was merged into `'a` then it's bucket references by dimensions of other vertices need to be remapped to `'a`.

`QProv.rename` should do the trick here.

djspiewak: Ah interesting. I’ll fix that

jsantos17:
Is there anything missing to merge this? It fixes a bunch of Mongo planner issues and I need it to start fixing the tests for Mongo in `qsu`.

PR #3142, opened 2017-11-21
Modify value access in reduction buckets on upper freemap coalesce

djspiewak:
Depends on #3139 

Title.


djspiewak:
Don't merge yet.  Something is weird…

PR #3146, opened 2017-11-22
[#3129] Compute the correct source for QSReduce

### LINKED ISSUE #3129: Extract mappable region from reduce source into reducer

wemrysi:
Fixes #3129.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/ApplyProvenance.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ApplyProvenance.scala
@@ -22,7 +22,7 @@ import quasar.Planner.{PlannerErrorME, InternalError}
 import quasar.contrib.scalaz.MonadState_
 import quasar.ejson
 import quasar.ejson.EJson
-import quasar.qscript.{ExcludeId, HoleF, IdOnly, IdStatus, RightSideF, SrcHole}
+import quasar.qscript.{ExcludeId, HoleF, IdOnly, IdStatus, RightSideF}
 import quasar.qscript.provenance._
 
 import matryoshka._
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/quasar/qscript/qsu/ApplyProvenance.scala
```diff
@@ -56,78 +56,80 @@ final class ApplyProvenance[T[_[_]]: BirecursiveT: EqualT] {
...
 
-      case Read(file) => dims.squash(segments(file).map(projStr).reverse).point[F]
+        case Subset((_, from), _, (_, count)) => dims.join(from, count).point[F]
 
-      case Subset(from, _, count) => dims.join(from, count).point[F]
+        case ThetaJoin((_, left), (_, right), _, _, _) => dims.join(left, right).point[F]
 
-      case ThetaJoin(left, right, _, _, _) => dims.join(left, right).point[F]
+        case Transpose((_, src), _, rot) =>
+          val tid: dims.I = Free.pure(Access.identity(root, root))
```
alissapajer: Can you explain this change here, going from `Access.identity(root, SrcHole)` to `Access.identity(root, root)`?

wemrysi: The identity type changed from `FreeMapA[Access[Hole]]` to `FreeMapA[Access[Symbol]]` so that we could track the name for value access as well. For identity and bucket, it just means the name happens to be duplicated in the `id` and `src` fields in this case.

PR #3149, opened 2017-11-22
Working unit tests in QSU

djspiewak:
Attention @jsantos17 and @rintcius!  This will create some (hopefully trivial) merge conflicts with what you're doing.


djspiewak:
@alissapajer I expect all the unit tests to pass.  Integration tests should run in travis.  I expect ITs to fail on every connector except mimir, but the failures should be relatively confined.  We might get a few passes.

PR #3151, opened 2017-11-22
Implemented reading of Union nodes from LP

djspiewak:
Depends on #3149 

Spark integration tests spring to life with this change.


alissapajer:
I'm expecting mimir and spark builds to be green with this PR. Going to wait on merge until all builds have run now that we have builds that are actually green :)

PR #3160, opened 2017-11-22
Download directories containing modules

jedesah:
- Add notion of `Node` in Filesystem to convey more information than simply `PathSegment`
- Simplify a few tests I happened to touch to use some new methods (saveThese)
- When downloading a directory, ignore any `Function` files so as to not throw an exception attempting to read from such files


jedesah:
So what are we doing about Appveyor build failures?

connector/src/main/scala/quasar/connector/BackendModule.scala
```diff
@@ -110,7 +110,8 @@ trait BackendModule {
           explanation <- QueryFileModule.explain(pp.repr)
         } yield ExecutionPlan(Type, explanation, pp.paths)).run.run
 
-      case QueryFile.ListContents(dir) => QueryFileModule.listContents(dir).run.value
+      case QueryFile.ListContents(dir) =>
+        QueryFileModule.listContents(dir).run.value.map(_.map(set => set.map(Node.fromSegment)))
```
jsantos17: I think it'd be slightly cleaner if you mapped `fromSegment` before running the transformer. Something like:

```
QueryFileModule.listContents(dir).map(set => set.map(Node.fromSegment)).run.value
```

core/src/main/scala/quasar/fs/mount/Mounting.scala
```diff
@@ -112,10 +112,18 @@ object Mounting {
       lookupConfig(path).flatMap(config =>
         EitherT.right(OptionT(viewConfig.getOption(config).map(ViewConfig.tupled).point[FreeS])))
 
+    def lookupViewConfigIgnoreError(path: AFile): OptionT[FreeS, ViewConfig] =
+      lookupViewConfig(path).run
```
jsantos17: I believe you can replace this with `OptionT(lookupViewConfig(path).toOption.run.run.map(_.join))`. Would be even nicer if we had `squash`. It'd be `lookupViewConfig(path).toOption.squash`.

Just a suggestion, change it if you want.

connector/src/main/scala/quasar/connector/BackendModule.scala
jedesah: Agreed! Thanks for pointing that out!

connector/src/main/scala/quasar/fs/package.scala
```diff
@@ -37,6 +39,52 @@ package object fs extends PhysicalErrorPrisms {
     def apply[F[_]](implicit F: MonadFsErr[F]): MonadFsErr[F] = F
   }
 
+  trait Node {
```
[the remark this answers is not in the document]
jedesah: Well, my thinking was that a connector can add an additional `Node` type in a typed manner. I don't know if that's worth the loss of exhaustivity checking though... What do you think?

core/src/main/scala/quasar/fs/mount/Mounting.scala
jedesah: Added `squash`. It sucks that (as I far as I can reason) it's not easy to make it generic in Scala but seems like it's worth defining it specifically for `OptionT`. Although I didn't look for them, I wouldn't be surprised if there were other parts of the codebase that could make use of `squash`

jsantos17:
@jedesah Approved. I'll merge this after conflicts are resolved.

PR #3161, opened 2017-11-23
Missed some test failures

djspiewak:
🐭🔨


alissapajer:
This is the official "green" PR. We won't merge it until all the builds are green, and from here on out all builds on all PRs to qsu must be green before merge!

alissapajer:
Keeping an eye on this, will merge when green!

Issue #3021, opened 2017-11-03
Provide mappable access to provenance-changing constructs


alissapajer:
In the `QScriptUniform` cases of `LPFilter`, `GroupBy`, `LPJoin`, and `Sort` we are required to create a mappable structure from a `QSUGraph` (a `FreeMap` or a `JoinFunc`) in order to construct `QScriptEducated`. Currently we are restricting these positions to only contain "freemappable" constructs, but this is too narrow a restriction according to MRA.

This issue is to expand the allowed structures to contain provenance-changing and non-freemappable constructs.

For example, since `LeftShift` can provide access in the `repair` to both the identities and the values, we could move a `LeftShift` into the source and then provide `FreeMap` access to that. Or in other cases we may need to insert an actual join.

alissapajer:
This currently manifests as any of the errors produced from `ExtractFreeMap`.

alissapajer:
I'm closing this now that #3278 has been merged. 

The remaining bit of this is tracked in #3286.

PR #3165, opened 2017-11-23
[WIP] planner spec

rintcius:
Depends on #3161 

Reinstate tracking for `PlannerSpec` where possible (several tests don't compile qscript though)

Not for review/merge yet


mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -53,29 +53,44 @@ class PlannerSpec extends
...
+      "filter with both index and key projections",
+      plan(sqlE"""select count(parents[0].sha) as count from slamengine_commits where parents[0].sha = "56d1caf5d082d1a6840090986e277d36d03f1859" """),
+      IList(ReadOp, MatchOp, SimpleMapOp, GroupOp))
 
     "having with multiple projections" in {
       plan(sqlE"select city, sum(pop) from extraSmallZips group by city having sum(pop) > 40000") must
         beRight.which(cwf => notBrokenWithOps(cwf.op, IList(ReadOp, GroupOp, MatchOp, ProjectOp)))
+        // FIXME Fails with:
+        // [error] x having with multiple projections (5 seconds, 908 ms)
+        // [error]  'Left(QScriptPlanningFailed(InternalError(Invalid filter predicate, 'qsu14, must be a mappable function of 'qsu8.,None)))' is not Right (PlannerSpec.scala:65)
```
rintcius: @djspiewak I encountered a number of qscript errors, which I annotated with FIXME's in this diff. I guess some of them are known issues, but not sure which. Can you indicate which ones are not yet known?

The FIXME in line 91 is not qscript related

djspiewak: Any qscript error which looks like the above ("invalid filter predicate" or "invalid group key") is already known, and @alissapajer is working on it.  Any *other* sorts of errors aren't known, and we want to hear about them.

alissapajer: Yes. The full list of errors which are covered in #3021 (which I'm currently working on) is:
- "Invalid filter predicate"
- "Invalid group key"
- "Invalid sort key"
- "Invalid join condition"

rintcius: @alissapajer @djspiewak ok thanks, here are 3 others. I'll create separate issues for these :

```
+    // FIXME errors with
 +    // [error]     java.lang.IndexOutOfBoundsException: 1
 +    // [error]     	at scala.collection.LinearSeqOptimized$class.apply(LinearSeqOptimized.scala:65)
 +    // [error]     	at scala.collection.immutable.List.apply(List.scala:84)
 +    // [error]     	at scala.collection.immutable.List.apply(List.scala:84)
 +    // [error]     	at scalaz.$bslash$div.fold(Either.scala:57)
 +    // [error]     	at quasar.qscript.analysis.OutlineInstances$$anon$3.quasar$qscript$analysis$OutlineInstances$$anon$3$$$anonfun$24(Outline.scala:263)
 +    // trackPending(
 +    //   "group by simple expression",
 +    //   plan(sqlE"select city, sum(pop) from extraSmallZips group by lower(city)"),
 +    //   IList(ReadOp, GroupOp, UnwindOp))
```

and 

```
    "expr3 with grouping" in {
        plan(sqlE"select case when pop > 1000 then city else lower(city) end, count(*) from zips group by city") must
          beRight
 +        //FIXME fails with: an implementation is missing (ReifyIdentities.scala:358)
      }.pendingUntilFixed
```

and there's a dangling reference for `select distinct loc || [pop - 1] as p1, pop - 1 as p2 from zips group by territory order by p2` which could either be a consequence of incorrect qscript or something in mongo

alissapajer: @rintcius awesome, individual issues for those are great! Thanks!

PR #3166, opened 2017-11-23
Rdbms MapFunc planner

kciesielski:


.travis.yml
```diff
@@ -29,6 +29,7 @@ env:
     - CONNECTOR=mongodb_read_only
     - CONNECTOR=spark_local_test
     - CONNECTOR=spark_hdfs
+    - CONNECTOR=postgres
```
rabbitonweb: this should be removed for now

rdbms/src/main/scala/quasar/physical/rdbms/planner/MapFuncCorePlanner.scala
```diff
@@ -0,0 +1,156 @@
...
+    case MFC.ToString(f) =>  notImplemented("ToString", this)
+    case MFC.Search(fStr, fPattern, fInsen) => notImplemented("Search", this)
+    case MFC.Substring(fStr, fFrom, fCount) => notImplemented("Substring", this)
+    case MFC.Split(fStr, fDelim) => notImplemented("Split", this)
+    case MFC.MakeArray(f) =>  notImplemented("MakeArray", this)
+    case MFC.MakeMap(key, value) =>
+      key.project match {
+              case Constant(Data.Str(keyStr)) =>
+                SQL.ExprWithAlias[T[SQL]](value, keyStr).embed.η[F]
+              case other =>
```
rabbitonweb: @wemrysi @alissapajer not part of this review, but general question, can `key` be something different then `String`? If yes, what should happen here? From what I can see from, we only expect here `Data.Str`

.travis.yml
mikolak-net: Done.

rdbms/src/main/scala/quasar/physical/rdbms/planner/MapFuncCorePlanner.scala
wemrysi: @rabbitonweb Yes, `key` can be any type could be another map, an array, int, etc. It is connector dependent as to what keys are actually supported efficiently. You can try and emulate support for keys that don't have native support or you can produce an error stating such a key type isn't supported.

Strings seem to be the most popular, so making sure those work performantly is high-priority.

rabbitonweb: ok, well, spark only supports strings, i wonder how much big of an issue that is

PR #3167, opened 2017-11-23
RDBMS Sort planner

kciesielski:


rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -47,11 +50,42 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
...
+      }
+
+      def orderByPushdown(in: T[SqlExpr]): F[T[SqlExpr]] = {
+        in.project match {
+          case s @ SqlExpr.SelectRow(_, from, _) =>
+            for {
+              orderByExprs <- order.traverse(createOrderBy(from.alias))
+              bucketExprs <- bucket.map((_, orderByExprs.head.sortDir)).traverse(createOrderBy(from.alias))
+            }
+              yield s.copy(orderBy = bucketExprs ++ orderByExprs.toList).embed
```
rabbitonweb: shouldnt that be connected with existing list of `orderBy`s rather then replacing it

rabbitonweb:
@wemrysi false alarm, this is not yet ready for review

kciesielski:
The pushdown scheme is incorrent, this will fail for queries like `select a as c from (select a from foo) as xxx order by c`

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -31,27 +33,58 @@ import matryoshka.patterns._
...
+  def processFreeMap(f: FreeMap[T],
+                     alias: SqlExpr.Id[T[SqlExpr]]): F[T[SqlExpr]] =
     f.cataM(interpretM(κ(alias.embed.η[F]), mapFuncPlanner.plan))
 
   def plan: AlgebraM[F, QScriptCore[T, ?], T[SqlExpr]] = {
     case qscript.Map(src, f) =>
       for {
-        generatedAlias <- genId[T[SqlExpr], F]
-        selection <- processFreeMap(f, generatedAlias)
+        fromAlias <- genId[T[SqlExpr], F]
```
rabbitonweb: As discussed, this can be simplified

Issue #3144, opened 2017-11-22
Appveyor build failure


rintcius:
Causes lots of Appveyor builds to fail right at the start:

```
Build started
git clone -q https://github.com/quasar-analytics/quasar.git C:\projects\quasar
git fetch -q origin +refs/pull/3138/merge:
git checkout -qf FETCH_HEAD
IF "%APPVEYOR_PULL_REQUEST_NUMBER%"=="" (.\sbt clean)
.\sbt ++"2.11.8" test:compile
BITSADMIN version 3.0 [ 7.7.9600 ]
BITS administration utility.
(C) Copyright 2000-2006 Microsoft Corp.
BITSAdmin is deprecated and is not guaranteed to be available in future versions of Windows.
Administrative tools for the BITS service are now provided by BITS PowerShell cmdlets.
Transfer complete.
Error while downloading https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.11/1.0.0-RC13/coursier-cli_2.11-1.0.0-RC13-standalone.jar: C:\Users\appveyor\AppData\Local\Coursier\cache\v1\.structure.lock (The system cannot find the path specified), ignoring it
Command exited with code 1
```

Restarting the build doesn't help

jedesah:
Anyone object to merging PRs even if AppVeyor is failing for the time being.

This has happened on every single of my PRs pretty much.

rintcius:
Same here. 

Yeah, I think most people are ignoring AppVeyor builds right now

PR #3169, opened 2017-11-23
Serialize module SQL properly so it can be parsed again

### LINKED ISSUE #3027: Module mounts with escaped identifiers as arguments fail to mount
This kind of thing:
``` sql
CREATE FUNCTION testModule(:`test`) BEGIN 
    SELECT :`test` 
END
```
Results in:
``` json
{
  "error": {
    "status": "Invalid mount.",
    "detail": {
      "message": "error while obtaining a mount config for CREATE FUNCTION card3633169b02d9448ab6dc3ab7ffed9964(:d.test)\n  BEGIN\n    (select :`d.test`)\n  END, operator ')' expected; but found `.'"
    }
  }
}
```
In the case it's occurring we actually have a name that does need escaping rather than `test`. 

This issue perhaps deserves a skull, since it's something we can't really work around in the frontend (well... maybe, we could apply some kind of name mangling, but that has its own problems), but is also needed to get it to behave correctly in some situations.

jedesah:
Fixes #3027 

- escape function param names that need to be escaped
- Add round-trip test for modules
- Add test to make sure we parse function with escaped param names properly
- Rename `_qq` to `escape`
- Add `Equal` instance to `ParsingError`
- Add convenience `pprintF` on `Statement` for common use case

### DIFF
--- a/frontend/src/main/scala/quasar/sql/ParsingError.scala
+++ b/frontend/src/main/scala/quasar/sql/ParsingError.scala
@@ -31,4 +31,9 @@ final case class ParsingPathError(error: PathError) extends ParsingError {
 
 object ParsingError {
   implicit val parsingErrorShow: Show[ParsingError] = Show.showFromToString
+  implicit val equal: Equal[ParsingError] = Equal.equal {
+    case (GenericParsingError(m1), GenericParsingError(m2)) => m1 ≟ m2
+    case (ParsingPathError(e1), ParsingPathError(e2))       => e1 ≟ e2
+    case _                                                  => false
+  }
 }
--- a/core/src/test/scala/quasar/fs/mount/MountConfigSpec.scala
+++ b/core/src/test/scala/quasar/fs/mount/MountConfigSpec.scala
@@ -19,8 +19,12 @@ package quasar.fs.mount
 import slamdata.Predef._
 import quasar.{Variables, VarName, VarValue}
 import quasar.sql._
+import quasar.sql.StatementArbitrary._
+
+import scala.Predef.$conforms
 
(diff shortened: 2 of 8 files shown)


frontend/src/main/scala/quasar/sql/ParsingError.scala
```diff
@@ -31,4 +31,9 @@ final case class ParsingPathError(error: PathError) extends ParsingError {
 
 object ParsingError {
   implicit val parsingErrorShow: Show[ParsingError] = Show.showFromToString
+  implicit val equal: Equal[ParsingError] = Equal.equal {
+    case (GenericParsingError(m1), GenericParsingError(m2)) => m1 ≟ m2
+    case (ParsingPathError(e1), ParsingPathError(e2))       => e1 ≟ e2
+    case _                                                  => false
```
jedesah: Is there a better way of doing this? I can't think of one but I don't think this is ideal, one could easily forget to add another case to the pattern match when adding a new type of `ParsingError`

djspiewak: Yeah I don’t know of a better way off hand. You could fold, but it would be really ugly.

jedesah:
I'm supposed to have permission to restart AppVeyor now right? I can't see anything in the UI though despite being logged in...

rintcius:
@jedesah - you have to be readded explicitly (by @zeeiyerWork ). Note that restarting won't work in this case, see #3144 (last line)

zeeiyerWork:
@jedesah - (jr@slamdata.com) exists on the appveyor account - sent you pwd on slack.
image

PR #3179, opened 2017-11-24
Added basic LPtoQSSpec

### LINKED ISSUE #3097: Add an LPtoQSSpec
Shouldn't put a ton of tests in here (they will be too fragile), but it would be good to see the whole read, rewrite, and graduation process work from front to back without having to plug it into a repl.  Every test case is expected to be of the form input logical plan => output qscript.

djspiewak:
Fixes #3097


djspiewak:
The marklogic_json failure is fixed by #3181.  The appveyor failure (should be) fixed by #3185

drostron:
Will merge once associated issues are resolved via a merge from master.

djspiewak:
@drostron Ready

PR #3180, opened 2017-11-24
Regression tests for various queries

### LINKED ISSUE #1929: Can't use subquery when there's a group by statement
I can't use a SELECT statement within another when a GROUP BY is involved.

For example, this query returns the total population:

SELECT SUM(pop) FROM zips

This query returns the population per state:

SELECT SUM(pop), state FROM zips GROUP BY state

I want a table with the population per state divided by the total population, so I can present the data in percentages. So I tried this:

This fails with the following error. Am I doing something wrong?

```
💪 $ SELECT 100 * SUM(pop) / (SELECT SUM(pop) FROM zips), state FROM zips GROUP BY state
Logical Plan (projections preferred):
Let('__tmp0)
├─ Read(/tutorial/zips)
╰─ Let('__tmp1)
   ├─ Invoke(GroupBy)
   │  ├─ Free('__tmp0)
   │  ╰─ Invoke(MakeArray)
   │     ╰─ ObjectProject('__tmp0{"state"})
   ╰─ Invoke(Take)
      ├─ Invoke(Squash)
      │  ╰─ Invoke(ObjectConcat)
      │     ├─ Invoke(MakeObject)
      │     │  ├─ Str("0")
      │     │  ╰─ Invoke(Divide)
      │     │     ├─ Invoke(Multiply)
      │     │     │  ├─ Constant(Int(100))
      │     │     │  ╰─ Invoke(Sum)
      │     │     │     ╰─ Let('__check1)
      │     │     │        ├─ ObjectProject('__tmp1{"pop"})
      │     │     │        ╰─ Typecheck(Int, or Dec)
      │     │     │           ├─ Free('__check1)
      │     │     │           ├─ Free('__check1)
      │     │     │           ╰─ Constant(NA)
      │     │     ╰─ Invoke(Squash)
      │     │        ╰─ Invoke(Sum)
      │     │           ╰─ Let('__check3)
      │     │              ├─ ObjectProject(/tutorial/zips{"pop"})
      │     │              ╰─ Typecheck(Int, Dec, or Interval)
      │     │                 ├─ Free('__check3)
      │     │                 ├─ Free('__check3)
      │     │                 ╰─ Constant(NA)
      │     ╰─ Invoke(MakeObject)
      │        ├─ Str("state")
      │        ╰─ Invoke(Arbitrary)
      │           ╰─ ObjectProject('__tmp1{"state"})
      ╰─ Constant(Int(11))

Quasar error: failed to merge:
GroupBuilder
├─ ExprBuilder
│  ├─ CollectionBuilder(Root())
│  │  ├─ $ReadF(tutorial; zips)
│  │  ╰─ Schema
│  ╰─ ExprOp({
│            "$cond": [
│              {
│                "$and": [
│                  { "$lt": [{ "$literal": null }, "$pop"] },
│                  { "$lt": ["$pop", { "$literal": "" }] }]
│              },
│              "$pop",
│              { "$literal": undefined }]
│          })
├─ By
│  ╰─ ExprBuilder
│     ├─ CollectionBuilder(Root())
│     │  ├─ $ReadF(tutorial; zips)
│     │  ╰─ Schema
│     ╰─ ExprOp("$state")
╰─ Content
   ╰─ -\/
      ╰─ AccumOp({ "$sum": "$$ROOT" })
GroupBuilder
├─ ExprBuilder
│  ├─ CollectionBuilder(Root())
│  │  ├─ $ReadF(tutorial; zips)
│  │  ╰─ Schema
│  ╰─ ExprOp({
│            "$cond": [
│              {
│                "$and": [
│                  { "$lt": [{ "$literal": null }, "$pop"] },
│                  { "$lt": ["$pop", { "$literal": "" }] }]
│              },
│              "$pop",
│              { "$literal": undefined }]
│          })
├─ By
╰─ Content
   ╰─ -\/
      ╰─ AccumOp({ "$sum": "$$ROOT" })
```

### LINKED ISSUE #2227: Pie chart renders incorrectly if SQL Query has aggregate, group by and order by
When building out a pie chart based on the following SQL query:

```sql
select
        count(*) as cnt
    ,   city
from
        `/marklogiclocal/patients`
where
        state = "NY"
group by
        city
order by
        count(*) desc
limit
        10
```

the pie chart is displayed incorrectly. It has just a single city with a sum of all the cities as it's value as shown below. It is expected to have the top 10 cities by number of records based off city.

incorrect pie chart

We can see that the query works and returns data as expected if we put a **Preview Table** card after the query:

correct preview table

Here is the query string parameters that are being passed in to the chart:

```
q=SELECT%20city%20AS%20category%2C%20SUM(cnt)%20AS%20measure%2C%20null%20AS%20donut%2C%20null%20AS%20parallel%20FROM%20%60%2Fmarklogiclocal%2Fpatients%20workspaces%2FUntitled%20Workspace%201.slam%2F.tmp%2Fout741f6a36-03d6-42c0-9ad7-c2d9be12e1dc%60%20GROUP%20BY%20city%20ORDER%20BY%20city%20ASC
```

Looking at the output of this, there are 1000 rows, and they are all the same, showing a `category` of **NEW YORK** and a `measure` of **44** which is the sum of the counts for the 10 rows.

This works in Old Mongo connector, but does not work in MarkLogic or Spark. When the original query is run in Couchbase we get an error stating `Failed to plan query.: unimplemented EquiJoin: Not currently mapped to N1QL's key join` which is a known issue (Quasar issue #2117 https://github.com/quasar-analytics/quasar/issues/2117)

### LINKED ISSUE #2775: Explosively horrible qscript from structure editor query
SQL:

```sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t2, "c"), {"test": OBJECT_CONCAT(t2.test, {"c": OBJECT_CONCAT(t2.test.c, t2.c) ?? t2.c}) ?? {"c": t2.c}}) ?? t2 FROM (SELECT OBJECT_CONCAT(DELETE_FIELD(t1, "b"), {"test": OBJECT_CONCAT(t1.test, {"b": OBJECT_CONCAT(t1.test.b, t1.b) ?? t1.b}) ?? {"b": t1.b}}) ?? t1 FROM (SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "a"), {"test": OBJECT_CONCAT(t0.test, {"a": OBJECT_CONCAT(t0.test.a, t0.a) ?? t0.a}) ?? {"a": t0.a}}) ?? t0 FROM `/spark-local/structure_test_deep.json` AS t0) AS t1) AS t2
```

Logical plan:

```
Let('__tmp0)
├─ Let('__tmp1)
│  ├─ Read(/spark_local/zips)
│  ╰─ Typecheck(Obj(Map[], Some(Top)))
│     ├─ Free('__tmp1)
│     ├─ Free('__tmp1)
│     ╰─ Constant(NA)
╰─ Let('__tmp2)
   ├─ Invoke(Squash)
   │  ╰─ Invoke(IfUndefined)
   │     ├─ Invoke(ObjectConcat)
   │     │  ├─ Invoke(DeleteField)
   │     │  │  ├─ Free('__tmp0)
   │     │  │  ╰─ Str("a")
   │     │  ╰─ Invoke(MakeObject)
   │     │     ├─ Str("test")
   │     │     ╰─ Let('__tmp3)
   │     │        ├─ ObjectProject(Free('__tmp0){"test"})
   │     │        ╰─ Typecheck(Obj(Map[], Some(Top)))
   │     │           ├─ Free('__tmp3)
   │     │           ├─ Invoke(IfUndefined)
   │     │           │  ├─ Invoke(ObjectConcat)
   │     │           │  │  ├─ Free('__tmp3)
   │     │           │  │  ╰─ Invoke(MakeObject)
   │     │           │  │     ├─ Str("a")
   │     │           │  │     ╰─ Let('__tmp4)
   │     │           │  │        ├─ ObjectProject(Free('__tmp0){"a"})
   │     │           │  │        ╰─ Typecheck(Obj(Map[], Some(Top)))
   │     │           │  │           ├─ Free('__tmp4)
   │     │           │  │           ├─ Let('__tmp5)
   │     │           │  │           │  ├─ ObjectProject(Free('__tmp0){"test"})
   │     │           │  │           │  ╰─ Typecheck(Obj(Map[], Some(Top)))
   │     │           │  │           │     ├─ Free('__tmp5)
   │     │           │  │           │     ├─ Let('__tmp6)
   │     │           │  │           │     │  ├─ ObjectProject(Free('__tmp5){"a"})
   │     │           │  │           │     │  ╰─ Typecheck(Obj(Map[], Some(Top)))
   │     │           │  │           │     │     ├─ Free('__tmp6)
   │     │           │  │           │     │     ├─ Invoke(IfUndefined)
   │     │           │  │           │     │     │  ├─ Invoke(ObjectConcat)
   │     │           │  │           │     │     │  │  ├─ Free('__tmp6)
   │     │           │  │           │     │     │  │  ╰─ Free('__tmp4)
   │     │           │  │           │     │     │  ╰─ ObjectProject(Free('__tmp0){"a"})
   │     │           │  │           │     │     ╰─ Constant(NA)
   │     │           │  │           │     ╰─ Constant(NA)
   │     │           │  │           ╰─ Constant(NA)
   │     │           │  ╰─ Invoke(MakeObject)
   │     │           │     ├─ Str("a")
   │     │           │     ╰─ ObjectProject(Free('__tmp0){"a"})
   │     │           ╰─ Constant(NA)
   │     ╰─ Free('__tmp0)
   ╰─ Let('__tmp7)
      ├─ Invoke(Squash)
      │  ╰─ Invoke(IfUndefined)
      │     ├─ Invoke(ObjectConcat)
      │     │  ├─ Invoke(DeleteField)
      │     │  │  ├─ Free('__tmp2)
      │     │  │  ╰─ Str("b")
      │     │  ╰─ Invoke(MakeObject)
      │     │     ├─ Str("test")
      │     │     ╰─ Let('__tmp8)
      │     │        ├─ ObjectProject(Free('__tmp2){"test"})
      │     │        ╰─ Typecheck(Obj(Map[], Some(Top)))
      │     │           ├─ Free('__tmp8)
      │     │           ├─ Invoke(IfUndefined)
      │     │           │  ├─ Invoke(ObjectConcat)
      │     │           │  │  ├─ Free('__tmp8)
      │     │           │  │  ╰─ Invoke(MakeObject)
      │     │           │  │     ├─ Str("b")
      │     │           │  │     ╰─ Let('__tmp9)
      │     │           │  │        ├─ ObjectProject(Free('__tmp2){"b"})
      │     │           │  │        ╰─ Typecheck(Obj(Map[], Some(Top)))
      │     │           │  │           ├─ Free('__tmp9)
      │     │           │  │           ├─ Let('__tmp10)
      │     │           │  │           │  ├─ ObjectProject(Free('__tmp2){"test"})
      │     │           │  │           │  ╰─ Typecheck(Obj(Map[], Some(Top)))
      │     │           │  │           │     ├─ Free('__tmp10)
      │     │           │  │           │     ├─ Let('__tmp11)
      │     │           │  │           │     │  ├─ ObjectProject(Free('__tmp10){"b"})
      │     │           │  │           │     │  ╰─ Typecheck(Obj(Map[], Some(Top)))
      │     │           │  │           │     │     ├─ Free('__tmp11)
      │     │           │  │           │     │     ├─ Invoke(IfUndefined)
      │     │           │  │           │     │     │  ├─ Invoke(ObjectConcat)
      │     │           │  │           │     │     │  │  ├─ Free('__tmp11)
      │     │           │  │           │     │     │  │  ╰─ Free('__tmp9)
      │     │           │  │           │     │     │  ╰─ ObjectProject(Free('__tmp2){"b"})
      │     │           │  │           │     │     ╰─ Constant(NA)
      │     │           │  │           │     ╰─ Constant(NA)
      │     │           │  │           ╰─ Constant(NA)
      │     │           │  ╰─ Invoke(MakeObject)
      │     │           │     ├─ Str("b")
      │     │           │     ╰─ ObjectProject(Free('__tmp2){"b"})
      │     │           ╰─ Constant(NA)
      │     ╰─ Free('__tmp2)
      ╰─ Invoke(Take)
         ├─ Invoke(Squash)
         │  ╰─ Invoke(IfUndefined)
         │     ├─ Invoke(ObjectConcat)
         │     │  ├─ Invoke(DeleteField)
         │     │  │  ├─ Free('__tmp7)
         │     │  │  ╰─ Str("c")
         │     │  ╰─ Invoke(MakeObject)
         │     │     ├─ Str("test")
         │     │     ╰─ Let('__tmp12)
         │     │        ├─ ObjectProject(Free('__tmp7){"test"})
         │     │        ╰─ Typecheck(Obj(Map[], Some(Top)))
         │     │           ├─ Free('__tmp12)
         │     │           ├─ Invoke(IfUndefined)
         │     │           │  ├─ Invoke(ObjectConcat)
         │     │           │  │  ├─ Free('__tmp12)
         │     │           │  │  ╰─ Invoke(MakeObject)
         │     │           │  │     ├─ Str("c")
         │     │           │  │     ╰─ Let('__tmp13)
         │     │           │  │        ├─ ObjectProject(Free('__tmp7){"c"})
         │     │           │  │        ╰─ Typecheck(Obj(Map[], Some(Top)))
         │     │           │  │           ├─ Free('__tmp13)
         │     │           │  │           ├─ Let('__tmp14)
         │     │           │  │           │  ├─ ObjectProject(Free('__tmp7){"test"})
         │     │           │  │           │  ╰─ Typecheck(Obj(Map[], Some(Top)))
         │     │           │  │           │     ├─ Free('__tmp14)
         │     │           │  │           │     ├─ Let('__tmp15)
         │     │           │  │           │     │  ├─ ObjectProject(Free('__tmp14){"c"})
         │     │           │  │           │     │  ╰─ Typecheck(Obj(Map[], Some(Top)))
         │     │           │  │           │     │     ├─ Free('__tmp15)
         │     │           │  │           │     │     ├─ Invoke(IfUndefined)
         │     │           │  │           │     │     │  ├─ Invoke(ObjectConcat)
         │     │           │  │           │     │     │  │  ├─ Free('__tmp15)
         │     │           │  │           │     │     │  │  ╰─ Free('__tmp13)
         │     │           │  │           │     │     │  ╰─ ObjectProject(Free('__tmp7){"c"})
         │     │           │  │           │     │     ╰─ Constant(NA)
         │     │           │  │           │     ╰─ Constant(NA)
         │     │           │  │           ╰─ Constant(NA)
         │     │           │  ╰─ Invoke(MakeObject)
         │     │           │     ├─ Str("c")
         │     │           │     ╰─ ObjectProject(Free('__tmp7){"c"})
         │     │           ╰─ Constant(NA)
         │     ╰─ Free('__tmp7)
         ╰─ Constant(Int(11))
```

QScript: https://gist.github.com/6911a89d74a44af852e3959f48a42ad8

Obviously a pretty gnarly query, but the qscript is lololololololol.

djspiewak:
Fixes #2227
Fixes #2775
Fixes #1929

### DIFF
--- a/it/src/main/resources/tests/demo/bug2227.test
+++ b/it/src/main/resources/tests/demo/bug2227.test
@@ -1,0 +1,17 @@
+{
+  "name": "bug2227",
+  "NB": "This test is basically charting on queries with reductions",
+  "backends": {
+    "mimir": "ignoreFieldOrder"
+  },
+  "data": "patients.data",
+  "query": "select city, sum(cnt) as c from (select city, count(*) as cnt from patients where state = \"NY\" group by city order by count(*) desc) as meh group by city order by c desc",
+  "predicate": "atLeast",
+  "ignoreResultOrder": true,
+  "NB": "ignoreResultOrder is needed for atLeast, which sucks",
+  "expected": [
+    { "city": "NEW YORK", "c": 4 },
+    { "city": "ROCHESTER", "c": 2 },
+    { "city": "SOUTHAMPTON", "c": 1 }
+  ]
+}
--- a/it/src/main/resources/tests/demo/bug2227.test
+++ b/it/src/main/resources/tests/demo/bug2227.test
@@ -5,10 +5,8 @@
     "mimir": "ignoreFieldOrder"
   },
   "data": "patients.data",
-  "query": "select city, sum(cnt) as c from (select city, count(*) as cnt from patients where state = \"NY\" group by city order by count(*) desc) as meh group by city order by c desc",
-  "predicate": "atLeast",
-  "ignoreResultOrder": true,
-  "NB": "ignoreResultOrder is needed for atLeast, which sucks",
+  "query": "select city, sum(cnt) as c from (select city, count(*) as cnt from patients where state = \"NY\" group by city order by count(*) desc) as meh group by city order by c desc, city desc",
+  "predicate": "initial",
   "expected": [
     { "city": "NEW YORK", "c": 4 },
     { "city": "ROCHESTER", "c": 2 },
--- a/it/src/main/resources/tests/demo/bug2227.test
+++ b/it/src/main/resources/tests/demo/bug2227.test
@@ -2,6 +2,8 @@
   "name": "bug2227",
   "NB": "This test is basically charting on queries with reductions",
   "backends": {
+    "couchbase": "ignoreFieldOrder",
+    "marklogic_json": "ignoreFieldOrder",
     "mimir": "ignoreFieldOrder"
   },
   "data": "patients.data",
--- a/it/src/main/resources/tests/subqueryGroupBy.test
+++ b/it/src/main/resources/tests/subqueryGroupBy.test
@@ -4,6 +4,7 @@
     "data": "zips.data",
     "query": "SELECT 100 * SUM(pop) / (SELECT SUM(pop) FROM zips), state FROM zips GROUP BY state ORDER BY state LIMIT 5",
     "predicate": "exactly",
+    "NB": "These results are incorrect; they will be rectified when #3184 is fixed",
     "expected": [
         { "0": 100, "state": "AK" },
         { "0": 100, "state": "AL" },
--- a/it/src/main/resources/tests/subqueryGroupBy.test
+++ b/it/src/main/resources/tests/subqueryGroupBy.test
@@ -1,0 +1,14 @@
+{
+    "name": "use subquery when there's a group by statement",
+    "NB": "#1929",
+    "data": "zips.data",
+    "query": "SELECT 100 * SUM(pop) / (SELECT SUM(pop) FROM zips), state FROM zips GROUP BY state ORDER BY state LIMIT 5",
+    "predicate": "exactly",
+    "expected": [
+        { "0": 100, "state": "AK" },
+        { "0": 100, "state": "AL" },
+        { "0": 100, "state": "AR" },
+        { "0": 100, "state": "AZ" },
+        { "0": 100, "state": "CA" }
+    ]
+}
--- a/it/src/main/resources/tests/structureTestDeep.data
+++ b/it/src/main/resources/tests/structureTestDeep.data
@@ -1,3 +1,0 @@
-{ "a": 1, "b": { "ba": { "innermost": true }, "bb": ["x", false, { "foo": "bar" }], "bc": 0.5 }, "c": [] }
-{ "a": 2, "b": { "ba": { "innermost": false }, "bb": ["y", true, { "foo": "baz" }], "bc": 60 }, "c": [true, false, "maybe"] }
-{ "a": 3, "b": { "ba": true, "bb": ["z", { "foo": "quux" }] }, "c": null }
--- a/it/src/main/resources/tests/structureTestDeep.data
+++ b/it/src/main/resources/tests/structureTestDeep.data
@@ -1,0 +1,3 @@
+{ "a": 1, "b": { "ba": { "innermost": true }, "bb": ["x", false, { "foo": "bar" }], "bc": 0.5 }, "c": [] }
+{ "a": 2, "b": { "ba": { "innermost": false }, "bb": ["y", true, { "foo": "baz" }], "bc": 60 }, "c": [true, false, "maybe"] }
+{ "a": 3, "b": { "ba": true, "bb": ["z", { "foo": "quux" }] }, "c": null }
--- a/it/src/main/resources/tests/structureTestDeep.test
+++ b/it/src/main/resources/tests/structureTestDeep.test
@@ -1,0 +1,8 @@
+{
+    "name": "structureTestDeep",
+    "data": "structureTestDeep.data",
+    "query": "SELECT MAP_CONCAT(DELETE_KEY(t2, \"c\"), {\"test\": MAP_CONCAT(t2.test, {\"c\": MAP_CONCAT(t2.test.c, t2.c) ?? t2.c}) ?? {\"c\": t2.c}}) ?? t2 FROM (SELECT MAP_CONCAT(DELETE_KEY(t1, \"b\"), {\"test\": MAP_CONCAT(t1.test, {\"b\": MAP_CONCAT(t1.test.b, t1.b) ?? t1.b}) ?? {\"b\": t1.b}}) ?? t1 FROM (SELECT MAP_CONCAT(DELETE_KEY(t0, \"a\"), {\"test\": MAP_CONCAT(t0.test, {\"a\": MAP_CONCAT(t0.test.a, t0.a) ?? t0.a}) ?? {\"a\": t0.a}}) ?? t0 FROM structureTestDeep AS t0) AS t1) AS t2",
+    "NB": "Right now, this test is just checking compilation. Need confirmation from @garyb on the expected results",
+    "predicate": "exactly",
+    "expected": []
+}
--- a/it/src/main/resources/tests/structureTestDeep.test
+++ b/it/src/main/resources/tests/structureTestDeep.test
@@ -1,8 +1,0 @@
-{
-    "name": "structureTestDeep",
-    "data": "structureTestDeep.data",
-    "query": "SELECT MAP_CONCAT(DELETE_KEY(t2, \"c\"), {\"test\": MAP_CONCAT(t2.test, {\"c\": MAP_CONCAT(t2.test.c, t2.c) ?? t2.c}) ?? {\"c\": t2.c}}) ?? t2 FROM (SELECT MAP_CONCAT(DELETE_KEY(t1, \"b\"), {\"test\": MAP_CONCAT(t1.test, {\"b\": MAP_CONCAT(t1.test.b, t1.b) ?? t1.b}) ?? {\"b\": t1.b}}) ?? t1 FROM (SELECT MAP_CONCAT(DELETE_KEY(t0, \"a\"), {\"test\": MAP_CONCAT(t0.test, {\"a\": MAP_CONCAT(t0.test.a, t0.a) ?? t0.a}) ?? {\"a\": t0.a}}) ?? t0 FROM structureTestDeep AS t0) AS t1) AS t2",
-    "NB": "Right now, this test is just checking compilation. Need confirmation from @garyb on the expected results",
-    "predicate": "exactly",
-    "expected": []
-}


it/src/main/resources/tests/subqueryGroupBy.test
```diff
@@ -0,0 +1,14 @@
...
+    "NB": "#1929",
+    "data": "zips.data",
+    "query": "SELECT 100 * SUM(pop) / (SELECT SUM(pop) FROM zips), state FROM zips GROUP BY state ORDER BY state LIMIT 5",
+    "predicate": "exactly",
+    "expected": [
+        { "0": 100, "state": "AK" },
+        { "0": 100, "state": "AL" },
+        { "0": 100, "state": "AR" },
+        { "0": 100, "state": "AZ" },
+        { "0": 100, "state": "CA" }
```
alissapajer: These are not the correct results for this query. Each row should have the percent of the total population that lives in the given state. So, it looks like we need to keep #1929 open.

it/src/main/resources/tests/demo/bug2227.test
```diff
@@ -0,0 +1,17 @@
...
+  "name": "bug2227",
+  "NB": "This test is basically charting on queries with reductions",
+  "backends": {
+    "mimir": "ignoreFieldOrder"
+  },
+  "data": "patients.data",
+  "query": "select city, sum(cnt) as c from (select city, count(*) as cnt from patients where state = \"NY\" group by city order by count(*) desc) as meh group by city order by c desc",
+  "predicate": "atLeast",
+  "ignoreResultOrder": true,
+  "NB": "ignoreResultOrder is needed for atLeast, which sucks",
```
alissapajer: We have the predicate `initial` that we can use here. Using that we could test that the expected first two rows. And then we won't need to `ignoreResultOrder`.

Since there are multiple rows with `"c": 1`, we could change the query to have `order by c, city` in the outer select to test a few more rows.

alissapajer: `order by c desc, city desc` or `order by c desc, city` is what we need, actually.

djspiewak: No. atLeast appears to just not work at all on this query without it ignoring order. The appropriate results are returned, but the marchers are buggy

alissapajer: I'm not sure I understand what you're saying. If `atLeast` is ignoring order, then I think we need to use a different predicate so that order won't be ignored. This query is requesting that we are ordering by `c`, so we need to make sure we are testing that. 

Additionally, there is no guarantee that `"SOUTHAMPTON"` will be the first `city` value returned, so I was suggesting how we could amend the query to include more expected results.

alissapajer: By adding the `ignoreResultOrder` flag you are forcing it to hit the `OrderIgnored` case of `AtLeast`. I agree that `AtLeast` with `OrderPreserved` is what we need, and we could use that without amending the query, but if that's buggy then I think we should use `Initial` instead, and make a note that `AtLeast` would be better but is buggy.

The query amendment I suggested is only required if we use `Initial` _and_ want to include more than two results in the expected output.

Can you open an issue for the buggy `AtLeast`? We have `PredicateSpec` so we should be able to test it as well.

djspiewak: I'll double check and make sure that I wasn't just seeing things with `atLeast`.  It wasn't an order-ignoring thing.  It actually messed up field comparisons.

it/src/main/resources/tests/subqueryGroupBy.test
djspiewak: I opened #3184 to track this issue, which is distinct from #1929 (and more general).

djspiewak:
@alissapajer Do you want me to add @garyb's revised testcase here?  It does run and produce results, but it's insanely slow even on mimir, and I don't even want to think about how slow it would be on other connectors.

it/src/main/resources/tests/subqueryGroupBy.test
alissapajer: I think we should remove this integration test since the result are incorrect. I don't think we should be tracking incorrect results. 

Unfortunately I don't have a way to get the correct results; my old mongo branch is on a different metastore version and I don't want to mess with that. Otherwise I'd suggest we just update the results to be correct, and make the tests pending.

alissapajer:
@djspiewak hm, since that compile+run time is still too slow for the business, I don't think we need to track it at the integration test level yet. We know it's a problem and will be opening specific issues in order to fix it. So I'm fine closing out #2775 without adding an integration test.

djspiewak:
Sounds good. This is ready for merge then. The build failures are addressed in three other PRs. I’ll open a fresh issue for the query explosion.

alissapajer:
@djspiewak I don't think we should be checking in `subqueryGroupBy.test ` because the results are incorrect. We should check it in when https://github.com/quasar-analytics/quasar/issues/3184 is fixed. Did you have a reason for checking it in sooner?

djspiewak:
Compilation test that the old issue is fixed. I’ll fix the results when I fix the new issue.

alissapajer:
@djspiewak can you add a note in the test itself saying the results are incorrect and will be fixed in #3184. Someone looking at the test needs to know that the results are wrong. Ideally we'd have a integration-level mechanism for testing compilation separate from results.

alissapajer:
@djspiewak also `demo/bug2227.test` is failing for a couple connectors due to field ordering.

PR #3187, opened 2017-11-25
[#3145] Fix how rewrites are applied during ApplyProvenance

### LINKED ISSUE #3145: Eliminating DimEdit in ApplyProvenance doesn't update enough identities
Currently when we eliminate a `DimEdit` in `ApplyProvenance` we update its dimensions, replacing any references to its name with the source's. However due to how we compute provenance in the `zygo`, the _old_ name is propagated to future vertices, it should be the _new_ name.

### LINKED ISSUE #3182: ReifyBuckets erroring
The minimal example
```sql
select sum(ct) from (select count(*) as ct from olympics group by country) as src group by country
```
produces the error
```
Quasar error: ReifyBuckets: Expected to find a mappable region in 'qsu17 based on one of NonEmpty['qsu12].
```
The integration test queries `reduceNestedSelect.test` and `nestedSelectWithFilter.test` are also hitting this error.

wemrysi:
Updates how symbol rewrites are applied in `ApplyProvenance` to be correct in the presence of nested `DimEdit`s. 

Fixes #3145.
Fixes #3182.

### DIFF
--- a/connector/src/test/scala/quasar/qscript/qsu/ApplyProvenanceSpec.scala
+++ b/connector/src/test/scala/quasar/qscript/qsu/ApplyProvenanceSpec.scala
@@ -22,26 +22,35 @@ import quasar.Qspec
 import quasar.Planner.PlannerError
 import quasar.contrib.pathy.AFile
 import quasar.ejson.{EJson, Fixed}
+import quasar.ejson.implicits._
+import quasar.fp._
 import quasar.qscript.construction
 import quasar.qscript.provenance.Dimensions
-import quasar.qscript.HoleF
+import quasar.qscript.{HoleF, ReduceFuncs}
 import quasar.qscript.MapFuncsCore.IntLit
 
+import matryoshka._
 import matryoshka.data.Fix
+import matryoshka.data.free._
 import org.specs2.matcher.{Expectable, Matcher, MatchResult}
 import pathy.Path, Path.{file, Sandboxed}
-import scalaz.{\/, Cofree, IList}
+import scalaz.{\/, Cofree, Equal, IList}
+import scalaz.syntax.applicative._
+import scalaz.syntax.equal._
+import scalaz.syntax.show._
+import scalaz.std.map._
 
 object ApplyProvenanceSpec extends Qspec with QSUTTypes[Fix] {
 
   import ApplyProvenance.AuthenticatedQSU
+  import QScriptUniform.DTrans
 
   type F[A] = PlannerError \/ A
   type QSU[A] = QScriptUniform[A]
 
   val J = Fixed[Fix[EJson]]
 
-  val qsu = QScriptUniform.AnnotatedDsl[Fix, String]
+  val qsu = QScriptUniform.AnnotatedDsl[Fix, Symbol]
   val func = construction.Func[Fix]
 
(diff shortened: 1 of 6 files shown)


djspiewak:
Some of the build failures are legit (tests that were pending are now passing).

alissapajer:
This brought alive `nestedSelectWithFilter.test` :tada: but not `reduceNestedSelect.test`, which was the other integration test that was hitting the #3182 error.

@wemrysi could you take a look at how `reduceNestedSelect.test` is behaving on this branch to see if we need to open an additional issue?

connector/src/test/scala/quasar/qscript/qsu/ApplyProvenanceSpec.scala
```diff
@@ -50,50 +59,81 @@ object ApplyProvenanceSpec extends Qspec with QSUTTypes[Fix] {
...
+                qsu.read('n0, afile),
+                DTrans.Group(func.ProjectKeyS(func.Hole, "x")))),
+              DTrans.Group(func.ProjectKeyS(func.Hole, "y")))),
+            func.ProjectKeyS(func.Hole, "pop"))),
+          ReduceFuncs.Sum(())))
+
+      tree must haveDimensions(SMap(
+        'n4 -> IList(
+          qprov.prov.value(Access.bucket('n4, 1, 'n4).point[FreeMapA])
+        , qprov.prov.value(Access.bucket('n4, 0, 'n4).point[FreeMapA]))
```
alissapajer: Thinking this all through, consider
```sql
select count(sum(pop)) from foobar group by x, y
```
Given our encoding of dimensionality, this query groups the `sum` by `x` followed by `y`, and then groups the `count` by just  `x`. Does that agree with your expectation of this query?

connector/src/test/scala/quasar/qscript/qsu/ApplyProvenanceSpec.scala
```diff
@@ -50,50 +59,81 @@ object ApplyProvenanceSpec extends Qspec with QSUTTypes[Fix] {
...
+      , 'n3 -> IList(
+          qprov.prov.proj(J.str("foobar"))
+        , qprov.prov.value(func.ProjectKeyS(Access.value('n0).point[FreeMapA], "y"))
+        , qprov.prov.value(func.ProjectKeyS(Access.value('n0).point[FreeMapA], "x")))
+      , 'n0 -> IList(
```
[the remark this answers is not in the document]
wemrysi: The `DimEdit`s are applied to the provenance of the `Read` node (and removed in the process), which is why this node includes the group keys.

connector/src/test/scala/quasar/qscript/qsu/ApplyProvenanceSpec.scala
wemrysi: Yes, that seems correct as each reduction "consumes" a dimension.

wemrysi:
@alissapajer `reduceNestedSelect.test` no longer throws an exception due to `ReifyBuckets` but it is still failing due to another issue uncovered (#3191) now that this issue is fixed.

wemrysi:
Grr, apparently I forgot about `mongodb_read_only`, rebuilding...

PR #3195, opened 2017-11-28
Switch CoreEffIORW from CoM to expanded Coproducts

drostron:
Isolated compilation of `RestApi.coreServices[CoreEffIORW]` went from ~130s down
to ~40s on a 2015 macbook pro

### DIFF
--- a/interface/src/main/scala/quasar/main/package.scala
+++ b/interface/src/main/scala/quasar/main/package.scala
@@ -208,14 +208,19 @@ package object main extends Logging {
   }
 
   /** Effect comprising the core Quasar apis. */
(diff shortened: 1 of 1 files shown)


drostron:
Towards speeding up downstream compilation leading to build timeouts slamdata/slamdata-backend#280

jedesah:
We can never have nice things... 😞

jedesah:
Just missing a version bump

jedesah:
I'm going to merge this as is because I want to compilation speedup and that way it will be available to SB sooner but I think a comment would be a good idea so it doesn't revert back to nicer/slower syntax

drostron:
@jedesah Comment added

PR #3198, opened 2017-11-28
Over-aggressive reduction coalescence

### LINKED ISSUE #3184: MinimizeAutoJoins should check branch provenance first
Right now, we over-aggressively try to merge even in cases where it's actually impossible.  This in turn results in some rather arbitrary choices being made.  An example is this query:

```sql
```

### LINKED ISSUE #3186: ReifyBuckets can produce duplicates
When the same `Value` provenance appears multiple times in the dimensions of the source to a reduce vertex, it results in duplicate buckets in the reduce. The duplicates should be elided.

djspiewak:
Fixes #3184 
Fixes #3186 

Things still aren't perfect.  I'm about to file an issue detailing other problems which arise even after this fix.  The integration test had to be marked as skipped because of rounding differences between the connectors, and unfortunately we cannot just add `round(...)` because of the issue I'm about to file.


wemrysi:
@djspiewak Looks like `mongodb_read_only` also needs to be `pending` in `unsupportedTests.test`.

wemrysi:
@djspiewak I think `unsupportedTests.test` dislikes you. 👻

it/src/main/resources/tests/unsupportedTests.test
```diff
@@ -1,26 +1,25 @@
...
+        "couchbase": "skip",
+        "marklogic_json": "pending",
+        "mimir": "pending",
+        "mongodb_2_6": "pending",
+        "mongodb_3_0": "pending",
+        "mongodb_3_2": "pending",
+        "mongodb_3_4": "pending",
+        "mongodb_read_only": "pending"
     },
+    "NB": "pending on mimir due to sparse arrays",
```
alissapajer: Is this comment valid anymore?

djspiewak: Probably not.  I haven't had a chance to dig into exactly what is happening now that the array stuff is gone.

alissapajer: No harm in leaving it for now. There could even be another place we're using arrays in post-qsu rewrites.

PR #3204, opened 2017-11-29
Ensure new reduce name is used when remapping dimensions in MinimizeAutoJoins

wemrysi:
Previously, we were remapping the dimensions using the name of the `Map` node around the new `Reduce`, which resulted in incorrect bucket access.


djspiewak:
PlannerSpec

wemrysi:
Looks like this is blocked on #3198 for the provenance normalization.

djspiewak:
Needs merge from master.

wemrysi:
@djspiewak `master` merged, also improved provenance normalization a bit.

PR #3212, opened 2017-11-29
Consider previously-failed autojoins as unmappable sources

### LINKED ISSUE #3199: Failed autojoin elimination results in a downstream "join explosion"
This query (from #3184) is relatively straightforward:

```sql
```

And it generates reasonable qscript, modulo issues with join association.  However, simply adding a mapfunc around the left term of the map makes everything go insane:

```sql
SELECT round(100 * SUM(pop) / (SELECT SUM(pop) FROM zips)), state FROM zips GROUP BY state
```

The resulting qscript is [here](https://gist.github.com/aef42c19aaa96c54f64afa9209210260) and involves features such as a) a zillion `ThetaJoin`s, and b) a `MakeMap` which has `LeftSide` for the key inside a joinfunc, resulting in a dynamic map creation out of thin air.  No idea what's causing this.

This is all on top of #3198

djspiewak:
Depends on #3198 
Fixes #3199 

Previously, on `MinimizeAutoJoins`…

We would fail to eliminate an `AutoJoin`, leaving it in the graph and moving on to the downstream vertices.  Any subsequent `AutoJoin`s which were freemappable from the *first* `AutoJoin` would then end up considering the original roots as part of their `candidates`.  But since those roots had already failed to merge, they certainly wouldn't merge on the second try, meaning that *all* subsequent `AutoJoin`s within a freemappable region would fail to eliminate, resulting in a cascading "join explosion" (particularly with `Guard`, which results in multiple `ThetaJoin`s at graduation).

This PR solves the issue by considering prior failures as maximal roots, despite the fact that we *could* freemap above them.  Since we handle the rewrites top-down, there shouldn't be any missed opportunities between the failed `AutoJoin` and its sources, so I believe this is probably optimal in lieu of join reassociation (which is another issue).


djspiewak:
@edmundnoble Build green; ready for merge.

edmundnoble:
Makes sense to me.

PR #3214, opened 2017-11-30
Improve error message when failed to switch metastores

jedesah:
Fixes https://github.com/slamdata/slamdata-backend/issues/254.

I don't know if this warrants a major version of bump. It is technically a breaking api change but a very minor one...


djspiewak:
@jedesah It's just returning a different error message right?  That doesn't feel like a breaking change to me.

jedesah:
@djspiewak Yup, that's all it's doing. Ya, our definition of breaking change seems to be fuzzy. It will break any client depending on the current format of the error message but agreed it's very minor

wemrysi:
We actually mention in the [wiki](https://github.com/quasar-analytics/quasar/wiki/Versioning) that changes to error message strings isn't breaking.

djspiewak:
IMO should be `version: revision` then.

jedesah:
Fine by me

web/src/main/scala/quasar/api/services/metastore.scala
```diff
@@ -44,7 +45,8 @@ object metastore {
           _              <- EitherT(meta.set(connConfig, initialize))
           newUrl         =  DbConnectionConfig.connectionInfo(connConfig).url
           initializedStr =  if (initialize) "newly initialized " else ""
-        } yield s"Now using ${initializedStr}metastore located at $newUrl").leftMap(msg => ApiError.fromMsg(BadRequest, msg)))
+        } yield s"Now using ${initializedStr}metastore located at $newUrl")
+          .leftMapF(msg => lift(BadRequest(Json("error" := Json("status" := "UninitializedMetastore", "detail":= Json("message" := msg))))).into[S]))
```
wemrysi: @jedesah `ApiError.fromMsg_` will get you want you want here.

jedesah: @wemrysi I don't think so. `fromMsg_` requires a `http4s.Status` as argument but @kRITZCREEK specifically requested the `status` field be populated with `"UninitializedMetastore"`.

There is a ticket to revisit the format of all error messages which hopefully I can get around to after `5.0`. This change bugs me a little because it's fixing one error message at the expense of uniformity with the rest but until we do the complete overhaul I think @kRITZCREEK was particularly interested in fixing this one.

kRITZCREEK: If we're getting the proper tag on the error messages that I can match on "soonish" I'll be fine. If this is too much of a hassle right now, do something else I can clearly identify.

jedesah: I don't think it's too much of a hassle. I was explaining the reason for bypassing the internal API that is meant to provide uniformity across error messages. I think this is an improvement and hopefully, we can get around to fixing everything else soon.

wemrysi: @jedesah ‘fromMsg_(BadRequest withReason “UninitializedMetastore”, msg)’

jedesah: @wemrysi Ah, thx that makes sense!

jedesah:
@wemrysi Made the change you suggested

Issue #3211, opened 2017-11-29
Demo integration tests


rabbitonweb:
Move work from #3166 to integration tests

rabbitonweb:
Not done yet

djspiewak:
@rabbitonweb It's because you used the phrase "partially resolves" in the PR description, so github auto-closed it.

rabbitonweb:
@djspiewak yeah, i know, i just put comment so we know why i'm opening it

PR #3215, opened 2017-11-30
Add it/tests based on queries generated from demo workspaces (part 1)

rabbitonweb:
Partially resolved https://github.com/quasar-analytics/quasar/issues/3211


rabbitonweb:
Running on CI, tests that will fail need to be marked pending. I will see which ones once CI is done

djspiewak:
We might be able to use the existing patients and olympics datasets.  Or at the least, not have patients multiple times.  :-)  Otherwise looks good!

rabbitonweb:
@alissapajer @djspiewak this is ready for review

PR #3230, opened 2017-12-04
Only use mapReduce to plan join branches if necessary

### LINKED ISSUE #3152: Random _id returned from query
This is on the qsu branch.  See the gist: https://gist.github.com/72074d7cbc6563d6913696236ca84317

jsantos17:
So looks like we were planning joins with mapReduce even when it wasn't necessary. Fixes #3152.


jsantos17:
Explanation of the bug fixed in this PR:

In order to join two collection, Quasar uses the `reduce` output option for a `mapReduce`. For this to work the `_id`s for each element we join together must collide. The collision will make Mongo call the reduce function while writing back. So just before we write back to the temp collection the result of each branch, `JoinHandler` runs a `GroupF` grouping by null. Then the `_id`s of each document is `null` so they collide and Mongo doesn’t generate a new `_id`.

However, the `GroupF` by null approach only works for join branches planned with aggregation. Join branches planned with mapReduce use a `MapF` instead. Due to a silly bug every join branch was using the mapReduce approach if join keys were empty. Because `left0.keys.map(HasThis.unapply).sequence` will return `Some` for empty key lists. Even if nothing actually needs JavaScript in the join branches. This led to broken write-back results and Mongo generating new ids that were then displayed to the user.

This is the code that picks whether to use `MapF` or `GroupF` to shape the result we want to write back.

https://github.com/quasar-analytics/quasar/pull/3230/files#diff-2a598731f642bc5c7bc5913d05ee2154R282

Issue #3075, opened 2017-11-13
Triage `distinct` unit tests


jsantos17:
Branched off #2667 

There are 8 disabled tests for queries including `distinct`. Some of them just need updated expectations while others might have QScript issues or issues in the Mongo planner. Queries with QScript issues can be verified by writing tests in `PlannerQScriptSpec` to bypass QScript generation.

jsantos17:
Just checked the state of distinct and most queries are correct and working correctly. The only two problems queries are:

| Query | Result|Plan|
|-------|------------|-----|
|`SELECT DISTINCT SUM(pop) AS totalPop, city, state FROM zips GROUP BY city`|Empty|[Link](https://gist.github.com/jsantos17/2e82cb7c18d1c7e703e8f03dea32804b)
|`SELECT DISTINCT SUM(pop) AS totalPop, city, state FROM zips GROUP BY city ORDER BY totalPop DESC`|Empty|[Link](https://gist.github.com/jsantos17/01013e63c9b68b919d8c68775f87a603)

These have autojoins in them that can probably be optimized away, but we should be getting correct results anyway.

@alissapajer, @djspiewak, @wemrysi: Any known issues in the two queries above? If not, this might be a bug in the Mongo connector.

These have `Guard`s on `ShiftedRead` so we might want to re-evaluate these two queries after #3048

djspiewak:
These are both the same issue.  I believe @wemrysi is working on this right now.  Neither are failing because of the `Guard`s (they fail on mimir as well).

alissapajer:
We are projecting out keys that we have not created. For example [here](https://gist.github.com/jsantos17/2e82cb7c18d1c7e703e8f03dea32804b#file-distinct-with-sum-and-group-txt-L72) we project out `"identities"` followed by `"qsu18_0"`, neither of which exist.

alissapajer:
Also to note, [here](https://gist.github.com/jsantos17/2e82cb7c18d1c7e703e8f03dea32804b#file-distinct-with-sum-and-group-txt-L76) we project `"city"` from the right side join branch, but in the right side join branch we create a map with key `"state"`.

wemrysi:
@alissapajer Yeah, the second bit you refer to is what I'm fixing for #3170 and the first bit should be fixed when #3204 is merged.

jsantos17:
Awesome. I'll re-evaluate once #3204 is merged. Thanks for all the input.

djspiewak:
[...]

Issue #3225, opened 2017-12-04
Remove special handling for mongo < v3.2


rintcius:
We still handle version 2.6+ for mongo. Since we don't support versions less than 3.2 anymore we can remove special handling for these unsupported versions.

PR #3236, opened 2017-12-05
Update expectations on fixed Mongo connector tests

jsantos17:
Relates to #3075. Not closing that just yet since I'm using it to keep track of the last two remaining, failing distinct queries.


mongodb/src/test/scala/quasar/physical/mongodb/PlannerSql2ExactSpec.scala
```diff
@@ -1821,21 +1807,21 @@ class PlannerSql2ExactSpec extends
       plan2_6(sqlE"select smallZips.city from zips join smallZips on zips.`_id` = smallZips.`_id`") must
```
jsantos17: This plan is for 2.6 which we no longer support, but I didn't delete it because it shows how close we're to planning joins correctly. This specific test needed very minor fixes to match the old Mongo spec.

rintcius: Yeah, we can address this in #3225

PR #3244, opened 2017-12-06
Define and/or to have "outer" semantics

### LINKED ISSUE #2986: Implement “outer” semantics for mimir or/and
Title

djspiewak:
Fixes #2986

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/InfixLib.scala
+++ b/mimir/src/main/scala/quasar/mimir/InfixLib.scala
@@ -236,8 +236,76 @@ trait InfixLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
         }
       }
 
-      val And = new BoolOp2("and", _ && _)
-      val Or  = new BoolOp2("or", _ || _)
+      // TODO find the commonalities and abstract these two
+      val And = new OpNFN(InfixNamespace, "and") {
+        val fn: FN = CFNP("builtin::infix::bool::and") {
+          case List(c: BoolColumn) =>
+            new BoolColumn {
+              def isDefinedAt(row: Int): Boolean = c.isDefinedAt(row) && !c(row)
+              def apply(row: Int): Boolean = c(row)
+            }
+
+          case List(c1: BoolColumn, c2: BoolColumn) =>
+            new BoolColumn {
+              def isDefinedAt(row: Int): Boolean = {
+                if (c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  true
+                else if (c1.isDefinedAt(row) && !c2.isDefinedAt(row))
+                  !c1(row)
+                else if (!c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  !c2(row)
+                else
+                  false
+              }
+
+              def apply(row: Int): Boolean = {
+                if (c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  c1(row) && c2(row)
+                else if (c1.isDefinedAt(row) && !c2.isDefinedAt(row))
+                  c1(row)
+                else if (!c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  c2(row)
+                else
+                  false
+              }
+            }
+        }
+      }
+
+      val Or = new OpNFN(InfixNamespace, "or") {
+        val fn: FN = CFNP("builtin::infix::bool::or") {
+          case List(c: BoolColumn) =>
+            new BoolColumn {
+              def isDefinedAt(row: Int): Boolean = c.isDefinedAt(row) && c(row)
+              def apply(row: Int): Boolean = c(row)
+            }
+
+          case List(c1: BoolColumn, c2: BoolColumn) =>
+            new BoolColumn {
+              def isDefinedAt(row: Int): Boolean = {
+                if (c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  true
+                else if (c1.isDefinedAt(row) && !c2.isDefinedAt(row))
+                  c1(row)
+                else if (!c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  c2(row)
+                else
+                  false
+              }
+
+              def apply(row: Int): Boolean = {
+                if (c1.isDefinedAt(row) && c2.isDefinedAt(row))
+                  c1(row) || c2(row)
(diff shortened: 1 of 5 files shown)


rabbitonweb:
This is more general comment (not specific to this PR), when I create PRs I always point to initial issue (which I also try to document if it was not documented by its creator). In the PR itself I at least try to give general feedback of what I tried to achieve. Question like: what is being fixed or extended, maybe what were the challenges. Doesn't have to be 700 long epos :) - just a general overview so that the reviewer has *context* of what is actually going on.

Here (in this PR) what I have is:
1. #2986 title "Implement “outer” semantics for mimir" which does not tell me much
2. #2986 description `"Title"` - that's literally all :)
3. 6 commits - small set so I will eventually get what's the idea behind that PR but small context would be beneficial 

^^ this is general remark, will probably post something on #backend about it

djspiewak:
@rabbitonweb Sorry, there was some assumed knowledge here.  The terminology refers to "inner" vs "outer" joins.  General explanation…

In MRA, any n-ary operator where *n > 1* has a choice as to how it wants to handle undefinedness in its values.  This is most easily seen in binary operators, where you could choose the result to be undefined if either input is undefined, or you could define some sort of recovery semantic.  The `+` operator, for example, is undefined if either of its inputs are undefined.  Conversely, the `,` operator (`ConcatMap`) is undefined only if *both* of its inputs are undefined; if only one is undefined, then that part of the resulting map will be retained.  This sort of recovery behavior is what we refer to when we say "outer" semantics.

By MRA's specification, *most* of our operators have "inner" semantics: they are undefined whenever either input is undefined.  This also allows more aggressive optimizations in the low-level connector implementation.  However, just as a practical matter, some operators do need to have outer semantics (either partially, as in the case of `??`, or totally as in the case of `,` and `or`/`and`).  This PR implements those semantics for `or` and `and`.

As a sidebar, I did update the tests.  When I review PRs, that's usually the first place I start looking, since it tells me the specification of what the main changes were trying to do.  Definitely worth having a conversation in #backend-team though if there's ever something about a PR or issue that doesn't make sense!

mimir/src/main/scala/quasar/mimir/InfixLib.scala
```diff
@@ -236,8 +236,76 @@ trait InfixLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
...
+          case List(c: BoolColumn) =>
+            new BoolColumn {
+              def isDefinedAt(row: Int): Boolean = c.isDefinedAt(row) && c(row)
+              def apply(row: Int): Boolean = c(row)
+            }
+
+          case List(c1: BoolColumn, c2: BoolColumn) =>
+            new BoolColumn {
+              def isDefinedAt(row: Int): Boolean = {
+                if (c1.isDefinedAt(row) && c2.isDefinedAt(row))
```
rabbitonweb: I just wonder if maybe this is more readable 
```
(c1.isDefinedAt(row), c2.isDefinedAt(row) match {
  case (true, true) => true
  case(true, false) => c1(row)
  case(false, true) => c2(row)
  case(false, false) => false
}
```

djspiewak: It could be. It's also at least two orders of magnitude slower.

rabbitonweb: @djspiewak can you point where is the pain https://gist.github.com/rabbitonweb/bbd8604b7edb2881a7716ccdeecbc42e ?

djspiewak:
```
       7: invokestatic  #20                 // Method scala/runtime/BoxesRunTime.boxToInteger:(I)Ljava/lang/Integer;
```

OH GOD NO BOXING MAKE IT STAHP

```
      10: invokeinterface #26,  2           // InterfaceMethod scala/PartialFunction.isDefinedAt:(Ljava/lang/Object;)Z
```

Say what now?  What's even partial here?  Oh yeah, pattern matching…

```
      18: invokestatic  #20                 // Method scala/runtime/BoxesRunTime.boxToInteger:(I)Ljava/lang/Integer;
```

OH NO IT HAPPENS MORE

```
      26: invokespecial #30                 // Method scala/Tuple2$mcZZ$sp."<init>":(ZZ)V
```

Not only do we box in `Integer`, but we take that box and put it into ANOTHER box!

```
      38: invokevirtual #36                 // Method scala/Tuple2._1$mcZ$sp:()Z
```

…only to pull it back out again.

Actually this is hilarious too.  If you look very carefully at the stack state (hinted by the type signature in instruction 38), you'll see that the boxing is *solely* to do the `isDefinedAt` check!  `Tuple2` is specialized.

```
      51: iload         5
      53: if_icmpne     67
      56: iconst_1
      57: iload         6
      59: if_icmpne     67
      62: iconst_1
      63: istore_3
      64: goto          223
      67: goto          73
      70: goto          73
      73: aload         4
      75: ifnull        119
```

Well…  That seems efficient.

```
     101: if_icmpne     116
     104: aload_1
     105: bipush        10
     107: invokeinterface #43,  2           // InterfaceMethod scala/PartialFunction.apply$mcZI$sp:(I)Z
```

Wait, how much undefinedness could we possibly have???

Also dear god, it's an `invokeinterface`…

```
     156: invokeinterface #43,  2           // InterfaceMethod scala/PartialFunction.apply$mcZI$sp:(I)Z
```

Oh christ, we're actually calling something in `PartialFunction`????  Good job it's specialized I suppose.

```
     213: new           #45                 // class scala/MatchError
     216: dup
     217: aload         4
     219: invokespecial #48                 // Method scala/MatchError."<init>":(Ljava/lang/Object;)V
```

Very helpful, scalac.  This is a total match.

In fairness, most of the boxing that is happening here is because you're using `PartialFunction`, which is vastly less efficient than the `Column` datatype that is inside of mimir.  But the point is that all of this is happening on a *per row* basis.  We need to count milliseconds here.  *Nanoseconds*, really.  The code I wrote is entirely unboxed and just passes primitives around.  The only fault you can find with it is the `invokevirtual` instructions, which is deeply unfortunate and will be removed when we rewrite the internals of mimir into a monomorphic, pre-compiled-to-bytecode eagerly fused evaluation model.

djspiewak:
@rabbitonweb I believe this is ready for merge?  Unless you had other objections?

PR #3250, opened 2017-12-07
Demo integration tests part3

rabbitonweb:
Fixes and finally 

closes https://github.com/quasar-analytics/quasar/issues/3211


alissapajer:
One more test needs to be pending for couchbase.

it/src/main/resources/tests/demo/flights_step1.test
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "flights_step1",
+    "data": "flights.data",
+    "backends": {
+       "couchbase":         "ignoreFieldOrder",
+       "marklogic_json":    "ignoreFieldOrder",
+       "mimir":             "ignoreFieldOrder"
+    },
+    "query": "SELECT source AS Source, destination AS Destination, fuel AS Cost FROM flights ORDER BY Source ASC, Destination ASC, Cost ASC LIMIT 6",
```
alissapajer: I'd like to verify that we are also sorting by `Cost`. The are no entries with the same `source` and `destination`, so there is no data to sort by `Cost`. Could you make a second dataset and integration test to test this?

it/src/main/resources/tests/demo/flights_step2.test
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "flights_step2",
+    "data": "flights.data",
+    "backends": {
+       "couchbase":         "ignoreFieldOrder",
+       "marklogic_json":    "ignoreFieldOrder",
+       "mimir":             "ignoreFieldOrder"
+    },
+    "query": "SELECT source, destination, SUM(fuel) AS weight FROM flights GROUP BY source, destination ORDER BY source ASC, destination ASC, weight asc LIMIT 6",
```
alissapajer: Again since there are no entries with the same `source` and `destination` each group in the `group by` has a single entry. This is good to test, but I'd also like to see this query run against data that has multiple entries per group bucket.

alissapajer:
Needs version label.

it/src/main/resources/tests/demo/flights_step1.test
djspiewak: Aren't these datasets drawn directly from the demo decks?  I agree that we should probably add an integration test which checks what you're talking about, that seems more like a separate PR, no?

alissapajer: I don't know where these datasets are coming from (if they are the full demo datasets or just subsets of them). I was thinking they might be subsets of the real data, otherwise these queries aren't doing as much as they could be doing.

If this is indeed the full dataset, I'm fine opening a new issue for these other queries I have suggested. I was just thinking it would be easier to add them as tests to this PR rather than file a separate issue. But I'm fine either way.

rabbitonweb:
I will mark that last test as pending
Version label added

rabbitonweb:
@alissapajer @djspiewak regarding ordering: those queries are (almost) exact as the ones generated by workspaces. I sometimes modify the `ORDER BY` calls to make results deterministic in order set predicate to `exactly`

rabbitonweb:
Also I would say `ORDER BY` is right now the most covered part o SQL^2 within integration tests :)

alissapajer:
@rabbitonweb what we were wondering is if the demo data is exactly the data used by the demo environments, or if that's been changed in any way?

rabbitonweb:
@alissapajer data is *exactly* the same, queries are modifed by changing `order by` parts a bit

it/src/main/resources/tests/demo/flights_step1.test
alissapajer: See #3259

it/src/main/resources/tests/demo/flights_step2.test
alissapajer: See #3259

djspiewak:
`demo/flights_step3.test` should be skipped on couchbase, since it doesn't have general join.  After that's fixed, :+1: for merge.

Issue #2329, opened 2017-06-14
couchbase integration test intermittently fails loading test data


alissapajer:
Seen on travis.
```
Running Streaming Queries [......................................................................................................................................................................]
[error] ! distinct * [./distinctStar.test] (11 seconds, 991 ms)
[error]  java.lang.RuntimeException: Failed loading test data './cities.data': null (QueryRegressionTest.scala:265)
[error] quasar.regression.QueryRegressionTest$$anonfun$loadTestData$1.applyOrElse(QueryRegressionTest.scala:265)
[error] quasar.regression.QueryRegressionTest$$anonfun$loadTestData$1.applyOrElse(QueryRegressionTest.scala:262)
[error] com.couchbase.client.java.CouchbaseAsyncBucket$13.call(CouchbaseAsyncBucket.java:475)
[error] com.couchbase.client.java.CouchbaseAsyncBucket$13.call(CouchbaseAsyncBucket.java:456)
[error] rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:69)
[error] rx.observers.Subscribers$5.onNext(Subscribers.java:235)
[error] rx.internal.producers.SingleProducer.request(SingleProducer.java:65)
[error] rx.Subscriber.setProducer(Subscriber.java:211)
[error] rx.internal.operators.OnSubscribeMap$MapSubscriber.setProducer(OnSubscribeMap.java:102)
[error] rx.Subscriber.setProducer(Subscriber.java:205)
[error] rx.subjects.AsyncSubject.onCompleted(AsyncSubject.java:105)
[error] com.couchbase.client.core.endpoint.AbstractGenericHandler.completeResponse(AbstractGenericHandler.java:364)
[error] com.couchbase.client.core.endpoint.AbstractGenericHandler.access$000(AbstractGenericHandler.java:69)
[error] com.couchbase.client.core.endpoint.AbstractGenericHandler$1.call(AbstractGenericHandler.java:382)
[error] rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
[error] rx.exceptions.OnErrorThrowable.addValueAsLastCause(OnErrorThrowable.java:109)
[error] rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:73)
[error] rx.observers.Subscribers$5.onNext(Subscribers.java:235)
[error] rx.internal.producers.SingleProducer.request(SingleProducer.java:65)
[...]

PR #3263, opened 2017-12-08
Reimplemented Filter => Cond rewrite to apply more conservatively

djspiewak:
We can now eliminate autojoins even in cases where common sources are being obscured by filters.  We do this by rewriting the `Filter` to a `Cond` within the autojoin heritage (we don't eliminate the node, obviously, so other descendants will continue to see it as a `Filter`).  This optimization applies recursively and can open the door to other collapsing.

I wish we had a `Scan` `MapFunc`; if we had that, I could do something similar with `Subset`.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
@@ -97,7 +97,7 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends 
 
       fms = branches.zipWithIndex map {
         case (g, i) =>
-          expandSecondOrder(MappableRegion[T](state.failed, g)).map(g => (g, i))
+          MappableRegion[T](state.failed, g).map(g => (g, i))
       }
 
       (remap, candidates) = minimizeSources(fms.flatMap(_.toList))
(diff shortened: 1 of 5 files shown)


alissapajer:
I'll review this this afternoon.

alissapajer:
@djspiewak there is another flaky couchbase test. I am all for skipping these when they flake. But I think we might ending skipping half the couchbase test suite because many different couchbase tests sometimes fail with "Failed loading test data". So at some point to maintain a verified couchbase connector we will just need to fix this. (See #2329.)

djspiewak:
@alissapajer Yeah I agree.  Not a windmill we can tilt right now, but if it goes too much further in this direction, we'll probably just mass-disable couchbase until polyrepo.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -110,23 +110,26 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
+        val fm2 = fm flatMap { i2 =>
+          if (i2 === i)
+            func.Cond(predicate.map(κ(i)), i.point[FreeMapA], func.Undefined[Int])
+          else
+            i2.point[FreeMapA]
+        }
 
-    /*fm flatMap {
-      case QSFilter(source, predicate) =>
-        // tune into 102.5 FM, The Source
```
alissapajer: oh noes we lost this radio joke :(

connector/src/test/scala/quasar/qscript/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -280,7 +280,23 @@ object MinimizeAutoJoinsSpec extends Qspec with TreeMatchers with QSUTTypes[Fix]
                 HoleF[Fix],
                 func.Undefined)))
       }
-    }.pendingUntilFixed
```
alissapajer: 🎉

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
alissapajer: Every university has a radio station or two, right?

alissapajer:
@djspiewak looks like we need to do #3276 in order to merge this (or probably any other PRs at this point).

djspiewak:
Lets ignore couchbase failures (ie not hold up PRs) until it’s done then. I’ll PR it tomorrow.

Issue #3272, opened 2017-12-08
Bad assumptions about wrapped buckets


djspiewak:
`subqueryGroupBy.test` and possibly also `unsupportedTests.test`.  This appears to be an issue on both mimir and mongo, but maybe different causes.

alissapajer:
We could also just add `Distinct` to qscript, either as an optional coproduct node, or as a required one (meaning we have to update each connector, but that might not be that hard).

djspiewak:
I'm legit terrified about the scala compilation time implications of that though.

alissapajer:
That's all too fair :(

Moving forward we can't have compile times preventing us from doing the right thing, but that's not something we're about to tackle pre-5.0. It might be worth adding the node unimplemented everywhere, and benchmarking the compile times, if someone has a computer with some CPU cycles to spare...

djspiewak:
Yeah that might be worth doing.  It's a particularly interesting case in the above, since neither of the queries in question involve `distinct`!  It's just that the pattern is the same as what gets generated from a `Distinct` node.

alissapajer:
Oh, that's actually really interesting. I wonder if there is a more general pattern we want to detect in qsu.

djspiewak:
I was thinking about it. I'm not yet convinced that interpreting it as `distinct` is universally the right thing to do though.  And in this particular case, it only exists because we don't have join reassociation.  But maybe…

I'd venture a guess that most backends are going to have a primitive `distinct` that is more efficient than reduction.

djspiewak:
I figured this out.  At least for mimir, it has nothing to do with `distinct` recognition.  Rather, it's the fact that we *always* used to get reduce buckets wrapped in an array.  Now, since we get them separated, we're hitting some code paths that were never checked before, and are in fact buggy.  I would venture to guess that mongo is having the same problem.  (@jsantos17 / @rintcius ?)

rintcius:
https://gist.github.com/rintcius/160625ed8181269ba39fd87449eee476 shows the same problem as `subqueryGroupBy.test` but is smaller. The problem is that the last operation in the pipeline is map-reduce. When this is the case the sorting is not done. We have a rewrite in place - `mapBeforeSort` - that avoids this in some cases, but not all. 

`unsupportedTests.test` has another issue, it doesn't return any results. Haven't looked at it in detail, but here's the plan: https://gist.github.com/rintcius/8e33fab7ac11ea9bcf42aa22b295d619

[...]

PR #3280, opened 2017-12-09
Fixed mistaken assumption about group keys always being array-wrapped

djspiewak:
Affects #3272

As mentioned on the issue, I'm guessing mongo has a similar problem.

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/Mimir.scala
+++ b/mimir/src/main/scala/quasar/mimir/Mimir.scala
@@ -284,7 +284,7 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
       if (specs.isEmpty)
         TransSpec1.Id
       else if (specs.lengthCompare(1) == 0)
-        specs.head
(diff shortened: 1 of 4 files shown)


djspiewak:
@drostron The only thing failing here is couchbase, which is disabled in master, so it's ready for review.

mimir/src/main/scala/quasar/mimir/Mimir.scala
```diff
@@ -284,7 +284,7 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
       if (specs.isEmpty)
         TransSpec1.Id
       else if (specs.lengthCompare(1) == 0)
-        specs.head
+        WrapArray(specs.head)
```
drostron: I take it there's no opportunity for mistakenly double wrapping then?

drostron: Question quelled via a quick conversation.

Issue #3176, opened 2017-11-24
Handle generated `_id`s when reading results from a collection Quasar wrote back


jsantos17:
The [basic analytic query test](https://github.com/quasar-analytics/quasar/blob/8c30259dffa4eaf56acca3d78783374907129793/it/src/main/resources/tests/resultFile/basicAnalyticQuery.test) is failing because of an extraneous `_id` in the result. This is unrelated to #3152.

The query is planned correctly and the test passes when run in streaming mode. The planner generates the correct plan and the correct results are returned (without `_id`). However, when Quasar writes back the results into a temporary collection, MongoDB generates `_id`s for each document. This, of course, makes the result no longer match the expectation.

This was working before QSU was merged into `master`. Investigating how this was working before might prove fruitful. Old Mongo worked around this problem by special handling of the `_id` field. The Old Mongo approach is fragile and sometimes deletes real `_id`s (when implemented in New Mongo).

cc @wemrysi

zeeiyerWork:
One approach could be to write a "special-field" when quasar writes - so that we can ignore the _id when read back (so we know for sure which is customer data versus synthetic temporary data?) ? @jsantos17 or @wemrysi - any thoughts on this approach?

zeeiyerWork:
@jsantos17 opines that this does not impact end-user queries (just tests that explicitly write back) - hence the deprioritization.

jsantos17:
Yes, this shouldn't affect any queries unless the user explicitly asks for the result of a query to be written into a collection. Intermediate result write back is unaffected.

So the question is, does the frontend ever make queries intended to be stored in a Mongo collection? @garyb, @beckyconning?

PR #3282, opened 2017-12-11
Add a couple integration tests

### LINKED ISSUE #2530: QScript incorrect when we join on distinct
```sql
select * from patients as p join (select distinct state from zips) as z on p.state = z.state
```

In the `QScript` we `ProjectField` of `"state"` in the result of the reducer, and then we `ProjectField` of `"state"` on the result of _that_ in the join condition. We only need one of these projections.

Additionally, on new mongo, the bucket of the reduction is equal to the reduction, yet we are supposed to detect this case and reference the bucket in the reduction instead of duplicating it. 

The `QScript` for `distinct` has changed dramatically on the new mongo branch (compared to master), so unfortunately we probably need to fix this on the new mongo branch.

### LINKED ISSUE #2720: Cannot use `?? {}` in `OBJECT_CONCAT` or `DELETE_FIELD` arguments
Example query that is rejected:

``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "b"), t0 ?? {}) FROM `/spark-local/structure_test_deep.json` AS t0
```

Error message from quasar:
``` json
{
  "error": {
    "status": "Error in query.",
    "detail": {
      "message": "Unknown arguments: List(Obj(Map(),Some(Top)), Coproduct(Obj(Map(),Some(Top)),Const(Obj(Map()))))"
    }
  }
}
```

However, using either `t0` or `{}` alone works fine:
``` sql
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "b"), {}) FROM `/spark-local/structure_test_deep.json` AS t0
SELECT OBJECT_CONCAT(DELETE_FIELD(t0, "b"), t0) FROM `/spark-local/structure_test_deep.json` AS t0
```

The issue seems to be specifically with using constant values with `??`, as using a map on the RHS with some value computed from t0 is fine, but other constant map values are not.

alissapajer:
Closes #2530
Closes #2720

### DIFF
--- a/it/src/main/resources/tests/ifundefinedInFunction.test
+++ b/it/src/main/resources/tests/ifundefinedInFunction.test
@@ -11,8 +11,8 @@
 
     "query": "SELECT MAP_CONCAT(DELETE_KEY(t0, \"b\"), t0 ?? {}) FROM flattenable AS t0",
 
+    "ignoredFields": ["_id"],
     "predicate": "exactly",
-
     "ignoreResultOrder": true,
 
     "expected": [
--- a/it/src/main/resources/tests/ifundefinedInFunction.test
+++ b/it/src/main/resources/tests/ifundefinedInFunction.test
@@ -2,8 +2,8 @@
     "name": "if undefined in function",
 
     "backends": {
-        "couchbase": "pendingIgnoreFieldOrder",
-        "marklogic_json": "ignoreFieldOrder",
+        "couchbase": "ignoreFieldOrder",
+        "marklogic_json": "pendingIgnoreFieldOrder",
         "mimir": "ignoreFieldOrder"
     },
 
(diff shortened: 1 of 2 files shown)


alissapajer:
@djspiewak the appveyor build was a success, but it hung at the very end. I'm okay merging as is, would you like it restarted?

it/src/main/resources/tests/ifundefinedInFunction.test
```diff
@@ -2,7 +2,7 @@
     "name": "if undefined in function",
 
     "backends": {
-        "couchbase": "ignoreFieldOrder",
+        "couchbase": "pendingIgnoreFieldOrder",
```
djspiewak: Isn't this actually being caused by a mongo bug?

alissapajer: If you meant to comment on the ignore `_id` line, this may be tracked in #3176.

alissapajer: Also, this is an outdated diff. What are you asking?

Issue #3222, opened 2017-12-01
Avoid thetajoins in cases with differing dimensions


djspiewak:
Easy example:

```sql
select a[*] + b from foo
```

This really shouldn't be a thetajoin.  We can rearrange the `ProjectKey` with `b` to happen after the shift on the `LeftSide` of the results.  This shows up in other areas as well.  For example:

```sql
select a[*][*] + b[*] + c[*][*][*] from foo
```

Same case, just bigger.  This is a generalization of autojoin elimination which often gets confused with #3092.  Resolving this issue is important for the search card.

PR #3287, opened 2017-12-11
Minimize autojoins in dimensional 1-2 cases

djspiewak:
Example:

```sql
select a[*] + b from foo
```

This is pretty narrowly tailored, but it does address part of #3222

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
@@ -168,12 +168,12 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends 
           // candidates.forall(_ ~= QSReduce)
           lazy val reducerCheck = reducerAttempt.lengthCompare(candidates.length) === 0
 
-          lazy val leftShift12Extract = candidates match {
-            case (ls @ LeftShift(src1, _, _, _, _)) :: src2 :: Nil if src1.root === src2.root =>
-              Some((ls.unfold, true))
+          lazy val leftShift12Extract: Option[(QSU.LeftShift[T, QSUGraph], Boolean)] = candidates match {
+            case (ls @ LeftShift(src1, struct, idStatus, repair, rot)) :: src2 :: Nil if src1.root === src2.root =>
+              Some((QSU.LeftShift(src1, struct, idStatus, repair, rot), true))
 
-            case src2 :: (ls @ LeftShift(src1, _, _, _, _)) :: Nil if src1.root === src2.root =>
-              Some((ls.unfold, false))
+            case src2 :: (ls @ LeftShift(src1, struct, idStatus, repair, rot)) :: Nil if src1.root === src2.root =>
+              Some((QSU.LeftShift(src1, struct, idStatus, repair, rot), false))
 
             case _ => None
           }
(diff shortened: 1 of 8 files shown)


djspiewak:
This is going to fail on appveyor because of cache invalidation issues (zinc has a bug w.r.t. the exact change that I made).

djspiewak:
Even aside from the build failures, don't merge this yet.  Reduction coalescence will be broken in some weird cases by this change.

djspiewak:
Ok correction: it doesn't break reduction coalescence, and thus can be merged.  It's just that recursive coalescence is quite naive right now, which results in missing optimization chances.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -159,13 +160,25 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
           coalesceToMap[G](qgraph, candidates, fm)
 
         case None =>
-          val reducerAttempt = candidates collect {
+          lazy val reducerAttempt = candidates collect {
```
wemrysi: `reducerCheck`, forced on L181, will in turn force this, so does the `lazy` actually provide any benefit?

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -159,13 +160,25 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
             case g @ QSReduce(source, buckets, reducers, repair) =>
               (source, buckets, reducers, repair)
           }
 
           // candidates.forall(_ ~= QSReduce)
-          if (reducerAttempt.lengthCompare(candidates.length) === 0) {
+          lazy val reducerCheck = reducerAttempt.lengthCompare(candidates.length) === 0
```
wemrysi: Always forced on L181.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -253,28 +266,35 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
-                  }
+                  _ <- updateForCoalesce[G](candidates, red.root)
                 } yield back
               } else {
                 failure[G](qgraph)
               }
             } yield back
           } else {
-            failure[G](qgraph)
+            leftShift12Extract match {
```
wemrysi: Looks like we need to start thinking about how to factor these transformations out, this is ok for now but not going to scale as we need add more.

wemrysi:
@djspiewak did you intentionally tag this with `version: feature`?

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
djspiewak: Probably not.  This is gone on my follow-up branch.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
djspiewak: Way ahead of you.  :-)  And I totally agree.

djspiewak:
@wemrysi I did.  Not sure exactly if it makes more sense as a `revision`.  It is optimizing a whole class of queries in a user-visible way.  Does that count as a feature?

wemrysi:
@djspiewak I don't recall if we've labeled optimizations as features in the past, but I'm not opposed to the idea.

Issue #3266, opened 2017-12-08
Provide Planner for QScriptCore for Postgres connector


EncodePanda:
Part of https://github.com/quasar-analytics/quasar/issues/3265

PR #3293, opened 2017-12-12
Provide LIMIT (aka Subset/Take) for Postgres

rabbitonweb:
Part of #3266


rabbitonweb:
@wemrysi this is ready for review, it fails on marklogic but (as you probably know) is down to some marklogic specific issues (not related to this change)

rabbitonweb:
Travis cache....

PR #3295, opened 2017-12-12
Add explain to integration tests

rabbitonweb:
With this change you can add flag `"explainQuery": true,` to your integration tests, like this:

```
{
  "name": "select with wildcard",
  "data": "basic.data",
  "query": "select * from basic",
  "predicate": "exactly",
  "ignoreResultOrder": true,
  "ignoredFields": ["_id"],
  "ignoreFieldOrder": true,

  "explainQuery": true,

  "expected": [
    {"name": {"first": "John", "last": "Smith"}, "num": 4},
    {"name": {"first": "Janina", "last": "Kowalska"}, "num": 42}
  ]
}

```

so when you run your integration tests, you will get the explanation of your physical plan. Below example for spark

```
> it/testOnly **.StreamingQueryRegressionSpec -- -ex wild --
[info] Compiling 1 Scala source to /Users/rabbit/projects/quasar/.targets/it/scala-2.11/test-classes...
Running Streaming Queries [.....................................................................helllo!!!!
ExecutionPlan[FileSystemType(spark-local)](inputs = ISet(rootDir </> dir("regression") </> dir("postgres") </> file("basic")))

(2) MapPartitionsRDD[3] at map at QScriptCorePlanner.scala:145 []
 |  MapPartitionsRDD[2] at map at SparkLocal.scala:129 []
 |  /Users/rabbit/temp/t/run6340e4df/regression/postgres/basic MapPartitionsRDD[1] at textFile at SparkLocal.scala:128 []
 |  /Users/rabbit/temp/t/run6340e4df/regression/postgres/basic HadoopRDD[0] at textFile at SparkLocal.scala:128 []
.................................................................................................................................................................................]
[info] StreamingQueryRegressionSpec
[info] 
```

### DIFF
--- a/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
+++ b/it/src/test/scala/quasar/regression/QueryRegressionTest.scala
@@ -128,6 +128,9 @@ abstract class QueryRegressionTest[S[_]](
 
   ////
 
+  private def queryExplain(expr: Fix[Sql], vars: Variables, basePath: ADir): Process[CompExecM, ExecutionPlan] =
+      fsQ.explainQuery(expr, vars, basePath).liftM[Process]
+
   /** Returns an `Example` verifying the given `RegressionTest`. */
   def regressionExample(
     loc: RFile,
@@ -140,7 +143,9 @@ abstract class QueryRegressionTest[S[_]](
       val data = testQuery(
         test.data.nonEmpty.fold(DataDir </> fileParent(loc), rootDir),
         test.query,
-        test.variables)
+        test.variables,
+        test.explainQuery
+      )
 
       (test.data.nonEmpty.whenM(ensureTestData(loc, test, setup)) *>
        verifyResults(test.expected, data, run, backendName))
@@ -265,7 +270,8 @@ abstract class QueryRegressionTest[S[_]](
   def testQuery(
     loc: ADir,
     qry: String,
-    vars: Map[String, String]
+    vars: Map[String, String],
+    explain: Boolean
   ): Process[CompExecM, Data] = {
     val f: Task ~> CompExecM =
       toCompExec compose injectTask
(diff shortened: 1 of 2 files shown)


rabbitonweb:
Not knowing what's going on is seriously a pain when writing new connector. This helps a lot.

Next step will be probably removing that flag and printing put explain whenever test fails (right after the assertions that failed). <-- but that is for another PR

rabbitonweb:
@drostron this PR is ready for review, it fails on Marklogic because there are some marklogic specific issues after Marklogic.

rabbitonweb:
@drostron I deal with bunch of not working tests, I really need this :)

rabbitonweb:
Well I'm happy to have this as a test runner argument (once I know what that means) BUT this is improvement against what we have right now (which is nothing) so.... can we have the wait it is right now so we can make improvements on it later?

all the tests will not have this flag set, so there is literally no change to the ppl who don't care about this feature, but  I will get feedback to the stuff I do right now.

i'm happy to improve it as test runner argument as a next PR, just would like to have that feedback asasp

drostron:
What makes this necessary for master until the time of the planned follow up PR? Feedback on Travis? If so, a topic branch could be useful since Travis instances against our personal repos are under-provisioned.

Specs2's CustomOutput might be what you're looking for:
https://etorreborre.github.io/specs2/guide/SPECS2-3.8.6/org.specs2.guide.CustomOutput.html

rabbitonweb:
Nothing for Travis. We create new connector (postgres) and we need constant feedback loop from tests that are failing. Right now we have repl but that is troublesome. Having this flag allows me to quickly inspect why given test failed for postgress (what was the physical plan for it).

drostron:
Right, so I'm suggesting this stay as a patch used for local debugging until there's time for the preferred approach. Is there too much challenge with this that I'm not seeing?

Also, it might not take that much time to get `CustomOutput` in place, although I haven't taken more than a cursory look.

alissapajer:
Closing this as it can be used as a local patch until implemented as a test runner argument.

@rabbitonweb can you open an issue to track this?

PR #3296, opened 2017-12-12
RDBMS explain support

kciesielski:
This should print output like:
```
SqlExprTree:
Select
├─ selection
│  ╰─ References
│     ├─ Id(_2)
│     ╰─ Constant(Str(a))
╰─ from(_2)
   ╰─ SelectRow
      ├─ selectionInRow(_1)
      │  ╰─ *
      ╰─ fromInRow(_0)
         ╰─ Table(db.foo)

Raw query:
select row_to_json(row) from (select _2.a from (select * from db.foo) _2) row

DB EXPLAIN output:
                       QUERY PLAN                       
--------------------------------------------------------
 Seq Scan on foo  (cost=0.00..41.88 rows=2550 width=32)
```
In case there's an error on calling the DB EXPLAIN query, the last part will be
```
DB EXPLAIN output: unavailable (error description)
```

### DIFF
--- a/rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsQueryFile.scala
+++ b/rdbms/src/main/scala/quasar/physical/rdbms/fs/RdbmsQueryFile.scala
@@ -18,17 +18,17 @@ package quasar.physical.rdbms.fs
 
 import slamdata.Predef._
 import quasar.contrib.pathy.{ADir, AFile, PathSegment}
-import quasar.Data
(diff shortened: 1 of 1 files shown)


rabbitonweb:
Waiting for marklogic to be  fixed before we can merge anything at the moment

rabbitonweb:
@kciesielski merge master to it and push

kciesielski:
@rabbitonweb updated with latest master, waiting for CI and should be good to go.

PR #3297, opened 2017-12-12
Generalize autojoin minimization to apply inductively

### LINKED ISSUE #3289: Recursive autojoin minimization misses adjacency cases
The recursive coalescence is quite naive right now.  As an example, it will miss the ability to optimize anything about the following query:

```sql
select count(a[*]) + sum(b) from foo
```

This will result in a thetajoin and two separate reads and reduce passes, when it should be no join on a single read and a single reduce.

The problem is simply that the recursive call to `coalesceToMap` is single-sourced, and so it can't really do anything other than fetch the `FreeMap` above.  Instead, it should be multisourced so that we can see that the sources can be made common with recursive coalescence.  Even better, this should be generalized so that it also applies to other rewrite types.  For example:

```sql
select count(a)[*] + sum(b) from foo
```

Same story, just inside out.  Also this query returns the empty set by definition, but ignore that for now.  This query demonstrates that we need the recursive coalescence mechanism to be generic and not triggered by specific rewrite types.

djspiewak:
Fixes #3289

This also introduces a real abstraction into `MinimizeAutoJoins` so that we have the ability to better compose these things.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
@@ -39,7 +39,7 @@ import quasar.qscript.qsu.ApplyProvenance.AuthenticatedQSU
 
 import matryoshka.{delayEqual, BirecursiveT, EqualT}
 import matryoshka.data.free._
-import scalaz.{Bind, Equal, Monad, Scalaz, StateT}, Scalaz._   // sigh, monad/traverse conflict
+import scalaz.{Bind, Equal, Monad, OptionT, Scalaz, StateT}, Scalaz._   // sigh, monad/traverse conflict
 
 final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends QSUTTypes[T] {
   import MinimizeAutoJoins._
@@ -69,7 +69,7 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends 
           case RightSide => 1
         }
 
-        coalesceToMap[G](qgraph, List(left, right), combiner2)
+        OptionT(coalesceToMap[G](qgraph, List(left, right), combiner2)).getOrElseF(failure[G](qgraph))
 
       case qgraph @ AutoJoin3(left, center, right, combiner) =>
         val combiner2: FreeMapA[Int] = combiner map {
@@ -78,7 +78,7 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends 
           case RightSide3 => 2
         }
 
-        coalesceToMap[G](qgraph, List(left, center, right), combiner2)
+        OptionT(coalesceToMap[G](qgraph, List(left, center, right), combiner2)).getOrElseF(failure[G](qgraph))
     }
 
     val lifted = back(MinimizationState[T](agraph.dims, Set())) map {
@@ -93,7 +93,7 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends 
       G[_]: Monad: NameGenerator: MonadState_[?[_], RevIdx]: MonadState_[?[_], MinimizationState[T]]: PlannerErrorME](
       qgraph: QSUGraph,
       branches: List[QSUGraph],
-      combiner: FreeMapA[Int]): G[QSUGraph] = {
+      combiner: FreeMapA[Int]): G[Option[QSUGraph]] = {
 
     for {
       state <- MonadState_[G, MinimizationState[T]].get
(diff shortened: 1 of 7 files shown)


connector/src/test/scala/quasar/qscript/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -476,11 +476,67 @@ object MinimizeAutoJoinsSpec extends Qspec with TreeMatchers with QSUTTypes[Fix]
...
           _) =>
 
-          struct must_=== HoleF[Fix]
-
+          struct must beTreeEqual(HoleF[Fix])
           repair must beTreeEqual(func.Add(LeftSideF, RightSideF))
       }
     }
+
+    "inductively coalesce reduces on coalesced shifts" in {
```
edmundnoble: I think it's very likely worth adding more tests here, especially for more than 2 reductions.

djspiewak: I agree.  Some more will be added when we merge, but we'll continue adding more going forward.  This is sort of par for the course for QSU work though: get the test harness in place, and write the tests you need to be sure things minimally work, and pray it holds together long enough for you to eventually write more tests.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -137,21 +142,24 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
       fm: => FreeMapA[Int],
-      candidates: List[QSUGraph]): G[QSUGraph] = candidates match {
+      candidates: List[QSUGraph]): G[Option[QSUGraph]] = candidates match {
 
     case Nil =>
-      updateGraph[G](QSU.Unreferenced[T, Symbol]()) map { unref =>
-        qgraph.overwriteAtRoot(QSU.Map[T, Symbol](unref.root, fm.map(κ(srcHole)))) :++ unref
+      updateGraph[T, G](QSU.Unreferenced[T, Symbol]()) map { unref =>
+        val back = qgraph.overwriteAtRoot(QSU.Map[T, Symbol](unref.root, fm.map(κ(srcHole)))) :++ unref
+        Some(back)
```
edmundnoble: Looks like an artifact of debugging.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -137,21 +142,24 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
       }
 
     case single :: Nil =>
       // if this is false, it's an assertion error
       // lazy val sanityCheck = fm.toList.forall(0 ==)
 
-      qgraph.overwriteAtRoot(QSU.Map[T, Symbol](single.root, fm.map(κ(srcHole)))).point[G]
+      qgraph.overwriteAtRoot(QSU.Map[T, Symbol](single.root, fm.map(κ(srcHole)))).point[G] map { back =>
+        Some(back)
```
edmundnoble: Ditto here.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -160,161 +168,96 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
+                Option(pair).point[G]
+
+              case (None, minimizer) =>
+                if (minimizer.couldApplyTo(candidates)) {
+                  val backM = for {
+                    extractions <- candidates traverse { c =>
+                      OptionT(minimizer.extract[G](c).point[G])
+                    }
+
+                    (exCandidates, exRebuilds) = extractions unzip {
```
edmundnoble: Ditto.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -160,161 +168,96 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
+                    singleSource <-
+                      OptionT(
+                        coalesceToMap[G](
+                          fakeAutoJoin,
+                          exCandidates,
+                          upperFM))
+
+                    (simplifiedSource, simplifiedFM) = singleSource match {
+                      case Map(src, fm) => (src, fm)
+                      case _ => (singleSource, func.Hole)
```
edmundnoble: Nice.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
```diff
@@ -160,161 +168,96 @@ final class MinimizeAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends
...
+
+                    upperMapFragments = (0 until exCandidates.length) map { i =>
+                      func.MakeMapS(i.toString, i.point[FreeMapA])
+                    }
+
+                    upperFM = upperMapFragments.reduceLeft(func.ConcatMaps(_, _))
+
+                    upperFMReduced = upperFM.map(κ(srcHole))
+
+                    fakeAutoJoinM = exCandidates match {
```
edmundnoble: Very interesting.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
djspiewak: Nope it's intentional.

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
djspiewak: Also intentional

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
djspiewak: Working around a bug in scalac here

connector/src/main/scala/quasar/qscript/qsu/MinimizeAutoJoins.scala
edmundnoble: 🐛

Issue #3110, opened 2017-11-15
Plan nested LeftShifts efficiently


jsantos17:
The Search Card and queries including flattening of multiple fields with a common source will be planned using nested left-shifts instead of autojoins. Because of this, the Mongo connector needs to plan nested left shifts efficiently. [More information.](https://gist.github.com/djspiewak/262a5fa0e56d69746dede59320b263f6)

This should NOT depend on SQL^2, but queries like `select loc[*][*] from /mngo/db/zips` and their current QScript can be used as a starting point:

```
LeftShift
├─ LeftShift
│  ├─ ShiftedRead(/db/zips, ExcludeId)
│  ├─ Struct
│  │  ╰─ Guard
│  │     ├─ ProjectKey
│  │     │  ├─ ○
│  │     │  ╰─ Constant(Str(loc))
│  │     ├─ Type(FlexArr(0, None, FlexArr(0, None, Top)))
│  │     ├─ ProjectKey
│  │     │  ├─ ○
│  │     │  ╰─ Constant(Str(loc))
│  │     ╰─ Undefined
│  ├─ IdStatus
│  │  ╰─ IdStatus(ExcludeId)
│  ╰─ Repair
│     ╰─ Guard
│        ├─ RightSide
│        ├─ Type(FlexArr(0, None, Top))
│        ├─ RightSide
│        ╰─ Undefined
├─ Struct
│  ╰─ ○
├─ IdStatus
│  ╰─ IdStatus(ExcludeId)
╰─ Repair
   ╰─ RightSide
```

Currently this QScript plans using `mapReduce`. We need to make sure QScript like this is planned efficiently in aggregation. It's worth noting the old Mongo connector plans this query efficiently with no need for mapReduce.

Importantly, the `Struct` functions in the previous QScript will not look like this for autojoin elimination using left-shifts. It'll instead include nested projections. Details on this are still in the air.

@djspiewak: when rewriting autojoins like this, does the inner `ShiftedRead` have `ExcludeId` or `IncludeId`? We might need to prioritize #3047 for the Search Card if we still need `IncludeId`.

djspiewak:
@jsantos17 So long as there is no actual autojoining on the results at the end (which there wouldn't be for the search card), then it *should* all be `ExcludeId`.  @wemrysi can speak to this more.

wemrysi:
[...]

PR #3298, opened 2017-12-13
Left shifts with unwind

jsantos17:
Here is what I have so far. This is not mergeable yet since there are four failing `it` tests.  The following queries are failing, but they have suboptimal QScript that is making it hard for me to know how to plan this correctly in aggregation:

1. `select loc[*] from zips where loc[*] < 0`. [QScript](https://gist.github.com/jsantos17/c734e652f38c240ad537546655056eb8)

2. `select city, state, loc from zips order by loc[*] desc` [QScript](https://gist.github.com/jsantos17/7bbca8267055b191d8ea6691a1e4da88)

3. `select _id as zip, loc[*] from zips order by loc[*] limit 10` [QScript](https://gist.github.com/jsantos17/55b69aaa603f3a0bbee436f277fe53ce)

4. `SELECT AVG(cnt) as measure, state as category FROM (SELECT COUNT(*) as cnt, state, gender FROM otherpatients WHERE codes[*].desc LIKE "%flu%" GROUP BY state, gender ORDER BY COUNT(*) DESC, state ASC) as meh` [QScript](https://gist.github.com/jsantos17/b3d46cb50f79df51a8986f36fb9ae8c5)

In all of these the `identities` map seems unnecessary and needlessly complicates the query. 

I could simplify the QScript by hand and see if that solves the problem but it might be easier to actually solve the QScript problem instead of working around it in the Mongo connector.

What do you think @wemrysi, @alissapajer, @rintcius, @djspiewak?

Relates to #3110


djspiewak:
So the qscript for a lot of these is going to get simplified radically when the shift coalescence work lands (hopefully tomorrow). We can see what things look like on top of that.

I’m not sure I understand what you mean about the identities map being unnecessary though. It’s very definitely being used.

jsantos17:
@djspiewak Ok, trying to understand this better. For example, [in this query](https://gist.github.com/jsantos17/c734e652f38c240ad537546655056eb8#file-unifyflattenedfields-txt-L303) I see we create a map with `identities` and we then preserve it in the `repair` of the inner `LeftShift`. Then I see the outer `LeftShift`'s `repair`' does not involve `LeftSide` so the `identities` is effectively ignored. Neither the outer nor the inner `LeftShift` `struct`s seem to use the ids either. I see the `ThetaJoin` does use the ids, but that's elided after `QScript (Educated)`. 

What's the flaw in my reasoning here?

djspiewak:
The elided join happens because of a rewrite in qscript (not qsu) that happens to catch this exact case. QSU doesn’t know about that, which is what Edmund and I are working to fix, but since it isn’t known, we have to emit a theta join along with all the associated trappings (including the identity map).

The qscript rewrite is going away eventually. It’s quite naive and probably actually breaks trees in some cases. Once Edmund and my work lands, the identity map won’t exist in cases where joins can be fully eliminated.

jsantos17:
@djspiewak makes sense. For now I'll simplify these by hand to write tests and implement other parts necessary to plan the `LeftShift`s efficently

jsantos17:
Build should be green at least, but this PR is still missing a few things:

1. Handle nested left shifts with `Outline`.
2. Massive refactor of `LeftShift` case handling

Some cases are still handled with `mapReduce` because of suboptimal QScript. Hopefully @djspiewak and @edmundnoble's multiple PRs will help here.

djspiewak:
Test failures in the mongo planner specs.

jsantos17:
@djspiewak this is the good kind of failure since it's pending test that's now (almost) fixed. I'll fix that (it should return `[undefined]` instead of `undefined` when the field is not an array), refactor `case LeftShift => ...`. Once that's done we can merge this and I'll fix more cases in subsequent PRs.

jsantos17:
The Mongo connector is now planning more queries efficiently. Build should be green. Notably, flattening in a predicate condition with references to both `LeftSide` and `RightSide` in the `LeftShift`'s `repair` are not working. Not sure if this is a QScript issue or a Mongo issue, but the QScript can definitely be simplified while keeping the `Filter` node.

Sample query: `select city, loc[*] from zips where loc[*] < -165`

`select city from zips where loc[*] < -165` work, same with `select loc[*] from zips where loc[*] < -165`.

djspiewak:
@rintcius Can you review and (hopefully) merge this asap in the morning?  This is one of the two outstanding PRs which comprise the final piece of 5.0.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -886,31 +892,83 @@ object MongoDbPlanner {
...
+                          builder,
+                          Set(StructureType.Array(DocVar.ROOT(), id))),
+                          repair.as(SrcHole)))
+                  case _ =>
+                    getExprBuilder[T, M, WF, EX](cfg.funcHandler, cfg.staticHandler)(src, struct) >>= (builder =>
+                      getExprBuilder[T, M, WF, EX](
+                        cfg.funcHandler, cfg.staticHandler)(
+                        FlatteningBuilder(
+                          builder,
+                          Set(StructureType.Object(DocVar.ROOT(), id))),
```
rintcius: Looks like these 2 pattern matches can be joined together, since the difference between both cases is small (just `StructureType.Object` vs `StructureType.Array`)

jsantos17:
Thanks for the review. I’ll definitely incorporate the fixes in the
subsequent PRs I have planned.

PR #3299, opened 2017-12-13
Generalized dimensional coalescence to arbitrarily-nested shifts

djspiewak:
This generalizes #3287 to arbitrary nestings (e.g. `a[*][*] + b`).  Depends on #3297.  Affects #3222.  This PR is actually just the single `HEAD` commit.

Assigning to @wemrysi.  Can you take a close look at the way that I'm rewriting root at the end of the shift chains?  Which is to say, lines 141 and 190.  Note that `updateGraph` calls `withName`, updates dimensions, and then returns that node.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
@@ -43,8 +43,11 @@ final class ShiftProjectBelow[T[_[_]]: BirecursiveT: EqualT] private () extends 
   import QSUGraph.Extractors._
 
   def couldApplyTo(candidates: List[QSUGraph]): Boolean = candidates match {
+    case LeftShift(_, _, _, _, _) :: LeftShift(_, _, _, _, _) :: Nil => false
+
     case LeftShift(_, _, _, _, _) :: _ :: Nil => true
     case _ :: LeftShift(_, _, _, _, _) :: Nil => true
+
     case _ => false
   }
 
--- a/connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript.qsu
+package minimizers
+
+import quasar.{NameGenerator, Planner}, Planner.PlannerErrorME
+import quasar.contrib.matryoshka._
+import quasar.contrib.scalaz.MonadState_
+import quasar.ejson.implicits._
+import quasar.fp._
+import quasar.fp.ski.κ
+import quasar.qscript.{
+  HoleF,
+  JoinSide,
+  LeftSide,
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
```diff
@@ -17,31 +17,35 @@
...
-final class ShiftProjectBelow[T[_[_]]: BirecursiveT: EqualT] private () extends Minimizer[T] {
+final class ShiftProjectBelow[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] private () extends Minimizer[T] {
   import MinimizeAutoJoins._
   import QSUGraph.Extractors._
 
+  private val func = construction.Func[T]
+
+  private val N = new NormalizableT[T]
+
   def couldApplyTo(candidates: List[QSUGraph]): Boolean = candidates match {
```
edmundnoble: Is it worth changing this to 
```scala
candidates.exists {
  case LeftShift(_, _, _, _, _) => true
  case _ => false
}
```
To make the reassociated joins easier to deal with in future? Seems clearer than what's happening here, as well.
Regardless, maybe we should have a ticket for the 3-ary case?

djspiewak: Oh yeah this does specifically ignore ternary cases doesn't it.  Crap.  Ok I'll think about that.  `exists` is almost certainly right.

edmundnoble: After offline chat, it's probably better to leave the code here as-is and make a ticket.

connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
```diff
@@ -90,42 +109,107 @@ final class ShiftProjectBelow[T[_[_]]: BirecursiveT: EqualT] private () extends
...
 
-        updateGraph[T, G](QSU.LeftShift[T, Symbol](src.root, struct, idStatus, repair2, rot)) map { rewritten =>
-          val back = qgraph.overwriteAtRoot(rewritten.vertices(rewritten.root)) :++ rewritten
+        if (rest.isEmpty) {
+          val repair2 = fm2 flatMap {
+            case 0 => repair
+            case 1 => func.LeftSide
+          }
+
+          // TODO I'm pretty sure this messes up dimension names
```
wemrysi: I think you're right as `updateGraph` will compute provenance, including the shift identity, for `rewritten.root`, but then `overwriteAtRoot` changes the meaning of `qgraph.root` without recomputing its provenance.

Nodes updated/created with `updateGraph` have their provenance recomputed, so maybe we just need to recompute provenance for the results of `Minimizers` in `MinimizeAutoJoins`?

connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
```diff
@@ -90,42 +109,107 @@ final class ShiftProjectBelow[T[_[_]]: BirecursiveT: EqualT] private () extends
...
+                val struct2 = struct.flatMap(κ(func.ProjectKeyS(func.Hole, "results")))
+
+                val repair2 = repair flatMap {
+                  case LeftSide => func.ProjectKeyS(func.LeftSide, "results")
+                  case RightSide => func.RightSide
+                }
+
+                // we use right-biased map concat to our advantage here and overwrite results
+                val repair3 = func.ConcatMaps(
+                  func.LeftSide,
```
wemrysi: As discussed offline, this may or may not be an advantage given the current limitations of `MapFunc` normalization not knowing the shape of `Hole/LeftSide/RightSide`.

connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
djspiewak: That's kind of what I'm thinking. It's interesting that we don't hit any failures because of this.

connector/src/main/scala/quasar/qscript/qsu/minimizers/ShiftProjectBelow.scala
djspiewak: Yeah I'll fix this up in another PR.

PR #3304, opened 2017-12-13
QSU Group Key Reification

### LINKED ISSUE #3170: Nested select bug
```sql
select city, cnt as c from (select city, count(*) as cnt from patients where state = "NY" group by city order by count(*) desc limit 10) as meh group by city order by c desc
```

The above query should return the same results as the bug reported in https://github.com/quasar-analytics/quasar/issues/2227, which manifests as the similar query:

```sql
select city, sum(cnt) as c from (select city, count(*) as cnt from patients where state = "NY" group by city order by count(*) desc limit 10) as meh group by city order by c desc
```

### LINKED ISSUE #3252: `flattenGroupedObject.test` failing for all connectors
```sql
select distinct commit{*}, count(*) from slamengine_commits where commit{*} like "http%" group by committer.login
```

Needs investigation into why it's failing. As it's failing for all connectors, it's probably a qscript bug.

wemrysi:
Group keys are now first-class identities along side shift ids and reducer buckets. This allows them to be properly reified when referenced in an autojoin.

Fixes #3170.
Fixes #3252.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/MapFuncCore.scala
+++ b/connector/src/main/scala/quasar/qscript/MapFuncCore.scala
@@ -398,6 +398,8 @@ object MapFuncCore {
 
       case ConcatMaps(Embed(lhs), Embed(StaticMap(Nil))) => lhs.some
 
+      case ConcatMaps(Embed(lhs), Embed(rhs)) if lhs ≟ rhs => lhs.some
+
       case _ => none
     }
 
--- a/connector/src/main/scala/quasar/qscript/provenance/Prov.scala
+++ b/connector/src/main/scala/quasar/qscript/provenance/Prov.scala
@@ -141,13 +141,13 @@ trait Prov[D, I, P] {
     })
   }
 
-  // eliminate duplicates within contiguous Both/OneOf and reassociate to the left
+  // eliminate duplicates within contiguous Both/OneOf and reassociate to the right
   def normalize(p: P)(implicit eqD: Equal[D], eqI: Equal[I]): P = {
     def normalize0(alternates: NEL[NEL[P]]): P =
       alternates
-        .map(_.distinctE1.foldLeft1(both(_, _)))
+        .map(_.distinctE1.foldRight1(both(_, _)))
         .distinctE1
-        .foldLeft1(oneOf(_, _))
+        .foldRight1(oneOf(_, _))
 
     val normalizeƒ: Algebra[PF, NEL[NEL[P]]] = {
       case Both(l, r)  => (l |@| r)(_ append _)
--- a/connector/src/main/scala/quasar/qscript/qsu/IdAccess.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/IdAccess.scala
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript.qsu
+
+import slamdata.Predef._
+import quasar.RenderTree
+import quasar.fp.symbolOrder
+
+import matryoshka._
+import monocle.{Prism, PPrism, Traversal}
+import scalaz.{Applicative, Equal, Order, Show, Traverse}
+import scalaz.std.anyVal._
+import scalaz.std.option._
+import scalaz.std.tuple._
+import scalaz.syntax.applicative._
+import scalaz.syntax.either._
+import scalaz.syntax.show._
+
+/** Describes access to the various forms of ids. */
+sealed abstract class IdAccess[D]
+
+object IdAccess extends IdAccessInstances {
+  final case class Bucket[D](of: Symbol, idx: Int) extends IdAccess[D]
+  final case class GroupKey[D](of: Symbol, idx: Int) extends IdAccess[D]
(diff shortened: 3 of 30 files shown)


connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -398,6 +398,8 @@ object MapFuncCore {
 
       case ConcatMaps(Embed(lhs), Embed(StaticMap(Nil))) => lhs.some
 
+      case ConcatMaps(Embed(lhs), Embed(rhs)) if lhs ≟ rhs => lhs.some
```
djspiewak: I also want to add a case to handle `ProjectKeyS(ConcatMaps(_, MakeMapS(key1, rhs)), key2) if key1 === key2`, which will be `rhs`.  That's unrelated to this PR though.  :-)

connector/src/main/scala/quasar/qscript/provenance/Prov.scala
```diff
@@ -141,13 +141,13 @@ trait Prov[D, I, P] {
     })
   }
 
-  // eliminate duplicates within contiguous Both/OneOf and reassociate to the left
+  // eliminate duplicates within contiguous Both/OneOf and reassociate to the right
```
djspiewak: Why?

connector/src/main/scala/quasar/qscript/qsu/IdAccess.scala
```diff
@@ -0,0 +1,112 @@
...
+import scalaz.syntax.show._
+
+/** Describes access to the various forms of ids. */
+sealed abstract class IdAccess[D]
+
+object IdAccess extends IdAccessInstances {
+  final case class Bucket[D](of: Symbol, idx: Int) extends IdAccess[D]
+  final case class GroupKey[D](of: Symbol, idx: Int) extends IdAccess[D]
+  final case class Identity[D](of: Symbol) extends IdAccess[D]
+  final case class Static[D](data: D) extends IdAccess[D]
```
djspiewak: Outstanding.

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -225,6 +232,12 @@ object QScriptUniform {
       right: A,
       combiner: FreeMapA[T, JoinSide3]) extends QScriptUniform[T, A]
 
+  final case class QSAutoJoin[T[_[_]], A](
```
djspiewak: Wait… what?

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -715,6 +738,12 @@ object QScriptUniform {
     def tread1(name: String): T[QSU] =
       tread(Path.rootDir </> Path.file(name))
 
+    // undefined
+    val undefined: Prism[T[QSU], Unit] =
```
djspiewak: :+1:

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
```diff
@@ -776,7 +805,7 @@ object QScriptUniform {
             .composeIso(envTIso[A, QSU, CoQSU])
 
         def lifting[S, B]: Prism[S, B] => Prism[(A, S), (A, B)] =
-	  _.second[A]
+          _.second[A]
```
djspiewak: lol…

connector/src/test/scala/quasar/qscript/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -426,8 +424,66 @@ object MinimizeAutoJoinsSpec extends Qspec with TreeMatchers with QSUTTypes[Fix]
...
+        func.Lower(func.ProjectKeyS(func.Hole, "city"))
+
+      val groupedGuardedRead =
+        qsu.dimEdit(
+          qsu.autojoin3((
+            qsu.read(afile),
+            qsu.read(afile),
+            qsu.undefined(),
+            _(MapFuncsCore.Guard(_, Type.AnyObject, _, _)))),
+          DTrans.Group(groupKey))
```
djspiewak: This is kind of weird, because I thought that `DimEdit` was eliminated by the time we get to `MinimizeAutoJoins`.  If it's not, then we're kind of in trouble, since we don't know how to merge `LPReduce` (only `QSReduce`).

connector/src/main/scala/quasar/qscript/MapFuncCore.scala
wemrysi: Yeah, I actually need to writeup an issue as we need more sophisticated handling of all of this, hopefully using `Outline` so we can see shapes through `Hole`.

connector/src/main/scala/quasar/qscript/provenance/Prov.scala
wemrysi: It is easier, for me at least, to read right-associated binops then left-associated ones.

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
wemrysi: It allows us to have a much more precise join condition, i.e. just identity comparisons, no `FreeMap`s. It also allows us to definitively handle identity map merging for autojoins in `ReifyIdentities` (we don't have to come up with a heuristic to decide if a join condition was for an autojoin).

connector/src/main/scala/quasar/qscript/qsu/QScriptUniform.scala
wemrysi: 😞

connector/src/test/scala/quasar/qscript/qsu/MinimizeAutoJoinsSpec.scala
wemrysi: It is, `runOn` applies `ApplyProvenance >>> ReifyBuckets >>> MinimizeAutoJoins`. It insulates us from having to care about the particular names of vertices.

wemrysi:
@djspiewak resolving them now...

wemrysi:
@djspiewak @edmundnoble Conflicts resolved.

djspiewak:
Will merge once built

PR #3307, opened 2017-12-13
Generalize shift below merge to multi left shift

### LINKED ISSUE #2133: QScript needs to merge `LeftShift` containing `Typecheck`
In the query
```
select * from nested_foo where (foo{*} = 5 or foo[*] = 6)
```
we need to merge the two sides of the `or` in qscript. Currently this fails to merge in qscript and creates an unnecessary theta join.

Here is the relevance `LogicalPlan`:
```
      ╰─ Invoke(Or)
         ├─ Invoke(Eq)
         │  ├─ Invoke(FlattenMap)
         │  │  ╰─ Let('__tmp2)
         │  │     ├─ ObjectProject(Free('__tmp0){"foo"})
         │  │     ╰─ Typecheck(Obj(Map[], Some(Top)))
         │  │        ├─ Free('__tmp2)
         │  │        ├─ Free('__tmp2)
         │  │        ╰─ Constant(Obj(Map( -> NA)))
         │  ╰─ Constant(Int(5))
         ╰─ Invoke(Eq)
            ├─ Invoke(FlattenArray)
            │  ╰─ Let('__tmp3)
            │     ├─ ObjectProject(Free('__tmp0){"foo"})
            │     ╰─ Typecheck(FlexArr(0, None, Top))
            │        ├─ Free('__tmp3)
            │        ├─ Free('__tmp3)
            │        ╰─ Constant(Arr(List(NA)))
            ╰─ Constant(Int(6))
```
This should enable the integration test `multitypeFlatten.test`. It is also possible that this work will enable a handful of other flatten-related integration tests.

### LINKED ISSUE #3092: Adjacent LeftShift collapse
Subsumes part of #2998 

Look for a join on two `LeftShift` nodes with a common source.  Put them together one-above-the-other (which one doesn't matter), ensuring that original value access works.  This will fix the issue in #2935 and likely also #2366

Obviously this will be implemented as a rewrite on the graph, probably after `ReifyProvenance`.

### LINKED ISSUE #3222: Avoid thetajoins in cases with differing dimensions

djspiewak:
I'm just going to leave this right here…

```sql
select a[*][*][*] + b - c[*] / d[*][*] from foo
```

QScript: https://gist.github.com/7024e91afdcaeee9b016ef427d5d5117

<kbd>Cmd</kbd>-<kbd>F</kbd> `theta`

*No Results Found*

:tada:

Ping @jdegoes @alissapajer 

Depends on #3299
Fixes #2133
Fixes #3092
Fixes #3222

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript.qsu
+package minimizers
+
+import quasar.{NameGenerator, Planner}
+import Planner.PlannerErrorME
+import quasar.contrib.matryoshka._
+import quasar.contrib.scalaz.MonadState_
+import quasar.fp._
+import quasar.fp.ski.κ
+import quasar.qscript.{construction, Hole, SrcHole}
+import quasar.qscript.qsu.{QScriptUniform => QSU}
+import slamdata.Predef._
+import matryoshka.{BirecursiveT, EqualT, ShowT, delayEqual}
+import slamdata.Predef
+
+import scalaz.{\/, -\/, \/-, Monad, Monoid, Scalaz}
+import Scalaz._
+
+final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT] private () extends Minimizer[T] {
+  import MinimizeAutoJoins._
+  import QSUGraph.Extractors._
+
+  val func = construction.Func[T]
+  val srcHole: Hole = SrcHole
+
+  def extractShifts(candidates: List[QSUGraph]): List[QSU.LeftShift[T, QSUGraph] \/ QSU.MultiLeftShift[T, QSUGraph]] =
+    candidates collect {
+      case LeftShift(source, struct, ids, repair, rot) =>
+        QSU.LeftShift(source, struct, ids, repair, rot).left
+      case MultiLeftShift(source, shifts, repair) =>
+        QSU.MultiLeftShift(source, shifts, repair).right
+    }
+
+  def couldApplyTo(candidates: Predef.List[QSUGraph]): Predef.Boolean =
+    candidates.lengthCompare(extractShifts(candidates).length) === 0
+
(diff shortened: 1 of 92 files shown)


djspiewak:
Ok this is an extremely bizarre commit list.  Also I was excited and filed my PR and didn't wait for @edmundnoble's conflict resolution.  So uh… don't merge yet.

djspiewak:
I think this is failing because of the `ShiftArray` issue.

jdegoes:
> select a[*][*][*] + b - c[*] / d[*][*] from foo

Sweeeet! 🎉

djspiewak:
@wemrysi If you can do a final review and merge on this tonight, that would be *greatly* appreciated.  I'd like to get it into master asap so we can get the build cycles through and have time tomorrow morning to run the downstream builds, which also take a fair amount of time.

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -0,0 +1,142 @@
...
+          case (-\/(ls), idx) => ls.repair.map {
+            case QSU.LeftTarget() => scala.sys.error("QSU.LeftTarget in CollapseShifts")
+            case QSU.AccessLeftTarget(access) => access.left[Int]
+            case QSU.RightTarget() => idx.right[QAccess[Hole]]
+          }
+          case (\/-(mls), idx) => mls.repair.map(_.map(_ + idx))
+        }
+        val shiftsLength = shifts.length
+        val repair = fm.flatMap { idx =>
+          // get is safe because fm's indices line up with updatedShifts'
```
wemrysi: Why not use it then?

connector/src/test/scala/quasar/qscript/qsu/ExpandShiftsSpec.scala
```diff
@@ -0,0 +1,250 @@
...
+            func.Eq(
+              func.AccessLeftTarget(Access.id(IdAccess.identity[Fix[EJson]]('qsu0), _)),
+              func.AccessLeftTarget(Access.id(IdAccess.identity[Fix[EJson]]('qsu1), _))),
+            func.ConcatMaps(
+              func.AccessLeftTarget(Access.valueHole(_)),
+              func.MakeMapS("1", func.RightTarget)),
+            func.Undefined
+          )
+        )
+        ok
```
wemrysi: Can this (and all the others) be removed?

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
djspiewak: It requires enabling more warts.

djspiewak:
The other comments are basically re @edmundnoble. Flagging for later attention

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
wemrysi: I'd consider that a feature 😄

Issue #3114, opened 2017-11-16
Fix autojoin key handling for 'Both' provenance


wemrysi:
Creating autojoin keys for `Both` provenance needs to properly handle the case where the left/right sides of the `Both` have the same provenance (say, from a self-join) and, for some matching candidate provenance `x` construct `(x == l) || (x == r)`.

This also means we need to resurrect constructing identity getters when creating join keys so we know how to access identities, even with unique symbols, since the same symbol may appear on either side of a join in the aforementioned self-join scenario.

djspiewak:
One of the demo queries hits this issue, btw:

```sql
SELECT COUNT(*) AS count FROM (SELECT o.order_key AS OrderKey, o.purchase_date AS PurchaseDate, c.last_name || ", " || c.first_name AS Name, c.city AS City, c.state AS State, i.clothing_size AS Size, i.clothing_type AS Type, i.clothing_color AS Color, i.price AS UnitPrice, oi.qty AS Quantity, i.price * oi.qty AS PricePaid, SUM(i.price * oi.qty) AS TotalPaid, i.clothing_brand AS Brand FROM (SELECT * FROM (SELECT * FROM `/hdfs/customers.ldjson`) AS res WHERE res.state IN ("AK")) AS c INNER JOIN `/hdfs/orders.ldjson` AS o ON c.customer_key = o.customer_key INNER JOIN `/hdfs/ordered_items.ldjson` AS oi ON oi.order_key = o.order_key INNER JOIN `/hdfs/inventory_items.ldjson` AS i ON i.item_key = oi.item_key GROUP BY i.price * oi.qty) AS tmp__1
```

wemrysi:
I experimented enabling the naive identity map merging on this query and it now compiles and runs, but returns a count of `0`, when it should be non-zero given my local data.

The resulting plan is quite complicated so I'm not sure if it is failing due to the identities or something else. This may mean that more queries are affected by this than I originally anticipated.

Issue #3309, opened 2017-12-14
Dependent dimensions missing during MinimizeAutoJoins


wemrysi:
In `v27.5.5`, attempting to compile
```sql
select * from (select pop, loc[_] as l, state from zips) as z inner join `demo/patients` as p on z.state = p.state
```
fails with
```
Quasar error: [MinimizeAutoJoins] Dependent dimensions not found: ['qsu0].
```

<details>
<summary><code>AuthenticatedQSU</code> handed to MinimizeAutoJoins</summary>

```
ReifyBuckets: AuthenticatedQSU {
QSUGraph('qsu36)[
  'qsu31 -> Map('qsu2, Constant(Str(left)))
  'qsu32 -> AutoJoin2('qsu28, 'qsu31, ProjectKey(LeftSide, RightSide))
  'qsu5 -> Map('qsu2, Constant(Str(pop)))
  'qsu6 -> AutoJoin2('qsu4, 'qsu5, ProjectKey(LeftSide, RightSide))
  'qsu7 -> AutoJoin2('qsu5, 'qsu6, MakeMap(LeftSide, RightSide))
  'qsu8 -> Map('qsu2, Constant(Str(l)))
  'qsu9 -> Map('qsu2, Constant(Str(loc)))
  'qsu10 -> AutoJoin2('qsu4, 'qsu9, ProjectKey(LeftSide, RightSide))
  'qsu11 -> AutoJoin3('qsu10, 'qsu10, 'qsu3, Guard(LeftSide3, FlexArr(0, None, Top), Center, RightSide3))
  'qsu12 -> LeftShift('qsu11, SrcHole, ExcludeId, RightSide, ShiftArray)
  'qsu13 -> AutoJoin2('qsu8, 'qsu12, MakeMap(LeftSide, RightSide))
  'qsu14 -> AutoJoin2('qsu7, 'qsu13, ConcatMaps(LeftSide, RightSide))
  'qsu0 -> Read(/zips)
  'qsu1 -> LeftShift('qsu0, SrcHole, ExcludeId, RightSide, ShiftMap)
  'qsu4 -> AutoJoin3('qsu1, 'qsu1, 'qsu3, Guard(LeftSide3, Obj(Map[], Some(Top)), Center, RightSide3))
  'qsu15 -> Map('qsu2, Constant(Str(state)))
  'qsu16 -> AutoJoin2('qsu4, 'qsu15, ProjectKey(LeftSide, RightSide))
[...]

PR #3310, opened 2017-12-14
Remap ThetaJoin condition in ReifyIdentities.

wemrysi:
If a `ThetaJoin` branch carries an IV map, we need to remap value access to that side in the join condition.

Also adds a nicer error message for the unimplemented case tracked in #3114.


wemrysi:
I attempted to add some integration tests that would trigger the issue fixed here, but the queries that I think will do so won't compile at the moment (see #3309).

I'll make sure to include tests for these cases in the main `ReifyIdentities` testing PR.

PR #3311, opened 2017-12-14
fix checksum marklogic failure

nicflores:
Fix checksum yum install failure of Marklogic.rpm.

### DIFF
--- a/it/src/main/resources/tests/demo/groupedSelectOfNestedReduction.test
+++ b/it/src/main/resources/tests/demo/groupedSelectOfNestedReduction.test
@@ -2,7 +2,7 @@
     "name": "select unreduced grouped values from nested select with reduction",
     "backends": {
         "couchbase":         "skip",
-        "marklogic_json":    "skip"
+        "marklogic_json":    "pending"
     },
     "NB": "disabled on couchbase due to lack of general join",
     "data": "patients.data",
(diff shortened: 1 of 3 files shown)


it/src/main/resources/tests/demo/groupedSelectOfNestedReduction.test
```diff
@@ -2,7 +2,7 @@
     "name": "select unreduced grouped values from nested select with reduction",
     "backends": {
         "couchbase":         "skip",
-        "marklogic_json":    "ignoreFieldOrder"
+        "marklogic_json":    "skip"
```
wemrysi: Can this just be `pending`?

it/src/main/resources/tests/flattenGroupedObject.test
```diff
@@ -2,8 +2,9 @@
     "name": "flatten a grouped object with filter",
     "backends": {
         "couchbase":         "pending",
-        "marklogic_json":    "pending",
-        "marklogic_xml":     "pending",
+        "marklogic_json":    "skip",
```
wemrysi: Same here?

it/src/main/resources/tests/flattenGroupedObject.test
```diff
@@ -2,8 +2,9 @@
...
+        "marklogic_xml":     "skip",
+        "mimir":             "pending",
```
wemrysi: Is this unintended, or did `mimir` actually start failing?

nicflores:
@wemrysi, thanks for taking a look. I've made the suggested changes and pushed.

it/src/main/resources/tests/flattenGroupedObject.test
djspiewak: It was made `skip` mostly because it's flaky.  It's right at the timeout boundary line, and will sometimes pass, sometimes fail.

Issue #3313, opened 2017-12-15
Integration tests reorganization


EncodePanda:
I want to clean up integration tests. I will divide them into three groups: regression, demo and qa.

```
`demo` - tests from workspaces
`regression` - tests from discovered bugs
`qa` - tests that allow you to iteratively build a connector (edited)
```

each group will have a subfolder
each test will be prefixed with the group so you can quickly run all integration tests from given group
also `qa` will be divided into steps (iterations) so that future connector creators can have little milestones when creating new connector
`qa/step1_simple_read` holding things like `select * from foo` `select a, b from foo`
`qa/step2_sort` holding things like `select a, b  from foo ordered by a asc, b desc`

each step will build in mind that previous steps are working

so `qa_step2_sort` assumes that reads are working because those were covered in `qa_step1_simple_read`

PR #3314, opened 2017-12-15
It tests refactor

rabbitonweb:
Part of #3313


rabbitonweb:
Reassigning bcs @alissapajer is OOO

djspiewak:
So the names of the tests that you moved got pretty universally changed to a space-free, semi-mangled encoding.  Is there a particular reason for that?  The names are meant to be human-friendly, so there's no reason to mangle them.

rabbitonweb:
@djspiewak actually in other branch I'm unifying those names like that - it's actually for a reason to be able to run a batch of tests prefixed with given phrase. Whitespaces makes it problematic (too many tests run when actually is needed). Does it make sense?

rabbitonweb:
@djspiewak I in future want to rename it to `demo_games_step1.test`, so
1. when I want to run all demo tests I do

```it/testOnly **.StreamingQueryRegressionSpec -- -ex demo -```
2. whenI want to run all "games" tests I do

```it/testOnly **.StreamingQueryRegressionSpec -- -ex demo_games_step --```

**Comment by wemrysi** [comment]
@rabbitonweb Could we maybe "tag" similar issues with a key, but keep the human readable names e.g. `[demo_games] Some nice description for a person.`?

**Comment by wemrysi** [comment]
Or maybe we add a `group` key to the test metadata and when we convert the test into a specs2 example, we "tag" the test using the group key automatically, when present. This would keep them as data, which would be nice.

**Comment by rabbitonweb** [comment]
@wemrysi I might go with "tag" first and then maybe in the futue do the second thing you proposed

**Comment by rabbitonweb** [comment]
@djspiewak @wemrysi Are you gusy ok with keeping this order I propse in folder structure:

```
|-demo
|-regression
|-qa
   |-s01
   |-s02
```
?

rabbitonweb:
@djspiewak if I "tag" and bring whitespaces back, will this be 👍 ?

djspiewak:
@rabbitonweb Yep!

rabbitonweb:
I'm closing this and I will open new PR with more changes because this is already too old and I dont want to backport those changes

PR #3327, opened 2017-12-18
Fix Mimir returning `NotFound` on root if empty

### LINKED ISSUE #2473: Empty mounts are listed as non-existent
Ran into this and thought it was a different issue (#2472).  If you create a mount that does not have any data in it (no dirs and no files), the repl will believe that it doesn't exist even though it displays as present.  The result will be the same error as if there was no mount at that path whatsoever.  Note that this is only relevant to data stores which don't have "default" datasets (so, not mongo) such as mimir and marklogic.

A workaround is to upload the test dataset.  But really, we should ensure that empty mounts are allowed and meaningful, since it doesn't make any sense for them to cause errors.

jedesah:
Fix #2473

- Add test for `HierarchicalFS` to make sure it handles the root of a mount properly
- Remove `pending` from "root dir" test for connectors
- Fix `externalFileSystems` to provide "non-chrooted" interpreters to it tests
- Fix `testBackendConfig` to avoid crashing when there are no external connectors to test
- Add `mimir` to `testing.conf.example`

### DIFF
--- a/core/src/test/scala/quasar/fs/mount/HierarchicalFileSystemSpec.scala
+++ b/core/src/test/scala/quasar/fs/mount/HierarchicalFileSystemSpec.scala
@@ -222,6 +222,11 @@ class HierarchicalFileSystemSpec extends quasar.Qspec with FileSystemFixture {
           runMntd(query.ls(rootDir </> dir("bar")).run)
             .eval(emptyMS) must_=== mnts.right
         }
+
+        "of empty mount should return empty set" >> {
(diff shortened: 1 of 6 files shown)


jedesah:
I half expect another connector to fail on the test I re-enabled for all connectors but I don't feel like running them all locally to find out which ones

jedesah:
Oh, wow, looks like all the other connectors have passed the newly enabled test!

PR #3328, opened 2017-12-19
Improve LeftShift Mongo plans

jsantos17:
This is a second implementation that allows for parts of the `LeftShift`'s `repair` function to be planned in `mapReduce` while planning the rest of the `repair` in aggregation. I expected some more tests to become green.

### DIFF
--- a/mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
+++ b/mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
@@ -401,14 +401,10 @@ class PlannerQScriptSpec extends
                 $field("loc"),
                 $arrayLit(List($literal(Bson.Undefined)))))),
           $unwind(DocField("f")),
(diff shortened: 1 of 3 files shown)


mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
```diff
@@ -401,14 +401,10 @@ class PlannerQScriptSpec extends
                 $field("loc"),
                 $arrayLit(List($literal(Bson.Undefined)))))),
           $unwind(DocField("f")),
+          $match(Selector.Doc((BsonField.Name("f")) -> Selector.Lt(Bson.Int32(-165)))),
```
jsantos17: This is a concrete case where the planner is able to re-order pipeline stages (`$match before $project`) for efficiency.

jsantos17:
@rintcius I don't think there's anything left to update. So this is ready for another look.

Issue #3319, opened 2017-12-15
Plan LeftShifts with IncludeId using `$unwind`'s includeArrayIndex


jsantos17:
Relevant documentation: https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/#pipe._S_unwind

When planning `LeftShift`s, the `Id` flag is simply passed through inside the newly created `FlatenningBuilder`. The choice of whether to plan the `FlatteningBuilder` with `$unwind` (agg) or `$simpleMap` (mapReduce) is made in the `toWorkflow` function.

zeeiyerWork:
In talking with @jsantos17 - several queries do need it (of the type -> `select a[*] + b[*] from foo`) and @rintcius is taking a look - hence the December/5.0.0 labels. Thanks.

PR #3330, opened 2017-12-19
add IT summing flattened int arrays

rintcius:
Relates to #3319


wemrysi:
Interesting that `mimir` is the only connector passing, I wonder if it has to do with the `Cond` in the repair.

rintcius:
@wemrysi yeah, mongo even gives an error (see qz-3329)

PR #3334, opened 2017-12-19
Use set intersection for multiple ids in autojoin conditions

wemrysi:
Corrects how autojoin conditions are generated when provenance involves multiple ids. Previously we were implementing cartesian equality when the correct thing to do is use set intersection.


djspiewak:
MarkLogic test failure.

PR #3335, opened 2017-12-20
Use ShiftType to plan LeftShift on arrays instead of shifty inference from Guards

jsantos17:


mongodb/src/test/scala/quasar/physical/mongodb/PlannerQScriptSpec.scala
```diff
@@ -507,11 +507,11 @@ class PlannerQScriptSpec extends
                   $lte($literal(Bson.Arr()), $field("0")),
                   $lt($field("0"), $literal(Bson.Binary.fromArray(scala.Array[Byte]())))),
                 $field("0"),
-                $literal(Bson.Undefined)))),
+                $arrayLit(List($literal(Bson.Undefined)))))),
           $unwind(DocField("0")),
           $project(reshape(
             sigil.Quasar -> $field("0")))))
-    }.pendingUntilFixed
+    }
```
jsantos17: Not done yet since double flattening with reference to `LeftSide`  still uses `mapReduce` for some reason. I'm unhappy with `rewriteUndefined` and I need to put some more thought on it.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -118,15 +118,15 @@ object MongoDbPlanner {
...
+    (fm: FreeMap[T], default: OutputM[PartialSelector[T]], galg: GAlgebra[(T[MapFunc[T, ?]], ?), MapFunc[T, ?], OutputM[PartialSelector[T]]])
     (implicit inj: EX :<: ExprOp)
       : OutputM[PartialSelector[T]] =
     fm.zygo(
       interpret[MapFunc[T, ?], Hole, T[MapFunc[T, ?]]](
         κ(MFC(MapFuncsCore.Undefined[T, T[MapFunc[T, ?]]]()).embed),
         _.embed),
       ginterpret[(T[MapFunc[T, ?]], ?), MapFunc[T, ?], Hole, OutputM[PartialSelector[T]]](
-        κ(defaultSelector[T].point[OutputM]), galg))
+        κ(default), galg))
```
jsantos17: By default a `SrcHole` indicates we need to extract a selector, however, a `SrcHole` does *not* imply we need to extract a *type* selector. Type selectors are generated by `Guard`s on `ProjectKey` on `SrcHole`, hence the refactor to provide a default selector for each phase.

PR #3338, opened 2017-12-20
2.12 migration

romanowski:
This PR represents  @lukaszwawrzyk's and mine work to migrate quasar to Scala 2.12.4 (what also means migrating away from Typelevel Scala).

This is planned to be merged as one big bang change so this branch will live for a while.

Reaming issues (list will be updated with every new issue faced):

- [ ] We need to properly deal with Spark-related modules (they are commented out ATM) 
- [x] scalaz-spec2 is incompatible with specs 4.0.2 that was included due to smock. We are waiting for our PR to be merged and new version released.
- [x] For now we disabled `-Xfatal-warnings` flag since 2.12 introduced new warnings
- [x] We need to slightly refactor the way how scalac flags are handled (I don't like using `-=` or `--=`). I was said that this is ok, so we may keep it
- [x] Possible issues from Travis build
- [x] Specs2 was really slow before 4.0.2 version


djspiewak:
We can update smock. That's not a hard problem.

Regarding spark, one solution is to cross build everything connector and above, maintaining backwards compatibility with Scala 2.11.  But I don't think that's a particular good solution, so we might have to just keep spark out of the aggregation for the time being.

Regarding scalac flags…  I've had to use `-=` and `--=` in almost every project I've ever worked on.  I agree it's pretty ugly, but I have yet to find a nicer way of factoring it that doesn't end up with a lot of redundancy.

lukaszwawrzyk:
@djspiewak It turns out that bumping specs to 3.9.1 leads to serious performance issues, going back to 3.8.6 seems to solve the problem but we would need smock released for scala 2.12 and specs2 3.8.6.

djspiewak:
@lukaszwawrzyk Runtime performance or compile time?  Also is there a reason we can't bump to 4.0?

lukaszwawrzyk:
@djspiewak It is runtime performance, in some cases even more than 10 times slower with 3.9.1 than with 3.8.6. This was the reason of failing tests on CI - timeouts. I didn't try bumping to 4.0, we went with the approach of bumping libraries by as little as possible to avoid potential problems caused by serious changes in their code.

djspiewak:
That's amazingly slower… What tests were timing out?  The integration tests?

lukaszwawrzyk:
Not the integration tests. For example `StdLibTypeCoherenceSpec`. For example commenting out everything but `RelationsLib` leads to tests running for 7s on 3.8.6 and over 70s on 3.9.1. I verified this also on master branch with scala 2.11. Each test runs with similar speed, but time between tests differs a lot. It seems that the reason is migrating away from scalaz in specs2, as with 3.9.1 a lot of time is spent in `org.spec2.control.eff.Attrs.apply` on `Vector` concatenation and unapply from `+:`. I didn't analyze it more deeply yet though. I also don't know yet if problem persists with 4.0.

romanowski:
@djspiewak @lukaszwawrzyk It seems that spec2 4.0.0 improve performance a lot especially for deeply nested test fragments when combined with exclude.
[This minimal project](https://github.com/romanowski/spec-bug) that mimic quasar tests needs **14 sec (sic!)** to execute 16 tests that simply returns `success` and only **1 sec** when 4.0.0 is used.

For now I remove exclude from tests parameters to see if tests will be successful in 3.9.5 and we need to migrate to 4.0.0 soon.

djspiewak:
@romanowski I'm all in favor of migrating forward to 4.0.0 asap.  Doesn't have to be in this PR, but if it's more convenient in this PR, then :+1: from me.

djspiewak:
@romanowski I will say though that I'm very :-1: on inlining dependencies just for the sake of an upstream upgrade.  It's ok for now as we see how things work, but before this lands, I'd rather move to a version of smock and scalaz-specs2 that supports 4.0.0.  I can upgrade smock pretty trivially, and it's easy enough for me to chase down people to make sure scalaz-specs2 gets upgraded.

romanowski:
@djspiewak Inlining of deps was just to quickly test if 4.0.2 will fix performance problems. So now next step is to migrate to smock and scalaz-specs versions based on specs 4.x. We will open PRs to both libs.

djspiewak:
<3

lukaszwawrzyk:
@djspiewak Here is the pr for smock https://github.com/djspiewak/smock/pull/3
I am working now on scalaz-specs2

djspiewak:
John let me know that this was ready for review, but it looks like it's still missing some commits?  The build should be passing, at a minimum.

romanowski:
@djspiewak Travis was [green](https://travis-ci.org/quasar-analytics/quasar/builds/324477265?utm_source=github_status&utm_medium=notification) but appveoyr for some reason (sby said build was 
 a success) was marked as red. @lukaszwawrzyk add dummy commits to retrigger builds.

The only missing thing was new version of smock so IMHO it is ready to be reviewed.

djspiewak:
Cool!  I'll get on reading through this today.

frontend/src/main/scala/quasar/sql/parser.scala
```diff
@@ -463,11 +464,15 @@ private[sql] class SQLParser[T[_[_]]: BirecursiveT]
       case k ~ h => GroupBy(k, h)
     }
 
-  def order_by: Parser[OrderBy[T[Sql]]] =
-    keyword("order") ~> keyword("by") ~> rep1sep(defined_expr ~ opt(keyword("asc") | keyword("desc")) ^^ {
+  def order_by: Parser[OrderBy[T[Sql]]] = {
+    def asOrdering[I]: PartialFunction[I ~ Option[String], (OrderType, I)] = {
```
djspiewak: Why is this a `PartialFunction`? Shouldn't it be total?

foundation/src/test/scala/quasar/ScalazSpecs2Instances.scala
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package quasar
+
+import org.scalacheck.Gen
+import scalaz.Monad
+import scalaz.Monoid
+
+trait ScalazSpecs2Instances extends org.specs2.scalacheck.GenInstances {
```
djspiewak: Doesn't this come via scalaz-specs2?

mongodb/src/test/resources/planner/plan_combination_of_two_distinct_sets.txt
```diff
@@ -3,7 +3,7 @@ Chain
 ├─ $SimpleMapF
 │  ├─ Map
 │  │  ╰─ JsCore(remove(_, "_id"))
-│  ╰─ Scope(Map())
+│  ╰─ Scope(ListMap())
```
djspiewak: We converted from `Map`?

project/Dependencies.scala
```diff
@@ -79,8 +80,8 @@ object Dependencies {
     // Removing this will not cause any compile time errors, but will cause a runtime error once
     // Quasar attempts to connect to an h2 database to use as a metastore
     "com.h2database"              % "h2"                        % "1.4.196",
-    ("org.tpolecat"               %% "doobie-specs2"             % doobieVersion % Test)
-      .exclude("org.specs2", "specs2-core_2.11") // conflicting version
+    ("org.tpolecat"               %% "doobie-specs2"            % doobieVersion % Test)
+      .exclude("org.specs2", "specs2-core_2.12") // conflicting version
```
djspiewak: Out of curiosity, does this still cause a conflict?  (if we remove the `exclude`)

project/Dependencies.scala
```diff
@@ -107,7 +108,7 @@ object Dependencies {
       "org.tpolecat" %% "doobie-h2"         % doobieVersion,
       "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.6" % Test,
       ("org.tpolecat" %% "doobie-specs2"     % doobieVersion % Test)
-        .exclude("org.specs2", "specs2-core_2.11") // conflicting version
+        .exclude("org.specs2", "specs2-core_2.12") // conflicting version
```
djspiewak: Ditto.

build.sbt
```diff
@@ -27,6 +27,14 @@ def exclusiveTasks(tasks: Scoped*) =
 
 lazy val buildSettings = commonBuildSettings ++ Seq(
   organization := "org.quasar-analytics",
+  scalaOrganization := "org.scala-lang",
```
djspiewak: We should file an issue against sbt-slamdata to provide better support for working with non-typelevel Scala.  (heads up, @drostron)

build.sbt
```diff
@@ -165,7 +170,7 @@ lazy val assemblySettings = Seq(
     cp filter { af =>
       val file = af.data
 
-      val excludeByName: Boolean = file.getName.matches("""scala-library-2\.11\.\d+\.jar""")
+      val excludeByName: Boolean = file.getName.matches("""scala-library-2\.12\.\d+\.jar""")
```
djspiewak: Do we still need this, since we're not on TLS?

build.sbt
```diff
@@ -27,6 +27,14 @@ def exclusiveTasks(tasks: Scoped*) =
 
 lazy val buildSettings = commonBuildSettings ++ Seq(
   organization := "org.quasar-analytics",
+  scalaOrganization := "org.scala-lang",
+  scalacOptions --= Seq(
+    "-Yliteral-types",
+    "-Xstrict-patmat-analysis",
+    "-Yinduction-heuristics",
```
djspiewak: Is there any reason to turn *off* `induction-heuristics`?  Have you tried with it on?

lukaszwawrzyk: It seems to be typelevel specific option. It gives `bad option: '-Yinduction-heuristics'` when enabled.

foundation/src/test/scala/quasar/ScalazSpecs2Instances.scala
lukaszwawrzyk: Sadly, not. As specs2 does not depend on scalaz anymore, the instances that were there before are gone. We decided not to add it to scalaz-specs2 as this is kind of custom solution. If we could simply create two implicit defs that convert one monad and one monoid to another without any conflicts or problems, we would probably put it into that library. However as explained in one comment there it causes comflicts when resolving implicits.

frontend/src/main/scala/quasar/sql/parser.scala
lukaszwawrzyk: It doesn't handle the case when we have Some(<something else than "asc" or "desc">), which is not possible to happen because of how parser is constructed, however having a total function causes a warning about not exhaustive pattern matching.

project/Dependencies.scala
lukaszwawrzyk: I forwarded this question to Travis :smile:

mongodb/src/test/resources/planner/plan_combination_of_two_distinct_sets.txt
lukaszwawrzyk: Nothing changed in the code regarding type of map, I guess `ListMap` was always there, but `println(scala.collection.immutable.ListMap())` prints `Map()` in scala 2.11 and `ListMap()` in scala 2.12 so I updated tests accordingly.

project/Dependencies.scala
lukaszwawrzyk: looks fine

build.sbt
lukaszwawrzyk: I guess not. I removed that.

romanowski:
@djspiewak is there anything else we need to do here or is there anyone else that needs to look at it?

build.sbt
djspiewak: Oh interesting!  Looks like it isn't in mainline 2.12.

djspiewak:
@romanowski Everything looks good to me!  I'm going to do some housekeeping on this end and merge.

romanowski:
@djspiewak Just checking if there is anything we can do. 

IMHO we should not delay merging this PR for long otherwise we will need to resolve potential conflicts further delaying merge.

djspiewak:
@romanowski I agree on not delaying the PR too long.  The conflict as it stands is trivial, but they eventually will not be.  We're working to get a final release of 5.0 out, and originally that was going to be today.  My intention was to merge thereafter, but with the revised timeline, I might merge before 5.0.  I won't let this sit for more than a week; regardless of almost anything else, it's landing this week.

djspiewak:
Ok so update on this… We can’t merge before 5.0, and that seems unlikely to happen this week for other reasons. I’m going to merge this PR on Monday afternoon barring any immediate emergency blockers in flight at that point, which is unlikely. If 5.0 is still outstanding, there will be a back port branch (as we’ve done in the past) to allow a freeze.

Sorry for the schenanigans. This would have been smoother if we could have done 5.0 today as planned, but as they say… mice and men.

romanowski:
@djspiewak Merging such breaking changes just before release is probably not the best idea so we can wait a bit longer.
Monday is fine :)

djspiewak:
@romanowski Did you just rebase?  As a broad policy, we do not rebase PRs except when they're explicitly marked as WIP, since you don't know who else might be basing work on it.  Once the PR is filed (in non-WIP status), we strongly prefer a merge strategy.  In theory, that should also make conflicts easier to address. :-)

romanowski:
Yeah, we've rebased the branch (habits from other projects). Do you have git/pr related policies somewhere? We don't want to violate any other rules in future :)

djspiewak:
@romanowski We really should write up our git/pr/etc etiquette. :-)  To the best of my knowledge, we haven't really documented that particularly well.  The tl;dr off the top of my head though (possibly not comprehensive, but probably):

- When in doubt, do whatever you would do for another OSS project
- Don't rebase public non-WIP work (i.e. branches you've pushed, and especially PRs)
  + Rebasing locally is encouraged!  Clean history is nice.  Not required though
- Basing on other people's public work is totally fine and encouraged (this is why we don't rebase publicly)
- Granular commits are great, even though they do sort of mess up git-bisect
- Don't mix whitespace changes with real changes.  In other words, it's ok to fix the formatting of lines you're touching, and it's ok to fix the formatting of unrelated lines, but don't fix unrelated lines in the same commits that are fixing related ones (it just makes PRs hard to review).  This is a broad rule of thumb for any OSS project, so it sort of goes without saying.  :-)

Anyway, that's basically all I can think of.  It's a pretty standard merge-driven OSS process.

djspiewak:
Christmas is coming early!  Merging now…

PR #3340, opened 2017-12-21
Misc precog cleanup

alissapajer:


precog/src/main/scala/quasar/precog/ToString.scala
```diff
@@ -1,22 +0,0 @@
...
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-package quasar.precog
-
-trait ToString {
```
djspiewak: What on earth…?

alissapajer:
@djspiewak appveyor succeeded and then threw an exception, so this is good to go unless you want to see that greener.

PR #3341, opened 2017-12-21
Allow 'null' values in MongoDB distinct results

wemrysi:
Fixes an issue where `null` values in the results of a `db.collection.distinct(...)` query would throw an exception.

### DIFF
--- a/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIOWorkflowExecutor.scala
+++ b/mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIOWorkflowExecutor.scala
@@ -31,7 +31,7 @@ import scala.Predef.classOf
 import com.mongodb._
 import com.mongodb.async.client._
 import com.mongodb.client.model.CountOptions
-import org.bson.{BsonDocument, BsonValue}
+import org.bson.{BsonDocument, BsonNull, BsonValue}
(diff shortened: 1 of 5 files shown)


mongodb/src/main/scala/quasar/physical/mongodb/MongoDbIOWorkflowExecutor.scala
```diff
@@ -66,8 +66,10 @@ private[mongodb] final class MongoDbIOWorkflowExecutor
     type DIT = DistinctIterable[BsonValue]
     type MIT = MongoIterable[BsonDocument]
 
+    // Fun fact: Even though BSON has a `null` type, the MongoDB java driver emits
+    // BSON nulls as Java `null`, because reasons.
     val wrapVal: BsonValue => BsonDocument =
-      new BsonDocument(field.asText, _)
+      bv => new BsonDocument(field.asText, Option(bv) getOrElse new BsonNull())
```
jsantos17: Wow, this is an obscure bug. Should we also open an issue against the Mongo Java driver?

wemrysi: We can try, this was apparently their fix to [a different bug](https://jira.mongodb.org/browse/JAVA-2386).

PR #3342, opened 2017-12-21
Remove authorities and permissioning from mimir.

alissapajer:
Removes everything in `quasar.precog.common.security` and `quasar.precog.common.accounts`.


djspiewak:
I think there's a ticket somewhere which referred to this removal.  We should probably find that at some point and update it.

alissapajer:
Yeah, that is one of the issues that didn't make it in the migration. Once the migration is complete, my plan was to go through all the `topic: Precog Debt` issues and close the ones that are done.

PR #3343, opened 2017-12-22
Integration test refactor part 3

rabbitonweb:
This is further work on integration tests refactor. Introducing:

* s08 - reduce
* s09 - more complex filters

Missing s07 and 06 are coming with different PR, being developed while working on postgress connector


jsantos17:
I'm not sure I'm following why some files are duplicated by this PR. Would you mind explaining?

Some items are already duplicate in `master`, like`inventory_items`. Any idea why we duplicated them like this?

```
~/R/s/quasar ❯❯❯ find it -iname inventory_items.data | xargs shasum
16e9e9567deb9112aaf93915a57944d666dcb125  it/src/main/resources/tests/inventory_items.data
16e9e9567deb9112aaf93915a57944d666dcb125  it/src/main/resources/tests/regression/inventory_items.data
```

alissapajer:
@rabbitonweb this is awaiting follow up from you.

rabbitonweb:
@alissapajer I'm pretty much aware of it (and I'm even working on it in between compile times) but postgres has higher priority on my list, thus work on this is going slower.

jsantos17:
@rabbitonweb I think we're still duplicating `olympics.data`. I see it has the exact same number of lines but a different checksum.

Checksums:

```
~/R/s/quasar ❯❯❯ find it -name olympics.data | xargs gsha256sum                                                                                                                                                                           ⏎
687bc5ce565ffdde325cbb5458c8af45374575f974c0da754b28490e671aab8c  it/src/main/resources/tests/demo/olympics.data
e7266ae6d6c52cbf073916041bf0d5ebbe6061e3743550f47c05a7b8efacd327  it/src/main/resources/tests/qa/olympics.data
e7266ae6d6c52cbf073916041bf0d5ebbe6061e3743550f47c05a7b8efacd327  it/src/main/resources/tests/olympics.dat
```

Lines:

```
~/R/s/quasar ❯❯❯ find it -name olympics.data | xargs wc -l
    2311 it/src/main/resources/tests/demo/olympics.data
    2311 it/src/main/resources/tests/qa/olympics.data
    2311 it/src/main/resources/tests/olympics.data
```

Are they meant to be different?

Also, red build is red.

rabbitonweb:
Those will be one day merged to one (once all tests are under `qa`)
```
    2311 it/src/main/resources/tests/qa/olympics.data
    2311 it/src/main/resources/tests/olympics.data
```
This   ```2311 it/src/main/resources/tests/demo/olympics.data``` might be different (not sure) as it was taken from demo environment and I want to keep it separate because it represents data from our demo environment

jsantos17:
Ah okay. I'll approve and merge this once the build is green @rabbitonweb.

alissapajer:
@rabbitonweb it's possible to reference data in an integration test that is in a different directory than the test. Is there a reason you can't do that for `qa/olympics.data` and `olympics.data`?

alissapajer:
Also, we should see what the differences are between the demo `olympics.data` and the one we have already checked in. If the differences are minor, then I'd prefer to just use the demo data for all the queries that require `olympics.data`. I think it's good to have just a single version of each dataset, when possible.

rabbitonweb:
@alissapajer my original idea was that there are three main groups `demo`, `qa` and `regression` - i just think they should have their own data (even if it sometimes duplicates itself).

rabbitonweb:
@jsantos17 there is one test failing (i dont know why yet), i will have a look tomorrow in the morning, it should not really fail as we only rearrange that - i need to look closer on what that is.

rabbitonweb:
@alissapajer @jsantos17 reagrind that data files: just to clarify, i think we should have three seperate groups:
`qa` running on its own data
`demo` running on data from workspaces, *EXACT* data as it is being used on demos
`regression` running with data that was used when bug was issued on youtrack

alissapajer:
@rabbitonweb I don't think we should be duplicating data files when we can avoid it, because that makes the tests harder to maintain. I agree we should use the exact demo data for all the demo queries. But if other integration tests can take advantage of that data, I think they should. What's the advantage of duplicating the files?

The categories `qa` and `regression` seem to represent the same underlying notion: a test that someone found to be problematic that we want to track. Why do we need to categorize them according to where the issue surfaced?

I would rather see the integration tests categorized in terms of what the tests are actually testing rather than who discovered the bug that led to the test being created. Maybe we can add some type flags to the `.test` json files that categorize what the test is testing, to make the tests easier to search (and dogfood into a slamdata dashboard). For example,
```
{ "type": ["join", "reduction", "grouping"] }
```
and
```
{ "type": ["select *", "ordering"] }
```

rabbitonweb:
@alissapajer unfortunately my github issue regarding those changes was never imported to youtrack . I will create tomorrow new issue in youtrack where I will describe what motivates those changes. Ok?

alissapajer:
@rabbitonweb is this the issue you were looking for? https://slamdata.myjetbrains.com/youtrack/issue/qz-3313

jsantos17:
@rabbitonweb the build failure looks legitimate too.

rabbitonweb:
Github notification woek me up :) Damn smartphones

alissapajer:
@rabbitonweb my remaining concern with this PR is that we are duplicating integration test data files. Regardless of the directory structure of the tests themselves, I think we should aim to not copy-paste the data files, even if tests in different paths are using the same data. Instead we can simply reference the path where the data is stored.

rabbitonweb:
TBD

```
test
|- qa
   |- s01 - selection
   |- s02 - function
   |- s03 - order
   |- basic.data
   |- sthn.data
   |- patients.data
|- demo
   |- patients.data
|- regression



test
|- qa
   |- s01 - selection
   |- s02 - function
   |- s03 - order
|- demo
|- regression
|- basic.data
|- sthn.data
|- patients.data


test
|- s01 - selection
|- s02 - function
|- s03 - order
|- demo
|- regression
|- basic.data
|- sthn.data
|- patients.data
|- patients.data
```

jsantos17:
@rabbitonweb would you mind explaining your comment? I still think we should get rid of the duplicates. Can we achieve what you want with symbolic links? Git will store the target of the symlink and restore it when you check out.

rabbitonweb:
Postgres has right now my top prior, thus I only added a comment to not forget some of my ideas.

I still believe  there might be more good then harm in haveing some data duplicated (as there is a different purpose for `qa`, `regression` and `demo` and there is no harm done in having data duplicated but actually there might be some good in it - because those folders will be **DECOUPLED** conceptually from each other.

However, since I was probably the only person that was caring about this refactor and having them cleaned up, I will adhere to your suggestions, because in future you guys will maintain it not me, so having pushing it my way makes no sense at this point.

rabbitonweb:
ok, somebody changed `largeZips.data` ? :)

rabbitonweb:
Ok, to have this work either completely reverted or merged. Here is my proposition.

Structure of `test` folder will be following:
```
test
|- s01 - selection
|- s02 - function
|- s03 - order
|- demo
|- regression
|- // here data files like basic.data, patients.data etc. 
```
Tests will be divided into three categories: 

* QA tests (folders `s0x`) - those tests hold simple building steps on which you can start implementing your connector. The idea here is that you will first implement `s01 - simple select` before moving to `s02 - functions` and so on. You grow your connector by gradually "close" one `s0x` after the other. Previously when working with new connector, integration tests were not helping, they were failing because some particular part of your connector was not yet implemented. It was really hard to track **progress** - the idea here is to **enable** that progress tracking.

 * Folder `regression`
This folder holds all tests that check that regression will never happen. You will add tests here most likely after discovering a bug, but before you fix it. Connector creators might run all tests in this folder after they've successfully passed all tests under `s0x`. 

* Folder `demo`

This folder will hold all tests that were "rebuild" from our demo environments. In theory once those tests will pass, given connector should have no issues being demoed to our clients ("In theory practice and theory are the same. In practice they are not")

What is above described could not possibly achieved in one PR (I mean it could but that would be overkill I think), that's why we have to make it in small steps. So while the tests are being moved to their categories (`s0x` folders) we will keep the `qa` folder (data will only be duplicated between `qa`, `demo` and `regressio`). And once all tests are moved to either `s0x` or `demo` or `regresson` we will flatten the structure by removing the `qa` and moving all `s0x` tests under `test` folder.

In other words we will eventually strive for this structure: 

```
test
|- s01 - selection
|- s02 - function
|- s03 - order
|- demo
|- regression
|- // here data files like basic.data, patients.data etc. 
```

but during the refactor process we will go with temporally with this structure:


```
test
|- qa
|    |- s01 - selection
|    |- s02 - function
|    |- s03 - order
|- demo
|- regression
|- // here data files like basic.data, patients.data etc. 
``` 

Does it make sense?

jsantos17:
Ping @rabbitonweb 

There are some conflicts in the data files now as well.

rabbitonweb:
I'm ooo

alissapajer:
@jsantos17 since I'm already doing some integration test work on next-major, I'm going to see what it takes to get this ready for merge. If/when I push commits directly to this branch, I'll ping you for review.

alissapajer:
@jsantos17 I'll also PR the removal of `PlannerSql2ExactSpec` (making sure all queries are in integration tests) to master, since that's a better place, rather than shoving even more into next-major :)

jsantos17:
Sounds good! Thanks @alissapajer!

rintcius:
@alissapajer @jsantos17 I don't think we can replace `PlannerSql2ExactSpec` with IT's (at least not yet).  They test more than what the ITs test (viz. stability of the plans).
As an example, I have added these tests because they fell over without anyone noticing and causing performance issues for clients: https://github.com/quasar-analytics/quasar/blob/master/mongodb/src/test/scala/quasar/physical/mongodb/PlannerSql2ExactSpec.scala#L229-L267

I'd like to get rid of `PlannerSql2ExactSpec` as it is tedious to maintain this, but I think we need to be able to track the plans used in ITs (see https://slamdata.myjetbrains.com/youtrack/issue/qz-2802) before we can get rid of it.

jsantos17:
@rintcius what do you think about replacing them with `it`s and performance tests? I see that approach as easier to maintain with similar (if not higher) value. The `toMetal` tests are very valuable, but now that I think about it, we should probably test them from QScript instead of SQL^2. With another test (probably outside the Mongo connector) to check the SQL^2 produces the expected QScript.

alissapajer:
Here's my plan, in order to keep all these things separate.
1) I'll push changes to this branch to get it mergeable. These changes won't touch `PlannerSql2ExactSpec`.
2) On next-major I'll update the expectations for the failing tests in `PlannerSql2ExactSpec`.
3) Any changes to `PlannerSql2ExactSpec` we will do separately against master after further discussion.

rintcius:
@jsantos17 I think there's a lot of value in tracking plans & this value is something that complements both performance and integration tests. It shouldn't take much to maintain these. 

Also I think there's more value in testing from SQL^2 than from QScript, because 
1) SQL^2 based tests are more stable than QScript-based tests (at the moment we have nothing in place to make sure that QScript-based tests are still relevant)
2) SQL^2 is more compact and easier to maintain than QScript

Maybe discuss some time in hangout?

alissapajer:
I'm going to close this since I'm not actively working on it. When we do some more integration test re-org we can base it on this branch if that still makes sense.

PR #3345, opened 2017-12-22
Delete moar

alissapajer:
Delete some more dead code in mimir and friends.


alissapajer:
appveyor passed and then reported thread errors. I wish I knew how to make it stop doing that.

djspiewak:
@alissapajer I wonder if something is actually wrong.  AppVeyor usually isn't unreliable; I suspect the build is actually just failing on windows.

alissapajer:
@djspiewak appveyor passed on #3346, which contains this PR. So if something is actually wrong, it's not consistently wrong.

alissapajer:
And according to the printout all the tests have passed, so whatever component is failing, it's not the tests themselves (or compilation).

PR #3348, opened 2017-12-22
Generate ExecutionIDs, record and dump basic timing information (fixes qz-3343)

edmundnoble:
First pass on benchmarking infrastructure.

1. All queries have an associated "execution identifier", a pair of an integer which increments on every query and a start `Instant` (timestamp).
2. All queries have a set of "timing scopes" like "full query execution pipeline" and "parse SQL" which are pairs of start and end timestamps with identifiers attached. Adding a timing scope is as simple as a call into a type class wrapping some part of the query. There need to be a *lot* more of these; we haven't even separated query planning from query execution yet.
3. There's a "timing repository" containing a map from execution identifiers to maps containing all of the execution's timing scopes. It's size-limited, and drops the oldest executions when it goes over the size specified (with age measured by start-time). The goal here is to keep something like 200 timing scopes in memory at all times so that we can dump the info and understand why queries are slow, even after the fact (especially for customers).
4. When an execution finishes, there's an option to print out its timing info immediately. This option isn't currently exposed to the web server (and I'm not sure if it should), but it is to the REPL; use `set printTiming = 1`.

Potential issues:
- It may be best to order executions by *end time* instead, which would require not eagerly adding an execution's info to the timing repository.
- The integer part of the execution identifier may end up being redundant, the intention was to use it to keep them distinct and make it easier to index them mentally.
- The timing repository could use some tests to ensure it behaves as specified. Manual testing seemed to show it was fine, though.
- Code that populates timing scopes might need to be tested to ensure that it's doing it properly.

Some sample output from the REPL:
```
(v27.6.6) 💪 $ select substring(city, 0, 1), count(*) from zips
Query time: 0.3s
 0  | 1      |
----|--------|
 A  |  29353 |
 C  |  29353 |
 B  |  29353 |
 B  |  29353 |
 B  |  29353 |
 B  |  29353 |
 C  |  29353 |
 C  |  29353 |
 C  |  29353 |
 C  |  29353 |
Query execution 5 began at 2017-12-22T20:55:24.337Z
phase parse SQL: 6 ms
phase resolve imports: 1 ms
phase execute query: 363 ms
phase total query pipeline (REPL): 370 ms
(v27.6.6) 💪 $
```
#qz-3343 done

### DIFF
--- a/effect/src/main/scala/quasar/effect/TimingScope.scala
+++ b/effect/src/main/scala/quasar/effect/TimingScope.scala
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.effect
+
+import slamdata.Predef._
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import scala.collection.immutable.TreeMap
+import quasar.fp._
+import quasar.fp.numeric.Natural
+
+import scalaz._, Scalaz._
+import scalaz.concurrent.Task
+
+/** Represents the ability to create an execution scope, within which timing scopes can be created.
+  */
+trait ScopeExecution[F[_], T] {
+  def newExecution[A](executionId: ExecutionId, action: ScopeTiming[F, T] => F[A]): F[A]
+}
+
+/** Represents the ability to create a timing scope.
+  */
+trait ScopeTiming[F[_], T] {
+  def newScope[A](scopeId: String, fa: => F[A]): F[A]
+}
+
+final case class ExecutionId(index: Long, startTime: Instant)
+object ExecutionId {
+  import scala.Ordering
+  implicit val executionIdOrdering: Ordering[ExecutionId] = Ordering.ordered[Instant](x => x).on(_.startTime)
+  implicit val executionIdShow: Show[ExecutionId] = Show.shows {
+    case ExecutionId(index, start) => s"Query execution $index began at $start"
+  }
+  def ofRef(ref: TaskRef[Long]): Task[ExecutionId] =
+    for {
+      i <- ref.modify(_ + 1)
+      now <- Task.delay(Instant.now())
+    } yield ExecutionId(i, now)
+}
+
+final case class ExecutionTimings(timings: Map[String, (Instant, Instant)])
+object ExecutionTimings {
+  implicit val executionTimingsShow: Show[ExecutionTimings] = Show.show {
+    executionTimings =>
+      executionTimings.timings.map {
+        case (identifier, (start, end)) =>
+          val millisBetween = start.until(end, ChronoUnit.MILLIS)
+          Cord("phase ") ++
+          Cord(identifier) ++
+          Cord(": ") ++
+          millisBetween.show ++
+          Cord(" ms")
+      }.mkString("\n")
+  }
+}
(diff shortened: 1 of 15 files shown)


effect/src/main/scala/quasar/effect/TimingScope.scala
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+}
+
+object ScopeTiming {
+  def forTask[T](executionId: ExecutionId, repo: TimingRepository): ScopeTiming[Task, T] =
+    new ScopeTiming[Task, T] {
+      def newScope[A](scopeId: String, f: => Task[A]): Task[A] = {
+        for {
+          start <- Task.delay(Instant.now())
```
wemrysi: You might be interested in `quasar.effect.Timing` here.

Also, it is more accurate to use the difference between successive `System.nanoTime` calls to measure elapsed time rather than the difference in wall-clock time.

djspiewak:
FYI, "fixes qz-3343" in the title will not trigger the automation.  :-)

jdegoes:
Should be `#qz-3343 done`

effect/src/main/scala/quasar/effect/TimingScope.scala
edmundnoble: @wemrysi I considered it, the only issue is that `Timing` isn't in a lot of the algebras here and I don't think it's likely to be worth mocking; but when I write tests, I suspect I'll learn to what extent I'm actually right on that.

With regard to `nanoTime`, I think I'm actually more interested in having accurate start and end times than having millisecond-precise deltas. There seems to be enough variation that `nanoTime` doesn't seem to give us any extra meaningful information, and knowing when a phase really started or stopped seems to be more valuable.

edmundnoble:
Thanks @jdegoes, @djspiewak; I wasn't sure if it provided that integration at all.

edmundnoble:
Whoops, tests don't compile.

edmundnoble:
This latest commit makes for the last work on this PR excepting review. Next thing I'll be working on is the infrastructure itself, and perhaps next breaking query planning down into phases; I'm doing that after because with the current state of the PR, I haven't been able to generate queries that are even slightly bottlenecked by query planning.

djspiewak:
> I'm doing that after because with the current state of the PR, I haven't been able to generate queries that are even slightly bottlenecked by query planning.

This doesn't surprise me at all.  QSU ftw.  :-D  I would imagine that for most queries, the bottleneck is primarily in the underlying datastore, but I also suspect that there is a massive amount of load that we generate that we don't need to be.

repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -233,24 +240,35 @@ object Repl {
...
+              block <- ST.newScope("resolve imports", DF.unattemptT(resolveImports(expr, state.cwd).leftMap(_.message)))
+              vars  = Variables.fromMap(state.variables)
+              _     <- n.cata(name =>
+                         {
+                           val out = state.cwd </> file(name)
+                           for {
+                             query <- ST.newScope("plan query",
+                               fsQ.executeQuery(block, Variables.fromMap(state.variables), state.cwd, out).pure[Free[S, ?]])
+                             results <- ST.newScope("evaluate query",
+                              runQuery(state, query)(κ(P.println("Wrote file: " + posixCodec.printPath(out)))))
```
djspiewak: We should add some of this scoping to `BackendModule` at some point, so we can get more granular information on compilation vs rewrites vs coalescence.

repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -233,24 +240,35 @@ object Repl {
...
+                           } yield results
+                         },
+                         {
+                           for {
+                             query <- ST.newScope("plan query",
```
djspiewak: On that note, it's worth mentioning that this isn't really where the query is being planned.  That would be `LPtoQS`.

repl/src/main/scala/quasar/repl/Repl.scala
edmundnoble: Absolutely will, I was actually going to add it to `QueryFile` because I want to also get an idea of constant query planning performance.

repl/src/main/scala/quasar/repl/Repl.scala
edmundnoble: The query is being planned by this call, but I agree that it makes more sense to push it all down to `LPtoQS` and all of the rewrites in the planner, which will be in a subsequent PR.

PR #3349, opened 2017-12-27
[WIP] RDBMS planner union

mikolak-net:
Adds support for UNION to RDBMS/Postgres. Depends on #3344 .


it/src/main/resources/tests/qa/s06_union/columns_self_union.test
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "[qa_s06] select columns self union",
+
+  "backends": {
+    "mimir": "ignoreResultOrder",
```
rabbitonweb: Please add `ORDER BY` clause too all tests so that results are sorted and asserted as `exactly`

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -131,9 +130,22 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
       }
     }
 
+    case qUnion@qscript.Union(src, left, right) =>
+      val hole = qscript.HoleQS[T]
+      (src.project, left, right) match {
+        case (_: Select[_], `hole`, `hole`) => src.point[F]
```
rabbitonweb: this is "optimalization" that postgres planner should handle, but we can leave it as it is for now (as long as reduce is not implementd)

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -131,9 +130,22 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
...
+        case _ =>
+          (compile(left, src) |@| compile(right, src))(Union(_,_).embed)
+      }
+
+    case reduce@qscript.Reduce(reduceSrc, _, _, _) =>
+      reduceSrc.project match {
+        case _: Union[_] | _: Select[_] => reduceSrc.point[F]
```
rabbitonweb: In future (when reduce is implemented) we should remove this and provide recognition of "distinct-pattern", meaning 

```
│     ├─ Bucket
│     │  ╰─ List
│     │     ╰─ ○
│     ├─ Reducers
│     │  ╰─ Arbitrary
│     │     ╰─ ○
│     ╰─ Repair
│        ╰─ ReduceIndex(\/-(0))
```

PR #3354, opened 2017-12-28
RDBMS: Optimize Search for const conditions

kciesielski:
All `like` expressions were translated to `when (true) ... else .... end`. To avoid `when (true)` and `when (false)`, such cases are now handled independently in the planner proper.

This optimization is not intended to increase performance, just readability of initial DB queries when debugging.


edmundnoble:
Won't the DB's planner optimize this out?

kciesielski:
@edmundnoble I believe so. However, this makes queries shorter and easier to read/debug, while it doesn't seem to increase complexity too much.

PR #3357, opened 2017-12-29
Mongo count

rintcius:
* make simple count queries use `count()` again
* move 2 ITs so that they get run for both streaming and resultFile ITs

Fixes sd-2687


mongodb/src/main/scala/quasar/physical/mongodb/bson.scala
```diff
@@ -118,6 +118,10 @@ object Bson {
   object Binary {
     def fromArray(array: Array[Byte]): Binary = Binary(ImmutableArray.fromArray(array))
   }
+
+  val _listMap =
```
jsantos17: Can we rename this prism to `_doc`? To keep it consistent with the other optics defined in this file.

mongodb/src/main/scala/quasar/physical/mongodb/execution.scala
```diff
@@ -89,6 +89,31 @@ private[mongodb] object execution {
     }
   }
 
+  object SingleListMap {
+    def unapply[A, B](lm: ListMap[A, B]): Option[(A, B)] =
+      if (lm.size <= 1) lm.headOption else None
+  }
+
+  object SimpleRedirect {
```
jsantos17: Can we rename this to `SimpleRename` or `KeyRename`? We could also add some comments explaining what these extractors are doing.

rintcius:
@jsantos17 updated

PR #3360, opened 2017-12-29
Extract selectors from LeftShift repair/struct

jsantos17:
qz-3334 done


jsantos17:
I believe I can generalize this to every `FreeMap`, but I believe we should merge it and explore generalization in a different issue. Then plan for the query in qz-3334 in this branch:

```
select codes[*] as codes, first_name from patients where state = "CO"
```

plans as

```
db.patients.aggregate(
  [
    { "$match": { "state": "CO" } },
    {
      "$project": {
        "s": "$$ROOT",
        "f": {
          "$cond": [
            {
              "$and": [
                { "$lte": [{ "$literal": [] }, "$codes"] },
                { "$lt": ["$codes", { "$literal": BinData(0, "") }] }]
            },
            "$codes",
            [{ "$literal": undefined }]]
        },
        "_id": false
      }
    },
    { "$unwind": "$f" },
    {
      "$project": { "codes": "$f", "first_name": "$s.first_name", "_id": false }
    }],
  { "allowDiskUse": true });
```

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -902,30 +908,66 @@ object MongoDbPlanner {
               case _ => none
             }
 
-            if (repair.contains(LeftSideF))
+            def elideCond[A]: CoMapFuncR[T, A] => Option[CoMapFuncR[T, A]] = {
+              case CoEnv(\/-(MFC(Cond(if_, then_, Embed(CoEnv(\/-(MFC(Undefined())))))))) =>
+                CoEnv(then_.resume.swap).some
```
rintcius: I don't understand why this is a valid rewrite. Can you explain?

rintcius:
Oh, since it fixes qz-3334 can you also add the query mentioned in there as IT?

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
jsantos17: The extracted matchers implement the `Cond` so we no longer need it in the QScript. I think we can improve this by removing whatever the extracted `InputFinder` found instead of indiscriminately removing all `Cond`s

jsantos17:
@rintcius great idea. I'll add the test.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
rintcius: Ah, ok. Yes would be great if it can be related to what the extractor found

jsantos17:
@rintcius what do you think about fixing the `Cond` rewrite later? I'll open an issue to generalize this to all `FreeMap`s instead of limiting it to `LeftShift` with a note to avoid the `Cond` eliding.

rintcius:
@jsantos17 that's fine. I saw you added a FIXME already so 👍

jsantos17:
@rintcius build failed because Marklogic's `Cond` implementation is incomplete. Just marked it as pending.

jsantos17:
@rintcius just added it! thanks for reminding me.

PR #3362, opened 2018-01-02
RDBMS join planner

kciesielski:


rabbitonweb:
@djspiewak this is being addressed in the uber test refactor that is currently in the making, see https://github.com/quasar-analytics/quasar/pull/3343

rabbitonweb:
@djspiewak eventually all data will fall into either `qa` or `tests` depending on how final discussion on #3343 will resolve, but for this test we assume its `qa` (if it will be `tests` it will be fixed in #3343)

rabbitonweb:
@djspiewak pls comment as we are dealing with an major issue with postgres (refer to my standups from yesterday) and we might have a working solution but we need current work (joins and reduce) in the master in order to apply the fixes as a separate PR (to see that it really is fixing all the issues as well as have more readable PR)

djspiewak:
👍

rabbitonweb:
@djspiewak this is ready to go :)

PR #3366, opened 2018-01-03
PostgreSQL Reduce planner

rabbitonweb:


rabbitonweb:
We are still missing few reducing functions, but those will be addressed in separate PR

case RF.First(a1)          => ???
case RF.Last(a1)           => ???
case RF.UnshiftArray(a1)   => ???
case RF.UnshiftMap(a1, a2) => ???

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -56,9 +59,10 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
 
   val unref: T[SqlExpr] = SqlExpr.Unreferenced[T[SqlExpr]]().embed
 
+  @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
   def plan: AlgebraM[F, QScriptCore[T, ?], T[SqlExpr]] = {
     case qscript.Map(`unref`, f) =>
-      processFreeMap(f, SqlExpr.Null[T[SqlExpr]])
+     processFreeMap(f, SqlExpr.Null[T[SqlExpr]])
```
wemrysi: Indentation.

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -125,28 +131,102 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
...
       notImplemented(s"QScriptCore: $other", this)
+  }
+
+  import quasar.qscript.{MapFuncCore => MFC}
+  import quasar.qscript.{MapFuncDerived => MFD}
+
+  object DisctinctPattern {
+
+    object BucketWithSingleHole {
+      def unapply(fms: List[FreeMap[T]]): Boolean =
```
wemrysi: Need to reindent this entire function.

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -125,28 +133,84 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
...
+    ).embed
+
+    case reduce@qscript.Reduce(src, bucket, reducers, repair) => for {
+      alias <- genId[T[SqlExpr], F]
+      gbs   <- bucket.traverse(processFreeMap(_, alias))
+      rds   <- reducers.traverse(_.traverse(processFreeMap(_, alias)) >>=
+        reduceFuncPlanner[T, F].plan)
+      rep <- repair.cataM(interpretM(
+        _.idx.fold(
+          idx => notImplemented("Reduce repair with left index, waiting for a test case", this): F[T[SqlExpr]],
```
djspiewak: This is worth writing some manual qscript to test.  We're absolutely going to be using this functionality in qscript in the not-too-distant future.

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
```diff
@@ -125,28 +133,84 @@ F[_]: Monad: NameGenerator: PlannerErrorME](
...
+    object ReducersWithSingleArbitraryHole {
+      def unapply(fms: List[ReduceFunc[FreeMap[T]]]): Boolean =
+        fms.headOption.exists {
+          case Arbitrary(fm) => isHoleOrGuardedHole(fm)
+          case _ => false
+        }
+    }
+
+    def unapply[A](qs: QScriptCore[T, A]): Option[qscript.Reduce[T, A]] = qs match {
+      case reduce@qscript.Reduce(_, BucketWithSingleHole(), ReducersWithSingleArbitraryHole(), _) => reduce.some
```
djspiewak: This needs to be more general.  If you `select distinct(foo) from bar`, I believe you'll get a more elaborate pattern that needs to be recognized.

rdbms/src/main/scala/quasar/physical/rdbms/planner/ReduceFuncPlanner.scala
```diff
@@ -0,0 +1,47 @@
...
+import scalaz._
+import Scalaz._
+
+final class ReduceFuncPlanner[T[_[_]]: BirecursiveT, F[_]: Applicative] extends Planner[T, F, ReduceFunc] {
+
+  def plan: AlgebraM[F, ReduceFunc, T[SqlExpr]] = planʹ >>> (_.embed.η[F])
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  val planʹ: Transform[T[SqlExpr], ReduceFunc, SqlExpr] = {
+    case RF.Arbitrary(a1)      => a1.project
```
djspiewak: So the constraint of `Arbitrary` is that it needs to return a single value (per bucket), just like the other reductions.  If I'm reading this right, you aren't returning a single value, because you're just applying whatever freemap is within the function.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/planner/PostgresRenderQuery.scala
```diff
@@ -170,13 +182,15 @@ object PostgresRenderQuery extends RenderQuery {
...
+      val orderByStr = if (order.nonEmpty) s" order by $orderStr" else ""
+
+      val groupByStr = ~(groupBy.flatMap{
+        case GroupBy(Nil) => none
+        case GroupBy(v) => v.map(_._2).intercalate(", ").some
+      }.map(v => s" GROUP BY $v"))
 
       val fromExpr = s" from ${from.v._2} ${from.alias.v}"
-      s"(select ${selection.v._2}$fromExpr$filter$orderByStr)".right
+      s"(select ${selection.v._2}$fromExpr$filter$groupByStr$orderByStr)".right
```
djspiewak: This would be a lot more readable with spaces between the variables, and it shouldn't matter to the semantics of the generated SQL.

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/planner/PostgresRenderQuery.scala
```diff
@@ -170,13 +182,15 @@ object PostgresRenderQuery extends RenderQuery {
         s"${o.v._2} $dirStr"
       }.mkString(", ")
 
-      val orderByStr = if (order.nonEmpty)
-        s" order by $orderStr"
-      else
-        ""
+      val orderByStr = if (order.nonEmpty) s" order by $orderStr" else ""
+
+      val groupByStr = ~(groupBy.flatMap{
```
djspiewak: What does prefix `~` mean?

wemrysi: @djspiewak [`getOrElse(mzero)`](https://github.com/scalaz/scalaz/blob/23b35ffb1a44ff8a157952b6aa1d02eff49926e3/core/src/main/scala/scalaz/syntax/std/OptionOps.scala#L69).

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
rabbitonweb: Ok, but this PR is not closing all the work on `Reduce` (we are still missing few reducing functions). It just enables majority of QA test to pass. 
I will include that in comment of "what else need to be implemented for `Reduce`"

it/src/main/resources/tests/qa/s08_reduce/distinct_asterix_simple.test
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "[qa_s08] distinct * simple",
+
+    "backends": {
+       "couchbase":         "ignoreFieldOrder",
+       "marklogic_json":    "ignoreFieldOrder",
+       "mimir":             "ignoreFieldOrder"
+    },
+    "data": "../basic2.data",
+    "query": "select distinct(*) from `../basic2`",
```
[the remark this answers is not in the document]
kciesielski: @wemrysi Right, I added a commit with a duplicate to this data set.

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
kciesielski: Fixed

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
kciesielski: Refactored and reindended.

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
rabbitonweb: `select distinct(foo) from bar` doesn't match this pattern, but it results in something like `SELECT foo from BAR group by foo` which effectively means the same thing.

rdbms/src/main/scala/quasar/physical/rdbms/planner/ReduceFuncPlanner.scala
rabbitonweb: `Arbitrary` is correctly implemented (we have tests that ... I wanna say "prove".. lets say show ... that it works).

`Arbitrary` is used as `reducing` function so it always ends up here : `select <<HERE>> from .... group by ....` (it is only used when reducing and is places in between `select` and `from` - which essentially means that it will generate SQL that gives that single value in return.

Example: 

`select c  from foo group by c` will create qscript where `redcures` have single element and that is `Arbitrary(ProjectKey)` so we return `ProjectKey` which is then rended as `c`

rdbms/src/main/scala/quasar/physical/rdbms/planner/QScriptCorePlanner.scala
rabbitonweb: After talking to @wemrysi I understand the general patter, but still would prefer to do this as optimalization maybe (after all integration tests are passing).

rdbms/src/main/scala/quasar/physical/rdbms/fs/postgres/planner/PostgresRenderQuery.scala
djspiewak: Intuitive.

PR #3368, opened 2018-01-04
Update VCache date handling.

alissapajer:
#qz-3358 done
#qz-3359 done

(Tickets will be closed manually since we are merging to next-major.)


alissapajer:
Only unit test failures in travis are in `mongodb`, which is as expected. Those will be addressed separately.

PR #3372, opened 2018-01-04
Add Mimir VFS VERSION

drostron:
#qz-2464 Done

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
@@ -18,6 +18,8 @@ package quasar.yggdrasil.vfs
 
 import quasar.contrib.pathy.{unsafeSandboxAbs, ADir, AFile, APath, RPath}
 import quasar.contrib.scalaz.stateT, stateT._
+import quasar.fp.ski.ι
+import quasar.fp.free._
 import quasar.fs.MoveSemantics
 
 import argonaut.{Argonaut, Parse}
@@ -27,10 +29,11 @@ import fs2.async, async.mutable
 import fs2.interop.scalaz._
 import fs2.util.Async
 
-import pathy.Path
+import pathy.Path, Path.file
 
-import scalaz.{~>, :<:, Coproduct, Free, Monad, NaturalTransformation, StateT}
+import scalaz.{~>, :<:, Coproduct, Equal, Free, Monad, NaturalTransformation, StateT}
 import scalaz.concurrent.Task
+import scalaz.std.anyVal._
 import scalaz.std.list._
 import scalaz.std.map._
 import scalaz.std.string._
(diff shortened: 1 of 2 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -64,8 +70,43 @@ object FreeVFS {
...
+
+  object VFSVersion {
+    implicit val equal: Equal[VFSVersion] = Equal.equalBy(_.v)
+  }
+
+  def handleVersion[S[_]](path: AFile)(implicit S0: POSIXOp :<: S, S1: Task :<: S): Free[S, Unit] = {
+    val writeVersion: Free[S, Unit] =
+      for {
+        verSink <- POSIX.openW[S](path)
+        verWriter = Stream.emit(ByteVector.fromInt(currentVFSVersion.v)).to(verSink).run
```
djspiewak: Isn't this writing a signed 32bit integer?  Also I have no idea what the endianness of this function is.  Seems like it would be better to have an actual `Codec[VFSVersion]` that gives us more direct control over the encoding.

drostron: `scodec-bits` defines `fromInt` as `def fromInt(i: Int, size: Int = 4, ordering: ByteOrdering = ByteOrdering.BigEndian): ByteVector = …` with the corresponding `toInt` in agreeance on size and endianness.

@djspiewak with that in mind does this still leave you with concern?

djspiewak: I still don't like that it's signed, but at least we have slightly more control.

drostron: I'm open to types other than `Int`. In discussion with @alissapajer `Int` was decided upon for simplicity. Alternatives could be refined's `Int Refined Positive`, shapeless's `Sized[List[Byte], _4]`, ….

drostron: And if we head that direction, I like the idea of defining a `Codec` instance.

alissapajer: Another option is
```scala
sealed trait VFSVersion
final case object version1 extends VFSVersion
```
and then have `Codec[VFSVersion]`. That would give us complete control over the codec, and a type for each version.

djspiewak: I would definitely prefer @alissapajer's suggested implementation. Having a nominal type for each version is extremely powerful.  Control over the codec would give us the ability to get rid of the signed bit, which we cannot avoid with `ByteVector.fromInt`.

drostron: That works. The world in which we have that many versions is interesting 😉.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -64,8 +70,43 @@ object FreeVFS {
...
+      for {
+        verStream <- POSIX.openR[S](path)
+        verBV <- verStream.runFoldMap(ι).mapSuspension(injectNT[POSIXOp, S] :+: injectNT[Task, S])
+        ver <- lift(Task.delay(VFSVersion(verBV.toInt())).handleWith { case e =>
+          Task.fail(
+            new RuntimeException("Unexpected VFS VERSION format", e)
+          )}).into[S]
+        _ <- (ver ≠ currentVFSVersion).whenM(
+          lift(Task.fail(new RuntimeException(
+            s"Unexpected VFS VERSION. Found ${ver.v}, current is ${currentVFSVersion.v}"
```
djspiewak: This doesn't matter for now, but we shouldn't ever be failing on any versions.

drostron: What is the desired behavior for this situation?

Issue #3265, opened 2017-12-08
PostgreSQL connector


EncodePanda:
- [x] ReadFile
- [x] WriteFile
- [x] ManageFile
- [x] QueryFile
- [ ] Planner
- [ ] integration tests
- [ ] demoable environment

kciesielski:
How about splitting the Planner point? Like:
[] Shifted Read
[] Core Planner (Map)
[] MapFuncCore  (Basic)
[] Sort 
[] Filter
[] Join
[] Reduce
[] MapFuncCore (remaining)
I'm not sure what else :)

PR #3376, opened 2018-01-05
RDBMS ProjectIndex

mikolak-net:
Part of #3265 .


mikolak-net:
@alissapajer : thanks, was in a hurry and somehow missed that :).

PR #3382, opened 2018-01-08
Repl instrumentation

edmundnoble:
Instrumenting the REPL to have more capabilities required for benchmarking; warmup and multiple-iteration runs as well as query naming in JSON output.

This also has the effect of drastically reducing the amount of contention over the repository map during simultaneous multiple query execution.

Depends on #3356.

### DIFF
--- a/repl/src/main/scala/quasar/repl/OutputFormat.scala
+++ b/repl/src/main/scala/quasar/repl/OutputFormat.scala
@@ -26,14 +26,12 @@ object OutputFormat {
   case object Precise extends OutputFormat
   case object Readable extends OutputFormat
   case object Csv extends OutputFormat
-  case object Nothing extends OutputFormat
 
   def fromString(str: String): Option[OutputFormat] = str.toLowerCase match {
     case "table"    => Table.some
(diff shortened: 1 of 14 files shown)


repl/src/main/scala/quasar/repl/OutputFormat.scala
```diff
@@ -26,12 +26,14 @@ object OutputFormat {
   case object Precise extends OutputFormat
   case object Readable extends OutputFormat
   case object Csv extends OutputFormat
+  case object Nothing extends OutputFormat
```
djspiewak: I've had some baaaaaad experiences reusing names for which scalac has special semantics…

edmundnoble: Lmao point taken.

edmundnoble:
Contention is drastically reduced because it used to be that adding a new timing scope would touch a single, centralized `TaskRef` which was used by other executions. Now, we have a `TaskRef` per execution and when they finish their results get added to the central `TaskRef`.

PR #3385, opened 2018-01-09
Evaluate and update join tests

jsantos17:
#qz-3351 done.


djspiewak:
@jsantos17 Needs assignee and version label.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -220,17 +212,15 @@ class PlannerSpec extends
...
         beRight.which(cwf => notBrokenWithOps(cwf.op, IList(ReadOp, SimpleMapOp, ProjectOp)))
     }
 
-    trackPending(
-      "object flatten",
-      plan(sqlE"select geo{*} from usa_factbook"),
-      // actual [ReadOp,ProjectOp,SimpleMapOp,ProjectOp]
-      IList(ReadOp, SimpleMapOp, ProjectOp))
-
+    "plan object flatten" in {
```
rintcius: just checking, this extra project is not something we should be able to optimize away?

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -352,57 +342,51 @@ class PlannerSpec extends
...
-      "simple left equi-join ($lookup)",
+    "plan simple left equi-join ($lookup)" in {
       plan3_4(
         sqlE"select foo.name, bar.address from foo left join bar on foo.id = bar.foo_id",
         defaultStats,
         indexes(collection("db", "bar") -> BsonField.Name("foo_id")),
-        emptyDoc),
-      // TODO: left/right joins in $lookup
-      // $unwind(DocField(JoinHandler.RightName)),  // FIXME: need to preserve docs with no match
-      // actual [ReadOp,MatchOp,ProjectOp,LookupOp,ProjectOp,UnwindOp,ProjectOp]
```
rintcius: I don't think left/right equi-joins are ok yet (since those FIXME's still hold)

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -352,57 +342,51 @@ class PlannerSpec extends
...
-    trackPending(
-      "simple right equi-join ($lookup)",
+    "plan simple right equi-join ($lookup)" in {
       plan3_4(
         sqlE"select foo.name, bar.address from foo right join bar on foo.id = bar.foo_id",
         defaultStats,
-        indexes(collection("db", "bar") -> BsonField.Name("foo_id")),
-        emptyDoc),
-      // TODO: left/right joins in $lookup
-      // $unwind(DocField(JoinHandler.RightName)),  // FIXME: need to preserve docs with no match
```
rintcius: same here

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
jsantos17: Now that I think more about it, we could inline the `$project` inside the `$simpleMap` as a `Submap`. I'll try to implement that simplification.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
jsantos17: I checked again and it looks like left and right joins in `$lookup` are already implemented. It was never implemented for old Mongo, it is for new Mongo.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
jsantos17: I reverted this change since we can definitely improve this plan. Left a `FIXME`.

PR #3386, opened 2018-01-09
Simplify static projections on `ProjectKey(Cond(...))`

jsantos17:
This substantially reduces the number of nodes in `repair` after `LeftShift`s are coalesced.


connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -380,6 +380,11 @@ object MapFuncCore {
         ExtractFunc(Constant(key))) =>
         map.reverse.find(_._1 ≟ key) ∘ (_._2.project)
 
+      case ProjectKey(
+        Embed(CoEnv(\/-(MFC(Cond(cond, Embed(StaticMap(map)), u @ Embed(CoEnv(\/-(MFC(Undefined()))))))))),
+        ExtractFunc(Constant(key))) =>
+        map.reverse.find(_._1 ≟ key) ∘ (_._2) >>= (v => rollMF[T, A](MFC(Cond(cond, v, u))).some)
```
djspiewak: In theory this should apply to the alternate as well as the consequent.  It also doesn't need to be specific to `Undefined()`.

connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -380,6 +380,16 @@ object MapFuncCore {
...
 
+      case ProjectKey(
+        Embed(CoEnv(\/-(MFC(Cond(cond, Embed(StaticMap(map)), alt))))),
+        ExtractFunc(Constant(key))) =>
+        map.reverse.find(_._1 ≟ key) ∘ (_._2) >>= (v => rollMF[T, A](MFC(Cond(cond, v, alt))).some)
+
+      case ProjectKey(
+        Embed(CoEnv(\/-(MFC(Cond(cond, cons, Embed(StaticMap(map))))))),
+        ExtractFunc(Constant(key))) =>
+        map.reverse.find(_._1 ≟ key) ∘ (_._2) >>= (v => rollMF[T, A](MFC(Cond(cond, cons, v))).some)
```
djspiewak: Needs a third case here for where *both* the consequent and the alternate are static maps.

wemrysi:
@jsantos17 @djspiewak I have a more general version of this in progress as we can always pull `Cond` where one branch is `Undefined` out of an expression, like we do with `Guards` this will allow the existing normalizations to apply without the special-casing here.

connector/src/main/scala/quasar/qscript/MapFuncCore.scala
rintcius: I think you should be able to `map (v => ...)` instead of `>>= (v => ... .some)` (2 times)

djspiewak:
@wemrysi Do you want us to just wait for your more general version?  Or do you want to just undo this in your general version PR?

wemrysi:
@djspiewak I think adding the 3rd case you mentioned (where both branches are static maps) and then merging is good, I'll remove the other two cases usurped by the general case.

connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -380,6 +380,23 @@ object MapFuncCore {
...
+        ExtractFunc(Constant(key))) =>
+        map.reverse.find(_._1 ≟ key) ∘ (_._2) ∘ (v => rollMF[T, A](MFC(Cond(cond, v, alt))))
+
+      case ProjectKey(
+        Embed(CoEnv(\/-(MFC(Cond(cond, cons, Embed(StaticMap(map))))))),
+        ExtractFunc(Constant(key))) =>
+        map.reverse.find(_._1 ≟ key) ∘ (_._2) ∘ (v => rollMF[T, A](MFC(Cond(cond, cons, v))))
+
+      case ProjectKey(
+        Embed(CoEnv(\/-(MFC(Cond(cond, Embed(StaticMap(consMap)), Embed(StaticMap(altMap))))))),
```
djspiewak: This case has to be above the other two, otherwise it will never be hit.

jsantos17: Oh snap! Just pushed a commit fixing it.

PR #3388, opened 2018-01-10
Web server now shuts down properly after pressing Enter

jedesah:


alissapajer:
In order to use the auto transitions, you need `#qz-3378 Done` in the PR body.

djspiewak:
> In order to use the auto transitions, you need #qz-3378 Done in the PR body.

Or in the commit.

PR #3389, opened 2018-01-10
Add func.StaticMap and clean up DSL usage

edmundnoble:
Adds `StaticMap` (and `StaticMapF`) to make maps with statically-known keys, a suggestion from Daniel, which is also offered by the old DSL. Also fixed some other DSL usage.


connector/src/main/scala/quasar/qscript/construction.scala
```diff
@@ -137,6 +137,13 @@ object construction {
       rollCore(MapFuncsCore.MakeMap(key, src))
     def MakeMapS[A](key: String, src: FreeMapA[T, A]): FreeMapA[T, A] =
       MakeMap(Constant(json.str(key)), src)
+    def StaticMap[A](pairs: (String, FreeMapA[T, A])*): FreeMapA[T, A] =
+      IList(pairs: _*).toNel.fold(Constant[A](json.map(Nil))) {
+        case NonEmptyList((hk, hv), t) =>
+          t.reverse.foldRight(MakeMapS(hk, hv)) { case ((k, v), b) => ConcatMaps(b, MakeMapS(k, v)) }
```
djspiewak: Why not just `foldLeft`?

connector/src/main/scala/quasar/qscript/construction.scala
```diff
@@ -137,6 +137,13 @@ object construction {
...
+      }
+    def StaticMapF[A, K](keys: K*)(f: K => FreeMapA[T, A], fk: K => String): FreeMapA[T, A] =
```
djspiewak: Shouldn't this be something more like `StaticMapGen`?  This would allow us to have a generalized variant which traverses: `def StaticMapGenM[M[_]: Monad, A, K](keys: K*)(f: K => M[FreeMapA[T, A]], fk: K => M[String]): M[FreeMapA[T, A]]`

connector/src/main/scala/quasar/qscript/qsu/ExpandShifts.scala
```diff
@@ -78,9 +78,9 @@ final class ExpandShifts[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes
       val shiftedG = sortedShifts match {
         case (struct, idStatus, rotation) :: ss =>
           val firstRepair: FreeMapA[QScriptUniform.ShiftTarget[T]] =
-            func.ConcatMaps(
-              func.MakeMapS("original", func.AccessLeftTarget(Access.valueHole(_))),
-              func.MakeMapS("0", func.RightTarget)
+            func.StaticMap(
+              ("original", func.AccessLeftTarget(Access.valueHole(_))),
```
djspiewak: Personally, I prefer `"original" -> func.AccessLeft ...` as syntax, since it avoids the extra parentheses and overloaded `,` meaning.

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -57,9 +56,9 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
 
     case graph @ Extractors.LPJoin(left, right, cond, jtype, lref, rref) =>
       val combiner: JoinFunc =
-        func.ConcatMaps(
-          func.MakeMap(StrLit[T, JoinSide](JoinDir.Left.name), func.LeftSide),
-          func.MakeMap(StrLit[T, JoinSide](JoinDir.Right.name), func.RightSide))
+        func.StaticMap(
+          (JoinDir.Left.name, func.LeftSide),
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -86,7 +85,7 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
       def autojoinVerts(head: Symbol, tail: List[Symbol]): F[(Symbol, QSUVerts[T])] = {
         freshName[F].flatMap { joinRoot =>
           val join: QSU[Symbol] = AutoJoin2(src.root, head,
-            func.ConcatMaps(func.MakeMapS("sort_source", func.LeftSide), func.MakeMapS(head.name, func.RightSide)))
+            func.StaticMap(("sort_source", func.LeftSide), (head.name, func.RightSide)))
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/ReifyAutoJoins.scala
```diff
@@ -90,9 +90,9 @@ final class ReifyAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends QSU
             prov.autojoinKeys(ldims, cdims)
 
           def lcCombiner: JoinFunc =
-            func.ConcatMaps(
-              func.MakeMap(StrLit(lName), LeftSideF),
-              func.MakeMap(StrLit(cName), RightSideF))
+            func.StaticMap(
+              (lName, LeftSideF),
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/ReifyIdentities.scala
```diff
@@ -177,9 +177,9 @@ final class ReifyIdentities[T[_[_]]: BirecursiveT: ShowT] private () extends QSU
       makeI[Id, A](sym -> id)
 
     def makeIV[A](initialI: FreeMapA[A], initialV: FreeMapA[A]): FreeMapA[A] =
-      func.ConcatMaps(
-        func.MakeMapS(IdentitiesK, initialI),
-        func.MakeMapS(ValueK, initialV))
+      func.StaticMap(
+        (IdentitiesK, initialI),
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -223,16 +223,16 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+            (ResultsField, repair))
 
           QSU.LeftShift[T, Symbol](src.root, struct, idStatus, repair2, rot)
 
         case \/-(QSU.MultiLeftShift(_, shifts, repair)) =>
-          val repair2 = func.ConcatMaps(
-            func.MakeMapS(OriginalField, accessHoleLeftF),
-            func.MakeMapS(ResultsField, repair))
+          val repair2 = func.StaticMap(
+            (OriginalField, accessHoleLeftF),
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -253,13 +253,12 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
               case QSU.RightTarget() => func.RightTarget
             }
 
-            val repair3 = func.ConcatMaps(
-              func.MakeMapS(
-                OriginalField,
+            val repair3 = func.StaticMap(
+              (OriginalField,
```
djspiewak: MOAR DITTO

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -280,11 +279,10 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
               case \/-(idx) => idx.right[QAccess[Hole]].point[FreeMapA]
             }
 
-            val repair3 = func.ConcatMaps(
-              func.MakeMapS(
-                OriginalField,
+            val repair3 = func.StaticMap(
+              (OriginalField,
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -429,9 +427,9 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
           val repairRAdj = fixSingleRepairForMulti(repairR, 1, RightSide)
 
           val repair =
-            func.ConcatMaps(
-              func.MakeMapS(LeftField, repairLAdj),
-              func.MakeMapS(RightField, repairRAdj))
+            func.StaticMap(
+              (LeftField, repairLAdj),
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -454,9 +452,9 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
           val repairRAdj = fixSingleRepairForMulti(repairR, offset, RightSide)
 
           val repair =
-            func.ConcatMaps(
-              func.MakeMapS(LeftField, repairLAdj),
-              func.MakeMapS(RightField, repairRAdj))
+            func.StaticMap(
+              (LeftField, repairLAdj),
```
djspiewak: Ditto

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -477,9 +475,9 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
           val repairRAdj = fixMultiRepair(repairR, 1, RightSide)
 
           val repair =
-            func.ConcatMaps(
-              func.MakeMapS(LeftField, repairLAdj),
-              func.MakeMapS(RightField, repairRAdj))
+            func.StaticMap(
+              (LeftField, repairLAdj),
```
djspiewak: Ok just basically all the things…

connector/src/main/scala/quasar/qscript/construction.scala
```diff
@@ -137,6 +137,13 @@ object construction {
       rollCore(MapFuncsCore.MakeMap(key, src))
     def MakeMapS[A](key: String, src: FreeMapA[T, A]): FreeMapA[T, A] =
       MakeMap(Constant(json.str(key)), src)
+    def StaticMap[A](pairs: (String, FreeMapA[T, A])*): FreeMapA[T, A] =
```
wemrysi: For completeness, could this be `(T[EJson], FreeMapA[T, A]) => FreeMapA[T, A]` and then have a `StaticMapS` variant like we do the other map functions?

connector/src/main/scala/quasar/qscript/construction.scala
wemrysi: Same here, exposing the general and more-common string-keyed variant.

connector/src/main/scala/quasar/qscript/construction.scala
djspiewak: I don't think we would ever use the ejson variant though.  This is strictly a utility which is replacing something that was *already* built with `MakeMapS`.

wemrysi: I agree it'll be rare if we ever use it, but I could see backends that support non-string keyed maps want to build them in tests. I don't feel all that strongly about this, though, we can always add it later.

However, we should probably be using `MapFuncCore.StaticMap` here to DRY things up and prevent them from diverging.

connector/src/main/scala/quasar/qscript/construction.scala
edmundnoble: Because I have the head of the list here already, in the wrong place to call `foldLeft` with.

connector/src/main/scala/quasar/qscript/construction.scala
edmundnoble: I'm not following @djspiewak. Why would an applicative variation be necessary to have? And are you suggesting `StaticMapGen` as a name? And @wemrysi, do you mean that

connector/src/main/scala/quasar/qscript/qsu/ExpandShifts.scala
edmundnoble: An old holdover from IntelliJ, which told me that `->` was invalid in Quasar :/ fixing.

connector/src/main/scala/quasar/qscript/construction.scala
edmundnoble: MapFuncCore's DSL is slated for removal, when fully replaced.

connector/src/main/scala/quasar/qscript/construction.scala
edmundnoble: Interestingly this seems to mean that `MapFuncCore.StaticMap` is incorrectly implemented. `MapFuncCore.StaticMap("a" -> x, "b" -> y)` actually gives you `ConcatMaps(MakeMapS("b", y), MakeMaps("a", x))`.

connector/src/main/scala/quasar/qscript/construction.scala
wemrysi: That's fine, but until that happens why not make it easy to ensure they produce compatible results?

edmundnoble: The problem with that is that `MapFuncCore.StaticMap` is incorrectly implemented. See my above comment to @djspiewak.

edmundnoble: Hang on actually. Still thinking about this one.

connector/src/main/scala/quasar/qscript/construction.scala
wemrysi: @edmundnoble Yes, for the same reason as for `StaticMap`. I think if we're going to have a `FreeMap` dsl it should allow us to express all `FreeMap`s. But, as I mentioned, if y'all disagree then I guess whoever needs the first non-stringly-keyed map can add the implementation.

connector/src/main/scala/quasar/qscript/construction.scala
wemrysi: If that's the case, then it's an argument _for_ implementing one in terms of the other. We don't want them to be different, so if there is a bug we should fix it.

connector/src/main/scala/quasar/qscript/construction.scala
djspiewak: How is `foldRight . reverse` different from `foldLeft`?

edmundnoble: Yup. I got a bit lost in mental tree diagrams. Fixing that :blush:

djspiewak:
@edmundnoble BTW, needs a version label

PR #3391, opened 2018-01-10
Fix key order in orderByDescAsc

jsantos17:


it/src/main/resources/tests/orderByDescAsc.test
```diff
@@ -1,26 +1,19 @@
...
     "backends": {
-        "couchbase":         "pending",
-        "marklogic_json":    "pending",
-        "marklogic_xml":     "pending",
-        "mongodb_3_2":       "pending",
-        "mongodb_3_4":       "pending",
-        "mongodb_read_only": "pending",
-        "spark_local":       "pending",
-        "spark_hdfs":        "pending"
+        "mimir":       "pending"
```
alissapajer: `mimir` should pass if you use `ignoreFieldOrder`. `mimir` doesn't guarantee the results will have field order the same as in the query.

I believe that `marklogic_json` and `couchbase` also do not guarantee field order in the results, but @wemrysi and @drostron (or whoever knows) should confirm.

drostron: Correct, `couchbase` does not guarantee field order.

jsantos17: Marklogic does guarantee field order.  I'll update `couchbase` and `mimir` to `ignoreFieldOrder`. Thanks for all the input!

wemrysi: @jsantos17 It does for `XML`, but I didn't think it did for `JSON`.

jsantos17: Oh right. I might've extrapolated from `XML`. I'll set `marklogic_json` to `ignoreFieldOrder` too.

jsantos17:
@alissapajer done! :)

PR #3395, opened 2018-01-11
Unit tests passing/pending on next-major

alissapajer:
Depends on #3393 

All the unit tests should be passing (made 5 temporarily `pending` in mongo), so the integration tests will run in travis now.


mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -74,18 +73,21 @@ object MongoDbPlanner {
...
-          Some(_))
+    typ => f.lift(typ).fold(
+      typ match {
+        case Type.Temporal =>
+          generateTypeCheck(or)(f)(Type.OffsetDateTime)
+        case Type.LocalDateTime ⨿ OffsetDateTime =>  // time_of_day
+          generateTypeCheck(or)(f)(Type.OffsetDateTime)
+        case Type.OffsetDateTime ⨿ Type.OffsetDate ⨿
+            Type.LocalDateTime ⨿ Type.LocalDate =>  // date_part
+          generateTypeCheck(or)(f)(Type.OffsetDateTime)
```
edmundnoble: It may be worth substituting this with `case Type.Temporal.superOf(_) =>`, but I don't quite follow why `LocalDateTime` type checks as `OffsetDateTime`.

alissapajer: Yeah, so this is (maybe) temporary in order to get most of the mongo planner specs passing so that integration tests run on travis. Mongo only has a single bson type for time, so the typechecks here are basically hacks so that in another part of the code mongo will not insert multiple typechecks for the same bson type. It looks like this hack is being used in master as well; after I dig into it a bit more, I'll get some input from people who've worked more on mongo as well.

I haven't really looked into the mongo planner yet; I just put in the minimal number of changes to get the most tests working. So if you're okay with it, I'd like to merge this as is, and then continue  the mongo work in another PR.

alissapajer: But also, if we ultimately go this way, I think using `case Type.Temporal.superOf(_) =>` would be better than having all the individual cases.

edmundnoble: I follow, I mean it's definitely mergeable to me regardless. We can't do this long-term all the way for offset cases because `Bson.Date` doesn't include a time zone offset, but right now I think the most serious consequences would be for methods that can produce `LocalDate sum LocalTime` having their results fed to something like `ExtractSecond` with nonsense results. If that's already happening with Mongo though (which I presume it would be if this is being used in master) then of course we can deal with that after next-major's merge.

PR #3396, opened 2018-01-11
Extract "filtering" Cond expressions in addition to Guards

wemrysi:
Extends the extracting of "filtering" conditional expressions (where one branch is `Undefined`) to `Cond` in addition to `Guard`.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/MapFuncCore.scala
+++ b/connector/src/main/scala/quasar/qscript/MapFuncCore.scala
@@ -342,6 +342,14 @@ object MapFuncCore {
           if index.isValidInt =>
         as.lift(index.intValue).map(_.project)
 
+      case ProjectIndex(
+        ExtractFunc(Cond(cond, Embed(StaticArrayPrefix(consArr)), Embed(StaticArrayPrefix(altArr)))),
+        ExtractFunc(Constant(Embed(EX(ejson.Int(index))))))
+          if index.isValidInt =>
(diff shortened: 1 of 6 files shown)


connector/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -375,34 +342,32 @@ object MapFuncCore {
...
+      case ProjectIndex(
+        ExtractFunc(Cond(cond, Embed(StaticArrayPrefix(consArr)), Embed(StaticArrayPrefix(altArr)))),
+        ExtractFunc(Constant(Embed(EX(ejson.Int(index))))))
+          if index.isValidInt =>
+        (consArr.lift(index.intValue) |@| altArr.lift(index.intValue)) {
           case (cons, alt) => rollMF[T, A](MFC(Cond(cond, cons, alt)))
         }
 
       case ProjectKey(
-        Embed(CoEnv(\/-(MFC(Cond(cond, Embed(StaticMap(map)), alt))))),
```
wemrysi: Even though `ExtractFiltering` only covers this case when `alt = Undefined`, this was still removed as it was incorrect (we have to `ProjectKey(alt, key)`) and in order to make it work smoothly we need an additional rewrite that collapses expressions containing `Undefined` to `Undefined` (sans a few special cases).

Once we have that rewrite we could restore a working version of these cases.

jsantos17: Hmm, I can see it now. Good catch!

PR #3397, opened 2018-01-11
Timing dump endpoint

edmundnoble:
Adds /timedump endpoint to fetch timing information as JSON. Also removes `printTiming` option from the web server because it's now pointless, and probably always was.

Also makes the timing repository use a `Queue` and not a `TreeSet`, for great ~~justice~~ constant factors.

Depends on #3382.


web/src/main/scala/quasar/api/services/timedump.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.http4s.dsl._
+import scalaz.{Zip => _, _}, Scalaz._
+import scalaz.concurrent.Task
+import argonaut.Json
+
+object timedump {
+
+  def service[S[_]](repo: TimingRepository)(
+    implicit
+    R: ReadFile.Ops[S],
```
wemrysi: Looks like everything but `Task :<: S` can be removed.

edmundnoble: Don't know how I missed that, thanks :+1:

effect/src/main/scala/quasar/effect/ScopeExecution.scala
```diff
@@ -19,58 +19,60 @@ package quasar.effect
...
+  def newExecution[A](executionId: ExecutionId, action: ScopeTiming[F, T] => F[A]): F[A]
 }
 
 object ScopeExecution {
   def ignore[F[_], T]: ScopeExecution[F, T] = new ScopeExecution[F, T] {
-    def newExecution[A](index: Long, action: ScopeTiming[F, T] => F[A]): F[A] =
+    def newExecution[A](executionId: ExecutionId, action: ScopeTiming[F, T] => F[A]): F[A] =
       action(ScopeTiming.ignore[F, T])
   }
   def forTask[T](repo: TimingRepository,
```
wemrysi: You might consider defining this (and other impls) in terms of `quasar.effect.Capture` to have a single implementation that can be instantiated at `Task` and `Task <: F => Free[F, ?]`, etc.

web/src/main/scala/quasar/api/services/RestApi.scala
```diff
@@ -71,7 +71,8 @@ object RestApi {
       "/query/fs"     -> query.execute.service[S, T](executionIdRef),
       "/invoke/fs"    -> invoke.service[S],
       "/schema/fs"    -> analyze.schema.service[S],
-      "/metastore"    -> metastore.service[S]
+      "/metastore"    -> metastore.service[S],
+      "/timedump"     -> timedump.service[S](timingRepo)
```
wemrysi: How about `/timings` or `/execution-timings` as the endpoint?

effect/src/main/scala/quasar/effect/ScopeExecution.scala
edmundnoble: Excellent.

web/src/main/scala/quasar/api/services/RestApi.scala
edmundnoble: `/timings` makes sense to me.

edmundnoble:
I agree that `ExecutionId` generation *could* be an effect, I just wasn't sure if it was worth it seeing as there's only one implementation. However, I don't agree that `TimingRepository` access makes sense as an effect. `ScopeExecution`'s job is to hide that `TimingRepository` from writers, and the time dump is the only possible reader of it, meaning that there's only one consumer of that effect.

With that said, one step on the way to more granular timing scopes *is* through a `QueryFile` interpreter transformer; I just don't understand the idea that we can exclude the particulars of timing from business logic. It seems to be that this is exactly where it has to be and should be.

wemrysi:
Not sure I agree with that, `TimingRepository` now appears in the signature of `coreServices` because there is another consumer of it: the `/timings` endpoint.

effect/src/main/scala/quasar/effect/ScopeExecution.scala
edmundnoble: Actually this isn't really possible. All `forX` methods mutate `TaskRef`s.

wemrysi:
And yeah, after reviewing the granularity at which we're timing, it will have to be mixed in with the logic (at least the scope demarcation) to some degree. It is another reminder that we need to extract more of the logic from the endpoints so it can be reused with the repl (a sort of quasar-client library if you will).

edmundnoble:
Right, that's what I mean. Timedump endpoint, timings endpoint, same thing. Still one reader of TimingRepository.

wemrysi:
Ah, yes, saying the same thing, I missed "time dump".

effect/src/main/scala/quasar/effect/ScopeExecution.scala
wemrysi: 😞 If we had a final encoding of `quasar.effect.AtomicRef` then we could implement in terms of `F: Capture: AtomicRef`.

edmundnoble: Definitely think that would be nice. Hell, I think you could even keep it as-is but parameterize it like `AtomicRef[F]` and make the methods and factory require a `Capture[F]`, because I can't see any other monads that make sense.

wemrysi:
@edmundnoble One more thing, would you update the `README` with the details of the new endpoint? Apologies for not noticing this earlier. Include `[ci skip]` in the commit to avoid a rebuild, I'll merge once that's ready.

PR #3400, opened 2018-01-12
Fix typechecking bug in IfUndefined.

alissapajer:
This fixes the broken `IfUndefined` integration tests.

#qz-3580 Done

### DIFF
--- a/frontend/src/main/scala/quasar/std/relations.scala
+++ b/frontend/src/main/scala/quasar/std/relations.scala
@@ -172,11 +172,7 @@ trait RelationsLib extends Library {
     noSimplification,
     partialTyper[nat._2] {
       case Sized(Type.Const(Data.NA), fallback) => fallback
-      case Sized(Type.Bottom,         fallback) => fallback
-      case Sized(value,               fallback) => 
-        if (value.contains(Type.Top)) fallback 
-        else if (fallback.contains(Type.Top)) value 
-        else value ⨿ fallback
+      case Sized(value, fallback) => Type.TypeOrMonoid.append(value, fallback)
     },
     basicUntyper)
 
--- a/frontend/src/main/scala/quasar/types.scala
+++ b/frontend/src/main/scala/quasar/types.scala
@@ -218,7 +218,8 @@ trait TypeInstances {
     def append(v1: Type, v2: => Type) = (v1, v2) match {
       case (Type.Bottom, that) => that
       case (this0, Type.Bottom) => this0
-      case _ => v1 ⨿ v2
+      case (this0, that0) if this0.contains(Type.Top) || that0.contains(Type.Top) => Type.Top
+      case (_, _) => v1 ⨿ v2
     }
   }
 


frontend/src/main/scala/quasar/std/relations.scala
```diff
@@ -172,11 +172,7 @@ trait RelationsLib extends Library {
     noSimplification,
     partialTyper[nat._2] {
       case Sized(Type.Const(Data.NA), fallback) => fallback
-      case Sized(Type.Bottom,         fallback) => fallback
-      case Sized(value,               fallback) => 
-        if (value.contains(Type.Top)) fallback 
-        else if (fallback.contains(Type.Top)) value 
-        else value ⨿ fallback
+      case Sized(value, fallback) => Type.TypeOrMonoid.append(value, fallback)
```
alissapajer: This is basically reverting to what is in master already.

edmundnoble: What's the bug here?

alissapajer: `if (value.contains(Type.Top)) fallback` is incorrect because just because one side containing `Type.Top` doesn't mean the result type is the type of the other side; it means that the best we can say about the result type is that it's `Type.Top`.

alissapajer: This was causing a query like
```sql
select city ?? false as c from zips
```
to create a constant plan `Constant(Map(List((Str(c),Bool(false)))))`

edmundnoble: Ah. Makes perfect sense to me. Here's the reason why I introduced this code: otherwise, we generate types like `Type.Top sum Type.Int`, which make the type system very odd (notably, there is not a single upper bound for all types, which means a lot of code has to check for `contains(Type.Top)` rather than `== Type.Top`). I believe the *best* fix here is to change `TypeOrMonoid` to check for `Type.Top` and avoid producing coproducts with top, and use it here. It's just not easy when we expose constructors like this to maintain these invariants.

edmundnoble: In the long-term it makes more sense to me to use prisms and not constructors for types to avoid breaking these invariants.

frontend/src/main/scala/quasar/types.scala
```diff
@@ -218,7 +218,8 @@ trait TypeInstances {
     def append(v1: Type, v2: => Type) = (v1, v2) match {
       case (Type.Bottom, that) => that
       case (this0, Type.Bottom) => this0
-      case _ => v1 ⨿ v2
+      case (this0, that0) if this0.contains(Type.Top) || that0.contains(Type.Top) => Type.Top
```
alissapajer: @edmundnoble is your suggestion above not what I did here?

edmundnoble: Oh! I didn't see this :blush: LGTM.

frontend/src/main/scala/quasar/std/relations.scala
alissapajer: Yeah, we have no way to enforce these invariants currently, and so we end up having various versions of type comparison that we use semi-arbitrarily in various places.

frontend/src/main/scala/quasar/types.scala
alissapajer: :smile:

edmundnoble:
Oh, also needs version label.

alissapajer:
This is against next-major.

edmundnoble:
Ah.

edmundnoble:
Right, can you PR the `TypeOrMonoid` change to master later?

alissapajer:
Yeah, I actually meant to open for discussion whether this change would be better PRd directly to master. I think I'll just close this PR and reopen the analogous one against master. And then when we merge in master to next-major, I'll update next-major to have the same typechecking as master.

edmundnoble:
`relations.scala` is still a reversion to master, no? Maybe only include that change in here, and we won't merge it until your PR to master is merged to next-major?

alissapajer:
It's not quite a reversion to master because in master we don't call to the monoid. So I think I'll just open a new PR against next-major including the commit that will be PRd to master. I think that has the most understandable history.

frontend/src/main/scala/quasar/std/relations.scala
wemrysi: Hopefully we can switch to the newer representation for these sooner than later which normalizes through known operations and doesn't use structural equality so `Top + Int === Top`.

PR #3402, opened 2018-01-12
Update the Cond typer to use TypeOrMonoid.

alissapajer:


alissapajer:
Dangit typo in the commit message - this is actually fixing `IfUndefined`... I'm going to close this an open yet another one... :)

PR #3403, opened 2018-01-12
Update the IfUndefined typer to use TypeOrMonoid.

alissapajer:


frontend/src/main/scala/quasar/std/relations.scala
```diff
@@ -146,9 +146,8 @@ trait RelationsLib extends Library {
     Func.Input2(Type.Top, Type.Top),
     noSimplification,
     partialTyper {
-      case Sized(Type.Bottom, fallback)                                    => fallback
```
edmundnoble: Realistically this would only have been called in case of bugged `Cond` and `Constantly`, because it's not checking for `contains(Type.Bottom)`. Oh dear.

frontend/src/main/scala/quasar/std/relations.scala
```diff
@@ -146,9 +146,8 @@ trait RelationsLib extends Library {
...
-      case Sized(t1, t2) if t1.contains(Type.Top) || t2.contains(Type.Top) => Type.Top
-      case Sized(value, fallback)                                    => value ⨿ fallback
+      case Sized(Type.Const(Data.NA), fallback) => fallback
```
edmundnoble: So at first glance this looks better than what we had before. But if undefined is returned anywhere in the simplifier we're going to have bugs anyway, seeing as we don't remove it from arrays and objects, so I think this is actually unnecessary.

frontend/src/main/scala/quasar/types.scala
```diff
@@ -205,7 +205,8 @@ trait TypeInstances {
     def append(v1: Type, v2: => Type) = (v1, v2) match {
       case (Type.Bottom, that) => that
       case (this0, Type.Bottom) => this0
-      case _ => v1 ⨿ v2
+      case (this0, that0) if this0.contains(Type.Top) || that0.contains(Type.Top) => Type.Top
+      case (_, _) => v1 ⨿ v2
```
edmundnoble: :+1:

frontend/src/main/scala/quasar/std/relations.scala
alissapajer: All types are going to pass the `contains(Type.Bottom)` check though, no?

alissapajer:
@edmundnoble ready for another look.

frontend/src/main/scala/quasar/std/relations.scala
edmundnoble: Right, that'd be `Type.Bottom.contains(_)`.

edmundnoble: And all types *shouldn't* pass that test because you can't assign a `Data.Set(Nil)` to a `Data.NA`.

PR #3405, opened 2018-01-15
RDBMS unwind/unnest/flatten

mikolak-net:


mikolak-net:
All tests added by this PR have been skipped for the "problematic" connectors, so merging this to master will not break anything beyond what's already broken by other changes.

PR #3407, opened 2018-01-15
Add 'onUndefined' property to LeftShift

wemrysi:
Adds an `onUndefined` property to `LeftShift` that indicates how to handle `Undefined` values produced by the `struct` expression from a `LeftShift`.

Updating individual connectors and properly emitting this value from `QSU` will come along shortly.

Part of #qz-3583.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/OnUndefined.scala
+++ b/connector/src/main/scala/quasar/qscript/OnUndefined.scala
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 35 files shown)


connector/src/main/scala/quasar/qscript/OnUndefined.scala
```diff
@@ -0,0 +1,67 @@
...
+        x.fold(omit, emit)
+
+      override val min: Option[OnUndefined] =
+        Some(Emit)
+
+      override val max: Option[OnUndefined] =
+        Some(Omit)
+
+      private val asBool: OnUndefined => Boolean =
+        _.fold(false, true)
```
djspiewak: This seems backwards.

wemrysi: It's not meant to have any semantic meaning, just using it for the Order instance. I could use an Int or something else if that'd be more clear.

PR #3408, opened 2018-01-15
Use the LeftShift#onUndefined property in QSU

djspiewak:
Depends on #3407

Whenever shifts are collapsed, we now set `onUndefined` to `Emit`, regardless of what the original value was.  This ensures that we carry along the original value, which may be shifted or used in subsequent computation.  In theory, the qscript for queries such as `select a[*], b from foo` should now be correct.

Part of #qz-3583.


wemrysi:
Looks like just needs a merge from master.

rintcius:
@djspiewak Added fix here #3426

PR #3409, opened 2018-01-15
Disable problematic test

djspiewak:
Should get the build functioning again.


alissapajer:
Let's merge this by merging #3410 so that we can confirm that the build is green for all connectors.

PR #3410, opened 2018-01-15
Display incremental regression test results in travis build

wemrysi:
We were previously only displaying errors, but now that we're using the incremental specs2 output as a quasi-heartbeat, we need to see all the output to avoid travis killing the job due to inactivity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ script:
   - |-
     ./sbt -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION \
       it/sideEffectTestFSConfig \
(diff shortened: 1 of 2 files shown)


djspiewak:
The caching is all screwy between stages when the overall build has yet to pass. I wish I had a good way to get around that issue.

rintcius:
Seems like we need to remove more `-- xonly failtrace`s

alissapajer:
Both ML builds received no output for 10 minutes while running integration tests, and so were killed by travis. 

We also run `marklogicIt/testOnly` in the integration test stage; locally that runs in a couple minutes for me, but maybe it takes 10 minutes on travis? I'm going to push a change removing `-- xonly failtrace` for `marklogicIt/testOnly` and see where that gets us.

PR #3411, opened 2018-01-16
Plan flatten arrays using $unwind

rintcius:
Depends on #3380 

#qz-3589 Done

PR #3412, opened 2018-01-16
Plan flatten objects using $unwind

rintcius:
Depends on #3411

#qz-3590 Done


mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -549,10 +482,116 @@ object WorkflowBuilder {
...
+    (field, acc) => {
+      (field, rest, queryModel) match {
+        case (StructureType.Array(field, quasar.qscript.ExcludeId), _, _) =>
+          $unwind[WF](base.toDocVar \\ field, None, None).apply(acc)
+        case (StructureType.Array(field, idStatus), Some(r), _) =>
+          flattenFieldArrayUnwindProject(base, field, idStatus, r, acc)
+        case (StructureType.Array(field, idStatus), _, _) =>
+          flattenFieldMapReduce(base, field, idStatus).apply(acc)
+        case (StructureType.Object(field, idStatus), Some(r), MongoQueryModel.`3.4.4`) =>
+          flattenFieldObject344(base, field, idStatus, r, acc)
```
jsantos17: This is fine for now, but we really need to think of how to version `WorkflowOp` long term.

jsantos17:
On a second thought, I think we need to the rewrite of `Undefined` to `MakeArray(Undefined)` here too. Consider the plan for `select backend{*} from newTests`:

```
Workflow (crystallized):
Chain
├─ $ReadF(jps; newTests)
├─ $ProjectF
│  ├─ Name("o" -> {
│  │       "$objectToArray": {
│  │         "$cond": [
│  │           {
│  │             "$and": [
│  │               { "$lte": [{ "$literal": {  } }, "$backends"] },
│  │               { "$lt": ["$backends", { "$literal": [] }] }]
│  │           },
│  │           "$backends",
│  │           { "$literal": undefined }]
│  │       }
│  │     })
│  ╰─ ExcludeId
├─ $UnwindF
│  ╰─ Path(DocField(BsonField.Name("o")))
╰─ $ProjectF
   ├─ Name("__quasar_mongodb_sigil" -> "$o.v")
   ╰─ ExcludeId
```

In reality, this rewrite should only be necessary when `onUndefined = Emit`. But anyway, that's something we ought to handle. Maybe not in this PR tho.

rintcius:
@jsantos17 Yeah, I was thinking we do the `onUndefined` related stuff separately.

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
rintcius: @jsantos17 Yes, it would be good if we come up with a solution for this. It's more complex than versioning just `WorkflowOp` though, because these are all `WorkflowOpCore` here. We need to incorporate the `ExprOp` version as part of this too.

PR #3417, opened 2018-01-16
Really fix the build this time

wemrysi:
By third time, I meant fourth.


mikolak-net:
Looks like Travis has decided to call it a day - all PR builds are in limbo at the moment (no status indicator for running builds seems to be updated, not even time elapsed).

PR #3421, opened 2018-01-17
LeftShift onUndefined for Mongo

rintcius:
Depends on #3412 and #3408 

#qz-3602 Done


rintcius:
@jsantos17 @wemrysi This implementation of `LeftShift.onUndefined` is simple, but looks like it does the job. Let me know if I have missed something.

wemrysi:
@rintcius Needs a version label.

PR #3422, opened 2018-01-17
Update ExtractFreeMap and ReifyBuckets to use withName

wemrysi:
Avoids direct manipulation of vertices and ensures identical structure will have identical names.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
@@ -101,6 +101,8 @@ final case class QSUGraph[T[_[_]]](
 
     import QScriptUniform._
 
+    val withName = QSUGraph.withName[T, F]("qsu")_
+
     if (src =/= target) {
 
       if (root === src) {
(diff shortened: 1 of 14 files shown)


connector/src/main/scala/quasar/qscript/qsu/QSUGraph.scala
```diff
@@ -101,20 +101,22 @@ final case class QSUGraph[T[_[_]]](
 
     import QScriptUniform._
 
+    val withName = QSUGraph.withName[T, F]("qsu")_
```
djspiewak: One of the few times when point-free style is actually useful in Scala!

connector/src/main/scala/quasar/qscript/qsu/ReadLP.scala
```diff
@@ -268,9 +268,8 @@ final class ReadLP[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T] {
       constr: (Symbol, Symbol, Symbol) => QSU[Symbol]): G[QSUGraph] =
     withName[G](constr(parent1.root, parent2.root, parent3.root)).map(_ :++ parent1 :++ parent2 :++ parent3)
 
-  private def withName[G[_]: Monad: NameGenerator: MonadState_[?[_], RevIdx]](
-      node: QSU[Symbol]): G[QSUGraph] =
-    QSUGraph.withName[T, G](node)
+  private def withName[G[_]: Monad: NameGenerator: RevIdxM](node: QScriptUniform[Symbol]): G[QSUGraph] =
+    QSUGraph.withName[T, G]("rlp")(node)
```
djspiewak: I read this as `rip`…

PR #3423, opened 2018-01-17
Merge master into next-major

alissapajer:
Depends on #3420


alissapajer:
This gets the build fixes into next-major.

PR #3424, opened 2018-01-17
Generalize selector phase

jsantos17:
Before generalizing for every `FreeMap` in the connector I generalized it for `LeftShift` since that's our immediate pain point. Still missing tests and a merge from master, hence WIP.

So, TODO:

What needs to be accomplished before merging this.

- [x] The `condSelector` algebra is backwards. It should attempt to use `selector` first and then the matchers inside `condSelector`. This makes some matchers suboptimal.
- [x] The `Undefined` => `MakeArray(Undefined)` rewrite is not applying always for some reason I don't fully understand yet.
- [x] Rework failing unit tests. (worth noting integration tests are working fine)
- [x] Add more tests.
- [x] `invoke2rel` is duplicated. Refactor missing.
- [x] Matches on `Cond`, `ConcatMaps` and `MakeMap` do not belong in the `selector` algebra. They should go in `condSelector` instead.

Can be deferred to a second PR:

- [ ] Figure out how to delete nodes pointed by `InputFinder` (this is necessary before further generalization)
- [ ] Only works for LeftShift. This needs to be generalized to `getBuilder`

#qz-3329 Done

Related to #qz-3560, but not done yet since generalization is not complete.


jsantos17:
At the moment this only works for array left shifts. But generalization to `getBuilder` should fix that too (and clean up the `case LeftShift =>` impl).

@rintcius Can we start reviewing this? I'll open another PR with the generalization and think about how to use `InputFinder` to delete extracted nodes.

it/src/main/resources/tests/groupByFlatten.test
```diff
@@ -4,9 +4,6 @@
         "couchbase":         "pending",
         "marklogic_json":    "pendingIgnoreFieldOrder",
         "mimir":             "ignoreFieldOrder",
-        "mongodb_3_2":       "pending",
-        "mongodb_3_4":       "pending",
-        "mongodb_read_only": "pending",
```
rintcius: 😆

it/src/main/resources/tests/sumFlattenedIntArrays.test
```diff
@@ -2,9 +2,6 @@
     "name": "sum flattened int arrays",
     "backends": {
         "couchbase": "pending",
-        "mongodb_3_2": "pending",
-        "mongodb_3_4": "pending",
-        "mongodb_read_only": "pending",
```
rintcius: You can add `#qz-3329 Done` to the description

rintcius: ... and a link to qz-3329

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -797,12 +822,6 @@ object MongoDbPlanner {
           case MFC(Not((_, v))) =>
             v.map { case (sel, inputs) => (sel andThen (_.negate), inputs.map(There(0, _))) }
 
-          case MFC(Cond((_, v), _, _)) =>
-            v.map { case (sel, inputs) => (sel, inputs.map(There(0, _))) }
-
-          case MFC(ProjectKey((Embed(MFC(Cond(_, _, _))), v), _)) =>
-            v.map { case (sel, inputs) => (sel, inputs.map(There(0, _))) }
```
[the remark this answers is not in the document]
jsantos17: This was a crutch I was using because relational operations `(Eq, Neq, Lt, Lte, Gt, Gte)` were nested under that construction. However, matching directly for them like in `condSelector` is far more reliable. This algebra is also used to plan filters to I wanted to remove anything not necessary for filters from it.

it/src/main/resources/tests/backendStatuses.test
```diff
@@ -4,9 +4,7 @@
         "couchbase":         "skip",
         "marklogic_json":    "skip",
         "marklogic_xml":     "skip",
-        "mongodb_3_2":       "skip",
-        "mongodb_3_4":       "skip",
-        "mongodb_read_only": "skip",
+        "mongodb_3_2":       "pending",
```
jsantos17: There's likely a bug in the mapReduce implementation for flatten with `IdOnly`/`IncludeId`. Any ideas @rintcius?

jsantos17: Something is definitely broken:

`select backends{_:} from newTests`

errors as well.

jsantos17: After much experimentation I believe the problem happens when `struct` evaluates to `undefined`. Then we attempt to assign to a key on that `undefined`, leading to `TypeError: can't convert undefined to object`. Dealing with JS errors is so painful.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -1033,31 +1037,21 @@ object MongoDbPlanner {
...
-                  getExprBuilder[T, M, WF, EX](cfg.funcHandler, cfg.staticHandler)(src, struct) >>= (builder =>
-                    getExprBuilder[T, M, WF, EX](
-                      cfg.funcHandler, cfg.staticHandler)(
-                      FlatteningBuilder(
-                        builder,
-                        Set(StructureType.Object(DocVar.ROOT(), id)),
-                        List().some),
-                        repair.as(SrcHole)))
-              }
+              getExprBuilder[T, M, WF, EX](cfg.funcHandler, cfg.staticHandler)(src, struct0) >>= (builder =>
```
rintcius: it now uses the rewrites `struct0` and `repair0` for `ShiftType.Map` as well. Is that correct?

jsantos17: Yes. However, there is no difference in the case of `ShiftType.Map` since we won't have typechecks of the form `Guard(_, Type.FlexArr(...), _)` in that case. It simply de-duplicates the logicc here.

I think we want to push the entire "OnUndef" handling to the `WorkflowBuilder` level anyway.

it/src/main/resources/tests/backendStatuses.test
jsantos17: Most definitely not a regression. I was thinking over the weekend how to plan this and I think it's very important we inline the `$project` inside the next `$simpleMap` in the chain.

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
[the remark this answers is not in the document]
rintcius: Ok, but I meant that the `ProjectKey` isn't added to `condSelector`

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
rintcius: Ah yes, gotcha

PR #3426, opened 2018-01-18
Fallback ls repair

rintcius:
Fix compilation error after merge

Part of #qz-3583


djspiewak:
This supersedes #3408.  Thanks, @rintcius!

PR #3427, opened 2018-01-18
mark groupByFlatten for mongo pending.. again

rintcius:


alissapajer:
What's causing this test to fail? This test is failing for mimir on next-major, it could be unrelated though.

rintcius:
@alissapajer there's an inconsistency between `$unwind` (indices are ints) and mapreduce's (indices are strings) flattening. See also Slack for some more info.

PR #3435, opened 2018-01-19
Updated to sbt 1.1

djspiewak:
#qz-2616 Done

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -111,7 +111,7 @@ lazy val backendRewrittenRunSettings = Seq(
       def trace(t: => Throwable): Unit = delegate.trace(t)
     }
 
-    r.run(main, (fullClasspath in Compile).value.files, args ++ backends, filtered) foreach sys.error
+    r.run(main, (fullClasspath in Compile).value.files, args ++ backends, filtered)
   })
 
 // In Travis, the processor count is reported as 32, but only ~2 cores are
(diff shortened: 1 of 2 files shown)


alissapajer:
Can you update our runner as well? https://github.com/quasar-analytics/quasar/blob/master/sbt

djspiewak:
@alissapajer It doesn't need to be updated.  The existing runner already supports sbt 1.1.

alissapajer:
@djspiewak in the runner we default to the sbt version found in `project/build.properties`, but we also have a default version we can force set here https://github.com/quasar-analytics/quasar/blob/master/sbt#L8. Is there a reason you don't want to update this version? I think we should actually just set that to the version found in `project/build.properties`. We can always force an sbt version with `-sbt-version  <version>` so I don't see a reason to set this (or `sbt_unreleased_version`) separately.

djspiewak:
@alissapajer The default version is used when that particular runner is used on a directory which doesn't have a pre-existing `build.properties`.  That situation never applies to us by definition, since we're only using the runner within quasar.  Additionally, updating it puts us out of sync with sbt-extras, which is where the runner is copied from.

alissapajer:
@djspiewak yes, it's clear that we will always use the version in `build.properties`, _unless_ someone tries to use any of the commands in the script that explicitly use `sbt_release_version`, for example `sbt-force-latest`. This will not use the actual latest version of sbt; it will use `0.13.16`, which is the latest version that sbt-extras supports. This version is effectively arbitrary, which renders `sbt-force-latest` a useless command that we should therefore delete. Why is it important that we maintain an exact copy of the script in sbt-extras?

djspiewak:
@alissapajer Mostly because I want to just blindly copy sbt-extras when I update it.  :-)  I don't want to maintain the runner script.  The only purpose for its existence in the repo is essentially to make the Travis build reproducible and devoid of external dependencies.

alissapajer:
@djspiewak I agree that having the build reproducible is important. I just wanted to point out that as we move ahead in sbt version of the script in sbt-extras, many of the commands in our version of the script don't make sense, or at least don't do what you might expect if you don't look closely. And so it will never make sense for anyone to use them. That's fine, because I agree the cost of maintaining the script ourselves would be high.

Can you add a comment in our runner that says that we maintain this exactly from sbt-extras, and so things like the latest version will be whatever the latest version in sbt-extras is. It seems disingenuous   to offer users options like `sbt-force-latest` which will use an sbt version earlier than the one we use in `build.properties`.

djspiewak:
@alissapajer Well I can't really add it to the runner without making it not an exact copy.  :-)  However, I could do some synthesis of a) add a note to the readme, or b) hide both of the runner scripts so that people just don't use them (they literally just exist for travis).

A lot of this comes down to the fact that sbt-extras includes some rather dubious commands (and always have), many of which only make sense in certain circumstances.  That's basically fine, but it means that some situations like this can be a bit weird.  sbt-extras is also designed to be a more general runner (e.g. you'll find it supports sbt 0.7 – which is prior to the complete rewrite in 0.9! – despite the fact that we would never use that functionality).

PR #3436, opened 2018-01-19
Make Range produce an array, implement it in Mimir

edmundnoble:
Maybe there were more goats than there could've been 😉 we'll see what CI has to say.

#qz-2625 Done

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/ArrayLib.scala
+++ b/mimir/src/main/scala/quasar/mimir/ArrayLib.scala
@@ -19,9 +19,10 @@ package quasar.mimir
 import quasar.yggdrasil.bytecode._
 import quasar.blueeyes._
 import quasar.precog.common._
-
 import quasar.yggdrasil._
 import quasar.yggdrasil.table._
+
+import scala.collection.immutable.NumericRange
 
 trait ArrayLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
   trait ArrayLib extends ColumnarTableLib {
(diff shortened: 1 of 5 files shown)


djspiewak:
Wow…  That was fast.

mimir/src/main/scala/quasar/mimir/ArrayLib.scala
```diff
@@ -219,6 +220,20 @@ trait ArrayLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
         between(List(dateToStrCol(target), dateToStrCol(lower), dateToStrCol(upper))).get
     }
 
+    lazy val range: CF2 = CF2P("std::array::range") {
+      case (lower: LongColumn, upper: LongColumn) =>
+        new HomogeneousArrayColumn[Long] {
```
djspiewak: Homogeneous arrays literally don't work throughout most of mimir, so if we start introducing them, it's just not going to work.  As a quick way to reproduce such an issue, try this query on your branch:

```sql
select (pop[0]..pop[1])[*] from extraSmallZips
```

It will crash.

edmundnoble: Huh. So... how can any function return an array?

djspiewak: I don't think it can be done in a column function.  It probably needs to be a primitive `TransSpec`.

edmundnoble: Ahhhhh. Right, I'll give that a go.

beckyconning: Thanks both.

djspiewak:
@edmundnoble I'm going to close this for now (general housekeeping).  That will also give you the opportunity to rebase later if you would like to when you get a chance to work on it again.

PR #3437, opened 2018-01-19
Add unit tests for datetime functions on next-major

alissapajer:
#qz-3593 Done

### DIFF
--- a/frontend/src/main/scala/quasar/std/date.scala
+++ b/frontend/src/main/scala/quasar/std/date.scala
@@ -220,6 +220,8 @@ trait DateLib extends Library with Serializable {
       },
       basicUntyper)
 
+  // FIXME `ZoneOffset.ofTotalSeconds` throws an exception if the integer
+  // input is not in the range [-64800, 64800]
   val SetTimeZone = setTimeZone(
     "Sets the timezone subfield in a date/time value (in seconds east of UTC).",
     ZoneOffset.ofTotalSeconds,
--- a/frontend/src/main/scala/quasar/std/date.scala
+++ b/frontend/src/main/scala/quasar/std/date.scala
@@ -222,6 +222,8 @@ trait DateLib extends Library with Serializable {
 
   // FIXME `ZoneOffset.ofTotalSeconds` throws an exception if the integer
   // input is not in the range [-64800, 64800]
+  //
+  // TODO deduplicate the set time zone impls from mimir's TimeLib
   val SetTimeZone = setTimeZone(
     "Sets the timezone subfield in a date/time value (in seconds east of UTC).",
     ZoneOffset.ofTotalSeconds,
--- a/frontend/src/main/scala/quasar/std/date.scala
+++ b/frontend/src/main/scala/quasar/std/date.scala
@@ -265,8 +265,7 @@ trait DateLib extends Library with Serializable {
     Mapping,
     "Returns the current time zone offset in total seconds - this must always return the same value within the same execution of a query.",
     Type.Int,
-    noSimplification
-  )
+    noSimplification)
 
   val OffsetDateTime = UnaryFunc(
     Mapping,
--- a/frontend/src/main/scala/quasar/std/date.scala
+++ b/frontend/src/main/scala/quasar/std/date.scala
@@ -231,8 +231,9 @@ trait DateLib extends Library with Serializable {
     "Sets the minute component of the timezone subfield in a date/time value.",
     ZoneOffset.ofHoursMinutes(0, _),
     { (i, zo) =>
-      val sec = zo.getTotalSeconds
-      ZoneOffset.ofTotalSeconds(i * 60 + (sec % 3600) * 3600 + sec / 60)
+      val totalSeconds: Int = zo.getTotalSeconds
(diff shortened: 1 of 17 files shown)


alissapajer:
Ooooo all the emojis! :)

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -661,9 +819,19 @@ abstract class StdLibSpec extends Qspec {
         }
       }
 
+      "Interval" >> {
+        unary(
+          Interval(_).embed,
+          Data.Str("1999-12-31T01:22:33.789"),
```
alissapajer: Should this string parse as a `DateTimeInterval`? Or does `DateTimeInterval.parse` only parse strings of the form `"P1Y2M3W4DT5H6M7.8S"`?

Looking at `DateTimeIntervalSpec` it looks like it just parses the variant starting the `"P"`?

frontend/src/main/scala/quasar/std/date.scala
```diff
@@ -229,8 +233,9 @@ trait DateLib extends Library with Serializable {
     "Sets the minute component of the timezone subfield in a date/time value.",
     ZoneOffset.ofHoursMinutes(0, _),
     { (i, zo) =>
-      val sec = zo.getTotalSeconds
-      ZoneOffset.ofTotalSeconds(i * 60 + (sec % 3600) * 3600 + sec / 60)
```
alissapajer: This was returning incorrect results (a java date time construction exception, iirc), so I reimplemented it.

frontend/src/main/scala/quasar/std/date.scala
```diff
@@ -220,6 +220,10 @@ trait DateLib extends Library with Serializable {
       },
       basicUntyper)
 
+  // FIXME `ZoneOffset.ofTotalSeconds` throws an exception if the integer
+  // input is not in the range [-64800, 64800]
```
alissapajer: This comment applies to other datetime `Typer`s as well. Some of the java functions we use expect restricted inputs and throw exceptions for values outside the accepted range. I haven't actually traced through yet if typer exceptions are returned to the user in a helpful fashion (or at all). But we need to test disallowed inputs to these functions (in sql^2) to make sure they return a meaningful error.

Also, a lot of this logic is duplicated in mimir's `TimeLib`, see comment below.

alissapajer:
Some unit tests failed in `quasar.std.SimplifyStdLibSpec`; before merge want to take a look at those to see what's up.

frontend/src/main/scala/quasar/std/date.scala
edmundnoble: They get to the user in a very non-helpful fashion, i.e. they have no idea where the error happened but they get the exception name and description.

I believe returning `undefined` if outside the valid range is a valid way to go about this.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
edmundnoble: It should definitely only parse if it begins with "P". Should work exactly as https://en.wikipedia.org/wiki/ISO_8601#Durations, except that e.g. more than 12 months is allowed.

frontend/src/main/scala/quasar/std/date.scala
edmundnoble: What exactly was incorrect about it? I also believe `(totalSeconds % 3600) / 60 * 60` can be simplified.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
alissapajer: :+1: I'll push a change.

frontend/src/main/scala/quasar/std/date.scala
alissapajer: Say the initial zone offset is 1 minute and 30 seconds, so `sec = 90`. Let `i = 5`, meaning we want the final zone offset to be 5 minutes and 30 seconds.
```scala
(5 * 60) + (90 % 3600) * 3600 + 90 / 60 = 324301  // incorrect (old code)
90 - ((90 % 3600) / 60 * 60) + (5 * 60) = 330  // correct (new code)
```
We cannot simplify `(totalSeconds % 3600) / 60 * 60` to `totalSeconds % 3600` because we are dividing integers in Scala, and integer division produces the integer quotient, for example `9 / 5 = 1`. This deserves a comment, because `x / y * y` not equaling `x` is not expected unless you're thinking in the right context.

I'm not sure what logic was applied to produce the first formula. What I did is take the original number of seconds, subtract out the current number of minutes (in units seconds), and add in the new number of minutes (in units seconds).

frontend/src/main/scala/quasar/std/date.scala
alissapajer: In the name of incremental progress, I'd be fine merging with this `FIXME` in place. I can make an issue to track it.

alissapajer:
@edmundnoble I've pushed another commit. Let me know if there are other changes you'd like to see or discuss before merge!

PR #3441, opened 2018-01-23
Exclude typelevel's scala-library

edmundnoble:
With this change (in addition to the already-merged scalaz 7.2.18 PR), the launcher works!


djspiewak:
@edmundnoble BTW, you forgot to give it a version label.

PR #3445, opened 2018-01-25
Redefine RDBMS error handling

kciesielski:
When using `.attempt(ex => emit(readFailed(...))`, our stream emits a `Left[FileSystemError]`, but this causes problems with database connections. Apparently this swallows the exception, which doesn't go through standard handling, which leaves a leaking connection and freezes integration tests after saturating the connection pool.
Letting the stream crash in a standard way fixes the problem with leaking connections, but I suppose it's not the right way to fail here, because we would prefer to get the `FileSystemError` wrapped in `Backend`?
Could someone confirm this and eventually offer a solution for correct handling?


kciesielski:
Build failed for unknown reasons on `marklogic_json` and `mimir` (timeouts?). Can you bump it @wemrysi ?

PR #3446, opened 2018-01-25
Greedily extract mappable regions prior to autojoining

wemrysi:
When building functions for various constructs (filter predicates, sort keys, group keys, reducer expressions), when we run into an operation that cannot be expressed as a function, we're forced to create an autojoin. However, there may be operations prior to the problematic one that _can_ be expressed as a function, this PR ensures those operations are extracted as a function prior to autojoining.

#qz-3320 Done
#qz-3331 Done

### DIFF
--- a/connector/src/main/scala/quasar/qscript/MapFuncCore.scala
+++ b/connector/src/main/scala/quasar/qscript/MapFuncCore.scala
@@ -353,13 +353,26 @@ object MapFuncCore {
   private def rewrite[T[_[_]]: BirecursiveT: EqualT, A: Equal]:
       CoMapFuncR[T, A] => Option[CoMapFuncR[T, A]] =
     _.run.toOption >>= (MFC.unapply _) >>= {
-      case Eq(v1, v2) if v1 ≟ v2 =>
-        rollMF[T, A](
-          MFC(Constant(EJson.fromCommon(ejson.Bool[T[EJson]](true))))).some
+      case And(BoolLit(true), b) => some(b.project)
+
+      case And(a, BoolLit(true)) => some(a.project)
+
+      case And(BoolLit(false), _) => some(BoolLit[T, A](false).project)
+
+      case And(_, BoolLit(false)) => some(BoolLit[T, A](false).project)
+
+      case Or(a, BoolLit(false)) => some(a.project)
+
+      case Or(BoolLit(false), b) => some(b.project)
+
+      case Or(_, BoolLit(true)) => some(BoolLit[T, A](true).project)
+
+      case Or(BoolLit(true), _) => some(BoolLit[T, A](true).project)
+
+      case Eq(v1, v2) if v1 ≟ v2 => some(BoolLit[T, A](true).project)
 
       case Eq(ExtractFunc(Constant(v1)), ExtractFunc(Constant(v2))) =>
-        rollMF[T, A](
-          MFC(Constant(EJson.fromCommon(ejson.Bool[T[EJson]](v1 ≟ v2))))).some
+        some(BoolLit[T, A](v1 ≟ v2).project)
 
       case DeleteKey(
         Embed(StaticMap(map)),
--- a/connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
@@ -20,12 +20,12 @@ import slamdata.Predef._
 
 import quasar.NameGenerator
 import quasar.Planner.{InternalError, PlannerErrorME}
-import quasar.common.SortDir
+import quasar.fp.symbolOrder
 import quasar.qscript.{construction, JoinSide, LeftSide, RightSide}
 import quasar.sql.JoinDir
 
 import matryoshka.BirecursiveT
-import scalaz.{\/, -\/, \/-, Monad, NonEmptyList => NEL, Scalaz, StateT}, Scalaz._
+import scalaz.{Monad, NonEmptyList, Scalaz, StateT}, Scalaz._
 
 /** Extracts `MapFunc` expressions from operations by requiring an argument
   * to be a function of one or more sibling arguments and creating an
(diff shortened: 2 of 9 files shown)


wemrysi:
@jsantos17 This PR fixes qz-3320, here is the [example query after the fix](https://gist.github.com/wemrysi/22057205a08abfe2c64200b1c1043d8d).

However, it appears to have caused the "subselect in where" regression tests to start failing for MongoDB due to "MongoDB Error: Command failed with error 13070: 'value too large to reduce'". I'm not sure why this is happening as the newer plans for the queries are much nicer and have fewer joins. Maybe the previous query was avoiding some pathological case in MongoDB somehow?

Here is the difference for the `subselectWhere.test`:
* [30.1.2 (master as of this PR)](https://gist.github.com/wemrysi/0546b0e9dabba09e69c3aa5626140aae)
* [30.1.2 + this PR](https://gist.github.com/wemrysi/f7184dd15d2e3181d3ad7f172d92f668)

Any ideas as to why this is happening? Did this QScript actually get worse for MongoDB? Are we not planning optimally in the MongoDB planner?

jsantos17:
@wemrysi looks like the joins are smaller in the previous plan. In order to plan this we might want to use `$lookup` even if there's no index for `state` in `extraSmallZips`. So yeah, we're not producing the most efficient plan here. Phrasing the query as an actual join produces correct results:

```
SELECT distinct patients.first_name, patients.last_name FROM patients join extraSmallZips on patients.state = extraSmallZips.state
```

The difference I see in the QScript is that the actual join produce an `EquiJoin` between two `ShiftedRead`s. The subquery version  has a `Reduce` in a branch of the `EquiJoin` which forces us to force the entire `patients` dataset into a single document before joining. This causes the "value too large to reduce".

<details>
<summary>QScript for subquery</summary>

```
QScript Mongo (Prefer Projection):
Map
├─ Filter
│  ├─ EquiJoin
│  │  ├─ Unreferenced
│  │  ├─ ShiftedRead(/jps/patients, ExcludeId)
│  │  ├─ Reduce
│  │  │  ├─ ShiftedRead(/jps/extraSmallZips, ExcludeId)
│  │  │  ├─ Bucket
│  │  │  │  ╰─ List
│  │  │  ├─ Reducers
│  │  │  │  ╰─ UnshiftArray
│  │  │  │     ╰─ ProjectKey
│  │  │  │        ├─ ○
│  │  │  │        ╰─ Constant(Str(state))
│  │  │  ╰─ Repair
│  │  │     ╰─ ReduceIndex(\/-(0))
│  │  ├─ List
│  │  ├─ Inner
│  │  ╰─ ConcatMaps
│  │     ├─ MakeMap
│  │     │  ├─ Constant(Str(left))
│  │     │  ╰─ LeftSide
│  │     ╰─ MakeMap
│  │        ├─ Constant(Str(right))
│  │        ╰─ RightSide
│  ╰─ Within
│     ├─ ProjectKey
│     │  ├─ ProjectKey
│     │  │  ├─ ○
│     │  │  ╰─ Constant(Str(left))
│     │  ╰─ Constant(Str(state))
│     ╰─ ProjectKey
│        ├─ ○
│        ╰─ Constant(Str(right))
╰─ ConcatMaps
   ├─ MakeMap
   │  ├─ Constant(Str(first_name))
   │  ╰─ ProjectKey
   │     ├─ ProjectKey
   │     │  ├─ ○
   │     │  ╰─ Constant(Str(left))
   │     ╰─ Constant(Str(first_name))
   ╰─ MakeMap
      ├─ Constant(Str(last_name))
      ╰─ ProjectKey
         ├─ ProjectKey
         │  ├─ ○
         │  ╰─ Constant(Str(left))
         ╰─ Constant(Str(last_name))
```
</details>

<details>
<summary>QScript for literal join</summary>

```
EquiJoin
├─ Unreferenced
├─ ShiftedRead(/jps/patients, ExcludeId)
├─ ShiftedRead(/jps/extraSmallZips, ExcludeId)
├─ List
│  ╰─ tuple
│     ├─ ProjectKey
│     │  ├─ ○
│     │  ╰─ Constant(Str(state))
│     ╰─ ProjectKey
│        ├─ ○
│        ╰─ Constant(Str(state))
├─ Inner
╰─ ConcatMaps
   ├─ MakeMap
   │  ├─ Constant(Str(first_name))
   │  ╰─ ProjectKey
   │     ├─ LeftSide
   │     ╰─ Constant(Str(first_name))
   ╰─ MakeMap
      ├─ Constant(Str(last_name))
      ╰─ ProjectKey
         ├─ LeftSide
         ╰─ Constant(Str(last_name))
```
</details>

What is the purpose of the `Reduce` with `UnshiftArray` in `Reduce` instead of using the join key?

djspiewak:
@jsantos17 `Within` operates on an array.  That's basically the purpose.  So the query in this case (the subquery select) is literally *asking* for an `UnshiftArray`.  The fact that we weren't planning it this way in the past is weird, and arguably wrong.

jsantos17:
All in all: I'm okay with merging this since the impact of qz-3320 is very severe for Mongo. But I still believe the QScript can be improved.

wemrysi:
@djspiewak @jsantos17 We _were_ (planning it with a reduce)[https://gist.github.com/wemrysi/0546b0e9dabba09e69c3aa5626140aae#file-gistfile1-txt-L49] prior to the fix.

The previous query had two joins and the new query appears to be, essentially, the inner of the two joins in the old query, which is why it beginning to fail is confusing to me.

jsantos17:
@wemrysi Mongo does write back to plan joins. Using an actual join, we have the information from the join keys that we're interested only in `state` from the `patients` collection so we can write to a temporary collection the `patients` grouped by `state`. 

In the case of the subquery, the `EquiJoin` has an empty join key so we have to group by `null` on both sides of the join (i.e problematically, storing the entire `patients` collection in a single document) to make the documents collide (and make Mongo call the reduce method to perform the join). This causes the "value too large to reduce" error.

PS: what are the semantics of an `EquiJoin` with empty `keys`?

wemrysi:
@jsantos17 
> PS: what are the semantics of an EquiJoin with empty keys?

A cartesian.

jsantos17:
> @djspiewak @jsantos17 We were planning it with a reduce prior to the fix.

Yes, [but the `patients` branch of the join has a `Map` that extracts `state`. This avoids pushing the entire `patients` into a single document](https://gist.github.com/wemrysi/0546b0e9dabba09e69c3aa5626140aae#file-gistfile1-txt-L47)

PR #3448, opened 2018-01-25
Elide superfluous guards in guarded expressions

wemrysi:
Simplifies nested guards of the form `Guard(inner @ Guard(ci, ti, ei, Undefined), t, inner, Undefined)` to `Guard(Guard(ci, ti, ei, Undefined), t, ei, Undefined)`, avoiding evaluating the inner condition twice.

Also elides guards from a guarded expression having the same condition as the outer guard.

#qz-3613 Done
#qz-3575 Done


wemrysi:
@alissapajer I realized I have another case I want to handle here, so don't merge yet. I'll reassign you when it's ready.

wemrysi:
Closing this in favor of a new PR.

PR #3450, opened 2018-01-29
[WIP] RDBMS/Postgres fixes

kciesielski:


kciesielski:
Closing due to incorrect branching.

alissapajer:
@kciesielski messaging here because I'm not sure where a better place would be :)

On the `next-major` branch we have overhauled all the quasar date/time types, and so it would be best to hold off further work on RDBMS involving dates and times, since that will all be changing when we land `next-major` in `master`. Everything outstanding regarding dates and times can remain unimplemented for RDBMS for now. Sorry I didn't think to mention this sooner :/

The new date/time types are here https://github.com/quasar-analytics/quasar/blob/next-major/frontend/src/main/scala/quasar/data.scala#L138.

kciesielski:
@alissapajer We've just completed most of our implementation of date/time support for current API. It's been just merged today with #3438. We're also just before the end of our contract. This Thursday will be the last day of our involvement. When are you planning to get `next-major` into `master`?

alissapajer:
@kciesielski `next-major` may not land this week. Since you're reaching the end of your contract, when we land `next-major` someone internally can update the date/time functionality then. And having your date/time implementations on `master` will definitely aid in that update. Thanks!

PR #3452, opened 2018-01-29
[qz-3635] Correct parsing of join conditions

wemrysi:
Corrects an issue when parsing join conditions where sub-queries weren't required to be enclosed in parentheses, causing subsequent terms from the outer query to be incorrectly considered part of the join condition.

#qz-3635 Done.


beckyconning:
🙌 nice!

PR #3454, opened 2018-01-29
Mappable shift coalescence

djspiewak:
Improves `ConsecutiveLeftShift` handling to see through mappable regions, allowing shifts to be aligned correctly in the general case.  Doesn't handle `Cond` rewriting yet (might never; depends on needs).  This unveils other pre-existing bugs within shift handling, necessitating the disabling of `groupByFlatten`.

#qz-3611 Done

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
@@ -53,8 +53,7 @@ import scalaz.{
 final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] private () extends Minimizer[T] {
   import MinimizeAutoJoins._
   import QSUGraph.Extractors._
-
-  private type ShiftGraph = QSU.LeftShift[T, QSUGraph] \/ QSU.MultiLeftShift[T, QSUGraph]
+  import ShiftGraph._
 
   private val func = construction.Func[T]
   private val N = new NormalizableT[T]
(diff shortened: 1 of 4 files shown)


connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -223,27 +226,31 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
       }
 
       for {
         init2 <- updateGraph[T, G](initPattern)
 
         reconstructed <- reversed.tail.foldLeftM[G, QSUGraph](init2) {
-          case (src, -\/(QSU.LeftShift(_, struct, idStatus, _, repair, rot))) =>
-            val struct2 = struct >> func.ProjectKeyS(func.Hole, ResultsField)
+          case (src, Left(QSU.LeftShift(_, struct, idStatus, _, repair, rot), regionFM)) =>
+            val struct2 = struct >> func.ProjectKeyS(regionFM, ResultsField)
```
wemrysi: Hm, shouldn't this be `struct >> regionFM >> func.ProjectKeyS(func.Hole, ResultsField)`? I may be misunderstanding as well, but as-is it will attempt to project the `ResultsField` from the result of applying `regionFM` to the `src` which doesn't seem correct.

connector/src/main/scala/quasar/qscript/qsu/minimizers/CollapseShifts.scala
```diff
@@ -725,17 +737,29 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
       case LeftShift(Read(_), _, _, _, _, _) =>
         None
 
-      case LeftShift(oparent @ Self(parent, inners), struct, idStatus, onUndefined, repair, rot) =>
-        Some((parent, -\/(QSU.LeftShift[T, QSUGraph](oparent, struct, idStatus, onUndefined, repair, rot)) <:: inners))
+      case
+        LeftShift(
```
wemrysi: Related to my other comment, it seems like we should be able to extract the `regionFM` and then just inline it into the `struct` in the returned value and leave the rest of the algorithm unchanged. If not, what am I missing here?

djspiewak: We can't actually do that since the `LeftSide` of the `repair` needs to access it as well.  Conceptually, `regionFM` is the `FreeMap` that would be inserted into a `Map` in the *parent* of the `LeftShift`. I wonder if pushing it into both the `struct` and the `repair` is viable though…

wemrysi: Ah, right. Yeah, seems like `struct >> regionFM` and `repair >>= { case LeftSide => regionFM as LeftSide; case RightSide => func.RightSide }` would do it, no?

djspiewak: Yep.  I'm fiddling with that approach now.  :-)

PR #3456, opened 2018-01-30
RDBMS fixes and comments

kciesielski:
This PR adds some more capabilities to the Postgres connector:
- String/Dec/Bool columns can be now transformed to Json columns when inserting data
- Comparision (eq, neq) to constant values should now use better type casts. This allows especially text to json comparisions
- UnshiftArray reduce function is now supported

This PR also adds a number of code comments to less immediately apparent parts.


alissapajer:
Can the Travis build for `"postgres"` be enabled in this PR so that we know where the integration tests stand?

alissapajer:
Merged master to get the build green on this. I'm fine merging this without enabling Travis, and then we can enable it in another pass. Since the Travis build has never been enabled, one more PR with it disabled seems harmless.

alissapajer:
Going to wait until the `hoarder` PR #3458 lands to try building this again.

PR #3458, opened 2018-02-01
Added hoarder

djspiewak:
Should be ready for merge once the build goes green.  We won't see the effects until it is in master.

#qz-3637 Done


project/plugins.sbt
```diff
@@ -1,8 +1,9 @@
...
-addSbtPlugin("org.scoverage"      % "sbt-scoverage" % "1.5.1")
-addSbtPlugin("com.slamdata"       % "sbt-slamdata"  % "0.5.7")
-addSbtPlugin("pl.project13.scala" % "sbt-jmh"       % "0.2.27")
+addSbtPlugin("com.eed3si9n"          % "sbt-assembly"  % "0.14.5")
+addSbtPlugin("com.eed3si9n"          % "sbt-buildinfo" % "0.7.0")
+addSbtPlugin("io.get-coursier"       % "sbt-coursier"  % "1.0.0-RC12")
+addSbtPlugin("org.scoverage"         % "sbt-scoverage" % "1.5.1")
+addSbtPlugin("com.slamdata"          % "sbt-slamdata"  % "0.5.7")
+addSbtPlugin("pl.project13.scala"    % "sbt-jmh"       % "0.2.27")
+addSbtPlugin("com.github.romanowski" % "hoarder"       % "1.0.2-RC2")
```
alissapajer: This resolves now for me. I wonder if the repo was actually down when I tried yesterday.

djspiewak: Nope. The artifact in their README doesn't exist.  This one isn't mentioned anywhere; I dug it out by querying Sonatype.

alissapajer: I tried the one in the `README` first, but then I went to sonatype. I only saw the release for `"hoardercore"` there though, so I might not have tried `"hoarder"` with the same version number.

Out of curiosity, did you actually find the `"hoarder"` artifact listed somewhere?

djspiewak: I searched for `romanowski` on sonatype. That pulled up `hoardercore` and `hoarder`, and I just took the most recent version of the latter.

alissapajer: Interesting. Compare https://search.maven.org/#search%7Cga%7C1%7Cromanowski to https://mvnrepository.com/search?q=romanowski.

I previously searched the later, which appears to be both lacking results and to have incorrect artifact id information.

djspiewak: LOL

djspiewak:
@alissapajer Ok, so it looks like hoarder only publishes caches *from mainline branches*, and it only uses them from PRs.  That actually makes a ton of sense as a workflow.  So we need to get this into master before we can see it working for sure.

alissapajer:
@djspiewak ok cool. I would still expect a green Travis build on this branch, so I'm restarting the build so we can get that before merge. Then I'm 👍 to merging this and seeing what happens.

djspiewak:
@alissapajer Actually I'm kind of waffling a bit.   The problem is that this *doesn't* solve the "caching doesn't work between stages" issue, which is what's really especially hurting us.  I'm trying to think what we can do about that…

alissapajer:
To summarize, this gets us (in theory) incremental caching at every stage from the master cache. Master builds will still compile from scratch every time, meaning this does not speed up our release process (once a PR is merged). And incremental caching at every stage means that each stage is not using the cache from a previous stage, so PR builds are still compiling more than is necessary.

We'd like to solve those last two problems, but this gets us at least some of the way there. So we're going to go ahead and merge this (when it's green) to speed up PR builds.

PR #3463, opened 2018-02-05
Add test verifying we $match  before $unwind(ing)

jsantos17:
#qz-1880 done

The test checks the query in the issue compile to QScript we know can be in turn be compiled into an efficient plan for Mongo.


djspiewak:
Stealing this one, since @alissapajer already has a zillion things.  :-)

PR #3464, opened 2018-02-06
Ensure guards are created for all nested Array/Map inflations.

wemrysi:
Runtime guards were only being ensured for the inputs to `FlattenArray` and `FlattenMap`, this extends them to also cover indices/keys and the corresponding `Shift` variants.

#qz-3659 Done.


frontend/src/main/scala/quasar/frontend/logicalplan/LogicalPlanR.scala
```diff
@@ -469,40 +531,6 @@ final class LogicalPlanR[T]
       _.cataM(liftTM(checkTypesƒ)).map(appConst(_, constant(Data.NA))).evalZero.validation)
   }
 
-  // TODO: Generalize this to Binder
-  def lpParaZygoHistoM[M[_]: Monad, A, B](
```
djspiewak: Omg it's gone!  :-P

PR #3466, opened 2018-02-08
Upgrade mongo driver to 3.6.3

rintcius:
#qz-3565 Done


djspiewak:
Mimir build failure was spurious. Mongo failures might be legit.

mongodb/src/main/scala/quasar/physical/mongodb/util.scala
```diff
@@ -104,7 +104,9 @@ object util {
         .applyConnectionString(cs)
         .build)
 
-      settings.credentialList(cs.getCredentialList)
+      val cred = cs.getCredential
```
jsantos17: I think we should handle this with `Option` instead of raw `null`. Could something like this work?

```
Option(cs.getCredential) ∘ settings.credential(_)
```

rintcius:
@djspiewak yeah, looking into that mongo failure

mongodb/src/main/scala/quasar/physical/mongodb/util.scala
rintcius: Hmm, for me the current code is more clear. I prefer the explicit use of `null` when interacting with 3rd party, non-functional API's that introduce it. That `null` has to be dealt with either way.

rintcius:
@djspiewak - I'm re-unassigning @jsantos17 until that driver problem is sorted (it's not ready for review yet)

rintcius:
@jsantos17 ready for review now

PR #3468, opened 2018-02-09
Add preBuild before connector-specific integration tests.

alissapajer:
#qz-3637 Done (maybe for reals this time)


djspiewak:
Merging this, because it's the right thing to do even though it doesn't appear to be working 100%.

PR #3469, opened 2018-02-09
Fix BitSet#setLength loop

edmundnoble:
I couldn't develop the benchmark dashboard without fixing this, so here it is :)

### DIFF
--- a/blueeyes/src/main/java/quasar/precog/BitSet.java
+++ b/blueeyes/src/main/java/quasar/precog/BitSet.java
@@ -505,7 +505,7 @@ public class BitSet {
      * @param newLength the new length of the table.
      */
     private final void setLength(final int newLength) {
-        int arrayLength = (int)Math.pow(2, 32 - Integer.numberOfLeadingZeros(newLength));
+        int arrayLength = 1 << (32 - Integer.numberOfLeadingZeros(newLength));
         if (arrayLength < 0 || arrayLength > (1 << 27)) {
             throw new RuntimeException(newLength + " is too large for a BitSet length");
         }
--- a/blueeyes/src/main/java/quasar/precog/BitSet.java
+++ b/blueeyes/src/main/java/quasar/precog/BitSet.java
@@ -505,19 +505,19 @@ public class BitSet {
      * @param newLength the new length of the table.
      */
     private final void setLength(final int newLength) {
-        if (bits.length < newLength) { // Resizes array.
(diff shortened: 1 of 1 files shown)


blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -505,19 +505,19 @@ public int hashCode() {
      * @param newLength the new length of the table.
      */
     private final void setLength(final int newLength) {
-        if (bits.length < newLength) { // Resizes array.
-            int arrayLength = bits.length;
-            while (arrayLength < newLength) {
-                arrayLength <<= 1;
-            }
+        int arrayLength = (int)Math.pow(2, 32 - Integer.numberOfLeadingZeros(newLength));
```
djspiewak: I'm pretty sure this is significantly slower than the `while` loop approach.

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -505,19 +505,19 @@ public int hashCode() {
...
+        if (arrayLength < 0 || arrayLength > (1 << 27)) {
```
djspiewak: Seems like a better solution would be to do a pre-check on the desired `newLength` to ensure that it's sane, but otherwise leave the original implementation in place.

blueeyes/src/main/java/quasar/precog/BitSet.java
edmundnoble: Integer.numberOfLeadingZeros is a single machine instruction, see https://bugs.java.com/view_bug.do?bug_id=6823354. More importantly, this always results in a power of 2. That's the important thing.

edmundnoble: Though I should just `1 << (32 - Integer.numberOfLeadingZeros(newLength))` for sure.

blueeyes/src/main/java/quasar/precog/BitSet.java
edmundnoble: This actually fixes the loop in the original implementation. I just wasn't sure if it was *always* caused by the bug.

blueeyes/src/main/java/quasar/precog/BitSet.java
djspiewak: Yeah `Math.pow` is what I was thinking of.

PR #3470, opened 2018-02-12
Fix LP long compilation times

nicflores:
Avoid comparing fully substituted LPs for equality.

Done qz-3604

### DIFF
--- a/sql/src/main/scala/quasar/sql/compiler.scala
+++ b/sql/src/main/scala/quasar/sql/compiler.scala
@@ -25,12 +25,11 @@ import quasar.contrib.shapeless._
 import quasar.common.SortDir
 import quasar.fp._
 import quasar.frontend.logicalplan.{LogicalPlan => LP, Let => LPLet, _}
-import quasar.std.StdLib
-import StdLib._
+import quasar.std.StdLib, StdLib._
 import quasar.std.TemporalPart
-import quasar.sql.{SemanticAnalysis => SA}
-import SA._
+import quasar.sql.{SemanticAnalysis => SA}, SA._
 import quasar.RenderTree
+// import quasar.RenderTree.ops._
 
 import matryoshka._
 import matryoshka.data._
@@ -257,30 +256,26 @@ final class Compiler[M[_], T: Equal]
       CIName("shift_array")             -> structural.ShiftArray,
       CIName("meta")                    -> structural.Meta)
 
-    def compileDistinct(t: T) = {
+    def compileDistinct(t: T) =
       CompilerState.freshName("distinct") ∘ (name =>
         lpr.let(name, t, agg.Arbitrary(set.GroupBy(lpr.free(name), lpr.free(name)).embed).embed))
-    }
 
     def compileCases
       (cases: List[Case[CoExpr]], default: T)
-      (f: Case[CoExpr] => M[(T, T)]) = {
+      (f: Case[CoExpr] => M[(T, T)]) =
       cases.traverse(f).map(_.foldRight(default) {
         case ((cond, expr), default) =>
           relations.Cond(cond, expr, default).embed
       })
-    }
 
     @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
     def flattenJoins(term: T, relations: SqlRelation[CoExpr]):
-        T = {
-      relations match {
+        T = relations match {
         case _: NamedRelation[_]             => term
         case JoinRelation(left, right, _, _) =>
           structural.MapConcat(
             flattenJoins(Left.projectFrom(term), left),
             flattenJoins(Right.projectFrom(term), right)).embed
-      }
     }
 
     // TODO: Customer Orders starting point
@@ -293,40 +288,30 @@ final class Compiler[M[_], T: Equal]
         Map[String, List[JoinDir]] = {
       @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
       def loop(rel: SqlRelation[CoExpr], acc: List[JoinDir]):
-          Map[String, List[JoinDir]] = {
-        rel match {
-          case t: NamedRelation[_] => {
-            val x = Map(t.aliasName -> acc)
-            x
-          }
-          case JoinRelation(left, right, tpe, clause) => {
+          Map[String, List[JoinDir]] = rel match {
+        case t: NamedRelation[_] => Map(t.aliasName -> acc)
+        case JoinRelation(left, right, tpe, clause) =>
             loop(left, Left :: acc) ++ loop(right, Right :: acc)
-          }
-        }
       }
 
       loop(relations, Nil)
     }
 
     def compileTableRefs(joined: T, relations: SqlRelation[CoExpr]):
-        Map[String, T] = {
-      val x = buildJoinDirectionMap(relations).map {
-        case (name, dirs) => {
+        Map[String, T] =
+      buildJoinDirectionMap(relations).map {
+        case (name, dirs) =>
           name -> dirs.foldRight(
             joined)(
             (dir, acc) => dir.projectFrom(acc))
-        }
       }
-      x
-    }
 
     def tableContext(joined: T, relations: SqlRelation[CoExpr]):
-        TableContext[T] = {
+        TableContext[T] =
       TableContext(
         Some(joined),
         () => flattenJoins(joined, relations),
         compileTableRefs(joined, relations))
-    }
 
     def step(relations: SqlRelation[CoExpr])
         : (Option[M[T]] => M[T] => M[T]) = {
@@ -343,7 +328,7 @@ final class Compiler[M[_], T: Equal]
             case r                               => BindingContext[T](Map())
           }
           next2    <- CompilerState.contextual(bc, tableContext(lpr.free(stepName), relations))(next)
-        } yield {lpr.let(stepName, current, next2)}
+        } yield lpr.let(stepName, current, next2)
       }.getOrElse(next)
     }
 
@@ -363,9 +348,8 @@ final class Compiler[M[_], T: Equal]
     }
 
     def compileFunction[N <: Nat](func: GenericFunc[N], args: Func.Input[CoExpr, N]):
-        M[T] = {
+        M[T] =
       args.traverse(compile0).map(func.applyGeneric(_).embed)
-    }
 
     def buildRecord(names: List[Option[String]], values: List[T]):
         T = {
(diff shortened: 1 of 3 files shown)


alissapajer:
`havingWithMultipleProjections.test` is failing on various connectors, looks like qz-3686.

alissapajer:
@nicflores can we add the test that was taking forever to compile as an integration test? Our integration test timeout is 5 min, so we'll at least know the test is completing in that amount of time.

Also ping @edmundnoble, this is a test we should add to the performance suite.

nicflores:
@alissapajer, good point! I'll add that Customer Orders query as a test and also add the query to benchmarking repo.

nicflores:
@alissapajer, I've got a test in for the Customer Orders query, but still dealing with build issues. Also, I've submitted a PR to benchmarking repo containing the query: https://github.com/slamdata/slamdata-benchmarks/pull/27

it/src/main/resources/tests/regression/bug3686.test
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "[reg] bug3686",
+  "data": ["customers.data", "orders.data", "ordered_items.data", "inventory_items.data"],
+  "backends": {
+      "couchbase":         "skip",
+      "marklogic_json":    "ignoreFieldOrder",
+      "mimir":             "pendingIgnoreFieldOrder"
```
alissapajer: Can you open an issue in youtrack reporting that this query isn't working in mimir? I want to see what's going on with that. No problem merging with this pending though.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -740,44 +744,70 @@ object Compiler {
...
+     R: RenderTree[T]): T = {
     // Step 0: identify key expressions, and rewrite them by replacing the
     // group source with the source at the point where they might appear.
-    def keysƒ(t: LP[(T, List[T])]):
-        (T, List[T]) =
-    {
-      @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
-      def groupedKeys(t: LP[T], newSrc: T): Option[List[T]] = {
-        t match {
+    def keysƒ(t: LP[(T, List[(T, ZFree[LP, Unit])])]): (T, List[(T, ZFree[LP, Unit])]) = {
```
alissapajer: Could you make a type alias for `(T, ZFree[LP, Unit])`? That's a relatively minor thing, but I think it would make the code a lot more readable, since you use that type a few times here and below.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -740,44 +744,70 @@ object Compiler {
...
 
-      (t.map(_._1).embed,
-        groupedKeys(t.map(_._1), t.map(_._1).embed).getOrElse(t.foldMap(_._2)))
+      val tf = t.map(_._1)
+      (tf.embed, groupedKeys(tf, tf.embed).getOrElse(t.foldMap(_._2)))
     }
 
-    // use `scalaz.IList` so we can use `scalaz.Equal[LP]`
-    val keys: IList[T] = IList.fromList(boundCata(tree)(keysƒ)._2)
+    val sources: List[(T, ZFree[LP, Unit])] = tree.cata(keysƒ)._2
```
alissapajer: I might be missing something, but it seems that, throughout this algorithm, instead of `ZFree[LP, Unit]` you could have `Option[LP[T]]`. Would that work?

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -17,24 +17,26 @@
...
 import quasar.common.SortDir
 import quasar.fp._
-import quasar.fp.binder._
-import quasar.frontend.logicalplan.{LogicalPlan => LP, _}
+import quasar.frontend.logicalplan.{LogicalPlan => LP, Let => LPLet, _}
 import quasar.std.StdLib, StdLib._
 import quasar.std.TemporalPart
 import quasar.sql.{SemanticAnalysis => SA}, SA._
+import quasar.RenderTree
+// import quasar.RenderTree.ops._
```
alissapajer: Leftover debug line.

sql/src/main/scala/quasar/sql/compiler.scala
wemrysi: @alissapajer Mind elaborating? I don't understand how we'd use one in place of the other.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -740,44 +744,70 @@ object Compiler {
...
+              if (lp.embed ≟ src) CoEnv[Unit, LP, T](-\/(()))
+              else CoEnv[Unit, LP, T](\/-(lp))
+            }) strengthL newSrc
         }
       }
 
-      (t.map(_._1).embed,
-        groupedKeys(t.map(_._1), t.map(_._1).embed).getOrElse(t.foldMap(_._2)))
+      val tf = t.map(_._1)
+      (tf.embed, groupedKeys(tf, tf.embed).getOrElse(t.foldMap(_._2)))
```
alissapajer: Can we just have `def groupedKeys(t: LP[T]): ...` since we're only calling it with `tf` and `tf.embed`?

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -740,44 +744,70 @@ object Compiler {
...
+
+    val KS = MonadState_[State[KeyState, ?], KeyState]
+
+    def makeKey(tree: T, flp: ZFree[LP, Unit]): T =
+      flp.cata(interpret[LP, Unit, T](_ => tree, _.embed))
 
     // Step 1: annotate nodes containing the keys.
-    val ann: Cofree[LP, Boolean] = boundAttribute(tree)(keys.element)
+    val ann: State[KeyState, Cofree[LP, Boolean]] = tree.transAnaM {
+      case let @ LPLet(name, expr, body) =>
```
alissapajer: `body` is unused so it can be `_`.

sql/src/main/scala/quasar/sql/compiler.scala
alissapajer: @wemrysi maybe a better way to phrase my question is the following: do we actually have a `LP` with `Unit`s at the leaves, warranting the `Free` structure? Or do we have `LP[T] \/ Unit`, meaning we're effectively only using one layer of the `Free` structure. Thinking it could be the later is what prompted my original question. 

But as I look at the code again, when a group key equals the source of the `GroupBy`, we replace the group key with `Unit` to create the `Free` structure. Then when we're in the context of a `Let` binding we replace both the grouping source and the grouping key with a free variable name, when we can (and also annotate the nodes). Then lastly we'll insert `Arbitrary`s, which is basically the point of this all.

Does this mean we don't insert free variables inside group keys during the logical plan logic that creates `Let` bindings?

wemrysi: @alissapajer We do insert the free variables inside group keys when constructing the LP, what we're doing in `keysƒ` is finding the key expression _up to_ the `src` of the `GroupBy`.

The `GroupBy` will look something like
```
Invoke(GroupBy, Sized(Free('tmp0), MakeArrayN(MapProject(Free('tmp0), "foo")))
```
and so, in this case, we'll extract a single key that looks like (modulo `Free` boilerplate) `MapProject((), "foo") : ZFree[LP, Unit]` and pair it with the original `GroupBy` tree to get a `(T, ZFree[LP, Unit])`.

Then when we're annotating, we do a limited form of substitution, just for group key sources. So if we see a `Let` where the bound expression equals the first component of the tuple for one of our keys, we replace it with the free variable. This is because any references to the original `GroupBy` will be through the free variable in the body of the `Let`. Thus we can substitute the same free variable in the `ZFree[LP, Unit]` we saved from earlier to obtain a "concrete" `T` that represents how the group key expression would look inside the `Let` body and search for expressions equal to it in order to identify group keys.

it/src/main/resources/tests/regression/bug3686.test
nicflores: @alissapajer, I checked and the query does work in mimir, I assumed it wouldn't since `bug2228.test` is also excluding `mimir`. The only difference in the data returned is that `mimir` returns `2008-01-16T10:28:32Z[UTC]` and the mongo's don't include the `[UTC]` part. Is there a way to drop the UTC in `mimir` dates? Or add it to the mongo dates?

alissapajer: Oh, that's annoying. Good to hear it basically works for mimir though :) Can you leave a `"NB:"` in this test saying that that's why it's not working for mimir? The test expectation you have is correct (the trailing `"Z"` is included in the original data so we want to include it in the results). 

The plan is to resolve mimir's datetime handling after next-major lands, and I don't want to touch anything datetime related on master. So we'll have to keep it pending until we resolve all that for mimir.

sql/src/main/scala/quasar/sql/compiler.scala
alissapajer: Aha, ok cool. 👍 That last bit in your explanation ("how the group key expression would look") clarifies a lot.

nicflores:
@alissapajer, there we go, finally got a successful build.

sql/src/main/scala/quasar/sql/compiler.scala
```diff
@@ -740,44 +743,71 @@ object Compiler {
...
+        for {
+          ks <- KS.get
+          (srcs, keys) = ks
+          srcs2 = srcs.map { case (t, flp) =>
+            val l = if (t ≟ expr) Free[T](name).embed else t
+            (l, flp)
+          }
+          keys2 = srcs2.map { case (t, flp) => makeKey(t,flp) }
+          _ <- KS.put((srcs2, keys2))
+        } yield EnvT((keys.element(let.embed), let: LP[T]))
```
alissapajer: Should this be `keys` or `keys2`? I haven't yet been able to convince myself one way or another, so I wanted to bring it to attention.

alissapajer:
Aside from my last question, LGTM!

Also waiting on word from @beckyconning regarding comments in qz-3686. Since there is a workaround, and previously `having` was only tractable on small data sets anyway, and this fix is important, I'd like to land this. But I want to make sure business is aware of the regression and knows what the workaround is.

sql/src/main/scala/quasar/sql/compiler.scala
wemrysi: Technically, it probably doesn't matter as the modifications made in `keys2` should never affect equality with the `Let`, however the intent is best described with `keys`. `keys2` has all sources equal to `expr` substituted with `Free(name)` and is meant to be the state used by the child nodes of the `Let`.

beckyconning:
I think merging this and raising this bug is the best way forwards. See https://slamdata.myjetbrains.com/youtrack/issue/qz-3686#comment=4-19927 for more info.

alissapajer:
Closing qz-3604 manually.

FYI @nicflores the correct syntax to auto-close an issue is `#qz-3604 Done`

PR #3471, opened 2018-02-13
Fix range

edmundnoble:
This should work, but I can't run tests locally because of some weird bugs.

### DIFF
--- a/frontend/src/test/scala/quasar/std/SimplifyStdLibSpec.scala
+++ b/frontend/src/test/scala/quasar/std/SimplifyStdLibSpec.scala
@@ -39,39 +39,41 @@ class SimplifyStdLibSpec extends StdLibSpec {
   val notHandled: Result \/ Unit = Skipped("not simplified").left
 
   def shortCircuit[N <: Nat](func: GenericFunc[N], args: List[Data]): Result \/ Unit = (func, args) match {
-    case (relations.Between, _) => notHandled
+      case (relations.Between, _) => notHandled
 
-    case (relations.Eq, List(Data.Date(_), Data.Timestamp(_))) => notHandled
-    case (relations.Lt, List(Data.Date(_), Data.Timestamp(_))) => notHandled
-    case (relations.Lte, List(Data.Date(_), Data.Timestamp(_))) => notHandled
-    case (relations.Gt, List(Data.Date(_), Data.Timestamp(_))) => notHandled
-    case (relations.Gte, List(Data.Date(_), Data.Timestamp(_))) => notHandled
-    case (relations.IfUndefined, _) => notHandled
+      case (relations.Eq, List(Data.Date(_), Data.Timestamp(_))) => notHandled
+      case (relations.Lt, List(Data.Date(_), Data.Timestamp(_))) => notHandled
+      case (relations.Lte, List(Data.Date(_), Data.Timestamp(_))) => notHandled
+      case (relations.Gt, List(Data.Date(_), Data.Timestamp(_))) => notHandled
+      case (relations.Gte, List(Data.Date(_), Data.Timestamp(_))) => notHandled
+      case (relations.IfUndefined, _) => notHandled
 
-    case (date.ExtractCentury, _) => notHandled
-    case (date.ExtractDayOfMonth, _) => notHandled
-    case (date.ExtractDecade, _) => notHandled
-    case (date.ExtractDayOfWeek, _) => notHandled
-    case (date.ExtractDayOfYear, _) => notHandled
-    case (date.ExtractEpoch, _) => notHandled
-    case (date.ExtractHour, _) => notHandled
-    case (date.ExtractIsoDayOfWeek, _) => notHandled
-    case (date.ExtractIsoYear, _) => notHandled
-    case (date.ExtractMicroseconds, _) => notHandled
-    case (date.ExtractMillennium, _) => notHandled
-    case (date.ExtractMilliseconds, _) => notHandled
-    case (date.ExtractMinute, _) => notHandled
-    case (date.ExtractMonth, _) => notHandled
-    case (date.ExtractQuarter, _) => notHandled
-    case (date.ExtractSecond, _) => notHandled
-    case (date.ExtractWeek, _) => notHandled
-    case (date.ExtractYear, _) => notHandled
-    case (date.Now, _) => notHandled
+      case (date.ExtractCentury, _) => notHandled
+      case (date.ExtractDayOfMonth, _) => notHandled
+      case (date.ExtractDecade, _) => notHandled
+      case (date.ExtractDayOfWeek, _) => notHandled
+      case (date.ExtractDayOfYear, _) => notHandled
+      case (date.ExtractEpoch, _) => notHandled
+      case (date.ExtractHour, _) => notHandled
+      case (date.ExtractIsoDayOfWeek, _) => notHandled
+      case (date.ExtractIsoYear, _) => notHandled
+      case (date.ExtractMicroseconds, _) => notHandled
+      case (date.ExtractMillennium, _) => notHandled
+      case (date.ExtractMilliseconds, _) => notHandled
+      case (date.ExtractMinute, _) => notHandled
+      case (date.ExtractMonth, _) => notHandled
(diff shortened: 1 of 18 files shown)


alissapajer:
Is there a youtrack issue for this?

edmundnoble:
@alissapajer Yep, https://slamdata.myjetbrains.com/youtrack/issue/qz-2625.

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
```diff
@@ -391,6 +391,90 @@ trait SliceTransforms[M[+ _]] extends TableModule[M] with ColumnarTableTypes[M]
...
+          import java.util.Arrays
+          import scala.collection.mutable
+          import scala.annotation.tailrec
+          composeSliceTransform2(upper).zip(composeSliceTransform2(lower)) { (upperS, lowerS) =>
+            val upperColumns = upperS.materialized.columns
+            val lowerColumns = lowerS.materialized.columns
+            val lowerC = lowerColumns.getOrElse(ColumnRef(CPath.Identity, CLong), {
+              val numCol = lowerColumns(ColumnRef(CPath.Identity, CNum)).asInstanceOf[ArrayNumColumn]
+              new ArrayLongColumn(numCol.defined, numCol.values.map(_.toLong))
+            }).asInstanceOf[ArrayLongColumn]
```
djspiewak: Misses the case where we have doubles which are coincidentally integers.

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
```diff
@@ -391,6 +391,90 @@ trait SliceTransforms[M[+ _]] extends TableModule[M] with ColumnarTableTypes[M]
...
+                (arrs, defineds) match {
+                  case (a :: as, d :: ds) => 
+                    val ref = ColumnRef(CPath(CPathIndex(i) :: Nil), CLong)
+                    val column = new ArrayLongColumn(d, a)
+                    val newAccum = accum + (ref -> column)
+                    rec(i + 1, as, ds, newAccum)
+                  case _ => accum
+                }
+              rec(0, arrs, defineds, Map.empty)
+            }
```
djspiewak: All of the above is actually quite a bit more complicated than I was expecting.  Building the bitsets should be the only real work that you're doing here.  The columns are trivial, because each column will simply be a constant of the same value repeated over and over again!  The first constant column will be the value from `lowerA` (which will be a constant in the only value(s) for which it is defined), while the last constant column will be the value from `upperA`, with each other column incrementing between.  The bitsets you build now mask out the appropriate values, though about half that code can be cut out if you're only building bitsets rather than bitsets *and* arrays.  The result should be considerably faster.

Obviously this only works because the evaluator is megamorphic, but all of this needs to be rewritten anyway when that change comes to pass.

edmundnoble: Each column is not a constant of the same value repeated over and over again.
I showed you sample results for `[7, 8, 9][*] .. 10`:
```
[0][0] | [0][1] | [0][2] | [0][3] |
--------|--------|--------|--------|
      7 |      8 |      9 |     10 |
      8 |      9 |     10 |        |
      9 |     10 |        |        |
```
With this in mind I don't understand how reusing a single array makes any sense at all. Each column here is a shifted version of lowerA with values masked out when the range has been fully enumerated.

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
edmundnoble: This was on purpose, why is it incorrect?

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
djspiewak: Oh I’m a derp. You’re correct. My way would work, but only after compaction. Your method precompacts. I’ll reconsider in that light.

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
djspiewak: Because `(2.4+1.6)..6` is a valid range. Since we have auto magical numeric coercion, we have to be careful to be consistent about it.

edmundnoble: Actually, I think we don't have auto magical numeric coercion. The typer doesn't do it. `substring("hello", 0, 5.0)` for example doesn't type check.

djspiewak: We should talk to John about this, but the design *was* to have magical numeric coercion.

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
```diff
@@ -391,6 +391,90 @@ trait SliceTransforms[M[+ _]] extends TableModule[M] with ColumnarTableTypes[M]
...
+              new ArrayLongColumn(numCol.defined, numCol.values.map(_.toLong))
+            }).asInstanceOf[ArrayLongColumn]
+            val lowerA = lowerC.values
+            val upperA = upperC.values
+            val minNumRows = 
+              if (lowerA.length > upperA.length) upperA.length 
+              else lowerA.length
+            val maxNumRows =
+              if (lowerA.length < upperA.length) upperA.length 
+              else lowerA.length
```
djspiewak: One last thought (aside from talking to John about numeric coercion)…  Should we be doing a check to ensure that we're not trying to create millions of columns?  Basically, is it worth tossing a sanity bound on things so that people don't try to create a range `Long.MinValue .. Long.MaxValue`?

edmundnoble: I guess that makes sense, do we do this on other operations? I'd also suggest making sure that there are no columns with size `Int.MaxValue`, because array tings.

djspiewak: Yeah there are several places where we *should* be doing this but are not right now.  I mostly just don't want to add to that list.  :-)

edmundnoble: Oh so actually the number of columns created is not directly relevant to the size of the range. It's actually relevant to the number of columns in the *inputs*. I think it makes sense not to bound this because we'd already have run into an issue if we had way too many columns.

I will still bound the maximum range bound difference though.

yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala
```diff
@@ -391,6 +391,90 @@ trait SliceTransforms[M[+ _]] extends TableModule[M] with ColumnarTableTypes[M]
             }
           }
 
+        case Range(lower, upper) =>
+          import java.util.Arrays
+          import scala.collection.mutable
+          import scala.annotation.tailrec
```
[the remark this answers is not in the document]
edmundnoble: Done.

frontend/src/test/scala/quasar/std/SimplifyStdLibSpec.scala
```diff
@@ -39,39 +39,42 @@ class SimplifyStdLibSpec extends StdLibSpec {
...
-    case (date.ExtractSecond, _) => notHandled
-    case (date.ExtractWeek, _) => notHandled
-    case (date.ExtractYear, _) => notHandled
-    case (date.Now, _) => notHandled
-
-    case (structural.MapProject, List(Data.Obj(fields), Data.Str(field))) if !fields.contains(field) => notHandled
-
-    case _ => ().right
-  }
+      case (relations.Between, _) => notHandled
```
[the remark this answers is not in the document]
edmundnoble: Done.

djspiewak:
I'm still concerned about the lack of double coercion, but we'll have to loop back to that.

PR #3473, opened 2018-02-14
Fix bug ReifyAutoJoins and add unit tests.

alissapajer:
#qz-3687 Done

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/ReifyAutoJoins.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/ReifyAutoJoins.scala
@@ -98,10 +98,10 @@ final class ReifyAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends QSU
             qsu.qsAutoJoin(l, c, lcKeys, lcCombiner)
 
           def projLeft[A](hole: FreeMapA[A]): FreeMapA[A] =
-            func.ProjectKey(StrLit(lName), hole)
+            func.ProjectKey(hole, StrLit(lName))
 
           def projCenter[A](hole: FreeMapA[A]): FreeMapA[A] =
-            func.ProjectKey(StrLit(cName), hole)
+            func.ProjectKey(hole, StrLit(cName))
 
           val combiner: JoinFunc = combiner3 flatMap {
             case LeftSide3 => projLeft[JoinSide](LeftSideF)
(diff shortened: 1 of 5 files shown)


connector/src/test/scala/quasar/qscript/qsu/ReifyAutoJoinsSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+          fmC must beTreeEqual(
+           func.ProjectKeyS(func.Hole, "bar"))
+
+          fmR must beTreeEqual(
+           func.ProjectKeyS(func.Hole, "baz"))
+
+          fmInner must beTreeEqual(
+           func.ConcatMaps(
+             func.MakeMap(func.Constant(J.str("leftAccess1")), func.LeftSide),
+             func.MakeMap(func.Constant(J.str("centerAccess2")), func.RightSide)))
```
alissapajer: Hard coding these names is kinda brittle. I could match on the `FreeMap` and extract the names, and then check that they're the same ones projected out below, if you think that's worth it.

djspiewak: We tend to do this in a bunch of places. Sadly, I don’t know if there’s a less brittle alternative right now.

connector/src/main/scala/quasar/qscript/qsu/ReifyAutoJoins.scala
```diff
@@ -98,10 +98,10 @@ final class ReifyAutoJoins[T[_[_]]: BirecursiveT: EqualT] private () extends QSU
             qsu.qsAutoJoin(l, c, lcKeys, lcCombiner)
 
           def projLeft[A](hole: FreeMapA[A]): FreeMapA[A] =
-            func.ProjectKey(StrLit(lName), hole)
+            func.ProjectKey(hole, StrLit(lName))
 
           def projCenter[A](hole: FreeMapA[A]): FreeMapA[A] =
-            func.ProjectKey(StrLit(cName), hole)
+            func.ProjectKey(hole, StrLit(cName))
```
djspiewak: Why not just `ProjectKeyS`?

connector/src/test/scala/quasar/qscript/qsu/ReifyAutoJoinsSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+import quasar.qscript.{construction, MapFuncsCore}
+import quasar.qscript.provenance.JoinKeys
+import quasar.Planner.PlannerError
+
+import matryoshka.data.Fix
+import pathy.Path
+import pathy.Path.Sandboxed
+import scalaz.{EitherT, INil, Need, StateT}
+
+object ReifyAutoJoinSpecs extends Qspec with TreeMatchers with QSUTTypes[Fix] {
```
djspiewak: Woot!

PR #3474, opened 2018-02-14
Remove Expires http response header

rintcius:
#qz-3657 Done

### DIFF
--- a/web/src/main/scala/quasar/api/VCacheMiddleware.scala
+++ b/web/src/main/scala/quasar/api/VCacheMiddleware.scala
@@ -16,15 +16,10 @@
 
 package quasar.api
 
-import slamdata.Predef._
 import quasar.effect.Timing
 import quasar.fs.mount.cache.VCache, VCache.VCacheExpR
 
-import org.http4s.Header
-import org.http4s.headers.Expires
-import org.http4s.util.Renderer
-import scalaz._, Scalaz._
-import scalaz.syntax.tag._
+import scalaz._
 
 object VCacheMiddleware {
   def apply[S[_]](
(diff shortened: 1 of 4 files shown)


web/src/main/scala/quasar/api/VCacheMiddleware.scala
```diff
@@ -34,13 +29,8 @@ object VCacheMiddleware {
...
-    QHttpService { case req =>
-      (service(req) ⊛ R.ask ⊛ T.timestamp) { case (resp, ex, ts) =>
-        val cacheHeaders = ex.unwrap.foldMap(e =>
-          Header(Expires.name.value, Renderer.renderString(e.v)) ::
-          ts.isAfter(e.v).fold(List(StaleHeader), Nil))
-
-        resp.modifyHeaders(_ ++ cacheHeaders)
-      }
-    }
+    //Due to qz-3657 cache specific response headers have been removed for the moment.
```
wemrysi: I think we should just delete this file. It is easy enough to recover from `git` history if we ever need to revive it, but until then it is just dead code.

rintcius: It's easy to just delete this file, but the reason I didn't do this is that it looks like there's more unused code than just this file (not sure about this though, first time I'm looking at this code). 

I just felt that the time needed to go through everything & decide what's exactly unused is too large compared to reimplementing it with ETags. If it's just this file though I'm happy to delete this.

wemrysi: If you're interested in preserving the existing impl and just disabling the headers, you could restore this file and instead delete the [line applying this middleware](https://github.com/quasar-analytics/quasar/blob/785adf54d480ddce417678b2525a80b5017b1488/web/src/main/scala/quasar/api/services/RestApi.scala#L76).

rintcius: yeah, I choose this option because this also preserves the idea of the tests, but at the cost of some runtime overhead. 

I will change that line that you mentioned though.

rintcius:
@wemrysi ah, you are right - I have reinstated them

wemrysi:
I've restarted the appveyor build, will merge when green.

PR #3475, opened 2018-02-16
Disallow mounting mimir on relative paths

edmundnoble:
I did this by accident from the runner and that was *very* odd. @djspiewak says this is never a valid thing to do in general.

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/Mimir.scala
+++ b/mimir/src/main/scala/quasar/mimir/Mimir.scala
@@ -105,8 +105,11 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
     O.optimize(reflNT[QSM[T, ?]])
   }
 
-  def parseConfig(uri: ConnectionUri): BackendDef.DefErrT[Task, Config] =
-    Config(new java.io.File(uri.value)).point[BackendDef.DefErrT[Task, ?]]
+  def parseConfig(uri: ConnectionUri): BackendDef.DefErrT[Task, Config] = {
+    val file = new java.io.File(uri.value)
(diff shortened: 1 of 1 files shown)


edmundnoble:
A session where I attempt to mount to a relative path:
```
$ curl -v -X PUT http://localhost:8080/mount/fs/mimir/target/ -d '{ "mimir": { "connectionUri":"tmp/precog/" } }'
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> PUT /mount/fs/mimir/target/ HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 46
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 46 out of 46 bytes
< HTTP/1.1 400 Invalid mount configuration.
< Content-Type: application/json; charset=UTF-8
< Date: Fri, 16 Feb 2018 19:14:10 GMT
< Content-Length: 132
<
* Connection #0 to host localhost left intact
{ "error": { "status": "Invalid mount configuration.", "detail": { "reasons": [ "Mimir cannot be mounted to a relative path" ] } } }%
```

alissapajer:
Needs version label.

edmundnoble:
Failure looks spurious.

djspiewak:
@edmundnoble For the record, when there's a spurious failure, you should restart the job (after noting down the error and ensuring a ticket is creating).  We generally don't want to merge PRs with red builds; if the failure truly is spurious, then the restart will make it go green, and the ticket will track the work to move to a more stable build.

PR #3476, opened 2018-02-16
Error `having` in sql parser.

alissapajer:
#qz-3692 Done

The single integration test we have that uses `having` (`havingWithMultipleProjections.test`) is already pending, with a note to qz-3686.


alissapajer:
Fixing a test error.

PR #3478, opened 2018-02-20
This doesn't show up locally, weirdly, but it should fix travis

djspiewak:


mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -102,11 +102,11 @@ class PlannerSpec extends
         beRight.which(cwf => notBrokenWithOps(cwf.op, IList(ReadOp, MatchOp, ProjectOp, MatchOp, GroupOp, ProjectOp)))
     }
 
-    trackPending(
+    /*trackPending(
```
alissapajer: Can you add `// TODO qz-3686` here as well.

alissapajer:
How did this make it through Travis in my PR?

alissapajer:
Do we blame `hoarder`?

alissapajer:
I think the contents of `pending` tests are inconsistently compiled or not compiled.

djspiewak:
@alissapajer Originally I was going to blame hoarder, but it fully compiles (after `clean`!) on my local machine as well, so I'm perplexed.

alissapajer:
I wonder if there is some test or runner non-referential transparency getting exposed because `trackPending` is a function.

djspiewak:
@alissapajer That's just it though: this is *compilation*. The only thing I can think of is some macro bizarreness, since `sqlE` is indeed a macro, but even then it should have been caught on clean.

alissapajer:
In my PR I actually saw this same thing, where the appveyor build failed to compile a different `sqlE` macro, and Travis compiled it. If the overhead of repeating a `clean` weren't so high locally, I'd be curious if this is nondeterministic locally.

alissapajer:
Rather than a macro I'd prefer a test util function that parses the sql and returns valid sql or a test error, instead of having this be a compile error. But that's not for this PR.

PR #3479, opened 2018-02-20
Quasar server no longer waits for input but simply runs forever instead

jedesah:
- `sbt` now forks a new jvm process to run the server which allows developers to kill Quasar using a command such as the `kill` Unix command. Hopefully `sbt` one day supports killing forked processes without having to lookup the process PID in another tab and `kill` it (which is the current recommendation for developers)
- Add a new stage `testPublish` to Travis which makes sure we can generate scaladoc and generate an assembly jar as well as run `testJar` to make sure all PRs that build cleanly are fit for publishing


jedesah:
Honestly, there are enough changes in `travis.yml` that I don't completely grok that I'd be surprised if this worked on the first try and passed everyone's scrutiny but got to start somewhere!

jedesah:
And of course the `travis.yml` diff is terrible...

djspiewak:
AppVeyor failure is due to the `having` issue, addressed in #3478

interface/src/main/scala/quasar/main/package.scala
```diff
@@ -499,10 +499,11 @@ package object main extends Logging {
 
   /** Either initialize the metastore or execute the start depending
     * on what command is provided by the user in the command line arguments
+    * @param start The body of logic, should return true whether to shutdown quasar or not
```
edmundnoble: I think you can remove "whether" and "or not" here.

edmundnoble: More importantly, why is this returning a boolean when we're always yielding `false`?

djspiewak:
Needs merge with master.

nicflores:
@jedesah, should we have a message saying run `kill <pid>` to go back to `sbt`?
Also I just built a web assembly jar and noticed that there's no message on how to exit, should we say something like `Ctrl-C to exit`?

.travis.yml
```diff
@@ -139,6 +142,16 @@ jobs:
 
         - scripts/testJar
```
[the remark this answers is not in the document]
jedesah: The one on line 135 isn't sufficient? I think this is just a terrible diff thing 😄

interface/src/main/scala/quasar/main/package.scala
jedesah: The `repl` returns `true`. Agreed on the wording improvement though

.travis.yml
```diff
@@ -139,6 +142,16 @@ jobs:
```
jedesah: Does anyone know why `./sbt -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION doc web/assembly` is between `'`? I guess I'll look it up in the Travis docs but in case someone knows off the top of their head

jedesah:
Not sure I understand the error message Travis is reporting on the failed `Testpublish` stage.

If the command exited with 0 then why did the build fail? `testJar` didn't run so obviously something went wrong but what is it the previous command finished successfully?

djspiewak:
Where is revolver?

jedesah:
@djspiewak I've tested with a globally installed revolver and it works as desired. I was going to wait for this PR to get merged and encourage others to install revolver and use it. I don't know if it's necessary to add revolver to the project build since individual developers can choose to use it or not and those who do can use it for all their projects. If there were some specific settings for quasar or something then I would have added it to the build, but since it's just a "productivity" tool, we've elected not to include such things in the project build in the past.

That being said, if there is a preference for it being part of the build for discoverability purposes or in anticipation of custom configuration, I woudn't object to adding it to the build.

djspiewak:
@jedesah Globally-installed plugins are *exceptionally* dangerous. The real question is whether or not the build continues to work without it, and the nature of global plugins always makes that a difficult question to answer. Forking in `run` is the *correct* thing to do in a vacuum (if only because of the resource issues), which is the litmus test for me: if a global plugin causes you to do something which is otherwise-wrong in a local build, then stop using the global plugin (or just make it part of the build).

jedesah:
@djspiewak If I wasn't convinced by your argument against global sbt plugins in general (I was semi-convinced, would like to discuss next time we are both in the office for lunch), I just found a practical problem which is that `revolver` does not seem to be playing nicely with `sideEffectTestFSConfig` which either means `revolver` just isn't an option or it might need some custom config which would necessitate `revolver` being part of the build. I'll have a look at whether I can make the two work well together. Also, now I'm wondering if `sbt` `fork := true` was working well with `sideEffectTestFSConfig`, I don't remember if I had tested that or not.

djspiewak:
Closing this, since quasar isn't going to be runnable at all after the new fs lands.

PR #3481, opened 2018-02-21
Nonmappable join condition

jsantos17:
#qz-3286 done

Still missing:

- [x] Refactor cases in `LPJoin` to reduce duplication
- [x] Refactor `UnifyTargets` to return a `Map[Symbol, FreeMap]` instead of matching target symbol and target `FreeMap`s positionally in `NonEmptyList`s
- [x] Write unit tests


it/src/main/resources/tests/qa/s07_join/nonJsJoinCondition.test
```diff
@@ -3,11 +3,6 @@
 
     "backends": {
         "couchbase":         "pending",
-        "marklogic_json":    "pending",
```
jsantos17: Mongo 3.4 is passing. Using Travis to check which conenctors pass/fail.

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -60,34 +61,104 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
...
       }
   }
 
+  private def combiner: JoinFunc =
+    func.StaticMapS(
+      JoinDir.Left.name -> func.LeftSide,
+      JoinDir.Right.name -> func.RightSide)
+
+  private def hasJoinRef(g: QSUGraph, refId: Symbol): Boolean =
+    Tag.unwrap[Boolean, Disjunction](g.foldMapUp(g => g.unfold.map(_.root) match {
```
djspiewak: Another place where you can just match on `g`, which also allows you to phrase this whole thing as a total-partial function (no `match`).

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -97,7 +168,7 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
       buildNode: (Symbol, NonEmptyList[FreeMap]) => QScriptUniform[Symbol]): F[QSUGraph] =
     UnifyTargets[T, F](withName[F](_))(graph, source, targets)(sourceName, targetPrefix) flatMap {
       case (newSrc, original, targetExprs) =>
-        val node = buildNode(newSrc.root, targetExprs)
+        val node = buildNode(newSrc.root, targetExprs.seconds)
```
djspiewak: Wait… what?

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -107,13 +178,50 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
...
+            case -\/(side) => Free.pure(side).some
+            case \/-(g) => g.root match {
+              case sym if map.isDefinedAt(sym) => map.get(sym).map(_.as(reshapeSide))
+              case _ => none
+            }
+          }.point[F]) >>= { on =>
+          val node = buildNode(newSrc.root, on, repair)
+
+          withName[F](node)
+            .map(inter => graph.overwriteAtRoot(node) :++ inter :++ newSrc)
```
djspiewak: What are you trying to do here?  `inter`'s map is empty, so the `:++ inter` is pointless, and `withName` is discarded because you're just taking the `node` and passing it into `overwriteAtRoot` (which uses the root's name).

connector/src/main/scala/quasar/qscript/qsu/ReifyBuckets.scala
```diff
@@ -60,7 +60,7 @@ final class ReifyBuckets[T[_[_]]: BirecursiveT: EqualT: ShowT] private () extend
                         original as access
                     })
 
-                  g.overwriteAtRoot(mkReduce(newSrc.root, buckets, reduce as reduceExpr.head)) :++ newSrc
+                  g.overwriteAtRoot(mkReduce(newSrc.root, buckets, reduce as reduceExpr.seconds.head)) :++ newSrc
```
djspiewak: "What?" again.

connector/src/main/scala/quasar/qscript/qsu/UnifyTargets.scala
```diff
@@ -40,20 +40,23 @@ final class UnifyTargets[T[_[_]]: BirecursiveT, F[_]: Monad] private (
   import QScriptUniform.AutoJoin2
 
   def apply(graph: QSUGraph, source: Symbol, targets: NonEmptyList[Symbol])
-      : F[(QSUGraph, FreeMap, NonEmptyList[FreeMap])] = {
+      : F[(QSUGraph, FreeMap, NonEmptyList[(Symbol, FreeMap)])] = {
```
djspiewak: What are these symbols of?  Also should they be `QSUGraph` instead?

connector/src/test/scala/quasar/qscript/qsu/ExtractFreeMapSpec.scala
```diff
@@ -316,6 +317,93 @@ object ExtractFreeMapSpec extends Qspec with QSUTTypes[Fix] with TreeMatchers {
...
+              projectCustomersKey.as[JoinSide](RightSide)))
+
+          repair must beTreeEqual(
+            makeMap("left", "right") >>= {
+              case LeftSide => func.ProjectKeyS(func.LeftSide, "left_source")
+              case side => Free.pure(side)
+            })
+        }
+      }
+    }
```
djspiewak: I'd really like to see a test here for when we have multiple non-mappable roots.  Something like this:

```sql
  foo 
join
  bar
on
  (foo union bar) +
  (foo.a union foo.b) +
  foo[*] +
  bar[*] +
  (bar.a union bar.b)
```

Unless I'm misunderstanding your implementation, I think that will break.

connector/src/main/scala/quasar/qscript/qsu/ReifyBuckets.scala
jsantos17: `UnifyTargets` used to take a  `NonEmptyList[Symbol]` and return a `NonEmptyList[FreeMap]`. I changed it to tuple the `FreeMaps` with their corresponding symbols. However, `ReifyBuckets` still expect the `NonEmptyList[FreeMap]` so I just ignore the symbols

connector/src/main/scala/quasar/qscript/qsu/UnifyTargets.scala
jsantos17: These are the same symbol passed as the `target` argument. I could make them `QSUGraph`s by refocusing `graph` on each target. Would that be better?

connector/src/test/scala/quasar/qscript/qsu/ExtractFreeMapSpec.scala
jsantos17: Thank for the example! I'll test this!

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -107,13 +178,50 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
...
+          case side if side === reshapeSide => original.as(side)
+          case other => Free.pure(other)
+        }
+
+        val map = SMap(targetMap.toList: _*)
+
+        OptionT(max.map(qg => partialRefReplace(qg, lref.id, rref.id)(qg.root))
+          .traverseM[Option, JoinSide] {
+            case -\/(side) => Free.pure(side).some
+            case \/-(g) => g.root match {
```
wemrysi: It looks like the `match` is unnecessary here, could replace with `map.get(g.root).map(_.as(reshapeSide))`.

connector/src/main/scala/quasar/qscript/qsu/ExtractFreeMap.scala
```diff
@@ -60,34 +61,104 @@ final class ExtractFreeMap[T[_[_]]: BirecursiveT] private () extends QSUTTypes[T
...
+              val rightUnify =
+                UnifyTargets[T, F](withName[F](_))(graph, right.root, rights)("right_source", "right_target")
+
+              (leftUnify |@| rightUnify) {
+                case ((leftGraph, leftOrig, leftMap0), (rightGraph, rightOrig, rightMap0)) => {
+                  val leftMap = SMap(leftMap0.toList: _*)
+                  val rightMap = SMap(rightMap0.toList: _*)
+
+                  val on: Option[JoinFunc] = max.traverseM[Option, JoinSide](qg => qg.root match {
+                    case sym if leftMap.isDefinedAt(sym) => leftMap.get(sym).map(_.as(LeftSide))
```
wemrysi: Similarly here, could replace with
```scala
leftMap.get(sym).map(_.as(LeftSide)) orElse rightMap.get(sym).map(_.as(RightSide))
```

connector/src/main/scala/quasar/qscript/qsu/ReifyBuckets.scala
djspiewak: Oh, gotcha.

connector/src/main/scala/quasar/qscript/qsu/UnifyTargets.scala
djspiewak: Oh I see due to the `targets` parameter.  Ok, that makes sense.  In general, both should have been `QSUGraph` rather than `Symbol`, but it's sort of ok for now.

connector/src/test/scala/quasar/qscript/qsu/ExtractFreeMapSpec.scala
jsantos17: I  arbitrarily replaced a `+` with `=` to make it evaluate to a boolean. The following query compiles in current `master` (and in this branch) and doesn't hit the non-mappable `LPJoin` case in `ExtractFreeMap`.

```
select foo.abc from foo join bar on (foo union bar) + (foo.a union foo.b) + foo[*] = bar[*] + (bar.a union bar.b)
```

However, I was able to break my implementation with `select foo.abc from foo join bar on bar.a + bar.b[*]  = foo.c[*]`

I'll try to figure out why that doesn't work.

it/src/main/resources/tests/qa/s07_join/changedTestStatus.test
```diff
@@ -5,7 +5,7 @@
         "couchbase":         "pending",
         "marklogic_json":    "pending",
         "marklogic_xml":     "pending",
-        "mimir":             "pendingIgnoreFieldOrder",
+        "mimir":             "skip",
         "mongodb_3_2":       "pendingIgnoreFieldOrder",
```
jsantos17: Not sure how these were passing given the query did not compile before this PR.

PR #3483, opened 2018-02-22
Fix PUTting existing cached view

rintcius:
Problem was that when updating a cached view `view_cache.last_update` was still filled with a value from the old entry, whereas `view_cache.refresh_after` was set to a value in the future calculated from `maxAge`. This meant that when checking for `staleCachedViews` (based on the query `SELECT *
FROM view_cache WHERE last_update IS NULL OR ($now > refresh_after)`) these entries were not going to be repaired until somewhere in the future (i.e. a broken cache for this period). They must be repaired immediately instead.

The fix sets `view_cache.refresh_after` to `now` when creating or updating the cache and simplifies the `staleCachedViews`check by stopping to look at `last_update` in the query.

#qz-3694 Done
#qz-3695 Done

### DIFF
--- a/web/src/main/scala/quasar/api/services/mount.scala
+++ b/web/src/main/scala/quasar/api/services/mount.scala
@@ -148,13 +148,13 @@ object mount {
       maxAge =  req.headers.get(`Cache-Control`).flatMap(_.values.list.collectFirst {
                   case `max-age`(s) => s
                 })
-      _      <- (refineType(path).toOption ⊛ MountConfig.viewConfig.getOption(bConf).map(MountConfig.ViewConfig.tupled) ⊛ maxAge)(createNewViewCache[S])
+      _      <- (refineType(path).toOption ⊛ MountConfig.viewConfig.getOption(bConf).map(MountConfig.ViewConfig.tupled) ⊛ maxAge)(createOrUpdateViewCache[S])
                   .getOrElse(().point[EitherT[Free[S, ?], ApiError, ?]])
     } yield exists
 
-  private def createNewViewCache[S[_]](
-    viewPath: AFile, 
-    viewConfig: MountConfig.ViewConfig, 
+  private def createOrUpdateViewCache[S[_]](
+    viewPath: AFile,
+    viewConfig: MountConfig.ViewConfig,
     maxAge: scala.concurrent.duration.Duration
   )(implicit
     MF: ManageFile.Ops[S],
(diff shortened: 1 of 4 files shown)


web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -163,30 +163,33 @@ object mount {
...
+    //   <existing code>
+    // else
+    //   // only update maxAgeSeconds and refreshAfter (relative to maxAgeSeconds)
+    //   old.copy(
+    //     maxAgeSeconds = nw.maxAgeSeconds,
+    //     refreshAfter = old.refreshAfter.plusSeconds(nw.maxAgeSeconds - old.maxAgeSeconds))
+    if (MountConfig.equal.equal(nw.viewConfig, old.viewConfig))
+      // Let's keep the old cacheReads & lastUpdate if toplevel definition
+      // is unchanged
+      nw.copy(
```
rintcius: This is a bit arbitrary (& a remnant of me trying to fix as described in the TODO above, before I realized that we need the fully resolved query of the old view for this)

Alternative is to also return `nw` here, so we can simplify the code. Happy to implement like this if you prefer this functionality.

web/src/main/scala/quasar/api/services/mount.scala
```diff
@@ -163,30 +163,33 @@ object mount {
...
-                            dataFile = newViewCache.dataFile))),
+                          κ(vcache.modify(viewPath, modifyViewCache(newViewCache))),
                           vcache.put(viewPath, newViewCache)).join.liftM[ApiErrT]
     } yield ()
+
+  private def modifyViewCache(nw: ViewCache)(old: ViewCache): ViewCache =
+    // TODO we should not recreate the cache if the ViewConfig hasn't changed
+    // However `if (MountConfig.equal.equal(nw.viewConfig, old.viewConfig))`
+    // only tests whether the toplevel definition has changed.
+    // At the moment we cannot detect any changes in underlying definitions.
```
wemrysi: Hm, are you saying that we can't do this because if a view references another view or module it may have changed? Wouldn't we want to invalidate caches when one of _those_ mounts are updated? Not this one?

Maybe I've misunderstood, but it seems like we can do what you proposed below now.

rintcius: Yes, that's what I mean. 

Your approach looks better but I don't think we currently invalidate caches yet when an underlying definition changes, right? I think we can only activate the code below when this is implemented because in our current implementation activating this would mean that caches don't get invalidated when they should.

E.g. consider the construction for cached views that the frontend uses:
```
PUT view foo: <some query> // the user can change this
PUT cache bar: select * from foo // the user cannot change this definition, it's generated from a cached card, so this is quite a stable definition
```

Whenever a user changes the definition of `foo`, `bar` needs to be invalidated, but this won't happen if we only activate the code below.

rintcius:
@wemrysi Thanks! I have replied to your comment https://github.com/quasar-analytics/quasar/pull/3483#discussion_r170215512 & changed the comment in the code

web/src/main/scala/quasar/api/services/mount.scala
wemrysi: Hm, I think I'm missing something. How would `bar` be invalidated now?

wemrysi: Oh! Do you mean that the frontend will reissue a `PUT` for `bar` everytime `foo` is changed? If so, then I understand the situation and why my proposed change would be worse than the status quo.

rintcius: Yes, that's my understanding - when you change a card in a deck then frontend will recalculate that card and all cards after it in that deck

PR #3487, opened 2018-02-23
Add new mongo versions

nicflores:
Include Mongo 3.4.3, 3.4.4, and 3.6 in build system and testing.

#qz-3569 Done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,7 @@ env:
     - CONNECTOR=mimir
     - CONNECTOR=mongodb_3_2
     - CONNECTOR=mongodb_3_4
-    - CONNECTOR=mongodb_3_4_3
-    - CONNECTOR=mongodb_3_4_4
+    - CONNECTOR=mongodb_3_4_13
     - CONNECTOR=mongodb_3_6
     - CONNECTOR=mongodb_read_only
     # - CONNECTOR=spark_local_test # no spark for 2.12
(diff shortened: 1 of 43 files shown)


edmundnoble:
Rebase on the multiple connector PR and add to the `mounts` file?

nicflores:
@rintcius, I went with MongoDB `v3.4.3`, `v3.4.4`, and `v3.6` (which really runs `v3.6.2`). I'll have you guys take a look and let me know if I'm missing anything.

@jsantos17, feel free to look though this if you have time as well.

rintcius:
@nicflores Right now, it looks like it doesn't yet run integration tests for the added mongo versions. I got triggered because I saw the entries added to travis, but without any changes to the IT `backends` json.

I don't mind about the docker names but in the ITs it would be nice if they can be referred to like this (see https://slamdata.myjetbrains.com/youtrack/issue/qz-3569#comment=4-19561 for context):

```
    "backends": {
        "mongodb_3_2":       "skip",
        "mongodb_3_4":       "skip",
        "mongodb_3_4_4":     "skip",
        "mongodb_3_6":       "skip",
        "mongodb_read_only": "skip"
    },
```

Not sure if it's easy to implement but I think it would be good to fail the build for unknown `CONNECTOR` values instead of silently running without ITs (we can create a separate issue for that if you prefer).

nicflores:
@rintcius, this should be ready for another look.

rintcius:
Looks good! 👍

PR #3488, opened 2018-02-26
Remove Data.Set, good riddance

edmundnoble:
🚽 and it's gone.

#qz-3624 Done
#qz-3713 Done
#qz-1914 Done


edmundnoble:
Oh, and we were planning `select 1 where false` as `1` previously (and a bunch of other hilariously incorrect constant plans) which this fixes.

alissapajer:
@edmundnoble can you check if this issue https://slamdata.myjetbrains.com/youtrack/issue/qz-1914 is fixed by this PR as well?

core/src/test/scala/quasar/fs/ConstantPlansSpec.scala
```diff
@@ -24,10 +24,10 @@ import scalaz.Scalaz._
 
 class ConstantPlansSpec extends quasar.Qspec {
   "asConstant should identify plans that are constant" >> {
-    "literal set with negatives" >> {
-      val expr = sqlE"select max((0, 1, -2, 5)) as maxlit from cars"
```
alissapajer: Why did this test change?

sql/src/test/scala/quasar/sql/compiler.scala
```diff
@@ -66,29 +66,6 @@ class CompilerSpec extends quasar.Qspec with CompilerHelpers {
           "b" -> lpf.constant(Data.Str("abc"))))
     }
 
-    "compile complex constant" in {
-      testTypedLogicalPlanCompile(sqlE"[1, 2, 3, 4, 5][*] limit 3 offset 1",
-        lpf.constant(Data.Set(List(Data.Int(2), Data.Int(3)))))
-    }
```
alissapajer: What do all these compile to now? Should we keep the tests around and change the test expectation?

web/src/test/scala/quasar/api/services/InvokeServiceSpec.scala
```diff
@@ -202,13 +202,13 @@ class InvokeServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s
       }
       "if query in function is constant even if not supported by connector" >>
         prop { (functionFile: PathOf[Abs, File, Sandboxed, AlphaCharacters]) =>
-          val constant = sqlE"select (1,2)"
```
alissapajer: Why the change? Is this no longer valid sql^2?

frontend/src/test/scala/quasar/std/set.scala
```diff
@@ -1,74 +0,0 @@
...
-class SetSpec extends quasar.Qspec with TypeArbitrary {
-  import SetLib._
-  import quasar.Data
-  import quasar.Type
-  import quasar.Type.Const
-
-  "SetLib" should {
-    "type taking no results" in {
-      val expr = Take.tpe(Func.Input2(Type.Int, Type.Const(Data.Int(0))))
-      expr should beSuccessful(Type.Const(Data.Set(Nil)))
```
alissapajer: Again, could we keep these test and change the expectations?

core/src/test/scala/quasar/fs/ConstantPlansSpec.scala
edmundnoble: `max` and other reductions are no longer possible to simplify.

sql/src/test/scala/quasar/sql/compiler.scala
edmundnoble: They just don't get intercepted by the constant planner; afaik that's all that this test was for.

web/src/test/scala/quasar/api/services/InvokeServiceSpec.scala
edmundnoble: `select (1, 2)` would produce a `Data.Set`.

frontend/src/test/scala/quasar/std/set.scala
edmundnoble: We can't keep these tests, because they all require `Data.Set` to be testing anything I believe.

edmundnoble:
To clarify; every test removed or altered in this PR will be restored in the event that multi-row datasets are allowed as constants in the future.

frontend/src/main/scala/quasar/std/set.scala
```diff
@@ -135,98 +116,40 @@ trait SetLib extends Library {
...
-  object JoinFunc {
-    def unapply(func: TernaryFunc): Option[TernaryFunc] =
-      func match {
-        case (InnerJoin | LeftOuterJoin | RightOuterJoin | FullOuterJoin) => Some(func)
-        case _ => None
-      }
-  }
-
-  // TODO: deprecated - delete when old mongo is deleted
-  val InnerJoin = TernaryFunc(
```
alissapajer: :tada: happy to see these gone! :)

frontend/src/main/scala/quasar/std/set.scala
```diff
@@ -135,98 +116,40 @@ trait SetLib extends Library {
...
-    Transformation,
-    "Returns a new set containing the pairs values from the two sets that satisfy the condition, plus all other values from the left set.",
-    Type.Top,
-    Func.Input3(Type.Top, Type.Top, Type.Bool),
-    noSimplification,
-    partialTyper[nat._3] {
-      case Sized(s1, _, Type.Const(Data.Bool(false))) =>
-        Type.Obj(Map(JoinDir.Left.name -> s1, JoinDir.Right.name -> Type.Null), None)
-      case Sized(Type.Const(Data.Set(Nil)), _, _) => Type.Const(Data.Set(Nil))
+  def joinTyper(ty: JoinType): Func.Typer[nat._3] = ty match {
```
alissapajer: :+1:

alissapajer:
@edmundnoble can you see what
```sql
select * from foo join bar on false
```
compiles to and evaluates as on this branch? In master it gives
```
Quasar error: Compilation failed: TypeError(Obj(Map(),Some(Top)),Const(Set(List())),None)
```
If we no longer get the error and the results are empty, can you add a test for this? Then this PR can close out https://slamdata.myjetbrains.com/youtrack/issue/qz-1914.

alissapajer:
Restarted a couple travis jobs. Mongo 3.2 timed out and ML-xml flaked out. I've seen that ML test flake out in another PR recently, actually, so next time we see it flake we should disable it. I just forgot to note the test name before restarting now.

edmundnoble:
I'm saying this is breaking because some queries that used to be evaluated as constant by the planner can now fall through to a connector which can't evaluate them.

PR #3490, opened 2018-02-27
Remove normalizeProjections to prevent field/collection name collision

edmundnoble:
Turns out, `normalizeProjections` was causing this issue and isn't necessary anymore.

#qz-3649 Done


alissapajer:
@edmundnoble can we add a new dataset with a field name the same as the dataset name and then project out the field, to show that it works?

Tests that return `[]` aren't bad to have, but we should strive to have their positive equivalent as well.

alissapajer:
`./regression/fieldtablecollision.test` is failing for mongo.

edmundnoble:
@alissapajer This test should be better, and works on Mongo locally.

it/src/main/resources/tests/regression/fieldtablecollision.test
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "[reg] field/table name collision",
+  "data": "fieldtablecollision.data",
+  "backends": {
+  },
+  "query": "SELECT fieldtablecollision.fieldtablecollision FROM fieldtablecollision",
```
alissapajer: :smile:

PR #3493, opened 2018-03-01
An SST-based random data generator 

wemrysi:
Introduces a new command-line utility, `datagen`, that generates a random dataset conforming to an SST.

There are a couple of known limitations which will be addressed in future work

1. Structure within strings is not recognized (phone numbers, SSNs, UUIDs, ISO8601, etc) and the values generated from these will not retain any of their structure. This can be improved by increasing the amount of information we capture about the distribution of characters within strings in SSTs and also possibly by explicitly recognizing certain common patterns for more declarative generation of the represented values ([qz-3719](https://slamdata.myjetbrains.com/youtrack/issue/qz-3719)).

2. Known tagged types are not generated directly. Since most of these are temporal types and are represented as maps, the resulting generated values are actually quite useful, though there is a chance they are invalid (i.e. 2017-02-30). This can be improved by noticing the type tags and using them to drive the semantics of generation ([qz-3720](https://slamdata.myjetbrains.com/youtrack/issue/qz-3720)).

#qz-3634 Done.


wemrysi:
I'll add issues in YouTrack for the aforementioned limitations.

datagen/src/main/scala/quasar/datagen/Main.scala
```diff
@@ -0,0 +1,99 @@
...
+import scalaz.concurrent._
+import scalaz.std.anyVal._
+import scalaz.syntax.either._
+import scalaz.syntax.show._
+import spire.std.double._
+
+object Main extends TaskApp {
+
+  override def run(args: ImmutableArray[String]) =
+    Stream.eval(CliOptions.parse[Task](args))
```
djspiewak: Noice

datagen/src/main/scala/quasar/datagen/dist.scala
```diff
@@ -0,0 +1,260 @@
...
+
+      Dist.weightedMix[List[(J, J)]](
+        (p, v map (j => List((k, j)))),
+        (1.0 - p, Dist.constant(List())))
+    } map (_.join)
+
+  private def leafDist[J](implicit J: Corecursive.Aux[J, EJson]): Dist[J] =
+    Dist.mix(
+      Dist[Boolean] map (EJson.bool(_))
+    , Dist[Byte] map (EJson.byte(_))
```
djspiewak: Arbitrarily chosen root for this comment…

OH GOD LEADING COMMAS WHY WHY WHY???!!! 😱

;-)

datagen/src/main/scala/quasar/datagen/package.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package quasar
+
+import slamdata.Predef.{RuntimeException, String}
+
+import fs2.Stream
+
+package object datagen {
+  /** A stream that is failed with the given reason. */
+  def failedStream[F[_], A](reason: String): Stream[F, A] =
```
djspiewak: Rip types

datagen/src/test/scala/quasar/datagen/GenerateSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+        val baz = J.str("baz")
+
+        val msst =
+          sst(
+            TypeStat.coll(10.0, Some(1.0), Some(3.0)),
+            TypeF.map(
+              IMap(
+                foo -> sst(TypeStat.count(10.0), TypeF.top())
+              , bar -> sst(TypeStat.count(5.0), TypeF.top())
+              , baz -> sst(TypeStat.count(3.0), TypeF.top())),
```
djspiewak: Omg leading *and* trailing commas!

datagen/src/test/scala/quasar/datagen/PreciseDataCodecSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  "precise data codec roundtrips" >> prop { d: Data =>
+    val encoded =
+      DataCodec.Precise.encode(d).map(_.nospaces)
+
+    Stream.pure(encoded)
+      .unNone
+      .through(codec.ejsonDecodePreciseData[Pure, J])
+      .through(codec.ejsonEncodePreciseData[Pure, J])
+      .toList must equal(encoded.toList)
```
djspiewak: Word of warning on this test…  I think it's fine as it stands, but if we ever broke `encode` such that it always (or almost-always) returned `None`, then this test would become vacuous, since it would just always pass.

foundation/src/main/scala/quasar/contrib/spire/random/dist.scala
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package quasar.contrib.spire.random
+
+import scalaz.{Functor, Monad, Zip}
+import spire.random.Dist
+
+trait DistInstances {
+  implicit val distMonad: Monad[Dist] =
```
djspiewak: Is there already a `cats.Monad` for `Dist`?  If so, then we could use shims instead of writing a contrib instance.

frontend/src/main/scala/quasar/tpe/TypeF.scala
```diff
@@ -150,14 +150,14 @@ object TypeF extends TypeFInstances {
     JR: Recursive.Aux[J, EJson]
   ): GCoalgebra[T \/ ?, TypeF[J, ?], (T, T)] = {
     type LR  = (T, T)
-    val T    = top[J, T]().embed
+    val ⊤    = top[J, T]().embed
```
djspiewak: Definitely nicer.

project/Dependencies.scala
```diff
@@ -23,6 +23,7 @@ object Dependencies {
   private val scalacheckVersion   = "1.13.4"
   private val scalazVersion       = "7.2.18"
   private val scalazStreamVersion = "0.8.6a"
+  private val scoptVersion        = "3.5.0"
```
djspiewak: Sigh… Note to self: need to refactor our crazy dependencies encoding within SBT.

frontend/src/main/scala/quasar/tpe/TypeF.scala
wemrysi: Yeah, I'm honestly surprised we've tolerated the old version for this long.

datagen/src/test/scala/quasar/datagen/PreciseDataCodecSpec.scala
wemrysi: True, though we're not really interested in the details of `encode` here just that these can roundtrip whatever it produces. If it doesn't produce anything, then there's not much to test and we'd catch the issue in the tests for the `Precise` codec.

PR #3495, opened 2018-03-06
Delete duplicate integration test data files.

alissapajer:
Confirmed each duplicate data file has an empty diff against its counterpart. We should not duplicate test data files.

The diff of `largeZips.data` and `qa/largeZips.data` is nonempty, so I left them both for now.


djspiewak:
Needs merge from master.

alissapajer:
Somehow this branch acquired 520e76bff5b9736a5949494f07459a0ab8440270 which is the commit that adds the temporal join test from #3496, so I'm closing this PR since #3496 now fully subsumes it.

PR #3500, opened 2018-03-10
Introduce RecFreeS

jsantos17:


djspiewak:
Originally I was envisioning this being in `quasar.fp.contrib` or something like that, but this is fine.  I doubt we're going to use it outside of qscript.

PR #3501, opened 2018-03-12
add test cases for date, timestamp and to_timestamp filters

rintcius:
add test cases for date, timestamp and to_timestamp filters

Relates to  #sdbe-314, #sdbe-315, #sdbe-316


djspiewak:
For some reason, github isn't detecting that this branch has actually been merged as part of the other PR.

rintcius:
@djspiewak I think github is correct because this PR is against `master` - the changes of this PR have now not landed in `master`

PR #3505, opened 2018-03-13
Update testing.conf and README.md to reflect changes to configuration

jsantos17:
I decided against renaming `quasar_mongodb_3_4` to `quasar_mongodb_3_4_3` and `quasar_mongodb_3_4_13` to `quasar_mongodb_3_4` since that'd be inconsistent with the architecture of the Mongo connector. 

The connector uses the convention ) that higher versions can handle everything lower versions can, which means `quasar_mongodb_3_4_3` has more capabilities than `quasar_mongodb_3_4` which is backwards. Worth noting this is strictly a convention, no code enforces this.

My idea now is to rename:

`quasar_mongodb_3_4_13` to `quasar_mongodb_3_4_4` 

and

`quasar_mongodb_3_4` to `quasar_mongodb_3_4_3`

WDYT: @rintcius, @nic? I don't like the current status quo since `mongob_3_6` and `mongodb_3_2` refers to the last minor versions of each, while `mongodb_3_4` refers specifically to MongoDB 3.4.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ Quasar supports the following datastores:
 quasar_mongodb_read_only
 quasar_mongodb_3_2
 quasar_mongodb_3_4
+quasar_mongodb_3_4_13
+quasar_mongodb_3_6
 quasar_metastore
 quasar_marklogic_xml
 quasar_marklogic_json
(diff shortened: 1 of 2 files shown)


rintcius:
Hmm, what is exactly this inconsistency with the architecture of Mongo connector that you are speaking of? AFAICS, the code needs to be changed but when using `3_4` and `3_4_3` in the code it will fit the interpretation of "<= maximize the unspecified part of the version" (just like 3_2 and 3_6).

I like `quasar_mongodb_3_4_3` and `quasar_mongodb_3_4_4` less because 
1) 3_4_4 won't fit in that interpretation
2) the container name `quasar_mongodb_3_4_4` doesn't match the version
 
So I still like this best:
```
docker container name      code
quasar_mongodb_3_4         3_4
quasar_mongodb_3_4_3       3_4_3
```

jsantos17:
@rintcius I like your proposal. I think I'll go with that.

jsantos17:
@rintcius good call! Will update naming in the code after `next-major` lands. I fear the merge conflicts might be intractable otherwise.

PR #3506, opened 2018-03-13
Implemented using local sbt in publishing

lukaszwawrzyk:
The goal is to use local sbt script everywhere. In quasar it was used very consistently. The only place was publishAndTag script that is created by sbt-slamdata. Plugin was updated so that it requires `$SBT` with sbt path to work.
I exported this variable from env.global section of travis file. As all scripts are run from the main directory, this should work.
Alternative way is to add `export` to the line in scripts/constants that I deleted. But I think that travis env looks more predictable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ env:
     # - CONNECTOR=spark_hdfs # no spark for 2.12
   global:
     - COURSIER_PROGRESS=0
+    - SBT="./sbt"
 
 # this is also the test stage :eyeroll:
 script:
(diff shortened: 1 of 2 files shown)


lukaszwawrzyk:
@djspiewak latest quasar build failed because of lack of $SBT This PR should have been merged before as sbt-slamdata 0.7.0 includes changes from 0.6.2 which requires this variable set.
I rebased on top of master, it also retriggered build that failed for some odd reason before.

djspiewak:
Change has already been implemented in master.

PR #3508, opened 2018-03-13
Provide types to get rid of scala pattern match warnings.

alissapajer:


djspiewak:
It should be. You're right that it didn't. Heads up, @jdegoes

PR #3515, opened 2018-03-15
Make LeftShift's struct into a RecFreeMap

jsantos17:
This PR changes `LeftShift`'s `struct` to be a `RecFreeMap` instead of a `FreeMap`. I made liberal use of `linearize` to get it to compile. I'm mostly opening this to get some feedback on my current progress. There are still some things I'd like to change before this is mergeable:

- [ ] Calls to `RecFreeS.fromFree` should be replaced by a new DSL to construct `RecFreeMap`
- [ ] Re-enable commented tests in `RenderQScriptDSL`

@djspiewak is this somewhat similar to what you had in mind?

### DIFF
--- a/connector/src/main/scala/quasar/qscript/RecFreeS.scala
+++ b/connector/src/main/scala/quasar/qscript/RecFreeS.scala
@@ -86,8 +86,4 @@ object RecFreeS {
   implicit final class LinearizeOps[F[_], A](val self: Free[RecFreeS[F, ?], A]) extends AnyVal {
     def linearize: Free[F, A] = RecFreeS.linearize(self)
   }
-
-  implicit final class ToRecOps[F[_], A](val self: Free[F, A]) extends AnyVal {
-    def toRec: Free[RecFreeS[F, ?], A] = RecFreeS.fromFree(self)
-  }
 }
--- a/connector/src/main/scala/quasar/qscript/RecFreeS.scala
+++ b/connector/src/main/scala/quasar/qscript/RecFreeS.scala
@@ -39,8 +39,6 @@ sealed trait RecFreeS[F[_], A] extends Product with Serializable {
 }
 
 object RecFreeS {
-  type RecFreeMapA[T[_[_]], A] = Free[RecFreeS[MapFunc[T, ?], ?], A]
-
   final case class Suspend[F[_], A](fa: F[A]) extends RecFreeS[F, A]
   final case class Fix[F[_], A](form: RecFreeS[F, A], rec: Free[RecFreeS[F, ?], Unit]) extends RecFreeS[F, A]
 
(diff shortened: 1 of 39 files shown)


connector/src/main/scala/quasar/qscript/RecFreeS.scala
```diff
@@ -64,6 +62,9 @@ object RecFreeS {
   def letIn[F[_], A](form: Free[F, A])(rec: Free[F, Unit]): Free[RecFreeS[F, ?], A] =
```
djspiewak: Btw, do you think it's clearer to use `Hole` here rather than `Unit`?

connector/src/test/scala/quasar/qscript/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -762,15 +766,15 @@ object MinimizeAutoJoinsSpec extends Qspec with TreeMatchers with QSUTTypes[Fix]
...
 
           repairInner must beTreeEqual(
             func.StaticMapS(
               "original" -> func.AccessLeftTarget(Access.valueHole(_)),
               "results" -> func.RightTarget))
 
-          structOuter must beTreeEqual(
-            func.ProjectKeyS(func.Hole, "results"))
+          structOuter.linearize must beTreeEqual(
+              func.ProjectKeyS(func.Hole, "results"))
```
djspiewak: Indentation?

connector/src/main/scala/quasar/qscript/construction.scala
```diff
@@ -32,10 +32,208 @@ import scalaz.std.tuple._
 import scalaz.syntax.bifunctor._
 
 object construction {
+  final case class RecFunc[T[_[_]]: BirecursiveT]() {
```
djspiewak: We should also add a constructor to build a let binding.  Something like:

```scala
def Let[A](form: RecFreeMapA[T, A], body: RecFreeMapA[T, Hole]): RecFreeMapA[T, A]
```

connector/src/test/scala/quasar/qscript/qsu/MinimizeAutoJoinsSpec.scala
djspiewak: This is still weird!

jsantos17: Will fix!

jsantos17:
I'd still like to add tests for the DSL to construct `RecFreeMap`  (hence the WIP). Will probably take a bit for merge `next-major` changes into this too.

jsantos17:
@djspiewak this is ready for another look. No longer a WIP

connector/src/main/scala/quasar/qscript/UnaryFunctions.scala
```diff
@@ -63,7 +63,7 @@ object UnaryFunctions {
               case Map(src, fm) =>
                 (f(fm)).map(Map(src, _))
               case LeftShift(src, struct, id, stpe, undef, repair) =>
-                (f(struct)).map(LeftShift(src, _, id, stpe, undef, repair))
+                (f(struct.linearize).map(RecFreeS.fromFree(_))).map(LeftShift(src, _, id, stpe, undef, repair))
```
jsantos17: I just realized this change breaks (at least) the identity law for `Traversal`. Using it undoes let bindings. Not sure what to do:

1. Leave it as is.
2. Modify the `Traversal` to not apply `f` to `struct`. Maybe this will break something?

Ideas @djspiewak?

djspiewak: The problem is mostly that `Traversal` promises a `FreeMap`, when really it needs to promise a `RecFreeMap`.  We can leave this as is for now, I think, but this is one of the things that will need to change in subsequent iterations.

PR #3523, opened 2018-03-21
Remove rmdir mvdir support from mongodb

rintcius:
#qz-3741 Done

### DIFF
--- a/it/src/test/scala/quasar/fs/FileSystemTest.scala
+++ b/it/src/test/scala/quasar/fs/FileSystemTest.scala
@@ -22,6 +22,8 @@ import quasar.contrib.pathy._
 import quasar.contrib.scalaz.eitherT._
 import quasar.fp._
 import quasar.fp.free._
+import quasar.fp.ski._
+import quasar.fs.FileSystemError._
 import quasar.fs.mount._, BackendDef.DefinitionResult, Fixture._
 import quasar.fs.mount.cache.VCache, VCache.VCacheKVS
 import quasar.effect._
@@ -52,9 +54,12 @@ import scalaz.stream.Process
   */
 abstract class FileSystemTest[S[_]](
   val fileSystems: Task[IList[SupportedFs[S]]]
-) extends quasar.Qspec {
+)(implicit Q: QueryFile :<: S, M: ManageFile :<: S) extends quasar.Qspec {
 
   sequential
+
+  val query_  = QueryFile.Ops[S]
+  val manage_ = ManageFile.Ops[S]
 
   type F[A]      = Free[S, A]
   type FsTask[A] = FileSystemErrT[Task, A]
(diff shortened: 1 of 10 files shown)


rintcius:
@wemrysi - I verified this with @beckyconning but I'll double-check with frontend team.

beckyconning:
Can we still rename collections?

beckyconning:
From reading the source it looks like yes.

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
```diff
@@ -56,32 +56,8 @@ object managefile {
     case MoveSemantics.FailIfMissing => RenameSemantics.Overwrite
   }
 
-  def moveDir(src: ADir, dst: ADir, sem: MoveSemantics): MongoFsM[Unit] = {
```
beckyconning: individual collection delete :+1:
individual collection rename :+1:
recursive collection rename :+1:
recursive collection delete :-1:

mongodb/src/main/scala/quasar/physical/mongodb/fs/managefile.scala
```diff
@@ -76,9 +76,7 @@ object managefile {
     if (src === dst)
       ().point[MongoFsM]
     else if (depth(src) == 1)
-      dbNameFromPathM(src) flatMap { dbName =>
-        moveAllUserCollections *> dropDatabase(dbName).liftM[FileSystemErrT]
-      }
+      unsupportedOperation("MongoDb connector does not support moving directories involving databases").raiseError[MongoFsM, Unit]
```
rintcius: @beckyconning @wemrysi - I interpreted that this was the outcome of the conversation between the 2 of you on Slack. Not 100% sure, alternative would be (but that'd make the move unclean IMO): 

```
if (src === dst)
  ().point[MongoFsM]	 
else
  moveAllUserCollections
```

rintcius:
@beckyconning @wemrysi Ready for another look, but not yet ready for merge: 

When testing I noticed that cleanup of tests was also using the functionality that got removed, so I changed that too. There's still a problem in the cleanup of test files: mongo ITs that use `tmp.gen_`-collections *and* result in an error are not cleaned up properly (there are 2 or 3 of such tests among which `s07_join/changedTestStatus`). I'll look into this further tomorrow.

rintcius:
Oh! That cleanup issue already exists in `master`. I have created qz-3753 for that.

beckyconning:
This looks great : ). In the future I think we will be more restrictive but this is perfect for right now.

it/src/test/scala/quasar/fs/FileSystemTest.scala
```diff
@@ -52,10 +54,13 @@ import scalaz.stream.Process
   */
 abstract class FileSystemTest[S[_]](
   val fileSystems: Task[IList[SupportedFs[S]]]
-) extends quasar.Qspec {
+)(implicit Q: QueryFile :<: S, M: ManageFile :<: S) extends quasar.Qspec {
```
wemrysi: These constraints should be on `doDelete` as that is the only function that requires them.

it/src/test/scala/quasar/fs/FileSystemTest.scala
```diff
@@ -134,6 +139,29 @@ abstract class FileSystemTest[S[_]](
   def execT[A](run: Run, p: Process[FileSystemErrT[F, ?], A]): FsTask[Unit] =
     p.translate[FsTask](runT(run)).run
 
+  def doDelete(run: Run, dir: ADir): FsTask[Unit] = {
```
wemrysi: :sob: I see why this is necessary, but seems so arbitrary. We really need to make loading and cleaning up test data first-class in the test suite.

PR #3525, opened 2018-03-22
Enable mimir fs tests

alissapajer:
These are all passing locally and do not run slowly (at least not more slowly than other tests in the suite).


alissapajer:
appveyor timed out. It doesn't seem to not time out anymore.

djspiewak:
We might need to enable sbt-hoarder for appveyor. Right now, it's just applied to Travis, since appveyor was just coincidentally working.

PR #3526, opened 2018-03-23
Implement auto PR to bump sdbe on release

lukaszwawrzyk:
This PR depends on https://github.com/slamdata/slamdata-backend/pull/325 and https://github.com/slamdata/sbt-slamdata/pull/46
plugin version needs to be bumped to a real one once it is released.


lukaszwawrzyk:
PR is ready, but should be merged after sdbe, because otherways it will try to bump the version file that won't exist yet (quasar-version).

PR #3528, opened 2018-03-26
Simplify some constant reductions in QSU

edmundnoble:
I'm not sure I'm doing the right thing with providence here, hence the unused `StateT` over `QDims`. If I am, I can delete that.

### DIFF
--- a/connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
+++ b/connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qscript.qsu
+
+import slamdata.Predef._
+import ApplyProvenance.AuthenticatedQSU
+import quasar._
+// import quasar.contrib.matryoshka._
+import quasar.ejson.EJson
+import quasar.fp._
+import quasar.qscript._
+import quasar.Planner._
+import ReduceFuncs.{Arbitrary, First, Last}
+
+import matryoshka.{Hole => _, _}
+import matryoshka.data.free._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+final class FoldConstantReductions[T[_[_]]: BirecursiveT: EqualT: RenderTreeT: ShowT] private () extends QSUTTypes[T] {
+  import QSUGraph.Extractors
+  private val func = qscript.construction.Func[T]
+  private val json = ejson.Fixed[T[EJson]]
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  def apply[F[_]: Monad: PlannerErrorME](aqsu: AuthenticatedQSU[T])
+      : F[AuthenticatedQSU[T]] = {
+    type G[A] = StateT[StateT[F, RevIdx, ?], QAuth, A]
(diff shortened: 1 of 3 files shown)


connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
```diff
@@ -0,0 +1,69 @@
...
+import matryoshka.data.free._
+import matryoshka.implicits._
+import scalaz._, Scalaz._
+
+final class FoldConstantReductions[T[_[_]]: BirecursiveT: EqualT: RenderTreeT: ShowT] private () extends QSUTTypes[T] {
+  import QSUGraph.Extractors
+  private val func = qscript.construction.Func[T]
+  private val json = ejson.Fixed[T[EJson]]
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
```
wemrysi: Is this necessary?

connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
```diff
@@ -0,0 +1,69 @@
...
+  def apply(aqsu: AuthenticatedQSU[T])
+      : AuthenticatedQSU[T] = {
+    type G[A] = State[QAuth, A]
+    val (auth, graph) = aqsu.graph.rewriteM[G](Function.unlift(extract)).run(aqsu.auth)
+    ApplyProvenance.AuthenticatedQSU(graph, auth)
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
```
wemrysi: Or this?

connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
```diff
@@ -0,0 +1,69 @@
...
+
+final class FoldConstantReductions[T[_[_]]: BirecursiveT: EqualT: RenderTreeT: ShowT] private () extends QSUTTypes[T] {
+  import QSUGraph.Extractors
+  private val func = qscript.construction.Func[T]
+  private val json = ejson.Fixed[T[EJson]]
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  def apply(aqsu: AuthenticatedQSU[T])
+      : AuthenticatedQSU[T] = {
+    type G[A] = State[QAuth, A]
```
wemrysi: Yeah, I don't think this is necessary since you aren't adding any new nodes and constant folding shouldn't have an effect on provenance.

it/src/main/resources/tests/demo/reductionsonnull.test
```diff
@@ -0,0 +1,16 @@
+{
```
wemrysi: Is this supposed to be related to the change? If so, it doesn't appear to feature any reductions over constants.

edmundnoble: It does generate reductions over constants, no less than three `Arbitrary(Null)`, in the grouping clause and fields.

connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
edmundnoble: :+1:

connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
edmundnoble: Nope.

connector/src/main/scala/quasar/qscript/qsu/FoldConstantReductions.scala
edmundnoble: Ditto.

it/src/main/resources/tests/demo/reductionsonnull.test
wemrysi: Huh, so it does. It didn't use to and I was using an older version of the repl to look at the plan for that query. The version after your changes looks much better than what's in master!

edmundnoble: Yeah, @djspiewak said that the reason they're there now is to make provenance line up inside QSU. We still don't eliminate grouping by a constant even with this PR; I'll make an issue for that specifically now.

edmundnoble:
Restarting, looks like a (potentially new?) intermittent double-related failure.

edmundnoble:
I guess we just have some build failures because of hoarder now? Not sure if this is mergeable.

rintcius:
@edmundnoble Those hoarder errors are just noise, they don't fail the build. Looks like you just need to mark `reductionsonnull` for some backends

edmundnoble:
Thanks @rintcius.

PR #3529, opened 2018-03-27
Fix flatten and shift map involving wildcard

rintcius:
#qz-3696 Done
#qz-3746 Done

--- Old description ---

The problem occurs when the `.deref` is `None` which causes `prj` to be omitted from the `Reshape`.

I have improved the behaviour by adding `BsonField.Name("value") -> \/-(prj)` in this case, but it still doesn't fix it correctly I think (`value` just happens to be a choice that isn't *too* bad):

```
(v36.0.1) 💪 $ select *{_:} from newTests
MongoDB:
db.newTests.aggregate(
  [
    { "$project": { "o": { "$objectToArray": "$$ROOT" }, "_id": false } },
    { "$unwind": { "path": "$o" } },
    { "$limit": NumberLong("10") },
    { "$project": { "value": "$o.k", "_id": false } }],
  { "allowDiskUse": true });


Query time: 0.2s
 value      |
------------|
 _id        |
 name       |
 backends   |
 data       |
 query      |
 predicate  |
 expected   |
 _id        |
 name       |
 backends   |
(v36.0.1) 💪 $ select distinct *{_:} from newTests
MongoDB:
db.newTests.aggregate(
  [
    { "$project": { "o": { "$objectToArray": "$$ROOT" }, "_id": false } },
    { "$unwind": { "path": "$o" } },
    { "$group": { "_id": { "0": { "value": "$o.k" } } } },
    { "$limit": NumberLong("10") },
    { "$project": { "__quasar_mongodb_sigil": "$_id.0", "_id": false } }],
  { "allowDiskUse": true });


Query time: 0.4s
 value              |
--------------------|
 nb                 |
 description        |
 data               |
 ignoreFieldOrder   |
 query              |
 backends           |
 ignoredFields      |
 name               |
 predicate          |
 ignoreResultOrder  |
```

I also think there is another problem in this code. The following line leaves out nesting information (because it takes `f.flatten.head`):

```
val fldName: Option[BsonField.Name] = (base.toDocVar \\ field).deref.map(f => f.flatten.head)
```

So I think in case there's a nested field the function will return wrong results.

And finally, `_id`s are included in the results. Not sure if this needs to be investigated further.

@wemrysi @jsantos17 Any ideas how to improve this fix?


wemrysi:
@rintcius 
> I have improved the behaviour by adding BsonField.Name("value") -> \/-(prj) in this case, but it still doesn't fix it correctly I think (value just happens to be a choice that isn't too bad)

I think you can use the sigil here (pretty much anytime you're forced into constructing a field to return a value, use the sigil) and it will be elided later.

> So I think in case there's a nested field the function will return wrong results.
Agreed, I think it should be `f.flatten.last` to get the proper name. i.e. right now if one flatten's `foo.bar.baz` the result will be like `{"foo" : ....}` when it should be `{"baz": ...}`.

> And finally, _ids are included in the results. Not sure if this needs to be investigated further.
This is technically correct as the document has an `_id` field, so I think it's fine. Since users sometimes store their data in this field (see `zips`) we can't really know when it is synthetic or not.

rintcius:
@wemrysi thanks!

> I think you can use the sigil here (pretty much anytime you're forced into constructing a field to return a value, use the sigil) and it will be elided later.

I tried using the `Sigil` and that works for `select *{_:}`, but in case of `select distinct *{_:}` it wasn't giving the correct result because of nested `sigil` that didn't get elided (at least in the repl), i.e. the result showed up as:

```
 __quasar_mongodb_sigil |
------------------------|
 nb                     |
...
```

This was why I choose `value`, assuming `sigil` was not meant to be used in these cases. But maybe I should look into whether this sigil can be elided too? (Thinking if there are multiple nested `sigil`s of the form `sigil`: { `sigil`: { `sigil`: {...}}}` then they can be elided)

> Agreed, I think it should be f.flatten.last to get the proper name. i.e. right now if one flatten's foo.bar.baz the result will be like {"foo" : ....} when it should be {"baz": ...}.

I have trouble coming up with an example query for this, but I was thinking that `foo.bar.baz` has to be preserved fully somehow (not sure though).

> This is technically correct as the document has an _id field, so I think it's fine. Since users sometimes store their data in this field (see zips) we can't really know when it is synthetic or not.

:+1:

wemrysi:
@rintcius Regarding the sigil, we'd need to look at why we end up with two. we try and detect if the result value is already included under the sigil, but maybe that detection is missing cases? I think we should avoid nesting them rather than elide the nested values.

> I have trouble coming up with an example query for this, but I was thinking that foo.bar.baz has to be preserved fully somehow (not sure though).

Isn't an example `select foo.bar.baz{_:}, quux from something` and we'd expect results like `{"baz": <key>, "quux": ...}`?

jsantos17:
My thoughts on this:

I believe this should be handled while compiling `QScript` to `WorkflowBuilder` instead of handling it between `WorkflowBuilder` to `Workflow`. Specifically it should be done when analyzing the `repair`.

My reasoning is as follows: the meaning of `RightSide` in a `LeftShift`'s `repair` changes depending on the `IdStatus` and `ShiftType` of the `LeftShift`.  Hence, an extra `ExprBuilder`/`DocBuilder` should be added inbetween the `FlatteningBuilder` and the `repair`'s `ExprBuilder`/`DocBuilder` to account for this difference. This new `WorkflowBuilder` node would be in charge of projecting either `k` or `v` depending on whether `IdStatus = IdOnly` or `IdStatus = ExcludeId` or `IdStatus = IncludeId`.  Doing it this way avoids the dummy `value` key.

The downside of this is that the `mapReduce` implementation of object flattening will need updating to return the key and/or value inside `k` and/or `v` keys. Otherwise it won't work in Mongo without `$objectToArray`.

rintcius:
@wemrysi 
I agree it would be better to avoid nesting sigils in the first place, but where's the code where "we try and detect if the result value is already included under the sigil"? I couldn't find it.

> Isn't an example select foo.bar.baz{_:}, quux from something and we'd expect results like {"baz": <key>, "quux": ...}?

No, for these kind of queries the results are like `{"0": <key>, "quux": ...}`

@jsantos17 
I was also thinking to fix the `value` issue with extra builders in `MongoDbPlanner`, but I can't follow how that relates to what you say about `RightSide`/`repair`. Maybe something to discuss in a hangout session somewhere today?

wemrysi:
@rintcius Ah, right, the numeric names.

The place I was referring to where we check for the sigil is in [workflowbuilder](https://github.com/quasar-analytics/quasar/blob/ac9d99a6571eaa6f5c69e9cc3fef4c2b8dc91ff6/mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala#L600).

rintcius:
Not sure if this is the only reason, but I think we can end up with nested sigils because `workflowbuilder.toWorkflow` is called recursively.
In `workflowbuilder.build`, the algebra may call `shift` to deal with sigils, but when `toWorkflow` is called recursively, we don't call (and cannot call) `shift`.

Some ways to solve this issue (from best to worst solution, but also from most to least amount of work):

1) Make `toWorkflow` integrate properly with `shift`. We'd need to get rid of recursive calls from `toWorkflow`, which is tech debt so improves the structure of the code as well as (hopefully) fix this issue.
2) Construct extra `DocBuilder` wrapping `$$ROOT`, to avoid having to call `FlatteningBuilder` with `$$ROOT`. And add an extra `ProjectKey` to select into wrapped doc.
3) Allow nested sigils to be constructed and descend recursively into sigils when getting the result 

Option 3 is implemented at this moment and pushed here right now and I'm busy implementing option 2. The implementation for 2 has some similarities with what @jsantos17 proposed, but _is_ done in MongoDbPlanner (i.e. QScript -> WorkflowBuilder) and doesn't need any changes in `mapReduce` implementation of object flattening. So I will discuss this with @jsantos17 

I'm currently not looking into option 1, but I'm curious to learn how we could get rid of the recursion in `toWorkflow`

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -690,15 +690,18 @@ object MongoDbPlanner {
...
+
+            } else {
 
-            else {
               val struct0 = struct.transCata[FreeMap[T]](orOriginal(rewriteUndefined[Hole]))
-              val repair0 = repair.as[Hole](SrcHole).transCata[FreeMap[T]](orOriginal(rewriteUndefined[Hole]))
+              val repair0 =
+                repair.as[Hole](SrcHole).transCata[FreeMap[T]](orOriginal(rewriteUndefined[Hole])) >>=
+                  κ(Free.roll(MFC(MapFuncsCore.ProjectKey[T, FreeMap[T]](HoleF[T], MapFuncsCore.StrLit("wrap")))))
+
```
rintcius: @jsantos17 & I discussed adding an `ExprBuilder` instead of changing this `repair0` but that didn't work out. 

I think still worth spending some time & see if we can get rid of the QScript rewrites done here

mongodb/src/main/scala/quasar/physical/mongodb/workflowbuilder.scala
```diff
@@ -314,10 +321,10 @@ object WorkflowBuilder {
    * fields - the fields to flatten
    * rest - the other fields of the structure or `None` if unknown
    */
-  final case class FlatteningBuilderF[F[_], A](src: A, fields: Set[StructureType[DocVar]], rest: Option[List[BsonField.Name]])
+  final case class FlatteningBuilderF[F[_], A](src: A, fields: Set[StructureType[BsonField.Name]], rest: Option[List[BsonField.Name]])
```
rintcius: Tightened the type for `fields`. This makes `DocVar.ROOT` as a field unrepresentable (and less importantly, also nested fields)

wemrysi: Nice!

mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -690,15 +690,18 @@ object MongoDbPlanner {
...
+              val wrapKey = BsonField.Name("wrap")
```
wemrysi: We should make `"wrap`" a constant and refer to that everywhere (including tests) instead of the string literal.

jsantos17:
I find it interesting that the `wrap` key is discarded when compiling `WorkflowBuilder` to `Workflow`:

```
Workflow Builder:
ExprBuilder
├─ FlatteningBuilder
│  ├─ DocBuilder
│  │  ├─ ExprBuilder
│  │  │  ├─ CollectionBuilder(Root())
│  │  │  │  ├─ $ReadF(jps; zips)
│  │  │  │  ╰─ Schema
│  │  │  ╰─ \&/(Both)
│  │  │     ├─ Ident(_)
│  │  │     ╰─ ExprOp("$$ROOT")
│  │  ╰─ Shape
│  │     ╰─ Name(wrap)
│  │        ╰─ \&/(Both)
│  │           ├─ Ident(_)
│  │           ╰─ ExprOp("$$ROOT")
│  ├─ Fields
│  │  ╰─ Object(BsonField.Name("wrap"))
│  │     ╰─ IdStatus(IdOnly)
│  ╰─ Rest([])
╰─ \&/(Both)
   ├─ JsCore(_.wrap)
   ╰─ ExprOp("$wrap")
```

is compiled to

```
Workflow (raw):
Chain
├─ $ReadF(jps; zips)
├─ $ProjectF
│  ├─ Name("o" -> { "$objectToArray": "$$ROOT" })
│  ╰─ ExcludeId
├─ $UnwindF
│  ╰─ Path(DocField(BsonField.Name("o")))
╰─ $ProjectF
   ├─ Name("__quasar_mongodb_sigil" -> "$o.k")
   ╰─ ExcludeId
```

This makes me think there might be a more direct way to express this through `WorkflowBuilder`

rintcius:
Yeah, `wrap` is extra structure that can very well be optimized away later on. That's because it's basically a workaround for solution 1 in this comment https://github.com/quasar-analytics/quasar/pull/3529#issuecomment-377546570

jsantos17:
@rintcius needs merge from master too.

PR #3530, opened 2018-03-27
Enable logging of HTTP request lines by default

wemrysi:
Adds a `RequestLogging` middleware for http4s and enables it by default to log HTTP request lines.

#sdbe-325 Done


edmundnoble:
@wemrysi Can you show me a sample log line output by this?

wemrysi:
@edmundnoble Sure
without headers
```
GET /foo/bar/baz?param=value&param=value HTTP/1.1
```
with headers
```
GET /foo/bar/baz?param=value&param=value HTTP/1.1 | Accept: */* | Host: localhost:8080
```

edmundnoble:
Nice.

PR #3531, opened 2018-03-27
Tweak local settings to allow for publication of branch snapshots

djspiewak:
This makes it considerably easier to publish branch snapshots.


nicflores:
To use this folks can run: `sbt publishSign`
@djspiewak, can you link to your gist to setting up PGP/GPG keys.

djspiewak:
Actually the task is `publishSigned`.  Close enough though.  :-) https://gist.github.com/djspiewak/ec8d4f3fa37d94cbbfdb39c42d9e3a40 There's a lot of random info in there that you probably don't need if not doing OSS work, but it's relatively comprehensive.

PR #3534, opened 2018-03-29
Local on disk file system implementation.

alissapajer:
#qz-3768 Done


connector/src/main/scala/quasar/fs/Local.scala
```diff
@@ -0,0 +1,473 @@
...
+          AFile.fromFile(file).toRight(
+            FileSystemError.unexpectedError(
+              None,
+              s"Error constructing an AFile from the created temp file $file.")))
+        .run
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  def queryFile = λ[QueryFile ~> Task] {
+    case QueryFile.ExecutePlan(lp, _) => ??? // not supported
```
wemrysi: These could return an `UnsupportedOperation` error.

connector/src/main/scala/quasar/fs/Local.scala
```diff
@@ -0,0 +1,473 @@
...
+                StandardCharsets.UTF_8,
+                StandardOpenOption.APPEND,
+                StandardOpenOption.CREATE,
+                StandardOpenOption.WRITE)
+            }
+          } match {
+            case Some(\/-(_)) =>
+              Vector[FileSystemError]()
+            case Some(-\/(err)) =>
+              Vector(FileSystemError.unexpectedError(
```
jsantos17: Would it be appropriate to use `WriteFailed` here?

web/src/main/scala/quasar/api/ToApiError.scala
```diff
@@ -96,6 +96,10 @@ sealed abstract class ToApiErrorInstances extends ToApiErrorInstances0 {
         e.toApiError :+ ("logicalplan" := lp.render)
       case QScriptPlanningFailed(e) =>
         e.toApiError
+      case ue @ UnexpectedError(err, reason) =>
+        fromMsg(
+          BadRequest withReason "Unexpected Error",
```
jsantos17: Would `InternalServerError` be more appropriate here?

connector/src/main/scala/quasar/fs/FileSystemError.scala
```diff
@@ -43,26 +43,39 @@ object FileSystemError {
...
   final case class ReadFailed private (data: String, reason: String)
     extends FileSystemError
+
   final case class PartialWrite private (numFailed: Int)
     extends FileSystemError
+
   final case class WriteFailed private (data: Data, reason: String)
     extends FileSystemError
+
+  final case class UnexpectedError(ex: Option[Throwable], reason: String)
```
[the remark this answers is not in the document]
alissapajer: Agreed, will add a constraint like `MonadError[LocalFileSystemError, ?]`.

web/src/main/scala/quasar/api/ToApiError.scala
alissapajer: I'm going to remove `UnexpectedError` in favor of reporting a `LocalFileSystemError` through a `MonadError` constraint.

connector/src/main/scala/quasar/fs/Local.scala
alissapajer: Yes. :+1:

connector/src/main/scala/quasar/fs/Local.scala
```diff
@@ -0,0 +1,473 @@
...
+                Files.readAllLines(jpath, StandardCharsets.UTF_8).asScala.toVector)(
+                s"Reading failed for $file.",
+                _.drop(offset.value.toInt)
+                  .take(limit.map(_.value.toInt).getOrElse(Int.MaxValue)) // FIXME horrible performance
+                  .traverse(str =>
+                    \/.fromEither(Data.jsonParser.parseFromString(str).toEither))
+                  .leftMap(err =>
+                    FileSystemError.readFailed(err.toString, s"Read for $file failed.")))
+            } else {
+              Vector[Data]().right
```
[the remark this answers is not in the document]
alissapajer: Yes, this returns the empty vector when we read a non-existent file. This is what the file system tests are expecting. I do agree it's not the behavior we want, and an error would make more sense. Though since the tests on master are expecting this, I think we should address it as a separate issue.

I'll add a comment here.

connector/src/main/scala/quasar/fs/Local.scala
alissapajer: `WriteFailed` and `ReadFailed` both require a `Data` instance and thus can only be used when we can trace the write error to a single `Data`. We are handed a `Vector[Data]` to write, so if we error on any write-related steps that involve processing an individual `Data` from that vector, we return `WriteFailed`. Otherwise if we fail to write to the file using the java file api and throw an exception, we return a `LocalFileSystemError`. 

^^ (That's how it will be after I push up my most recent changes.)

connector/src/main/scala/quasar/fs/FileSystemError.scala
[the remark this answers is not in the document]
alissapajer: Since we can't abstract over `Task` I'm tracking the error in the implementation's monad, which is now of type `EitherT[F, LocalFileSystemError, A]` (where `F` and `A` depend on the file system component).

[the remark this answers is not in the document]
alissapajer: and `F` explicitly is or "contains" `Task`. All will be clear when I push :)

alissapajer:
This is working on @djspiewak's windows VM. :crossed_fingers: that it works on appveyor too.

alissapajer:
@djspiewak appveyor passed!!!! :tada:

alissapajer:
@jsantos17 this is ready for review (had to restart a single travis job).

connector/src/main/scala/quasar/fs/Local.scala
```diff
@@ -0,0 +1,547 @@
...
+    case ManageFile.Move(pair, sem @ MoveSemantics.FailIfExists) => pair match {
+      case ManageFile.PathPair.DirToDir(src, dst) => delayEither[FSError[Unit]] {
+        val jsrc: JPath = toJPath(src)
+        val jdst: JPath = toJPath(dst)
+
+        failIfTargetExists(src, dst) match {
+          case Some(err) => err.left.right
+          case None =>
+            \/.fromTryCatchNonFatal[Unit]({
+              val contents: List[JFile] = Files.list(jsrc)
```
jsantos17: This fragment seems to be repeatead in a few places. Might be a good idea to refactor it out?

```scala
val contents: List[JFile] = Files.list(jsrc)
  .collect(Collectors.toList()).asScala.toList
  .map(_.toFile)
  .filter(_.isFile)
```

edmundnoble: It looks like these effects aren't encapsulated in `Task`.

alissapajer: @edmundnoble the `delayEither` on line https://github.com/quasar-analytics/quasar/pull/3534/files/9f0c981c483c6b1a6f88decdcbe32289e6af9b8b#diff-8208766527a17d4d27896035c0b5627dR186 wraps in a `Task.delay`.

edmundnoble: Oh, okay.

alissapajer:
@jsantos17 pushed a commit with the refactoring.

alissapajer:
The build is failing due to the parsing errors which should be fixed in https://github.com/quasar-analytics/quasar/pull/3545.

alissapajer:
@jsantos17 merged in master, this should be ready to go now!

PR #3535, opened 2018-03-30
Reduce constant reductions in LP

edmundnoble:
Turns out rewriting constant reductions in QSU causes some odd bugs in ExtractFreeMap when it comes to group key reification. So, I'm doing it in LP with some simplifiers. This exposed constant arrays to RewriteGroupByArrays, which needed to be adjusted for that.

Oh, and this removed grouping by constants anyway during ShiftRead, somehow.

#qz-3766 Done


connector/src/main/scala/quasar/qscript/qsu/RewriteGroupByArrays.scala
```diff
@@ -57,7 +60,13 @@ final class RewriteGroupByArrays[T[_[_]]: BirecursiveT: ShowT] private () extend
         Unary(last, MFC(MapFuncsCore.MakeArray(SrcHole))),
         MapFuncsCore.ConcatArrays(LeftSide, RightSide)) => Some(elems.toVector :+ last)
 
+      case AutoJoin2C(
+        NAryArray(elems @ _*),
+        c @ Unary(_, MFC(MapFuncsCore.Constant(json.arr(_)))),
+        MapFuncsCore.ConcatArrays(LeftSide, RightSide)) => Some(elems.toVector :+ c)
```
djspiewak: Can we put in the reciprocal case as well, since this commutes?

connector/src/main/scala/quasar/qscript/qsu/RewriteGroupByArrays.scala
```diff
@@ -57,7 +60,13 @@ final class RewriteGroupByArrays[T[_[_]]: BirecursiveT: ShowT] private () extend
...
+
       case Unary(last, MFC(MapFuncsCore.MakeArray(SrcHole))) => Some(Vector(last))
+      case c @ Unary(_, MFC(MapFuncsCore.Constant(json.arr(_)))) => Some(Vector(c))
```
djspiewak: Needs unit tests for both this and the other added case.

connector/src/main/scala/quasar/qscript/qsu/RewriteGroupByArrays.scala
edmundnoble: Doesn't seem to be one for the static array counterpart. I guessed that this was written like this because this is always how flattened groupby is associated by LP.

connector/src/main/scala/quasar/qscript/qsu/RewriteGroupByArrays.scala
edmundnoble: :+1: and an integration test.

connector/src/main/scala/quasar/qscript/qsu/RewriteGroupByArrays.scala
djspiewak: I suppose that's true. Let's punt on it for now.  Can you toss a TODO on there to note that we haven't forgotten about the commuted cases, we're just ignoring them for now?

djspiewak:
I’m pretty sure you don’t use cata anywhere. Also using it would be highly dangerous, since it computes a spanning tree, which is seldom what you want. The extractors already do a better job at handling this scenario for examination of subgraphs (or tests), and foldMapUpM is more general, so I don’t see the use of the function.

edmundnoble:
@djspiewak I am *not* merging any QSUGraph manipulation tools in this PR. I wrote those just so I can actually *see* the resulting `QSUGraph`s in the failing tests (I also have a `Delay[RenderTree, QScriptUniform[T, ?]]` instance). That's why `para` is going to compute an annotated tree which I can render, and I'm going to add that as a separate PR. Even if `para` is only useful as a private tool to implement the rendering, I think it deserves inclusion (again, in a separate PR; the only change I really want to make to this branch is to add tests).

djspiewak:
@edmundnoble `RenderTree[QSUGraph]` defined in that way is wrong though, and it's not going to yield the correct information.  `Show[QSUGraph]` is in scope and you can use it.  It doesn't render as a tree, but that's accurate, because it's not a tree.

I'm not averse to *a* `RenderTree` for `QSUGraph`, but just doing it via folds is going to render a spanning tree, not the graph, which is a strongly meaningful distinction given the way that QSU works.

edmundnoble:
Here's the signature of the method I'm writing for this, to give you an idea:
```scala
  def linearizeAsAnnotatedTree[A](
    implicit CR: Corecursive.Aux[A, EnvT[Symbol, QScriptUniform[T, ?], ?]]
  ): A = {
    para[A] { (root, unfold, newVertices) =>
      CR.embed(EnvT((root, unfold.map(newVertices(_)))))
    }
  }
```

djspiewak:
@edmundnoble Can you clarify what is insufficient about the existing `Show[QSUGraph]`?

edmundnoble:
What's insufficient is multiple things. Firstly what's insufficient is that there's no `RenderQSUGraphDSL`. Secondly, what's insufficient is that there is no easy way to view QSU graphs which are *already linearized* as *linearized trees*, which would be immensely convenient for testing because that's actually how we write them down (most of the time).

djspiewak:
@edmundnoble `RenderQSUGraphDSL` is a particularly strong reason *not* to do this, because taking a `QSUGraph`, rendering it via the DSL, and then running that code will not round trip on realistic graphs. It will round trip on graphs which happen to already be linearized, but that is most definitely not all graphs.  The graphs in the unit tests actually aren't linearized, even though they're written that way.  Look at the `Show` output.  :-)  `fromTree` collapses subgraphs, which is really important because if it didn't do that, then a huge swath of the unit tests would fail (e.g. literally all of `MinimizeAutoJoins`).

To be clear, I'm not morally opposed to having more tools for viewing graphs, even in an inaccurate way.  There's an old ticket somewhere that I filed that asks us to implement exactly `RenderTree[QSUGraph]`.  But given the fact that the `Show` instance exists, is 100% accurate, and isn't actually that difficult to read with practice (especially if we reorder topologically, which Emrys suggested and should be pretty trivial), I don't buy the argument that we need to spend time on it right now since it's just not going to be a particularly significant gain.

edmundnoble:
`RenderQSUGraphDSL` can declare variables, so no, it actually will round-trip.

It would certainly be nice to have a topologically-sorted `Show` instance; I think that's probably what I really need.

djspiewak:
> RenderQSUGraphDSL can declare variables, so no, it actually will round-trip.

If those variables are of type `QSUGraph`, then yes, it will work.  That's a different DSL than the one we use in testing though, and it would be too imprecise for use in QSU itself.  So it would essentially only exist to support itself.

> It would certainly be nice to have a topologically-sorted Show instance; I think that's probably what I really need.

IMO that seems like a better use of time.  It's pretty trivial to change the implementation to be topological (we just haven't done it yet), especially since I think `foldMapUpM` will literally just do it for you.  I would make that change and see if it improves the readability of things for you.

edmundnoble:
All the tests and topologically-ordered `Show` have been added.

PR #3536, opened 2018-04-02
Fix discrepancy between view and inline query

jedesah:
This is sort of a double negative makes one positive where the removal of typechecks in the "inline" query when resolving views (ironically the inline query has no views) masks the bug in the compiler by removing a non-sensical typecheck.

So although going through the view produces the wrong result, that's because there is only one bug in play as opposed to two canceling each other out.

Adding a simplifier for the equal operation fixes the problem of the erroneous typecheck though I'm not sure it's the proper fix. Looking for feedback on that as well as my question in the comment regarding why we remove typechecks when resolving views in the first place.

#qz-3656 Done


core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -274,6 +274,7 @@ object view {
 
     // NB: simplify incoming queries to the raw, idealized LP which is simpler
     //     to manage.
+    // JR: Why is this easier to manage exactly? Perhaps we could add a more detailed explanation here
```
wemrysi: I'm not sure it is "simpler to manage" (or what that even means), but it may just be to avoid redundant typechecks as we optimize and typecheck the substituted plan once all views have been resolved.

@jedesah Also, I think I accidentally deleted your original comment when trying to delete a prevous version of my own, apologies!

core/src/main/scala/quasar/fs/mount/view.scala
```diff
@@ -274,6 +274,7 @@ object view {
...
     val cleaned = plan.cata(optimizer.elideTypeCheckƒ)
```
wemrysi: @jedesah So instead of always removing typechecks here, how about if we do the substitution and track whether we actually change the resulting plan. If we did change the plan, _then_ we elide all typechecks and call `preparePlan`, if not, we just return the original `LP`.

frontend/src/main/scala/quasar/std/relations.scala
```diff
@@ -31,7 +31,15 @@ trait RelationsLib extends Library {
     "Determines if two values are equal",
     Type.Bool,
     Func.Input2(Type.Top, Type.Top),
-    noSimplification,
+    new Func.Simplifier {
```
jedesah: So this is only masking the bug in the typechecker but since the typechecker is soon going to be replaced, improving constant plan optimization seems like an acceptable solution in order to sidetrack the bug that @cryogenian is running into in regards to this query.

@edmundnoble You removed a lot of the "simpliers" recently I was told. Do you have objections to the "fix" proposed by this PR?

edmundnoble: Unfortunately, we have two meanings of the term "simplifier", both the actual `Func.Simplifier` and the action of `Typer`s on constant values. I only removed a bunch of typer stuff which used `Data.Set`. With that said; if you're going to add a simplifier for this because the type checker is broken, I'd recommend removing the corresponding simplication case in the typer to avoid confusion (https://github.com/quasar-analytics/quasar/pull/3536/files/7898a47f401c1f8b3e023a9fd35e753f5d80d743#diff-581dc1037e31b411143bae4dfbac2119R45).

alissapajer: To clarify, we plan to replace the typechecker at some point, but it probably won't happen this year since there are other higher priorities. Maintaining the old typechecker has a non-zero cost, but I wouldn't be surprised if there continue to be other higher priority tasks during the next few months.

jedesah: @edmundnoble Ah, I see. So the reason why we normally don't need a simplifier is that the expression is going to get simplified when typechecked?

edmundnoble: @jedesah Right, exactly. So what I'm getting from this PR is that what the type checker is doing isn't working, so I'd remove it if you're going to replace it with a simplifier.

jedesah: Well, it's not that the typechecker is not simplifying this expression, it is. It's that it's adding a date/time typecheck before simplifying the expression and then that typecheck becomes non-sensical. My current understanding based on your latest comment is that by simplifying the expression without typechecks, we get rid of the date/time (specifically a call to `date_part`) stuff that is in a dead code path and then when we typecheck we don't get an erroneous date/time typecheck

jedesah: btw, @edmundnoble. For reference, this is the query that misbehaves:

```
create function agg(:prj, :at) begin 
  case 
    when :at = "count" then count(:prj) 
    when :at = "round" then round(:prj)
    when :at = "hour" then date_part("hour", :prj)
    else :prj
  end
end;

Removing `date_part` case magically removes differences.
select agg(weight, "round"), round(weight) from `/mongodb/data/patients` group by state 
```

core/src/main/scala/quasar/fs/mount/view.scala
jedesah: Done!

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
```diff
@@ -52,6 +52,11 @@ package object scalaz {
       OptionT(self.run.run.map(_.join))
   }
 
+  implicit final class Nested3FunctorsOps[F[_], G[_], H[_], A](val self: F[G[H[A]]]) extends AnyVal {
```
wemrysi: This can also be achieved via composition, `Functor[F].compose[G].compose[H]`.

jedesah: Good point. Although, I still think this is useful in terms of making the usage site concise and thus more readable

alissapajer: I don't think that reimplementing operations that can be derived from primitives is a good idea because it opens up the possibility of bugs being introduced.

Also, the call site may have fewer characters required (just `map3`) but the user then has to go find the implicit class that made `map3` available, which I think is a more convoluted path of understanding that just seeing `Functor[F].compose[G].compose[H]` at the call site.

frontend/src/main/scala/quasar/std/relations.scala
edmundnoble: I am really confused. So what it sounds like you're saying is that typecheck nodes are being added, then the query is being simplified with the current "simplifier" in the equals function's typer. But as far as I remember that's not how typechecking works. Typecheck nodes only get synthesized because the function's typer has been run to determine that they need to be there; the function's typer has already done the simplification by the time they're added. This sounds like a bug in either MapFunc normalization (after the simplification occurs) or the typechecking phase.

jedesah: It's definitely a bug in the typechecking phase and this fix is sidestepping the issue by simplifying the expression during the "optimize" phase before the typechecker has the opportunity to get to it and mess up. Does that match your understanding?

jedesah: @edmundnoble Okay, okay, you've convinced me to have a quick look at the typer to see if I can't find the bug there. It's just that I was told we were going to rewrite the typer soon anyway but as @alissapajer pointed out, that might not happen all that soon after all.

foundation/src/main/scala/quasar/contrib/scalaz/package.scala
alissapajer: I generally still disagree and prefer `Functor[F].compose[G]` better, but I won't hold up merge for it.

frontend/src/main/scala/quasar/std/relations.scala
jedesah: >  I'd recommend removing the corresponding simplication case in the typer to avoid confusion

@edmundnoble In the interest of better understanding the optomizer/typechecker, could you explain to me your rationale behind the above comment. For instance, `Cond` has this kind of logic in both its `Simplifier` and `Typer`, would you suggest removing the simplication logic from the `Simplifier` there as well? If so, what's the point of `Simplifier`, what's an example instance where one would write a simplifier instead of including simplification logic in the typer?

edmundnoble: The point of a simplifier is to do simplifications that the typer can't. So, for example, when your simplification is based on the structure of an expression rather than the types are involved: `-(-1)` for example, the simplifier must inspect the tree and find `neg(neg(x))`.

In the case of `Cond`, you're right. We do have both. I'm not sure why, but one possible reason leaps to mind; perhaps other simplifications yield opportunities for this `Cond` simplification, *and* the `Cond` typer yields opportunities for other simplifications.

Us having both there does suggest though that you should have both here, perhaps. But there should be *no semantic difference either way*, so that's still likely worth investigating.

jedesah: I see, that clarifies things. Ya, I think it makes sense in that case to have them in both places. Some instances of `neg(neg(x))` (like `neg(if(1 == 1) neg(1) else 1`) will be missed if we wait on typechecking to do some of these simplications since we don't call `optimize` again after typechecking.

Actually, now that I think of it, I question whether this logic should even appear in the typer like you suggested earlier since we should never hit that case if we assume that we always call "simplify" before "typechecking". Perhaps that is a bad assumption, but on the other hand, it's unfortunate to have duplicate logic. Am I missing something, is there a reason we want constant folding logic in the typer regardless?

Perhaps as a subsequent PR or even as part of this one I can go through all these functions and make them all consistent with regards to wtv decision we come up with (duplicate everywhere or move from typer to simplifier). Hopefully that would avoid confusion in the future.

edmundnoble: Afaik the reason why we want constant folding is that it exposes simplifier opportunities. Beyond that I don't know. It would certainly simplify the type system a little to remove constant types entirely and perhaps write reductions more directly, with the added benefit that we can more easily manipulate multiple-row datasets.

jedesah: > Edit: To be clear, as I understand it we type check before simplification.

@edmundnoble I don't think that's the case: https://github.com/quasar-analytics/quasar/blob/master/core/src/main/scala/quasar/package.scala#L75

jedesah:
I think this is good to go now, @alissapajer could you give it another review?

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
```diff
@@ -369,6 +369,17 @@ class DataServiceSpec extends quasar.Qspec with FileSystemFixture with Http4s {
...
+        "qz-3656 inline query and view do not behave the same" >> {
+          val data = Vector(Data.Obj(ListMap("weigth" -> Data.Dec(89.9), "state" -> Data.Str("WA"))))
+          val viewFile = rootDir </> file("view")
+          val dataFile = rootDir </> file("data")
+          val sql = sqlB"""create function agg(:prj, :at) begin case when :at = "round" then round(:prj) when :at = "hour" then date_part("hour", :prj) else :prj end end; select agg(weight, "round"), round(weight) from `/data` group by state"""
+          val view = MountConfig.viewConfig0(sql)
+          val mounts = Map((viewFile: APath) -> view)
+          val service0 = service(InMemState.fromFiles(Map(dataFile -> data)), MountingsConfig(mounts))
+          val response = service0(Request(uri = pathUri(viewFile))).unsafePerformSync
+          response.as[String].unsafePerformSync must_=== """{ "0": 90.0, "1": 90.0 }"""
```
alissapajer: This test is only testing that a view returns the expected result. We should also assert that a query returns the expected result.

Also, how is this test working at all given the typo in the data above?

frontend/src/main/scala/quasar/std/relations.scala
edmundnoble: True, well-spotted.

web/src/test/scala/quasar/api/services/DataServiceSpec.scala
alissapajer: Well, it's `pending` :)

alissapajer:
@jedesah the PR body says `#qz-3656 Done` but the unit test that's attempting to verify this is `pending`. Is this PR fixing that issue?

frontend/src/main/scala/quasar/std/relations.scala
```diff
@@ -30,11 +30,16 @@ trait RelationsLib extends Library {
...
+      (orig: LP[T])
+      (implicit TR: Recursive.Aux[T, LP], TC: Corecursive.Aux[T, LP]) =
+        orig match {
+          case Invoke(_, Sized(Embed(Constant(x)), Embed(Constant(y)))) => constant(Data.Bool(x === y)).some
+          case _                                                       => None
+        }
+    },
     partialTyper[nat._2] {
-      case Sized(Type.Const(data1), Type.Const(data2)) =>
-        Type.Const(Data.Bool(data1 == data2))
```
alissapajer: I'd prefer not to delete this. The control flow through the optimizer and typechecker is not well-managed or composed in such a way that it _can_ be well-tested, so I don't feel comfortable ad-hoc deleting cases in the `Typer`s. 

Is this causing something to break? If so, we should understand what's going on with that, as it should not be breaking anything. If not, then let's just leave it.

But to be clear, @jedesah I completely agree that the duplication between the simplifiers and typecheckers is quite sub-optimal and ultimately should be remedied. I just think we need a detailed proposal before we dive into that, and that plan will likely include moving all of this logic to `qsu`. /cc @djspiewak

djspiewak: I believe the reason this logic got removed is it is actually changing the semantics of the typing. The way that the short circuiting is implemented, it is effectively converting ill-typed queries into well-typed ones.  An example in terms of Scala:

```scala
val result: String = 42 match {
  case i: Int => "howdy"
  case s: String => false
}
```

Is this assignment well-typed?  Of course not.  Just because the second branch is dead code doesn't mean the Scala typer shouldn't reject it.  The same thing is happening here.  And the problem is that the boolean short-circuiting is resulting in dead branches *sometimes* being removed.

Hence, why some of the simplifier cases are arguably unsound.

alissapajer: Do you think, given the current state of things, we can actually reason about that soundly for all simplifier/typer cases? I was claiming that we cannot, and that this whole typer/simplifier implementation is a glorified trial-and-error implementation, where we fix up cases when we hit them as bugs.

alissapajer: So if we're "sure" that we can remove this case, okay. But I don't have confidence in the state of this code as a whole to confidently handle all these various cases by hand, especially since we are not testing the short-circuiting thoroughly (as evidenced by this bug).

djspiewak: Oh I'm quite certain that we *cannot* reason about it soundly. I mean, we could probably take that particular case off the table by simply removing short-circuiting of `case` statements entirely, but there might be other, similar situations that we haven't hit yet.  It really is kind of a glorified trial-and-error implementation.

djspiewak: In theory, *removing* cases from the simplifier should be safe. It will result in some queries being rejected that would have previously been accepted, but the fact that we're in that state in the first place is the whole root of this bug.  Adding cases is what's honestly scarier.

alissapajer: :+1: @djspiewak I think we're on the same page. This should go on the list of code debt that needs attention when we have time :)

@jedesah I see the logic for removing this, and constant folding operations in theory should always be an optimization, so I'm okay with this diff.

djspiewak: Among many other things, this is part of the reason that typing *always* happens in a separate pass and prior to any constant folding in normal compilers.  We conflate the two, which is why we're in this pickle.

alissapajer:
@jedesah some tests failed:
```
[error] 	quasar.std.ConstantPlanStdLibSpec
[error] 	quasar.std.RelationsSpec
```

frontend/src/main/scala/quasar/std/relations.scala
djspiewak: I'm going to get the syntax wrong, but bear with me…

```sql
create function foo(cmd, x)
  case
    when :cmd = "add1" then :x + 1
    when :cmd = "add1string" then integer(:x) + 1
  end
end function

select foo("add1", 42) + foo("add1string", "24")
```

See the problem?  `:x` is being magically polymorphic.  It only works because the `case` is short-circuited when the equality comparison is fully statically known.  If it weren't fully statically known, then the typechecks that would be inferred would be contradictory (integer for one branch, string for the other).  Right now, this results in completely invalid typechecks; more ideally it would result in an error.

Either way, it's 100% wrong.  Constant folding needs to happen after type checking.  Full stop.  The above query is simply invalid.

edmundnoble: I don't understand how they're "contradictory", any more than `if` and `else` branches being "contradictory". This is valid to do, and it *should* short-circuit and remove the branch during typing. Though I don't see why we have constant folding in two different places, it seems that the optimizer's (simplifier's) constant folding could never even reduce this `case` statement because it doesn't know about the static type of `:cmd`.

djspiewak: @edmundnoble Look at the Scala snippet I posted.  Does that type check (in scalac)?  Why not?  Should it?

djspiewak: Here's another quasar example (one that is actually closer to what originated this ticket):

```sql
create function foo(cmd, x)
  case
    when :cmd = "add1" then :x + 1
    when :cmd = "reduce" then count(:x)
  end
end function

select foo("add1", 42) + foo("reduce", *) from baz
```

Now the *dimensionality* of the `foo` function is polymorphic.  Where does it end?

djspiewak: I have a call with Becky in a few minutes. Can jump on afterward though.

edmundnoble: /cc @alissapajer 

I had a call with @djspiewak, where we hashed this out.

The problem appears to be that we float type-check nodes out from `case`. So the different branches which require different types (and thus, different type-check nodes) are replaced with a single type check node and multiple branches. Perhaps this is what's most worth investigating @jedesah? If that's not it, I think we don't have a grasp on what's really the problem here.

The typecheck-floating is actually *not* valid to do for QScript. We're trying to use static types to reduce the amount of dynamically-typed performance killing we have to do; so in general adding an extra case branch *may* widen the type of a function argument, meaning typecheck floating only works if every case expects the *same* type.

Also, we've agreed that the typer should stop doing constant folding. I think we should leave that to later; mimir should really be doing it, and the simplifier should "massage" the expressions into a form which makes them easier for mimir to evaluate rather than doing its job for it.

jedesah: Sorry, I wanted to jump in earlier, but I was busy running around and don't like typing too much on my phone.

> are replaced with a single type check node and multiple branches

@edmundnoble That's not quite correct. It's true that the typechecks occur outside the `case` block, but they are both still there, they do not get "widened" into a single typecheck. And since they are checking for two different types in the case of this query, we are guaranteed to never actually execute anything in the block since no value is ever going to be a number and a date at the same time.

I was going to chime in on how I think it makes sense to treat the value polymorphically given that we are dealing with non-structured data but sounds like you guys talked about it and came to that conclusion as well.

I agree that the solution is to make sure the typechecking occurs inside the various branches so that this query can function properly even in the case where we are branching on a dynamic value (and constant folding would not solve the issue unlike for this specific query). But that change appeared non-trivial to me yesterday which is why I asked @djspiewak when and why we might move the typechecker to QScript and whether that would involve throwing out this code or translating it over.

Since the answer was we would most likely start from scrath, I just wanted to check in again to make sure this was worthwhile work considering we might end up throwing it out. But if so, I'm happy to get started on making the change. I'll assume it's a go ahead unless I hear otherwise.

edmundnoble: > It's true that the typechecks occur outside the case block, but they are both still there, they do not get "widened" into a single typecheck. 

I didn't even consider that as a possibility. I assumed `case` was treated specially in typer, but no, `case` is compiled to nested `Cond`, so this is an emergent property of the typer, which is registering multiple uses of a variable at different types in the same program, which has to yield something that conforms to Type.Bottom` (in this case, `Data.NA`). The problem with the typer is, it doesn't know that *only one* of the branches will execute at runtime, it thinks *both will* (it's *strict* in the programming language sense).

> I agree that the solution is to make sure the typechecking occurs inside the various branches so that this query can function properly even in the case where we are branching on a dynamic value (and constant folding would not solve the issue unlike for this specific query). 

Agreed here.

So if I understand correctly, the way to do this is to keep the type information gained inside nested `Cond` applications independent, and `lub` all of the variable type info gained when some LP outside of those nested `Cond` applications asks for types of the accessed variables.

This seems to imply statefulness in the typer (and/or untyper), which may not be too time-consuming: it would require special-casing `inferTypes` and/or `checkTypesf`, and/or possibly changing `ConstrainedPlan` to allow for disjunctive constraints as well as conjunctive.

Also, I'm not sure to what extent `Cond` is special here; it seems to apply inside every branching construct, including `Typecheck`.

Not much I can say about the time cost here, but I believe it's likely something we'll have to understand for the new type checker anyway.

alissapajer: Given that all this code is in need of some major redesign, I'm concerned that if we dig into this now, we'll be going down a deep rabbit hole with some possible yak shaving on the way. We'd basically be doing some ad-hoc patching on an already patched-up piece of code.

As such, we should bring @beckyconning in on this so we can make the decision based on how this impacts business needs. It's possible that this kind of query will be something we don't need to support yet, and so we can fix this in the new qscript-based typechecker.

alissapajer: @jedesah could you add pending unit tests with examples of this problem?

edmundnoble: Just to be clear, this affects all queries for which the dynamic type of one expression depends on the dynamic type (or value) of another expression.

alissapajer: @edmundnoble what are some simple SQL2 queries that are going to hit this and what is going to go wrong? 

In `select foo + bar from baz`, `foo` and `bar` are dynamically typed, yet this is a case we avoid handling currently, relying on the connector to return undefined when we sum two non-summable types. We don't currently have a way to express the idea in the typer that "if `foo` is numeric, then `bar` has to be as well". We've known this for a long time, so why do we suddenly need to solve this problem now?

edmundnoble: That's not what I mean exactly. It's not so much "if a is an int, b is a double", which provides us with *optimization* (in that we have one less typecheck node). It's more like, if you have this JSON structure:

```json
{
  "fields": [
    { "name": "customerCount", "value": 20 },
    { "name": "fullName", "value": "Edmund Noble" }
  ]
}
```

And you write code which does something different depending on whether your field has `name` `customerCount` or `fullName` (which treats `value` as alternately an `Int` or a `String`), then your code *will not work*. At all. Everything will be `undefined`.

Put another way; we can't actually deal with heterogeneously-typed data at *all*.

alissapajer: Um, your last statement is hyperbole, no? Can we not run `select foo + bar from baz` where `foo` and `bar` are heterogeneous columns?

edmundnoble: Er, sorry, "heterogeneously-typed data" is too general here probably. Rather, data that may have different types depending on which branch of an if-expression they're accessed in.

edmundnoble: How about "data with out-of-band type tags"? I'm not sure, but I think this is fairly common among JSON data.

alissapajer: Ok, so what I don't know is exactly what subset of SQL2 we need to support (both initially and longer term). So we should confirm what we need to support with @beckyconning before making any changes that could be potentially supporting something we don't (yet) need.

alissapajer: And also we need to confirm what kinds of data that SQL2 will be running against.

jedesah: @alissapajer I'll add a query that exemplifies this problem. It's basically going to be something similar to the query that I already added as a pending test for this PR already but with branching on a dynamic value instead of static value so that the constant folding fix doesn't solve the problem (since sounds like we want to eventually perform constant folding after typechecking)

jedesah:
Btw, @alissapajer, to answer your previous question about why the test is pending. It's not because the fix doesn't work (which is a subset of the real problem as elaborately discussed in the other thread), it's because the testing infrastructure at the web layer simply doesn't make it easy to test this because querying isn't supported by the in memory interpreter.

However, it's now clear that there are two very distinct problems here. First, there is the fact that queries get "un-typechecked" because of views which leads to masking of this bug. And second there is our improper support for branching logic (in regards to typechecking). I think at this point it makes sense to move tests regarding that to either integration tests or unit tests in `core` somewhere.

Making sure that views and queries behave the same can be tested here but until we have an in-memory evaluator that can resolve queries, it's a little tricky to test. This test is a "this is how I would like to be able to test this" but can't without query support test. Maybe that's not very useful. In any case, sounds like views might be dropped going forward or changed dramatically, so this is probably a less important problem.

djspiewak:
Going to close this for now, since views are getting changed quite considerably.

PR #3539, opened 2018-04-04
Revised hoarder's Travis semantics to take our build stages into account

djspiewak:
This might fix hoarder for integration tests as well, now that I think about it…

### DIFF
--- a/project/CachedCi.scala
+++ b/project/CachedCi.scala
@@ -1,3 +1,9 @@
+import org.romanowski.hoarder.actions.CachedCI
 import org.romanowski.hoarder.actions.ci.TravisPRValidation
 
-object CachedCi extends TravisPRValidation.PluginBase
+object CustomTravisValidation extends TravisPRValidation() {
+  override def shouldUseCache(): Boolean =
+    super.shouldUseCache() || sys.env.get("TRAVIS_BUILD_STAGE_NAME").map(_ != "Compile").getOrElse(false)
+}
+
+object CachedCi extends CachedCI.PluginBase(CustomTravisValidation)


project/CachedCi.scala
```diff
@@ -1,3 +1,9 @@
+import org.romanowski.hoarder.actions.CachedCI
 import org.romanowski.hoarder.actions.ci.TravisPRValidation
 
-object CachedCi extends TravisPRValidation.PluginBase
+object CustomTravisValidation extends TravisPRValidation() {
+  override def shouldUseCache(): Boolean =
+    super.shouldUseCache() || sys.env.get("TRAVIS_BUILD_STAGE_NAME").map(_ != "Compile").getOrElse(false)
```
alissapajer: Is your theory that `super.shouldUseCache()` is returning `false` when we want it to return `true`?

alissapajer: To clarify, is `super.shouldUseCache()` returning `false` when it _should_ (according to... something) be returning `true`? Basically, I'm just asking for more info about how and when `shouldUseCache` is set.

djspiewak: `shouldUseCache` is defined [here](https://github.com/romanowski/hoarder/blob/09452293e81e296427dfac0904fb099b23f68420/hoarder/src/main/scala/org/romanowski/hoarder/actions/ci/TravisPRValidation.scala#L26).  I know for a fact it's `false` in the `Unit test` and `Test` stages because hoarder logs when this is the case.  I haven't yet looked to see if that's happening in PRs for the `Test` stage, but it's very possible.

PR #3540, opened 2018-04-04
Experimental webhook support for discord

djspiewak:


djspiewak:
Worth noting that the webhook will not work on the PR because of how the encryption was done.  I'll think about how to address that.  But in the interim, it'll be interesting to see what it does in the build.

PR #3543, opened 2018-04-05
Change Map's function to be a RecFreeMap

jsantos17:
Opening this so Travis will run the test suite.

#qz-3772 done.


jsantos17:
@djspiewak this is ready for review now. However, I have once concern that I'd like to discuss further:

There are some places (such as normalization) that apply transformations to `FreeMap`s. For now, to apply these transforms I `linearize` the `FreeMap`, apply the transformation and reconstruct the `RecFreeMap` with `RecFreeS.fromFree`. Obviously this undoes any let bindings that might've existed in the `RecFreeMap`. Which leads me to the question:

Should I keep it as is or figure out how to apply `FreeMap` transformations to the bound form and body separately?

connector/src/main/scala/quasar/qscript/RecFreeS.scala
```diff
@@ -84,7 +87,7 @@ object RecFreeS {
   }
 
   // FIXME: Display the bound form and body separately
-  implicit def renderTree[F[_], A](implicit F: RenderTree[F[A]], FR: RenderTree[Free[F, A]]): RenderTree[RecFreeS[F, A]] =
+  implicit def renderTree[F[_], A](FR: RenderTree[Free[F, A]]): RenderTree[Free[RecFreeS[F, ?], A]] =
```
djspiewak: `implicit` modifier got lost.  This isn't going to behave the way you think it is.

connector/src/main/scala/quasar/qscript/QScriptCore.scala
```diff
@@ -250,7 +250,7 @@ object QScriptCore {
             v match {
               case Map(src, f) =>
                 NonTerminal("Map" :: nt, None,
-                  RA.render(src) :: f.render :: Nil)
+                  RA.render(src) :: f.linearize.render :: Nil)
```
djspiewak: Random note: we really should do a `RenderTree` for `RecFreeMap` which has some marker for the let bindings, so we know where they are in output.

connector/src/main/scala/quasar/qscript/rewrites/Rewrite.scala
```diff
@@ -127,7 +128,7 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] extends TTypes[
 
   def elideNopQC[F[_]: Functor]: QScriptCore[T[F]] => Option[F[T[F]]] = {
     case Filter(Embed(src), BoolLit(true)) => some(src)
-    case Map(Embed(src), mf) if mf ≟ HoleF => some(src)
+    case Map(Embed(src), mf) if mf.linearize ≟ HoleF => some(src)
```
djspiewak: We have an `Equal` for `RecFreeMap`, right? So we should be able to do this sort of thing without linearizing.

connector/src/main/scala/quasar/qscript/RecFreeS.scala
jsantos17: Oh snap! This explains a lot. Fixing!

connector/src/main/scala/quasar/qscript/QScriptCore.scala
jsantos17: Working on it!

connector/src/main/scala/quasar/qscript/rewrites/Rewrite.scala
jsantos17: Yes we can. I `linearize`d this particular `FreeMap` because I figured the linearizing only affects analysis and not code generation.

jsantos17:
@djspiewak I reduced usage of `linearize` in `Rewrite` itself and `Coalesce`. Fixed the `implicit` that got lost too.

Is there anything else preventing this from being merged? There's still some non-trivial `linearize`ing we can avoid, but I'd prefer to address that in a separate PR.

jsantos17:
@djspiewak btw I added a honest-to-god `RenderTree` instance [here](https://github.com/quasar-analytics/quasar/pull/3543/files#diff-efe94c99061d321819bc4d91ca4e5d47R93).

djspiewak:
Trivial merge conflict.

jsantos17:
@djspiewak merge conflict fixed.

djspiewak:
More merge conflicts!  Also I think this should be `version: breaking` due to the changes to the internal API.

jsantos17:
@djspiewak this is green now and merge conflicts (caused by myself merging a diff. PR) have been resolved.

PR #3544, opened 2018-04-06
FuncHandler as AlgebraM

rintcius:
#qz-3330 Done


mongodb/src/main/scala/quasar/physical/mongodb/MongoDbPlanner.scala
```diff
@@ -843,9 +698,11 @@ object MongoDbPlanner {
   def getExpr[
     T[_[_]]: BirecursiveT: ShowT,
     M[_]: Monad: ExecTimeR: MonadFsErr, EX[_]: Traverse: Inject[?[_], ExprOp]]
```
wemrysi: Probably no longer need `ExecTimeR` or `MonadFsErr` here either, guessing this is true for other signatures in this file.

PR #3545, opened 2018-04-06
Fix JSON format of flattenNestedMap.test

jsantos17:
#qz-3783 done

### DIFF
--- a/it/src/main/resources/tests/demo/nestedreductionsonnull.test
+++ b/it/src/main/resources/tests/demo/nestedreductionsonnull.test
@@ -4,12 +4,12 @@
   "backends": {
     "marklogic_json":    "skip",
     "marklogic_xml":     "skip",
-    "mimir":             "ignoreFieldOrder",
-    "mongodb_3_2":       "ignoreFieldOrder",
-    "mongodb_3_4":       "ignoreFieldOrder",
-    "mongodb_3_4_13":    "ignoreFieldOrder",
-    "mongodb_3_6":       "ignoreFieldOrder",
-    "mongodb_read_only": "ignoreFieldOrder"
+    "mimir":             "pending",
+    "mongodb_3_2":       "pending",
+    "mongodb_3_4":       "pending",
+    "mongodb_3_4_13":    "pending",
+    "mongodb_3_6":       "pending",
+    "mongodb_read_only": "pending"
   },
   "query": "SELECT AVG(weight) AS measure, city AS category, gender AS stack, null AS parallel FROM patients AS card468278e5ec90410e8d702ed8f6967066_ WHERE true GROUP BY null, gender, city ORDER BY null ASC, gender ASC, city ASC",
   "predicate": "exactly",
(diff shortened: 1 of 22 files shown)


djspiewak:
TIL @djspiewak doesn't understand JSON syntax.

alissapajer:
@jsantos17 `flattenNestedMap.test` came to life for mimir so that's cool? I don't understand how it has been passing for other connectors previously though. Errors are being swallowed in strange scary ways.
```
[error]     x flatten a map inside a field projection [./flattenNestedMap.test] (183 ms)
[error]      Fixed now, you should remove the 'pendingUntilFixed' marker (QuasarSpecification.scala:77)
```
And then a couple other tests are actually just failing now. And I don't have an explanation for this yet...
```
[error]     x nestedreductionsonnull [./demo/nestedreductionsonnull.test] (40 seconds, 531 ms)
[error]      ; Actual result does not match expected value. 
[error]      
[error]       Diff: 
[error]       Obj
[error]      ╰─ Map
[error]         ├─ Key("parallel")
[error]         │  ╰─ Null
[error]         ├─ Key("measure")
[error]         │  ╰─ Dec(116.0)
[error]         ├─ Key("stack")
[error]         │  ╰─ Str("female")
[error]         ╰─ Key("category")
[error]            ├─ >>> Str("ABERDEEN PROVING GROUND")
[error]            ╰─ <<< Str("ABERDEEN") 
```

jsantos17:
@alissapajer me neither. Not sure what's going on. Confirming master is green locally at the moment. Will report back.

alissapajer:
Yeah, if this diff is actually what caused the other tests to fail, then the error reporting from our integration test suite is quite messed up. :scream:

jsantos17:
Interestingly, without this change, the test suite doesn't run locally. 😱

jsantos17:
@djspiewak @alissapajer 

So it looks like we had some real failures unreported because of the invalid JSON. Looks like the test suite gets to the test with invalid JSON, fails and then the build is green. Fixing the JSON uncovered the other problems.

I'm not sure I follow the meaning behind the null-related tests, but Mongo is not return the expectations. Hence marked as `pending`.

alissapajer:
:+1: I created https://slamdata.myjetbrains.com/youtrack/issue/qz-3779 to track these two failing `null`-related tests.

@jsantos17 do you have evidence that the integration tests that were supposed to run _after_ the json parsing failed were run at all? Or I wonder, did we just silently give up on all subsequent tests after the json parsing failed?

jsantos17:
@alissapajer still investigating. I'll report back when I get some more info. Looks like `postgres` was silently failing a bunch of tests too. Hence the red build. We really need to audit the test suite to prevent this kind of problems in the future.

jsantos17:
@alissapajer with regard to your question: I don't think we were running *any* tests due to the failure to parse JSON. 

It seems to me that even green builds skipped the test suite.

Check out these two logs for green builds:

https://api.travis-ci.org/v3/job/364626419/log.txt
https://api.travis-ci.org/v3/job/364626423/log.txt

I don't see the test suite running in either of them.

alissapajer:
@jsantos17 I looked a those builds and I agree :scream:

alissapajer:
Confirmed that `StreamingQueryRegressionSpec` is running for the integration test jobs.

PR #3547, opened 2018-04-09
Increased travis_wait timeout. This is a good idea.

djspiewak:
There is no way I will regret this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ env:
   global:
     - COURSIER_PROGRESS=0
     - SBT=./sbt
+    - BUILD_HACK_TIMEOUT=60
     # DISCORD_WEBHOOK_TOKENS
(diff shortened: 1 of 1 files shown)


alissapajer:
Mandatory :scream:

nicflores:
@djspiewak,
What travis/build error is this suppose to fix?

alissapajer:
@nicflores this is "fixing" the travis timeouts we get while compiling `connector`. I'm going to break `connector` into subprojects because the compile times are out of hand, but this is still fine to land imo.

nicflores:
Thanks @alissapajer. Merging...

rintcius:
Hmm, isn't this supposed to avoid build failures like https://travis-ci.org/quasar-analytics/quasar/jobs/365050824 ?
Looks like the property was set, but still terminating the build after 10 mins:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

PR #3551, opened 2018-04-12
Refactor MongoDbPlanner

rintcius:
#qz-3771 Done


mongodb/src/main/scala/quasar/physical/mongodb/planner/getFilterBuilder.scala
```diff
@@ -0,0 +1,129 @@
...
+import quasar.physical.mongodb.planner.selector._
+import quasar.qscript._
+
+import matryoshka._
+import matryoshka.data._
+import matryoshka.implicits._
+import matryoshka.patterns._
+import scalaz._, Scalaz._
+
+object getFilterBuilder {
```
jsantos17: Can we merge this with `selector`? They're very closely related.

jsantos17: Can we merge this with `selector`? They're closely related.

rintcius: My reasoning to put it in a separate file is that on one hand `getfilterBuilder` fits both with `selector.scala` (because, like you say, they are closely related) and `workflow.scala` (with respect to the level it's operating on, e.g. compare signatures with the other builders), but on the other hand it also doesn't fit that well with either of those: `getFilterBuilder` operates on a level above the functions of `selector` & it uses dependencies that the functions in `workflow` don't use.

Also merging it with either of them would make the compilation unit unnecessarily big. I guess it's a personal thing but I prefer opening a new file above scrolling up and down. For me, the size of this function is big enough to warrant its own file

PR #3554, opened 2018-04-13
Fix LocalDateTime property test in mongodb std lib 

rintcius:
#qz-3778 Done

### DIFF
--- a/mongoIt/src/test/scala/quasar/physical/mongodb/MongoDbStdLibSpec.scala
+++ b/mongoIt/src/test/scala/quasar/physical/mongodb/MongoDbStdLibSpec.scala
@@ -34,6 +34,7 @@ import quasar.time.TemporalPart
 import scala.sys
 
 import java.time.{Instant, OffsetDateTime => JOffsetDateTime, ZoneOffset}
+import java.time.temporal.ChronoUnit
 
 import matryoshka.{Hole => _, _}
 import matryoshka.data.Fix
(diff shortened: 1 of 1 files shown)


mongoIt/src/test/scala/quasar/physical/mongodb/MongoDbStdLibSpec.scala
```diff
@@ -84,7 +85,8 @@ abstract class MongoDbStdLibSpec extends StdLibSpec {
         Data.Str(time.toString)
       case Data.LocalDateTime(time) =>
         Data.OffsetDateTime(JOffsetDateTime.of(
-          time.withNano(scala.math.round(time.getNano.toDouble / 1000000).toInt * 1000000), // round to millis
+          time.truncatedTo(ChronoUnit.SECONDS)
+            .plusNanos(scala.math.round(time.getNano.toDouble / 1000000).toLong * 1000000), // round to millis
```
alissapajer: What's the difference between `withNano` versus `truncatedTo` and `plusNanos`? I thought that `withNano` would be equivalent to those two. What difference did you find?

rintcius: just a difference in a corner case: `withNano(1000000000)` fails with `Invalid value for NanoOfSecond (valid values 0 - 999999999): 1000000000` whereas `truncatedTo` & `plusNanos` allow it

alissapajer: :+1: could you add a comment here that says that? Otherwise someone might revert it in the future.

PR #3568, opened 2018-04-24
Re-org integration tests

alissapajer:
Re-org integration tests, removing the `qa` directory in favor of directories such a `temporal` and `joins`. This still isn't an ideal organization of these tests, as many tests fall into multiple categories, for example temporal joins :) Ideally we could have all the test files at the same file system hierarchy level, but then add some additional json to the `.test` files tagging them with their types, for example:
```
"types": ["temporal", "join", "reduce"]
```
Then we could query the json files to find all tests of a certain type.

Also deletes duplicate `qa/largeZips.data` file.


alissapajer:
See https://slamdata.myjetbrains.com/youtrack/issue/qz-3796 for organization of integration tests idea flushed out.

alissapajer:
:+1: to using tags on the benchmark server.

PR #3569, opened 2018-04-25
New Quasar client API

wemrysi:
New client APIs and associated types for Quasar, authored with @nicflores.

The intent is that the repl and web layers are implemented in terms of this api, there will be additional abstractions introduced to enable implementations of these types as needed.


api/src/main/scala/quasar/api/ConflictResolution.scala
```diff
@@ -0,0 +1,66 @@
...
+package quasar.api
+
+import slamdata.Predef.{Int, Some}
+
+import scalaz.{Enum, Show}
+import scalaz.std.anyVal._
+import scalaz.syntax.order._
+
+sealed trait ConflictResolution {
+  def fold[A](preserve: => A, replace: => A): A =
```
edmundnoble: This is just my personal opinion, but I think if you introduce a `fold` function, you might as well just use booleans instead of a custom type because your boolean blindness is actually even *worse*; there's no way you'll always use the continuation you want. No need to be afraid of pattern matching.

api/src/main/scala/quasar/api/DataSourceError.scala
```diff
@@ -0,0 +1,87 @@
...
+  sealed trait ExternalError[C] extends DataSourceError[C]
+
+  final case class DataSourceUnsupported[C](kind: MediaType, supported: ISet[MediaType])
+      extends ExternalError[C]
+
+  final case class MalformedConfiguration[C](kind: MediaType, config: C, reason: String)
+      extends ExternalError[C]
+
+  final case class InvalidConfiguration[C](kind: MediaType, config: C, reasons: NonEmptyList[String])
+      extends ExternalError[C]
```
edmundnoble: What is the difference between these two above?

api/src/main/scala/quasar/api/DataSourceError.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  sealed trait StaticError[C] extends DataSourceError[C]
+
+  final case class MalformedContent[C](reason: String)
+      extends StaticError[C]
+
+  sealed trait CreateError[C] extends ExternalError[C] with StaticError[C]
+
+  final case class DataSourceExists[C](name: ResourceName)
```
edmundnoble: Considered `DataSourceAlreadyExists`? This doesn't look like an error constructor at first glance.

api/src/main/scala/quasar/api/DataSources.scala
```diff
@@ -0,0 +1,80 @@
...
+    * or an error if it doesn't exist.
+    */
+  def lookup(name: ResourceName): F[CommonError[C] \/ (StaticDataSource \/ (ExternalMetadata, C))]
+
+  /** Removes the specified datasource, making its data unavailable. If the
+    * specified datasource is static, its content is deleted.
+    *
+    * An error is returned if no datasource exists having the specified name.
+    */
+  def remove(name: ResourceName): F[CommonError[C] \/ Unit]
```
edmundnoble: What is a static datasource? We aren't allowing deleting databases, right?

api/src/main/scala/quasar/api/ExternalStatus.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import monocle.Prism
+import scalaz.{Cord, Show}
+import scalaz.std.option._
+import scalaz.std.string._
+import scalaz.syntax.show._
+
+sealed trait ExternalStatus
+
+object ExternalStatus extends ExternalStatusInstances {
```
edmundnoble: This is an interesting type, but I can't help but wonder why not use `new Exception(message, cause)`.

api/src/main/scala/quasar/api/QueryEvaluator.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package quasar.api
+
+import quasar.api.ResourceError.CommonError
+
+import scalaz.\/
+
+trait QueryEvaluator[F[_], Q, R] extends ResourceDiscovery[F] {
```
edmundnoble: Hahaaaaa yes. This is starting to look like the benchmark runner.

api/src/main/scala/quasar/api/ResourceAccess.scala
```diff
@@ -0,0 +1,32 @@
...
+package quasar.api
+
+import slamdata.Predef.Byte
+import quasar.api.ResourceError.ReadError
+
+import fs2.Stream
+import scalaz.\/
+
+/** Provides for accessing the content of resources in various formats. */
+trait ResourceAccess[F[_], G[_]] extends ResourceDiscovery[F] {
```
edmundnoble: I hope that once we're using fs2 proper, we can remove `G[_]`.

api/src/main/scala/quasar/api/ResourceDiscovery.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  /** Returns the children of the specified resource path or an error if it
+    * does not exist.
+    */
+  def children(path: ResourcePath): F[CommonError \/ IMap[ResourceName, ResourcePathType]]
+
+  /** Returns the descendants of the specified resource path or an error if it
+    * does not exist.
+    */
+  def descendants(path: ResourcePath): F[CommonError \/ Tree[(ResourceName, ResourcePathType)]]
```
edmundnoble: Can we have something more like this, as we discussed in discord:
```scala
def children(path: ResourcePath): F[CommonError \/ IMap[ResourceName, ResourcePathType]] \/
                                  F[CommonError \/ Tree[(ResourceName, ResourcePathType)]]
```

api/src/main/scala/quasar/api/ResourceDiscovery.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  /** Returns the type of the specified resource path or an error if it
+    * does not exist.
+    */
+  def pathType(path: ResourcePath): F[CommonError \/ ResourcePathType]
```
edmundnoble: Can we have a version of this which accommodates data stores that you can't query for path types? Got S3 in mind.

api/src/main/scala/quasar/api/ResourceError.scala
```diff
@@ -0,0 +1,55 @@
...
+  sealed trait ReadError extends ResourceError
+
+  final case class MediaTypeUnsupported(unsupported: MediaType, supported: ISet[MediaType])
+      extends ReadError
+
+  final case class NotAResource(path: ResourcePath) extends ReadError
+
+  sealed trait CommonError extends ReadError
+
+  final case class PathNotFound(path: ResourcePath) extends CommonError
```
edmundnoble: What's the difference between NotAResource and PathNotFound?

api/src/main/scala/quasar/api/ResourceName.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package quasar.api
+
+import slamdata.Predef.String
+
+import scalaz.{Order, Show}
+import scalaz.std.string._
+
+final case class ResourceName(value: String) extends scala.AnyVal
```
edmundnoble: `AnyVal` is satanic. Every occurrence in a generic position results in *more* boxing than if you don't use it; if we don't do opaque type-style newtypes, I'd just remove `extends AnyVal`.

api/src/main/scala/quasar/api/ResourcePath.scala
```diff
@@ -0,0 +1,74 @@
...
+import quasar.contrib.pathy.{AFile, APath}
+import quasar.fp.ski.{ι, κ}
+
+import monocle.Prism
+import pathy.Path._
+import scalaz.{Cord, Order, Show}
+import scalaz.syntax.show._
+
+/** Identifies a resource in a datasource. */
+sealed trait ResourcePath extends Product with Serializable {
```
edmundnoble: Thanks for this.

api/src/main/scala/quasar/api/ResourcePathType.scala
```diff
@@ -0,0 +1,60 @@
...
+import scalaz.{Cord, ISet, OneAnd, Order, Show}
+import scalaz.std.option._
+import scalaz.syntax.show._
+
+/** Describes what a `ResourcePath` refers to. */
+trait ResourcePathType extends Product with Serializable
+
+object ResourcePathType extends ResourcePathTypeInstances {
+  case object ResourcePrefix extends ResourcePathType
+  final case class Resource(formats: OneAnd[ISet, MediaType]) extends ResourcePathType
```
edmundnoble: Is `OneAnd[ISet, MediaType]` a `NonEmptySet`? What if the first element is also present in the set?

build.sbt
```diff
@@ -285,6 +285,16 @@ lazy val foundation = project
   .settings(excludeTypelevelScalaLibrary)
   .enablePlugins(AutomateHeaderPlugin, BuildInfoPlugin)
 
+/** Types and interfaces describing Quasar's functionality. */
+lazy val api = project
```
edmundnoble: :+1:

api/src/main/scala/quasar/api/ConflictResolution.scala
djspiewak: I generally agree (re: pattern matching), though the fold isn't *quite* as bad as an `if`/`else` since you can use named parameters to make things more clear.

api/src/main/scala/quasar/api/ResourceAccess.scala
djspiewak: We can't. This is the solution to the coproduct hack that we needed in mimir's `FreeVFS`.

api/src/main/scala/quasar/api/ResourceName.scala
djspiewak: Are we using it in a generic context?

api/src/main/scala/quasar/api/ConflictResolution.scala
edmundnoble: Yeah, my only issue with that is just that you *can* use named parameters, and aren't forced to.

api/src/main/scala/quasar/api/ResourceName.scala
edmundnoble: https://github.com/quasar-analytics/quasar/pull/3569/files#diff-147f92a10da81398b37f23e23edb2b25R76, also in `children`.

api/src/main/scala/quasar/api/ResourceAccess.scala
edmundnoble: Ahhhh. Gotcha.

api/src/main/scala/quasar/api/ConflictResolution.scala
wemrysi: @edmundnoble Agree that `fold` can result in blindness at call sites without named params, disagree that this negates all the value of the custom type (its not interchangeable with other booleans and self-documents, to name two).

Anyhow, I'm fine with getting rid of `fold` since the types don't force correct usage, it is there mostly as a convenience.

api/src/main/scala/quasar/api/DataSourceError.scala
wemrysi: "Malformed" is structural, "Invalid" is semantic.

api/src/main/scala/quasar/api/DataSourceError.scala
wemrysi: I didn't, but if others find that preferable, :+1: (though doesn't "exists" imply "already"?).

api/src/main/scala/quasar/api/DataSources.scala
wemrysi: A static datasource consists of data uploaded to quasar by a user. As opposed to a datasource consisting of a connection to an external store.

api/src/main/scala/quasar/api/ExternalStatus.scala
wemrysi: Are you suggesting `final case class Error(ex: Exception) extends ExternalStatus`? or replacing `ExternalStatus` entirely? If the latter, what about `Ok`? I could be open to the former, but if there is no cause exception then a user is forced to construct a `RuntimeException` or similar just to provide a message.

api/src/main/scala/quasar/api/ConflictResolution.scala
edmundnoble: True, it doesn't negate *all* of the utility, that's for sure.

api/src/main/scala/quasar/api/ExternalStatus.scala
edmundnoble: I'd probably say `ExternalStatus` could just be `Exception \/ Unit`. Though, really, I'd love it if we had a custom type for `E \/ Unit`; it makes me uneasy.

api/src/main/scala/quasar/api/ResourceDiscovery.scala
wemrysi: Yes, I think actually having this just be
```
def exists(path: ResourcePath): F[Option[OneAnd[ISet, MediaType]]]
```
would be better, as the intent is to answer the question of "does this path refer to a resource".

api/src/main/scala/quasar/api/ResourceName.scala
djspiewak: Meh… You're correct. I guess the one nice thing about the `extends AnyVal` is it does effectively document that this is *intended* to be a newtype, even if scalac isn't going to optimize it as such.  Neither of those generic contexts are in the hot path, so the performance hit is meaningless, so it really does come down to whether or not the effective documentation of the newtyping is valuable.

api/src/main/scala/quasar/api/ExternalStatus.scala
wemrysi: Hm, if we had a better type for `E \/ Unit` then I'd be good with that. I'll come up with one and give it a go.

api/src/main/scala/quasar/api/ResourceDiscovery.scala
wemrysi: Possibly, but does this sort of thing belong in a client interface? i.e. as a user, if I really do just want the children, then I'm going to have to prune the rhs. Maybe this sort of interface would be better as an implementation trait.

api/src/main/scala/quasar/api/ConflictResolution.scala
alissapajer: :+1: I'd prefer pattern matching over the `fold`.

api/src/main/scala/quasar/api/DataSourceError.scala
alissapajer: The only way that "exists" wouldn't imply "already" is if the data source was created successfully by the call that produced this error, which doesn't make sense since this is a `CreateError`. So I'm fine with keeping the name as is, but if anyone feels strongly for changing it I also won't object.

api/src/main/scala/quasar/api/ConflictResolution.scala
```diff
@@ -0,0 +1,66 @@
...
+
+object ConflictResolution extends ConflictResolutionInstances {
+  case object Preserve extends ConflictResolution
+  case object Replace extends ConflictResolution
+
+  val preserve: ConflictResolution =
+    Preserve
+
+  val replace: ConflictResolution =
+    Replace
```
alissapajer: These are here to get you `Preserve` and `Replace` typed as a `ConflictResolution`?

api/src/main/scala/quasar/api/QuasarError.scala
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package quasar.api
+
+/** The supertype of all Quasar errors.
+  *
+  * TODO: Rename to `QuasarError`
+  */
+trait QuasarErrorNG
```
alissapajer: What does `NG` stand for?

api/src/main/scala/quasar/api/SchemaDiscovery.scala
```diff
@@ -0,0 +1,27 @@
...
+package quasar.api
+
+import slamdata.Predef.Option
+
+/** Provides a facility for discovering the structure of datasets. */
+trait SchemaDiscovery[F[_], Q, C, S] {
+  /** Returns a description of the structure of the dataset represented
+    * by the given query.
+    */
+  def schema(query: Q, compressionSettings: C): F[Option[S]]
```
alissapajer: Having this param specified by name as `compressionSettings` seems like it's exposing an implementation detail of our particular schema discovery tool. Maybe just name this param `settings`?

api/src/main/scala/quasar/api/ResourcePathType.scala
alissapajer: `OneAnd` is neat, but I think a custom type would be more useful here to avoid strangeness like set semantics not holding.

alissapajer: @wemrysi did you have an example in mind of a `Resource` with multiple `MediaType`s? I wouldn't want to allow for multiple if all of our use cases will only have one.

api/src/main/scala/quasar/api/ResourceError.scala
alissapajer: At first I was thinking that this could be solved by requiring more precise types for the params of `children` and `read`. So:
```scala
def children(path: ResourcePathPrefix)
def read(format: MediaType, path: ResourcePathLeaf)
```
But I then realized that we don't know if a `ResourcePath` is a "prefix" or a "leaf" until we actually ask the connector. An alternate approach would be to ask the connector, but that adds in another API call and another layer of complexity. So I'm not convinced that's the way to go, just was thinking about it.

alissapajer: I do think the distinction between `NotAResource` and `PathNotFound` is meaningful and will allow us to provide more specific errors to the user. Otherwise if we had a single error type for these two cases, we'd have to special case the errors to the user.

api/src/main/scala/quasar/api/ResourceDiscovery.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package quasar.api
+
+import quasar.api.ResourceError.CommonError
+
+import scalaz.{\/, IMap, Tree}
+
+/** Provides for discovering the existence and types of resources in a datasource. */
+trait ResourceDiscovery[F[_]] {
```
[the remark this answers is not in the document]
alissapajer: Plus connection management.

api/src/main/scala/quasar/api/ResourcePathType.scala
wemrysi: @alissapajer For datasources that aren't mimir, I think there is always at least two possible: `eprec` and whatever the native format of the resource is.

For example, an s3 bucket full of xml would return `application/vnd.quasar.dataset.eprec` and `application/xml` as options for an object therein.

However, there is also the option of making queries the _only_ means of retrieving data and just dropping `ResourceAccess` altogether. Then the only format would be whatever queries use (currently planned to be `eprec`). We could possibly extend this type later on to support "medium" weight connectors and have formats other than `eprec` for when a query is executable completely within the connector (no mimir needed), but I'm not even sure how often that will be or if anyone will care about results in those other formats.

api/src/main/scala/quasar/api/ConflictResolution.scala
wemrysi: Aye.

api/src/main/scala/quasar/api/QuasarError.scala
wemrysi: "Next Generation"

alissapajer:
@wemrysi did you have more changes you plan to push or have you pushed what you planned to?

api/src/main/scala/quasar/api/ResourceDiscovery.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  /** Returns the children of the specified resource path or an error if it
+    * does not exist.
+    */
+  def children(path: ResourcePath): F[CommonError \/ IMap[ResourceName, ResourcePathType]]
```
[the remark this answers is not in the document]
wemrysi: Yeah, though this is why I was suggesting an implementation trait as you could also have a default implementation of `descendants` that recursively calls `children`.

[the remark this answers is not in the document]
alissapajer: :+1: sounds good

wemrysi:
@alissapajer Ok, I think I've addressed everything, should be ready for a final review.

api/src/main/scala/quasar/api/Condition.scala
```diff
@@ -0,0 +1,139 @@
...
+    Condition.pAbnormal.modify(f)(this)
+
+  def orElse(other: => Condition[E]): Condition[E] =
+    this match {
+      case Condition.Abnormal(_) => this
+      case Condition.Normal()    => other
+    }
+}
+
+object Condition extends ConditionInstances {
```
alissapajer: `Condition` is general enough that it could live in `foundation` rather than in `api`.

wemrysi: Good point :+1:

wemrysi:
@alissapajer Actually, sorry, I just realized that without `ResourceAccess`, `ResourceDiscovery` becomes a bit more simple (as we don't need `MediaType`s there any longer). Let me make that change before merging.

api/src/main/scala/quasar/api/MediaTypes.scala
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.api
+
+import scalaz.{ISet, Order, Semigroup, Show}
+import scalaz.syntax.show._
+
+final class MediaTypes private (val element: MediaType, rest: ISet[MediaType]) {
```
alissapajer: There is nothing special about `MediaType` here, right? So instead this could be
```scala
final class NonEmptySet[A] private (val element: A, rest: ISet[A])
```
and also moved to `foundation`?

wemrysi: Maybe, but it doesn't really have any set operations (and didn't really want to add them), the idea was that if you wanted those you'd just `toISet` and go from there.

However, I'm actually about to delete this.

alissapajer:
@wemrysi I might have missed the conversation about this: why was `ResourceAccess` removed?

wemrysi:
@alissapajer After some [discussion with @beckyconning](https://github.com/slamdata/research/pull/3#issuecomment-384982245), it looks like querying is the only interface we need to resources, for now at least.

wemrysi:
@alissapajer Ok, I think we're set. There were a few more changes (mostly deletions) resulting from removing `ResourceAccess`.

alissapajer:
I'll merge when green!

PR #3570, opened 2018-04-30
Selector refactoring

rintcius:
Refactoring part of an attempt to fix #qz-3679


alissapajer:
@rintcius try merging in master to avoid the cache uploading timeouts, just merged #3571.

rintcius:
@alissapajer @nicflores still timeouts, now on downloading the cache - I have added `travis_wait 30` to download calls as well (in this PR)

alissapajer:
@rintcius can you PR the `travis_wait` commit to master? That way it can get in sooner for other PRs as well.

mongodb/src/main/scala/quasar/physical/mongodb/planner/QScriptCorePlanner.scala
```diff
@@ -170,22 +186,23 @@ class QScriptCorePlanner[T[_[_]]: BirecursiveT: EqualT: ShowT] extends
...
-            case \&/.This(js) => WB.filter(src0, Nil, {
-              case Nil => Selector.Where(js(jscore.ident("this")).toJs)
-            })
-          }
-      }
+      ((selectors |@| typeSelectors.point[M])((s, t) =>
+        (s.toOption, t.toOption) match {
+          case (None, Some(typeSel)) => filterBuilder(src0, typeSel)
+          case (Some(sel), None) => filterBuilder(src0, sel)
+          case (Some(sel), Some(typeSel)) => filterBuilder(src0, typeSel) >>= (filterBuilder(_, sel))
```
rintcius: This case is hit by the failing test. See https://gist.github.com/rintcius/9f9e905c38afbe7c1b6964a80aebaf3c#file-gistfile1-txt-L292 for the query (2nd revision is mongo 3.6/new/incorrect, 1st revision is mongo3.4/old/correct). 

I see the problem, but I'm not sure how to fix this. The vars being used inside the `$expr` in l.292-448 all miss a `src` prefix, eg `$author` needs to be `$src.author`.
I _think_ what's wrong above is that the `sel` should depend on the `typeSel` but currently they are independent: sometimes it needs an extra `rebaseWB` though to correct the `cond` a la

```
        rebaseWB[T, M, WF, EX](cfg, projSrc, _) >>=
          (s => filterBuilder(s, sel))
```
where `val projSrc: FreeQS[T] = dsl.free.Map(dsl.free.Hole, recFunc.ProjectKeyS(recFunc.Hole, "src"))`. The problem is to decide _when_ that needs to be added exactly though.

Or alternatively, can the implemention of `exprSelector` be improved such that `selectors` can still be kept independent of `typeSelectors`? See for the implementation https://github.com/quasar-analytics/quasar/pull/3570/files#diff-d88e6e29b6974cef6eb0d5beebd5d2e9R69 & for the call https://github.com/quasar-analytics/quasar/pull/3570/files#diff-6418a64b7d5149da15342803ff0cde6fR167

mongodb/src/main/scala/quasar/physical/mongodb/planner/QScriptCorePlanner.scala
```diff
@@ -152,40 +153,49 @@ class QScriptCorePlanner[T[_[_]]: BirecursiveT: EqualT: ShowT] extends
...
+          partialSel,
+          fm)
+
+      ((selectors(cond.linearize) |@| typeSelectors(cond.linearize).point[M])(
+        {
+          case (None, Some(typeSel)) => filterBuilder(src0, typeSel, cond.linearize)
+          case (Some(sel), None) => filterBuilder(src0, sel, cond.linearize)
+          case (Some(sel), Some(typeSel)) =>
+            filterBuilder(src0, typeSel, cond.linearize) >>= (filterBuilder(_, sel, cond.linearize))
+          case (None, None) => src0.point[M]
```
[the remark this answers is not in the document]
wemrysi: I understand `(None, None)` as "we were unable to plan the predicate using selectors". That would seem to mean we need to plan it via an alternative method, like what was being done before. Either via the project -> match approach or map-reduce.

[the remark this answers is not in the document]
rintcius: Hmm, in that interpretation the 4th case looks inconsistent to the other 3 cases. In the first 3 cases I can only interpret `None` as "no filter necessary" because we are just omitting the filters there (I also thought you were referring to the "no filter necessary" interpretation when we discussed the 1st case not being hit in any of our ITs, but that it could theoretically).

In a consistent "unable to plan" interpretation I'd expect there to be only 2 cases:
```
          case (Some(sel), Some(typeSel)) =>
            filterBuilder(src0, typeSel, cond.linearize) >>= (filterBuilder(_, sel, cond.linearize))
          case  _ => old code
```

Anyway, I can look into that selector code in more detail, next week. I can imagine that we need to represent both the "unable to plan the predicate using selectors" situation and the "no filter necessary" situation.

rintcius:
Need to start working on Mimir now. I wrote the current status in qz-3679. 

I still would like to create a PR with just the refactoring part of this. I'll revert the change that I made to the 4th case of the pattern match, because I want to exclude any behaviour changes for now. 
@jsantos17 @wemrysi are you ok with this?

As an aside, there are at least a couple of cases that return `None` with the meaning "unable to plan", but I still think that the current pattern match is inconsistent with it.

jsantos17:
@rintcius sounds good. I'll look into your comment. Thanks for the write up!

mongodb/src/main/scala/quasar/physical/mongodb/mapreduce.scala
```diff
@@ -19,6 +19,7 @@ package quasar.physical.mongodb
 import slamdata.Predef._
 import quasar.common.SortDir
 import quasar.javascript._
+import quasar.physical.mongodb.selector.Selector
```
jsantos17: Why was this import added?

mongodb/src/main/scala/quasar/physical/mongodb/selector/selector.scala
```diff
@@ -51,9 +53,9 @@ sealed abstract class Selector {
     }
 
   def negate: Selector = {
-    def expr(x: Selector.SelectorExpr): Selector.SelectorExpr = x match {
-      case Selector.Expr(cond) => Selector.NotExpr(cond)
-      case Selector.NotExpr(cond) => Selector.Expr(cond)
+    def expr(x: Selector.CondExpr): Selector.CondExpr = x match {
+      case Selector.CExpr(cond) => Selector.NotCExpr(cond)
```
jsantos17: What does `CExpr` stand for? Can we expand it into something more descriptive? I'm familiar with this part of the codebase and still find it hard to know what `CExpr` is.

mongodb/src/test/scala/quasar/physical/mongodb/OptimizeSpec.scala
```diff
@@ -22,6 +22,7 @@ import quasar.common.SortDir
 import quasar.physical.mongodb.accumulator._
 import quasar.physical.mongodb.expression._
 import quasar.physical.mongodb.optimize.pipeline._
+import quasar.physical.mongodb.selector.Selector
```
jsantos17: Why these imports?

mongodb/src/main/scala/quasar/physical/mongodb/mapreduce.scala
rintcius: This import is added because I moved `Selector` into its own package

mongodb/src/main/scala/quasar/physical/mongodb/selector/selector.scala
rintcius: I renamed it because of `Expr` being added (and then removed). `CExpr` & `NotCExpr` are extending `CondExpr` and `C` stands for `Cond` (I would have used `Cond` but that gives a duplicate with the parent). 
What do you think about renaming the parent to `ConditionExpr` and then use `CondExpr` & `NotCondExpr`?

rintcius:
@jsantos17 There's a bunch of different things:
- make spec names consistent
- add and use `plan3_6` in planner specs
- move `Selector` into `selector` package
- improve the type defs in `planner.selector`
- make `rewriteRefs` monadic (`rewriteRefsM`)
- add some `RenderTree` constraints
- deduplicate `filterBuilder`
- swap args in functions like this: 
```
  def processMapFuncExpr
    [T[_[_]]: BirecursiveT: ShowT, M[_]: Monad, EX[_]: Traverse, A]
    (funcHandler: AlgebraM[M, MapFunc[T, ?], Fix[EX]], staticHandler: StaticHandler[T, EX])
    (fm: FreeMapA[T, A])
    (recovery: A => Fix[EX])
```
I placed `fm` after `recovery` to make them more suitable for calling them point-free style. In general: most stable args come first.

re. dirty commit history, that's true. I'll clean it up & open a new PR after we have agreed on the new names for `CExpr`

mongodb/src/main/scala/quasar/physical/mongodb/selector/selector.scala
jsantos17: Sounds good!

jsantos17:
@rintcius sounds good. thanks for the explanation!

rintcius:
Superseded by https://github.com/quasar-analytics/quasar/pull/3599

PR #3573, opened 2018-05-02
Publish test classes of `quasar-fs-internal`

jedesah:


alissapajer:
Why do we need this?

jedesah:
SlamData Backend depends on some of these classes

PR #3575, opened 2018-05-02
DataSources and their modules

wemrysi:
Introduces `DataSource` and {light,heavy}weight versions along with their respective modules.

This is intended to be the new connector API, replacing `BackendModule` and associated types. The idea being one can implement either of the modules and we'll load both kinds at startup.

Not included in this PR, but planned before release
* Logging
* Replace `Stream[F, Data]` with `Stream[F, ByteVector]` once `EPrec` is finished
* Possibly a new `QScript` variant that uses `ResourcePath` as the `Read` type.


wemrysi:
@edmundnoble Yes to both :+1:

edmundnoble:
The problem is really with the signature `F[Stream[F, A]]`, and even more so `F[ReadError \/ Stream[F, A]]`. After I've executed the outer `F[_]`, there is *almost always* a finalizer associated with the `Stream` that *has* to run after. For example, implementing `read` for the S3 connector using http4s, I only make a single HTTP request that attempts to read the file, but still returns `F[Option[Stream[F, A]]]` (analogous to `F[ReadError \/ Stream[F, A]]`). But the lifetime of the inner stream is *tied* to the outer `Task`. If you store `Stream[F, A]` expecting it to be possible to use later, or just throw it away, the connection is likely to both a) stay open longer than you want and b) time out before you do eventually use it.

The type we really want is `Stream[F, A]`, even after inlining things. The error must be provided *in* the stream; once you start streaming, you can listen for errors, because a) it ties the lifetime of the stream to the lifetime of the request, and b) we don't want the weird case where the outer `F` fails and the finalizer isn't run.

wemrysi:
@edmundnoble I see what you're saying. The problem we've had in the past with `Stream[F, A]` is when using it with something like `http4s`, you have to "uncons" the stream anyway to evaluate the "acquire" effect to see what the response type should be, as once you've written out the body, it's too late to respond differently if there was an error, say, planning the query that resulted in the stream.

The idea with `F[Stream[F, A]]` is to allow users to be explicit with this distinction, though they must be careful to avoid the scenario you describe.

Would `F[Stream[G, A]]` be an improvement here, to enforce the distinction?

edmundnoble:
In that case, perhaps `F[DisposableResponse[G]]`?

edmundnoble:
I don't follow what the `F/G` distinction gives us here though, I thought that was for having `Task` in a `Free` coproduct.

wemrysi:
@edmundnoble In this case, the distinction would only be nominal, more for documentation purposes. As for `F[DisposableResponse[G]]`, I don't follow how that would solve the problem?

wemrysi:
@edmundnoble Speaking of `http4s`, they appear to make a similar distinction https://github.com/http4s/http4s/blob/0bd20db589c7c129dea7745af69ccf9029809da6/core/src/main/scala/org/http4s/package.scala#L29

edmundnoble:
I believe the reason they do that is less for documentation, and more to allow polymorphic middlewares (this also may, for example, allow us to do `QHttpService` in a nicer way eventually).

The reason I suggest `DisposableResponse` is because if the API actually *can't* prevent users from doing the wrong thing with the response, it might be nice if the API made it explicit what they *should* do by providing that in the signature rather than just a `Stream`.

wemrysi:
@edmundnoble I was more referring to the distinction between the effects required to obtain a result vs the effects of the result itself.

Something analogous to `DisposableResponse` sounds good :+1:, I'll add it.

api/src/main/scala/quasar/api/DataSourceError.scala
```diff
@@ -30,17 +30,19 @@ sealed trait DataSourceError[C] extends QuasarErrorNG
...
+      extends InitializationError[C]
+
+  final case class InvalidConfiguration[C](kind: DataSourceType, config: C, reasons: NonEmptyList[String])
+      extends InitializationError[C]
 
   final case class ConnectionFailed[C](message: String, cause: Option[Throwable])
-      extends ExternalError[C]
+      extends InitializationError[C]
 
   sealed trait StaticError[C] extends DataSourceError[C]
```
nicflores: @jedesah and I just realized this, but why does `StaticError` need a `C` if in the creation of a static datasource we don't use/need a config?

api/src/main/scala/quasar/api/DataSourceError.scala
```diff
@@ -30,17 +30,19 @@ sealed trait DataSourceError[C] extends QuasarErrorNG
 object DataSourceError {
```
nicflores: Should `DataSourceError` extend `DataSourceErrorInstances`?

wemrysi: Yep, nice catch!

api/src/main/scala/quasar/api/DataSourceError.scala
wemrysi: It doesn't, but I'm not sure how to have it and `ExternalError` share a common parent without it or forcing the config type to become existential, reducing its utility.

jedesah: Could we make `DataSourceError` covariant so that we could say `StaticError extends DataSourceError[Nothing]` or is there a drawback to that?

wemrysi:
@edmundnoble I've added `Disposable` and used it in the result types of the datasources, let me know what you think.

api/src/main/scala/quasar/api/DataSourceType.scala
```diff
@@ -0,0 +1,50 @@
...
+import scalaz.std.string._
+import scalaz.std.tuple._
+import scalaz.syntax.show._
+import shapeless.{Witness => W}
+
+@Lenses
+final case class DataSourceType(name: DataSourceType.Name, version: Positive)
+
+object DataSourceType extends DataSourceTypeInstances {
+  type Name = String Refined MatchesRegex[W.`"[a-zA-Z0-9-]+"`.T]
```
alissapajer: Why the restrictions on the `Name`?

wemrysi: Since it is more of a categorical designation and not really "data", my thought was to restrict it to allow it to be used verbatim in more contexts.

PR #3577, opened 2018-05-07
Remove unmanaged jar path for jdbm snapshot

alissapajer:
We were never actually putting the jar at that path; instead we manage the jar:
```
"org.apache.jdbm" %  "jdbm" % "3.0-alpha5"
```


alissapajer:
This is technically changing the classpath, even though the resource at this path was never available. So we could assign this as a feature version.

djspiewak:
Revision seems fine to me.

PR #3579, opened 2018-05-09
FederatingQueryEvaluator

wemrysi:
Adds a `QueryEvaluator` capable of executing queries over multiple sources. Handles the routing of operations to sources and ensuring all sources referenced in a query exist.

The actual work of executing a `FederatedQuery` is left up to an instance of `QueryFederation`, to eventually be implemented for `mimir`.

### DIFF
--- a/api/src/main/scala/quasar/api/ResourceDiscovery.scala
+++ b/api/src/main/scala/quasar/api/ResourceDiscovery.scala
@@ -16,7 +16,7 @@
 
 package quasar.api
 
-import slamdata.Predef.Boolean
+import slamdata.Predef.{Boolean, Stream}
 import quasar.api.ResourceError.CommonError
 
 import scalaz.{\/, IMap, Tree}
@@ -32,7 +32,7 @@ trait ResourceDiscovery[F[_]] {
   /** Returns the descendants of the specified resource path or an error if it
     * does not exist.
     */
-  def descendants(path: ResourcePath): F[CommonError \/ Tree[(ResourceName, ResourcePathType)]]
+  def descendants(path: ResourcePath): F[CommonError \/ Stream[Tree[ResourceName]]]
 
   /** Returns whether the specified resource path refers to a resource. */
   def isResource(path: ResourcePath): F[Boolean]
--- a/api/src/main/scala/quasar/api/ResourcePath.scala
+++ b/api/src/main/scala/quasar/api/ResourcePath.scala
@@ -16,14 +16,13 @@
 
 package quasar.api
 
-import slamdata.Predef.{Product, Serializable, Unit}
-import quasar.contrib.pathy.{AFile, APath}
+import slamdata.Predef._
+import quasar.contrib.pathy.{firstSegmentName, rebaseA, stripPrefixA, AFile, APath}
 import quasar.fp.ski.{ι, κ}
 
 import monocle.Prism
 import pathy.Path._
-import scalaz.{Cord, Order, Show}
-import scalaz.syntax.show._
+import scalaz.{Order, Show}
 
 /** Identifies a resource in a datasource. */
 sealed trait ResourcePath extends Product with Serializable {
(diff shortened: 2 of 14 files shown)


alissapajer:
A `Local` file system test failed in appveyor. There is a race condition in `Local` that I have fixed locally, so that's likely what was hit. Since this PR doesn't touch anything related to that code, I restarted the build.

api/src/main/scala/quasar/api/ResourceDiscovery.scala
```diff
@@ -16,7 +16,7 @@
 
 package quasar.api
 
-import slamdata.Predef.Boolean
+import slamdata.Predef.{Boolean, Stream}
```
djspiewak: 😱

Are we *sure* we want to do this???

api/src/main/scala/quasar/api/ResourcePath.scala
```diff
@@ -44,8 +43,20 @@ sealed trait ResourcePath extends Product with Serializable {
         ResourcePath.leaf(rootDir </> file(name.value))
     }
 
+  def /: (name: ResourceName): ResourcePath =
```
djspiewak: Space

api/src/main/scala/quasar/api/ResourcePath.scala
```diff
@@ -44,8 +43,20 @@ sealed trait ResourcePath extends Product with Serializable {
...
+
+      case ResourcePath.Root =>
+        ResourcePath.leaf(rootDir </> file(name.value))
+    }
+
   def toPath: APath =
     fold(ι, rootDir)
+
+  def uncons: Option[(ResourceName, ResourcePath)] =
+    ResourcePath.leaf.getOption(this) map (ResourcePath.unconsLeaf)
```
djspiewak: Space (and lack of `.`)

api/src/test/scala/quasar/api/MockQueryEvaluator.scala
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.api
+
+import slamdata.Predef.{tailrec, Boolean, None, Option, Some, Stream, Unit}
```
djspiewak: Uh, more `Stream`…

api/src/main/scala/quasar/api/ResourceDiscovery.scala
wemrysi: Not particularly, it's just what `scalaz.Tree` uses. We could use `Cofree[List, ResourceName]` or even `Cofree[fs2.Stream[G, ?], ResourceName]`.

djspiewak: Ah, I see. Yeah… bleh.

core/src/main/scala/quasar/FederatingQueryEvaluator.scala
```diff
@@ -0,0 +1,178 @@
...
+/** A `QueryEvaluator` capable of executing queries against multiple sources. */
+final class FederatingQueryEvaluator[T[_[_]]: BirecursiveT, F[_]: Monad, S, R] private (
+    queryFederation: QueryFederation[T, F, S, R],
+    sources: F[IMap[ResourceName, (ResourceDiscovery[F], S)]])
+    extends QueryEvaluator[F, T[QScriptRead[T, ?]], R] {
+
+  // TODO[scalaz]: Shadow the scalaz.Monad.monadMTMAB SI-2712 workaround
+  import WriterT.writerTMonadListen
+
+  def children(path: ResourcePath) =
```
alissapajer: Could you put the return types on the functions here that make up the implementation of `QueryEvaluator`? Even though we can look them up, I think it makes it more readable.

alissapajer:
@wemrysi is there more you planned to push, or is this ready for another look?

wemrysi:
@alissapajer No, I think that's it, ready for another look. If we decide we don't want to use `scalaz.Tree` I'd prefer to swap it out in a separate PR.

PR #3582, opened 2018-05-09
LWC

alissapajer:
#qz-3449 done

### DIFF
--- a/foundation/src/main/scala/quasar/fp/TaskRef.scala
+++ b/foundation/src/main/scala/quasar/fp/TaskRef.scala
@@ -32,9 +32,13 @@ import scalaz.concurrent.Task
 sealed abstract class TaskRef[A] {
   def read: Task[A]
   def write(a: A): Task[Unit]
+
   def compareAndSet(oldA: A, newA: A): Task[Boolean]
+
   def modifyS[B](f: A => (A, B)): Task[B]
   def modifyT[B](f: A => Task[(A, B)]): Task[B]
+  def twoPhaseModifyT[B](modify: A => (Task[A], Task[B])): Task[B]
+
   def modify(f: A => A): Task[A] =
     modifyS(a => f(a).squared)
 }
@@ -59,6 +63,9 @@ object TaskRef {
           back <- if (bool) Task.now(b) else modifyS[B](f)
         } yield back
 
+      /* The effect that results from running `f` may be run multiple
+       * times, in the case that the `compareAndSet` fails to set.
+       */
       @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
       def modifyT[B](f: A => Task[(A, B)]): Task[B] =
         for {
@@ -67,6 +74,21 @@ object TaskRef {
           (a1, b) = tup
           bool <- compareAndSet(a0, a1)
           back <- if (bool) Task.now(b) else modifyT[B](f)
+        } yield back
+
+      /* The effect on the left may be run multiple times, in the
+       * case that the `compareAndSet` fails to set. The effect on
+       * the right will only be run when `compareAndSet` successfully
+       * sets.
+       */
+      @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+      def twoPhaseModifyT[B](modify: A => (Task[A], Task[B])): Task[B] =
+        for {
+          a0 <- read
+          (a1T, bT) = modify(a0)
+	  a1 <- a1T
+          bool <- compareAndSet(a0, a1)
+          back <- if (bool) bT else twoPhaseModifyT[B](modify)
         } yield back
     }
   }
--- a/foundation/src/main/scala/quasar/fp/TaskRef.scala
+++ b/foundation/src/main/scala/quasar/fp/TaskRef.scala
@@ -86,7 +86,7 @@ object TaskRef {
         for {
           a0 <- read
           (a1T, bT) = modify(a0)
-	  a1 <- a1T
+          a1 <- a1T
           bool <- compareAndSet(a0, a1)
           back <- if (bool) bT else twoPhaseModifyT[B](modify)
         } yield back
--- a/fs/src/main/scala/quasar/fs/Local.scala
+++ b/fs/src/main/scala/quasar/fs/Local.scala
@@ -50,13 +50,13 @@ import scalaz.concurrent.Task
   *
   * The implementations are not guaranteed to be atomic.
   */
-class Local private (baseDir: JFile) {
+object Local {
 
-  case class ReadState(
+  final case class ReadState(
     current: Long,
     handles: Map[Long, (AtomicBoolean, Natural, Option[Positive])])
 
-  case class WriteState(
+  final case class WriteState(
     current: Long,
     handles: Set[Long])
 
@@ -534,8 +534,4 @@ class Local private (baseDir: JFile) {
     write compose writeFile,
     resultToTask compose manageFile)
   }
-}
-
-object Local {
-  def apply(baseDir: JFile): Local = new Local(baseDir)
 }
--- a/fs/src/main/scala/quasar/fs/Local.scala
+++ b/fs/src/main/scala/quasar/fs/Local.scala
@@ -21,10 +21,11 @@ import slamdata.Predef._
 import quasar.{Data, DataCodec}
 import quasar.common.PhaseResult
 import quasar.contrib.pathy.{prettyPrint, AFile, APath, PathSegment}
-import quasar.contrib.scalaz.stateT.StateTask
 import quasar.fp.TaskRef
 import quasar.fp.numeric.{Natural, Positive}
 import quasar.fp.ski.κ
+import quasar.fs.ReadFile.ReadHandle
+import quasar.fs.WriteFile.WriteHandle
 
 import java.io.{File => JFile}
 import java.nio.charset.StandardCharsets
@@ -42,7 +43,7 @@ import java.util.stream.Collectors
 import scala.collection.JavaConverters._
 
 import pathy.Path.{file1, refineType, DirName, FileName}
-import scalaz.{~>, \/, -\/, \/-, EitherT, StateT}
+import scalaz.{~>, \/, -\/, \/-, EitherT}
 import scalaz.Scalaz._
 import scalaz.concurrent.Task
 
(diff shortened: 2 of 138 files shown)


alissapajer:
This is only a WIP because if Travis fails to run or fails any of the lwc tests, I plan to rebase. If Travis passes, I'll remove the WIP status.

alissapajer:
Tagged this `version: breaking` mostly because the connection uri for mimir has changed. It has changed is  only because mimir is a lwc itself and I didn't special case it. We could special case it if we want.

alissapajer:
This was developed in parallel with the ongoing file system changes. I'd like to get this in sooner rather than later, porting this over to the new API incrementally after it lands in master. This PR gives us a baseline lwc API with integration tests running. Additionally when this lands we'll have a quasar artifact we can point to for in the S3 and Azure (and other) connectors.

djspiewak:
> We could special case it if we want.

IMO it's worth special-casing, since it's really easy to do and is a very nice QoL, but I won't stamp my feet about it.

mimir/src/main/scala/quasar/mimir/MimirCake.scala
```diff
@@ -37,10 +37,12 @@ object MimirCake {
 
   type Cake = Precog with Singleton
 
-  type MT[F[_], A] = Kleisli[F, Cake, A]
-  type CakeM[A] = MT[Task, A]
+  type CakeMT[F[_], A <: LightweightFileSystem, B] = Kleisli[F, (Cake, A), B]
+  type CakeM[A <: LightweightFileSystem, B] = CakeMT[Task, A, B]
```
djspiewak: Do we actually need the specific typing of `A` here? Why not just `Kleisli[F, (Cake, LightweightFileSystem), B]`?

mimir/src/main/scala/quasar/mimir/ShiftedReadPlanner.scala
```diff
@@ -18,44 +18,93 @@ package quasar.mimir
...
+                      Task.now(right: FileSystemError \/ P.Table)
+
+                    // read from the lwc
+                    case -\/(_) =>
+                      lwfs.read(path) flatMap {
+                        case Some(stream) => for {
+                          // FIXME this pages the entire fs2.Stream into memory
+                          values <- stream
+                            .map(data => RValue.fromJValueRaw(JValue.fromData(data)))
+                            .runLog
```
djspiewak: Uh…

Don't we have a function around here somewhere which converts an `fs.Stream[F, A]` to a `StreamT[F, A]`?  If we don't, it's easy to write.  We really really really need to be using it, rather than paging things into memory.

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -59,13 +60,15 @@ import scala.concurrent.Future
 
 import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicLong}
 
-object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
+trait SlamDB extends BackendModule with Logging with DefaultAnalyzeModule {
```
djspiewak: Since we're renaming for marketing, shouldn't it be `SlamEngine`?

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -97,30 +101,59 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
...
+  // the data directory for mimir and the connection uri used to connect to the lightweight connector
+  final case class Config(dataDir: java.io.File, uri: ConnectionUri)
 
   def optimize[T[_[_]]: BirecursiveT: EqualT: ShowT]
       : QSM[T, T[QSM[T, ?]]] => QSM[T, T[QSM[T, ?]]] = {
     val O = new Optimize[T]
     O.optimize(reflNT[QSM[T, ?]])
   }
 
+  // { "dataDir": "/path/to/data/", "uri": "<uri_to_lwc>" }
```
djspiewak: Nice

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -133,7 +166,7 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
 
     def mapFuncPlanner[F[_]: Monad] = MapFuncPlanner[T, F, MapFunc[T, ?]]
 
-    def qScriptCorePlanner = new QScriptCorePlanner[T, Backend](
+    def qScriptCorePlanner = new QScriptCorePlanner[T, Backend, LightweightFileSystem](
```
djspiewak: This type seems to support the hypothesis that `CakeMT` doesn't need to carry the *specific* subtype of `LightweightFileSystem`.

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -245,73 +278,104 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
...
-        precog <- cake[Backend]
+        connectors <- cake[Backend, LightweightFileSystem]
+        (precog, lwfs) = connectors
 
-        exists <- precog.fs.exists(dir).liftM[MT].liftB
+        existsPrecog <- precog.fs.exists(dir).liftM[MT].liftB
+        childrenLwfs <- lwfs.children(dir).liftM[MT].liftB
 
-        _ <- if (exists)
+        _ <- if (existsPrecog || !childrenLwfs.isEmpty)
```
djspiewak: Can we restructure this so that the `childrenLwfs` effect is only run if `existsPrecog` is `false`?

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -245,73 +278,104 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
...
+        backLwfs <- childrenLwfs.getOrElse(Set[PathSegment]()).point[Backend]
+
+      } yield backPrecog ++ backLwfs
     }
 
     def fileExists(file: AFile): Configured[Boolean] =
-      cake[M].flatMap(_.fs.exists(file).liftM[MT]).liftM[ConfiguredT]
+      cake[M, LightweightFileSystem].flatMap {
+        case (precog, lwfs) =>
+          ((precog.fs.exists(file) |@| lwfs.exists(file))(_ || _)).liftM[MT]
```
djspiewak: Similar to the `childrenLwfs` above. I would structure this monadically rather than with the applicative so that the effects short-circuit and we don't incur the added overhead of the LWC read when the local filesystem is overriding.

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -245,73 +278,104 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
...
+      }.liftM[ConfiguredT]
   }
 
   object ReadFileModule extends ReadFileModule {
     import ReadFile._
 
-    private val map = new ConcurrentHashMap[ReadHandle, Precog#TablePager]
+    private val readMap =
+      new ConcurrentHashMap[ReadHandle, Precog#TablePager \/ (AtomicBoolean, Option[Stream[Task, Data]])]
```
djspiewak: Random thought… we should consider a type alias here to improve clarity.

mimir/src/main/scala/quasar/mimir/SlamDB.scala
```diff
@@ -245,73 +278,104 @@ object Mimir extends BackendModule with Logging with DefaultAnalyzeModule {
...
           case None =>
             MonadError_[Backend, FileSystemError].raiseError(unknownReadHandle(h))
         }
 
-        chunk <- pager.more.liftM[MT].liftB
+        chunk <- reader.fold(
+          _.more,
+          {
+            case (bool, data) if !bool.getAndSet(true) =>
+              data.fold(Vector[Data]().point[Task])(_.runLog)
```
djspiewak: This is really scary. What's happening here is you're re-chunking the stream which comes from the LWC read, paging it all into memory, and returning it. So it has the same limitation as the `ShiftedRead` implementation, but without the TODO.

It might almost be easier just to restructure things so that `ReadFileModule#read` delegates to a function which reads a path into a `Table`, then uses `TablePager` on it, and that same function can be the implementation for `ShiftedRead`. I say this because the only way to take an `fs.Stream` and *unsafely* page it (which is required by the `ReadFileModule` API) is to drain it into a `Queue` and then take from the queue one at a time, which is exactly what `TablePager` does.

This implementation would also considerably simplify the type signature of `readMap`, while also removing the need to have a separately-managed `AtomicBoolean` state machine.

mimir/src/main/scala/quasar/mimir/TablePager.scala
```diff
@@ -62,6 +62,7 @@ trait TablePagerModule extends ColumnarTableModule[Future] {
         } yield ()
       }
 
+      // enqueue the empty vector so ReadFile.scan knows when to stop scanning
```
djspiewak: 👍 for this documentation

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -628,8 +628,7 @@ trait ColumnarTableModule[M[+ _]]
             }
           }
         },
-        ExactSize(values.length)
-      )
+        ExactSize(values.length))
```
djspiewak: :+1:

yggdrasil/src/main/scala/quasar/yggdrasil/util/IdSourceScannerModule.scala
```diff
@@ -34,17 +34,12 @@ trait IdSourceScannerModule {
...
-        def apply(row: Int)       = id
-      }
-
       val seqCol = new LongColumn {
         def isDefinedAt(row: Int) = defined(row)
         def apply(row: Int): Long = pos + row
       }
 
-      (pos + range.end, Map(ColumnRef(CPath(CPathIndex(0)), CLong) -> seqCol, ColumnRef(CPath(CPathIndex(1)), CLong) -> idCol))
+      (pos + range.end, Map(ColumnRef(CPath(CPathIndex(0)), CLong) -> seqCol))
```
djspiewak: Interesting…

mimir/src/main/scala/quasar/mimir/MimirCake.scala
alissapajer: This might be left over from a more complicated time. I'll see what I can do.

yggdrasil/src/main/scala/quasar/yggdrasil/util/IdSourceScannerModule.scala
alissapajer: This `Scanner` was not used, so I just repurposed for my this use case.

alissapajer:
Interesting error in mimir integration test, I'll investigate.

mimir/src/main/scala/quasar/mimir/SlamDB.scala
alissapajer: Good idea.

alissapajer:
Also, the appveyor build is failing; I'll fix that.

alissapajer:
@wemrysi the connection uri for `mimir` will not be changing. Thoughts on if this should be `version: feature` or `version: breaking`? It doesn't technically break anything anymore, but it is a landmark feature, which we could landmark with a breaking version.

wemrysi:
@alissapajer I think a `version: feature` is fine.

alissapajer:
@edmundnoble any further review that you'd like to add before merge?

foundation/src/main/scala/quasar/fp/TaskRef.scala
```diff
@@ -68,6 +75,21 @@ object TaskRef {
...
           back <- if (bool) Task.now(b) else modifyT[B](f)
         } yield back
+
+      /* The effect on the left may be run multiple times, in the
+       * case that the `compareAndSet` fails to set. The effect on
+       * the right will only be run when `compareAndSet` successfully
+       * sets.
+       */
+      @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+      def twoPhaseModifyT[B](modify: A => (Task[A], Task[B])): Task[B] =
```
edmundnoble: Very interesting.

fs/src/main/scala/quasar/fs/Local.scala
```diff
@@ -42,373 +43,452 @@ import java.util.stream.Collectors
...
 
-    case QueryFile.More(h) => delayRight {
+    case QueryFile.More(h) => EitherT.rightT(Task.delay {
       FileSystemError.unsupportedOperation(
         "QueryFile.More not supported by the local file system.").left
-    }
+    })
 
-    case QueryFile.Close(_) => delayRight(())
+    case QueryFile.Close(_) => EitherT.rightT(Task.point(()))
```
edmundnoble: :+1:

alissapajer:
@wemrysi this should be ready to go!

PR #3583, opened 2018-05-10
Update SBT to 1.1.5

edmundnoble:
There's apparently big improvements in file watcher latency and even build times for 1.1.5, so I thought it might be worth upgrading.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.1.0
+sbt.version=1.1.5


alissapajer:
Wonder if it will fix the file handle resource leak I hit sometimes.

edmundnoble:
@alissapajer Worth a try! I don't see anything *directly* alluding to leaks or handles in https://developer.lightbend.com/blog/2018-05-09-sbt-1-1-5-patch-release/, but who knows.

PR #3584, opened 2018-05-15
Delete a bunch of stuff from the old precog

alissapajer:
The only reason everything in `quasar.blueeyes.json.serialization` is still around is because it's used in `ProjectionState`. I'd much like to get rid of it.


djspiewak:
Technically this is breaking. Do we care about versioning w.r.t. the `blueeyes` project?

djspiewak:
Oh also, I'll take this since @edmundnoble is going to be on an airplane for the next week. 😜

PR #3585, opened 2018-05-15
Refactor intra-project dependencies

wemrysi:
A first pass at attempting to make intra-project dependencies a bit more manageable. The crux is moving more of the combining-of-parts into core, namely the compiler. This allows types, LP, Sql, SSTs, etc to be independent and all the knowledge that ties them together goes into `core` (or things that depend on core).


wemrysi:
There is more work that can be done along this line, but I decided to stop here since there will be some package shuffling once quasar is running on top of `quasar.api._` and we get rid of the previous filesystem, http api, etc. That will give us another opportunity to revisit this organization after deleting some of that code.

djspiewak:
Swapped revision for breaking since we're versioning by the binary API.

build.sbt
```diff
@@ -238,24 +238,26 @@ lazy val root = project.in(file("."))
...
-        frontend,    precog,
-//    |/    /    \       |             |
-     fs, sql, datagen, blueeyes,
-//    |   |              |             |
-// ___|___|              |             |
-// |  |                  |             |
+//  /      |        \   /           \
+                  common,
+// /       |     /  |   \             \
+        frontend,  sql, precog,
```
alissapajer: :+1:

build.sbt
```diff
@@ -238,24 +238,26 @@ lazy val root = project.in(file("."))
...
-// |  |                  |             |
-     qsu,
-// |     \               |             |
+// |  |          ||      |_____________|__
+     qsu,      core,
+// \____\            \   |             | |
          connector,   yggdrasil,
-// |     /   |   \______|______________|_________
-//  \   /    |         /     \         \         \
-    core, skeleton, mimir, marklogic, mongodb, couchbase,
```
alissapajer: :+1:

PR #3586, opened 2018-05-15
Local LWC

alissapajer:
Includes the local file system LWC in the core distribution.

An example of adding a `local_file_system` mount follows. You can choose any directory with write access for the `connectionUri`; query-related data will only be written to this directory if a table is prepared (though nihdb will write some version and metadata files upon initialization).
```
curl -v -X PUT http://localhost:8090/mount/fs/localfs/ -d '{ "local_file_system": { "connectionUri":"/path/to/slamengine/storage/" } }'
```
 You automatically have access to query your entire local file system via the LWC (modulo actual file permissions). To query a local path, omit the leading `/`. This isn't particularly intuitive; if this were to be user-facing we'd want to change this so it behaves like the terminal.
```
(v41.5.2-715e2f8) 💪$ select * from `complete/local/path/to/foobar.txt`
```


wemrysi:
Travis build timed out on mongo tests and appveyor had a spurious failure during service tests, restarted both.

alissapajer:
@wemrysi green!

PR #3588, opened 2018-05-16
[WIP] Implement top-down rewriteM

jsantos17:
This is my implementation of top-down rewriteM. It's very similar to bottom up rewriteM and reuses the  vertices collapse implementation from it.

I'm not 100% sure I understand how the vertices collapse works, so I would like someone else to double check it makes sense.  This still needs tests, hence the `[WIP]`; but comments on the implementation of `rewriteDownM` are more than welcome. cc @wemrysi, @alissapajer, @djspiewak, @edmundnoble

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/QSUGraph.scala
+++ b/qsu/src/main/scala/quasar/qsu/QSUGraph.scala
@@ -156,6 +156,50 @@ final case class QSUGraph[T[_[_]]](
   def overwriteAtRoot(qsu: QScriptUniform[T, Symbol]): QSUGraph[T] =
     QSUGraph(root, vertices.updated(root, qsu))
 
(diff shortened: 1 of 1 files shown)


djspiewak:
Ok I see why `rewriteDown` isn't working in its current form. Also I would prefer `corewriteM` as a name just to avoid directional confusion. Anyway…

The issue is that the pre/post-image inference was sort of copied from the bottom-up version, meaning that you're completely missing the computation of collapse on overlapping subgraphs affected by the rewrite *itself*.  In fact, I'm pretty sure the pre/post-image computation in its current form is not going to do anything useful. In the bottom-up algorithm, all of the subgraph fiddling can be identified after the recursion because it stems from prior steps. In the top-down algorithm, subgraph fiddling is exclusively performed by the rewrite itself, while the recursive steps may be simply trusted (I think).

jsantos17:
Closing this for now.

PR #3591, opened 2018-05-17
Add QData and implement a QData[Data]

alissapajer:


djspiewak:
Are we putting eprec into the quasar project?

qdata/src/main/scala/quasar/qdata/QType.scala
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qdata
+
+sealed trait QType
+
+final case object QLong extends QType
```
wemrysi: Mind moving these to the `QType` companion?

wemrysi:
@djspiewak @alissapajer and I thought `QData` is rather tightly coupled to Quasar, so it'd live in this repo. I'm not opposed to it in its own repo, however, if we think it has utility elsewhere. We'd need to extract the temporal types we've introduced as well.

djspiewak:
@wemrysi Isn't `QData` the typeclass which allows the eprec codecs to be written? I feel like the architecture has probably evolved since I last saw it, so take my thoughts with a grain of salt. :-) I just don't see a way to have `QData` in quasar without *also* having eprec itself in quasar.

wemrysi:
@djspiewak It is, I just missed the distinction you were making between `QData` (formerly known as `EfficientPrecise`) and "eprec". Yes, `QData` in quasar implies eprec in quasar.

PR #3592, opened 2018-05-18
Destroy the future

edmundnoble:
#qz-3535 done

A little bigger than I thought, git diff lied to me.

Small changes:
a) Remove variance from all `M[+_]`.
b) Display stack traces for uncaught exceptions. This was necessary for debugging, and will prove extremely useful in general.
c) Introduce shim instances of fs2 and cats-effect type classes.
d) Remove unnecessary shadowing imports of `EitherT.eitherTMonad`.

Big changes:
a) Introduce dependencies on fs2, shims, and cats-effect to many modules.
b) Replace almost all occurrences of `Task` in mimir and yggdrasil with `IO`. A notable boundary between `IO` and `Task` is the `Precog` `BackendModule`.
c) Replace all occurrences of `Future` in mimir and yggdrasil with `IO`.

Remaining issues:
a) All queries involving equijoins return no results. This may point to a problem with `cogroup` and execution order.
b) Use of the laziness in `M.point(x)` was ubiquitous, though it isn't really lazy because `Future` eagerly executes. I tried replacing it with `M.bind(M.point(()))(_ => M.point(x)` but that's a crap solution; at least `Sync` is needed, especially for JDBM3 interaction, but at that point the abstraction is totally useless and we should just use `IO`. We should do that anyway, because it helps to unravel the cake.
c) Unused imports.
d) Upgrading fs2 and http4s.

Edit: See my comment later on about further changes to the PR.


rintcius:
I am concerned about the `cats-effect` and `shims` dependencies being added. I guess we share a dislike about being stuck on old versions of `scalaz`, `http4s` etc which makes it more difficult to get rid of `Future` and other legacy classes, but I __really__ don't like us entering a mixed `scalaz` and `cats` situation. 
This confuses the architecture (eg a gliding scale of how much `cats` is allowed and where) & before we know it someone either unknowingly uses a shim on a performance critical path or is forced to use it.

Can you explain what the exact reasons are for adding this & what alternatives have been considered?  I hope we don't really need it, but if we do, I think it's important that someone adds a guideline to e.g. the wiki about when we can/should/can't/shouldn't use it.

djspiewak:
Let the future die. Kill it, if you have to.

mimir/src/test/scala/quasar/mimir/MathLibSpecs.scala
```diff
@@ -1,2052 +0,0 @@
...
- * limitations under the License.
- */
-
-package quasar.mimir
-
-import quasar.yggdrasil._
-import quasar.precog.common._
-import scalaz._
-
-trait MathLibSpecs[M[+_]] extends EvaluatorSpecification[M]
```
alissapajer: Why remove these tests? These are the only tests we have for the mimir libraries.

djspiewak:
@rintcius Sorry for the delay on responding. You raise some important points, so here's some specific responses…

> but I really don't like us entering a mixed scalaz and cats situation.

This is unavoidable because we use Scala. The ecosystem has bifurcated, with some frameworks standardizing on cats, and others on scalaz. Any realistic Scala endware project that isn't either a) avoiding FP, or b) severely ideologically biased is going to have both cats and scalaz on the classpath, and this is going to be true for the next several years (if not closer to a decade). We have to just get used to this.

This is precisely why shims exists.

> This confuses the architecture (eg a gliding scale of how much cats is allowed and where) […]

Cats is allowed. Simplifying this down: use the right tool for the job. Period. Now, we're not going to go through and search-replace scalaz with cats in quasar; we just don't have the time to do something like that, and there's no really compelling reason to do so. Scalaz 7 and cats are quite close together in most ways, and the ways in which they aren't (e.g. `Eval` and cats-mtl) we're already working around in some way (i.e. we don't really need `Trampoline` or `Need` since we use `Free` and `Task` so much, and we already have things like `MonadState_`). But just because our primary FP framework is and will remain scalaz 7 doesn't mean that we can't also use cats where it makes sense.

For example, http4s is the best web framework for Scala, full stop. So we're using http4s. Http4s depends on cats. fs2 is a very similar situation. Thus, cats ends up on our classpath, and we end up interacting with its types in a few places. This is just fine, and shims should make the interaction points quite painless (if it doesn't, definitely file a bug).

*By convention*, I would say that we should continue to use the `scalaz` variants of typeclasses as much as we possibly can. In other words, we shouldn't be defining methods in quasar which have something like `F[_]: cats.Monad`; use `scalaz.Monad` instead and lean on shims if you're interacting with something that expects a cats type. However, in the cases where the cats variant is essential or more convenient, feel free to use it. For example, you should use `cats.effect.Effect` if you need to, since scalaz has absolutely no peer to that typeclass. You should also use `cats.effect.LiftIO` instead of `scalaz.LiftIO` if relevant, since scalaz's `LiftIO` is kind of worthless, while the cats-effect version is well integrated into the ecosystem. And obviously, if you have a `cats.effect.Effect` in scope, you should not *also* bring in a `scalaz.Monad`, since `Effect` will materialize `Monad` via shims. You get the idea.

There are also a few other *extremely* niche situations where breaking convention and using cats types makes sense. For example, `cats.data.StateT` is stack-safe (at the cost of performance), while `scalaz.StateT` is not. So if you start running into stack issues while using `StateT`, feel free to use the cats version instead. But… uh, document it in some comments. And you probably won't be in that scenario.

Long-term, we'll have to make some decisions about frameworks, since scalaz 8 is a major departure and we're going to get left behind. We've got at least a year or two before we need to worry about that though.

> […] & before we know it someone either unknowingly uses a shim on a performance critical path or is forced to use it.

Shims is extremely lightweight by design. It's basically just a bunch of delegates, and not even done in a particularly fancy or polymorphic fashion. The JVM should basically just see through all of them (though they won't be completely inlined due to existing polymorphism on the call sites). It really shouldn't add any measurable overhead above and beyond what is *already* present at any call-site which is using typeclass based abstraction.

This actually gets at a more critical point: if you're in a performance-critical path, you shouldn't be using things like `Monad` or similar *in the first place.* Forget the overhead of shims: if you're in a critical path, just the overhead of scalaz/cats is *already* unacceptably high.

So basically, I wouldn't worry about shims, really. If you do find it causes some problems, it is *definitely* a bug and please file an issue on the project itself. It really should be, for all practical intents and purposes, a nearly-zero-cost and invisible compatibility layer.

> Can you explain what the exact reasons are for adding this & what alternatives have been considered?

`cats.effect.IO` is simply the best IO type for purposes, and so it definitely falls under the rubric of "use the best tool for the job".  The only other real alternative is Monix `Task`, but cats-effect is already going to be there because we're using fs2, and we don't need the other features that Monix's `Task` brings. `scalaz.concurrent.Task` was a good tool at the time it was introduced, but it's woefully obsolete at this point.  Prefer `IO` over `Task` for all new code (i.e. anything that isn't *already* assuming `Task`). We will eventually replace `Task` altogether.

Does that mostly answer your questions? This is a really important topic of conversation, so I'm glad you raised these points.

djspiewak:
@edmundnoble Regarding the `M.point` usage…  I think the correct thing to do is assume that *all* uses of `point` in the mimir code are tainted. Instead of replacing them with `pure`, you should change the constraint on `M[_]` from `scalaz.Monad` to `cats.effect.Async`. Then, search-replace `M.point` with `M.delay`. Obviously, this is *also* wrong, so we're going to need to go back and find the non-side-effecting constructions and replace them with `pure`, but at least it will be safe.

rintcius:
Thanks for the details @djspiewak, much appreciated! That answers most of the questions indeed.
I guess in my mind shims were taking more performance overhead than they do in reality (would still be nice if we can see this confirmed in a benchmark... one day). The mental overhead of mixing cats/scalaz is still there, but yeah I do appreciate that shims keeps us out of a deadlock situation here.

Am I right that we won't upgrade http4s (with the switch from scalaz to cats) until we'll make those long-term decisions, i.e. is that upgrade part of the "decisions about frameworks"?

djspiewak:
> I guess in my mind shims were taking more performance overhead than they do in reality (would still be nice if we can see this confirmed in a benchmark... one day).

I would love to see that as well. PRs accepted! 😉

> Am I right that we won't upgrade http4s (with the switch from scalaz to cats) until we'll make those long-term decisions, i.e. is that upgrade going to be part of the "decisions about frameworks"?

Actually we're upgrading http4s already on the new filesystem branch. The longer term decisions about frameworks is really referring to when scalaz 8 comes out and we have to decide if we're going to bite the bullet and switch to scalaz 8 (which is a large jump) or bite the bullet and switch to cats.

mimir/src/test/scala/quasar/mimir/MathLibSpecs.scala
edmundnoble: Yeah. This PR requires what I think is a giant patch to the evaluator otherwise. I think we should incorporate these tests into StdlibSpec, or throw them away. I can do that in this PR.

edmundnoble:
For the record, though runtime performance may be an issue, I've found practically that introducing shims really causes issues with compile time. That's probably more worth benchmarking.

edmundnoble:
When I introduced shims to slamdata-benchmarks and mimir, incremental compiles went from a couple seconds up to around eight seconds; not a controlled environment by any means, public benchmarks would probably help eliminate them.

rintcius:
compile time is the run time for developers ;)

djspiewak:
> public benchmarks would probably help eliminate this as a concern.

PRs accepted! ;-)

mimir/src/test/scala/quasar/mimir/MathLibSpecs.scala
alissapajer: `StdLibSpec` is testing the expected implementations of the `MapFunc` in quasar, whereas these are unit tests for mimir's standard library functions. In many cases there is a 1-1 mapping from mimir function to `MapFunc`, but not in all cases. Furthermore, from quasar's point of view, the mimir standard library is an implementation detail. mimir is a standalone evaluator that quasar just happens to use; we could interpret from something else entirely to `Table` and that would still be valid. And so I don't want mimir's unit tests tied to `MapFunc` in quasar.

The entire evaluator is an artifact from the old precog and only exists in so that we can test these functions in mimir. Ideally we could rip out most of that code and rewrite it to just contain the infrastructure needed to test the mimir standard library. If we decide to go that route, I'd prefer to do that in a separate PR.

rintcius:
> Actually we're upgrading http4s already on the new filesystem branch

Where can I find that branch? Or is it not pushed yet?

mimir/src/test/scala/quasar/mimir/MathLibSpecs.scala
edmundnoble: @alissapajer It actually turns out that removing the polymorphism in `M[_]` was all I needed to do. So the tests are back!

edmundnoble:
I've removed the polymorphism in `M[_]` from mimir and yggdrasil entirely. Everything is in IO. I also fixed `cogroup`, which was actually just broken because I broke it temporarily and forgot to reinstate the real version :blush:

An update on the PR description:
a) All queries involving equijoins and cogroup in general should now work, pending CI.
b) Use of the laziness in `M.point(x)` was carefully considered at every point; however, *no tests failed* when switching. This is because the tests *already used `Need` instead of `Future`* so whatever bugs in mimir that were caused by the incongruence were *never possible* to catch in tests anyway. /cc @djspiewak 
c) Unused imports are still a thing, which I will tackle with a follow-up PR, as well as
d) Upgrading fs2, because this PR is way too big now already.

alissapajer:
@edmundnoble failed to compile.

Can you get rid of the unused imports in this PR? Since they are an artifact of the work you did here, I think it makes sense to remove them now.

alissapajer:
:+1: to upgrading fs2 in another PR.

edmundnoble:
@alissapajer Fixed compilation, it was the extra (now redundant) shadowing imports on master.

The thing is I don't actually know which imports are unused because of my work and which are unused because there are just a lot of unused imports in mimir/yggdrasil. That's why I want to follow up; removing all of the unrelated unused imports seems out of scope and is probably *much* faster than going over imports with a fine toothed comb, but let me know.

foundation/src/main/scala/quasar/contrib/scalaz/concurrent/package.scala
```diff
@@ -16,10 +16,16 @@
 
 package quasar.contrib.scalaz
```
alissapajer: The additions in this file shouldn't be in the `contrib.scalaz` package because I don't think scalaz is going to accept PRs that depend on cats.

foundation/src/main/scala/quasar/contrib/scalaz/concurrent/package.scala
```diff
@@ -31,4 +37,51 @@ package object concurrent {
...
+  implicit class toOps[F[_], A](fa: F[A]) {
+    def to[G[_]](implicit F: Effect[F], G: Async[G], ec: scala.concurrent.ExecutionContext): G[A] =
+      Async[G].async { l =>
+        Effect[F].runAsync(fa)(c =>
+          IO(l(c))
+        ).unsafeRunSync
+      }
+  }
+
+  implicit def taskEffect: Effect[Task] = new Effect[Task] {
```
alissapajer: I'm surprised that an `Effect[Task]` doesn't exist in a library somewhere already.

foundation/src/main/scala/quasar/contrib/scalaz/fs2/package.scala
```diff
@@ -0,0 +1,79 @@
...
+import slamdata.Predef._
+
+import _root_.cats.effect.IO
+
+import _root_.fs2.util.{Async, Attempt}
+import Async.Ref
+import _root_.fs2.{Strategy, Task}
+
+import _root_.scalaz.{~>, Functor}
+import _root_.scalaz.syntax.functor._
```
alissapajer: Can you get rid of the `_root_` in the imports here?

foundation/src/main/scala/quasar/contrib/scalaz/fs2/package.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import shims._
+
+package object fs2 {
+  implicit val asyncCatsIO: Async[IO] =
```
alissapajer: We can get this here: https://github.com/functional-streams-for-scala/fs2-cats/blob/master/shared/src/main/scala/fs2/interop/cats/IOAsyncInstances.scala#L28

alissapajer:
@wemrysi @nicflores is there a "new filesystem branch" somewhere where we are using the latest version of `http4s`? @nicflores where have you been doing your api work? /cc @rintcius

foundation/src/main/scala/quasar/contrib/scalaz/concurrent/package.scala
djspiewak: Yoink! https://christopherdavenport.github.io/scalaz-task-effect/

foundation/src/main/scala/quasar/contrib/scalaz/fs2/package.scala
edmundnoble: Deleting the file, after you linked fs2-cats.

foundation/src/main/scala/quasar/contrib/scalaz/concurrent/package.scala
edmundnoble: Now in contrib.cats.effect.

foundation/src/main/scala/quasar/contrib/scalaz/concurrent/package.scala
edmundnoble: Added this dependency.

foundation/src/main/scala/quasar/contrib/scalaz/fs2/package.scala
edmundnoble: Done.

edmundnoble:
@alissapajer Turns out there were very few unused imports, so I just snipped them out.

wemrysi:
@alissapajer @nicflores @rintcius Yes, but it's over in @nicflores 's fork of slamdata-backend, nothing has been committed to upstream yet, afaik.

nicflores:
@alissapajer, as @wemrysi mentions, I have a local branch where we are using http4s version v0.18.9 within a new module of slamdata-backend. I've pushed some of the work to my fork here: https://github.com/nicflores/slamdata-backend/tree/wip-add-new-data-service
I still need to write tests for the `/datasource` and `/datasources` endpoints before I PR.

mimir/src/main/scala/quasar/mimir/ReductionLib.scala
```diff
@@ -51,7 +51,7 @@ class LongAdder {
   def total(): BigDecimal = ts.foldLeft(BigDecimal(0))(_ + _) + t
 }
 
-trait ReductionLibModule[M[+ _]] extends ColumnarTableLibModule[M] {
```
alissapajer: :+1: to removing `M` from all the modules; I don't think we were gaining any benefits from it.

alissapajer:
I added "#qz-3535 done" to the PR description.

mimir/src/main/scala/quasar/mimir/ReductionLib.scala
edmundnoble: That, and every abstraction I *did* find over `M[_]` just required `Monad`, which basically has to be an improper usage of `F.point`.

mimir/src/main/scala/quasar/mimir/TablePager.scala
```diff
@@ -78,11 +83,12 @@ trait TablePagerModule extends ColumnarTableModule[Future] {
...
 
   object TablePager {
-    def apply(table: Table, lookahead: Int = 1): Task[TablePager] = {
+    def apply(table: Table, lookahead: Int = 1): IO[TablePager] = {
       for {
-        q <- async.boundedQueue[Task, Throwable \/ Vector[Data]](lookahead)
-        slices = table.slices.trans(λ[Future ~> Task](_.toTask))
-        back <- Task.delay(new TablePager(slices, q))
+        q <- async.boundedQueue[Task, Throwable \/ Vector[Data]](lookahead).to[IO]
+        // ambiguity between M and effect-derived monad
```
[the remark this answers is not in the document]
edmundnoble: This is actually wrong now; I had an implicit ambiguity error in the `trans` call. I'll PR a fix tomorrow.

PR #3593, opened 2018-05-21
Remove support for static datasources from the api.

wemrysi:
Turns out uploading data directly into Quasar is no longer a feature, thus we no longer need static datasources.


wemrysi:
While this does modify the Quasar API, it is not a portion that has yet been released, so I figured a `version: revision` would be appropriate. If other would rather it be `breaking`, that's fine too.

nicflores:
Looks good @wemrysi, I like the simplification.

PR #3596, opened 2018-05-23
Remove dependency on scalaz-specs2 in favor of built-in support.

wemrysi:
`specs2` now includes support for assertions about `scalaz.\/` and `scalaz.Validation` that is more flexible and conducive to more correct tests than those included in `scalaz-specs2`. This removes the dependency on the latter, preserving the `scalaz.Equal`-based matcher and the utility functions for checking `scalaz` typeclass laws.

### DIFF
--- a/sql/src/test/scala/quasar/sql/SqlParserSpec.scala
+++ b/sql/src/test/scala/quasar/sql/SqlParserSpec.scala
@@ -146,7 +146,7 @@ class SQLParserSpec extends quasar.Qspec {
 
     "parse basic select" in {
       parse("select foo from bar") must
-        beRightDisjunction(
+        be_\/-(
           SelectR(
             SelectAll,
             List(Proj(IdentR("foo"), None)),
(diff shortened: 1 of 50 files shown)


sql/src/test/scala/quasar/sql/SqlParserSpec.scala
```diff
@@ -626,24 +626,24 @@ class SQLParserSpec extends quasar.Qspec {
...
-      parse(q) must beLeftDisjunction
+      parse(q) must be_-\/
     }
 
     "parse function declaration" in {
       val funcDeclString = "CREATE FUNCTION ARRAY_LENGTH(:foo) BEGIN COUNT(:foo[_]) END"
-      fixParser.parseWithParser(funcDeclString, fixParser.func_def) must beRightDisjunction(
-        FunctionDecl(CIName("ARRAY_LENGTH"),List(CIName("foo")),Fix(invokeFunction(CIName("count"),List(Fix(Unop(Fix(vari[Fix[Sql]]("foo")),ShiftArrayValues)))))))
+      fixParser.parseWithParser(funcDeclString, fixParser.func_def) must be_\/-(
+        equalTo(FunctionDecl(CIName("ARRAY_LENGTH"),List(CIName("foo")),Fix(invokeFunction(CIName("count"),List(Fix(Unop(Fix(vari[Fix[Sql]]("foo")),ShiftArrayValues))))))))
```
rintcius: just curious, why is this `equalTo` sometimes needed?

PR #3597, opened 2018-05-24
Handle Leftshift OnUndefined.Emit in SlamEngine

rintcius:
#qz-3763 Done


it/src/main/resources/tests/flattenOrElseField.test
```diff
@@ -2,10 +2,8 @@
     "name": "flatten or else field",
     "backends": {
         "couchbase":         "pending",
-        "lwc_local":         "pending",
         "marklogic_json":    "pending",
-        "marklogic_xml":     "pending",
-        "mimir":             "pending"
```
alissapajer: :tada:

it/src/main/resources/tests/flattenValueAndIndexWithField.test
```diff
@@ -0,0 +1,21 @@
...
+    "predicate": "exactly",
+    "ignoreResultOrder": true,
+    "expected": [
+      {"a": 1, "b": true, "i": 0},
+      {"a": 1, "b": true, "i": 1},
+      {"a": 1, "b": true, "i": 2},
+      {"a": 2, "b": false, "i": 0},
+      {"a": 2, "b": false, "i": 1},
+      {"a": 3, "i": 0},
+      {"a": 4, "i": 0}
```
alissapajer: Are these last two lines correct? Shouldn't they instead be 
```json
{"a": 3 },
{"a": 4 }
```
When we call `b[*:]` where the value of `b` at a given row isn't an array, shouldn't the array index be undefined, not `0`?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1481,27 +1484,46 @@ trait ColumnarTableModule[M[+ _]]
...
 
         val innerHeads = {
           val unsorted = focused.keys.toVector flatMap {
             case ColumnRef(path, _) => path.head.toVector
           }
 
+          // make sure there's at least 1 element when emitOnUndef
+          val unsorted0 =
+            if (unsorted.isEmpty && emitOnUndef)
+              Vector(CPathIndex(0))
```
alissapajer: This is where the `0`s in the integration test are coming from. This seems confusing (and incorrect) because now if we join on array index suddenly these rows that acquired this `0` will join with other rows that originated from an actual array.

alissapajer: Can we emit a value so that the semantics of the shift are correct, but then not actually return that value?

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
```diff
@@ -214,22 +214,23 @@ final class QScriptCorePlanner[T[_[_]]: BirecursiveT: EqualT: ShowT, F[_]: Monad
...
-    case qscript.LeftShift(src, struct, idStatus, _, onUndef, repair) =>
+    case qscript.LeftShift(src, struct, idStatus, shiftType, onUndef, repair) =>
       import src.P.trans._
 
       for {
+
         structTrans <- interpretMapFunc[T, F](src.P, mapFuncPlanner[F])(struct.linearize)
-        wrappedStructTrans = InnerArrayConcat(WrapArray(TransSpec1.Id), WrapArray(structTrans))
+        wrappedStructTrans =
+          OuterObjectConcat(WrapObject(TransSpec1.Id, "src"), WrapObject(structTrans, "f"))
```
alissapajer: Can we make these strings `"src"` and `"f"` constant `val`s in scope instead of referencing the same string in multiple places?

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
```diff
@@ -241,7 +242,8 @@ final class QScriptCorePlanner[T[_[_]]: BirecursiveT: EqualT: ShowT, F[_]: Monad
             },
             mapFuncPlanner[F].plan(src.P)[Source1](TransSpec1.Id)))
 
-        shifted = src.table.transform(wrappedStructTrans).leftShift(CPath.Identity \ 1)
+        emit = (shiftType === ShiftType.Array) && (onUndef === OnUndefined.Emit)
```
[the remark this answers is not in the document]
alissapajer: We need to add analogous integration tests (and unit tests) for shifting objects. @rintcius can we do that in this PR?

it/src/main/resources/tests/flattenValueAndIndexWithField.test
rintcius: @alissapajer Hmm, my interpretation of "emit" is that it's emitting an identity. I also used mongo as reference. I tried it with other queries and with this interpretation, the mimir implementation is consistent with mongo. I'll add ITs for this, so that we can first agree what these tests should return.

If it must not emit an identity then that will complicate things...
I initially went down this path but failed to implement it. One idea I didn't fully try was to introduce some fake `undefined` value, to make it distinguishable from a real `undefined`.

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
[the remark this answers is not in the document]
rintcius: @djspiewak Only arrays because that's in line with mongo. Also I thought it didn't make sense to take a default identity here (see my previous comment for context).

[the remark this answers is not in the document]
rintcius: @alissapajer I'll add ITs. With respect to unit tests, I think they are already there? Just some expectations need to be changed, most importantly what needs to be returned in case of `undefined`, eg here: https://github.com/quasar-analytics/quasar/pull/3597/files/cef9563f1c193a350b39fc715dc66a442ee05093#diff-a2173b7af9f5af76ab9831ab0de27c6eR58

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1481,27 +1484,46 @@ trait ColumnarTableModule[M[+ _]]
...
+            else
+              unsorted
+
           // sort the index lexicographically except in the case of indices
-          unsorted sortWith {
+          // also: indices come first
```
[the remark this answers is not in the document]
rintcius: This is a leftover of an earlier attempt. I don't think it's strictly needed indeed. I was just liking this a bit more, so I left it, but ok to change it back if you prefer.

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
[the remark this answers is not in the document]
alissapajer: I think mongo may be incorrect wrt object shifting, but I haven't looked at any examples to be sure. We should concretize this with examples to make sure we're on the same page. For example, if we have
```sql
select a{*}, b from foo
```
then we want returned all rows with a value at `b`, even if the corresponding value at `a` wasn't an object.

@rintcius what's an example where mongo handles object shifting differently from array shifting?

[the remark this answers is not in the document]
rintcius: @alissapajer I pushed a new commit here https://github.com/quasar-analytics/quasar/pull/3597/commits/f8e779b7597be6b0e23ae57e91f3d0dfd0d6e82d - these tests pass on both mimir and mongo 3.6

it/src/main/resources/tests/heterogenousFlattenValueAndIndexWithField.test
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "field and flattened field with value and index on heterogenous",
```
alissapajer: "field and flattened array..."?

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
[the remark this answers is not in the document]
alissapajer: Ooo those are awesome, thanks! I think they should behave identically, and that mongo is buggy. I don't see anything in MRA that treats arrays and objects differently in this situation. @wemrysi @djspiewak ?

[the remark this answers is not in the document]
wemrysi: @alissapajer I'd agree that shifting a value that isn't of the correct type being shifted should behave the same regardless of type :+1:

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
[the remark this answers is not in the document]
djspiewak: It's fine either way.

it/src/main/resources/tests/flattenValueAndIndexWithField.test
djspiewak: @alissapajer is correct: those two rows should be:

```json
{"a": 3 }
{"a": 4 }
```

Basically, you should think of shifting (with or without identities) as being an operation like any other which returns `undefined` when it is invalid. Thus, the `RightSide` of the `struct` in `LeftShift` ends up being `Undefined()`, which when fed through `MapConcat(LeftSide, _)` yields just `LeftSide`.

it/src/main/resources/tests/heterogenousFlattenValueAndIndexWithField.test
rintcius: I used without "array" because that naming standard is used in other ITs, i.e. if it is a map flattening then map is in the name. If it doesn't mention map or array then array is implied. But I agree that's it's better to be explicit. I'll add it here & in other spots where I find the same omission.

it/src/main/resources/tests/flattenValueAndIndexWithField.test
rintcius: ok, thanks for spotting the issue @alissapajer ! I'm happy that I added this test

rintcius:
@alissapajer @djspiewak not sure if it will actually be needed, but would it matter if I have to change the impl of `leftShift` such that it also takes the ShiftType into account? Eg then this unit test will have different results for `Array` vs `Map`  https://github.com/quasar-analytics/quasar/pull/3597/files#diff-a2173b7af9f5af76ab9831ab0de27c6eR108

djspiewak:
@rintcius I'm ok taking `ShiftType` into account on `leftShift`, but I don't see why it would be necessary. For one thing, you actually know what the types of things are by looking at the slice schema.

rintcius:
Looking at the unit tests and the basic ITs (`select a, b[*]`, `select a, b[*:]`, `select a, b{*}` and `select a, b{*:}`), the implementation of `ColumnarTableModuel.leftShift` looks correct now AFAICS (turned out to be easier than I imagined, indeed without having to take `ShiftType` into account).

The more complex ITs that I have added now return less rows than I previously thought were correct. I can reason my way backward to explain these results but I'm not sure it's valid. There may also something be wrong in the plan, for example.

See these tests (they are all green on mimir)
https://github.com/quasar-analytics/quasar/pull/3597/files#diff-e0c61996a6e4b5659f89996d2cf1c6f0
https://github.com/quasar-analytics/quasar/pull/3597/files#diff-f1990961d709764ed83550677762282e
https://github.com/quasar-analytics/quasar/pull/3597/files#diff-786634c34f15a209d9650503ea2634b0
https://github.com/quasar-analytics/quasar/pull/3597/files#diff-0c963d481490feaa8e6f645a88b6e1dd

it/src/main/resources/tests/heterogeneousFlattenArrayValueAndIndexWithField.test
```diff
@@ -0,0 +1,31 @@
...
+    "expected": [
+      { "a": 2, "b": "u", "i": 0 },
+      { "a": 2, "b": "v", "i": 1 },
+      { "a": 6, "b": "w", "i": 0 },
+      { "a": 10, "b": [ 4, 5, 6 ], "i": 0 },
+      { "a": 10, "b": { "m": 1 }, "i": 1 },
+      { "a": 10, "b": null, "i": 2 },
+      { "a": 10, "b": "foo", "i": 3 },
+      { "a": 10, "b": {  }, "i": 4 },
+      { "a": 10, "b": [], "i": 5 }
```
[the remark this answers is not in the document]
alissapajer: I haven't looked at the plans, but it's definitely possible that the plan is incorrect. We haven't tested theses sorts of queries thoroughly before, so something could be wrong in one of the compiler phases.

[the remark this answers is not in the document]
rintcius: Ok, I created qz-3833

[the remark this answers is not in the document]
rintcius: BTW there's still `select a, b[*], b{*}` of which I'm unsure whether the results are correct (for now included in the same ticket). 
The question is whether `{ "a": 3 }` and `{ "a": 4 }` need to be included or not given these rows:

```
{ "a": 3, "b": {} }
{ "a": 4, "b": [] }
```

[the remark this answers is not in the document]
alissapajer: Ah, sorry I missed that one! Given those rows as input, `{ "a": 3 }` and `{ "a": 4 }` should be included in the result set.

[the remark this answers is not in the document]
alissapajer: For both those input rows, the values of `b[*]` and `b{*}` both go to undefined, which is why only the value for `a` is returned.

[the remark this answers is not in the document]
rintcius: It's different though, in that the complete rows that are missing are different between array and map flattening. The `3` row is missing in 1 case and the `4` row in the other. I was wondering how that combines in the `select`.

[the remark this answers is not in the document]
alissapajer: `select a, b{*}, b[*]` is really shorthand for `select {"a": a, "b": b{*}, "b": b[*]}`. We create the result rows from a `select` by listing out all possible values for each of `a`, `b{*}`, and `b[*]` and crossing them (with "outer" semantics). In the case of the `3` row, we have a single value for `a`, and no values for `b{*}` and `b[*]`. So we return `{ "a": 3 }`. The undefined we get by considering `b{*}` when the value of `"b"` is `{}` is no different from the undefined we get by considering `b[*]` when the value of `"b"` is an object (empty or not).

Technically, according to the MRA spec, we can only select a value that is _not_ undefined. Maybe that's a better way to think about it?

Were you thinking that we should not return the rows `{ "a": 3 }` and `{ "a": 4 }`? Or that we should return multiple copies of them?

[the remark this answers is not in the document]
alissapajer: Ok, what I said about the MRA spec was technically incorrect but the spirit of it was still correct. So a sql selection is actually a MRA mapping (not a MRA selection), and a mapping can produce undefined values. We then autojoin the mapped values with "outer" semantics.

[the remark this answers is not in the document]
rintcius: Thanks for the explanation @alissapajer , got it now!

> Were you thinking that we should not return the rows { "a": 3 } and { "a": 4 }? Or that we should return multiple copies of them?

I was not sure if the rows `{ "a": 3 }` and `{ "a": 4 }` needed to be included or not.

rintcius:
@alissapajer @djspiewak ready for review

it/src/main/resources/tests/heterogeneousFlattenArrayValueAndFlattenMapValueWithField.test
```diff
@@ -0,0 +1,44 @@
+{
+    "name": "field and flattened array value and flattened map value on heterogenous",
+    "backends": {
+        "couchbase":         "pending",
+        "lwc_local":         "pendingIgnoreFieldOrder",
+        "marklogic_json":    "pending",
+        "marklogic_xml":     "pending",
+        "mimir":             "pendingIgnoreFieldOrder",
```
alissapajer: Do you know why this isn't working for mimir?

it/src/main/resources/tests/heterogeneousFlattenArrayValueAndIndexWithField.test
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "field and flattened array value and index on heterogenous",
+    "backends": {
+        "couchbase":         "pending",
+        "lwc_local":         "pendingIgnoreFieldOrder",
+        "marklogic_json":    "pending",
+        "marklogic_xml":     "pending",
+        "mimir":             "pendingIgnoreFieldOrder",
```
alissapajer: Same for this one.

it/src/main/resources/tests/heterogeneousFlattenMapValueAndKeyWithField.test
```diff
@@ -0,0 +1,36 @@
+{
+    "name": "field and flattened map value and key on heterogeneous",
+    "backends": {
+        "couchbase":         "pending",
+        "lwc_local":         "pendingIgnoreFieldOrder",
+        "marklogic_json":    "pending",
+        "marklogic_xml":     "pending",
+        "mimir":             "pendingIgnoreFieldOrder",
```
alissapajer: And this one.

it/src/main/resources/tests/flattenArrayValueAndIndexWithField.test
```diff
@@ -0,0 +1,24 @@
...
+    "query": "select a, b[*], b[*:] as i from flattenable",
+    "predicate": "exactly",
+    "ignoreResultOrder": true,
+    "expected": [
+      {"a": 1, "b": true, "i": 0},
+      {"a": 1, "b": true, "i": 1},
+      {"a": 1, "b": true, "i": 2},
+      {"a": 2, "b": false, "i": 0},
+      {"a": 2, "b": false, "i": 1}
+    ]
```
alissapajer: These results are incorrect. The query should return two additional rows:
```
{"a": 3}
{"a": 4}
```

it/src/main/resources/tests/heterogeneousFlattenArrayValueAndFlattenMapValueWithField.test
rintcius: @alissapajer I don't know, but I have created qz-3833 for this one and the other 2 you commented on - I was thinking this should be prioritized separately from this PR

it/src/main/resources/tests/flattenArrayValueAndIndexWithField.test
rintcius: ah, yes - forgot to update this one, thanks!

it/src/main/resources/tests/heterogeneousFlattenArrayIndexWithField.test
```diff
@@ -0,0 +1,35 @@
...
+      { "a": 6, "i": 0 },
+      { "a": 7 },
+      { "a": 8 },
+      { "a": 9 },
+      { "a": 10, "i": 0 },
+      { "a": 10, "i": 1 },
+      { "a": 10, "i": 2 },
+      { "a": 10, "i": 3 },
+      { "a": 10, "i": 4 },
+      { "a": 10, "i": 5 }
```
alissapajer: These results are missing a row. They should additionally include the row `{"a": 4}`. Maybe this bug is related to why the other integration tests are failing.

alissapajer: @rintcius also, btw, big :+1: to the incrementing values in the `"a"` field in `heterogeneous.data`; that makes it a lot easier to verify the results!

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
```diff
@@ -213,22 +213,26 @@ final class QScriptCorePlanner[T[_[_]]: BirecursiveT: EqualT: ShowT, F[_]: Monad
         }
       } yield MimirRepr(src.P)(table)
 
-    // FIXME: Handle `onUndef`
-    case qscript.LeftShift(src, struct, idStatus, _, onUndef, repair) =>
+    case qscript.LeftShift(src, struct, idStatus, shiftType, onUndef, repair) =>
       import src.P.trans._
 
+      val s = "src"
+      val f = "f"
```
djspiewak: I would have preferred more descriptive names (e.g. `Source` and `Focus`), but this is fine for now.

alissapajer:
@rintcius is this ready for another look, or do you have more you plan to push?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1469,38 +1469,63 @@ trait ColumnarTableModule
...
       val slices2: StreamT[IO, Slice] = slices.map(_.materialized) flatMap { slice =>
-        val (focused, unfocused) = lens(slice.columns)
+        val (focused, emit, unfocused) = lens(slice)
 
         val innerHeads = {
           val unsorted = focused.keys.toVector flatMap {
             case ColumnRef(path, _) => path.head.toVector
           }
 
+          // make sure there's at least 1 element when emitOnUndef
```
djspiewak: I'm actually not sure this semantic is correct. If you have a slice in which *nothing* is defined at all (both inside or outside the focus), which I believe is what happens when `unsorted.isEmpty`, then I think you just emit nothing.

Are there any tests which hit this scenario?

yggdrasil/src/test/scala/quasar/yggdrasil/LeftShiftSpec.scala
```diff
@@ -115,10 +188,13 @@ trait LeftShiftSpec extends TableModuleTestSupport with SpecificationLike {
...
 
-    toJson(table.leftShift(CPath.Identity \ 1)).getJValues mustEqual expected
+    toJson(table.leftShift(CPath.Identity \ 1, emitOnUndef = emit)).getJValues mustEqual expected
   }
 
   // replaces SampleData.toRecord to avoid ordering issues
   def toRecord(indices: Array[Int], jv: JValue): JValue =
     JArray(JArray(indices.map(JNum(_)).toList) :: jv :: Nil)
+
+  def toRecord0(indices: Array[Int]): JValue =
```
djspiewak: Another instance of the `0` suffix. This one actually doesn't delegate to each other. Is there a reason this can't be done with a default parameter? For example:

```scala
   def toRecord(indices: Array[Int], jv: Option[JValue] = None): JValue =
      JArray(JArray(indices.map(JNum(_)).toList) ::: jv.toList)
```

rintcius: I used to like default args, but now I avoid them, e.g. it increases chances of bugs when refactoring.
See e.g. https://github.com/wartremover/wartremover/issues/116#issuecomment-51280344 (and other comments in this issue, but this particular reflects my opinion best)

rintcius:
@alissapajer yes, now that it's green it's ready for review. 

BTW I have updated qz-3833 to reflect the remaining issue.

yggdrasil/src/test/scala/quasar/yggdrasil/LeftShiftSpec.scala
djspiewak: I agree with all the points in that comment, but does this situation fall afoul of any of them? The points are good but there are uses of default parameters which are *not* covered by any of those arguments and seem quite sound to me.

rintcius: I'm interpreting point 4 (faux overloading) as "the all other cases", see also a later comment https://github.com/wartremover/wartremover/issues/116#issuecomment-51326211

If you prefer I'm also happy to merge those 2 functions together like your proposal but without the 2nd arg having a default.

djspiewak: Ok, I see what you mean. So in this case then, I would keep the functions separate, but name them something more descriptive. The `0` suffix can be useful, but when it starts showing up in public APIs it's just unhelpful. So maybe something like `toRecordWithLast` and `toRecord` would be better.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
alissapajer: Yeah, I'd like to know what goes wrong if you remove this check.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1646,11 +1673,18 @@ trait ColumnarTableModule
               ColumnRef(focus \ path, tpe) -> col
           }
 
+          // additional definedness bitset:
+          // we need to keep every first expanded row
+          // in unfocused that's marked to be kept according to the emit bitset
+          val keep: BitSet = definedness |
+            BitSetUtil.filteredRange(0, slice.size * highWaterMark)(i =>
+              (i % highWaterMark == 0) && emit(i / highWaterMark))
```
alissapajer: Why is the first row special? What goes wrong if we just use `definedness` here? We used `definedness` in other places to determine what rows are defined, so I'm concerned that this is maybe a solution that's masking another underlying problem?

(Also, for good measure we should use `===` there. But we use `==` in lots of places in yggdrasil, so if it's a hassle to change we can do that later.)

rintcius: First row is only special in the sense that, that's always there in the expanded version of `unfocused` (i.e. `unfocusedExpanded`) and easiest to implement. It's being expanded  by a factor of `highWatermark`, so every 1st row upto every `highWatermark`-th row should work.

Regarding `keep` vs `definedness`: 
It is confusing, but I think it is correct like this: in case of `emit` we need to change the `definedness` _only_ for the unfocused part,  _not_ for the focused part (the other uses of `definedness` relate to the focused part).

rintcius: Oh, regarding `===`: isn't that `==` in yggdrasil for performance reasons? Probably overhead isn't that much, but still. 

This is BTW another thing I like to have a better feel for: where's the threshold of performance overhead that we can consider neglectable. Was thinking we could discuss this next week somewhere.

alissapajer: Oh yeah that's a good point. We can discuss that next week. The very general rule is that any operations performed _within_ a single slice should be as fast as possible. I agree it would be good to start setting down some bounds and thresholds. :+1:

djspiewak: Yeah definitely we should sit down next week and talk about how performance is achieved in mimir and where the hot paths are. Mimir is a nice project from that standpoint because the hot path is so clearly delineated, and so it's generally pretty easy to tell when you should use abstracted FP things and when you need to dive down into `while` loops and `eq` and stuff.

PR #3599, opened 2018-05-25
Selector refactoring

rintcius:
Refactoring part of an attempt to fix #qz-3679

There's a bunch of different things:
- make spec names consistent
- add and use `plan3_6` in planner specs
- move `Selector` into `selector` package
- improve the type defs in `planner.selector`
- make `rewriteRefs` monadic (`rewriteRefsM`)
- add some `RenderTree` constraints
- deduplicate `filterBuilder`
- rename `SelectorExpr` to `ConditionExpr` (and its children `(Not)Expr` -> `(Not)CondExpr`)
- swap args in functions like this: 
```
  def processMapFuncExpr
    [T[_[_]]: BirecursiveT: ShowT, M[_]: Monad, EX[_]: Traverse, A]
    (funcHandler: AlgebraM[M, MapFunc[T, ?], Fix[EX]], staticHandler: StaticHandler[T, EX])
    (fm: FreeMapA[T, A])
    (recovery: A => Fix[EX])
```
`fm` is now placed after `recovery` to make them more suitable for calling them point-free style. In general: most stable args come first.


mongodb/src/main/scala/quasar/physical/mongodb/planner/selector.scala
```diff
@@ -49,45 +48,48 @@ object selector {
       a.resume.fold(fa => next(fa.toList.apply(index)), κ(a))
   }
 
-  type Partial[T[_[_]], In, Out] = (PartialFunction[List[In], Out], List[InputFinder[T]])
```
jsantos17: 👏

PR #3606, opened 2018-05-30
Use Equal.equalA instead of Equal.equalRef

alissapajer:


alissapajer:
@djspiewak this tests commit association in ch.

edmundnoble:
Nice, almost everywhere it's sound to use `equalRef` is somewhere that `equals` is implemented with it anyway.

PR #3610, opened 2018-06-03
Avoid evaluating Slices effects multiple times when performing cross-joins

wemrysi:
The previous `isEmpty` check on `ltail` and `rtail` evaluates the first effect from each, however one or both tails are then passed to one of the join handling functions, resulting in that same effect being evaluated again. This causes problems when the intermediate `Step` effects are not repeatable (like consuming bytes from a network socket).

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
@@ -1432,37 +1432,27 @@ trait ColumnarTableModule
         val left  = this.canonicalize(Config.minIdealSliceSize, Some(Config.maxSliceSize))
         val right = that.canonicalize(Config.minIdealSliceSize, Some(Config.maxSliceSize))
 
-        left.slices.uncons flatMap {
-          case Some((lhead, ltail)) =>
-            right.slices.uncons flatMap {
-              case Some((rhead, rtail)) =>
-                for {
-                  lempty <- ltail.isEmpty //TODO: Scalaz result here is negated from what it should be!
-                  rempty <- rtail.isEmpty
+        (left.slices.uncons |@| right.slices.uncons).tupled flatMap {
+          case (Some((lhead, ltail)), Some((rhead, rtail))) =>
+            (ltail.uncons |@| rtail.uncons).tupled flatMap {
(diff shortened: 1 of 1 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1432,37 +1432,27 @@ trait ColumnarTableModule
         val left  = this.canonicalize(Config.minIdealSliceSize, Some(Config.maxSliceSize))
         val right = that.canonicalize(Config.minIdealSliceSize, Some(Config.maxSliceSize))
 
-        left.slices.uncons flatMap {
-          case Some((lhead, ltail)) =>
-            right.slices.uncons flatMap {
-              case Some((rhead, rtail)) =>
-                for {
-                  lempty <- ltail.isEmpty //TODO: Scalaz result here is negated from what it should be!
```
alissapajer: I'm not sure I want to know what this comment meant :scream: , but I'm glad to see it gone!

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1432,37 +1432,27 @@ trait ColumnarTableModule
...
+                // left side is a small set, so restart it in memory
+                IO(crossLeftSingle(lhead, rhead :: rthead :: rttail)(transform.initial))
+
+              case (Some((lthead, lttail)), None) =>
+                // right side is a small set, so restart it in memory
+                IO(crossRightSingle(lhead :: lthead :: lttail, rhead)(transform.initial))
+
+              case (Some((lthead, lttail)), Some((rthead, rttail))) =>
+                // both large sets, so just walk the left restarting the right.
+                IO(crossBoth(lhead :: lthead :: lttail, rhead :: rthead :: rttail))
```
djspiewak: This looks fantastic, though would you mind swapping `IO.apply` for `IO.pure`? These should have never been in `apply`.

djspiewak: Actually. I'll just merge. We'll figure out `pure` stuff later.

PR #3611, opened 2018-06-04
Iotaz migration 2

lukaszwawrzyk:
In this PR I refactored rest of the coproducts to iotaz (except those in mongo (Workflow and ExprOp) and anything that depends on quasar.fs).

My two small concerns are
* location of copkArbitrary needed for tests. I put it in quasar.fp.Helpers.
* `:<<:` and `ACopK` types from previous PR are in quasar.fp. Should I maybe move them to `quasar.contrib.iota`?

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
@@ -23,7 +23,7 @@ import argonaut.{Argonaut, CodecJson, DecodeResult}
 
 import fs2.util.Catchable
 
-import scalaz.{~>, Free}
+import scalaz.{~>, Free, :<:}
 import scalaz.concurrent.Task
 import iotaz.{CopK, TNilK}
 import iotaz.TListK.:::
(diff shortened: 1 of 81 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
```diff
@@ -54,11 +58,17 @@ package object vfs {
...
 
   object POSIXWithTask {
-    def generalize[S[_]]: GeneralizeSyntax[S] = new GeneralizeSyntax[S] {}
+    def generalize[S[a] <: ACopK[a]]: GeneralizeSyntax[S] = new GeneralizeSyntax[S] {}
 
-    trait GeneralizeSyntax[S[_]] {
-      def apply[A](pwt: POSIXWithTask[A])(implicit IP: POSIXOp :<: S, IT: Task :<: S): Free[S, A] =
-        pwt.mapSuspension(λ[Coproduct[POSIXOp, Task, ?] ~> S](_.run.fold(IP.inj, IT.inj)))
+    private val JP = CopK.Inject[POSIXOp, POSIXWithTaskCopK]
+    private val JT = CopK.Inject[Task,    POSIXWithTaskCopK]
```
djspiewak: For the record, I would prefer not having the vertical alignment. Doesn't hold up the merge of this PR, but I wanted to call it out as something I would rather avoid.

djspiewak:
@lukaszwawrzyk To answer your questions, yes, I think both of those types should be in `quasar.contrib.iota`. I think `copkArbitrary` is fine where it is so long as it didn't pull ScalaCheck as a main dependency.

lukaszwawrzyk:
@djspiewak I did not need to touch any dependencies to implement `Arbintrary[CopK]`.
I fixed the rest of the things. If there are be no more things to change, this is ready to merge if build passes.

PR #3621, opened 2018-06-06
Leftshift OnUndefined.Emit in SlamEngine (part 2)

rintcius:
leftovers that didn't make it in #3597


rintcius:
@alissapajer ready for review (when it's green).

PR #3625, opened 2018-06-07
Upgrade fs2 and cats-effect to 1.x

wemrysi:


foundation/src/main/scala/quasar/contrib/scalaz/catchable.scala
```diff
@@ -34,6 +35,15 @@ trait CatchableInstances extends CatchableInstances0 {
 trait CatchableInstances0 {
   implicit def injectableTaskCatchable[S[_]](implicit I: Task :<: S): Catchable[Free[S, ?]] =
     catchable.freeCatchable[Task, S](injectNT[Task, S], projectNT[Task, S])
+
+  implicit val catsIOCatchable: Catchable[IO] =
+    new Catchable[IO] {
+      def attempt[A](ioa: IO[A]): IO[Throwable \/ A] =
+        ioa.attempt.map(_.disjunction)
```
djspiewak: Random heads up: you can also do `map(_.asScalaz)` and shims will do things. Doesn't matter here, but for future reference.

mimir/src/main/scala/quasar/mimir/TablePager.scala
```diff
@@ -18,77 +18,71 @@ package quasar.mimir
...
-      val ta = driver >> queue.enqueue1(\/-(Vector.empty))
+      val ta = driver >> queue.enqueue1(Right(Vector.empty))
 
-      ta unsafePerformAsync {
-        case -\/(t) => queue.enqueue1(-\/(t)).unsafePerformAsync(_ => ())
-        case \/-(_) => ()
+      ta unsafeRunAsync {
+        case Left(t) => queue.enqueue1(Left(t)).unsafeRunAsync(_ => ())
+        case Right(_) => ()
       }
```
djspiewak: These lines give me hives. I can't wait until we can get rid of them.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
```diff
@@ -19,18 +19,14 @@ package quasar.yggdrasil.vfs
...
-import fs2.util.Suspendable
 
 import pathy.Path
 
-import scalaz.{~>, -\/, \/-, Free}
-import scalaz.concurrent.Task
-import scalaz.std.list._
-import scalaz.syntax.traverse._
-import iotaz.CopK
+import scalaz.{~>, -\/, \/-, Scalaz}, Scalaz._
```
djspiewak: Do we need the `Scalaz._` import here?

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
```diff
@@ -41,40 +37,40 @@ import java.util.UUID
 object RealPOSIX {
   import POSIXOp._
 
-  def apply(root: File): Task[POSIXOp ~> Task] = {
+  def apply[F[_]](root: File)(implicit F: Sync[F]): F[POSIXOp ~> F] = {
```
djspiewak: We should replace this with `Async` and use `readAllAsync`/`writeAllAsync` below.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
```diff
@@ -84,13 +80,13 @@ object RealPOSIX {
...
         case MkDir(target) =>
-          Task.delay(canonicalize(target).mkdirs()).void
+          F.delay(canonicalize(target).mkdirs()).void
 
         case LinkDir(src, target) =>
           val psrc = canonicalize(src).toPath()
           val ptarget = canonicalize(target).toPath()
 
-          Task delay {
+          F.delay {
```
djspiewak: Random style note: when using newlines, I like to use curly braces (like here) and spaces. So rather than `F.delay {` it would be `F delay {`.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -20,32 +20,21 @@ import slamdata.Predef.{SuppressWarnings, Array}
...
-import scalaz.std.map._
-import scalaz.std.string._
-import scalaz.std.vector._
-import scalaz.syntax.equal._
-import scalaz.syntax.monad._
-import scalaz.syntax.monoid._
-import scalaz.syntax.show._
-import scalaz.syntax.traverse._
-import scalaz.syntax.std.boolean._
+import scalaz.{~>, Equal, Free, Monad, Show, StateT, Scalaz}, Scalaz._
```
djspiewak: `Scalaz._` :-(

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -54,9 +43,13 @@ import scodec.bits.ByteVector
 import scodec.codecs, codecs.uint16
 import scodec.interop.scalaz.ByteVectorMonoidInstance
 
+import shims.monoidToCats
```
djspiewak: Why the specific import here?

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -533,95 +528,100 @@ object FreeVFS {
...
     root: File,
     @volatile private var current: VFS,
-    worker: mutable.Queue[Task, Task[Unit]],    // this exists solely for #serialize
-    interp: POSIXWithTask ~> Task) {
+    worker: mutable.Queue[F, F[Unit]],    // this exists solely for #serialize
+    interp: POSIXWithIO ~> F)(
+    implicit F: Concurrent[F]) {
 
   import SerialVFS.S
+  import shims.monadToScalaz
```
djspiewak: Why the specific import?

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
```diff
@@ -42,57 +48,51 @@ object FreeVFSSpecs extends Specification with DisjunctionMatchers {
...
         }
 
-        _ <- H.pattern[Unit] {
-          case CPR(ta) => ta
-        }
-
-        _ <- H.pattern[Unit] {
-          case CPR(ta) => ta
-        }
+        _ <- ioPattern.replicateM(11)
```
djspiewak: `11` 😢

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
```diff
@@ -254,50 +250,44 @@ object FreeVFSSpecs extends Specification with DisjunctionMatchers {
...
               List(Path.dir("META"))
             }
         }
       } yield ()
 
-      val vfs = interp(FreeVFS.init[S](BaseDir)).unsafePerformSyncAttempt
+      val vfs = interp(FreeVFS.init[S](BaseDir)).attempt.unsafeRunSync
 
-      vfs.leftMap(_.getMessage) must be_-\/("Unexpected VERSION, 0100000000000000")
+      vfs.disjunction.leftMap(_.getMessage) must be_-\/("Unexpected VERSION, 0100000000000000")
```
djspiewak: Why not just…

```scala
vfs.leftMap(_.getMessage) must beLeft("Unexpected VERSION, 0100000000000000")
```

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
```diff
@@ -375,35 +363,39 @@ object FreeVFSSpecs extends Specification with DisjunctionMatchers {
...
-            Task delay {
+            IO {
               target mustEqual (headDir </> Path.file("paths.json"))
 
               Stream(ByteVector("{}".getBytes))
             }
         }
 
-        _ <- H.pattern[Stream[POSIXWithTask, ByteVector]] {
+        _ <- ioPattern.replicateM(15)
```
djspiewak: `15`…

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/SerialVFSSpecs.scala
```diff
@@ -16,86 +16,69 @@
...
-
-          _ <- vfs.commit(blob, version)
-          _ <- vfs.link(blob, Path.rootDir </> Path.file("foo"))
-        } yield ()
-
-        _ <- Stream.eval(ta)
-      } yield ()
-
-      test.take(1).run.unsafePerformSync
+      val test = SerialVFS[IO](base).mapAsync(1)(vfs => for {
```
djspiewak: What is `mapAsync`?

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
wemrysi: Yeah, if I remember correctly, I added it because I needed both `Traverse` and `Monad` syntax and didn't want to fiddle with the overlapping import exclusions required to make `Functor` syntax work.

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/SerialVFSSpecs.scala
wemrysi: It's like `evalMap` but runs the effects asynchronously. I tried all manner of things to avoid this, but without it we deadlock. I'm not sure what changed between versions to cause this, but something did (or I made something worse).

The only place we call `SerialVFS` in production code we `unsafeRunAsync` the outer effect, so we coincidentally sidestep the problem, but something about sequencing usage of the `vfs` within the stream used to obtain it causes deadlocks.

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
wemrysi: Sure, was just attempting to keep diffs small.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
wemrysi: Yeah, I tend to always just use spaces, no matter what the context, but you've dinged me before in PRs for that, so I was attempting to avoid that, to my peril, apparently ;)

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
djspiewak: Yeah, IMO that's the reason to use `Scalaz._`.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
djspiewak: LOL! My current convention (which I got from Martin) is this:

- If it's a single line thing (e.g. `foo.map(_.bar)`), then use dots and parens
- If it's a multi-line thing (i.e. you need multiple lines in the lambda) use spaces and curlies

This style allows you to chain concise things really nicely and avoids the association ambiguity of dots on curlies (as well as postfix methods), while simultaneously encouraging you to *not* chain multi-line things and avoiding issues with semi-colon inference. So I think it's the best of both worlds.

Basically, you *almost* did it right (according to my style), you just had `F.delay` rather than `F delay`. :-D If you had bee doing something like wrapping a `println`, then `F.delay(println("foo"))` would have been correct.

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
djspiewak: Gotcha. I'm ok either way.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
alissapajer: We really need a formatter that works...

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
alissapajer: What are these incrementing? I feel like I should be crying too.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
alissapajer: Yeah, that's why I've been largely okay with using `import Scalaz._`, even though ideally I'd still much prefer the individual `syntax` imports.

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
wemrysi: You will be, you...will be.

This is to "drain" the `smock.Harness` of `IO` suspensions that were introduced via various interactions with `Stream[POSIXWithIO, A]`. "How did you determine it should be `11`?", you ask. Trial and error, naturally.

alissapajer: :sob:

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
wemrysi: Aye, we should, but I don't think it's that easy. For the read side, we could just use `IO` and then `translate` that into `POSIXWithIO`. However, for the write side, since the effect type appears invariantly, we'd need an `Effect[POSIXWithIO]` (and a `Timer[POSIXWithIO]`) which we can't write.

Maybe we could change the `OpenR` and `OpenW` cases to just deal in `Stream[IO, Byte]` and `Sink[IO, Byte]`, respectively?

Either way I'd prefer to leave this for another PR.

djspiewak: Why can't we write `Timer[POSIXWithIO]`? It should be straightforward to lift.

You're right that we would need an `Effect`. Nuts. *That* one I think we can't write. This flipping of type bounds is really really frustrating and constraining.

Agreed that we should deal with this in a future PR. For whatever reason I thought it would be trivial.

djspiewak:
@wemrysi Floating an idea regarding the smock woes… What if we provide a utility function in smock which applies when you're using a coproduct inside your `Free` where one side matches the effect monad of the test itself? Such a function would then transparently sequence all effects which appear on the impure side of the coproduct, allowing you to write resilient tests on the pure side.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
wemrysi: @djspiewak the "can't write" was regarding `Effect`, not `Timer`.

djspiewak: Gotcha. I totally agree then.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
wemrysi: The wildcard import causes all sorts of havoc with the implicit `Sync[Free[POSIXWithIO, ?]]` instance in scope.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
wemrysi: See above.

wemrysi:
@djspiewak Aye, that sounds good, much more declarative.

djspiewak:
@wemrysi Race you to a smock PR. ;-)

djspiewak:
Swapped to feature since it's a compatibility-affecting change due to the classpath shift.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
djspiewak: What kind of havoc?

djspiewak: OOOOOH! It's because `Sync <: Monad`. I get it. Damn it.

mimir/src/main/scala/quasar/mimir/LightweightLocal.scala
```diff
@@ -24,6 +24,7 @@ import quasar.fs.mount.ConnectionUri
 
 import eu.timepit.refined.auto._
 import fs2.Stream
+import io.chrisdavenport.scalaz.task._
```
djspiewak: How much of using `Effect[Task]` do you suppose we're doing in the test codepath? Any?

edmundnoble: Certainly some. Old FS2 uses it for basic stuff.

mimir/src/main/scala/quasar/mimir/SlamEngine.scala
```diff
@@ -246,13 +251,13 @@ trait SlamEngine extends BackendModule with Logging with DefaultAnalyzeModule {
 
         populatorWithTermination = populator.flatMap(_ => q.enqueue1(Vector.empty))
 
-        ingestor = repr.P.ingest(path, q.dequeue.takeWhile(_.nonEmpty).flatMap(Stream.emits)).run
+        ingestor = repr.P.ingest(path, q.dequeue.takeWhile(_.nonEmpty).flatMap(Stream.emits(_).covary[IO])).run
 
         // generally this function is bad news (TODO provide a way to ingest as a Stream)
-        _ <- Task.gatherUnordered(Seq(populatorWithTermination.to[Task], ingestor.to[Task])).to[IO]
+        _ <- Parallel.parSequence_[List, IO, IO.Par, Any](List(populatorWithTermination, ingestor))
```
djspiewak: I'm not sure I trust `Parallel`, but I doubt it's causing any huge issues.

edmundnoble: You and me both.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
```diff
@@ -16,58 +16,82 @@
...
   type POSIX[A] = Free[POSIXOp, A]
-  type POSIXWithTaskCopK[A] = CopK[POSIXOp ::: Task ::: TNilK, A]
-  type POSIXWithTask[A] = Free[POSIXWithTaskCopK, A]
+  type POSIXWithIOCopK[A] = CopK[POSIXOp ::: IO ::: TNilK, A]
+  type POSIXWithIO[A] = Free[POSIXWithIOCopK, A]
 
   // this is needed kind of a lot
-  private[vfs] implicit def catchableForS[S[a] <: ACopK[a]](implicit I: Task :<<: S): Catchable[Free[S, ?]] = {
-    val delegate = catchable.copKinjectableTaskCatchable[S]
+  private[vfs] implicit def syncForS[S[a] <: ACopK[a]](implicit I: IO :<<: S): Sync[Free[S, ?]] =
```
djspiewak: This is like, the most performance-suspicious thing I've seen so far, but the VFS operations really aren't in the hot path at all (if they were, we wouldn't be using `Free`), so I don't understand.

wemrysi: Yeah, and in the newer test suite, I don't think the `vfs` is used at all.

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/SerialVFSSpecs.scala
djspiewak: Summoning @mpilquist! *does a sinister-looking dance and drops a smoke pellet*

`SerialVFS.apply` returns something constructed with this form:

```scala
Stream.emit(svfs).merge(worker.dequeue.evalMap(t => t).drain)
```

`worker` is constructed as `async.boundedQueue[F, F[Unit]](10)` and is populated by the following:

```scala
  private def serialize[A](fa: F[A]): F[A] = {
    for {
      ref <- Deferred[F, Either[Throwable, A]]
      _ <- worker.enqueue1(F.attempt(fa).flatMap(ref.complete))
      r <- ref.get
      a <- F.fromEither(r)
    } yield a
  }
```

Basically, I can't for the life of me think of why `evalMap` would deadlock but `mapAsync` *wouldn't*, and it's quite concerning. I'm hesitant to say that this is a bug in fs2, but I can't think of anything else.

mpilquist: I'll take a closer look but on a lark, could you try `concurrently` in place of `merge`?

djspiewak: @mpilquist We actually thought about that. The problem with concurrently is it terminated immediately when one side halts, which would be right away since we’re just using emit.

The pattern we’re encoding here is where you have some resource with associated concurrent activity that happens independently and produces no results (only effects). So the resource gets bracketed and then merged with the drained stream that runs the effects.

wemrysi: @djspiewak @mpilquist I just tried `concurrently` again and it actually works when we're playing by the rules in `SerialVFSSpecs` and I'm able to change all the `mapAsync`s there to `evalMap`s.

However, since our production usage runs the outer stream due to reasons, _that_ usage halts things immediately (i.e. our regression tests don't run).

mpilquist:
Interesting- how about Stream(x, y).join(2) in place of merge (just as a test)?

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/SerialVFSSpecs.scala
djspiewak: @wemrysi I wonder if we could just thread the stream out of the Precog constructor. It’s not the only Stream there if I recall, and with your new file system, don’t we have the ability to return the file system init within Stream?

wemrysi: @djspiewak Yeah, that'd definitely be preferable to what we're currently doing and yes we can use a stream in the new setup as well.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -610,19 +611,27 @@ object SerialVFS {
    * Returns a stream which will immediately emit SerialVFS and then
    * continue running until the end of the world.  When the stream ends,
    * SerialVFS will no longer function.
+   *
+   * TODO: Return a `Stream[F, SerialVFS[F]]` to allow for better resource
+   *       handling once we're rid of `BackendModule`.
    */
-  def apply[F[_]: Concurrent](root: File): Stream[F, SerialVFS[F]] = {
+  def apply[F[_]: Concurrent](root: File): F[Disposable[F, SerialVFS[F]]] = {
```
djspiewak: `Disposable` is back?

wemrysi: For the time being, I think so. It is slightly better than `(A, F[Unit])` and that’s what we’re stuck with until we can get rid of the places that can’t use `Stream`, like `FileSystemDef`. The new constructs will just turn it into a stream, however.

wemrysi:
So the failure in the `lwc_local` tests is due to the lib we’re using to provide a `cats.effect.Effect[Task]` instance not supporting `cats-effect-1.0.0-RC2`. I’m looking into adding said support.

wemrysi:
I've replaced our usage of https://github.com/christopherdavenport/scalaz-task-effect with a local implementation compatible with `cats-effect-1.0.0-RC2`. I have [a PR](https://github.com/christopherdavenport/scalaz-task-effect/pull/19) open upstream and we can switch back to it once it's been merged if we're still using `Task` at that point.

PR #3628, opened 2018-06-12
[WIP] maxSliceColumns for canonicalize

rintcius:
not for merge


yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -797,17 +784,22 @@ trait ColumnarTableModule
...
+      * Returns a table where each slice (except maybe the last) has slice size `sliceRows`.
+      * Also removes slices of size zero. If an optional `maxSliceRows` size is provided,
+      * then the slices need only land in the range between `sliceRows` and `maxSliceRows`.
       * For slices being loaded from ingest, it is often the case that we are missing a
       * few rows at the end, so we shouldn't be too strict.
       */
-    def canonicalize(length: Int, maxLength0: Option[Int] = None): Table = {
-      val minLength = length
-      val maxLength = maxLength0 getOrElse length
+    def canonicalize(sliceRows: Int, maxSliceRows: Option[Int], maxColumns: Int): Table = {
```
rintcius: @alissapajer @djspiewak I have changed `canonicalize` with the extra assumption for `maxSliceColumns` that you told me about last week. Can you review? 

NB: It's not based on https://github.com/wemrysi/quasar/pull/11 yet (was planning to do that once that's in master), but the logic should not change. Also see changes in `CanonicalizeSpec`.

yggdrasil/src/test/scala/quasar/yggdrasil/CanonicalizeSpecs.scala
```diff
@@ -18,50 +18,62 @@ package quasar.yggdrasil
...
-      {"foo":14}
+      {"foo":11, "baz": 11},
+      {"foo":12, "baz": 12},
+      {"foo":13, "baz": 13},
+      {"foo":14, "bar14": 14}
     ]""")
 
     val sample = SampleData(elements.toStream)
-    fromSample(sample)
+    fromSample(sample, Some(3))
```
rintcius: note: 3 is the max number of rows per slice (i.e this sample has 5 slices)

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -797,17 +784,22 @@ trait ColumnarTableModule
...
+      val minRows = sliceRows
```
alissapajer: Should be name the `sliceRows` function parameter `minRows` so we don't have to rename it here?

rintcius: The reason that I choose to keep a similar naming pattern as the old code was because that makes the code a bit more consistent (eg note that the same arg names are used in the interface & that `sliceRows` takes the role of both min and max if `maxSliceRows = None`)

djspiewak:
@rintcius @alissapajer Are we going to try to land this for 0.10? We're not calling `canonicalize` at all right now, right? So we can put this off?

alissapajer:
@djspiewak I'm not going to have time to look at this PR this week. Bumping it over to you. I think we can address this after July 23. Correct me if there is some reason we need this urgently.

alissapajer:
@djspiewak haha I actually didn't see your comment above. You beat me by 13 minutes!

alissapajer:
In any case, agreed that we can put this off.

alissapajer:
@djspiewak we actually do call canonicalize in `cross` [here](https://github.com/slamdata/quasar/blob/master/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala#L1448-L1450)

rintcius:
@djspiewak I intended to pick it up after I have implemented the must haves for coming release, but that was assuming that `canonicalize` is being called. If it's not being called then I'll discuss with @alissapajer how/when to deal with this. In either way, I don't think there will be time for this in this week.

alissapajer:
I'm going to close this so it's not sitting open in our PR queue. When we work on this again, we can reopen a new PR with the changes cleaned up and rebased.

PR #3630, opened 2018-06-13
Unsafely don't page the entire lwc dataset into memory

alissapajer:


alissapajer:
[ch72]

alissapajer:
@djspiewak ready. `lwc_local` took ~27 min which is as it has been.

PR #3633, opened 2018-06-19
Remove some unused imports and an unused val

alissapajer:


alissapajer:
@djspiewak bumping this over to you.

PR #3634, opened 2018-06-19
Delete all the things

alissapajer:


alissapajer:
[ch160]

alissapajer:
This is WIP because if Travis doesn't pass I'm going to rebase it so that each removal is in its own commit.

We can also push a tag for each of the the CB and ML removals.

alissapajer:
Ready for review. Travis ran as expected. This leaves us with 5 matrix builds: `lwc_local`, `mimir`, `mongodb_3_4`, `mongodb_3_4_13`, and `mongodb_3_6`.

djspiewak:
I think we can also drop the old mongo 3.4 build and only retain the latest 3.4.x (x ≥ 4). So that'll drop us down another matrix build.

alissapajer:
For posterity, here's what I ran to delete backends from the `.test` files:
```bash
$ find . -type f -name "*.test" -print0 | xargs -0 sed -i '' -e '/\S*"mongodb_read_only.*$/d'
```

PR #3635, opened 2018-06-19
In ExpandShifts also repair when focus is undefined

rintcius:


rintcius:
[ch 183]

rintcius:
[ch183]

djspiewak:
@rintcius There are conflicts now, but I'm pretty sure that the build errors are also legitimate (which is why I've held off on merging).

rintcius:
@djspiewak yeah, it's difficult to say for sure with no error output, but I have just pushed a commit based on a guess that one of the tests times out. 

If this is the case we should look into fixing this: the test should timeout and fail before the whole build times out.

PR #3640, opened 2018-06-19
Remove support for mongo 3.4 (we still support 3.4.13)

alissapajer:
We still support mongo `3.4.13`.


djspiewak:
Do the `Plan2ExactSpec` tests exist against 3.4?

alissapajer:
@djspiewak probably not. The mongo unit tests should be written such that for each single test there is a list of expectations, one for each mongo version supported. Instead they are written such that each single test _can_ only test one mongo version. Thus we don't have an easy way to keep all the tests enabled when we remove support for old mongo versions.

The alternative to deleting these tests would be to run them instead against 3.6 and then blindly update their expectations. That was more work than I felt was warranted, though. So if we want to keep these tests around, I can revert 03b9139 and we can deal with them later. But to be fair, they were not testing `3.4.13` or `3.6` and so were not testing anything our customers would ever use.

rintcius:
Some of these tests - like the ones that test map-reduce fallback mechanism - should be migrated. We need to find different queries to test the fallback to map-reduce here. For other queries it looks like they can either be removed or be upgraded to test against 3.6.

I think it's better to revert 03b9139 for now and deal with it later. How about dealing with this when we pick up qz-2802 / qz-3699 which we have discussed in the past? These planner tests are very useful, but when integrated with ITs they'd become far less tedious to maintain and more useful too (they will test against the query that's generated instead of an artifact in the middle of the pipeline & they can be used for all connectors instead of just mongo)

alissapajer:
@djspiewak reverted the commit. This should be ready to go.

PR #3642, opened 2018-06-20
Include the impl project in root aggregation

wemrysi:
I forgot to aggregate the new `impl` project in the `root` project, which meant some tests weren't being run.


djspiewak:
Stealing this.

PR #3644, opened 2018-06-21
Table.fromRValuesStream, with column awareness

edmundnoble:


yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1894,16 +1894,95 @@ object Slice {
...
+          }
+        } else {
+          val flattened = next.head.flattenWithPath
+          val newAcc = updateRefs(flattened, acc, rows, maxRows)
+          val newCols = newAcc.size
+          if (newCols > maxColumns && cols > 1) {
+            // println("we have too many columns, cutting slice early")
+            (new Slice {
+              val size = rows
+              val columns = acc
```
rintcius: Continuing discussion of previous PR..

> `I think we don't want to make slices smaller than minSliceRows.`

Not by default, but if we have a column overflow then we can end up with smaller slices, even with just 1 row. Looks to me like it's worth down-sizing the array allocations in that scenario.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1894,16 +1894,95 @@ object Slice {
...
+
+    rec(Slice.ArraySliced(new Array[RValue](0), 0, 0), values).stream
+  }
+
   def fromJValues(values: Stream[JValue]): Slice =
-    fromRValues(values.flatMap(RValue.fromJValue))
+    fromRValues(values.map(RValue.fromJValueRaw))
 
+  // don't use this anymore. It doesn't limit the slice size properly,
+  // unlike allFromRValues and fromRValuesStep.
```
rintcius: Can't we get rid of this function yet?

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -71,6 +72,221 @@ class SliceSpec extends Specification with ScalaCheck {
...
+        COffsetDateTime(OffsetDateTime.parse("+28785-12-16T09:45:14.815967029+16:18:49")), CNull,
+        CInterval(DateTimeInterval.parse("P-9036285Y44M-411DT584566H4M1.912409135S").get), CDouble(-2.3703641119979175E233), CLocalDate(LocalDate.parse("+4396657-12-11")),
+        CNull, CNull, CEmptyObject, CInterval(DateTimeInterval.parse("PT550632H50M7.334061745S").get), CLocalTime(LocalTime.parse("17:00:02.808168924")),
+        CLong(-2051368316550135838L), COffsetDateTime(OffsetDateTime.parse("-9897897-05-10T00:02:09.498265366+04:48:25")), CDouble(1.0859578724304194E-147),
+        CNull, CTrue, CLong(2512754280505120381L), CLocalDateTime(LocalDateTime.parse("-7951878-05-18T11:03:35.812478165")), CEmptyArray, CEmptyObject)
+
+    "construct slices from pathological example" in {
+      // this example is pathological because when dealing with it there's a difference between cutting off
+      // a slice only when the column limit is reached and deciding beforehand whether adding a value would
+      // exceed the column limit.
```
rintcius: looks like a nice test, but I don't get your comment, can you elaborate a bit?

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -71,6 +72,221 @@ class SliceSpec extends Specification with ScalaCheck {
...
+  "fromRValues" should {
+
+    val v = List(CString("x"), CNum(42))
+
+    "construct slices from a simple vector" in {
+      "fits in 1 slice" >> testFromRValuesFittingIn1Slice(v)
+      "maxSliceRows < nrRows" >> testFromRValuesMaxSliceRowsOverflow(v)
+      "maxSliceRows = 1" >> testFromRValuesMaxSliceRows1(v)
+      "maxSliceColumns = nrColumns of biggest value" >> testFromRValuesMaxSliceColumnsEqualsBiggestValue(v)
+      // "maxSliceColumns < nrColumns of biggest value" >> testFromRValuesMaxSliceColumnsLowerThanBiggestValue(v)
```
rintcius: why is this commented out?

foundation/src/main/scala/quasar/contrib/fs2/stream.scala
```diff
@@ -16,8 +16,9 @@
 
 package quasar.contrib.fs2
 
-import slamdata.Predef.{Throwable, Unit}
+import slamdata.Predef.Throwable
 
+import scala.Unit
```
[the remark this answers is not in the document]
djspiewak: It is.

wemrysi:
@edmundnoble Did downgrading `cats-effect`/`fs2` actually have an effect on performance, I thought we determined the issue was still present in the `fs2-0.10.x` series.

wemrysi:
@edmundnoble Oh, nevermind, I was thrown off track by one of the commit messages, looks like you undid all the reverting.

edmundnoble:
@wemrysi I undid it all, yeah. I decided to leave that stuff to later.

it/src/main/resources/tests/temporal/dateFilter.test
```diff
@@ -1,10 +1,6 @@
 {
     "name": "filter on date part",
 
-    "backends": {
-        "lwc_local": "pending",
-        "mimir": "pending"
-    },
```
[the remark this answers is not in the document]
edmundnoble: It was making `fromRValues` not go through `JValue` first, because `JValue` destroys date/time values.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I was told we should leave this for the old API.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
edmundnoble: The difference here is between two algorithms for deciding when the column limit should force us to move to the next slice:
a) I inspected an RValue and decided it would put us above the column limit, so we'll leave it for the next slice, and
b) We always add values if we're currently below the column limit.

I implemented b) first, but it failed one of your tests (the random one), and this was the sample which did it.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
edmundnoble: Good point.

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
```diff
@@ -44,7 +39,7 @@ import scalaz._, Scalaz._
 
 final class FederatedShiftedReadPlanner[
     T[_[_]]: BirecursiveT: EqualT: ShowT,
-    F[_]: LiftIO: Monad: PlannerErrorME: MonadFinalizers[?[_], IO]](
+    F[_]: LiftIO: Monad: PlannerErrorME: MonadTell_[?[_], DList[IO[Unit]]]](
```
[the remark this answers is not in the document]
djspiewak: I think there's also a type alias, right?

alissapajer:
[ch71]

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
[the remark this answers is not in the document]
rintcius: @djspiewak those aliases are removed in this PR, see https://github.com/slamdata/quasar/pull/3644/files#diff-2ee44a217677c1a18c6b494972360fdcL30 - personally I am not that opinionated about these aliases, just a slight preference to keep them in, but happy to go with other opinions

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
rintcius: ok, thanks for clarifying @edmundnoble

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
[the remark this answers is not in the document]
alissapajer: I'd prefer to keep the aliases in; I think the name helps clarify the type since it's a type that includes `Unit`.

yggdrasil/src/main/scala/quasar/yggdrasil/TableModule.scala
```diff
@@ -35,6 +36,7 @@ import java.time.{LocalDate, LocalDateTime, LocalTime, OffsetDateTime, OffsetTim
 
 // TODO: define better upper/lower bound methods, better comparisons,
 // better names, better everything!
+// TODO: investigate adding columns
```
alissapajer: What does this comment mean?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -33,9 +33,25 @@ trait DefinedAtIndex {
 
 trait ArrayColumn[@specialized(Boolean, Long, Double) A] extends DefinedAtIndex with ExtensibleColumn {
   def update(row: Int, value: A): Unit
+  def resize(size: Int): ArrayColumn[A]
 }
 
-class ArrayHomogeneousArrayColumn[@specialized(Boolean, Long, Double) A](val defined: BitSet, values: Array[Array[A]])(val tpe: CArrayType[A])
+object ArrayColumn {
+  def resizeArray[A](arr: Array[A], size: Int)(implicit A: ClassTag[A]): Array[A] = {
```
alissapajer: This will lose values from `arr` if a `size` is chosen that is smaller than the number of values in `arr`. Is that the intention?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -33,9 +33,25 @@ trait DefinedAtIndex {
...
+    val newArr = new Array[A](size)
+    System.arraycopy(arr, 0, newArr, 0, Math.min(size, arr.length));
+    newArr
+  }
+
+  def resizeBitSet(bs: BitSet, size: Int): BitSet = {
+    val arr = new Array[Long](size)
+    // 64 bits per long, 2^6 == 64
+    System.arraycopy(bs.getBits(), 0, arr, 0, size >> 6)
```
alissapajer: Say `size` is `200`. Then `size >> 6` is `3`. This means we'd copy over three `Long` values. 

What's interesting here is that we are copying over the bits in `Long`-sized (64-bit sized) chunks. So, when `size` is `200` meaning (unless I've misunderstood) that we want to copy over `200` bits, we'd actually only copy over `64 * 3 = 192` bits.

alissapajer: Also, we're creating an `Array[Long]` of `size` but then only copying over `size >> 6` elements.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1894,16 +1894,95 @@ object Slice {
     }
   }
 
+  final case class ArraySliced[A](arr: Array[A], start: Int, size: Int) {
+    def head: A = arr(start)
+    def tail: ArraySliced[A] = ArraySliced(arr, start + 1, size - 1)
+  }
+
+  def fromRValuesStep(values: ArraySliced[RValue], maxRows: Int, maxColumns: Int): (Slice, ArraySliced[RValue]) = {
```
alissapajer: This is production code, so we should have some unit tests for this function specifically. The logic is gnarly, and we're only hitting it in the `lwc_local` integration tests. We should make sure to test the edge cases.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: Once the old API is gone, we should be able to move this function into a `test/` dir.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: This function reminded me of this poem.
```
Parsley
Is gharsley.
--Ogden Nash
```

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1894,16 +1894,95 @@ object Slice {
...
+      }
+    }
+    inner(values, 0, 0, Map.empty, 64)
+  }
+
+  def allFromRValues(values: fs2.Stream[IO, RValue], maxRows: Option[Int] = None, maxColumns: Option[Int] = None): fs2.Stream[IO, Slice] = {
+    val maxRowsC = maxRows.getOrElse(Config.maxSliceRows)
+    val maxColumnsC = maxColumns.getOrElse(Config.maxSliceColumns)
+    // println(s"maxRows: $maxRowsC, maxCols: $maxColumnsC")
+    def rec(next: Slice.ArraySliced[RValue], values: fs2.Stream[IO, RValue]): fs2.Pull[IO, Slice, Unit] =
```
alissapajer: Can we make this `@tailrec`?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1894,16 +1894,95 @@ object Slice {
     }
   }
 
+  final case class ArraySliced[A](arr: Array[A], start: Int, size: Int) {
+    def head: A = arr(start)
+    def tail: ArraySliced[A] = ArraySliced(arr, start + 1, size - 1)
+  }
+
+  def fromRValuesStep(values: ArraySliced[RValue], maxRows: Int, maxColumns: Int): (Slice, ArraySliced[RValue]) = {
+    def inner(next: ArraySliced[RValue], rows: Int, cols: Int, acc: Map[ColumnRef, ArrayColumn[_]], _size: Int): (Slice, ArraySliced[RValue]) = {
```
alissapajer: Same, can this be `@tailrec`?

alissapajer: @djspiewak (or anyone) hm, are there performance concerns we need to think about regarding tailrec-ness and doing computations on the stack as we recurse?

djspiewak: `@tailrec` is considerably faster than regular recursion when you're talking about the hot path. It doesn't have to allocate new stack frames (which is much faster than heap allocation, but still a cost) and doesn't have to re-check polymorphism on the recursion, etc. It literally turns into a `goto`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1894,16 +1894,95 @@ object Slice {
...
+              rec(Slice.ArraySliced(chunkArr, 0, chunkArr.length), next)
+            case None                =>
+              // println("Finished top-level loop")
+              fs2.Pull.done
+          }
+        } yield ()
+      } else {
+        // println(s"extracting slice from data with size ${next.size}")
+        val (nextSlice, remainingData) = Slice.fromRValuesStep(next, maxRowsC, maxColumnsC)
+        fs2.Pull.output1(nextSlice) >> rec(remainingData, values)
```
alissapajer: I haven't followed all the logic through completely yet, but here's one potential concern: Let's say we take a chunk from the `Stream` and this chunk has 1000 rows, the first 998 of which we turn into a `Slice`. Using this algorithm, won't those remaining two rows turn into a `Slice` of size two? 

If so, then we'd have to `canonicalize` in order to get a `Table` with correctly-sized `Slice`s. We have to pay the cost at some point of reassociating slice boundaries (whether those be actual `Slice`s or chunks coming in from `fs2.Stream`). I'm not sure yet if that cost is better paid within this function, or calling `canonicalize` is okay.

edmundnoble:
Much of the code in `fromRValuesStep` is completely incorrect. While working on performance tests for mimir, I found a hint that this function was implemented incorrectly when it crashed with OOM when I was producing large columns. Here's a list of the things that were wrong:

- `updateRefs` was only updating the index at `maxRows`. This meant all rows were empty but one.
- We weren't keeping track of the row counts per column separately. This means that having three columns each with 10 rows would lead to the method thinking you had 30 rows in the last column.
- Rows were added *after* checking that the *current* row count didn't exceed `_size` or `maxRows`, leading to array indexing exceptions. This wasn't noticeable until the previous bug was fixed.

Here's a list of the things that were noticed that were performance bugs:
- Minimum slice size of 64 is unnecessarily pessimistic for flat data. We can find out how much of the data is scalar and use that to make a larger estimate when it's feasible.
- ArrayBoolColumn didn't have an `empty(size: Int)` constructor, leading to runtime resizing of the `BitSet`.
- Tuples (`->`) in `updateRefs`, `foldLeft` in `updateRefs`.

And for @alissapajer:
- `ArrayColumn.resize` was made to return the same column instead of making the column smaller.

I've also found a bug, for which I've left a comment:
```scala
          // this shouldn't work. `updateRefs` mutates the underlying arrays of `acc`,
          // so just "throwing away" `newAcc` later on when we decide we don't
          // want it anymore is not possible.
          val newAcc = updateRefs(flattened, acc, rowsInCol, _size)
          val newCols = newAcc.size
          if (newCols > maxColumns && cols > 1) {
            (new Slice {
              val size = rows
              val columns = acc
            }, next)
          } else {
            inner(next.tail, newRowsInCol, newRows, newCols, newAcc, _size)
          }
        }
```

foundation/src/main/scala/quasar/contrib/fs2/convert.scala
```diff
@@ -23,7 +23,7 @@ import java.util.Iterator
 import java.util.stream.{Stream => JStream}
 import scala.util.{Either, Left, Right}
 
-import cats.effect.{Concurrent, Sync}
+import cats.effect.{ConcurrentEffect, Sync}
```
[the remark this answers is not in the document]
edmundnoble: For `runAsync` in `chunkQ`, it seems.

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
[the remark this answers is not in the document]
edmundnoble: The aliases are causing compiler errors because dealiasing is somehow interacting with implicit search, I think.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
edmundnoble: Ooh, missing data, with the exception fixed by incorrect array sizing. This is starting to become a pattern.

edmundnoble: So really this function will only be called for powers of two, for which this is accurate. However, in case that isn't the case, I've made it round up when necessary.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I'll definitely write these.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I actually was under the impression we didn't want to do this right now, from what we went through in the meeting room. However, that's not incredibly difficult to implement.

@alissapajer I believe paying the cost in this function is significantly better, because otherwise we would be allocating the slices twice.

foundation/src/main/scala/quasar/contrib/fs2/convert.scala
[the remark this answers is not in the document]
wemrysi: I think we’re using `Concurrent#start` in `master`, that’d obviate the need for `ConcurrentEffect`.

djspiewak:
Test failures look legit

edmundnoble:
Just to repeat what I said in a tiny, probably now hidden comment to Alissa: please don't merge this without unit tests, which I am working on today.

yggdrasil/src/main/scala/quasar/yggdrasil/TableModule.scala
edmundnoble: I meant specifically adding them to TableSize, I've tried now to make that more explicit.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
edmundnoble: Fixed, now does nothing in that case.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
edmundnoble: All fixed now, even another edge-case I discovered during testing.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I've now marked this `@tailrec`, but just so you know @alissapajer the compiler would optimize the tail-recursion even without the annotation.

foundation/src/main/scala/quasar/contrib/fs2/convert.scala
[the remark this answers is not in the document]
edmundnoble: Fixed.

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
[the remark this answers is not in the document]
edmundnoble: I also don't know where to put them anymore; they're used in yggdrasil now, not just mimir, and there's already an yggdrasil package object but it's just in the test project (and you can't have two). I could put it in foundation but it's not used anywhere else.

edmundnoble:
Unit tests are in; let me know if I missed any, but otherwise I believe I've addressed all of the feedback, though @wemrysi if you have any opinions on the finalizer type alias stuff I'd like to push those into this PR, because it would only reduce the number of lines changed.

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -490,6 +490,21 @@ public boolean equals(Object obj) {
         return true;
     }
 
+    public String toString() {
```
alissapajer: Why was this added?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -69,6 +107,28 @@ class ArrayBoolColumn(val defined: BitSet, val values: BitSet) extends ArrayColu
...
+      val newValues = ArrayColumn.resizeBitSet(values, size)
+      val newDefined = ArrayColumn.resizeBitSet(defined, size)
+      new ArrayBoolColumn(newDefined, newValues)
+    }
+  }
+
+  override def equals(other: Any): Boolean = {
+    if (other.isInstanceOf[ArrayBoolColumn]) {
+      val otherC = other.asInstanceOf[ArrayBoolColumn]
+      // this is incorrect, we should filter by definedness first.
```
alissapajer: Can we make this do the right thing in this PR? Someone might use this function expecting it to do the right thing without reading this comment.

Also, I was curious, were the `equals` methods you implemented here used just for testing, or are they used elsewhere too?

alissapajer:
This failed to compile. Also, merge master to get the temporary fix (I hope) for the SOE!

blueeyes/src/main/java/quasar/precog/BitSet.java
edmundnoble: Testing.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
edmundnoble: Fixed this in the subsequent commit. They're just for testing.

alissapajer:
The travis build was finding a stale cache so I deleted the cache for this PR build and restarted the build :crossed_fingers:

alissapajer:
@edmundnoble some tests are fixed now in the `lwc_local` travis build.

alissapajer:
@rintcius @wemrysi since you've both requested changes, I won't merge this until each of you approve.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -49,7 +49,9 @@ import shims.monoidToCats
 import java.io.File
 import java.util.UUID
 
-import scala.util.Either
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Promise
+import scala.util.{Failure, Success}
```
alissapajer: Are these imports used?

wemrysi:
@edmundnoble Regarding the `MonadFinalizer` alias, all I have to recommend is the `yggdrasil` package object, which would mean refactoring the tests to put their constructs somewhere else.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
alissapajer: Could you instead say something like:

"If we add an `RValue` to the current `Slice` without checking if its addition will exceed the column limit, it's possible that we will indeed exceed the column limit. This test case hit that bug, so we include it here."

The comment as it is doesn't say what the bug was, which is why I think it was a little unclear.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,284 @@ object Slice {
...
+    else
+      highestOneBit << 1
+  }
+
+  def fromRValuesStep(
+    values: ArraySliced[RValue], maxRows: Int, maxColumns: Int, startingSize: Int
+  ): (Slice, ArraySliced[RValue]) = {
+    @tailrec
+    def inner(
+      next: ArraySliced[RValue], rows: Int, cols: Int,
```
alissapajer: It doesn't look like `cols` is used anywhere.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,284 @@ object Slice {
...
+      acc: Map[ColumnRef, ArrayColumn[_]], _size: Int
```
alissapajer: What do you think of renaming `_size` to `allocatedColSize`? There are a few different "sizes" here so I think a more specific name would help readability.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,284 @@ object Slice {
...
+          inner(next, rows, cols, acc.mapValues { _.resize(newSize) }, newSize)
+        } else {
+          // we *may* have enough space in this slice for the
+          // next `RValue`. we're going to flatten the `RValue`
+          // out to find out how many columns we would have
+          // if we added that `RValue` to the current slice.
+          val flattened = next.head.flattenWithPath
+          val newAcc = updateRefs(flattened, acc, rows, _size)
+          val newCols = newAcc.size
+          val newRows = rows + 1
```
alissapajer: This is only used in the `else` clause on line 2001 so it can be within that context.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,284 @@ object Slice {
...
+        } else if (rows >= _size) {
+          // we'd have more rows than `_size` if we added this
+          // row to the slice. `_size` is the size in rows of
+          // all of the columns we've accumulated, so we'll
+          // have to resize those columns to add further data.
+          // `_size` should always be a power of two, so we multiply
+          // by two.
+          // println("we're resizing the columns because we have too many rows")
+          val newSize = _size * 2
+          inner(next, rows, cols, acc.mapValues { _.resize(newSize) }, newSize)
```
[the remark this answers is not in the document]
edmundnoble: This is very valuable to know, thank you.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: :+1:

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,284 @@ object Slice {
...
+      // we can make a guess at a nice starting array size;
+      // if we know that `n` values at the start of the array are
+      // scalars, we know we're going to have at least `min(n, maxRows)`
+      // scalars in this slice.
+      // otherwise, we go with the constant limit we've been passed, because
+      // we have some vector data coming up, so who knows how many rows we'll
+      // get before we have too many columns.
+      var ctr = values.start
+      var r = values.head
+      while (ctr < values.arr.length - 1 && ctr < (values.start + maxRows) && RValue.toCValue(r).nonEmpty) {
```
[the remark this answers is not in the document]
edmundnoble: Good point on always receiving scalars, but do we know for sure this will always be the case? Especially for LWC's.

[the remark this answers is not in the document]
alissapajer: I view this as an optimization for the specific case of non-nested data. It won't always be the case that our customers send nested data, but in my understanding of the product, it will be rare for customers to send raw values not nested in any structure since we're a solution for "heterogeneous messy data". And so we'd be optimizing for the scalar case at the cost of the non-scalar case when really we should be doing the opposite. 

@djspiewak if my assumptions above are incorrect please advise. :)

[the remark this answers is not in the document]
edmundnoble: I don't actually believe this is a significant cost to pay, is the thing. Type checks like this are fairly cheap, and they allow us to do things orders of magnitude faster when it's possible. But I have benchmarks now that can provide real evidence, and I'd like to revisit this with that benchmark PR.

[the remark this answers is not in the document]
alissapajer: This is an optimization for a corner case in customer data. Sometimes optimizations like this are worth it. But for now it's extra logic that we need to maintain, reason about, debug, and test. And so unless we know it's going to be giving us real benefit, I think it should be omitted.

I'm also not convinced that we would see a performance benefit in the case of data that is entirely comprised of scalars, since we'll be looping through the `while` loop for every single value. It definitely could be faster, but we shouldn't be adding anything like this without benchmarks.

Could we omit it in this PR and then benchmark it? Then we can have a discussion with the benchmarks in hand. I think it's going to be hard to discuss this further without benchmark data.

alissapajer:
@rintcius any other review to add?

it/src/main/resources/tests/orderByIfUndefined.test
```diff
@@ -2,7 +2,7 @@
     "name": "order by if undefined",
     "backends": {},
     "data": "smallZips.data",
-    "query": "select city ?? \"foo\" as name from smallZips order by city ?? \"foo\"",
+    "query": "select city ?? \"aaa\" as name from smallZips order by city ?? \"aaa\"",
```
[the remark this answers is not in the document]
edmundnoble: Well not exactly need; however, in the case that any `city` actually *was* `undefined`, it would not show up in the initial results (and the predicate is only on initial results) because `foo` is alphabetically so far down.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -69,6 +100,27 @@ class ArrayBoolColumn(val defined: BitSet, val values: BitSet) extends ArrayColu
...
+      val otherC = other.asInstanceOf[ArrayBoolColumn]
+      val newValues = values.copy()
+      val newOtherValues = otherC.values.copy()
+      newValues.and(defined)
+      newOtherValues.and(otherC.defined)
+      newValues == newOtherValues
+    } else {
+      false
+    }
+  }
```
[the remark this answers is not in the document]
edmundnoble: This is only for testing.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -69,6 +100,27 @@ class ArrayBoolColumn(val defined: BitSet, val values: BitSet) extends ArrayColu
...
+
+  def clear(row: Int): Unit = defined.clear(row)
+
+  def resize(size: Int): ArrayColumn[Boolean] = {
+    val newValues = ArrayColumn.resizeBitSet(values, size)
+    val newDefined = ArrayColumn.resizeBitSet(defined, size)
+    new ArrayBoolColumn(newDefined, newValues)
+  }
+
+  override def equals(other: Any): Boolean = {
```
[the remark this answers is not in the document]
edmundnoble: Similarly, this is only for testing.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -93,6 +145,24 @@ class ArrayLongColumn(val defined: BitSet, val values: Array[Long]) extends Arra
...
+    val newValues = ArrayColumn.resizeArray(values, size)
+    val newDefined = ArrayColumn.resizeBitSet(defined, size)
+    new ArrayLongColumn(newDefined, newValues)
+  }
+
+  override def equals(other: Any): Boolean = {
+    if (other.isInstanceOf[ArrayLongColumn]) {
+      val otherC = other.asInstanceOf[ArrayLongColumn]
+      ArrayColumn.filterDefined(defined, values).deep ==
+        ArrayColumn.filterDefined(otherC.defined, otherC.values).deep
```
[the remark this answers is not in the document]
edmundnoble: Ditto. I don't think there's any reason to use this in production; if we did, there are many other optimizations we could perform.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -73,6 +73,14 @@ trait Slice { source =>
     defined
   }
 
+  override def equals(other: Any): Boolean =
+    if (other.isInstanceOf[Slice]) {
+      val otherSlice = other.asInstanceOf[Slice]
+      size == otherSlice.size && columns == otherSlice.columns
+    } else {
+      false
+    }
```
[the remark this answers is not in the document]
edmundnoble: Ditto.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,283 @@ object Slice {
...
 
-  def updateRefs(rv: RValue, into: Map[ColumnRef, ArrayColumn[_]], sliceIndex: Int, sliceSize: Int): Map[ColumnRef, ArrayColumn[_]] = {
-    rv.flattenWithPath.foldLeft(into) {
-      case (acc, (cpath, CUndefined)) => acc
-      case (acc, (cpath, cvalue)) =>
-        val ref = ColumnRef(cpath, (cvalue.cType))
+  def updateRefs(rv: List[(CPath, CValue)], into: Map[ColumnRef, ArrayColumn[_]], sliceIndex: Int, sliceSize: Int): Map[ColumnRef, ArrayColumn[_]] = {
+    var acc = into
+    var cursor = rv
+    while (cursor.isInstanceOf[::[(CPath, CValue)]]) {
```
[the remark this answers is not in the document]
edmundnoble: It's faster, I've benchmarked this for scalaz previously.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1784,126 +1792,283 @@ object Slice {
...
+    def head: A = arr(start)
+    def tail: ArraySliced[A] = ArraySliced(arr, start + 1, size - 1)
+    // have to override this because of `Array.equals`
+    override def equals(other: Any): Boolean =
+      if (other.isInstanceOf[ArraySliced[A]]) {
+        val otherA = other.asInstanceOf[ArraySliced[A]]
+        arr.deep == otherA.arr.deep && start == otherA.start && size == otherA.size
+      } else {
+        false
+      }
```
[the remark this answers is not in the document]
edmundnoble: Ditto.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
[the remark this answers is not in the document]
djspiewak: Still worth updating `hashCode`. Or perhaps defining a method *other* than `equals`, since it's just for testing.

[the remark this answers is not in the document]
alissapajer: If this is only for testing we can name it `testEquals`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
edmundnoble: I actually just checked with benchmarks, and this doesn't make anything faster. I removed it in a follow-up PR.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
edmundnoble: Will do.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
rintcius: ok, can you add a comment for that?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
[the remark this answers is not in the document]
rintcius: At the moment not, but we don't know for the future and if you don't know the history of the codebase then things like this gets missed easily.

Anyway, I'm ok with renaming `equals` to something like `testEquals` (and avoid any changes to `hashCode` and the impl of `equals`)

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
edmundnoble: Oh compared to checking `Nil`, not pattern matching. They should be the same.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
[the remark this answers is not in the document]
rintcius: I think it is a bit better if we move `equals` under the test source root instead, but then it will be a bit more work. I'm also fine with just renaming.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
djspiewak: It'll be slightly faster than checking `eq Nil` since it won't have to do a field lookup. It's negligible though. Naturally it's faster than pattern matching. You basically shouldn't use anything other than `@switchcase` pattern matching in the hot path.

djspiewak:
@edmundnoble Branch is now conflicted.

edmundnoble:
I *believe* this is ready for merge after Appveyor, if @alissapajer and @rintcius have no more feedback.

I removed the premature optimization for slice size, and moved the `equals` stuff into `SliceSpec`. Also fixed unused imports.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
alissapajer: If this is just used for testing we should move it into the test directory.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
alissapajer: Same, is this just used for testing?

rintcius:
I have approved. There's just a couple of `equals` that are still in prod code as @alissapajer already pointed out. I'll leave that to her.

edmundnoble:
Good point @rintcius, moving that now.

PR #3648, opened 2018-06-26
Apply MinimizeAutoJoins top-down

jsantos17:


mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
```diff
@@ -175,7 +175,7 @@ class PlannerSpec extends
         beRight.which { cwf =>
           rootPushes(cwf.op) must_== Nil
         }
-    }
+    }.pendingUntilFixed
```
jsantos17: This test is now pending, but it was passing by coincidence before. This attempted to test that source minimization didn't push the entire dataset into memory when coalescing reductions with flat sources (akin to the `a[*] + b` case in `CollapseShifts`, but with reductions instead of shifts). 

The old Mongo connector performed this optimization, but QSU does not implement this. [We have an open ticket about this in YouTrack](https://slamdata.myjetbrains.com/youtrack/issue/qz-3332). All in all, this test is trying to test something that is currently unimplemented so it's rightly pending. This should be tested at the QSU level too, not here.

@djspiewak should I move qz-3332 ticket to CH?

it/src/main/resources/tests/subqueryGroupBy.test
```diff
@@ -1,9 +1,10 @@
 {
     "name": "use subquery when there's a group by statement",
     "backends": {
-        "mimir":             "ignoreFieldOrder",
-        "mongodb_3_4_13":    "pending",
-        "mongodb_3_6":       "pending"
+        "mimir":          "pending",
+        "lwc_local":      "pending",
```
alissapajer: `mimir` and `lwc_local` should be `pendingIgnoreFieldOrder` since SlamEngine doesn't guarantee field order.

But with that said, I'd like to figure out what's going on here before we merge this. Then we can decide if we want to fix it in this PR or separately. I'll take a look to see if I can figure anything out.

jsantos17: Changed it to `pendingIgnoreFieldOrder`. I'd definitely appreciate another set of eyes on this. The problem might be provenancy in nature and my lack of understanding of provenance makes it hard to debug.

mongodb/src/test/scala/quasar/physical/mongodb/PlannerSpec.scala
djspiewak: @jsantos17 We can if you feel it's important, but remember that the strategy outlined in that issue is *terrible* for any database evaluator which isn't backed by map-reduce. The way that things are right now (the cartesian with the results of the reduce) is pretty close to optimal for mimir.

djspiewak:
It's not really clear to me from reading the code… Does this fix the duplicated left shifts problem, as I suspected it would? If so, can we add a test?

jsantos17:
@djspiewak unfortunately no. I think the problem is that there is duplication in the graph before `MinimizeAutoJoins`:

<details>
<summary>Duplicate vertices in graph</summary>

```
ReifyBuckets: AuthenticatedQSU {
QSUGraph('rlp15)[
  'rlp15 -> AutoJoin2('rlp9, 'rlp13, ConcatMaps(LeftSide, RightSide))
  'rlp9  -> AutoJoin2('rlp5, 'rlp8, MakeMap(LeftSide, RightSide))
  'rlp8  -> LeftShift('rlp7, SrcHole, ExcludeId, Omit, RightTarget, FlattenArray)
  'rlp13 -> AutoJoin2('rlp10, 'rlp12, MakeMap(LeftSide, RightSide))
  'rlp10 -> Map('rlp2, Constant(Str(a0)))
  'rlp12 -> LeftShift('rlp11, SrcHole, ExcludeId, Omit, RightTarget, FlattenArray)
  'rlp11 -> Map('rlp6, TypeCheck(SrcHole:FlexArr(0, None, Top)))
  'rlp7  -> Map('rlp6, TypeCheck(SrcHole:FlexArr(0, None, Top)))
  'rlp6  -> AutoJoin2('rlp4, 'rlp5, ProjectKey(LeftSide, RightSide))
  'rlp4  -> Map('rlp1, TypeCheck(SrcHole:Obj(Map[], Some(Top))))
  'rlp1  -> LeftShift('rlp0, SrcHole, ExcludeId, Omit, RightTarget, ShiftMap)
  'rlp0  -> Read(/foo)
  'rlp5  -> Map('rlp2, Constant(Str(a)))
  'rlp2 -> ⊥
]
```

(`rlp11` and `rlp7` are the duplicate vertices)

Query is "select a[*], a[*] from foo"
</details>

It should be fairly easy to fix now that all candidate `LeftShift`s are part of the candidates passed to `CollapseShifts` in a single go. But I don't know if that's a workaround and fixing the duplication in the graph is the proper solution.

Said that, what this does fix is the need for manual join re-association. In master this query `select a, b, c, d, e, f, g from foo group by a, g` has six joins in it and manually re-ordering the projections reduces the number of joins. In this branch, there are always two joins, irrespective of the order of the projections.

djspiewak:
Solving join reassociation is very impressive. I'm curious as to exactly what logic decides the exact association. I would assume that between multiple candidate associations, each with the same number of joins, the choice right now is simply arbitrary?

Also yeah, the duplicated nodes there is really annoying. And tricky. I'm kind of surprised that duplication made it through the collapse process. Those nodes should have originally been `Unary`, which compares correctly for reference equality, meaning that they should have collapsed in `ReadLP`.

djspiewak:
Oh, can we add a test for the new join reassociation behavior? Just a quick one in `MinimizeAutoJoinSpecs`

it/src/main/resources/tests/subqueryGroupBy.test
```diff
@@ -1,9 +1,9 @@
 {
     "name": "use subquery when there's a group by statement",
     "backends": {
-        "mimir":             "ignoreFieldOrder",
```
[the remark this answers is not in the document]
alissapajer: I've started looking into this. The qscript is definitely more complex than it should be. There are 6 theta joins! I haven't looked at the qsu graphs yet. I'm currently generating the qscript against `master` to compare.

Here it is against this branch: https://gist.github.com/alissapajer/4e8e31c4cf8e235cef55c1076152a711

djspiewak:
@jsantos17 Looking into this a bit more… It looks like `ReadLP` has an issue where it just *trusts* that any nodes involved in `lp.Let` are already deduplicated. That isn't the case this particular query.

djspiewak:
[ch251]

alissapajer:
@jsantos17 can you create an issue for the integration test (`subqueryGroupBy.test `) that we had to disable in this PR? We won't need it fixed by end of July since 0.1.0 doesn't have grouping, but I don't want to lose track of it.

jsantos17:
This is on my radar. Will address review comments once I'm done with the S3 `DataSource`

PR #3649, opened 2018-06-26
Implementation of DataSourceControl and DataSourceErrors

wemrysi:
Implements the rest of the dependencies of `DefaultDataSources`. Also adds a `ConnectionFailed` case to `InitializationError` after the SlamFest conversation where we decided that datasources will eagerly connect in certain cases.


alissapajer:
I see there's a conflict with `MockDataSourcesSpec`. Did you change anything meaningful in it?

At this point I'm pretty much sure that `MockDataSourcesSpec` is the cause of the SOE in travis [ch270] since we haven't seen the error since disabling the test in master. I don't know exactly what's wrong yet though.

There's always the chance that it was fixed as a side-effect of some other change. So, one option is that we keep it disabled in this PR, and then we open a separate PR based off this one that enables the test. Then we run the unit test stage of that PR repeatedly seeing if we can still trigger the SOE. I'd be happy to do all that. I just want to make sure it's still reproducible after every change that could affect it, so that I'm not chasing down a bug that doesn't exist anymore (since I can't reproduce it locally).

wemrysi:
@alissapajer The main change in `MockDataSourcesSpec` was to move it to the `api` project as it was in `impl` for some reason. The rest were just cosmetic.

I'm fine with leaving it disabled in this PR and taking a more methodical approach to reenabling it :+1:

api/src/test/scala/quasar/api/MockDataSourcesSpec.scala
```diff
@@ -14,39 +14,38 @@
...
-final class MockDataSourcesSpec /*extends DataSourcesSpec[DefaultM, String]*/ {
-
-  def datasources: DataSources[DefaultM, String] = quasar.api.MockDataSources[DefaultM, String](acceptedSet, errorCondition)
-  def supportedType = DataSourceType("s3", 3L)
-  def validConfigs = ("one", "five")
-  def run: ~>[DefaultM, Id.Id] = λ[DefaultM ~> Id]( _.eval(IMap.empty).unsafeRunSync )
+import MockDataSourcesSpec.DefaultM
 
+// FIXME: Disabled due to being a suspect in the case of
+//        https://app.clubhouse.io/data/story/270/soe-on-travis-trace-through-specs2
```
alissapajer: :+1: good idea with the comment :)

api/src/test/scala/quasar/api/MockDataSourcesSpec.scala
```diff
@@ -14,39 +14,38 @@
...
   val mongo: DataSourceType = DataSourceType("mongodb", Positive(3).get)
   val acceptedSet: ISet[DataSourceType] = ISet.fromList(List(s3, azure, mongo))
-  def errorCondition[C](
-      rn: ResourceName,
-      dst: DataSourceType,
-      config: C
-      ): Condition[InitializationError[C]] = Condition.normal()
 
+  def datasources: DataSources[DefaultM, String] =
+    quasar.api.MockDataSources[DefaultM, String](acceptedSet, (_, _, _) => Condition.normal())
```
nicflores: this way of passing the error function is definitely much nicer!

impl/src/main/scala/quasar/impl/datasources/DataSourceManagement.scala
```diff
@@ -0,0 +1,193 @@
...
+      F[_]: Monad: MonadError_[?[_], Throwable], G[_], Q, R](
+      report: Condition[Exception] => F[Unit],
+      underlying: DataSource[F, G, Q, R])
+      extends DataSource[F, G, Q, R] {
+
+    def kind = underlying.kind
+    def shutdown = underlying.shutdown
+    def evaluate(query: Q) = reportCondition(underlying.evaluate(query))
+    def children(path: ResourcePath) = reportCondition(underlying.children(path))
+    def descendants(path: ResourcePath) = reportCondition(underlying.descendants(path))
```
nicflores: @wemrysi, @garyb confirmed that we didn't need descendants for now, but maybe this is good to still keep around?

wemrysi: I'd say let's wait and see once we've got the repl and the http api running if we actually use it, if not we can drop it so that datasource authors don't have to implement it.

PR #3650, opened 2018-06-26
Disable MockDataSourcesSpec as it might be the cause of a SOE

alissapajer:
This is my current best guess for which test is causing the SOE [ch270]. 

I'd like to restart the unit test build a few times in this PR to see if the SOE happens. If it doesn't happen after a few restarts, then let's merge this and I'll dig in to figure out what the problem is. If it does happen, I'll close this PR.

I've been unable to reproduce it locally so we might have to iterate with Travis a bit to track this down.

### DIFF
--- a/impl/src/test/scala/quasar/impl/datasources/MockDataSourcesSpec.scala
+++ b/impl/src/test/scala/quasar/impl/datasources/MockDataSourcesSpec.scala
@@ -26,11 +26,11 @@ import quasar.impl.datasources.MockDataSourcesSpec.DefaultM
 
 import eu.timepit.refined.auto._
 import scalaz.{IMap, ISet, Id, StateT, ~>}, Id.Id
-import scalaz.std.string._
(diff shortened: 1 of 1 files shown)


edmundnoble:
Ah, I ran into this locally. Not seeing it anymore, though; seemed to be a fluke.

alissapajer:
@edmundnoble do you by chance know what test or project you were testing when you saw the SOE?

alissapajer:
Unit test stage passed once.

alissapajer:
Unit test stage passed twice.

alissapajer:
Unit test stage passed thrice.

alissapajer:
@wemrysi three times is good enough for me. What do you think of merging this? If we see another SOE I'll revert immediately. I don't know what about this test is causing the problem yet, but since the SOE is blocking all builds I'd like to get this in in the interim.

PR #3652, opened 2018-06-27
Fix publishing

alissapajer:


scripts/testJar
```diff
@@ -15,16 +15,17 @@ then
     cp $(dirname $0)/../example-quasar-config.json $CONFIG_FILE
 fi
 
-QUASAR_SUCCESS_OUTPUT="Server started listening on port"
+QUASAR_VERSION=$(cat $WS_DIR/version.sbt | sed -n -e 's/^ *version.*:= "\(.*\)"[^\n]*/\1/p')
+QUASAR_SUCCESS_OUTPUT="(v$QUASAR_VERSION) 💪 $"
```
alissapajer: This is hacky, but not more hacky than what we had before... I ran this script locally and it works.

repl/src/main/scala/quasar/repl/CliOptions.scala
```diff
@@ -46,7 +46,7 @@ object CliOptions {
       c.copy(config = Some(x))
     } text("path to the config file to use")
 
-    opt[(String, Seq[File])]("backend").required.unbounded validate { x =>
+    opt[(String, Seq[File])]("backend").unbounded validate { x =>
```
alissapajer: It's valid to run the repl without any plugin backends enabled. Not sure why this was `required` previously.

PR #3655, opened 2018-06-27
Add sbt-project-graph, lexicographically order projects

edmundnoble:
I saw that I would have to edit the "dependency tree" in the comments of build.sbt when I added a benchmarks project for mimir, which seemed an incredibly daunting task. Then I thought, how many projects have we not added because of that feeling of dread? Then I thought, how much of this tree is even accurate? Why not ask SBT to do it? So I've added a plugin which does this exactly, and purged our "tree comment".


project/plugins.sbt
```diff
@@ -1,9 +1,12 @@
...
-addSbtPlugin("com.eed3si9n"          % "sbt-assembly"  % "0.14.5")
-addSbtPlugin("com.eed3si9n"          % "sbt-buildinfo" % "0.7.0")
-addSbtPlugin("io.get-coursier"       % "sbt-coursier"  % "1.1.0-M4")
-addSbtPlugin("com.slamdata"          % "sbt-slamdata"  % "1.2.0")
-addSbtPlugin("pl.project13.scala"    % "sbt-jmh"       % "0.2.27")
-addSbtPlugin("com.github.romanowski" % "hoarder"       % "1.0.2-RC2")
+addSbtPlugin("com.eed3si9n"          % "sbt-assembly"      % "0.14.5")
+addSbtPlugin("com.eed3si9n"          % "sbt-buildinfo"     % "0.7.0")
+addSbtPlugin("io.get-coursier"       % "sbt-coursier"      % "1.1.0-M4")
+addSbtPlugin("io.get-coursier"       % "sbt-coursier"      % "1.1.0-M4")
```
alissapajer: This is doubled up.

alissapajer:
Since the plugin isn't a requirement of the build, I'd prefer to not include it here. Anyone who wants it can include it in their global plugins.

project/plugins.sbt
edmundnoble: Fixed! :)

edmundnoble:
@alissapajer Donezo, just removed the tree from the comments.

alissapajer:
So, I've really enjoyed the 5 minutes I'd spend every few weeks editing the ASCII graph, and it was nearly correct. I was actually just looking at it today thinking about how I could fix the last bit that wasn't correct :)

But, I agree it requires maintenance which is entirely unnecessary. I just drew the graph with the plugin and it's actually pretty nice. So I'm :+1: to merge :)

edmundnoble:
Alright awesome. 

I could totally see how it would be fun to do, heh.

edmundnoble:
Here's a copy of what I got when generating it, for the curious:

image

alissapajer:
Is there any way to omit `root` (and thereby omit all the dashed lines?). I didn't see one in the docs.

PR #3658, opened 2018-06-28
[WIP] REPL based on quasar.api and jline3

rintcius:
TODO's:

- incorporate #3669 
- investigate why data source configs are not properly saved on shutdown. During a repl session adding/removing/editing data sources works fine, but when a new repl session is started the changes are mostly not remembered (after the 1st restart *some* seems to be remembered, but after subsequent ones nothing)
- CLI parsing and sbt integration: at the moment repl is started with fixed arguments (no external plugins, always the same mimir dir)
- properly print errors. We can do this in a separate PR. I propose to add a `def message: String` to new errors like we did for other errors.
- log4j classloader problem. This doesn't seem to cause any problem in usage, but a huge error is logged. See https://gist.github.com/rintcius/ff282eea131499c97ef02eb190314af7 for a stripped version. Note: I haven't seen this on a first start of `repl/run` in an `sbt` session, only after exiting and then restarting, so I think this won't be an issue when starting from the launcher. 
- when showing query results currently the following restrictions apply:
  - `summaryCount` is ignored. We can cut off client side or add a `limit` to the query but these are workarounds. We should account for this properly at some time, but that depends on the priorities.
  - `debug` is ignored. Currently results are always shown as if `debug = 0`. `phaseFormat` and `timingFormat` are also ignored
- `..` is not interpreted as parent yet. So `cd ..` will try to go down a prefix called `..` instead of up

Nice to have:
- command completion
- show current dir in prompt


repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+import cats.effect.concurrent.Ref
+import scalaz._, Scalaz._
+import shims._
+
+import cats.syntax.applicative._
+import cats.data.StateT
+import quasar.contrib.scalaz.MonadState_
+
+object Main extends IOApp {
+  // TODO clean up once we can work with real datasources instead of MockDataSources
```
rintcius: based on similar code in `slamdata-backend`

repl/src/main/scala/quasar/repl2/Evaluator.scala
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private def ensureNormal[E: Show](c: Condition[E]): F[Unit] =
+      c match {
+        case Condition.Normal() => F.unit
+        case Condition.Abnormal(err) => raiseEvalError(err.shows)
+      }
+
+    private def findType(tp: DataSourceType.Name): F[Option[DataSourceType]] =
+      stateRef.get.map(_.supportedTypes).map(_.toList.find(_.name == tp))
```
rintcius: Couldn't get `===` working here (it's a refined type)

repl/src/main/scala/quasar/repl2/Evaluator.scala
```diff
@@ -0,0 +1,195 @@
...
+          _   <- stateRef.update(_.copy(supportedTypes = tps))
+          s   =  tps.toList.map(tp => s"${tp.name} (${tp.version})")
+                   .mkString("Supported datasource types:\n", "\n", "").some
+        } yield s
+
+      case DatasourceAdd(name, tp, cfg, onConflict) =>
+        for {
+          tps <- sources.supported
+          dsType <- findTypeF(tp)
+          c <- sources.add(name, dsType, cfg.asInstanceOf[C], onConflict)
```
rintcius: We'd need some `String => C` conversion here, as opposed to the `asInstanceOf`. What's the vision here?

repl/src/main/scala/quasar/repl2/ReplState.scala
```diff
@@ -0,0 +1,81 @@
...
+
+object ReplState {
+
+  type DatasourceStore = IMap[ResourceName, DataSourceMetadata]
+
+  def mk: ReplState = ReplState(
+    rootDir,
+    DebugLevel.Normal,
+    PhaseFormat.Tree,
+    refineMV[Positive](10).some,
```
wemrysi: Should be able to replace this with `10L`.

repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+       IMap[ResourceName, (DataSourceMetadata, String)]((ResourceName("tests3"), (DataSourceMetadata(s3, Condition.Normal()), "test-s3-config")))
+
+     (dataSources, initialStore)
+  }
+
+  def run(args: List[String]): IO[ExitCode] = {
+    val datasources = mock._1
+    val l = Ref.of[DataSourcesMonadT, ReplState](ReplState.mk) >>=
+      (ref => Repl.mk[DataSourcesMonadT, String](datasources, ref).loop)
+    l.runA(mock._2)
```
wemrysi: This is why you're not seeing changes persist. You need to store the next state in a ref and use it as the input when evaluating the next command, etc. Similar to https://github.com/slamdata/slamdata-backend/pull/377/files#diff-d6289c36ccbdcb2f82b79e4a70e28614R51

rintcius:
@wemrysi maybe we can first discuss what needs to be included in this PR? I'll at least incorporate #3669 

BTW note that as of this PR it won't be possible to start the old repl anymore.

alissapajer:
If it wont' be possible to start the old repl anymore, should we delete it in this PR?

djspiewak:
Closing since #3683 opened

PR #3660, opened 2018-06-28
Yggdrasil benchmarks, phase 1 (fromRValues)

edmundnoble:
This is based on #3644, because it benchmarks the streaming `Slice.fromRValues` against the new `allFromRValues`. It's probably best not to review this as a diff, instead just look at the `yggdrasilPerf` folder and the new `build.sbt` stuff (as well as the sbt-jmh update).

I used these benchmarks along with `AsyncProfiler` (https://github.com/ktoso/sbt-jmh#using-async-profiler) to find and fix performance and correctness issues with #3644, so they're already paying off.

Here's what I use to run it (you can tack the profiler options on the end):

```
yggdrasilPerf/jmh:run -f1 -t2 -i 10 -wi 10 -jvmArgsAppend -Xmx4g -jvmArgsAppend -Xms4g
```

Here's some results from my machine:

```
Benchmark                                                 (streaming)  Mode  Cnt    Score    Error  Units
FromRValuesBenchmark.ingestArrsWithFittingColumns                true  avgt   10   73.694 ±  4.580  ms/op
FromRValuesBenchmark.ingestArrsWithFittingColumns               false  avgt   10  102.447 ±  5.470  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns            true  avgt   10  560.016 ± 32.047  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns           false  avgt   10  184.641 ± 10.767  ms/op
FromRValuesBenchmark.ingestLongs                                 true  avgt   10   71.825 ±  3.044  ms/op
FromRValuesBenchmark.ingestLongs                                false  avgt   10   93.332 ±  3.883  ms/op
FromRValuesBenchmark.ingestObjectsWithDistinctColumns            true  avgt   10  364.844 ± 16.069  ms/op
FromRValuesBenchmark.ingestObjectsWithDistinctColumns           false  avgt   10  637.279 ± 17.848  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns             true  avgt   10  130.653 ±  4.718  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns            false  avgt   10  116.869 ±  1.142  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns         true  avgt   10  708.024 ± 20.109  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns        false  avgt   10  332.660 ±  3.064  ms/op
```


edmundnoble:
Right now I'm making them a bit more parameterized, stand by for another half hour or so.

djspiewak:
It's interesting that the streaming results are so much faster than the non-streaming except in the overflowing column case. That would seem to hint that we're a bit more allocation-bound than expected.

edmundnoble:
I don't think I understand what you mean, @djspiewak. `streaming` here means "new version", in case that isn't clear (it isn't that clear now that I think about it, I'll change the name).

djspiewak:
@edmundnoble Oh you mean you're benchmarking the old vs the new? Is that even fair though, benchmarking the old one on overflowing columns? It just didn't handle that case at all.

edmundnoble:
Right, overflowing columns isn't exactly fair, but I hoped it might show in some cases that the extra columns actually result in huge slices which take *more* time to allocate without snipping the slice.

edmundnoble:
I've also just realized that there's an issue with this benchmark; to be really comparative I have to use the `Table` methods because `Slice.fromRValues` doesn't row-limit, the table version does.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
```diff
@@ -18,8 +18,9 @@ package quasar
 package yggdrasil
 package perf
 
-import quasar.yggdrasil.table._
+import quasar.contrib.cats.effect.liftio._
```
djspiewak: Do we need this import?

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
```diff
@@ -29,12 +30,17 @@ import java.util.concurrent.TimeUnit
 import org.openjdk.jmh.annotations.{Benchmark, BenchmarkMode, Mode, OutputTimeUnit, Param, Scope, State}
 import org.openjdk.jmh.infra.Blackhole
 
+import scalaz._, Scalaz._
```
djspiewak: Underscores! Why? In general, `shims` (and other packages which are imported for implicits) are the only imports which should be using underscores *unless* running into the `Applicative`/`Traverse` syntax ambiguity.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
```diff
@@ -61,50 +67,53 @@ class FromRValuesBenchmark {
...
 
-  def createAndConsumeSlices(data: Stream[List[RValue]], bh: Blackhole): IO[Unit] = {
-    val slices: fs2.Stream[IO, Slice] =
+  def createAndConsumeTable(data: Stream[List[RValue]], bh: Blackhole): IO[Unit] = {
+    val table: WriterT[IO, List[IO[Unit]], P.Table] =
       if (streaming) {
-        Slice.allFromRValues(fs2.Stream.fromIterator[IO, fs2.Stream[IO, RValue]](data.iterator.map(fs2.Stream.emits(_).covary)).flatMap(x => x))
+        P.Table.fromRValueStream[WriterT[IO, List[IO[Unit]], ?]](
+          fs2.Stream.fromIterator[IO, fs2.Stream[IO, RValue]](
+            data.iterator.map(fs2.Stream.emits(_).covary)
```
djspiewak: Please ascribe the annotation to `covary` here. This sort of type inference is extremely expensive.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/SliceTools.scala
```diff
@@ -20,16 +20,21 @@ package perf
...
+
 object SliceTools {
 
   def consumeSlice(slice: Slice, bh: Blackhole): IO[Unit] =
     IO(bh.consume(slice.materialized))
 
-  def consumeSlices(slices: fs2.Stream[IO, Slice], bh: Blackhole): IO[Unit] =
-    slices.compile.fold(IO.unit)((i, o) => i.flatMap(_ => consumeSlice(o, bh))).flatMap(x => x)
+  def consumeSlices(slices: StreamT[IO, Slice], bh: Blackhole): IO[Unit] =
+    slices.foldLeftRec(IO.unit)((i, o) => i.flatMap(_ => consumeSlice(o, bh))).flatMap(x => x)
```
djspiewak: I hate `StreamT`…

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
alissapajer: We have a lot of old(er) code in quasar that does exactly this, and this was the company default for a while. I do 100% agree that we should be explicitly importing from scalaz. I'm just not sure we had ever talked about this as a team :)

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
edmundnoble: Yes, in order to use `Table.fromRValuesStream`, which requires a `LiftIO` constraint on `WriterT`.

edmundnoble:
I've managed to optimize the code to deal with rows that have very large amounts of columns. It used to take three or four times as long compared to the old code; now it takes around two times as long, and I also eliminated the pathology @alissapajer pointed out. Though the code still has many places it could be improved; notably, we will end up re-traversing an `RValue` if it causes us to go over the column limit, and we don't go cross-chunk to create slices.

I also noticed that the CPU governor was substantially affecting the previous results; I've fixed that for these.

In addition, I've noticed that removing the optimization which @alissapajer was doubtful over (pre-allocating larger slices if we can see that lots of the data is scalar) actually doesn't change the benchmark results for `ingestLongs` at all, so I've taken it out.

Without the optimization:
```
Benchmark                                                 (streaming)  Mode  Cnt     Score     Error  Units
FromRValuesBenchmark.ingestArrsWithFittingColumns                true  avgt    5    98.485 ±   5.376  ms/op
FromRValuesBenchmark.ingestArrsWithFittingColumns               false  avgt    5   127.335 ±   1.751  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns            true  avgt    5   862.551 ±  32.254  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns           false  avgt    5   329.226 ±   3.285  ms/op
FromRValuesBenchmark.ingestLongs                                 true  avgt    5    97.735 ±   9.289  ms/op
FromRValuesBenchmark.ingestLongs                                false  avgt    5   215.191 ±   3.708  ms/op
FromRValuesBenchmark.ingestObjectsWithDistinctColumns            true  avgt    5   553.001 ±  70.297  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns             true  avgt    5   157.418 ±   4.215  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns            false  avgt    5   146.931 ±   1.517  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns         true  avgt    5  1240.071 ± 209.534  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns        false  avgt    5   414.461 ±   9.030  ms/op'
```

With the optimization:
```
Benchmark                                                 (streaming)  Mode  Cnt    Score     Error  Units
FromRValuesBenchmark.ingestArrsWithFittingColumns                true  avgt    5  100.108 ±   1.995  ms/op
FromRValuesBenchmark.ingestArrsWithFittingColumns               false  avgt    5  127.198 ±   3.476  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns            true  avgt    5  458.486 ±  12.713  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns           false  avgt    5  233.668 ±   1.515  ms/op
FromRValuesBenchmark.ingestLongs                                 true  avgt    5   96.670 ±   6.142  ms/op
FromRValuesBenchmark.ingestLongs                                false  avgt    5  210.891 ±   0.605  ms/op
FromRValuesBenchmark.ingestObjectsWithDistinctColumns            true  avgt    5  563.716 ±  18.554  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns             true  avgt    5  158.302 ±   6.457  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns            false  avgt    5  145.974 ±   1.488  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns         true  avgt    5  726.000 ± 157.575  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns        false  avgt    5  416.839 ±   7.474  ms/op
```

djspiewak:
@edmundnoble I think the test failures are legit.

edmundnoble:
@djspiewak Fixed.

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
```diff
@@ -44,7 +39,7 @@ import scalaz._, Scalaz._
 
 final class FederatedShiftedReadPlanner[
     T[_[_]]: BirecursiveT: EqualT: ShowT,
-    F[_]: LiftIO: Monad: PlannerErrorME: MonadFinalizers[?[_], IO]](
+    F[_]: LiftIO: Monad: PlannerErrorME: MonadTell_[?[_], List[IO[Unit]]]](
```
djspiewak: Why not use the type alias?

mimir/src/main/scala/quasar/mimir/evaluate/MimirQScriptEvaluator.scala
```diff
@@ -50,7 +51,7 @@ import shims._
 
 final class MimirQScriptEvaluator[
     T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT,
-    F[_]: LiftIO: Monad: PlannerErrorME: MonadFinalizers[?[_], IO]] private (
+    F[_]: LiftIO: Monad: PlannerErrorME: MonadTell_[?[_], List[IO[Unit]]]] private (
```
djspiewak: Ditto on type alias?

mimir/src/main/scala/quasar/mimir/evaluate/package.scala
```diff
@@ -16,18 +16,13 @@
...
 
 import scalaz.Kleisli
 
 package object evaluate {
   type Associates[T[_[_]], F[_], G[_]] = AFile => Option[Source[QueryAssociate[T, F, G]]]
   type AssociatesT[T[_[_]], F[_], G[_], A] = Kleisli[F, Associates[T, F, G], A]
-
-  type Finalizers[F[_]] = List[F[Unit]]
-  type MonadFinalizers[F[_], G[_]] = MonadTell_[F, Finalizers[G]]
-  def MonadFinalizers[F[_], G[_]](implicit ev: MonadFinalizers[F, G]): MonadFinalizers[F, G] = ev
```
djspiewak: Ok, that's why you didn't use them. Why are these being deleted?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -33,17 +33,48 @@ trait DefinedAtIndex {
...
   def update(row: Int, value: A): Unit
+  def clear(row: Int): Unit
+  def resize(size: Int): ArrayColumn[A]
 }
 
-class ArrayHomogeneousArrayColumn[@specialized(Boolean, Long, Double) A](val defined: BitSet, values: Array[Array[A]])(val tpe: CArrayType[A])
+object ArrayColumn {
+  private[table] def resizeArray[A](arr: Array[A], size: Int)(implicit A: ClassTag[A]): Array[A] = {
+    val newArr = new Array[A](size)
+    System.arraycopy(arr, 0, newArr, 0, Math.min(size, arr.length));
```
djspiewak: Random semicolon?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
```diff
@@ -33,17 +33,48 @@ trait DefinedAtIndex {
...
+    newArr
+  }
+
+  private[table] def resizeBitSet(bs: BitSet, size: Int): BitSet = {
+    // 64 bits per long, 2^6 == 64
+    val howManyLongs = ((size - 1) >> 6) + 1
+    val arr = new Array[Long](howManyLongs)
+    System.arraycopy(bs.getBits(), 0, arr, 0, Math.min(howManyLongs, bs.getBits().length))
+    new BitSet(arr, howManyLongs)
```
djspiewak: Rather than making the `BitSet` private constructor public, why not just put a `resize` function on `BitSet`?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+    def head: A = arr(start)
+    def tail: ArraySliced[A] = ArraySliced(arr, start + 1, size - 1)
+  }
+
+  object ArraySliced {
+    val noRValues = Slice.ArraySliced(new Array[RValue](0), 0, 0)
+  }
+
+  // doesn't return the next power of two
+  // when the input is already a power of two
```
djspiewak: Why not? I mean, I see why in the implementation, but you're special-casing. Wouldn't it be better to always return the next power of 2? Doesn't this semantic lead to more special-casing a call sites?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+  def nextPowerOfTwo(i: Int): Int = {
+    val highestOneBit = Integer.highestOneBit(i)
+    if (i == highestOneBit)
+      i
+    else
+      highestOneBit << 1
+  }
+
+  def fromRValuesStep(
```
djspiewak: Why is this public?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+    def tail: ArraySliced[A] = ArraySliced(arr, start + 1, size - 1)
+  }
+
+  object ArraySliced {
+    val noRValues = Slice.ArraySliced(new Array[RValue](0), 0, 0)
+  }
+
+  // doesn't return the next power of two
+  // when the input is already a power of two
+  def nextPowerOfTwo(i: Int): Int = {
```
djspiewak: This also shouldn't be public.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+      (Slice.empty, ArraySliced.noRValues)
+    } else {
+      val size = Math.min(maxRows, startingSize)
+      inner(values, 0, false, Map.empty, size)
     }
   }
 
+  def allFromRValues(
+    values: fs2.Stream[IO, RValue], maxRows: Option[Int] = None, maxColumns: Option[Int] = None
+  ): fs2.Stream[IO, Slice] = {
```
djspiewak: Style note: this is incorrectly indented. Not a big deal, but just for the record, the following is more conventional for quasar:

```scala
  def allFromRValues(
    values: fs2.Stream[IO, RValue],
    maxRows: Option[Int] = None, 
    maxColumns: Option[Int] = None)
    : fs2.Stream[IO, Slice] =
```

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+        val (nextSlice, remainingData) = Slice.fromRValuesStep(next, maxRowsC, maxColumnsC, 32)
+        fs2.Pull.output1(nextSlice) >> rec(remainingData, values)
+      }
+
+    rec(Slice.ArraySliced(new Array[RValue](0), 0, 0), values).stream
+  }
+
   def fromJValues(values: Stream[JValue]): Slice =
-    fromRValues(values.flatMap(RValue.fromJValue))
+    fromRValues(values.map(RValue.fromJValueRaw))
```
djspiewak: Why?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
 
+  // don't use this anymore. It doesn't limit the slice size properly,
+  // unlike allFromRValues and fromRValuesStep.
   def fromRValues(values: Stream[RValue]): Slice = {
```
djspiewak: Why are we keeping this around?

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -594,14 +592,12 @@ final class SerialVFS[F[_]] private (
...
-      r <- ref.get
-      a <- F.fromEither(r)
-    } yield a
-  }
+  private def serialize[A](fa: F[A]): F[A] = for {
+    ref <- Deferred[F, Either[Throwable, A]]
+    _ <- worker.enqueue1(F.attempt(fa).flatMap(ref.complete))
+    r <- ref.get
+    a <- F.fromEither(r)
+  } yield a
```
djspiewak: Why was this reformatted? I don't mind whitespace changes with other things, but you didn't touch this file at all aside from removing the import on `Either`. (which actually surprises me, since `Either` is used here, so somehow it's coming from somewhere else?

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -69,6 +71,371 @@ class SliceSpec extends Specification with ScalaCheck {
...
+    val allSliceValues: List[CValue] = sliceValues.foldLeft(List.empty[CValue])(_ ++ _.flatten)
+    allSliceValues must_===(values.flatMap(_.flattenWithPath.map(_._2)))
+    sliceValues.exists(_.isEmpty) must_===(false)
+  }
+
+  def sliceEqualityAtDefinedRows(slice1: Slice, slice2: Slice): Result = {
+    def colValues(col: ArrayColumn[_]): Iterator[CValue] = {
+      col match {
+        case ac: ArrayLongColumn           =>
+          (0 until ac.values.length).iterator.map(i => if (ac.defined(i)) CLong(ac.values(i)) else CUndefined)
```
djspiewak: No. Don't use `CUndefined`. I'm removing it. Use `Option[CValue]` instead.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -69,6 +71,371 @@ class SliceSpec extends Specification with ScalaCheck {
...
+        CLocalTime(LocalTime.parse("23:43:47.888386818")), COffsetTime(OffsetTime.parse("09:52:36.954641932+13:34:32")),
+        CLocalDateTime(LocalDateTime.parse("-5205533-11-05T23:40:19.080489826")), CFalse, CLocalDate(LocalDate.parse("+1751043-09-25")),
+        CEmptyArray, CLong(8180114872082381267L), COffsetDate(OffsetDate.parse("-3678224-01-01+15:25:19")),
+        COffsetDateTime(OffsetDateTime.parse("+28785-12-16T09:45:14.815967029+16:18:49")), CNull,
+        CInterval(DateTimeInterval.parse("P-9036285Y44M-411DT584566H4M1.912409135S").get), CDouble(-2.3703641119979175E233), CLocalDate(LocalDate.parse("+4396657-12-11")),
+        CNull, CNull, CEmptyObject, CInterval(DateTimeInterval.parse("PT550632H50M7.334061745S").get), CLocalTime(LocalTime.parse("17:00:02.808168924")),
+        CLong(-2051368316550135838L), COffsetDateTime(OffsetDateTime.parse("-9897897-05-10T00:02:09.498265366+04:48:25")), CDouble(1.0859578724304194E-147),
+        CNull, CTrue, CLong(2512754280505120381L), CLocalDateTime(LocalDateTime.parse("-7951878-05-18T11:03:35.812478165")), CEmptyArray, CEmptyObject)
```
djspiewak: 😨

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -69,6 +71,371 @@ class SliceSpec extends Specification with ScalaCheck {
...
+          {"foo":6},
+          {"foo":7, "bar7": 7, "baz": 7, "quux": 7},
+          {"foo":8, "baz": 8},
+          {"foo":9, "bar9": 9},
+          {"foo":10},
+          {"foo":11, "baz": 11},
+          {"foo":12, "baz": 12},
+          {"foo":13, "baz": 13},
+          {"foo":14, "bar14": 14}
+        ]""".stripMargin)
```
djspiewak: Remove `stripMargin`; it isn't actually doing anything here since you don't have a margin.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
```diff
@@ -0,0 +1,121 @@
...
+package yggdrasil
+package perf
+
+import quasar.contrib.cats.effect.liftio._
+import quasar.precog.common.{CValue, CLong, RArray, RObject, RValue}
+import quasar.yggdrasil.table.TestColumnarTableModule
+
+import cats.effect.IO
+
+// Must not be in default package
```
djspiewak: What does this mean?

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/SliceTools.scala
```diff
@@ -0,0 +1,40 @@
...
+import org.openjdk.jmh.infra.Blackhole
+
+import cats.effect.IO
+import scalaz.StreamT
+import shims._
+
+object SliceTools {
+
+  def consumeSlice(slice: Slice, bh: Blackhole): IO[Unit] =
+    IO(bh.consume(slice.materialized))
```
djspiewak: What does `bh.consume` do again?

yggdrasil/src/main/scala/quasar/yggdrasil/TableModule.scala
```diff
@@ -153,7 +155,8 @@ trait TableModule extends TransSpecModule {
     def constEmptyObject: Table
     def constEmptyArray: Table
 
-    def fromRValues(values: Stream[RValue], maxSliceSize: Option[Int] = None): Table
+    def fromRValues(values: Stream[RValue], maxSliceRows: Option[Int] = None): Table
+    def fromRValueStream[M[_]: Monad: MonadTell_[?[_], List[IO[Unit]]]: LiftIO](values: fs2.Stream[IO, RValue]): M[Table]
```
djspiewak: Style sidebar: I would prefer masking `scala.Stream` (e.g. renaming to something like `SStream`) and having `fs2.Stream` being the meaning of "`Stream`". Not a big deal, but something to think about for the future. `scala.Stream` just shouldn't be used at all.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -619,8 +621,20 @@ trait ColumnarTableModule
...
+      for {
+        d <- LiftIO[M].liftIO(convert.toStreamT(sliceStream))
+
+        _ <- MonadTell_[M, List[IO[Unit]]].tell(List(d.dispose))
+
+        slices = d.unsafeValue
+      } yield Table(slices, UnknownSize)
+    }
+
+    def fromRValues(values: Stream[RValue], maxSliceRows: Option[Int] = None): Table = {
```
djspiewak: This function should be deprecated along with the `Slice` version.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+
+          acc = acc.updated(ref, updatedColumn)
+      }
+      cursor = cursor.tail
+    }
+    acc
+  }
+
+  final case class ArraySliced[A](arr: Array[A], start: Int, size: Int) {
+    def head: A = arr(start)
```
djspiewak: `A` needs to be `@specialized` if you're going to do this (unless you're 100% sure that you aren't ever putting primitives in here, in which case you should probably comment that).

Sidebar, I think `ArrayProjection` or `ArrayView` would be a more appropriate name for this type. `ArraySliced` makes it sound like something related to `Slice`.

Other sidebar… make this `private` unless it's useful outside this file. If it's useful outside this file, put it in its own file.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+          } else {
+            // we *may* have enough space in this slice for the
+            // next `RValue`. we're going to flatten the `RValue`
+            // out to find out how many columns we would have
+            // if we added that `RValue` to the current slice.
+            // if we have no other rows, we'll start with a two-row
+            // slice, just in case we have too many columns immediately.
+            val flattened = next.head.flattenWithPath
+            val newAcc = updateRefs(flattened, acc, rows, allocatedColSize)
+            val newCols = newAcc.size
```
djspiewak: This is an extremely wasteful way to do this check. You can find out the distinct paths which are in the `RValue` and then (efficiently) intersect them with the current column set in the slice. This approach performs significant throwaway work whenever column limits are hit.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+              val chunkArr = chunk.toArray
+              rec(Slice.ArraySliced(chunkArr, 0, chunkArr.length), next)
+            case None                =>
+              // println("Finished top-level loop")
+              fs2.Pull.done
+          }
+        } yield ()
+      } else {
+        // println(s"extracting slice from data with size ${next.size}")
+        val (nextSlice, remainingData) = Slice.fromRValuesStep(next, maxRowsC, maxColumnsC, 32)
```
djspiewak: Extract this constant into a named value. Something like `val DefaultMinRows = 32` or ish.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1798,126 +1798,276 @@ object Slice {
...
+          // data we have already as a slice, and pass
+          // the row back to the caller.
+          // println(s"we have the maximum ($maxRows) rows, cutting slice early")
+          (new Slice {
+            val size = rows
+            val columns = acc
+          }, next)
+        } else if (rows >= allocatedColSize && !colsOverflowed) {
+          // we'd have more rows than `allocatedColSize` if we added this
+          // row to the slice. `allocatedColSize` is the size in rows of
```
djspiewak: > `allocatedColSize` is the size in rows

Shouldn't it be called something other than "allocated **col** size" then?

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
```diff
@@ -19,13 +19,14 @@ package quasar.yggdrasil
 object Config {
   val idSource = new FreshAtomicIdSource
 
-  def hashJoins         = true
-  def maxSliceSize: Int = 20000
+  def hashJoins            = true
+  def maxSliceRows: Int    = 16384
+  def maxSliceColumns: Int = 100
```
djspiewak: `maxSliceColumns` should be 128, not 100

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
```diff
@@ -19,13 +19,14 @@ package quasar.yggdrasil
...
 
   // This is a slice size that we'd like our slices to be at least as large as.
-  def minIdealSliceSize: Int = maxSliceSize / 4
+  def minIdealSliceRows: Int = maxSliceRows / 4
 
   // This is what we consider a "small" slice. This may affect points where
   // we take proactive measures to prevent problems caused by small slices.
-  def smallSliceSize: Int = 50
+  def smallSliceRows: Int = 50
```
djspiewak: I would set this to 32, since that's the magic "default min slice size" we use in `allFromRValues`.

djspiewak:
I think GitHub may have failed to pull out things that were already merged in #3644, so it's possible my review walked on some redundant paths. Either way, some things should probably be fixed.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
edmundnoble: True that.

mimir/src/main/scala/quasar/mimir/evaluate/package.scala
edmundnoble: This is from the `tablefromrvaluesstream` branch. As I said there, there's no longer anywhere to put `MonadFinalizers` because then it would be used in yggdrasil, which has a package object in its `test` folder and not its `main` folder, meaning I would have to reorganize the `test` folder and I saw it as too late in that PR's life to do so there.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ArrayColumn.scala
edmundnoble: Good idea.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: It actually leads to more special-casing to do what you said, that special-casing being to check whether `nextPowerOfTwo(x) == x` and if so to use `x` instead.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Testing.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Good point.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: For the old API, as we established on that PR.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
edmundnoble: This was an artifact of changing fs2 version earlier and changing it back. I'll revert this.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
edmundnoble: Okay.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/SliceTools.scala
edmundnoble: `bh.consume` says "JIT, I promise this object is not unused".

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: Then why is it public? `private[table]`

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
edmundnoble: Makes sense to me.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: I looked for that discussion and couldn't find it. Link?

I'd rather see this renamed if we can't delete it, but I honestly don't see why we can't delete it. All call sites can be trivially rewritten.

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
edmundnoble: Why? I had a thought earlier as well; isn't our limit really conceptually on `rows * cols`, perhaps multiplying `cols` by a constant > 1 to signify that we dislike very sparse slices?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: It's the size of the columns, though. `allocatedRowSize` would be `cols`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I'll put this in `Config.scala`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: It occurs to me that we might still be using the old function (this one) at some of the call sites.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
edmundnoble: Sure.

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
djspiewak: I suppose we're not going to be at the column max very often. And actually thinking about it, since we're using a Bagwell trie to represent the column map, we don't need the power of two, so this can stay.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I'll make it `private[table]`, though.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: The previous version silently adjusted `Num`s into `Double`s or `Long`s if they fit, randomly.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
edmundnoble: Please just be explicit. Are you saying I should've shrunk this example to make it less random? Fine, I'll do that.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I'll make it `private[table]` in its own file. 

I am not putting primitives in here. I have no way of promising that I never will, but I can't see why I would.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: @alissapajer All existing call-sites can be replaced with one of the two:

```scala
Slice.allFromRValues(Stream.emits(values))
// or
Slice.allFromRValues(Stream.fromIterator(values.iterator))
```

Depending on laziness needs.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: I think I see what you mean. Took a few re-reads. You're talking specifically about the `updateRefs` call. Instead of that though, you're saying I can do `(acc.keys ++ flattened.map(_._1).toSet).size` to find out `newCols` instead of calling `updateRefs` and building new slices. That, I can do in this PR.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: I mean, that's technically not random, it was by design. Numeric autocercion is part of the design. By making this change, you actually remove test coverage on those column types. I'm actually in the process of deprecating `fromJValueRaw` on my branch which deletes `CUndefined`.

edmundnoble:
@djspiewak All of the code except for one comment in your review is from #3644. I'd like to put up a smaller cleanup PR before this one lands to address most of your feedback, though a bit of it also touches on performance so I'll address that part in this one.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: My point wasn't that they are hard to replace. My point was that they might not have been replaced, meaning we may still be running production and test code against the old function.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
djspiewak: No, that's not what I meant. :-) It was an attempt at levity. I think it's fine to leave this here.

mimir/src/main/scala/quasar/mimir/evaluate/package.scala
edmundnoble: I'll handle this in the cleanup PR.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/FromRValuesBenchmark.scala
edmundnoble: That's... a good question. I don't remember. I believe I had this benchmark in the default package and it didn't work or something. Either way, this comment is probably not useful at all and should be removed.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: Ah I see. It's just confusing because of the horizontal/vertical flip.

djspiewak:
@edmundnoble I'm definitely very :+1: on splitting things into multiple PRs. I didn't get a chance to do a more comprehensive review on #3644, which is part of why I didn't catch a lot of this stuff.

I would definitely make the performance change here (the throwaway work at the column boundary), but otherwise we can have separate PRs for the other stuff.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: Wouldn't this be more efficient?

```scala
(acc.keys & flattened.map(_._1).toSet).isEmpty
```

Or even:

```scala
(acc.keys -- flattened.map(_._1)).size == acc.size
```

Or something like that. Difference between all three should be relatively small.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: Yeah, that too. I was thinking about `Sql2RegressionSpec`, but we're probably still hitting it in other pathways.

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
edmundnoble: So I was thinking, when it comes to slice limiting, our metric for "large slices" matters a lot. I argue that it's not the row count or column count individually, at least not directly, which makes for a "large slice" - rather, it's the number of CValues.

There's a caveat there though; we *do* prefer to (say) have a slice which is a single column and 16384 rows, compared to a slice which is 16384 columns in a single row. That's what I meant by the weighting of `cols` as having a larger contribution to the size of a slice.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Yeah. It used to be called `size`, which didn't incur that mental cost, at the additional cost of being much less explicit. Only winning move is not to play.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Good thing we have benchmarks :wink:

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Good point. Though I suppose then that it's inconsistent that `fromRValues` doesn't do this now.

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
djspiewak: It's not just the number of `CValue`s though, because for the primitives, an undefined value takes the same memory footprint as a defined one. Additionally, the contiguous nature of the pointer arrays means that even a large set of undefined non-primitive scalars is going to impose a significant cost.

And of course it all plays against the fact that slice-level operations are extremely expensive and we want to pay those penalties as infrequently as possible.

Realistically, all of these things should be weighted in some way. We're pretty naive right now. Once we get more comprehensive benchmarks in place to measure the impact of slice configuration on overall performance. We might not actually need to be *too* optimal here, given the performance impact on reads of doing a better job at estimating. Measurements will tell, for sure.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: It doesn't? You mean in master before the `fromRValues` PR landed?

yggdrasil/src/main/scala/quasar/yggdrasil/Config.scala
edmundnoble: Sorry, I misspoke w.r.t `CValue`, I meant array elements.

Right, sparseness is part of it too.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: 👌

djspiewak:
So to summarize (since github is actually kinda terrible at managing sprawling PR discussions):

- [ ] There will be another cleanup PR to address some of the lighter concerns (e.g. the random reformatting and similar)
- [ ] The sub-optimal performance of the column overflow detection will be addressed in this PR

I'm 👍 on merging this once the column overflow detection performance point has been addressed in more detail (or falsified – hurray for objective benchmarks!).

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Oh I had a question about this. How do you indent the body? Is it one further in from `: fs2.Stream[IO, Slice]`, or is it on the same level?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Right, I mean that if you call `fromRValues` it doesn't call some "autocoerceRValues" function.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: @djspiewak I'd argue that what's conventional for quasar is to put the return type line 
```scala
: fs2.Stream[IO, Slice] =
```
four spaces indented to the right of the `def`, and then put the body two spaces indented (same indentation as the parameters).

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
edmundnoble: Also... the fact that it works without the `Either` import is disturbing.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: @djspiewak There is actually another use. In `ReductionLib`, a couple of the `extract` functions call `fromRValues`. What is a suitable replacement?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: @alissapajer Oh you're correct. I was wrong. It's 4 spaces on the parameters and type, then 2 spaces on the body.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: `Slice.allFromRValues(Stream.emit(value))`? The reductions are just using it as a convenience to build a singleton slice. We could also write a function that built a singleton slice from a single `RValue`, but that's probably overkill.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
djspiewak: Yeah I'm quite confused about that. There's probably an `Either` in a package object somewhere, because that's how much the universe hates us.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: @djspiewak I was thinking of:
```scala
def allFromRValues(
  values: fs2.Stream[IO, RValue],
  maxRows: Option[Int] = None, 
  maxColumns: Option[Int] = None)
    : fs2.Stream[IO, Slice] = {
  val x = 17
  moarStuffHere
}
    
```

djspiewak: @alissapajer Quad-indenting the parameters is more conventional since it remains uniform with the following situation:

```scala
def foo(
    a: Int,
    b: Int,
    c: Int): Int = {

  a + b + c
}
```

We can *always* split the type to the next line, but then we have issues with local members which don't always ascribe return types.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: It actually looks like this change makes `fromRValues` much slower. Instead of speculatively doing more work than we need (and it being reversed when we hit column bounds), now we're doing work *always*. This did inspire me to add a couple of new benchmarks which I thought would show an advantage, but didn't in the end.

```
old

Benchmark                                                     (streaming)  Mode  Cnt     Score     Error  Units
FromRValuesBenchmark.ingestArrsWithFittingColumns                    true  avgt   10   106.724 ±   1.659  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns                true  avgt   10   891.529 ±  64.093  ms/op
FromRValuesBenchmark.ingestLongs                                     true  avgt   10    32.149 ±   0.869  ms/op
FromRValuesBenchmark.ingestObjectsWithDistinctColumns                true  avgt   10   779.295 ±  45.963  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns                 true  avgt   10   138.205 ±   6.648  ms/op
FromRValuesBenchmark.ingestObjectsWithNarrowScrollingColumns         true  avgt   10    89.481 ±   1.444  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns             true  avgt   10  1365.056 ± 198.752  ms/op
FromRValuesBenchmark.ingestObjectsWithWideScrollingColumns           true  avgt   10   766.302 ±  42.630  ms/op
```

```
new

Benchmark                                                     (streaming)  Mode  Cnt     Score     Error  Units
FromRValuesBenchmark.ingestArrsWithFittingColumns                    true  avgt   10   246.509 ±   3.554  ms/op
FromRValuesBenchmark.ingestArrsWithOverflowingColumns                true  avgt   10  1241.031 ±  90.570  ms/op
FromRValuesBenchmark.ingestLongs                                     true  avgt   10   137.812 ±   0.758  ms/op
FromRValuesBenchmark.ingestObjectsWithDistinctColumns                true  avgt   10  1090.314 ±  62.502  ms/op
FromRValuesBenchmark.ingestObjectsWithFittingColumns                 true  avgt   10   215.296 ±   4.580  ms/op
FromRValuesBenchmark.ingestObjectsWithNarrowScrollingColumns         true  avgt   10   156.971 ±   1.571  ms/op
FromRValuesBenchmark.ingestObjectsWithOverflowingColumns             true  avgt   10  1828.117 ± 114.758  ms/op
FromRValuesBenchmark.ingestObjectsWithWideScrollingColumns           true  avgt   10  1078.779 ±  65.012  ms/op
```

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: Then I'd have to run the effects inside `allFromRValues`, and assume that it returns a single slice, potentially by using `compile.last`. I'm not sure exactly why a singleton slice is being built there, though; it also seems from initial inspection that it won't necessarily be a singleton.

For now I've postponed this in https://github.com/slamdata/quasar/pull/3687.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: @edmundnoble Fascinating! That legitimately surprises me. I wonder if it wasn't possible to implement it in exactly the way I was thinking. Either way, thanks for trying it and measuring.

edmundnoble:
@djspiewak Ready to push the button?

djspiewak:
@edmundnoble Almost. You can definitely base work on this under the assumption it will be merged. I need to look at it again really quickly and I've been slammed this morning.

PR #3661, opened 2018-06-29
Upgrade qdata

alissapajer:
Also adds recursive generator for `Data` so we can generate `Data` that has nested objects/arrays.

### DIFF
--- a/frontend/src/test/scala/quasar/DataGenerators.scala
+++ b/frontend/src/test/scala/quasar/DataGenerators.scala
@@ -32,8 +32,6 @@ import java.time.{
 trait DataGenerators {
   import DataGenerators._
 
-  def genKey = Gen.alphaChar ^^ (_.toString)
-
   implicit val dataArbitrary: Arbitrary[Data] = Arbitrary(data)
 
   implicit def dataShrink(implicit l: Shrink[List[Data]], m: Shrink[ListMap[String, Data]])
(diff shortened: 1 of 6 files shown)


frontend/src/test/scala/quasar/DataGenerators.scala
```diff
@@ -67,19 +65,24 @@ object DataGenerators extends DataGenerators {
...
+  def genData(maxDepth: Int, atomic: Gen[Data]): Gen[Data] =
+    if (maxDepth < 1)
+      atomic
+    else
+      Gen.oneOf(genNested(maxDepth - 1, genKey, atomic), atomic)
 
-  val data: Gen[Data] = Gen.oneOf(
-    simpleData,
-    genNested(genKey, simpleData))
+  val data: Gen[Data] = genData(3, simpleNonNested)
```
alissapajer: I set the recursion depth to `3`. The unit tests take 10 minutes 17 seconds to complete at this depth. When I set it to `5` the unit tests took ~100 minutes to complete!

The depth can always be bumped up/down for individual specs as well by explicitly setting the generator in the test.

frontend/src/test/scala/quasar/DataGenerators.scala
```diff
@@ -67,19 +65,24 @@ object DataGenerators extends DataGenerators {
...
-  val simpleData: Gen[Data] = genAtomicData(Gen.alphaStr, defaultInt, defaultDec, defaultId)
+  val simpleNonNested: Gen[Data] = genNonNested(Gen.alphaStr, defaultInt, defaultDec, defaultId)
+
+  ////
+
+  def genData(maxDepth: Int, atomic: Gen[Data]): Gen[Data] =
+    if (maxDepth < 1)
+      atomic
+    else
+      Gen.oneOf(genNested(maxDepth - 1, genKey, atomic), atomic)
```
djspiewak: Do we want to bias towards shallow data? At present, the average generated `Data` will have depth 2. If we want to bias, we can use `Gen.frequency` here weighted by `maxDepth` such that we asymptotically approach the bound.

project/Dependencies.scala
```diff
@@ -30,7 +30,7 @@ object Dependencies {
   private val spireVersion        = "0.14.1"
   private val akkaVersion         = "2.5.1"
   private val fs2Version          = "1.0.0-M1"
-  private val qdataVersion        = "1.0.1"
+  private val qdataVersion        = "1.0.4"
```
djspiewak: Do we want to setup auto-bump for qdata in quasar?

frontend/src/test/scala/quasar/DataGenerators.scala
alissapajer: What's the benefit of biasing toward shallow data?

project/Dependencies.scala
alissapajer: Hmmm.

When the PR is closed, is the branch deleted from the repo? What if the PR is closed without merge, is the branch still deleted?

If the build does pass, then we can just merge the PR as is. But if the build doesn't pass, then I'd want to rebase so that the changes fixing the build are in the same commit as the version bump itself. If we do that, then version bump PRs should always be treated as WIP. Unless there's an argument for having a commit that _just_ bumps the version?

frontend/src/test/scala/quasar/DataGenerators.scala
djspiewak: Faster tests.

project/Dependencies.scala
djspiewak: > When the PR is closed, is the branch deleted from the repo? What if the PR is closed without merge, is the branch still deleted?

In neither case is the branch deleted. It should be possible to modify the functionality to use a fork (e.g. on the @slamdata-bot account), but it doesn't do so now.

> If the build does pass, then we can just merge the PR as is. But if the build doesn't pass, then I'd want to rebase so that the changes fixing the build are in the same commit as the version bump itself. If we do that, then version bump PRs should always be treated as WIP. Unless there's an argument for having a commit that just bumps the version?

I agree with this. We can change the autobumper to prepend [WIP] easily enough.

frontend/src/test/scala/quasar/DataGenerators.scala
alissapajer: I'll increase the max depth but bias toward shallow data and see how long the unit tests take. If we can find a happy medium there I'll go with it.

djspiewak: Yeah, I think that would be interesting. Something like:

```scala
Gen.frequency(
  (if (maxDepth < 0) 0 else math.exp(maxDepth).toInt) -> genNested(maxDepth - 1, genKey, atomic),
  20 -> atomic)  // e^3
```

Ish?  So we bias towards creating at least one level of nesting (at a 27:10 ratio), but we rapidly fall off and almost never create the fourth level.

frontend/src/test/scala/quasar/DataGenerators.scala
```diff
@@ -65,22 +64,39 @@ object DataGenerators extends DataGenerators {
...
+    Gen.oneOf[Char]("0123456789abcdef") * 24 ^^ (_.mkString)
+
+  val simpleNonNested: Gen[Data] =
+    genNonNested(Gen.alphaStr, defaultInt, defaultDec, defaultId)
+
+  ////
 
-  // TODO: make this very conservative so as likely to work with as many backends as possible
-  val simpleData: Gen[Data] = genAtomicData(Gen.alphaStr, defaultInt, defaultDec, defaultId)
+  private val maxDepthDefault: Int = 4
```
alissapajer: I set the `maxDepthDefault` to `4` and added `Gen.frequency`. The unit tests just completed in 13 minutes and 3 seconds.

PR #3663, opened 2018-07-02
Don't trust LP's let bindings

djspiewak:
Depends on #3648 

The `Let` construction in `LogicalPlan` isn't guaranteed to fully-deduplicate things. Specifically, it will miss any duplication which is hidden by adjacency within child nodes, as one would expect from a process which uses recursion without `State`.

The fix is to have special recursion handling for `Let` in `ReadLP`. Rather than using `cataM`, we use our own version of it which interleaves scope management between the recursive processing of `form` and `in`. Recursion schemes ftl. ☹️ Anyway, the state in particular carries along the renaming of the LP-chosen `Let` symbol to the one which is chosen by `withName`, which allows the duplication to collapse recursively all the way up the graph. This replaces the old rebase-and-rename scheme, which only worked at the leaves.


djspiewak:
[ch190]
[ch222]
[ch357]

PR #3668, opened 2018-07-03
Added forking to all the things, as it should have been

djspiewak:


edmundnoble:
I suggest that in addition to this, we add all of `.jvmopts` into the `javaOptions` key.

djspiewak:
Oh that's definitely worth doing. I'll take care of that quickly…

edmundnoble:
Hmmm I suppose this is good for now though eventually we may want different JVM options for SBT itself and the forked tasks.

djspiewak:
Don't merge this yet. Terrible things are happening

edmundnoble:
Perhaps we should only merge this after the fixed specs2 version is available?

djspiewak:
@edmundnoble Agreed

alissapajer:
@djspiewak there is here that possibly needs to change https://github.com/slamdata/quasar/blob/master/build.sbt#L502-L505

djspiewak:
@alissapajer Oh crap, yes. That's not what's causing the mongoIt failure though (I know what is, I just haven't fixed it yet).

djspiewak:
[ch270]

djspiewak:
Closing. I'll loop back to this someday soon.

PR #3669, opened 2018-07-05
Downgrade to fs2-0.10.5 and cats-effect-0.10.1

wemrysi:
These are compatible with the versions that `http4s` uses.


wemrysi:
[ch363]

djspiewak:
@wemrysi Are those integration test link errors legit? Or just incremental compilation failing us… again…

wemrysi:
@djspiewak Ugh, might be legit as I don't think I ran the regression tests locally. Will take a look.

wemrysi:
@djspiewak So looks like these errors are probably due to caching as I don't see them locally, however a couple of the time-related `lwc_local` regression tests fail, which I don't really understand. Going to merge from `master` and try agian.

PR #3671, opened 2018-07-06
Initial addition of Tables

alissapajer:


alissapajer:
There is still some ongoing discussion in the gist that can be incorporated here.

alissapajer:
I didn't add `Equal` or `Show` instances anywhere yet. I can do this once the API solidifies a bit more (in this PR or later).

The next step is to add a mock so that development work can begin on the REST API. If people prefer to see this merged along with an implementation I can do that here; otherwise I can do it in a separate PR.

api/src/main/scala/quasar/api/TableAttribute.scala
```diff
@@ -0,0 +1,32 @@
...
+package quasar.api
+
+import slamdata.Predef.{Product, Serializable, String}
+
+sealed trait TableAttribute extends Product with Serializable
+
+object TableAttribute {
+  final case class Name(value: String) extends TableAttribute
+  final case class Sql2(value: String) extends TableAttribute
+  final case class BlackBox(value: String) extends TableAttribute
```
alissapajer: Originally @wemrysi and I were thinking that `Sql2` and `BlackBox` would be combined into a single constructor but @rintcius suggested that splitting them apart could be better. 

Do we want to semantically tie `Sql2` and `BlackBox`? That's really the question. We know that `BlackBox` will be semantically tied (at least right now) to `Sql2`, but this might not always be the case. How much of the frontend's usage of `BlackBox` do we want to encode in these types...

api/src/main/scala/quasar/api/TableAttribute.scala
```diff
@@ -0,0 +1,32 @@
...
+}
+
+final case class Table(
```
alissapajer: Ideas for a better home for this?

api/src/main/scala/quasar/api/PreparationStatus.scala
```diff
@@ -0,0 +1,43 @@
...
+  final case class Preparing(startedAt: OffsetDateTime)
+      extends Inaccessible
+
+  final case class Repreparing(startedAt: OffsetDateTime, previous: Finished)
+      extends Accessible
+
+  final case class Prepared(startedAt: OffsetDateTime, duration: Duration, size: PreparationSize)
+      extends Accessible with Finished
+
+  final case class Errored(startedAt: OffsetDateTime, duration: Duration, error: PreparationError)
```
alissapajer: I'm not convinced we need `PreparationError` to be included here. Any error in preparation we could return in the `F` in `def preparationStatus(tableId: I): F[TableError \/ PreparationStatus]`. If we get an error during preparation, it would be because the disk is full, or because nihdb had a bug and a write failed, or because... ? Are those anticipated errors within the scope of the application which the user could fix (in which case we put them in `PreparationError`) or errors outside the scope of the application (in which case we put them in `F`).

(And also, do people agree with using that as a criteria for where to house errors? There are probably better words to use to describe the distinction...)

api/src/main/scala/quasar/api/Tables.scala
```diff
@@ -0,0 +1,50 @@
...
+  def setTableAttributes(tableId: I, attributes: NonEmptyList[TableAttribute]): F[Condition[TableError]]
+
+  // errors: table not found, preparation in progress
+  def prepareTable(tableId: I): F[Condition[TableError]]
+
+  // errors: table not found
+  def preparationStatus(tableId: I): F[TableError \/ PreparationStatus]
+
+  // errors: table not found
+  def preparedData(tableId: I): F[TableError \/ PreparationResult[D]]
```
alissapajer: Discussion point: should we add a `format` parameter to `preparedData` as broached by @wemrysi ?

alissapajer:
Also, we could make this a `version: feature` now or when the mock gets added. I was thinking it's not usable until there is the mock, so I kept it at `version: revision`.

api/src/main/scala/quasar/api/PreparationStatus.scala
rintcius: Hmm, I think `preparationStatus` should return `Errored` no matter what the reason for this error was. Otherwise it can become confusing whether the error came from the preparation or from obtaining the preparation status. 

In order to distinguish between different preparation errors, we could for example add an optional field `nested: Throwable` to `PreparationError` and/or add subtypes to `PreparationError`.

How about changing the signature to 
`def preparationStatus(tableId: I): F[PreparationStatus]`? In that way it's clear that the error came from the preparation. 

I don't think this is necessary now, but if we still need an error to be returned here then it could be something like `def preparationStatus(tableId: I): F[PreparationStatusError \/ PreparationStatus]`

api/src/main/scala/quasar/api/TableAttribute.scala
rintcius: Even with the current semantics I don't think they are fully tied. Eg in the repl we will typically care only for `Sql2` and not `BlackBox`.

wemrysi: @alissapajer Yeah, I think splitting them is fine, it places the control of how coupled they are into the frontend's hands.

On that note, and I haven't thought this through yet, what would you think about removing `Blackbox` from `Tables` and we have a separate API for managing the data the frontend associates with a table? The http endpoint would stay the same, but we'd separate the concerns at this level?

api/src/main/scala/quasar/api/Tables.scala
wemrysi: I'm fine either way, if we want to add it now or later. I think we only need to support `csv` for 0.1.0 so we can also defer until next iteration.

alissapajer: Let's defer until next iteration so as not to over design and to keep things as simple as possible.

api/src/main/scala/quasar/api/TableAttribute.scala
alissapajer: So one thing that comes to mind is that the frontend might want to store more than just a single blob of data (if not now, then in future product versions). And so we could then grow the quasar API to accommodate other data (table-related or not) that they might want to store in nihdb.

That way the `Tables` API represents what quasar is concerned with. And we can then separately manage whatever data the frontend cares about. 

So, I think I like the idea! Any counterarguments... hrm...

alissapajer: Following that through, this `FrontendBlackBox` (better name?) API could then serve data to multiple http endpoints. But since that data would never be used in quasar we could implement it as a key-value store.

api/src/main/scala/quasar/api/PreparationStatus.scala
alissapajer: What I have now is
```scala
def preparationStatus(tableId: I): F[ExistenceError[I] \/ PreparationStatus]
```
where `ExistenceError` can only be `TableNotFound`. All the methods that require a table to exist error with `TableNotFound` so this is consistent across `Tables`. And this is an error that had nothing to do with preparation so I think it makes sense.

My outstanding question is: given that a table exists and a preparation begins, in what situations will it then error? Those are the situations we could create a `PreparationError`. A couple examples that would hit this situation are that the disk is full, or nihdb had a bug and a write failed. Those are fatal scenarios (even if the whole application doesn't crash), and so I didn't see them as errors specific to preparation and so they wouldn't be `PreparationError`s. Are there actually any valid `PreparationError`s?

alissapajer:
@rintcius @wemrysi ready for another look.

api/src/main/scala/quasar/api/PreparationStatus.scala
wemrysi: @alissapajer I think you're correct that there aren't any errors unique to preparation. However, preparation is unique (thus far) as it is the only first-class asynchronous operation we have. So another question would be, do we reify the fact that this process can fail, or do we make that implicit? One attribute of an async process is whether or not it is in progress and, if not, why. Since success or failure are the primary partitions of "not in progress" including them both in the type seems reasonable. Maybe we have `Errored(startedAt: OffsetDateTime, duration: Duration, cause: Throwable)` so we can capture whatever caused the failure?

Reporting the async failure via the `F[_]` effect may also be a bit awkward since the `preparationStatus` api is informing the caller of the status, not performing the preparation itself.

wemrysi:
LGTM :+1: , I'm going to merge to keep things moving along. @rintcius if you have any more comments, there should be plenty of time to address them in future PRs.

PR #3672, opened 2018-07-06
Include lwc_local_insert config in example config

alissapajer:


alissapajer:
It is very much poor design that the `lwc_local` connector will silently run against data in nihdb if the `lwc_local_insert` config isn't provided. This is an artifact of the design of our integration test suite. If the `lwc_local` tests are not running against data in the local filesystem, they should not run at all. We just don't have time to fix this now.

alissapajer:
Also I want to mention that Travis has always been doing the right thing. It adds (and always has added) the `lwc_local_insert` config. So this PR is just clarifying how to run `lwc_local` tests locally.

alissapajer:
From travis logs:
```
The command "docker/scripts/assembleTestingConf -i quasar_$CONNECTOR" exited with 0.
0.01s$ cat it/testing.conf
postgresql_metastore="{\"host\":\"localhost\",\"port\":5432,\"database\":\"metastore\",\"userName\":\"postgres\",\"password\":\"\"}"
lwc_local="{\"mimirDataDir\": \"/tmp/lwc-local/\"}"
lwc_local_insert="{}"
```

PR #3673, opened 2018-07-06
update travis cache workflow

nicflores:


nicflores:
The idea here is to try to resolve the travis caching errors we've been seeing lately.

djspiewak:
Why remove `.hoarder-cache`? That basically makes hoarder useless.

Removing `.targets` makes sense, and I'm relatively certain this is part of the issue which underlies a lot of our cache problems, but removing `.hoarder-cache` is just going to make all PRs much slower to build.

alissapajer:
@djspiewak the idea was that we could use s3 to fetch the hoarder artifacts for _all_ jobs. The Compile stage would fetch the hoarder artifacts from s3 from the last successful master build. Then subsequent stages (and jobs within stages) would all fetch the hoarder artifacts from s3 that were put there during this build when the Compile stage succeeded.

alissapajer:
If we're not going to make the change to fetch from s3 for all jobs, then I agree we should keep in `.hoarder-cache`.

djspiewak:
@alissapajer Why would we though? Travis already does this for us, literally using S3 in fact. Why would we recreate that infrastructure?

wemrysi:
@djspiewak The difference would be that the `compile` stage would _always_ fetch the master cache, travis will use the previous PR cache on subsequent builds, no?

alissapajer:
I thought travis used the master cache if there wasn't one available for a given PR.

djspiewak:
@wemrysi @alissapajer Oh, you want to always use the master cache! That is indeed a difference. Why?

alissapajer:
@djspiewak the master cache is a stable build that we know the given PR is based off (or should/could be based off). PR caches include all sorts of WIP work (including library up/downgrades) that, even if compiling, might never land in master. Master is what's stable.

alissapajer:
This are the caches travis checks, in order:
https://docs-staging.travis-ci.com/user/caching/#Pull-request-builds-and-caches

djspiewak:
@alissapajer I mean, that explains why you fetch from master for the first PR build. But why would you fetch from the master cache for subsequent PR builds? Sure, the whole PR is built off master, but we have a more recent cache than that.

alissapajer:
@djspiewak within a single PR, someone could still upgrade and downgrade a dependency, or make some other kind of non-binary-compatible change.

djspiewak:
@alissapajer Why is that a problem? You can do that locally, too. SBT handles these things really really well. I'm less convinced that *hoarder* handles them as well, but we don't have much direct evidence of that. So far, every time we've seen hoarder cause incremental compilation weirdness (both in travis and locally), it has been in conjunction with copying the target classes, which hoarder actually doesn't say to do.

Now, hoarder could still be broken, and if that's the case then we need to get rid of it entirely and just deal with the fact that PR builds are going to be about 30 minutes longer than they could otherwise be. But I don't see the reason to keep hoarder but *only* cached from master.

djspiewak:
Just to be really clear, SBT's incremental compilation (which is the mechanism that hoarder hooks into and the databases it is caching) is considerably more sophisticated than a simple binary compatibility check. It understands the dependencies between individual compilation units on a very granular level and it is quite conservative with invalidating those dependencies when things *could* be broken.

wemrysi:
@djspiewak Ok, so you think that ceasing to cache the compilation artifacts should solve the issues we're seeing?

djspiewak:
@wemrysi I think so. It's at least worth a shot. If that *doesn't* solve the issue, then hoarder itself is responsible and we shouldn't be fetching any of its caches at all, master or otherwise.

alissapajer:
@djspiewak stepping back, our goal is to fix the caching problems we've seen in travis. I am on board with incrementally working our way toward a solution. If hoarder doesn't say we should be copying over the target classes, then let's start there, by not copying them over. If we still have problems, then we can figure out what next steps to take.

nicflores:
Ok, I'll add hoarder back into the travis cache we can go from there.

PR #3674, opened 2018-07-06
Delete TestDirective.Timeout

alissapajer:
It was unused. We can use `"skip"` instead.


alissapajer:
@jsantos17 could I get a review on this? It's not product-urgent but if it gets out of date then that will require time to get it up to date.

(Also, this will be the last PR I submit before July 23 that isn't extremely directly related to `v0.1.0`, I promise :stuck_out_tongue: )

jsantos17:
Totally missed this. Sorry :(

PR #3676, opened 2018-07-09
Removed hoarder

djspiewak:
I've confirmed locally that hoarder cannot handle quasar for some reason. I'm not sure why, but it appears to be the case. Fortunately, @nicflores's S3 stuff should ensure that we're not completely bonkers on build times, but it's going to be unfortunate. The PR compilation stage will always be from scratch now. *grumble grumble grumble*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,6 @@ script:
   - docker/scripts/assembleTestingConf -i quasar_$CONNECTOR
   - cat it/testing.conf
 
-  - ./sbt preBuild
-
   - |-
     SPECIFIC_DELEGATE=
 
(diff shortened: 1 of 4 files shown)


alissapajer:
We should also update the `scripts/s3cache` script so that it doesn't upload/download the hoarder artifacts.

alissapajer:
@djspiewak travis compile staged failed with our good old friend the scalac SOE. I'll let you restart the build.

nicflores:
@alissapajer and @djspiewak, I can update the s3cache script to not include hoarder.

alissapajer:
@nicflores that was already done in a8657cb

nicflores:
@djspiewak, yep, just saw the commit! thanks Daniel.

PR #3677, opened 2018-07-09
Boxless Table#renderCsv

djspiewak:
By convention, we sort of assume that things *inside* the cake are not part of the binary compatibility signature, so I'm versioning this as feature. Let the arguments commence!

This is a complete reimplementation of `renderCsv` in the style of `renderJson`, which doesn't suffer from this issue. Apparently the old `renderCsv` paged the *entire* dataset into a single `String`! Woot. Ostensibly, this was done to handle situations where the dataset was heterogeneous divided by slice boundaries (and thus invisibly at the start of the stream). However, as evidenced by the old (incorrect) test assertions, even this situation was improperly handled! Basically, it was broken, and it would cause crashes if we used it on large datasets.

Oh, and it actually paged the dataset into memory *twice*: once to build a bunch of `Array[String]` objects, and then once to produce the big `String`-o-doom. And that's not counting the incremental slice paging. It was bad.

The new one avoids boxing, allocation, and polymorphism in the hot path at the expense of considerable ugliness. It also uses a cute trick to avoid accidental over-specialization on the PIC of a critical path function, which makes things look a bit uglier than they need to. Also I wish we had untagged unions in Scala. I didn't comment things over-aggressively, but I did try to add some explanation when things were particularly bizarre.

The new `renderCsv` takes a parameter, `assumeHomogeneous`, which governs whether or not we make an unsafe assumption about the nature of the dataset. If the parameter is `true`, we'll use the schema of the first slice to dictate the full rendering. *This is very likely to be the case in production!* However, this assumption *is* unsound since we could have a dataset which is heterogeneous such that the first slice has an entirely different schema than the second slice. Column headers need to be emitted at the start, as do column separators within rows. In order to deal with this scenario (which probably won't happen in production anyway), we traverse the dataset twice in the `false` case. We use the first traversal to infer the schema, and then the second traversal to actually emit. This also changes some of the semantics of the data structures and bounds checks within the hot path (changing `Array[Column]` to `Array[Array[Column]]`, for example).

Anyway, requesting a review from @edmundnoble as well, just to get a second pair of eyes on this incredibly hairy piece of code that I wrote in a rush mid-road-trip.


djspiewak:
[ch427]

djspiewak:
Oh, pro-tip for the reviewers: the diff is *extremely* difficult to read in linear form due to the fact that I had to move around large chunks of code and the entire implementation shifted from `ColumnarTableModule` to `Slice`. Sorry.

I highly recommend reviewing in side-by-side view, ignoring whitespace.

djspiewak:
@alissapajer Yeah I don't really get it. I'm going to try running it locally as soon as I have a computer with which to do so. I tried clearing the PR caches and it didn't help.

rintcius:
@djspiewak FYI I haven't forgotten about this, just planning to have the repl PR up first

djspiewak:
@rintcius That's fine, we just can't let this sit for very long due to the extensive changes to `Slice`. It's going to become conflicted in a hurry. If you don't have time to deal with it now, feel free to assign it to someone else.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -122,202 +119,93 @@ object ColumnarTableModule extends Logging {
...
-      * Since we know in advance how many rows we have, we can return
-      * an array of lines.
-      */
-    def renderSlice(pastIndices: Option[Indices], slice: Slice): (Indices, CharBuffer) = {
+                val candidate = candidateHead + candidateTail
 
-      val indices = indicesForSlice(slice)
-      val height  = slice.size
-      val width   = indices.size
+                if (candidate.indexOf('"') >= 0 &&
```
rintcius: shouldn't this be `||`-ed together?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -988,163 +992,213 @@ trait Slice { source =>
...
+              val c = column.asInstanceOf[LocalDateColumn]
+              ctx.renderLocalDate(c(row))
+
+            case CInterval =>
+              val c = column.asInstanceOf[IntervalColumn]
+              ctx.renderInterval(c(row))
+
+            case CArrayType(_) => ???
+
+            case CUndefined =>
```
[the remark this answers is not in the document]
djspiewak: How is that more clear?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
djspiewak: They should. Good catch.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -2197,4 +2210,140 @@ object Slice {
...
+    }
+
+    def renderString(str: String): Unit = {
+      if (_renderJsonString eq null)
+        _renderCsvString(this, str)
+      else
+        _renderJsonString(this, str)
+    }
+
+    def renderLong(ln: Long): Unit = {
```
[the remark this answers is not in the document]
djspiewak: `ln.toString` doesn't handle `MinValue` correctly, and can also be quite slow in certain cases. `renderLong` is much more optimized.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1162,228 +1216,77 @@ trait Slice { source =>
...
+                case '"' => ctx.pushStr("\\\"")
+                case '\\' => ctx.pushStr("\\\\")
+                case '\b' => ctx.pushStr("\\b")
+                case '\f' => ctx.pushStr("\\f")
+                case '\n' => ctx.pushStr("\\n")
+                case '\r' => ctx.pushStr("\\r")
+                case '\t' => ctx.pushStr("\\t")
+
+                case c =>
+                  if ((c >= '\u0000' && c < '\u001f') || (c >= '\u0080' && c < '\u00a0') || (c >= '\u2000' && c < '\u2100')) {
```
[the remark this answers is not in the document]
djspiewak: I can't.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
rintcius: Just to make explicit that nothing has been forgotten, but feel free to leave as-is if that's your preference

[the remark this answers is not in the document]
djspiewak: I prefer its current form, so I'll leave it. I generally wouldn't put `()` in a no-op side-effecting `case`, stylistically.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
rintcius: curious, what's the bug for `MinValue`? I couldn't find any difference, maybe it was solved in the meantime?

ok fair enough, though I'm not the biggest fan of this kind of code, even in the hot path

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -122,202 +119,93 @@ object ColumnarTableModule extends Logging {
...
-        if (n == 0) return ()
-        sb.append(a(0))
-        var i = 1
-        val len = n
-        while (i < len) { sb.append(','); sb.append(a(i)); i += 1 }
-        sb.append("\r\n")
-      }
-    }
+  def renderCsv(slices: StreamT[IO, Slice], assumeHomogeneous: Boolean): StreamT[IO, CharBuffer] = {
+    val schemaFirstTailOptM: IO[Option[(Map[CPath, Set[CType]], Slice, StreamT[IO, Slice])]] = {
```
alissapajer: Thanks for the annotation. :+1:

djspiewak: I figured that if I was having trouble remembering the type *as I wrote the code*, it was worth putting in an explicit ascription.

alissapajer:
@djspiewak I'm not a fan of how we pass around `assumeHomogeneous` and subsequently check it with an `if` statement in a handful of places. It makes the code a harder to reason about than if we were to factor out the commonalities between the the cases and then write separate functions.

I understand this was written hastily. But this function will be called after every preparation since csv is the only rendering option we're offering, and so we have to get it right. We do have to carefully choose how we spend our time during the next few days, but I would like to see this code path reorganized.

Though will all that said, which assumption were we planning on using in production? I don't see how we could ever `assumeHomogeneous = true` in production because if that turns out to be an incorrect assumption, the csv will also be incorrect and likely unreadable by any upstream applications. `assumeHomogeneous` is a nice feature to have when we absolutely know for sure we are homogeneous, but we won't know that in production (at least not currently), meaning we always have to be pessimistic and set it to `false`. So should we just omit this flag entirely now to keep the code simpler?

djspiewak:
@alissapajer Actually we're almost always going to know that we're homogeneous in production because of the column picker. That's why this flag exists, and why it will generally default to `true`. Originally, it didn't get passed around everywhere. Instead, it was handled entirely within `Table#renderCsv` in the construction of the schema. However, it was less efficient to manage the differing codepaths within `Slice#renderCsv` without the boolean (since it would have required more frequent type tests).

The decision to keep everything in a single function mostly has to do with concision and also having as much overlap between the codepaths as possible. That overlap allows us to lean more heavily on the existing tests, rather than having to exhaustively test each function independently. There might be a way to reorganize this such that we still get good test reuse and minimal repetition without the current type union obfuscation, but I'm not 100% convinced of it. The dependencies between the code paths are more intrinsic than they initially seem.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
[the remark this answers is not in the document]
djspiewak: It may have been solved in the meantime.

I could remove it, but I basically just copied it from `renderJson`, where it was hand-tuned by (I think) @d6. I would be leery of removing it now without benchmarks.

alissapajer:
@djspiewak how does the column picker tell us anything about homogeneity? There still very well might be holes in the data. Does the column picker force the user to fill in those holes with some value (default or derived)?

djspiewak:
@alissapajer I believe holes are fine in `assumeHomogeneous`, it's just that type unions are not.

djspiewak:
To clarify, if *both* the `String` and `Boolean` types are selected from a projection, it gets put into the table builder as two separate columns. There's no way to build a heterogeneous column in the output table.

alissapajer:
@djspiewak so if a user selected both the `String` and the `Boolean` values from projection `foo.bar.baz`, then they would be required in the column picker to create _two_ distinct columns (each with its own name of course).

alissapajer:
I hadn't appreciated that subtlety about the column picker before. It's actually (among other things) a tool that separates out type unions (i.e. turns heterogeneous columns into homogeneous columns).

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -122,202 +119,93 @@ object ColumnarTableModule extends Logging {
     * becomes:
     *
     * "the fox said: ""hello, my name is fred."""
+    *
+    * If `assumeHomogeneous` is true, the columns from the first slice
+    * will dictate the header only one column from a polymorphic
+    * `ColumnRef` locus (e.g. a field which can be String or Long) will
+    * be used. If `assumeHomogeneous` is false, then the dataset will be
+    * fully traversed to infer the schema, which will then be used to
+    * render the CSV bytes in a second traversal.
```
edmundnoble: What is the behavior if you use `assumeHomogeneous` and the data actually ends up heterogeneous? Do we just see the values in the schema of the first slice? Could you comment to that effect?

edmundnoble:
I'm trying to absorb all of this right now, it seems at least somewhat straightforward but could really use more comments describing the procedure, especially around the more meaty parts (`Slice.scala`).

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1786,6 +1653,152 @@ trait Slice { source =>
   }
 
   override def toString = (0 until size).map(toString(_).getOrElse("")).mkString("\n", "\n", "\n")
+
+  private[this] def sliceSchema: Option[SchemaNode] = {
+    if (columns.isEmpty) {
+      None
+    } else {
+      def insert(target: SchemaNode, ref: ColumnRef, col: Column): SchemaNode = {
```
edmundnoble: Would it make sense to expose these to the rest of `Slice` for use in other operations, like `updateRef`?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
djspiewak: Several things break if data is *actually* heterogeneous. Off the top of my head:

- The value at a given locus in a type union will be nondeterministically chosen. This could result in holes in the data
- Speaking of holes, we would probably crash or at least hit some JVM-defaulted values since we assume density goes along with homogeneity (this needs to be fixed before merge, per the discussion with @alissapajer)
- Schemas that break across slices will end up with runtime exceptions dereffing the `columns` map
- Probably some other things

It's meant to be a flag only used when we have a *guarantee* that the data is homogeneous by definition.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
edmundnoble: `updateRefs` would benefit from using `SchemaNode`, too.

djspiewak: But… how?

edmundnoble: Every time we use `fromRValues` we build giant hashmaps representing each RValue. These are incredibly expensive in terms of allocation; on the heap profile, the slice arrays don't even show up relative to the hashmap copies, which require full array copies on each update.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
edmundnoble: Perhaps comment to the effect that "if the data isn't really homogeneous, any number of weird things including errors can happen".

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: Oh I see what you mean. Cool. I'll get on that.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
djspiewak: That's fair.

alissapajer:
@djspiewak let's say we have called this function assuming homogenous data, and that assumption is correct. _But_ the data in one column `foo.bar` is extremely sparse, and in the first slice there aren't any values in `foo.bar`. But there are values in subsequent slices. Then won't we assume that the `foo.bar` column doesn't exist and end up with incorrect CSV?

Could we solve this by allowing more columns to be added as we see subsequent slices? We'd have to hold some state as we did this. But it wouldn't be nearly as complex as the state required to merge columns in the heterogeneous case.

djspiewak:
> But the data in one column foo.bar is extremely sparse, and in the first slice there aren't any values in foo.bar.

Then we're in very serious trouble. There's no way to handle this in general aside from `assumeHomogeneous = false`, which does a full table scan before emitting (which in turn is a second table scan).

> Could we solve this by allowing more columns to be added as we see subsequent slices?

How? We have to emit the columns first, and then every single row is aligned to that column header. We would have go back and change everything *already emitted* in order to add columns after the fact. This is why CSV is hard.

djspiewak:
BTW the AppVeyor build failure is spurious and caused by cache issues that will resolve themselves once this is in master.

alissapajer:
@djspiewak if we hit the very sparse column case, is this function going to crash? Or are we silently going to omit incorrect CSV?

alissapajer:
@djspiewak then we should think of this as a placeholder that will get us through some alpha testing but won't hold up in the long run. If we don't implement the functionality that gets the schema during preparation soon, we can ask for the SST here. We can't afford to give our customers incorrect CSV; that seems like it would make the product untrustworthy, imo.

djspiewak:
@alissapajer The placeholder is to use `assumeHomogeneous = true`. Which is like… horrible. There's not really anything we can do otherwise though. The thing though is that we pervasively have this danger in the product. If the dataset has horrible things hidden late, we're just going to miss them in most of our functionality.

alissapajer:
Yeah good point. We can definitely store all the 'CPath's along with a table's preparation and then have that accessible for anyone to consume.

alissapajer:
@djspiewak pending green builds, anything else you plan to push to this or are you :+1: to merge?

PR #3682, opened 2018-07-10
Tables minor updates

alissapajer:


api/src/main/scala/quasar/api/table/PreparationResult.scala
```diff
@@ -23,5 +23,5 @@ sealed trait PreparationResult[A] extends Product with Serializable
 
 object PreparationResult {
   final case class Available[A](since: OffsetDateTime, value: A) extends PreparationResult[A]
-  final case class Unavailable[A](reason: Inaccessible) extends PreparationResult[A]
```
alissapajer: @wemrysi what do you think of this change? It semantically separates the preparation result from the preparation status. If a preparation result is unavailable, the preparation status will have to be queried for separately.

alissapajer: Since we've now encoded an `Option[Prepared]` in the `PreparationStatus.Errored` I don't think it makes sense to use a `PreparationStatus` as the `reason`. If we want a reason, we could create a new algebra that's one of `Unprepared` or `Errored` but I'm not sure that we need that right now.

wemrysi: @alissapajer Or how about `final case class Unavailable[A](condition: Condition[Throwable]) extends PreparationResult[A]`. If the current status is `Unprepared` then we'd have `Condition.normal`, if it was an error then `Condition.abnormal(err)`?

alissapajer: I like that. :+1:

PR #3683, opened 2018-07-11
initial version of repl based on quasar.api and jline3

rintcius:
cleaned up version of #3658 with all commits squashed into 1

### DIFF
--- a/repl/src/main/scala/quasar/repl2/Command.scala
+++ b/repl/src/main/scala/quasar/repl2/Command.scala
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar
+package repl2
+
+import slamdata.Predef._
+import quasar.api._
+import quasar.repl._
+import quasar.sql.Query
+
+import eu.timepit.refined.auto._
+import scalaz._, Scalaz._
+
+sealed abstract class Command
+
+object Command {
+  private val NamePattern                  = "[a-zA-Z0-9-]+"
+
+  private val ExitPattern                  = "(?i)(?:exit)|(?:quit)".r
+  private val HelpPattern                  = "(?i)(?:help)|(?:commands)|\\?".r
+  private val CdPattern                    = "(?i)cd(?: +(.+))?".r
+  private val LsPattern                    = "(?i)ls(?: +(.+))?".r
+  private val SetPhaseFormatPattern        = "(?i)(?:set +)?phaseFormat *= *(tree|code)".r
+  private val SetTimingFormatPattern       = "(?i)(?:set +)?timingFormat *= *(tree|onlytotal)".r
+  private val DebugPattern                 = "(?i)(?:set +)?debug *= *(0|1|2)".r
+  private val SummaryCountPattern          = "(?i)(?:set +)?summaryCount *= *(\\d+)".r
+  private val FormatPattern                = "(?i)(?:set +)?format *= *((?:table)|(?:precise)|(?:readable)|(?:csv))".r
+  private val SetVarPattern                = "(?i)(?:set +)?(\\w+) *= *(.*\\S)".r
+  private val UnsetVarPattern              = "(?i)unset +(\\w+)".r
+  private val ListVarPattern               = "(?i)env".r
+  private val DataSourcesPattern           = "(?i)datasources".r
+  private val DataSourceTypesPattern       = "(?i)types".r
+  private val DataSourceAddPattern         = s"(?i)(?:add +)($NamePattern)(?: +)($NamePattern)(?: +)(replace|preserve)(?: +)(.*\\S)".r
+  private val DataSourceLookupPattern      = "(?i)get +([\\S]+)".r
+  private val DataSourceRemovePattern      = "(?i)rm +([\\S]+)".r
+
+  final case object Exit extends Command
+  final case object Help extends Command
+  final case object ListVars extends Command
+  final case class Cd(dir: ReplPath) extends Command
+  final case class Select(query: Query) extends Command
+  final case class Ls(dir: Option[ReplPath]) extends Command
+  final case class Debug(level: DebugLevel) extends Command
+  final case class SummaryCount(rows: Int) extends Command
+  final case class Format(format: OutputFormat) extends Command
+  final case class SetPhaseFormat(format: PhaseFormat) extends Command
+  final case class SetTimingFormat(format: TimingFormat) extends Command
+  final case class SetVar(name: VarName, value: VarValue) extends Command
+  final case class UnsetVar(name: VarName) extends Command
+
+  final case object DataSources extends Command
+  final case object DataSourceTypes extends Command
+  final case class DataSourceLookup(name: ResourceName) extends Command
+  final case class DataSourceAdd(name: ResourceName, tp: DataSourceType.Name, config: String, onConflict: ConflictResolution) extends Command
+  //final case class DataSourceMove
+  final case class DataSourceRemove(name: ResourceName) extends Command
+
+  implicit val equalCommand: Equal[Command] = Equal.equalA
+
+  def parse(input: String): Command =
+    input match {
+      case ExitPattern()                            => Exit
+      case CdPattern(ReplPath(path))                => Cd(path)
+      case CdPattern(_)                             => Cd(ReplPath.Absolute(ResourcePath.Root))
+      case LsPattern(ReplPath(path))                => Ls(path.some)
+      case LsPattern(_)                             => Ls(none)
+      case DebugPattern(code)                       => Debug(DebugLevel.int.unapply(code.toInt) | DebugLevel.Normal)
+      case SetPhaseFormatPattern(format)            => SetPhaseFormat(PhaseFormat.fromString(format) | PhaseFormat.Tree)
+      case SetTimingFormatPattern(format)           => SetTimingFormat(TimingFormat.fromString(format) | TimingFormat.OnlyTotal)
+      case SummaryCountPattern(rows)                => SummaryCount(rows.toInt)
+      case FormatPattern(format)                    => Format(OutputFormat.fromString(format) | OutputFormat.Table)
+      case HelpPattern()                            => Help
+      case SetVarPattern(name, value)               => SetVar(VarName(name), VarValue(value))
+      case UnsetVarPattern(name)                    => UnsetVar(VarName(name))
+      case ListVarPattern()                         => ListVars
+      case DataSourcesPattern()                     => DataSources
+      case DataSourceTypesPattern()                 => DataSourceTypes
+      case DataSourceLookupPattern(n)               => DataSourceLookup(ResourceName(n))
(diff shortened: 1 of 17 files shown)


repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+      ref <- Ref[IO, ReplState](ReplState.mk)
+      repl <- Repl.mk[IO, IO](ref, q.dataSources, q.queryEvaluator)
+      l <- repl.loop
+    } yield l
+
+  def run(args: List[String]): IO[Unit] =
+    runQuasar(quasarStream, repl)
+
+  def main(args: Array[String]): Unit =
+    run(args.toList).unsafeRunSync
```
rintcius: I think there is now something incorrect here, after the downgrade of cats-effect. Maybe I should re-add the code that became unavailable due to the downgrade (e.g. `IOApp`, `ExitCode`) to `quasar.contrib`?

repl/src/main/scala/quasar/repl2/Paths.scala
```diff
@@ -0,0 +1,55 @@
...
+import scalaz._, Scalaz._
+import shims._
+
+object Paths {
+  val userHome = "user.home"
+
+  val tmpQuasarReplDirNames = NonEmptyList("tmp", "quasar-repl")
+  val quasarDirNames = NonEmptyList(".config", "quasar")
+  val quasarDataDirName = "data"
+  val quasarPluginsDirName = "plugins"
```
djspiewak: Minor note: it is more conventional to use `CamelCase` for constants.

rintcius:
[ch175]

repl/src/main/scala/quasar/repl2/Command.scala
```diff
@@ -0,0 +1,99 @@
...
+  private val SetPhaseFormatPattern        = "(?i)(?:set +)?phaseFormat *= *(tree|code)".r
+  private val SetTimingFormatPattern       = "(?i)(?:set +)?timingFormat *= *(tree|onlytotal)".r
+  private val DebugPattern                 = "(?i)(?:set +)?debug *= *(0|1|2)".r
+  private val SummaryCountPattern          = "(?i)(?:set +)?summaryCount *= *(\\d+)".r
+  private val FormatPattern                = "(?i)(?:set +)?format *= *((?:table)|(?:precise)|(?:readable)|(?:csv))".r
+  private val SetVarPattern                = "(?i)(?:set +)?(\\w+) *= *(.*\\S)".r
+  private val UnsetVarPattern              = "(?i)unset +(\\w+)".r
+  private val ListVarPattern               = "(?i)env".r
+  private val DataSourcesPattern           = "(?i)datasources".r
+  private val DataSourceTypesPattern       = "(?i)types".r
```
wemrysi: How about "supportedDatasources"? `types` seems too general, though "supportedDatasources" is a bit long.

Alternatively, we could prefix all the datasource operations with "ds", something like
* `ds` (list datasources)
* `dsTypes` (list supported datasource types)
* `dsAdd` (add a datasource)
* `dsRemove` (remove a datasource)
* `dsInspect` (details of a datasource)

repl/src/main/scala/quasar/repl2/Command.scala
```diff
@@ -0,0 +1,99 @@
...
+  final case class SetPhaseFormat(format: PhaseFormat) extends Command
+  final case class SetTimingFormat(format: TimingFormat) extends Command
+  final case class SetVar(name: VarName, value: VarValue) extends Command
+  final case class UnsetVar(name: VarName) extends Command
+
+  final case object DataSources extends Command
+  final case object DataSourceTypes extends Command
+  final case class DataSourceLookup(name: ResourceName) extends Command
+  final case class DataSourceAdd(name: ResourceName, tp: DataSourceType.Name, config: String, onConflict: ConflictResolution) extends Command
+  //final case class DataSourceMove
```
wemrysi: Can get rid of this, the "move" operation is actually going away soon.

repl/src/main/scala/quasar/repl2/Evaluator.scala
```diff
@@ -0,0 +1,329 @@
...
+          F.pure(s"Set debug level: $level".some)
+
+      case SummaryCount(rows) =>
+        val count: Option[Option[Int Refined Positive]] =
+          if (rows === 0) Some(None)
+          else refineV[Positive](rows).fold(κ(None), p => Some(Some(p)))
+        count match {
+          case None => F.pure("Rows must be a positive integer or 0 to indicate no limit".some)
+          case Some(c) => stateRef.modify(_.copy(summaryCount = c)) *>
+            F.pure(s"Set rows to show in result: $rows".some)
```
wemrysi: We should have a different message in the case where `rows === 0` indicating unlimited rows will be shown.

repl/src/main/scala/quasar/repl2/Evaluator.scala
```diff
@@ -0,0 +1,329 @@
...
+
+      case Cd(path: ReplPath) =>
+        for {
+          cwd <- stateRef.get.map(_.cwd)
+          dir = newPath(cwd, path)
+          _ <- ensureValidDir(dir)
+          _ <- stateRef.modify(_.copy(cwd = dir))
+        } yield s"cwd is now ${printPath(dir)}".some
+
+      case Ls(path: Option[ReplPath]) =>
```
wemrysi: Given that we're emulating a filesystem with our choice of commands (`cd`, `ls`), how about printing prefixes with a trailing `/` and resources without, like dirs and files?

repl/src/main/scala/quasar/repl2/Evaluator.scala
```diff
@@ -0,0 +1,329 @@
...
+      stateRef.get.map(_.supportedTypes) >>=
+        (_.map(_.point[F]).getOrElse(doSupportedTypes))
+
+    private def toADir(path: ResourcePath): ADir =
+      path.fold(f => fileParent(f) </> dir(fileName(f).value), rootDir)
+
+}
+
+object Evaluator {
+  //TODO change back to exitCode: Option[ExitCode] once we are back on
```
wemrysi: Maybe we use `fs2.StreamApp.ExitCode` in the meantime?

repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  implicit val ioQuasarError: MonadError_[IO, QuasarError] =
+    MonadError_.facet[IO](QuasarError.throwableP)
+
+  def quasarStream: Stream[IO, Quasar[IO, IO]] =
+    for {
+      basePath <- Paths.getBasePath[Stream[IO, ?]]
+      dataDir = basePath.resolve(Paths.quasarDataDirName)
+      _ <- Paths.mkdirs[Stream[IO, ?]](dataDir)
+      dataPath <- Stream.eval(IO(dataDir))
```
wemrysi: This line doesn't appear to be necessary.

repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+      pluginDir = basePath.resolve(Paths.quasarPluginsDirName)
+      _ <- Paths.mkdirs[Stream[IO, ?]](pluginDir)
+      pluginPath <- Stream.eval(IO(pluginDir))
```
wemrysi: Neither does this.

repl/src/main/scala/quasar/repl2/Main.scala
wemrysi: As I mention above, I think we can use `fs2.StreamApp` instead of `QuasarApp`, which happens to include an `ExitCode`.

repl/src/main/scala/quasar/repl2/Repl.scala
```diff
@@ -0,0 +1,114 @@
...
+import argonaut.Json
+import cats.effect._
+import fs2.Stream
+import fs2.async.Ref
+import org.apache.commons.io.FileUtils
+import org.jline.reader._
+import org.jline.terminal._
+import scalaz._, Scalaz._
+
+final class Repl[F[_]: Monad: ConcurrentEffect](
```
wemrysi: `ConcurrentEffect` is a `cats.Monad` so you should be able to drop the `scalaz.Monad` constraint. If you just need the syntax available for for-comprehensions, you can `import cats.syntax.flatMap._`, if you actually need a `scalaz.Monad` instance somewere, then you can `import shims._`.

repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import cats.effect.IO
+import fs2.Stream
+import fs2.async.Ref
+import scalaz._, Scalaz._
+import shims._
+
+object Main extends QuasarApp {
```
wemrysi: I don't think we need `QuasarApp`, you should be able to use `fs2.StreamApp`, get the quasar stream and then `quasarStream.flatMap(q => Stream.eval(repl(q))`.

repl/src/main/scala/quasar/repl2/Repl.scala
rintcius: ah, I thought I saw this convention being used, but am not able to find it now

repl/src/main/scala/quasar/repl2/Main.scala
rintcius: ah right, `fs2.StreamApp` actually integrates better with our use case than `IOApp` did. `QuasarApp` was a remnant of using `IOApp`.

Only small advantage of using `IOApp` is that `StreamApp` always does a `sys,exit` call, whereas `IOApp` only calls it in case of non-success exit code.
As a consequence, you now see a warning when exiting:

```
(v49.0.4) 💪 $ exit
Exiting...

Exception: sbt.TrapExitSecurityException thrown from the UncaughtExceptionHandler in thread "run-main-1"
[success] Total time: 15 s, completed Jul 13, 2018 12:44:37 PM
sbt:root>
```

repl/src/main/scala/quasar/repl2/Evaluator.scala
rintcius: So given that `StreamApp` integrates better than `IOApp` here and `StreamApp` still uses its own `ExitCode` in `fs2 series/1.0` branch, I think we don't have to switch back to `IOApp` and `cats-effect`'s `ExitCode` after upgrading `fs2` and `cats-effect`

repl/src/main/scala/quasar/repl2/Command.scala
rintcius: Yeah, not only `types` but also `add`, `rm` and `get` refer silently to datasources. I was hesitating about this but went for generic but short names for ease of use. 

The alternative I had in mind was to use a subcommand syntax with a prefix `ds`, which resembles your 2nd proposal. Advantage of subcommands is that it is still intuitive in combination with short names, eg `ds rm`. I have implemented that now.

repl/src/main/scala/quasar/repl2/Main.scala
wemrysi: That appears to be due to `sbt`, so it shouldn't show up when running via the `jar` or the launcher.

rintcius:

Still have to change the `testjar` procedure (see #3684 ). I merged in `master` so repl startup is tested now (but I think it will fail without an update)

repl/src/main/scala/quasar/repl2/Main.scala
```diff
@@ -21,47 +21,50 @@ import slamdata.Predef._
...
-import fs2.Stream
+import fs2.{Stream, StreamApp}, StreamApp.ExitCode
 import fs2.async.Ref
 import scalaz._, Scalaz._
 import shims._
 
-object Main extends QuasarApp {
+object Main {
+
+  private class ReplStreamApp(s: Stream[IO, ExitCode]) extends StreamApp[IO] {
```
wemrysi: As an aside, `StreamApp` is customarily used like
```scala
object Main extends StreamApp[IO] {
  def stream(args: List[String], requestShutdown: IO[Unit]) =
    ...
}
```

wemrysi:
@rintcius Looks great, I'll merge when the build is green!

PR #3684, opened 2018-07-11
add a testjar stage to travis for PRs

rintcius:


rintcius:
[ch481]

rintcius:
as suggested by @alissapajer here https://github.com/slamdata/quasar/pull/3683#issuecomment-404249213

.travis.yml
```diff
@@ -115,6 +117,19 @@ jobs:
             "testOnly -- xonly failtrace" \
             "exclusive:testOnly -- xonly failtrace"
 
+    - stage: testjar
```
rintcius: @djspiewak I opted for copying from `publish` stage, instead of making `scripts/quasarPublishAndTag` conditional __inside__ that script (and not having found a good way to add a conditional __outside__ that script).

djspiewak: `testJar` is already called from the publish stage. I must be missing something… why do we need to add a redundant stage here?

.travis.yml
```diff
@@ -115,6 +117,19 @@ jobs:
...
+      env:
+      script:
+        - set -e
```
djspiewak: Worth noting that, with this line, a failure in `testJar` will never actually fail the build. So this stage doesn't do anything.

.travis.yml
rintcius: in line 17 there's the opposite condition as line 20

.travis.yml
rintcius: Hmm, doesn't it just immediately exit with an error status?

```
  -e  Exit immediately if a command exits with a non-zero status.
```

.travis.yml
alissapajer: @djspiewak we don't run the publish stage in PR builds.

.travis.yml
djspiewak: You're correct. I think I had it backwards.

PR #3690, opened 2018-07-16
various IT improvements

rintcius:
Various improvements found during auditing ITs


rintcius:
relates to [ch402]

alissapajer:
@rintcius by the way, including a clubhouse story wrapped in square brackets in a PR comment like above will link this PR to that story regardless of the surrounding text. I've been referencing clubhouse stories without the brackets when I don't want the story to close along with the PR.

I moved ch402 back to develop state in clubhouse because I believe that was the intent?

rintcius:
@alissapajer Ah yes, that was the intent, thanks!

PR #3691, opened 2018-07-17
fix NPE converting from string to date/time in mimir

rintcius:
Also, in repl don't exit in case a bug is hit but show the exception instead

### DIFF
--- a/it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
+++ b/it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
@@ -1,0 +1,16 @@
+{
+    "name": "convert to datetime functions with invalid data",
+
+    "backends": {
+        "mimir": "pending"
+    },
+
+    "data": "../types.data",
+
+    "query": "select localdatetime(foo), date(foo), localdate(foo), time(foo), localtime(foo), offsetdatetime(foo), offsetdate(foo), offsettime(foo), timestamp(foo) from `../types.data`",
+
+    "predicate": "exactly",
+    "ignoreResultOrder": true,
+
+    "expected": []
+}
--- a/it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
+++ b/it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
@@ -9,10 +9,31 @@
 
     "data": "../types.data",
 
-    "query": "select localdatetime(foo), date(foo), localdate(foo), time(foo), localtime(foo), offsetdatetime(foo), offsetdate(foo), offsettime(foo), timestamp(foo) from `../types.data`",
+    "query": "select num, localdatetime(foo), date(foo), localdate(foo), time(foo), localtime(foo), offsetdatetime(foo), offsetdate(foo), offsettime(foo), timestamp(foo), interval(foo) from `../types.data`",
 
     "predicate": "exactly",
     "ignoreResultOrder": true,
 
-    "expected": []
+    "expected": [
+      { "num": 1 },
+      { "num": 2 },
+      { "num": 3 },
+      { "num": 4 },
+      { "num": 5 },
+      { "num": 6 },
+      { "num": 7 },
+      { "num": 8 },
+      { "num": 9 },
+      { "num": 10 },
+      { "num": 11 },
+      { "num": 12 },
+      { "num": 13 },
+      { "num": 14 },
+      { "num": 15 },
+      { "num": 16 },
+      { "num": 17 },
+      { "num": 18 },
+      { "num": 19 },
+      { "num": 20 }
+    ]
 }
--- a/it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
+++ b/it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
@@ -2,7 +2,9 @@
     "name": "convert to datetime functions with invalid data",
 
     "backends": {
-        "mimir": "pending"
+        "mimir": "pending",
+        "mongodb_3_4_13": "pending",
+        "mongodb_3_6": "pending"
     },
 
     "data": "../types.data",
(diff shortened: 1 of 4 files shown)


it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "convert to datetime functions with invalid data",
+
+    "backends": {
+        "mimir": "pending"
```
rintcius: note: this is pending in mimir because some of the functions don't return `undefined` where I think they should. E.g. 
`localdatetime(someLocalDatetime)` returns `someLocalDatetime`. This issue is tracked in ch500

it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "convert to datetime functions with invalid data",
+
+    "backends": {
+        "mimir": "pending",
```
rintcius: note: this is pending in mimir because some of the functions don't return undefined where I think they should. E.g.
localdatetime(someLocalDatetime) returns someLocalDatetime. This issue is tracked in ch500

alissapajer: :+1: we won't be reading any data out of mimir for `v0.1.0` so all that matters for now is that everything works for `lwc_local`.

djspiewak:
I'm a little confused. How are we getting `null`s? Like at all?

it/src/main/resources/tests/temporal/toDatetimeFunctionsInvalidData.test
alissapajer: Could we change this test query so that it returns an actual result? Maybe add a single selection in the query that selects something that is defined? Even though the test is testing that we don't crash with invalid data, it's possible that an empty result set could be received for some other reason, and this will limit the chance that the success is a false positive.

alissapajer:
That is a valid question though: what data would produce `null` and not a `DateTimeParseException`?

djspiewak:
@alissapajer I feel like it should only happen if we're calling `apply` on the results of the `CF1P` when the row is undefined.

alissapajer:
@rintcius my current thinking is that there might be an underlying problem causing us to get `null`s which we should address. For example what @djspiewak suggested above. So before just merging this, we should figure out where the `null`s are coming from.

rintcius:
@djspiewak @alissapajer 
Let me know if I'm misunderstanding you, but these are fixes in `isDefinedAt`, not in `apply`.  See e.g.
https://github.com/rintcius/quasar/blob/6bd6a22e5f4d73093b422d8364763e715f4e04fc/mimir/src/main/scala/quasar/mimir/TimeLib.scala#L195
Does that unconfuse you or is there still confusion left?

djspiewak:
@rintcius Oh, I see what's going on. This is the wrong way around though. In fact, this is actually going to yield incorrect results for `isDefinedAt`. That's not the fault of your fix; the function was just always wrong.

`isDefinedAt` here needs to delegating to `c.isDefinedAt(row)`. If the result from that function is `false`, then we shouldn't do the datetime conversion *at all*. This will avoid ever seeing `null` values, and will also correct this function for other cases.

For example, I'm relatively certain that the following query will result in incorrect results right now:

```sql
select localtime(select foo from bar where foo != "somedatetimestring")
```

Imagine real values there. The `"somedatetimestring"` result will magically *reappear* in the converted dataset. This bug, as well as the bug where we see `null` values at all, will be fixed by correctly delegating to `c.isDefinedAt(row)`. There may be a utility which does this for us (like `Op1SomethingOrOther`).

alissapajer:
@djspiewak the filter will be run before calling `localtime`, so those values will already have gone to undefined meaning that I think catching the `NullPointerException` would be enough. But that's a good test to add to our suite in any case :)

And I agree with @djspiewak that checking `c.isDefinedAt(row)` is the better way to solve this; else we'd have to catch `NullPointerExpection` in all of our mimir functions.

djspiewak:
@alissapajer Sending a value to undefined doesn't make it `null`. In this case, `c.isDefinedAt(row)` will be `false`, but `c(row)` will *not* be `null`, which is why the value will "reappear".

alissapajer:
@djspiewak oh I see what you're saying. Agreed. The `BitSet` filter _is_ what gives us the holes.

alissapajer:
@rintcius following up, I think that the query that @djspiewak suggested
```sql
select localtime(select foo from bar where foo != "somedatetimestring")
```
won't be doing the right thing, and that's a query that could conceivably be generated by the frontend. So we should fix this for `v0.1.0`.

rintcius:
@alissapajer indeed that's the same underlying issue to what I posted on discord

djspiewak:
BTW, we really need to fix the `isDefinedAt` issue on these functions before 0.10 if we expect them to be used at all. It's simply broken.

rintcius:
@djspiewak @alissapajer PR ready for re-review

As discussed with Daniel, I couldn't reproduce the incorrect result regarding `select localtime(select foo from bar where foo != "somedatetimestring")`

```
(v53.1.0) 💪 $ select offsetdatetime(select str from `days0.data` where str != "2014-08-17T06:00:00Z")
{ "$offsetdatetime": "2015-09-18T07:00:00.000000000Z" }
(v53.1.0) 💪 $ select * from `days0.data`
{ "num": 1, "str": "2014-08-17T06:00:00Z" }
{ "num": 2, "str": "not something dateish" }
{ "num": 3, "str": "2015-09-18T07:00:00Z" }
{ "num": 4 }
```

This is using https://github.com/slamdata/quasar/pull/3691 to get the results of the correct type when selecting date/times & on top of that a `catch NPE` instead of deferring to `c.isDefinedAt(r)`  (i.e. my first impl). The query shows the correct results though.

So I'm leaving that query out of this PR now (also because it needs #3691). The other comments I have processed, I think.

rintcius:
[ch493]

PR #3694, opened 2018-07-17
Add toRValue benchmarks

edmundnoble:
Actually very interesting results, which are so interesting I can't help but think that either the benchmark is broken, or `toRValue` is kind of broken.

```
Benchmark                                  (vectorLength)  Mode  Cnt    Score    Error  Units
ToRValuesBenchmarkVector.arraysToRValues                5  avgt    5    0.146 ±  0.008  ms/op
ToRValuesBenchmarkVector.arraysToRValues               20  avgt    5    0.300 ±  0.006  ms/op
ToRValuesBenchmarkVector.arraysToRValues               80  avgt    5    1.508 ±  0.035  ms/op
ToRValuesBenchmarkVector.arraysToRValues              320  avgt    5   18.392 ±  0.093  ms/op
ToRValuesBenchmarkVector.arraysToRValues             1280  avgt    5  326.650 ± 11.335  ms/op
ToRValuesBenchmarkVector.objectsToRValues               5  avgt    5    0.156 ±  0.007  ms/op
ToRValuesBenchmarkVector.objectsToRValues              20  avgt    5    0.254 ±  0.014  ms/op
ToRValuesBenchmarkVector.objectsToRValues              80  avgt    5    0.574 ±  0.023  ms/op
ToRValuesBenchmarkVector.objectsToRValues             320  avgt    5    2.132 ±  0.027  ms/op
ToRValuesBenchmarkVector.objectsToRValues            1280  avgt    5    8.947 ±  0.257  ms/op
ToRValuesBenchmarkScalar.scalarsToRValues                  avgt    5    0.138 ±  0.019  ms/op

```

### DIFF
--- a/yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/ToRValuesBenchmark.scala
+++ b/yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/ToRValuesBenchmark.scala
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar
+package yggdrasil
+package perf
+
+import quasar.contrib.cats.effect.liftio._
+import quasar.precog.common._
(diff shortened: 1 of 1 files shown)


djspiewak:
Given that the measured hot path is *also* including `fromRValuesStream`, could it be that we're seeing a bit of a performance pathology there as well? Isn't 1280 over the column limit?

edmundnoble:
1280 and 320 are both over the column limit, so yes. But the array/object difference doesn't exist in `fromRValuesStream`.

edmundnoble:
I believe this has to do with how we deal with arrays in `unsafeInsert`.

edmundnoble:
Yeah, this is definitely because we need full list copies to add elements to the end of arrays. That's obscenely slow.

djspiewak:
Yeah this is really weird. It actually lines up a lot better with what I would have expected for *both* benchmarks, but obviously it's only displayed here. Are we sure we're consuming things correctly in the `fromRValuesStream` benchmarks?

edmundnoble:
Fairly sure. `fromRValuesStream` doesn't have to incrementally construct the `List[RValue]` inside an `RArray`; it just produces a `List[Column]` from it.

PR #3696, opened 2018-07-17
Add mock tables with specs

alissapajer:


alissapajer:
This doesn't include tests for preparation because, without having more of the actual preparation implementation complete, I wasn't sure exactly how to mock out the various preparation states and still have a meaningful test. I'll come back to this when the actual preparation implementation is added.

alissapajer:
[ch452]

api/src/main/scala/quasar/api/table/Table.scala
```diff
@@ -16,8 +16,31 @@
...
+object TableName {
+  implicit val equalTableName: Equal[TableName] = Equal.equalA
+  implicit val showTableName: Show[TableName] = Show.showFromToString
+}
+
+object Table {
+  import TableName._
+
+  implicit def equalTable[Q: Equal]: Equal[Table[Q]] =
+    Equal.equal {
```
wemrysi: Alternatively, `Equal.equalBy(t => (t.name, t.query))`

api/src/main/scala/quasar/api/table/Table.scala
```diff
@@ -16,8 +16,31 @@
 
 package quasar.api.table
 
-import slamdata.Predef.String
+import scalaz.{Cord, Equal, Show}
+import scalaz.syntax.equal._
+import scalaz.syntax.show._
+import slamdata.Predef._
```
wemrysi: Minor, but we usually make `slamdata.Predef` the first import.

api/src/test/scala/quasar/api/table/MockTables.scala
```diff
@@ -0,0 +1,160 @@
...
+import quasar.Condition
+import quasar.contrib.scalaz.MonadState_
+import quasar.contrib.std.uuid._
+import scalaz.{\/, IMap, Monad}
+import scalaz.syntax.either._
+import scalaz.syntax.monad._
+import scalaz.syntax.std.option._
+import slamdata.Predef.{List, None, Option, Some, String, Unit}
+
+final class MockTables[F[_]: Monad: MockTables.TablesMockState[?[_]]]
```
wemrysi: Minor, no need for the `?[_]` when the type-constructor is the only type argument.

foundation/src/main/scala/quasar/contrib/std/uuid.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import scalaz.{Order, Ordering, Show}
+import scalaz.std.anyVal._
+import scalaz.syntax.equal._
+
+object uuid {
+
+  implicit val uuidOrder: Order[UUID] = new Order[UUID] {
+    def order(x: UUID, y: UUID): Ordering = {
+      val compare = x.compareTo(y)
```
wemrysi: Alternatively, `Ordering.fromInt(x.compareTo(y))`.

api/src/test/scala/quasar/api/table/TablesSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+
+abstract class TablesSpec[F[_]: Monad: Sync, G[_], I: Equal: Show, Q: Equal: Show, D]
+    extends Qspec
+    with ConditionMatchers {
+
+  import Table._
+
+  def tables: Tables[F, G, I, Q, D]
+
+  // `table1` and `table2` must have distinct names
```
wemrysi: Not necessary, but you could add an example that verifies this so the test will fail if they ever happen to be the same.

alissapajer: Good idea.

api/src/main/scala/quasar/api/table/Table.scala
alissapajer: All the imports in `quasar.api.table` were strictly alphabetized. I fixed them to comply with what we normally do in quasar.

alissapajer:
@wemrysi ready for another look (pending CI).

foundation/src/test/scala/quasar/QuasarSpecification.scala
```diff
@@ -55,6 +55,7 @@ trait QuasarSpecification extends AnyRef
 
   implicit class Specs2ScalazOps[A : Equal : Show](lhs: A) {
     def must_=(rhs: A) = lhs must equal(rhs)
+    def must_!=(rhs: A) = lhs must not(equal(rhs))
```
wemrysi: :+1:

PR #3698, opened 2018-07-18
[WIP] more RValue, less JValue

rintcius:
`Data` <-> `JValue` conversion is broken for date/time types, so taking the `Data` <-> `RValue` route.

Waiting for travis to get back to me (with fingers in my ears, because I won't be surprised if something exploded here and there).


djspiewak:
Why is this WIP?

rintcius:
@djspiewak waiting for travis feedback

rintcius:
@djspiewak `toRJsonElements` was semi-copied from `toJsonElements` above it. Happy to hear what you don't like about it.

djspiewak:
@rintcius I mean, the concept of the function in general is pretty horrible, but that's not your fault. :-) It's mostly the name that struck me as bizarre. I don't think it should hold anything up, and I don't have any particularly constructive suggestions to fix it.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -1656,6 +1656,19 @@ abstract class Slice { source =>
     rec(0, Vector())
   }
 
+  def toRJsonElements: Vector[RValue] = {
+    @tailrec def rec(i: Int, acc: Vector[RValue]): Vector[RValue] = {
```
alissapajer: Is this faster than folding?

djspiewak:
@rintcius I actually just realized that `Slice#toRValues` is a thing.

rintcius:
[ch568]

PR #3699, opened 2018-07-18
IndexedStore

wemrysi:
Introduces the `IndexedStore` abstraction and includes an implementation backed by `mimir`.

### DIFF
--- a/foundation/src/test/scala/quasar/EffectfulQSpec.scala
+++ b/foundation/src/test/scala/quasar/EffectfulQSpec.scala
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar
+
+import slamdata.Predef.{Either, String, Throwable}
+
+import scala.concurrent.ExecutionContext
+
+import cats.effect.{Effect, IO}
+import fs2.async.Promise
+import org.specs2.execute.AsResult
+import org.specs2.specification.core.Fragment
+
+abstract class EffectfulQSpec[F[_]: Effect](implicit ec: ExecutionContext) extends Qspec {
+  /** Provides syntax for defining effectful examples:
+    *
+    * "some example name" >>* {
+    *   <example body returning F[A]>
+    * }
+    */
+  implicit class RunExample(s: String) {
+    def >>*[A: AsResult](fa: => F[A]): Fragment = {
+      val run = for {
+        p <- Promise.empty[IO, Either[Throwable, A]]
+        _ <- Effect[F].runAsync(fa)(p.complete(_))
+        r <- p.get
+        a <- IO.fromEither(r)
+      } yield a
+
+      s >> run.unsafeRunSync
+    }
+  }
+}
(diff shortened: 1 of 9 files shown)


impl/src/main/scala/quasar/impl/storage/IndexedStore.scala
```diff
@@ -0,0 +1,144 @@
...
+
+          def insert(i: I, v: V): B[Unit] =
+            f(fa.insert(i, v))
+
+          def delete(i: I): B[Unit] =
+            f(fa.delete(i))
+        }
+    }
+
+  implicit def indexInvariantFunctor[F[_]: Functor, V]: InvariantFunctor[IndexedStore[F, ?, V]] =
```
djspiewak: Are you sure you want this? It's never going to be selected, given that the parameter is not right-most. It's also technically incoherent with the right-associated one below.

wemrysi:
[ch519]

mimir/src/main/scala/quasar/mimir/storage/MimirIndexedStore.scala
```diff
@@ -0,0 +1,150 @@
...
+      rpath <- Stream.emits(rpaths).covary[F]
+
+      rfile <- Stream.emit(Path.refineType(rpath).toOption).unNone.covary[F]
+
+      s = Path.fileName(rfile).value
+
+      v <- Stream.eval(lookup(s)).unNone
+    } yield (s, v)
+
+  def lookup(s: String): F[Option[RValue]] = {
```
djspiewak: Shouldn't we at least newtype the key?

mimir/src/main/scala/quasar/mimir/storage/MimirIndexedStore.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  import precog.trans.constants._
+
+  val entries: Stream[F, (String, RValue)] =
+    for {
+      rpaths <- Stream.eval(precog.vfs.ls(tablesPrefix).to[F])
+
+      rpath <- Stream.emits(rpaths).covary[F]
+
+      rfile <- Stream.emit(Path.refineType(rpath).toOption).unNone.covary[F]
```
djspiewak: I'm in favor of newlines, but, subjectively… I would collapse these three lines together.

mimir/src/test/scala/quasar/mimir/storage/MimirIndexedStoreSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+    } yield result
+
+    init.unsafeRunSync
+  }
+
+  val emptyStore =
+    IO(Random.nextInt(100000)) map { i =>
+      MimirIndexedStore(
+        P.unsafeValue,
+        rootDir </> dir("store") </> dir(i.toString))
```
djspiewak: I would propose that we standardize on using `.store` and similar for "meta" storage within the mimir vfs, just to help us keep it straight. The counter-argument here is that we don't really put things raw into mimir anymore, so maybe it's not necessary?

wemrysi: Doesn't matter to me, the production locations are specified in constants somewhere very close to `Main` and they're not exposed to any clients, so the specifics don't matter terribly, but :+1: to standardizing on something.

foundation/src/test/scala/quasar/EffectfulQSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+
+abstract class EffectfulQSpec[F[_]: Effect](implicit ec: ExecutionContext) extends Qspec {
+  /** Provides syntax for defining effectful examples:
+    *
+    * "some example name" >>* {
+    *   <example body returning F[A]>
+    * }
+    */
+  implicit class RunExample(s: String) {
+    def >>*[A: AsResult](fa: => F[A]): Fragment = {
```
alissapajer: :+1:

impl/src/main/scala/quasar/impl/storage/IndexedStore.scala
wemrysi: Yeah, I know it won't be selected normally, but good point about the incoherence. I'll just make it a function.

mimir/src/test/scala/quasar/mimir/storage/MimirIndexedStoreSpec.scala
djspiewak: On further thought, I'm fine with just `store`

PR #3701, opened 2018-07-18
Check to see if we can down-coerce BigDecimal when converting to CValue

djspiewak:


djspiewak:
[ch73]

alissapajer:
@djspiewak a couple integration tests failed legit in `lwc_local`.

djspiewak:
@alissapajer I'm glad they failed! We had a bug in `trunc` that was hidden by the fact that we never ever produced `DoubleColumn` in lwc_local.

PR #3702, opened 2018-07-18
Expose Tables in Quasar

alissapajer:


alissapajer:
@wemrysi to save on iteration time, is there anything else you can think of that I'll need to expose in quasar in order to access a `Tables` in sdbe?

alissapajer:
Going to just instantiate a `MockTables` directly in sdbe instead of getting handed it via a `Quasar`.

PR #3703, opened 2018-07-19
Implement ptable store

djspiewak:
This should allow us to actually store prepared tables in nihdb. Note that the dependent types are a little fiddly, especially within `for`-comprehensions.

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
+++ b/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.mimir.storage
+
+import slamdata.Predef._
+
+import quasar.contrib.fs2.convert._
+import quasar.contrib.pathy.{ADir, AFile}
+import quasar.mimir.MimirCake.Cake
+import quasar.yggdrasil.ExactSize
+import quasar.niflheim.NIHDB
+import quasar.yggdrasil.nihdb.NIHDBProjection
+
+import cats.arrow.FunctionK
+import cats.effect.{IO, LiftIO}
+
+import fs2.Stream
+
+import pathy.Path
+
+import scalaz.{Monad, OptionT}
+import scalaz.syntax.monad._
+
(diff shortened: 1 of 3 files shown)


djspiewak:
[ch590]

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
```diff
@@ -0,0 +1,115 @@
...
+    val cake: Cake,
+    tablesPrefix: ADir)(
+    implicit ME: MonadError_[F, ResourceError]) {
+
+  import cake.{Table => PTable}
+
+  // has overwrite semantics
+  def write(key: StoreKey, table: PTable): Stream[F, Unit] = {
+    val ios = Stream.bracket(cake.createDB(keyToFile(key)).map(_.toOption))({
+      case Some((blob, version, db)) =>
```
alissapajer: Can be `(_, _, db)`.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
```diff
@@ -0,0 +1,115 @@
...
+      table = PTable(proj.getBlockStream(None), ExactSize(proj.length))
+
+      back = table.transform(cake.trans.constants.SourceValue.Single)
+    } yield back
+
+    ioaOpt.run
+  }
+
+  def exists(key: StoreKey): F[Boolean] =
+    read(key).map(_.isDefined)
```
alissapajer: When the frontend is displaying all the tables that exist, they'll also be displaying the preparation status of each table. Those calls will go through here, to determine if the table is prepared. If it turns out that this is too slow (the user is waiting too long to see the list of their tables in the UI), we can consider populating a read-through cache on startup.

mimir/src/test/scala/quasar/mimir/storage/MimirPTableStoreSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+import scalaz.syntax.traverse._
+
+import shims._
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.Random
+
+import java.nio.file.Files
+
+object MimirPTableStoreSpec extends EffectfulQSpec[IO] {
```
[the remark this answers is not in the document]
alissapajer: Can you also add a test that reads a nonexistent key and errors?

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
djspiewak: Oh yeah, I guess it can. Woot.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
djspiewak: `read(key).map(_.isDefined)` doesn't load any of the table data. It's darn close to just checking to see if the directory exists in the VFS. I can make it use `fs.exists` instead if it's too slow.

alissapajer: Sounds good. We can keep it as is for now. If displaying the tables page is too slow we can switch over.

alissapajer:
@djspiewak if you see this green before I do feel free to merge.

PR #3704, opened 2018-07-19
add ITs for union all

rintcius:


rintcius:
[ch597]

it/src/main/resources/tests/union/union_all_1_field.test
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "[qa_s06] union all merging 1 field",
```
rintcius: relates to ch63 - Selecting a single field should return the result wrapped in an object

wanted to validate this explicitly

djspiewak:
@rintcius This was a WIP PR until a minute ago. Can you rebase please? These two commits should be one.

In general, WIP PRs should be avoided if at all possible. When they are used though (e.g. like in this case for getting Travis to run the tests), you should still treat it like a local branch and clean up the history prior to removing the WIP moniker. It's almost never correct to open a WIP PR and then remove the moniker without changing the commit set in some way.

rintcius:
@djspiewak Hmm, I like to have a clear indication that a PR is not for review yet. It's a waste of time if people start reviewing & making comments on things that are still worked on. I started using [WIP] to avoid that, also because it _is_ work in progress, just in a different sense. Anyway, happy to use some other indicator.

Also that's hardly an unclean commit to my taste, in the sense that it's full of those commits in this repos and when viewing the commit log it doesn't really matter whether it came from a WIP PR or not.

As an aside, I removed the [WIP] ahead of time (i.e. Travis still busy) because I went to sleep & thought it'd be good to have it reviewed despite chances of a failing Travis. That was solely because of the TZ difference though. More concretely, when you guys leave a review for the next morning basically no time is lost when you are in the same TZ, but in my case it means the PR delays by a complete day.

it/src/main/resources/tests/union/heterogeneous_union_all.test
```diff
@@ -0,0 +1,51 @@
...
+    { "num": 12, "b": 1.12345 },
+    { "num": 13, "b": "13" },
+    { "num": 14, "b": "zap" },
+    { "num": 15, "b": true },
+    { "num": 16, "b": false },
+    { "num": 17, "b": null },
+    { "num": 18, "b": { "$offsetdatetime": "2009-02-13T23:31:30.011000000+12:15:18" } },
+    { "num": 19, "b": { "$offsetdate": "2009-02-13+12:15:18" } },
+    { "num": 20, "b": { "$offsettime": "23:31:30.011000000+12:15:18" } },
+    { "b": { "$localdatetime": "2009-02-13T23:31:30.011000000" }, "num": 21 },
```
alissapajer: The `num` and `b` fields should be switched since field order is significant in sql.

it/src/main/resources/tests/union/union_all_1_field.test
alissapajer: Can you add a `"NB"` here that these results are what we currently expect but they do not follow the sql specification?

PR #3705, opened 2018-07-19
Doubled compiler stack size (to help avoid intermittent SOE in patmat)

djspiewak:


alissapajer:
Thanks, I hope this helps. We've been restarting easily 10+ jobs per day.

PR #3709, opened 2018-07-19
implement to_timestamp in mimir

rintcius:


frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -1509,6 +1509,13 @@ abstract class StdLibSpec extends Qspec {
             Data.Int(BigInt(1234567890011L)),
             Data.OffsetDateTime(JOffsetDateTime.parse("2009-02-13T23:31:30.011Z")))
         }
+
+        "arbitrary long" >> prop { (l : Long) =>
```
djspiewak: `(l: Long) =>`

mimir/src/main/scala/quasar/mimir/TimeLib.scala
```diff
@@ -359,6 +359,30 @@ trait TimeLibModule extends ColumnarTableLibModule {
     val TruncWeek = Trunc(TemporalPart.Week)
     val TruncYear = Trunc(TemporalPart.Year)
 
+    val ToTimestamp = new Op1F1 {
+      def f1: F1 = CF1P {
+        case c: DoubleColumn =>
```
rintcius: I am not clear on how the unified `number` type exactly interacts with the code here in some cases. In particular:

- is it correct that these 3 columns need to be implemented? `DoubleColumn`, `LongColumn` and `NumColumn`. I think I have to, but it also seems a bit weird because the domain is `Int -> OffsetDateTime` in `frontend`.
- The point above raises the question how to deal with non integer numbers in `DoubleColumn` and `NumColumn`: rounding (like I did now using `.longValue`) or overriding `isDefinedAt` and return false there for non integers?
- I haven't implemented this yet, but I think `DoubleColumn` also needs to include a check based on `Stdlib.doubleIsDefined` in `isDefinedAt`?

rintcius:
[ch574]

mimir/src/main/scala/quasar/mimir/TimeLib.scala
```diff
@@ -359,6 +359,30 @@ trait TimeLibModule extends ColumnarTableLibModule {
...
+          }
+        case c: NumColumn =>
+          new Map1Column(c) with OffsetDateTimeColumn {
+            def apply(row: Int) = {
+              time.epochMilliToOffsetDateTime(c(row).longValue)
+            }
+          }
+      }
+      val tpe = UnaryOperationType(JNumberT, JOffsetDateTimeT)
+    }
```
djspiewak: This isn't quite right. First off, it needs to be undefined if `c(row) < 0`. Second, it should only be defined for integer values. So that requires an additional check in `DoubleColumn` and `NumColumn`, at least in the `isDefinedAt` implementation (the `apply` is fine). Note that if you override `isDefinedAt` make sure to delegate to `super` as well, since that's where it inherits column definedness.

rintcius: regarding undefined if `c(row) < 0`, why? isn't that unnecessarily limiting? Also isn't that contradicting the spec here (which was already there)? https://github.com/rintcius/quasar/blob/5da4e539d4fd325c35e81a972565bfd558ac4b83/frontend/src/test/scala/quasar/std/StdLibSpec.scala#L1499-L1504

regarding 2nd: yeah, those were basically my main questions above. But can you confirm (or not) my last point there too?

alissapajer: At this point it's up to us to decide if the existing tests are correct and/or what we want to do in less rigorously defined situations like this one. In this case I think supporting negative millis values is the way to go.

djspiewak: I agree. It just initially strikes me as weird. I think it's the correct thing to do though.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
alissapajer: This test is failing for mongo. We should disable it for mongo as we don't need it working now.

alissapajer:
@djspiewak since you did the initial review on this, I'm bumping it over to you for a second look.

PR #3710, opened 2018-07-20
implement to_string for date/times in mimir

rintcius:


rintcius:
[ch635]

mimir/src/main/scala/quasar/mimir/StringLib.scala
```diff
@@ -132,6 +132,41 @@ trait StringLibModule extends ColumnarTableLibModule {
...
+        case c: LocalDateColumn =>
+          new StrColumn {
+            def apply(row: Int) = c(row).toString
+            def isDefinedAt(row: Int) = c.isDefinedAt(row)
+          }
+        case c: IntervalColumn =>
+          new StrColumn {
+            def apply(row: Int) = c(row).toString
+            def isDefinedAt(row: Int) = c.isDefinedAt(row)
+          }
```
djspiewak: Each of the above can be `new Map1Column(c) with StrColumn`, I think. Feel like doing that in this PR, or sometime in the future? If the latter, then I'm 👍 for merge.

PR #3712, opened 2018-07-20
quasar.api reorganization

wemrysi:
Reorganizes some of the operations in `quasar.api`, namely
* Removes `ResourceDiscovery`, moving discovery operations to `Datasources`
* Adds discovery methods to `Datasource`
* Reduces to a single effect type in `{Light,Heavy}WeightDatasourceModule`
* Datasources are no longer identified by name, instead an opaque identifier is generated when one is added
* Moves all the `Resource*` types to `quasar.common.resource`
* Expands `ResourcePathType` to allow for paths that are both a prefix and a resource


wemrysi:
[ch408]
[ch504]

alissapajer:
@wemrysi due to the large diff and the shortage of time, are there a couple places you could put comments to point reviewers toward the more semantically important stuff, like the opaque identifiers, for example?

wemrysi:
@alissapajer Sure thing.

api/src/main/scala/quasar/api/datasource/Datasources.scala
```diff
@@ -16,52 +16,58 @@
...
-import scalaz.{\/, IMap, ISet}
+import scalaz.{\/, ISet}
 
-trait Datasources[F[_], C] {
+/** @tparam F effects
+  * @tparam G multple results
+  * @tparam I identity
+  * @tparam C configuration
+  */
+trait Datasources[F[_], G[_], I, C] {
```
wemrysi: Here is where the opaque identifier is introduced for `Datsources` as well as the addition of the discovery methods, lots of fallout over in `quasar.impl.datasources` due to this.

connector/src/main/scala/quasar/connector/HeavyweightDatasourceModule.scala
```diff
@@ -33,8 +33,7 @@ trait HeavyweightDatasourceModule {
 
   def heavyweightDatasource[
       T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT,
-      F[_]: ConcurrentEffect: PlannerErrorME: Timer,
-      G[_]: ConcurrentEffect: Timer](
+      F[_]: ConcurrentEffect: PlannerErrorME: Timer](
       config: Json)
-      : F[InitializationError[Json] \/ Datasource[F, Stream[G, ?], T[QScriptEducated[T, ?]], Stream[G, Data]]]
+      : F[InitializationError[Json] \/ Disposable[F, Datasource[F, Stream[F, ?], T[QScriptEducated[T, ?]], Stream[F, Data]]]]
```
wemrysi: Here and over in the lightweight variant, the shutdown effect is now external to `Datasource`, look ma, representable state!

connector/src/test/scala/quasar/connector/DatasourceSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+abstract class DatasourceSpec[F[_]: Effect, G[_]]
+    extends EffectfulQSpec[F] {
+
+  def datasource: Datasource[F, G, _, _]
+
+  def nonExistentPath: ResourcePath
+
+  def gatherMultiple[A](fga: G[A]): F[List[A]]
+
+  "resource discovery" >> {
```
wemrysi: These were previously housed in `ResourceDiscovery` spec, no semantic changes, just relocated.

impl/src/main/scala/quasar/impl/datasource/ByNeedDatasource.scala
```diff
@@ -121,17 +98,24 @@ object ByNeedDatasource {
   object NeedState {
     final case class Uninitialized[F[_], A](init: F[A]) extends NeedState[F, A]
     final case class Initialized[F[_], A](a: A) extends NeedState[F, A]
-    final case class Shutdown[F[_], A]() extends NeedState[F, A]
```
wemrysi: :tada:

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
```diff
@@ -16,121 +16,147 @@
...
+import cats.effect.Sync
+import fs2.Stream
+import scalaz.{\/, -\/, \/-, EitherT, Equal, ISet, OptionT}
+import scalaz.syntax.either._
 import scalaz.syntax.equal._
 import scalaz.syntax.monad._
+import scalaz.syntax.std.boolean._
+import shims._
+
+final class DefaultDatasources[F[_]: Sync, I: Equal, C: Equal] private (
```
wemrysi: The bulk of the semantic changes are here as this is our `Datasources` implementation.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -43,30 +46,49 @@ import pathy.Path._
 import scalaz._, Scalaz._
 import shims._
 
-final class Evaluator[F[_]: Effect, G[_]: Effect](
-  stateRef: Ref[F, ReplState],
-  sources: Datasources[F, Json],
-  queryEvaluator: QueryEvaluator[F, Stream[G, ?], SqlQuery, Stream[G, Data]]) {
+final class Evaluator[F[_]: Effect](
```
wemrysi: Decent amount of changes here as well, mostly around mimicking a filesystem on top of `Datasources`.

run/src/main/scala/quasar/run/ResourceRouter.scala
```diff
@@ -0,0 +1,68 @@
...
+import cats.effect.{Effect, IO}
+import cats.syntax.applicative._
+import cats.syntax.functor._
+import fs2.Stream
+import scalaz.~>
+import scalaz.std.option._
+import scalaz.syntax.foldable._
+
+/** Translates resource paths from queries into actual sources. */
+object ResourceRouter {
```
wemrysi: This is new and handles routing paths that look like `/datasource/<uuid>/path/local/to/datasource` to the appropriate `Datasource`.

connector/src/main/scala/quasar/connector/HeavyweightDatasourceModule.scala
djspiewak: I still wish we could do this with `Stream` rather than `Disposable`, since that would actually prevent improper usage, but one step at a time!

foundation/src/main/scala/quasar/contrib/cats/effect/effect.scala
```diff
@@ -0,0 +1,37 @@
...
+  def unsafeRunEffect[F[_]: Effect, A](fa: F[A])(implicit ec: ExecutionContext): A = {
+    val ioa = for {
+      p <- Promise.empty[IO, Either[Throwable, A]]
+      _ <- Effect[F].runAsync(fa)(p.complete(_))
+      r <- p.get
+      a <- IO.fromEither(r)
+    } yield a
+
+    ioa.unsafeRunSync()
+  }
```
djspiewak: You could also do `fa.to[IO].unsafeRunSync()`.

impl/src/main/scala/quasar/impl/datasource/ByNeedDatasource.scala
```diff
@@ -35,47 +35,30 @@ import cats.syntax.functor._
...
 import fs2.async
 import fs2.async.mutable
-import scalaz.\/
 
 import ByNeedDatasource.NeedState
 
 final class ByNeedDatasource[F[_], G[_], Q, R] private (
     datasourceType: DatasourceType,
-    mvar: mutable.Queue[F, NeedState[F, Datasource[F, G, Q, R]]],
+    mvar: mutable.Queue[F, NeedState[F, Disposable[F, Datasource[F, G, Q, R]]]],
```
djspiewak: It's called `mvar` but it's a queue? Maybe obvious question, but why is it a `Queue` rather than an `MVar` or a `Signal`?

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasource.scala
```diff
@@ -52,90 +50,65 @@ import shims._
   * @param root the scope of this datasource, all paths will be considered relative to this one.
   * @param readChunkSizeBytes the number of bytes per chunk to use when reading files.
   */
-final class LocalDatasource[F[_]: Sync, G[_]: Effect: Timer] private (
+final class LocalDatasource[F[_]: Timer] private (
     root: JPath,
     readChunkSizeBytes: Int,
-    pool: ExecutionContext)
-    extends LightweightDatasource[F, Stream[G, ?], Stream[G, Data]] {
+    pool: ExecutionContext)(
```
djspiewak: Why isn't this implicit?

alissapajer:
Decided this was ready for merge. If there is any follow up review it can be addressed in a separate PR.

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
```diff
@@ -16,121 +16,147 @@
...
+    control: DatasourceControl[F, Stream[F, ?], I, C])
+    extends Datasources[F, Stream[F, ?], I, C] {
+
+  def addDatasource(ref: DatasourceRef[C]): F[CreateError[C] \/ I] =
+    for {
+      i <- freshId
+      c <- addRef[CreateError[C]](i, ref)
+    } yield Condition.disjunctionIso.get(c).as(i)
+
+  def allDatasourceMetadata: F[Stream[F, (I, DatasourceMeta)]] =
```
djspiewak: Why isn't this type signature just `Stream[F, (I, DatasourceMeta)]`? Is the outer effect significant in any context?

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
```diff
@@ -16,121 +16,147 @@
...
-    def checkSrc: EitherT[F, ExistentialError, Unit] =
-      EitherT(lookupConfig(src)).leftMap(commonIsExistential).void
+  /** Add the ref at the specified id, replacing any running datasource. */
+  private def addRef[E >: CreateError[C] <: DatasourceError[I, C]](
+      i: I, ref: DatasourceRef[C])
+      : F[Condition[E]] = {
 
-    def renamed: EitherT[F, ExistentialError, Unit] = for {
-      _ <- checkSrc
+    type T[X[_], A] = EitherT[X, E, A]
```
djspiewak: I personally prefer the convention `MT` (for the name), when defining the lift version of the `M` type. This keeps them tied together and avoids confusion with `T[_[_]]` in other parts of the code.

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
```diff
@@ -16,121 +16,147 @@
...
-    OptionT(configs.lookup(name))
-      .toRight(DatasourceNotFound(name): CommonError)
-      .run
+  private def verifyNameUnique[E >: CreateError[C] <: DatasourceError[I, C]](
+      name: DatasourceName,
+      currentId: I)
+      : F[E \/ Unit] =
+    refs.entries
+      .exists(t => t._2.name === name && t._1 =/= currentId)
+      .compile.fold(false)(_ || _)
```
djspiewak: Any idea if it's better to `fold` after `compile` or before? Both are an option and I don't have any intuition as to which is more correct in which context.

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
```diff
@@ -102,17 +100,15 @@ final class FederatedShiftedReadPlanner[
...
 
   private def tableFromStream(s: Stream[IO, Data]): F[P.Table] = {
     val dataToRValue: Data => RValue =
       d => RValue.fromData(d).getOrElse(sys.error("There is no representation of CUndefined in SlamDB as a value"))
 
-    for {
-      // TODO{fs2}: Chunkiness
-      table <- P.Table.fromRValueStream[F](s.mapChunks(_.map(dataToRValue).toSegment))
-    } yield {
+    // TODO{fs2}: Chunkiness
```
djspiewak: I think we can get rid of this todo, right?

repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -62,18 +63,18 @@ final class Repl[F[_]: ConcurrentEffect](
...
     new Repl[F](prompt, reader, evaluator)
 
-  def mk[F[_]: ConcurrentEffect, G[_]: Effect](
-    ref: Ref[F, ReplState],
-    datasources: Datasources[F, Json],
-    queryEvaluator: QueryEvaluator[F, Stream[G, ?], SqlQuery, Stream[G, Data]])
+  def mk[F[_]: ConcurrentEffect](
+      ref: Ref[F, ReplState],
+      datasources: Datasources[F, Stream[F, ?], UUID, Json],
+      queryEvaluator: QueryEvaluator[F, SqlQuery, Stream[F, Data]])
```
djspiewak: 🎉

api/src/main/scala/quasar/api/datasource/DatasourceError.scala
```diff
@@ -17,105 +17,152 @@
...
 
+import monocle.Prism
 import scalaz.{Cord, Equal, ISet, NonEmptyList, Show}
 import scalaz.std.option._
 import scalaz.std.string._
 import scalaz.std.tuple._
 import scalaz.syntax.show._
 
-sealed trait DatasourceError[+C] extends Product with Serializable
+sealed trait DatasourceError[+I, +C] extends Product with Serializable
```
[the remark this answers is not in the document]
wemrysi: It allows us to build a fairly precise error hierarchy, the sum of which includes both the identity and the config, but not all cases refer to both of them.

connector/src/main/scala/quasar/connector/HeavyweightDatasourceModule.scala
wemrysi: Agreed, I'm just not sure how to do so yet given the ad-hoc nature of the scope of the resource.

impl/src/main/scala/quasar/impl/datasource/ByNeedDatasource.scala
wemrysi: It was an `MVar` when we were using `cats-effect-1.x`, but that doesn't exist in the downgraded version. It really is an `MVar`, if you squint a bit.

I don't think I can get the exclusivity semantics of `take` from a signal (without some extra work at least), so I choose the `Queue` instead, inspired from the actual `MVar` docs that describe it as essentially a singleton queue :smiley:

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasource.scala
wemrysi: I'm not a fan of implicit pools at this "module" level, feel free to convince me otherwise.

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
wemrysi: Yeah, I'm not sure either, `exists` only emits one value so maybe not a big deal in this instance? And if you `fold` before, you still have a partial result and have to provide an alternative.

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
wemrysi: This is an implementation class where we've refined the result type to `Stream[F, ?]` it is abstract in the interface.

mimir/src/main/scala/quasar/mimir/evaluate/FederatedShiftedReadPlanner.scala
wemrysi: This `TODO` appears everywhere I've had to add non-obvious things to preserve chunks, so as to make them easy to find. Maybe they need to be more descriptive towards that end :smile:.

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasource.scala
djspiewak: Well, it's implicit down below. :-) Basically, it comes down to whether or not `ExecutionContext` is a valid use of implicits *at all*, and I would argue the answer is "yes". The benefits outweigh the harm, in my view, since there is a definitively "default" pool in all circumstances, and you probably want to be forced to be extremely explicit if you're not using that default. Given the number of operations which involve pool threading (directly or indirectly, such as via `Timer`), it becomes syntactically really bulky if you don't thread things implicitly.

wemrysi: Yeah, I hear you, and I think the implicit threading on methods/functions is reasonable for the reasons you've cited. I feel more confident knowing I have to be explicit when instantiating something that requires a threadpool as at least then I can make the choice to use the "default" pool or an explicit one. Maybe at this level it isn't worth it, but what about at the level of, say, `quasar.run.Quasar`?

My goal was to make sure we're considering where we're running things, not to add unnecessary boilerplate, so if there's a better way to do that then :+1:

PR #3713, opened 2018-07-20
Remove G[_] from Tables and update signature of allTables

alissapajer:


alissapajer:
@djspiewak @wemrysi assigned this to both of you. Whoever gets there first can have it. :racing_car:

alissapajer:
#3712 doesn't touch anything in `quasar.api.table` so I didn't base on that. (@wemrysi correct me if I'm wrong, I didn't see any diff there...)

PR #3717, opened 2018-07-23
PreparationsManager

djspiewak:
A `JobManager` by any other name…

### DIFF
--- a/impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
+++ b/impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
@@ -27,7 +27,7 @@ import fs2.{async, Stream}
 import fs2.async.mutable.Queue
 
 // monad/traverse syntax conflict
-import scalaz.{Equal, OptionT, Scalaz}, Scalaz._
+import scalaz.{-\/, \/-, \/, Equal, OptionT, Scalaz}, Scalaz._
 
 import shims._
 
@@ -47,11 +47,8 @@ class PreparationsManager[F[_]: Effect, I, Q, R] private (
 
   import PreparationsManager._
 
-  private val pending: ConcurrentHashMap[I, F[Unit]] =
-    new ConcurrentHashMap[I, F[Unit]]
-
-  private val ongoing: ConcurrentHashMap[I, Preparation[F]] =
-    new ConcurrentHashMap[I, Preparation[F]]
+  private val status: ConcurrentHashMap[I, F[Unit] \/ Preparation[F]] =
+    new ConcurrentHashMap[I, F[Unit] \/ Preparation[F]]
 
   val F = Effect[F]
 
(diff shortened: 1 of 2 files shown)


impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
```diff
@@ -0,0 +1,259 @@
...
+class PreparationsManager[F[_]: Effect, I, Q, R] private (
+    evaluator: QueryEvaluator[F, Q, R],
+    notificationsQ: Queue[F, Option[PreparationsManager.TableEvent[I]]],
+    background: Stream[F, Nothing] => F[Unit])(
+    runToStore: (I, R) => F[Stream[F, Unit]])(
+    implicit ec: ExecutionContext) {
+
+  import PreparationsManager._
+
+  private val pending: ConcurrentHashMap[I, F[Unit]] =
```
wemrysi: Would a `state: ConcurrentHashMap[I, Preparation[F] \/ F[Unit]]` (or equivalent), work instead? If so, this would eliminate the chance of being in neither state and simplify some of the other code.

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
```diff
@@ -0,0 +1,259 @@
...
+    new ConcurrentHashMap[I, F[Unit]]
+
+  private val ongoing: ConcurrentHashMap[I, Preparation[F]] =
+    new ConcurrentHashMap[I, Preparation[F]]
+
+  val F = Effect[F]
+
+  val notifications = notificationsQ.dequeue.unNoneTerminate
```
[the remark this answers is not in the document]
djspiewak: Good call.

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
djspiewak: As far as I can tell, it's only possible to be in *neither* state if the job has never actually been submitted. Things move to `ongoing` before they are removed from `pending`. With that said, I'll have to think about whether or not this actually covers things. The state management here is tricky by definition simply because a preparation is not "started" until the stream which evaluates and pipes it to disk is running, and that is done via the parallel worker.

djspiewak:
[ch651]

djspiewak:
@alissapajer I think this is actually set, pending re-review from @wemrysi. Tests are, I believe, complete.

alissapajer:
@djspiewak awesome. I won't get to it tonight but I'll look tomorrow morning.

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
```diff
@@ -0,0 +1,220 @@
...
+  @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+  def cancelAll(implicit I: Equal[I]): F[Unit] = {
+    for {
+      keys <- F.delay(status.keys.asScala.toList)
+      results <- keys.traverse(cancelPreparation)
+
+      _ <- if (results.isEmpty)
+        ().point[F]
+      else
+        cancelAll   // handle race condition of new enqueue
```
wemrysi: I wonder if this is actually needed? It isn't possible to "cancel all current and future preparations" if one is still allowed to initiate new ones so this method always operates in a "snapshot" fashion (i.e. it cancels the preparations in existence at some distinguished point in time, roughly the start of this method).

In the pathological case, we could blackhole a series of rapidly created preparations if they happened at just the right time when another actor initiated a `cancelAll`.

impl/src/test/scala/quasar/impl/table/PreparationsManagerSpec.scala
```diff
@@ -0,0 +1,362 @@
...
+        case Some(Status.Started(_)) => ok
+      }
+    }
+
+    "return Status.Pending before evaluation" in {
+      val Id = "table-id"
+
+      val results = for {
+        manager <- PreparationsManager[IO, String, PreparationsManager[IO, String, _, Unit], Unit](
+          { manager =>    // is this not clever? (actually it's probably really stupid; feel free to say so)
```
wemrysi: :clap:

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
djspiewak: I guess I kind of envisioned a guarantee around `cancelAll` where, at some snapshot in time after it is invoked, there are no table ids such that `preparationStatus` is *not* `Unknown`. Without the recursion, we don't have that guarantee.

impl/src/test/scala/quasar/impl/table/PreparationsManagerSpec.scala
```diff
@@ -0,0 +1,362 @@
...
+      } yield event
+
+      val event = results.compile.last.unsafeRunSync
+
+      event must beLike {
+        case Some(TableEvent.PreparationErrored(_, _, _, TestException)) => ok
+      }
+    }
+
+    // TODO this is flaky
```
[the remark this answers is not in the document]
djspiewak:
```scala
// TODO this comment is obsolete
```

PR #3718, opened 2018-07-23
Removed build stages, cleaned things up, presto!

djspiewak:


djspiewak:
[ch674]

alissapajer:
@djspiewak do you know what work needs to be done to land this? Or are there still some things to figure out?

djspiewak:
@alissapajer Travis is being inexplicably weird. It's like it's timing out while the build is running and I don't get it. One possibility is that the log buffer is too large, which I've seen cause issues with Travis in the past. Not sure exactly though.

.travis.yml
```diff
@@ -36,80 +24,29 @@ before_script:
...
+    ./sbt -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION \
+      doc \
+      repl/assembly
 
-    - stage: cache cleanup
-      env:
-      script:
-        - travis_wait 30 scripts/s3Cache cleanup
+  - scripts/testJar
+  - '[ $TRAVIS_PULL_REQUEST == "false" ] && [[ "$TRAVIS_BRANCH" =~ backport/v.*|master ]] && scripts/quasarPublishAndTag'
```
alissapajer: We want to run `scripts/testJar` for both PR builds and master builds (this is how it was previously).

alissapajer: Oh, we are. I mis-parsed the above lines.

.travis.yml
```diff
@@ -36,80 +24,29 @@ before_script:
   - nproc 2>/dev/null && python -c 'import multiprocessing;print multiprocessing.cpu_count()'
   - free && cat /proc/meminfo
 
-jobs:
-  include:
-    - stage: clean
-      env:
-      script:
-        - travis_wait 30 ./scripts/s3Cache cacheCleanUp
-        - ./sbt -DisIsolatedEnv=${ISOLATED_ENV:=false} clean
```
alissapajer: @djspiewak did you purposefully omit this line? We are not running `sbt clean` anymore.

djspiewak: There's no need to run `sbt clean` since the filesystem comes to us pre-cleaned.

.travis.yml
```diff
@@ -36,80 +24,29 @@ before_script:
...
+    ./sbt -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION \
+      checkHeaders \
+      test:compile
 
-        # release to sonatype
-        - scripts/quasarPublishAndTag
+  - |-
+    ./sbt -DisIsolatedEnv=${ISOLATED_ENV:=false} ++$TRAVIS_SCALA_VERSION \
+      "testOnly -- failtrace" \
+      "exclusive:testOnly -- failtrace"
```
alissapajer: `exclusive` ensures that this task is run in isolation. But how is running the entirety of `testOnly` in isolation different than just `testOnly`? The spec doesn't seem to imply that tasks _within_ the provided task are run in isolation.

.travis.yml
```diff
@@ -1,22 +1,10 @@
...
-  - name: compile
-  - name: test
-  - name: testjar
-    if: type = pull_request
-  - name: publish
-    # weirdly, we have to add the extra PR check
-    if: NOT type = pull_request
-  - name: cache cleanup
+dist: trusty
+sudo: false
```
alissapajer: This line seems potentially suspect. @djspiewak can you try toggling this back to `required`? This change switched us from [sudo-enabled](https://docs.travis-ci.com/user/reference/overview/#sudo-enabled) to [container-based](https://docs.travis-ci.com/user/reference/overview/#container-based). I'm wondering if there are other (not directly `sudo`-related environment changes that affect our tests.

djspiewak: OMG this is almost certainly the exact issue.

.travis.yml
djspiewak: I really don't know, honestly. I cargo-culted these lines from the previous iteration of the build.

alissapajer: @wemrysi I traced our use of `exclusive` back to [here](https://github.com/slamdata/quasar/commit/17a92bf7bb0ebd8b7a59f5b777f74922b2da534c#diff-596d2114d55ed0d2d1f87659c57d700eR56). I'm a little confused though because we currently have `it/test | testOnly | testQuick` tagged with `ExclusiveTests`, but we no longer have an [`ExclusiveExecution`](https://github.com/slamdata/quasar/commit/17a92bf7bb0ebd8b7a59f5b777f74922b2da534c#diff-ca09976c39cd4ea09951f3ed1350b323), so I'm not convinced that any of this is actually tied together anymore.

.travis.yml
alissapajer: :crossed_fingers:

.travis.yml
alissapajer: Also, proof that `exclusive:testOnly` is a no-op is that, as far as I can tell from the logs, each test is run only once on travis.

.travis.yml
alissapajer: We made it through tests! Building docs and stuff now. Hooray!

.travis.yml
wemrysi: @alissapajer agreed, the `exclusive:_` should have been removed along with the trait.

alissapajer: And my follow-up would be: I think we can get rid of `ExclusiveTest` in `build.sbt`, unless there is some usage connection I missed.

alissapajer:
@djspiewak travis just failed for a totally different reason.

.travis.yml
alissapajer: @wemrysi cool, thanks for the confirmation!

@djspiewak I'd vote to remove `exclusive` and friends in a separate PR (after this one lands) so as not to conflate concerns.

.travis.yml
alissapajer: @djspiewak when `[ $TRAVIS_PULL_REQUEST == "false" ]` evaluates to `false`, it evaluates to `1` and so `1` is returned, killing the travis build.

djspiewak: Yep, I just figured it out. It's fixed by adjusting the `set +e` location.

alissapajer: Ha, you pushed the commit as I was typing this :)

.travis.yml
```diff
@@ -27,26 +27,19 @@ before_script:
 script:
   - set -e
 
-  # we split the following across multiple sbt instances because it makes travis happier
```
alissapajer: :+1:

alissapajer:
@djspiewak we made it further. This time
```
Unable to send webhook.
```
when we were trying to send the webhook to discord.

PR #3721, opened 2018-07-24
Fix IT's & add ITs for flatten/shift

rintcius:


rintcius:
[ch665]
[ch71]
[ch402]

rintcius:
@alissapajer ready for review, still some tests to process (the "to be analyzed" ones that you commented on) but was planning to do that in a separate PR (hence ch402 is in develop)

alissapajer:
:+1: review is in progress for this. Didn't finish it tonight and want to make sure to think through the queries carefully enough. Will continue tomorrow.

it/src/main/resources/tests/joins/symmetricNonJsJoinCondition.test
```diff
@@ -2,8 +2,8 @@
     "name": "[qa_s07] flattening both sides of a join condition",
 
     "backends": {
-        "lwc_local":         "pending",
-        "mimir":             "pending"
+        "lwc_local": "ignoreFieldOrder",
+        "mimir":     "pending"
```
alissapajer: `ignoreFieldOrder`.

it/src/main/resources/tests/nestedSelectWithFilter.test
```diff
@@ -1,22 +1,22 @@
 {
     "name": "nested select with filter",
     "backends": {
-        "lwc_local":         "pending",
-        "mimir":             "pending"
+        "lwc_local": "ignoreFieldOrder",
+        "mimir":     "pendingIgnoreFieldOrder"
```
alissapajer: `ignoreFieldOrder`.

it/src/main/resources/tests/nestedSelectWithFilter.test
```diff
@@ -1,22 +1,22 @@
...
     },
-    "data": "olympics.data",
+    "data": "demo/olympics.data",
```
alissapajer: Are both `olympics.data` and `demo/olympics.data` still in use? If not, we can remove `olympics.data`. If it's still in use, we should remove it, but that can be done in a separate PR.

it/src/main/resources/tests/demo/nestedreductionsonnull.test
```diff
@@ -1,17 +1,17 @@
 {
-  "name": "nestedreductionsonnull",
+  "name": "nested reductions on null",
   "data": "patients.data",
   "backends": {
-    "lwc_local":         "pending",
+    "lwc_local":         "ignoreFieldOrder",
     "mimir":             "pending"
```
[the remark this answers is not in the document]
rintcius: ok, yeah I didn't want to change `mimir` because we don't get feedback on whether the change is right or not, but I do like aiming for the "best" status that should work (`ignoreFieldOrder` here) so that we can consider failures as a bug.

it/src/main/resources/tests/nestedSelectWithFilter.test
rintcius: Yes, they are both in use and more importantly, they turned out to be different!
The type of `year`  is `number` in `demo/olympics.data` but `string` in `olympics.data`. The query wasn't working because aggregations on strings don't work in mimir, but that's already tested in a couple of other ITs. So I changed it to a green test because the value of a green test is higher than a red one (given that there's at least 1 red test for an issue).

I do have in mind to spend some more time improving the tests on both of these datasets (hopefully when I change the last couple of ITs as part of ch402). If we need to keep both datasets then it should at least be more clear how they differ.

rintcius:
@alissapajer PR ready for another look

PR #3724, opened 2018-07-26
support showing phase results in repl

rintcius:
* support `explain <query>` (alias `compile <query>`) in repl
* support setting `debug` when running a query in repl
* log `QScriptEducated` phase result


repl/src/main/scala/quasar/repl/DebugLevel.scala
```diff
@@ -39,6 +39,9 @@ object DebugLevel {
       case Verbose => 2
     }
 
-  implicit val equal: Equal[DebugLevel] =
-    Equal.equalBy(int(_))
+  implicit val order: Order[DebugLevel] = new Order[DebugLevel] {
```
wemrysi: :golf: `Order.orderBy(int(_))`

run/src/main/scala/quasar/run/Sql2QueryEvaluator.scala
```diff
@@ -54,6 +56,6 @@ object Sql2QueryEvaluator {
 
       lp  <- queryPlan[F, T, T[LP]](sql, sqlQuery.vars, sqlQuery.basePath, 0L, None)
 
-      qs  <- LPtoQS[T].apply[StateT[F, Long, ?]](lp).eval(0)
+      qs  <- phaseM[F]("QScriptEducated", LPtoQS[T].apply[StateT[F, Long, ?]](lp).eval(0))
```
djspiewak: Shouldn’t this be “QScript (educated)”, following the convention of the other variants?

common/src/main/scala/quasar/common/package.scala
```diff
@@ -44,6 +44,7 @@ package object common {
 
   type PhaseResultW[A] = Writer[PhaseResults, A]
   type PhaseResultT[F[_], A] = WriterT[F, PhaseResults, A]
+  type PhaseResultCatsT[F[_], A] = cats.data.WriterT[F, PhaseResults, A]
```
wemrysi: Not sure if this is warranted since we only use it in one place.

run/src/main/scala/quasar/run/Sql2QueryEvaluator.scala
rintcius: right, it was "QScript (Educated)" in old repl - capital E. 
Shall we settle for that?

repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -19,44 +19,63 @@ package repl
...
-    MonadTell_.ignore[IO, PhaseResults]
+  type IOT[A] = PhaseResultCatsT[IO, A]
 
-  implicit val ioQuasarError: MonadError_[IO, QuasarError] =
-    MonadError_.facet[IO](QuasarError.throwableP)
+  implicit val iotQuasarError: MonadError_[IOT, QuasarError] =
+    MonadError_.facet[IOT](QuasarError.throwableP)
 
-  val quasarStream: Stream[IO, Quasar[IO]] =
+  implicit val iotListen: PhaseResultListen[IOT] =
```
wemrysi: This and `iotTell` shouldn't be necessary if you `import quasar.contrib.cats.writerT._` above.

rintcius:
[ch645]
[ch647]

run/src/main/scala/quasar/run/Quasar.scala
```diff
@@ -106,9 +107,10 @@ object Quasar {
 
       federation = MimirQueryFederation[Fix, F](precog)
 
-      queryEvaluator =
+      (queryEvaluator: QueryEvaluator[F, SqlQuery, Stream[IO, Data]]) =
         Sql2QueryEvaluator(FederatingQueryEvaluator(federation, ResourceRouter(running.get)))
 
-    } yield new Quasar(datasources, queryEvaluator)
+    } yield new Quasar(datasources,
+        QueryEvaluator.functor[F, SqlQuery].map(queryEvaluator)(_.translate(λ[IO ~> F](_.to[F]))))
```
wemrysi: Stylistically, I'd probably bind this to a name to make it easier to read
```scala
  queryEvaluatorIO = Sql2QueryEvaluator(...)

  queryEvaluator = queryEvaluatorIO.map(...)

} yield new Quasar(datasources, queryEvaluator)
```

common/src/main/scala/quasar/common/package.scala
rintcius: I tend to add these defs by analogy, not so much by actual usage (which can change)

repl/src/main/scala/quasar/repl/Main.scala
rintcius: yeah, I guess it's due to the trouble I had to get this working, but I added this to add clarity.

common/src/main/scala/quasar/common/package.scala
wemrysi: Yeah, though the others are probably worth an audit as well. They were there from another time where we used transformers much more frequently.

repl/src/main/scala/quasar/repl/Main.scala
wemrysi: I see. Ironically, they actually confused me because they're adjacent to two instances that we have to create locally whereas I'd expect these to be implicitly resolved given the listen/tell implementations elsewhere in this PR.

Generally, we only add local instances like this when necessary.

repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -19,44 +19,63 @@ package repl
...
-      repl <- Repl.mk[IO](ref, q.datasources, q.queryEvaluator)
+      ref <- Ref[F, ReplState](ReplState.mk)
+      repl <- Repl.mk[F](ref, q.datasources, q.queryEvaluator)
       l <- repl.loop
     } yield l
 
-  override def stream(args: List[String], requestShutdown: IO[Unit]): Stream[IO, ExitCode] =
-    quasarStream >>= (q => Stream.eval(repl(q)))
-
+  override def stream(args: List[String], requestShutdown: IO[Unit]): Stream[IO, ExitCode] = {
```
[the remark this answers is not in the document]
rintcius: ah, that's nice!

changing this means that usage of `PhaseResultCatsT` becomes 2 now, right? because sadly I can't use `IOT` type in  `object Main extends StreamApp[IOT]`

repl/src/main/scala/quasar/repl/Main.scala
rintcius: ok, getting rid of it

wemrysi:
@rintcius Will merge once conflicts are resolved.

rintcius:
@wemrysi conflicts resolved & build is green

PR #3727, opened 2018-07-27
[DO NOT MERGE] Revert slow repair in ExpandShifts

jsantos17:
I'm no longer certain we should revert this change, given the number of regressions it would imply. But performance is still an order of magnitude better with this change. Let's wait for the performance fixes @djspiewak mentioned before merging this?


jsantos17:
Closing this. I can reopen if we end up reverting.

PR #3728, opened 2018-07-27
Accept the Precog instance as an argument to Quasar

wemrysi:


alissapajer:
@wemrysi what's the upstream situation which requires this?

wemrysi:
@alissapajer Specifically, we need a `Precog` to setup table description storage over in `slamdata-backend`.

alissapajer:
@wemrysi alternatively, we could instantiate a `MimirIndexedStore` here in quasar and then expose _that_ instead of a `Precog`. Is there a reason exposing a `Precog` is better?

PR #3729, opened 2018-07-30
Implemented IfUndefined as a primitive in mimir

djspiewak:
Unit tests are a bit lacking because I'm in a hurry but I'm relatively certain that our test coverage is sufficient via the other test suites (especially `StdLibSpec`).

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
@@ -590,6 +590,36 @@ abstract class Slice { source =>
     }
   }
 
+  def ifUndefined(default: Slice): Slice = {
+    val defined = {
+      val bsing = source.columns.values map {
+        case col: BitsetColumn => col.definedAt.copy
+        case col: DefinedAtIndex => col.defined.copy
+        case col => col.definedAt(0, size)
+      }
+
+      val back = bsing.reduceOption(_ | _).getOrElse(new BitSet)
+      back.flip(0, size)
+      back
+    }
+
+    val masked = default.columns lazyMapValues { col =>
+      cf.util.filter(0, size, defined)(col).get
+    }
+
+    val merged = masked.foldLeft(source.columns) {
+      case (acc, (ref, col)) if acc.contains(ref) =>
+        acc.updated(ref, cf.util.UnionRight(col, acc(ref)).get)
+
+      case (acc, pair) => acc + pair
+    }
+
+    new Slice {
+      val size = source.size
+      val columns = merged
+    }
+  }
+
   def compact(filter: Slice, definedness: Definedness): Slice = {
     new Slice {
       private val cols = filter.columns
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/cf/Util.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/cf/Util.scala
@@ -31,95 +31,69 @@ object util {
   val UnionRight = CF2P {
     case (c1: BoolColumn, c2: BoolColumn) =>
       new UnionColumn(c1, c2) with BoolColumn {
-        def apply(row: Int) = {
-          if (c2.isDefinedAt(row)) c2(row) else if (c1.isDefinedAt(row)) c1(row) else sys.error("Attempt to retrieve undefined value for row: " + row)
-        }
+        def apply(row: Int) = if (c2.isDefinedAt(row)) c2(row) else c1(row)
       }
 
     case (c1: LongColumn, c2: LongColumn) =>
       new UnionColumn(c1, c2) with LongColumn {
-        def apply(row: Int) = {
-          if (c2.isDefinedAt(row)) c2(row) else if (c1.isDefinedAt(row)) c1(row) else sys.error("Attempt to retrieve undefined value for row: " + row)
-        }
+        def apply(row: Int) = if (c2.isDefinedAt(row)) c2(row) else c1(row)
       }
 
     case (c1: DoubleColumn, c2: DoubleColumn) =>
       new UnionColumn(c1, c2) with DoubleColumn {
-        def apply(row: Int) = {
-          if (c2.isDefinedAt(row)) c2(row) else if (c1.isDefinedAt(row)) c1(row) else sys.error("Attempt to retrieve undefined value for row: " + row)
-        }
+        def apply(row: Int) = if (c2.isDefinedAt(row)) c2(row) else c1(row)
       }
 
     case (c1: NumColumn, c2: NumColumn) =>
       new UnionColumn(c1, c2) with NumColumn {
-        def apply(row: Int) = {
-          if (c2.isDefinedAt(row)) c2(row) else if (c1.isDefinedAt(row)) c1(row) else sys.error("Attempt to retrieve undefined value for row: " + row)
-        }
+        def apply(row: Int) = if (c2.isDefinedAt(row)) c2(row) else c1(row)
       }
 
     case (c1: StrColumn, c2: StrColumn) =>
       new UnionColumn(c1, c2) with StrColumn {
-        def apply(row: Int) = {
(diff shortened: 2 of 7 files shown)


djspiewak:
[ch779]

yggdrasil/src/main/scala/quasar/yggdrasil/table/cf/Util.scala
```diff
@@ -31,95 +31,69 @@ object util {
...
       }
 
     case (c1: HomogeneousArrayColumn[a], _c2: HomogeneousArrayColumn[_]) if c1.tpe == _c2.tpe =>
       val c2 = _c2.asInstanceOf[HomogeneousArrayColumn[a]]
       new UnionColumn(c1, c2) with HomogeneousArrayColumn[a] {
         val tpe = c1.tpe
-        def apply(row: Int) = {
-          if (c2.isDefinedAt(row)) c2(row) else if (c1.isDefinedAt(row)) c1(row) else sys.error("Attempt to retrieve undefined value for row: " + row)
-        }
+        def apply(row: Int) = if (c2.isDefinedAt(row)) c2(row) else c1(row)
```
alissapajer: Strange the `sys.error` was here. I agree with the diff. Maybe `UnionColumn` had different semantics previously?

alissapajer: Although this doesn't change the semantics; it just throws a more explicit error. And that error is implicit in the contract that we'll call `isDefinedAt` prior to calling `apply`.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -60,7 +60,6 @@ class SliceSpec extends Specification with ScalaCheck {
   }
 
   def toCValues(slice: Slice) = sortableCValues(slice, Vector.empty) map (_._2)
-
```
alissapajer: :disappointed:

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
```diff
@@ -601,6 +600,50 @@ class SliceSpec extends Specification with ScalaCheck {
       }
     }
   }
+
+  "ifUndefined" should {
+    "merge slices with holes" in {
```
alissapajer: Could you add another test just like this one, except make the `"a"` column heterogeneous as well as having holes?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -590,6 +590,36 @@ abstract class Slice { source =>
     }
   }
 
+  def ifUndefined(default: Slice): Slice = {
+    val defined = {
+      val bsing = source.columns.values map {
+        case col: BitsetColumn => col.definedAt.copy
+        case col: DefinedAtIndex => col.defined.copy
+        case col => col.definedAt(0, size)
```
alissapajer: `size` throughout this function is calling the abstract method `size` on `Slice` and not directly calling the method through the self reference (you'd need `source.size` to reference that). I've been unable to insert myself into the cake enough to determine what the consequences of this will be. Though previously the cake-y thing to do was to call `source.size`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -590,6 +590,36 @@ abstract class Slice { source =>
...
+      }
+
+      val back = bsing.reduceOption(_ | _).getOrElse(new BitSet)
+      back.flip(0, size)
+      back
+    }
+
+    val masked = default.columns lazyMapValues { col =>
+      cf.util.filter(0, size, defined)(col).get
```
alissapajer: We're masking the `default.columns` by the rows of `source.columns` that are undefined, not defined. So I think `defined` should be named `undefined`. I was about to write a note saying we need to flip the bitset when I discovered line 602 :)

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -590,6 +590,36 @@ abstract class Slice { source =>
...
+  def ifUndefined(default: Slice): Slice = {
+    val defined = {
+      val bsing = source.columns.values map {
+        case col: BitsetColumn => col.definedAt.copy
+        case col: DefinedAtIndex => col.defined.copy
+        case col => col.definedAt(0, size)
+      }
+
+      val back = bsing.reduceOption(_ | _).getOrElse(new BitSet)
+      back.flip(0, size)
```
alissapajer: I think there is a bug. Say that `default` has _more_ rows than `source`. When we call `back.flip(0, source.size)` we are only flipping the bits up to row `source.size`. But really we want all the rows after `source.size` in `default` to have definedness value `1` when we `filter` below. But this won't be the case.

A unit test should be able to expose this.

yggdrasil/src/main/scala/quasar/yggdrasil/table/cf/Util.scala
djspiewak: Removing the `sys.error` allowed me to remove the call to `c1.isDefinedAt(row)`, which nearly halves the runtime of columns resulting from `UnionRight`. So this was a performance change. You're quite correct that the contract is the *endpoint* will call `isDefinedAt` prior to calling `apply`.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
djspiewak: Spoooooooky! I didn't delete this. At least, I have no memory of doing so. I certainly agree it's far better to keep it.

yggdrasil/src/test/scala/quasar/yggdrasil/table/SliceSpec.scala
djspiewak: Not today. Since this doesn't appear to make a major performance difference, we can keep this open pending more tests in mimir, but there's no way I can get to it right now. I'm actually comfortable merging this as-is due to `StdLibSpec` (as I mentioned in the PR description). I'd still like more tests specifically here but we can live without them for now. Or not, either way.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: `Slice` isn't in the cake. `source.size` is exactly the same as `size` (as in, same bytecode).

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: Yeah, I can flip the name. I was originally thinking about it in terms of image manipulation and photoshop masks and finding where the `source` slice was defined, but you're correct that it works out more to finding where it's *undefined*. I feel less comfortable thinking in those negative terms since, technically, the set of rows for which `source` is *not* is infinite.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
djspiewak: If `default` has more rows than `source`, this function is simply not defined. This is the case for most pairwise functions on `Slice` due to the fact that `SliceTransform` will *guarantee* the slices are the same size. So any unit test which exposes this issue is in fact an invalid test.

alissapajer:
@djspiewak I knew exactly the unit test I wanted, so I pushed it up.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
alissapajer: Yeah that's a fair point, and there's no one mental model that is the best for thinking about this kind of bitset stuff. I'm fine keeping the name is as well.

PR #3730, opened 2018-07-30
Eager(er) leftShift in mimir

djspiewak:
What I did here is make all of the definedness calculation eager, which avoids the largest hotspot that we saw in the stack traces. I also optimized for the extremely common homogeneous shift case, which previously caused a quadratic expansion of delegation through `UnionRight`. Finally, `emitOnUndefined = true` now is ultimately handled by a very tight and mostly-intrinsified loop within `BitSet` for any shift with a high water mark less than 64 (and above 64 it's still pretty fast).


djspiewak:
[ch771]

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -256,6 +298,40 @@ public void flip(int fromIndex, int toIndex) {
         }
     }
 
+    /**
+     * Flips all bits for which index % mod == 0 such that
+     * the bits (i % mod >= 0 && i % mod < mod) are 0. The
+     * invariant here is that, for all ranges of i such
+     * that (i % mod >= 0 && i % mod < mod), there is *at least*
+     * one value of i such that get(i) == true.
+     */
```
alissapajer: If we're flipping all the bits for which `i % mod == 0`, it's possible that we flip some of those bits from `1` to `0`. As such, I don't see how the invariant that `get(i) == true` holds for certain ranges?

Say we have `[1, 0, 0, 1, 0, 0, 1]` and we call `flipByMod(3)`. Do we not get a bitset of all `0`'s?

djspiewak: `flip` is the wrong word. You're actually pointing out a bug that this function had previously. You'll notice that the actual implementation uses `|=` rather than `^=`.

alissapajer: Once a `1` always a `1`.

alissapajer: I'd argue that "flip" is a technical term when it comes to bits, so renaming this would be helpful. Maybe `setByMod`?

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -256,6 +298,40 @@ public void flip(int fromIndex, int toIndex) {
...
+        }
+
+        // we special-case this because it has a much faster implementation (and also our main impl assumes mod > 1)
+        if (mod == 1) {
+            final long replacement = 0xFFFFFFFFFFFFFFFFL;
+            for (int i = 0; i < _length; i++) {
+                bits[i] = replacement;
+            }
+            return;
+        }
```
alissapajer: An alternate way to implement this would be to call `setBits` and provide an array of length `_length` filled entirely with `replacement`. That would actually require allocating a second array though, whereas what you have here is edit-in-place. So :+1:

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -256,6 +298,40 @@ public void flip(int fromIndex, int toIndex) {
...
+
+        long flipper = 1L;
+        long mask = (1L << mod) - 1L;  // bitmask which covers everything >= 0 and < mod
+        for (int i = 0; i < _length; i++) {
+            do {
+                if ((bits[i] & mask) == 0L) {
+                    bits[i] |= flipper;
+                }
```
alissapajer: I have parsed this to be:

`mask` is all `1`'s for an interval of length `mod`. We thus enter this `if` body when _none_ of the bits of `bits[i]` in that interval are `1`, in which case we set the first bit in that interval to `1` with the `flipper`. If we didn't enter the `if` body, we know at least one bit in that interval is already `1`, but it might not be the first one.

Is this what you're going for? Furthermore, the `0`th long array in `bits` is masked by a mask starting with the masking interval at the low-order. Then the `1`th long array in `bits` is masked with the interval rotated over once toward the high-order by `mod`, etc.

... this is nifty to think about. How does it fit in the larger picture of `Column`? :smile:

blueeyes/src/main/java/quasar/precog/BitSet.java
alissapajer: I've actually changed my mind. This is kind of a flipper. Just not the flipper you might be expecting.

djspiewak: Is it the flipper we deserve?

yggdrasil/src/test/scala/quasar/yggdrasil/LeftShiftSpec.scala
```diff
@@ -199,6 +199,33 @@ trait LeftShiftSpec extends TableModuleTestSupport with SpecificationLike {
     toJson(table.leftShift(CPath.Identity \ 1, emitOnUndef = emit)).getJValues mustEqual expected
   }
 
+  def testUnevenHomogeneousArraysEmit = {
+    // closely drawn from intArrays.data
+    val table =
+      fromSample(
+        SampleData(
+          Stream(
+            toRecord(Array(0), JObject("b" -> JArray(JNum(1) :: JNum(2) :: Nil), "c" -> JArray(JNum(11) :: JNum(12) :: Nil)).some),
```
rintcius: could improve the test a bit by making each record have a different index, i.e. next line becomes:

```
toRecord(Array(1), ...
```

blueeyes/src/main/java/quasar/precog/BitSet.java
djspiewak: I can add the examples. The condition though is much more precise than the examples, and it translates directly into the scalacheck conditional. If you prefer, I could add phrasing to describe the condition as "`i` projected into the mod ring".

blueeyes/src/main/java/quasar/precog/BitSet.java
djspiewak: This is indeed what I was going for! However, I just realized that there's a bug in it! I wonder if this is what's creating the failing tests. `mask` rotates around and will end up sometimes masking high-order *and* low-order bits simultaneously. This is in and of itself correct, but the high-order bits from `bits[i]` should be masked while the low-order bits from `bits[i + 1]` are masked, whereas right now everything is from `bits[i]` until it fully wraps around. This could be creating the bugs that we're seeing.

> How does it fit in the larger picture of Column? 😄

Great question! Think about the columns in a `LeftShift` hole which are not included in the `struct`. These columns need to be inflated such that each row in the input corresponds to at most *n* rows in the output, where *n* is the number of columns described by the struct (e.g. the width of the array/object). Note that the dataset may be heterogeneous, so *n* may not be the right size for *every* row, but it's the definitive high water mark.

Now, when we are doing `OnUndefined.Omit`, we can take the definedness bitset from the results of the shift and use that to mask the inflated unfocused columns (the ones which expand to size *n*). This gives us "inner" semantics on the map which is passed into the `repair` and strips out rows for which the shift was undefined. This is actually really easy and efficient.

When we do `Emit` though, we have to figure out *which* of the rows to hang onto. If `RightSide` is defined for a given *input* row, then we should throw away all of the others. If `RightSide` is undefined, then we want to keep *exactly* one row from the `LeftSide`. This is where it gets a little tricky, and it's why I wrote the soon-to-be-renamed `setByMod`.

When we're in the `Emit` case, we take the definedness bitset from the `RightSide` and call `setByMod` on it, where the `mod` is the high water mark (*n* from earlier). So this ensures that, for all input rows, *at least* one row will be defined in the unfocused columns. If a row is already defined (e.g. because `RightSide` is defined for that row), then we use it. Otherwise, we arbitrarily pick the first row and bring it back to life (it's valid to pick any of the rows within the mod ring, the first is just convenient).

blueeyes/src/main/java/quasar/precog/BitSet.java
djspiewak: Yeah I thought about that approach. I *think* the way I'm doing it is faster, especially since it's extremely transparent to both HotSpot and the CPU itself, but I don't have measurements to back up that assumption.

blueeyes/src/main/java/quasar/precog/BitSet.java
rintcius: Yeah, I meant the examples just as an illustration of the text.

>  If you prefer, I could add phrasing to describe the condition as "i projected into the mod ring"

👍

djspiewak:
Mostly for my own notes… Here's a minimized version of the failing test: https://gist.github.com/26fa058cf0a5a390fc4981067329d62a

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -1402,49 +1400,56 @@ trait ColumnarTableModule
...
+        (remapped, unfocused)
       }
 
-      def merge(focused: Map[ColumnRef, Column], innerIndex: Map[CPathNode, Int], expansion: CF1, highWaterMark: Int)
+      def merge(
+          focused: Map[ColumnRef, Column],
+          innerIndex: Map[CPathNode, Int],
+          expansion: CF1,
+          highWaterMark: Int)
           : Map[ColumnRef, Column] = {
```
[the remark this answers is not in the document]
djspiewak: It is not. The standard is the way that I formatted it.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+        val mod = math.abs(mod0) % 20    // don't let scalacheck go crazy
+
+        (mod > 0) ==> {
+          bs.setByMod(mod)
+          val bound = bs.length << 6
+
+          (0 until (bound / mod) by mod) forall { i =>
+            (0 until mod) must contain { j: Int =>
+              if (i + j < bound)    // mod might not divide bound
+                bs(i + j) must beTrue
```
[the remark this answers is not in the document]
djspiewak: For the scalacheck property, I was only checking the guaranteed invariant 1. Checking invariant 2 is probably a good idea as well, but it's much harder to quantify. I check this in the other tests.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+        val mod = math.abs(mod0) % 20    // don't let scalacheck go crazy
+
+        mod > 2 ==> {
+          val offset = math.abs(offset0) % mod
+
+          val sparsened = bs.sparsenByMod(offset, mod)
+          val bound = bs.length << 6
+
+          (0 until bound) forall { i =>
+            bs(i) mustEqual sparsened(i * mod + offset)
```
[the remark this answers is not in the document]
djspiewak: This property checks the only *guarantee* of the function.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+      bs
+    })
+  }
+
+  "bitset (java)" should {
+    "sparsenByMod" in prop { (bs: BitSet, offset0: Int, mod0: Int) =>
+      (mod0 > Int.MinValue && offset0 > Int.MinValue) ==> {
+        val mod = math.abs(mod0) % 20    // don't let scalacheck go crazy
+
+        mod > 2 ==> {
```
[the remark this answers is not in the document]
djspiewak: I think I could change it to `mod > 1`. Allowing `mod = 1` would be weird and doesn't actually line up with what we care about.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+          (0 until bound) forall { i =>
+            bs(i) mustEqual sparsened(i * mod + offset)
+          }
+        }
+      }
+    }
+
+    "setByMod" in prop { (bs: BitSet, mod0: Int) =>
+      (mod0 > Int.MinValue) ==> {
+        val mod = math.abs(mod0) % 20    // don't let scalacheck go crazy
```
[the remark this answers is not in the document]
djspiewak: I'm probably going to push up this value, yes. The other tests are hitting a lot of the bizarre cases.

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -217,6 +217,48 @@ public BitSet copy() {
         return new BitSet(_bits, _length);
     }
 
+    /**
+     * Inflates the bitset by a factor of `mod` such that
+     * bs.contains(i) == bs.sparsenByMod(o, m).contains(i * m + o)
+     * for all values of bs, i, o, m.
+     */
+    public BitSet sparsenByMod(final int offset, final int mod) {
+        long[] _bits = new long[_length * mod];
```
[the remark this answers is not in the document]
djspiewak: I could, but there's actually no point since we're never going to give it those values and anything `mod < 0` will lead to an exception anyway. I was actually thinking about removing the checks from `setByMod` as well for the same reason. This is not intended to be safe code.

blueeyes/src/main/java/quasar/precog/BitSet.java
```diff
@@ -256,6 +298,83 @@ public void flip(int fromIndex, int toIndex) {
...
+     * Ensures that all disjoint index mod rings contain at least one
+     * set bit. If a disjoint index mod ring *already* contains a set
+     * bit, then its contents are guaranteed to be left unmodified.
+     * More rigorously, the first invariant is captured by:
+     *
+     * ∀ i . ∃ j >= 0 && j < mod . bits(i + j) == 1
+     */
+    public void setByMod(int mod) {
+        if (mod >= 64) {
+            throw new IllegalArgumentException("this algorithm doesn't work for mod > 64; got " + mod);
```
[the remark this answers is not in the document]
djspiewak: Yeah the conditional is wrong, while the message is correct. Doesn't matter because the whole conditional is going away.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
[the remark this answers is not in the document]
rintcius: hmm, why isn't there a guarantee that the bits of `sparsened` that are not checked in l.47 are not set?

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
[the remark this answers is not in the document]
rintcius: i actually thought it'd be _easier_ to verify. Just need to iterate over the bits:

- if it's possibly changed (i.e. the first bit in each range), then skip testing it
- for all other bits in these ranges they must be equal to the original bit

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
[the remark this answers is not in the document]
rintcius: ok, written to memory :) there are places that don't follow this format, but the scalafmt PR did indeed format it like that.

blueeyes/src/main/java/quasar/precog/BitSet.java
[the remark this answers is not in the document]
rintcius: hmm, I think it's good to have the checks, so that we are explicit in how it can be called & fail fast if someone calls it in a wrong way.

djspiewak:
[ch769]

This one should *also* be fixed by this change.

djspiewak:
[ch731]

djspiewak:
[ch746]

alissapajer:
@djspiewak before closing stories that are specific to particular queries failing, I think we should make sure there is a test hitting that query.

djspiewak:
@alissapajer Both of those only hard-fail due to externalities (timeouts and paging, respectively). I gave it some thought but I'm not sure how to write a test for them within our existing harnesses. Both could be integration tested using a hypothetical harness that sits on top the rest API in sdbe, but we don't have that harness right now. So I hate to say it, but we have to leave those to QA. :-(

blueeyes/src/main/java/quasar/precog/BitSet.java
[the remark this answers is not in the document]
djspiewak: Some of this is performance-sensitive code though. In this case, I think it might not matter much, but it's something to keep in mind. Checks aren't free.

djspiewak:
@rintcius I think I've addressed all your PR review feedback. Due to the pressing release schedule, I'm going to reassign this to someone in a western timezone so we can merge it tonight. Please feel free to leave more review feedback at your discretion and I'll address asap in a follow-up PR.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+      bs.setByMod(35)
+
+      bs(35) must beTrue
+
+      bs(350) must beTrue
+      (351 to 384).map(bs(_)) must contain(beFalse).forall
+    }
+
+    "setByMod works with mod 3" in {
+      val bs = new BitSet(128)    // ensure we have seven longs
```
wemrysi: s/seven/two/

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+
+      // 64 % 35 = 29, and 64 * 6 = 384, and 384 % 35 = 0
```
wemrysi: Copy-paste artifact?

rintcius:
@djspiewak awesome! The main comment that wasn't addressed was regarding the spec of `sparsenByMod`. I thought to just PR it to both verify my understanding of that function and explain what I mean - see #3738 

While fiddling with that spec I found a bug in `sparsenByMod`, see that PR for details.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
djspiewak: It is.

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
djspiewak: They're basically the same number…

PR #3731, opened 2018-07-31
Verify from string conversions

rintcius:
replace incorrect `coercions` test with `conversionsNonDate` and correct expectation of `arrayFlattenAsInt`

Rebased version of #3692 with comments addressed


rintcius:
relates to ch156

PR #3732, opened 2018-07-31
Use MimirRepr instead of Data in mimir evaluator

alissapajer:


alissapajer:
This probably isn't going to work because of the dependent types. Hold off on merge @wemrysi

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
```diff
@@ -119,7 +119,9 @@ final class Sql2QueryRegressionSpec extends Qspec {
       (q, i) = t
 
       f = (squery: UUID => SqlQuery) =>
-        Stream.force(q.queryEvaluator.evaluate(squery(i)))
+        Stream.force {
+          q.queryEvaluator.evaluate(squery(i)).map(_.map(mimir.tableToData))
+        }.flatMap(s => s)
```
djspiewak: I don't normally make formatting comments, but… `}.` should always be a *huge* red flag because the precedence of that operation varies tremendously depending on what operation happens immediately prior to the opening brace (`{`). For the above, I would prefer one of the following:

```scala
Stream.force(
  q.queryEvaluator.evaluate(squery(i)).map(_.map(mimir.tableToData))).flatMap(s => s)

// or (you'll need to wrap the function starting on line 121 in curly braces, which you should do anyway)

val s = Stream force {
  q.queryEvaluator.evaluate(squery(i)).map(_.map(mimir.tableToData))
}

s.flatMap(x => x)

// or 

Stream force {
  q.queryEvaluator.evaluate(squery(i)).map(_.map(mimir.tableToData))
} flatMap { s => s }
```

A good rule of thumb: use spaces with curly braces, dots with parens, use curly braces for multi-line, parens for single-line, and never *mix* styles.

mimir/src/main/scala/quasar/mimir/package.scala
```diff
@@ -50,4 +53,10 @@ package object mimir {
 
   def slicesToStream[F[_]: Functor](slices: StreamT[F, Slice]): Stream[F, RValue] =
     convert.fromChunkedStreamT(slices.map(s => Chunk.indexedSeq(SliceIndexedSeq(s))))
+
+  def tableToData(repr: MimirRepr): Stream[IO, Data] =
```
djspiewak: Are we still using this function somewhere? We shouldn't be if it's at all avoidable.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -39,7 +40,8 @@ import java.lang.Exception
 import scala.util.control.NonFatal
 
 import argonaut.{Json, JsonParser, JsonScalaz}, JsonScalaz._
-import cats.effect._
+import cats.~>
```
djspiewak: I've been using cats' `cats.arrow.FunctionK` instead of their `~>` alias in scopes which also have scalaz, so I don't get confused with `scalaz.~>`. I don't have strong opinions on it though.

mimir/src/main/scala/quasar/mimir/evaluate/MimirQScriptEvaluator.scala
```diff
@@ -81,12 +80,8 @@ final class MimirQScriptEvaluator[
...
     Unicoalesce.Capture[T, QS[T]]
 
-  def execute(repr: Repr): M[Stream[IO, Data]] =
-    mimir.slicesToStream(repr.table.slices)
-      // TODO{fs2}: Chunkiness
-      .mapSegments(s =>
-        s.filter(_ != CUndefined).map(RValue.toData).force.toChunk.toSegment)
-      .point[M]
+  def execute(repr: Repr): M[Stream[IO, Repr]] =
+    Stream(repr).covary[IO].point[M]
```
wemrysi: Maybe we just make the output type `Repr`? Doesn't look like we need the `Stream` here.

mimir/src/main/scala/quasar/mimir/package.scala
wemrysi: Looks like its used in the tests and in the repl. We'll probably always need something like this for the tests unless we rewrite the predicates, the repl we could render results from the table but doing so isn't terribly urgent.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -57,7 +59,7 @@ import spire.std.double._
 final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResultTell](
     stateRef: Ref[F, ReplState],
     sources: Datasources[F, Stream[F, ?], UUID, Json, SstConfig[Fix[EJson], Double]],
-    queryEvaluator: QueryEvaluator[F, SqlQuery, Stream[F, Data]]) {
+    queryEvaluator: QueryEvaluator[F, SqlQuery, Stream[F, MimirRepr]]) {
```
wemrysi: If we left this as-is and transformed the `QueryEvaluator` before passing it to the `Evaluator`, we wouldn't have to change anything here and it'd localize the change to `Main`.

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
djspiewak: That works too, though it's definitely my least favorite option of all of the possibilities.

mimir/src/main/scala/quasar/mimir/package.scala
djspiewak: If it's just tests and repl then I'm 👍 on keeping it.

alissapajer: It's just tests and repl.

alissapajer:

I'm going to cast (in just one place) to get around the dependent type problem for now.

PR #3733, opened 2018-07-31
Leftshift benchmark

rintcius:
See https://gist.github.com/rintcius/9552dc741a547e9e5b60a7d37a1a6f10 for preliminary results (I can run with more iterations tomorrow if #3730 is ready then).

Overall the fixes in #3730 improve leftshift performance considerably. Exception is `scalarOrUndefined` - this became quite a lot slower.

### DIFF
--- a/yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/LeftShiftBenchmark.scala
+++ b/yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/LeftShiftBenchmark.scala
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar
+package yggdrasil
+package perf
+
+import quasar.contrib.cats.effect.liftio._
+import quasar.fp.ski.κ
+import quasar.precog.common.{CLong, CPath, RArray, RObject, RValue}
+import quasar.yggdrasil.table.TestColumnarTableModule
+
+import cats.effect.IO
+
+import java.util.concurrent.TimeUnit
+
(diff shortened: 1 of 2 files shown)


rintcius:
[ch757]

djspiewak:
@rintcius Unless I'm misreading those results, `bug/slow-shift` looks radically *slower* than master for everything except `distinctFieldsObjects`. Am I missing something? @jsantos17's testing seemed to show a pretty substantial improvement, so I feel like I'm either misreading your gist, or there's something wrong with our benchmark assumptions here.

rintcius:
@djspiewak you have to compare the 2nd and 3rd sections, those are both consuming json. The first section is just there for a comparison with the 2nd (compares json vs materialize consumption on master).

djspiewak:
Oh good! I *was* misreading then. Looks like a 78% improvement in the best case, a 22% improvement in the worst positive case, and a 97% regression in the worst negative case. I wonder why `scalarOrUndefined` is so bad. What does the data look like in that case?

This is definitely exciting. I really appreciate you running it on my PR.

rintcius:
`scalarOrUndefined` is basically testing cases like `a = 7 ` 
 and `a = 13` here: https://github.com/slamdata/quasar/blob/master/it/src/main/resources/tests/heterogeneous.data#L7
on queries where you flatten/shift `b` (i.e. the result is `undefined`)

djspiewak:
@rintcius Ah interesting. That's not going to happen much in practice but it's also easy to fast-path. Something to think about for the future.

beckyconning:
> That's not going to happen much in practice

Are we sure? Maybe we can have a chat about this to see how it fits in with the product.

djspiewak:
> Are we sure? Maybe we can have a chat about this to see how it fits in with the product.

beckyconning:
I'm not saying we necessarily need to fix this now nor that we can't avoid this case somehow but that I believe unless its specifically mitigated against this case will be quite common in usage. E.g.

Consider the dataset

```
{ x: [1,2,3] }
{ x: 1 }
```

These columns are picked from this dataset:

A) All indices > Column Number
B) Column Number

A = x[_] :: Number
B = x :: Number

rintcius:
@beckyconning The confusion may be that `scalarOrUndefined` is a case in which __all__  values are scalars or `undefined`. There is also the `heterogeneous` test case. In this case half of the values are scalars or undefined & in this case scores are:

```
heterogeneous     10000   5   true  avgt   1216.961  ms/op (master)
heterogeneous     10000   5   true  avgt    658.807  ms/op (daniels branch)
```

So here the scores are already better in Daniel's branch. I think @djspiewak may be just speaking about the probability that they are _all_ scalars/undefined, not for the heterogenous case. 

Still worth to find out why this is happening. Just a thought, I can imagine that the wrong results are partly due to this (because at the moment too many results are returned, so more data to process)

PR #3738, opened 2018-08-03
test leftshift

rintcius:
* improve `sparsenByMod` spec
* note: there's a bug in `sparsenByMod` which can be shown if you use `val mod = math.abs(mod0) % 100 //was 20 ` in the test (happens both in your version and my version)
* add a couple of leftshift test cases on very heterogenous data (of which half of them are pending because they give wrong results; I can imagine they are related to the bug mentioned above but haven't tracked that down)


blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -48,10 +48,14 @@ object BitSetSpec extends Specification with ScalaCheck {
           val offset = math.abs(offset0) % mod
 
           val sparsened = bs.sparsenByMod(offset, mod)
-          val bound = bs.length << 6
 
-          (0 until bound) forall { i =>
-            bs(i) mustEqual sparsened(i * mod + offset)
+          (sparsened.length << 6) must beEqualTo((bs.length << 6) * mod)
+          
+          (0 until (sparsened.length << 6)) forall { i: Int =>
```
rintcius: this tests all bits in `sparsened` instead of `1 / mod` of the bits

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
```diff
@@ -48,10 +48,14 @@ object BitSetSpec extends Specification with ScalaCheck {
...
+            if (i % mod == offset)
+              bs(i / mod) must beEqualTo(sparsened(i)).setMessage(s"bs(${i / mod}) != sparsened($i)")
+            else
+              sparsened(i) must beEqualTo(false).setMessage(s"sparsened($i) != false")
```
rintcius: `beEqualTo(false)` instead of `beFalse` because otherwise the message isn't altered when there's a failure

it/src/main/resources/tests/veryHeterogeneousShiftArrayValueWithFieldOnMap.test
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "field and shifted array value on very heterogenous maps",
+    "backends": {
+        "lwc_local": "pendingIgnoreFieldOrder",
+        "mimir":     "pendingIgnoreFieldOrder"
+    },
+    "data": "very_heterogeneous_array.data",
+    "query": "select a, b[_] as v from `very_heterogeneous_map.data`",
```
rintcius: returns just 1 row

it/src/main/resources/tests/veryHeterogeneousFlattenArrayIndexWithField.test
```diff
@@ -0,0 +1,117 @@
+{
+    "name": "field and flattened array index on very heterogenous arrays",
+    "backends": {
+        "lwc_local": "pendingIgnoreFieldOrder",
+        "mimir":     "pendingIgnoreFieldOrder"
+    },
+    "data": "very_heterogeneous_array.data",
+    "query": "select a, b[*:] as i from `very_heterogeneous_array.data`",
```
rintcius: returns 100 rows instead of 104 also nog equally distributed over the `a`'s

rintcius:
[ch925]

blueeyes/src/test/scala/quasar/precog/BitSetSpec.scala
djspiewak: Odd. `beFalse.setMessage` works for me.

alissapajer:
@rintcius would it make sense to add a `sparsenByMod` unit test that's `pending` that exhibits the bug you found?

rintcius:
@alissapajer - what @djspiewak mentioned indeed. Just some background info, I left the value at 20 (i.e. non pending test) to show that the test is working similar to the previous version.
I was thinking an extra test would make sense in case this bug isn't fixed soon, but I assumed it's important enough to fix soon.

rintcius:
created ch994

PR #3739, opened 2018-08-03
add NoOp command in repl 

rintcius:
add NoOp command in repl and avoid escaping in regex where possible


rintcius:
@djspiewak I was busy with repl code & remembered your feature (or rather revision) request :)

PR #3740, opened 2018-08-03
Log plugin version when loading

jsantos17:
[ch939]

this should help debug version incompatibilities.

### DIFF
--- a/impl/src/main/scala/quasar/impl/external/ExternalDatasources.scala
+++ b/impl/src/main/scala/quasar/impl/external/ExternalDatasources.scala
@@ -152,7 +152,7 @@ object ExternalDatasources extends Logging {
       backendModuleAttr <- jarAttribute[F](mainJar, Plugin.ManifestAttributeName)
       versionModuleAttr <- jarAttribute[F](mainJar, Plugin.ManifestVersionName)
 
-      (_: Stream[F, Nothing]) <- versionModuleAttr match {
+      _ <- versionModuleAttr match {
         case None => warnStream[F](s"No '${Plugin.ManifestVersionName}' attribute found in Manifest for '$pluginFile'.", None)
         case Some(version) => infoStream[F](s"Loading $pluginFile with version $version")
       }
@@ -180,6 +180,6 @@ object ExternalDatasources extends Logging {
   private def warnStream[F[_]: Sync](msg: => String, cause: Option[Throwable]): Stream[F, Nothing] =
     Sync[Stream[F, ?]].delay(cause.fold(log.warn(msg))(log.warn(msg, _))).drain
 
-  private def infoStream[F[_]: Sync](msg: => String): Stream[F, Nothing] =
-    Sync[Stream[F, ?]].delay(log.info(msg)).drain
+  private def infoStream[F[_]: Sync](msg: => String): Stream[F, Unit] =
+    Sync[Stream[F, ?]].delay(log.info(msg))
 }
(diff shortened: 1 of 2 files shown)


impl/src/main/scala/quasar/impl/external/ExternalDatasources.scala
```diff
@@ -149,12 +149,13 @@ object ExternalDatasources extends Logging {
...
-      backendModuleAttr <-
-        S.delay(Option(
-          mainJar
-            .getManifest
-            .getMainAttributes
-            .getValue(Plugin.ManifestAttributeName)))
+      backendModuleAttr <- jarAttribute[F](mainJar, Plugin.ManifestAttributeName)
+      versionModuleAttr <- jarAttribute[F](mainJar, Plugin.ManifestVersionName)
+
+      (_: Stream[F, Nothing]) <- versionModuleAttr match {
```
jsantos17: Not sure why I need this type annotation here, but it doesn't build without it.

wemrysi: What error is raised without it?

impl/src/main/scala/quasar/impl/external/Plugin.scala
```diff
@@ -36,6 +36,7 @@ final case class Plugin(mainJar: Path, classPath: ClassPath) {
 object Plugin {
   val FileExtension = "plugin"
   val ManifestAttributeName = "Datasource-Module"
+  val ManifestVersionName = "Specification-Version"
```
wemrysi: Are we setting this in `quasar-s3`? What value does it have?

jsantos17: Not explicitly. I think this is set by `sbt package`. But now that you mention it I'm not sure this will be included by all datasources. Maybe we need to add a `Datasource-Version` key?

wemrysi: I saw that as well, it seems like we'd want `Specification-Version` to be the `quasar` `major.minor` version compiled against and then we could actually use that when loading plugins to compare against the running quasar to determine compatability (assuming we trust our versioning discipline).

Is the `Implementation-Version` set? Or `Package-Version`, one of those seems more appropriate to use for the version of the plugin itself.

wemrysi: Maybe something like
```
Specification-Title: Quasar Datasource
Specification-Vendor: SlamData, Inc
Specification-Version: 59.4
```
And then we could either be pessimistic and just compare against the current version or we could keep track of what quasar version contained the last change that affected the `Datasource` API in any way.

jsantos17: `Implementation-Version` certainly seems more appropriate and it's set. I'll change it to that. 

It looks like this now:

```
Manifest-Version: 1.0
Implementation-Title: quasar-s3
Implementation-Version: 0.2.1
Specification-Vendor: SlamData Inc.
DataSource-Module: quasar.physical.s3.S3DataSourceModule$
Specification-Title: quasar-s3
Implementation-Vendor-Id: com.slamdata
Specification-Version: 0.2.1
Implementation-URL: https://github.com/slamdata/quasar-s3
Implementation-Vendor: SlamData Inc.
```

For this PR I think I'll just change it to read `Implementation-Version` rather than `Specification-Version`

PR #3741, opened 2018-08-03
Corrected future-derived IO continuations remaining on wrong pool

djspiewak:
This turned out to be quite easy to knock off. General rule: whenever we're touching `ExecutionContext`-involved code, try to keep pushing it towards the boundaries. Ultimately, we want to create these things when we create `Quasar` and thread them all the way through. We're using `global` in quite a few places right now, which is definitely not ideal, though not an urgent thing to fix.

[ch591]

### DIFF
--- a/foundation/src/main/scala/quasar/contrib/cats/effect/package.scala
+++ b/foundation/src/main/scala/quasar/contrib/cats/effect/package.scala
@@ -16,16 +16,26 @@
 
 package quasar.contrib.cats
 
+import slamdata.Predef.AnyVal
+
 import cats.effect._
+import cats.syntax.functor._
+
+import scala.concurrent.Future
 
 package object effect {
 
-  implicit class toOps[F[_], A](fa: F[A]) {
+  implicit class toOps[F[_], A](val fa: F[A]) extends AnyVal {
     def to[G[_]](implicit F: Effect[F], G: Async[G]): G[A] =
       Async[G].async { l =>
         Effect[F].runAsync(fa)(c =>
           IO(l(c))
         ).unsafeRunSync
       }
+  }
+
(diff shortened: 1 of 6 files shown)


wemrysi:
@djspiewak With these changes, looks like the only non-test code that uses `global` is `LocalDatasourceModule`, `repl/Main.scala` and `datagen/Main.scala` which seems reasonable.

I'm not sure whether it is more appropriate to hand datasource modules an `EC` or rely on them to provide their own? More involved plugins will likely be creating threadpools (think the `mongodb` or `postgresql` async drivers, etc).

wemrysi:
Oh, and `NIHDBActor`.

djspiewak:
> With these changes, looks like the only non-test code that uses global is LocalDatasourceModule, repl/Main.scala and datagen/Main.scala which seems reasonable.

Oh, I didn't look; we create our own `ExecutionContext` in `Quasar`? To be clear, the problem with `global` is not really that it's bad per se, but that a *lot* of random code can use it in dependencies and things. It's relatively normal for people to misuse that pool, causing starvation in unrelated code.

> I'm not sure whether it is more appropriate to hand datasource modules an EC or rely on them to provide their own? More involved plugins will likely be creating threadpools (think the mongodb or postgresql async drivers, etc).

I think what we want to do is hand them the CPU pool and allow them to create their own pools which handle their own blocking operations, etc. The CPU pool should always be global across the entire JVM (well, the subset of the JVM running code that we can trust not to abuse it, anyway). If they have their own blocking operations or a need for a custom async dispatcher, they can construct and manage that locally.

wemrysi:
`Quasar` accepts an implicit `ExecutionContext`.

PR #3743, opened 2018-08-06
Fixed post-projection of sparsenByMod index when mod > 64

djspiewak:
Assigning to both @rintcius and @alissapajer in the interest of urgency (I don't know what Rintcius' schedule looks like for the rest of the day). Either of you should feel free to merge. Any review comments made post-merge will be addressed in a follow-up PR. I'll try to address anything pre-merge as quickly as possible.

[ch994]


rintcius:
@djspiewak Just having a quick view now, I'll leave the rest to @alissapajer . The IT `veryHeterogeneousShiftArrayValueWithFieldOnMap.test`  is still pending after the fix?

PR #3744, opened 2018-08-07
Enable preserving top K keys during SST compression

wemrysi:


alissapajer:
@wemrysi is there a clubhouse story for this?

impl/src/test/scala/quasar/impl/schema/ExtractSstSpec.scala
```diff
@@ -189,7 +185,51 @@ final class ExtractSstSpec extends quasar.Qspec {
...
+        IMap(
+          J.str("a") -> fourOnes,
+          J.str("b") -> fourOnes),
+        Some((envT(
+          TypeStat.coll(Real(4), Real(3).some, Real(4).some),
+          TagST[J](Tagged(
+            strings.StructuralString,
+            envT(
+              TypeStat.coll(Real(4), Real(3).some, Real(4).some),
+              TypeST(TypeF.arr[J, S](IList(
```
alissapajer: This question is probably outside the scope of this PR, but I'm wondering what product use, if any, keeping track of the frequency of the characters at each string index in the object key names has? Is this a remnant of the original design of SSTs? It seems the frontend only makes use of the entire keys we preserve, and telling them that keys start with `'b'` more frequently than `'q'` might not be useful? I'm wondering if we want to continue preserving this information and if SST generation would be faster without it?

wemrysi:
@alissapajer there is (ch669), but I don't want this PR to close it b/c we need to still expose it in sdbe, so I was going to have _that_ PR close it.

PR #3745, opened 2018-08-07
Only warn when loading fails for all datasource kinds

wemrysi:
Previously we warned on any, which meant we always warned.


alissapajer:
Does this also take care of ch490 ?

wemrysi:
[ch490]

wemrysi:
@alissapajer it does indeed! Thanks for noticing :)

PR #3747, opened 2018-08-07
[WIP] Delete quasar's binary data type

alissapajer:


alissapajer:
@wemrysi I'll rebase this after we figure out what's wrong with one of the tests. But it is up for review.

sst/src/test/scala/quasar/sst/CompressionSpec.scala
```diff
@@ -364,20 +361,20 @@ final class CompressionSpec extends quasar.Qspec
 
   "narrowUnion" >> {
     "reduces the largest group of values having the same primary type" >> prop {
```
alissapajer: This test spins indefinitely locally and I'm not sure why. Any ideas?

alissapajer:
I also can't run all the tests easily because of that one test spinning, so I want to let travis have a go at this.

alissapajer:
@wemrysi in particular if you could give the `sst` changes a careful review, that'd be ace.

wemrysi:
@alissapajer geeze, I wonder if that is due to minification taking forever?

ejson/src/main/scala/quasar/ejson/Z85.scala
```diff
@@ -1,80 +0,0 @@
...
-
-package quasar.ejson
-
-import slamdata.Predef._
-
-import monocle.Prism
-import scalaz._, Scalaz._
-import scodec.bits.ByteVector
-
-package object z85 {
```
wemrysi: :tada:

sst/src/test/scala/quasar/sst/CompressionSpec.scala
wemrysi: It spins because it fails and I'm guessing scalacheck is trying really hard to minimize.

It fails because strings aren't compressed into `SimpleType.Str` by default, they're compressed into a `char[]`, so the expectation is wrong. To make the test pass, I'd swap the roles of `String` and `Char` in the test, using an `ISet[Char]` to exercise the reduction and `s1, s2, s3` as `String`s we'd expect to be preserved.

alissapajer: Thanks, I got it working locally! Just pushed up the change. When travis passes, I'll close this and then then reopen it rebased into a single commit (because I'd like this change to be a single unit).

alissapajer: Minimization is a good thought for how it spends its 17 minutes.

PR #3749, opened 2018-08-13
support setting mode in repl

rintcius:
* repl is now expecting stream of `MimirRepr` instead of `Data`
* new command `set mode = file | console`;  `console` is the default and more or less the same as before (though rewritten in terms of `MimirRepr`); `file` will stream the results to a temp file and should not be memory bound, except when `format = table`
* default `format` is now `precise` instead of `table`

### DIFF
--- a/repl/src/main/scala/quasar/repl/Evaluator.scala
+++ b/repl/src/main/scala/quasar/repl/Evaluator.scala
@@ -32,10 +32,13 @@ import quasar.fp.ski._
 import quasar.frontend.data.DataCodec
 import quasar.impl.schema.{SstConfig, SstSchema}
 import quasar.mimir.MimirRepr
+import quasar.precog.common.{ColumnRef, CNull}
 import quasar.run.{QuasarError, MonadQuasarErr, Sql2QueryEvaluator, SqlQuery}
 import quasar.run.ResourceRouter.DatasourceResourcePrefix
 import quasar.run.optics.{stringUuidP => UuidString}
 import quasar.sst._
+import quasar.yggdrasil.table.{Column, CScanner}
+
 
 import java.lang.Exception
 import java.nio.CharBuffer
@@ -414,7 +417,9 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
         case OutputFormat.Readable =>
           convert(repr.table.renderJson(precise = false))
         case OutputFormat.Csv =>
-          convert(repr.table.renderCsv(assumeHomogeneous = true))
+            import repr.P.trans._
+            val table2 = repr.table.transform(Scan(TransSpec1.Id, NullRemover): TransSpec[Source1])
+            convert(table2.renderCsv(assumeHomogeneous = true))
       }
     }
 
@@ -433,6 +438,23 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
       s.through(fs2.text.utf8Encode)
         .through(fs2.io.file.writeAll(path))
         .compile.drain
+
+    // duplicated from slamdata-backend
+    // someday, this should probably move into mimir...
+    private object NullRemover extends CScanner {
+      type A = Unit
+      val init = ()
+
+      def scan(u: Unit, cols: Map[ColumnRef, Column], range: Range): (Unit, Map[ColumnRef, Column]) = {
+        val cols2 = cols flatMap {
+          case (ColumnRef(_, CNull), _) => Nil
+          case pair => List(pair)
+        }
+
+        ((), cols2)
+      }
+    }
+
 }
 
 object Evaluator {
--- a/repl/src/main/scala/quasar/repl/Evaluator.scala
+++ b/repl/src/main/scala/quasar/repl/Evaluator.scala
@@ -139,6 +139,10 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
         stateRef.modify(_.copy(format = fmt)) *>
           F.pure(s"Set output format: $fmt".some)
 
+      case Mode(mode) =>
+        stateRef.modify(_.copy(mode = mode)) *>
+          F.pure(s"Set output mode: $mode".some)
+
       case SetPhaseFormat(fmt) =>
         stateRef.modify(_.copy(phaseFormat = fmt)) *>
           F.pure(s"Set phase format: $fmt".some)
@@ -417,6 +421,7 @@ object Evaluator {
       |  (explain | compile) [query]
       |  set debug = 0 | 1 | 2
       |  set format = table | precise | readable | csv
+      |  set mode = console | file
       |  set summaryCount = [rows]
       |  set [var] = [value]
       |  env
--- a/repl/src/main/scala/quasar/repl/Evaluator.scala
+++ b/repl/src/main/scala/quasar/repl/Evaluator.scala
@@ -21,6 +21,7 @@ import slamdata.Predef._
 import quasar.api._, datasource._, resource._
 import quasar.common.{PhaseResultListen, PhaseResultTell, PhaseResults}
 import quasar.common.data.Data
+import quasar.contrib.fs2.convert.fromStreamT
 import quasar.contrib.iota._
 import quasar.contrib.pathy._
 import quasar.contrib.std.uuid._
@@ -30,15 +31,18 @@ import quasar.fp.minspace
 import quasar.fp.ski._
 import quasar.frontend.data.DataCodec
 import quasar.impl.schema.{SstConfig, SstSchema}
+import quasar.mimir.MimirRepr
 import quasar.run.{QuasarError, MonadQuasarErr, Sql2QueryEvaluator, SqlQuery}
 import quasar.run.ResourceRouter.DatasourceResourcePrefix
 import quasar.run.optics.{stringUuidP => UuidString}
 import quasar.sst._
 
 import java.lang.Exception
+import java.nio.CharBuffer
 import scala.util.control.NonFatal
 
 import argonaut.{Json, JsonParser, JsonScalaz}, JsonScalaz._
+import cats.arrow.FunctionK
 import cats.effect._
 import eu.timepit.refined.refineV
 import eu.timepit.refined.api.Refined
@@ -57,7 +61,7 @@ import spire.std.double._
 final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResultTell](
     stateRef: Ref[F, ReplState],
     sources: Datasources[F, Stream[F, ?], UUID, Json, SstConfig[Fix[EJson], Double]],
-    queryEvaluator: QueryEvaluator[F, SqlQuery, Stream[F, Data]]) {
+    queryEvaluator: QueryEvaluator[F, SqlQuery, Stream[F, MimirRepr]]) {
 
   import Command._
   import DatasourceError._
@@ -227,8 +231,13 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
         stateRef.get.map(s => printPath(s.cwd).some)
 
       case Select(q) =>
-        def convert(format: OutputFormat, s: Stream[F, Data]): F[Option[String]] =
-          s.compile.toList.map(ds => renderData(format, ds).some)
+        def convert(format: OutputFormat, s: Stream[F, MimirRepr]): F[Option[String]] = {
+          val rs: Stream[F, String] = s.flatMap { repr =>
+            val rendered = render(format, repr)
+            fromStreamT(rendered).map(_.toString).translate(λ[FunctionK[IO, F]](_.to[F]))
+          }
+          rs.compile.toList.map(_.mkString("\n").some)
+        }
 
         for {
           state <- stateRef.get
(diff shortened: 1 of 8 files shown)


repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -344,6 +364,30 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
         case PhaseFormat.Code => results.map(_.showCode).mkString("\n\n")
       }
 
+    private def printQueryResults(maxLines: Option[Int], stream: Stream[F, String]): F[String] = {
```
rintcius: I put this code here for now, using `Table.renderCsv` as is but if we need this use-case in performance-critical prod code (outside repl) then we should think about adding an option to emit each line (as well), so that we don't have to inspect the items in the stream

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -375,6 +434,27 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
...
       path.fold(f => fileParent(f) </> dir(fileName(f).value), rootDir)
 
+    private def writeToPath(path: JPath, s: Stream[F, String]): F[Unit] =
+      s.through(fs2.text.utf8Encode)
+        .through(fs2.io.file.writeAll(path))
+        .compile.drain
+
+    // duplicated from slamdata-backend
+    // someday, this should probably move into mimir...
+    private object NullRemover extends CScanner {
```
rintcius: duplicated for now, but happy to move this into mimir

rintcius:
[ch1146]

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -356,17 +400,32 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
             t.getStackTrace.mkString("\n  ", "\n  ", "")).some
       }
 
-    private def renderData(format: OutputFormat, ds: List[Data]): String =
-      (format match {
+    private def renderMimirRepr(format: OutputFormat, repr: MimirRepr): Stream[F, String] = {
```
djspiewak: Why is this converting to `String`? `CharBuffer` is correct and should be retained whenever going to file.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -356,17 +400,32 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
...
+          convert(repr.table.renderJson(precise = true))
         case OutputFormat.Readable =>
-          ds.map(formatJson(DataCodec.Readable)).unite
+          convert(repr.table.renderJson(precise = false))
         case OutputFormat.Csv =>
-          Prettify.renderValues(ds).map(CsvWriter(none)(_).trim)
-      }).mkString("\n")
+            import repr.P.trans._
+            val table2 = repr.table.transform(Scan(TransSpec1.Id, NullRemover): TransSpec[Source1])
+            convert(table2.renderCsv(assumeHomogeneous = true))
```
djspiewak: We should have two modes: `Csv` and `HomogeneousCsv`, since they really are separate renderers. We can't assume that arbitrary REPL queries are going to be homogeneous.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -375,6 +434,27 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
     private def toADir(path: ResourcePath): ADir =
       path.fold(f => fileParent(f) </> dir(fileName(f).value), rootDir)
 
+    private def writeToPath(path: JPath, s: Stream[F, String]): F[Unit] =
+      s.through(fs2.text.utf8Encode)
+        .through(fs2.io.file.writeAll(path))
```
djspiewak: This should use `writeAllAsync` rather than `writeAll`, which results in blocking threads.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -375,6 +434,27 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
     private def toADir(path: ResourcePath): ADir =
       path.fold(f => fileParent(f) </> dir(fileName(f).value), rootDir)
 
+    private def writeToPath(path: JPath, s: Stream[F, String]): F[Unit] =
```
djspiewak: Input here should be `Stream[F, CharBuffer]`.

repl/src/main/scala/quasar/repl/Evaluator.scala
djspiewak: It's fine to duplicate for now. Do we have a ticket for moving it to mimir? I can't remember…

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -227,16 +239,24 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
...
+            res <- state.mode match {
+            case OutputMode.Console =>
+              val maxLines = state.summaryCount
+                .map(_.value + OutputFormat.headerLines(state.format))
+              printQueryResults(maxLines, rendered).map(log + _)
+            case OutputMode.File =>
+              for {
+                tmpFile <- Paths.createTempFile("results", ".txt")
+                _ <- writeToPath(tmpFile, Stream(log).covary[F] ++ rendered)
+              } yield s"Results written to ${tmpFile.toFile.getAbsolutePath}"
```
djspiewak: I believe this results in printing the `"Results written to ..."` line *after* the writing has fully completed! This makes it much, much less useful. The file we're sending results to needs to be printed *before* the rendering starts so that we can `tail -f` the file in a separate buffer. Right now, I believe what will happen is the rendering will start and the repl will effectively freeze while rendering proceeds, giving no indication that anything is actually evaluating.

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: oh.. silly mistake.. I actually meant `assumeHomogeneous = false`. You actually want both or is it ok to just change `true` to `false`?

djspiewak: I do want both if possible, since they're separate renderers. We're going to want to be able to easily test both.

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: That was because I didn't find anything in fs2 lib for files based on `CharBuffer`. I'll have a look at writing it myself (or let me know if I've missed something)

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: Right, i thought about that but as a nice to have. Maybe best to make the file name configurable then, eg via `set file = ...`?

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: ok will add both

repl/src/main/scala/quasar/repl/Evaluator.scala
alissapajer: It's [here](https://github.com/slamdata/slamdata-backend/blob/master/server/src/main/scala/slamdata/backend/server/encoders.scala#L102) in sdbe

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: main thing is that it doesn't conform to the normal sequence in `read evaluate print loop`.. so need to implement something that works around that (maybe breaking the command in 2 or something).

what do you find suboptimal about it?

djspiewak: If `evaluate` produces `Stream[F, String]` which is then printed (so `print: Stream[F, String] => F[Unit]`), it should work within that same structure.

It's suboptimal because you need to constantly come up with new files to drop things into. It makes the common case (quickly testing a query multiple times) harder and more cumbersome.

rintcius: hmm interesting idea... I'll have a look if I can make that work

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: I created a ticket now, ch1258

repl/src/main/scala/quasar/repl/Evaluator.scala
rintcius: Copied @alissapajer's  impl to quasar, so we can reuse that in SDBE after this PR lands.

rintcius:
@djspiewak I think I have addressed your comments. Your proposal to make `evaluate` in repl streaming worked out nicely in the end.

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -212,45 +229,63 @@ final class Evaluator[F[_]: Effect: MonadQuasarErr: PhaseResultListen: PhaseResu
...
+                  .map(_.value + OutputFormat.headerLines(state.format))
+                Stream.emit(log) ++ printQueryResults(maxLines, rendered.map(_.toString))
+              }
+            case OutputMode.File =>
+              Paths.createTempFile("results", ".txt").map { tmpFile =>
+                Stream.emit(s"Writing results to ${tmpFile.toFile.getAbsolutePath}") ++
+                  Stream.eval {
+                    for {
+                      (log, rendered) <- doSelect(sql, state)
+                      _ <- writeToPath(tmpFile, Stream(CharBuffer.wrap(log)).covary[F] ++ rendered)
```
djspiewak: Sidebar… We could someday get cute here and display a bit of a spinner if we felt like being cool. Completely unnecessary and not worth spending time on, but it would be fun. :-D

repl/src/main/scala/quasar/repl/Evaluator.scala
```diff
@@ -408,6 +515,7 @@ object Evaluator {
       |  (explain | compile) [query]
       |  set debug = 0 | 1 | 2
       |  set format = table | precise | readable | csv
```
djspiewak: This should be updated to include `homogeneouscsv`

repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -44,24 +45,23 @@ import scalaz._, Scalaz._
...
   private val read: F[Command] = F.delay(Command.parse(reader.readLine(prompt)))
 
-  private def eval(cmd: Command): F[Evaluator.Result] =
+  private def eval(cmd: Command): F[Evaluator.Result[Stream[F, String]]] =
     F.start(evaluator(cmd)) >>= (_.join)
 
-  private def print(string: Option[String]): F[Unit] =
-    string.fold(F.unit)(s => F.delay(println(s)))
+  private def print(strings: Stream[F, String]): F[Unit] =
+    strings.observe1(s => F.delay(println(s))).compile.drain
```
djspiewak: Minor note, you can also achieve this in the following way:

```scala
strings.map(_.toBytes).observe(fs2.io.stdout).compile.drain
```

Doesn't achieve anything, but it's kind of cute. :-) Your call as to whether you like it better than the existing line; either is fine IMO.

rintcius: I like `fs2.io.stdout` but choose `observe1` above `observe` because that has better performance, see e.g. https://github.com/functional-streams-for-scala/fs2/blame/ee6f700fcab9c1e72c3e83890ced60a51a6e865a/docs/migration-guide-1.0.md#L121

rintcius: BTW... any disadvantages if we make `renderCsv`/ `renderJson`  return a `ByteBuffer` instead of `CharBuffer`? That saves us from having to do the extra encoding in both SDBE and repl

djspiewak: > I like fs2.io.stdout but choose observe1 above observe because that has better performance

Performance doesn't really matter here, so it's really whatever you feel is cleanest. This is actually a general rule of thumb for interacting with fs2: performance matters, but only to a point. Most things that we're doing with fs2 are outside the pale of the hot path. This is particularly true of the repl. So you generally would want to pick the mode of interaction with fs2 which is cleanest and most maintainable, not necessarily the one that yields the best performance. Obviously, this only holds to a point. If something is 10000x slower than an alternative, you should probably pick the faster one, but if something is 2x slower *in fs2*, it probably doesn't matter much, and I believe `observe` is much less than 2x slower than `observe1`.

> BTW... any disadvantages if we make renderCsv/ renderJson return a ByteBuffer instead of CharBuffer?

Significant disadvantages. This would require both those functions to take a `Charset` and deal with character encoding. This adds complexity to a pair of functions which are already somewhat complex, particularly when it comes to dealing with chunk boundaries and multi-byte codepoints (e.g. 🤦🏼‍♂️). It does avoid an intermediate copy of the bytes, which might not be an insignificant win and so we might end up doing it eventually, but it's almost certainly not something we want to worry about right now.

PR #3750, opened 2018-08-13
catch LinkageError and raise as IncompatibleDatasourceException

rintcius:
[ch934]

### DIFF
--- a/connector/src/main/scala/quasar/connector/package.scala
+++ b/connector/src/main/scala/quasar/connector/package.scala
@@ -16,8 +16,6 @@
 
 package quasar
 
-import slamdata.Predef._
-import quasar.api.datasource.DatasourceType
 import quasar.contrib.scalaz.MonadError_
 
 package object connector {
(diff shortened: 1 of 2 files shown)


connector/src/main/scala/quasar/connector/package.scala
```diff
@@ -16,11 +16,17 @@
...
+import quasar.api.datasource.DatasourceType
 import quasar.contrib.scalaz.MonadError_
 
 package object connector {
   type MonadResourceErr[F[_]] = MonadError_[F, ResourceError]
 
   def MonadResourceErr[F[_]](implicit ev: MonadResourceErr[F])
       : MonadResourceErr[F] = ev
+
+  final case class IncompatibleDatasourceException(kind: DatasourceType) extends java.lang.RuntimeException {
```
wemrysi: I think we can define this in `quasar.impl.datasources` or even the `DatasourceManagement` companion as it is implementation specific.

PR #3756, opened 2018-08-15
Misc precog reorg and deletions

alissapajer:
Mostly because I find this a relaxing way to end my day :)


blueeyes/src/main/scala/quasar/blueeyes/LazyMap.scala
```diff
@@ -14,15 +14,24 @@
  * limitations under the License.
  */
 
-package quasar.precog
+package quasar.blueeyes
```
alissapajer: These two files had nothing in common. Not sure why github thinks they did, characterizing this as a file rename.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -42,6 +42,7 @@ import java.time.{LocalDate, LocalDateTime, LocalTime, OffsetDateTime, OffsetTim
 
 import scala.annotation.tailrec
 import scala.collection.immutable.Set
+import scala.concurrent.ExecutionContext.Implicits.global
```
djspiewak: This should be an abstract `implicit def`. I believe we have this in a couple other places as well. It is populated by the `Precog` concrete class: https://github.com/djspiewak/quasar/blob/master/mimir/src/main/scala/quasar/mimir/Precog.scala#L56 We shouldn't be bringing in the global `ExecutionContext` except in tests.

alissapajer:

PR #3757, opened 2018-08-15
Collapse specs2 test results in Travis logs

djspiewak:
Don't merge this yet. Let's see what Travis does…

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,6 +45,20 @@ lazy val buildSettings = Seq(
 
   console := { (console in Test).value },
 
+  Global / testFrameworks := {
+    if (isTravisBuild.value)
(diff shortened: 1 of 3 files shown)


djspiewak:
@alissapajer Something is still wrong with the parallel interleaving I think. It’s collapsing but in broken ways. Seems fixable in the morning.

alissapajer:
@djspiewak I'm still not seeing any tucking :/

djspiewak:
@alissapajer I think it might be bugging out on Travis' end. We're printing the appropriate markers and doing so sequentially, but Travis isn't rendering it properly. I see the tucks being identified, and I can click on them to explicitly tuck, though the browser pane kind of freaks out a bit. I'll close this for now. We can play with it later.

PR #3758, opened 2018-08-16
Collapse shifts with the same locus

jsantos17:
[ch257]

I had to change the typers to be less precise in order to get fewer shifts. With the typechecks in `master` it's not possible to coalesce `a{:_}, a{:_}{_:}` into two shifts. The first `a{:_}` typechecks checks the type `Obj(Map[], Some(Top))`, while the second `a{:_}` typechecks `Obj(Map[], Some(Obj(Map[], Some(Top))))`

Also fixed all the calls to `linearize` in `MinimizeAutoJoinsSpec` since they were bothering me.


djspiewak:
Assignee and version, please. Also hype! Emrys thinks this might remove *all* of the cases where we were joining in ExpandShifts, which is exciting to say the least. It means we no longer will generate sparse slices coming out of shifts since we will no longer be inserting Conds.

frontend/src/main/scala/quasar/std/structural.scala
```diff
@@ -312,7 +312,7 @@ trait StructuralLib extends Library {
           failureNel(invalidArgumentError("internal error: arrayLike, but no arrayType")))(
           success)
     },
-    untyper[nat._1](tpe => success(Func.Input1(FlexArr(0, None, tpe)))))
+    untyper[nat._1](tpe => success(Func.Input1(FlexArr(0, None, Top)))))
```
djspiewak: Hmm, is there a way we can just strip the typechecks from the `LeftShift` struct instead (e.g. when we do the equality comparison)? If this really is faster, then it's ok for now, it just makes a bit more work for us in future when we do projection pushdown. I want to avoid having to redo the typer in QSU in the short-term.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -436,11 +440,56 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
         }
       }
 
+      def fixCompatible(repair: FreeMapA[ShiftTarget[T]], idStatus: IdStatus, resultIdStatus: IdStatus): FreeMapA[ShiftTarget[T]] =
```
djspiewak: Is this ever called with `resultIdStatus != IncludeId`?

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -436,11 +440,56 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+
       (left, right) match {
         case
-          (
-            -\/(QSU.LeftShift(fakeParent, structL, idStatusL, _, repairL, rotL)) :: tailL,
-            -\/(QSU.LeftShift(_, structR, idStatusR, _, repairR, rotR)) :: tailR) =>
+            (
+              (lshift @ -\/(QSU.LeftShift(fakeParent, structL, idStatusL, onUndefinedL, repairL, rotL))) :: tailL,
+              (rshift @ -\/(QSU.LeftShift(_, _, idStatusR, onUndefinedR, repairR, _))) :: tailR)
+          if compatibleShifts(lshift, rshift) =>
```
djspiewak: If this is being done pair-wise, does it end up being subject to join reassociation?  For example:

```sql
select a{_}, b{_}, a{_:} from foo
```

Will the two `a` shifts rewrite together in this branch?

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -1537,6 +1538,51 @@ object MinimizeAutoJoinsSpec extends Qspec with TreeMatchers with QSUTTypes[Fix]
...
+            func.StaticMapS(
+              "0" -> func.ProjectIndexI(RightTarget[Fix], 0),
+              "1" -> func.ProjectIndexI(RightTarget[Fix], 1)))
+
+          outerMap must beTreeEqual(
+            recFunc.StaticMapS(
+              "0" -> recFunc.ProjectKeyS(recFunc.Hole, "0"),
+              "1" -> recFunc.ProjectKeyS(recFunc.Hole, "1")))
+      }
+    }
```
djspiewak: 🎉

djspiewak:
Another question: will this also work for identical shifts at the same locus? I'm thinking of something like this:

```sql
select a{_}.c, b{_}.d from foo
```

That should be one `LeftShift`, ideally with `ExcludeId`. Does that work?

frontend/src/main/scala/quasar/std/structural.scala
wemrysi: Yeah, I'd also prefer handling this locally, if possible.

@djspiewak speaking of typechecks, do we actually need them at all? Our only evaluator is `mimir` and it doesn't need them, no?

djspiewak:
@wemrysi I did. Whoops. Editing for clarity.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
jsantos17: Yes! Specifically in the query mentioned by @wemrysi 

```
select a{_}.c, a{_}.d from foo
```

The result of the coalesce should be `ExcludeId` and the repairs should not include `ProjectIndex` in this case. This was a tricky bug and made a few `it`s fail.

And yes, that query is planned with a single `LeftShift`.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
jsantos17: Yes, it's done pairwise and `select a{:_}, a{_:}, b{_}` plan with two `LeftShift`s, while `a{:_}, b{_}, a{_:}` plans with three. We could sort the list of candidates before attempting the coalesce, to make sure that compatible left shifts are next to each other.

frontend/src/main/scala/quasar/std/structural.scala
jsantos17: This is indeed faster, but eliding the `Guard`s shouldn't take a long time. If it's preferrable to elide the `Guard`s I'm happy to push a new commit with that implementation instead.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
djspiewak: I think we should probably do the sorting. We're going to get a *ton* of queries which require this kind of reassociation.

frontend/src/main/scala/quasar/std/structural.scala
djspiewak: @jsantos17 I think it's definitely preferable to just elide the `Guard` (and `Typecheck`) nodes in the comparison, rather than removing them in the typer.

@wemrysi We don't need the typechecks at all (except in `Filter`), but we will take advantage of them to infer projection pushdown. Or at least, that's the current plan. Leaving them in place allows us to avoid re-writing the typer entirely as part of that effort.

djspiewak:
@jsantos17 Version label. :-)

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -436,11 +440,56 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
         }
       }
 
+      def fixCompatible(repair: FreeMapA[ShiftTarget[T]], idStatus: IdStatus, resultIdStatus: IdStatus): FreeMapA[ShiftTarget[T]] =
+        (idStatus, resultIdStatus) match {
+          case (IdOnly, IncludeId) => repair flatMap {
+            case ShiftTarget.LeftTarget() => scala.sys.error("ShiftTarget.LeftTarget in CollapseShifts")
+            case ShiftTarget.RightTarget() => {
+              val hole = Free.pure[MapFunc, ShiftTarget[T]](ShiftTarget.RightTarget())
```
wemrysi: Alternatively `RightTarget[T]`, which is defined in the `qsu` package object.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -436,11 +440,56 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
         case
-          (
-            -\/(QSU.LeftShift(fakeParent, structL, idStatusL, _, repairL, rotL)) :: tailL,
-            -\/(QSU.LeftShift(_, structR, idStatusR, _, repairR, rotR)) :: tailR) =>
+            (
+              (lshift @ -\/(QSU.LeftShift(fakeParent, structL, idStatusL, onUndefinedL, repairL, rotL))) :: tailL,
+              (rshift @ -\/(QSU.LeftShift(_, _, idStatusR, onUndefinedR, repairR, _))) :: tailR)
+          if compatibleShifts(lshift, rshift) =>
+
+          val idStatusAdj = Semigroup[IdStatus].append(idStatusL, idStatusR)
```
wemrysi: This is fine, but you may find `idStatusL |+| idStatusR` easier on the eye.

djspiewak:
Outstanding. Do you want to still address the `Guard`-removal vs typer implementation in this PR?

jsantos17:
@djspiewak yes, I want to implement `Guard`-remove in this PR. @wemrysi provided me with an example where the typer changes are not enough. Example is `multitypeFlatten.test`

jsantos17:
@djspiewak @wemrysi quick update on this. I tried to discard the `Guard`s and `Typecheck`s in every `LeftShift` struct, but that made ~19 integration tests fail. Even some very basic ones like `select * from nested_foo.data where foo{*} = 15` started failing without `Guard`s. For the simplest cases I verified the only difference are typechecks in the struct. 

So instead of that, I implemented a rewrite that simplifies the typechecks in `struct`s right before `MAJ`. This worked much better, but I still have two integration test failures. The failures are:

- `multitypeFlatten.test`
- `flattenGroupedMap.test`

djspiewak:
@jsantos17 That surprises me a *lot*! (the fact that eliding guards didn't work) Really honestly a lot. Somewhat concerns me, too, since it implies that I'm not fully understanding what's going on. I'll have to investigate that further.

wemrysi:
@djspiewak Yeah, that surprises me too. @jsantos17 were you eliding the typechecks just for comparison purposes or dropping them completely?

jsantos17:
@djspiewak @wemrysi I dropped `Guards` altogether initially, not just for comparison.

wemrysi:
@jsantos17 Ah, I'd try just dropping them for comparison purposes, but leaving them in the query itself.

alissapajer:
@jsantos17 a couple integration tests are failing. Interesting they passed in appveyor...

alissapajer:
Oh that explains why appveyor passed:
```
[info] Sql2QueryRegressionSpec
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
```
Anyone know why we aren't running these on appveyor?

jsantos17:
@djspiewak this should be ready for another look. There's one failing integration test that's making the build red, but it's fixed by #3767. So I suggest we wait for that to merge before merging this PR.

alissapajer:
Yay! I'm finished up the review on #3767 now!

alissapajer:
@jsantos17 #3767 is in!

jsantos17:
Merged `master` on this branch. Next build should be green :)

alissapajer:
Yay it's green!

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -668,14 +713,103 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+      val cs0: List[(QSUGraph, ShiftDefinition)] =
+        toReorder.map(c => firstShiftDefinition(c) match {
+          case Some((shifting, struct, rot)) =>
+            (c, (shifting, elideGuards(struct), rot)).some
+          case None =>
+            none
+        }).unite
+
+
+      // FIXME: quadratic complexity.
```
alissapajer: What kind of queries are going to hit this `FIXME`? Should we make an issue for it or do you think the comment suffices?

jsantos17: It's quadratic on the number of candidates we have to coalesce. Something like:

```
select a_1[*], a_2[*]... a_n[*] from foo
```

For this query we will need to do up to n^2 comparisons. Ideally I'd use something like `Map[(QSUGraph, FreeMap. Rotation), QSUGraph]` but the hashing doesn't work because of the `FreeMap`s. I also tried scalaz's `IMap`, but that has an `Ord` constraint on `K`.

We could fix this by splitting the struct into a separate `Map` node and then use `Map[(Symbol, Symbol, Rotation), QSUGraph]`. I can open a ticket about this.

wemrysi: If we switch to using provenance to determine compatibility, we could define and ordering for that and be able to use `IMap` or even `stdlib.Map` as I think it's case classes all the way down.

jsantos17: @wemrysi can you please add details on that implementation in the [Clubhouse ticket](https://app.clubhouse.io/data/story/1555/avoid-quadratic-comparisons-when-coalescing-leftshifts)? Really curious

qscript/src/main/scala/quasar/qscript/rewrites/Normalizable.scala
```diff
@@ -132,6 +132,10 @@ class NormalizableT[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] extends T
   def freeMF[A: Equal: Show](fm: Free[MapFunc, A]): Free[MapFunc, A] =
     fm.transCata[Free[MapFunc, A]](MapFuncCore.normalize[T, A])
 
+  // freeMF without the Show constraint
+  def freeMF0[A: Equal](fm: Free[MapFunc, A]): Free[MapFunc, A] =
```
alissapajer: Was the problem that you didn't want to have to thread the `Show` constraint through a bunch of new places?

jsantos17: We didn't have a`Show[ShiftTarget]`, but I just noticed #3767 introduced the instance. Will push a commit.

jsantos17: Fixed!

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
djspiewak: @wemrysi Now that provenance includes path, I think we probably *could* use provenance to determine compatibility rather than approximating it by looking at the hole and the struct.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -607,7 +652,7 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
               sym,
               shiftsRAdj,
               OnUndefined.Emit,
-              repair)
+              N.freeMF(repair))
```
alissapajer: I'm a little wary of an approach that normalizes in specific places like this, rather than an approach that takes a qscript and traverses it to normalize all the `FreeMap`s inside. Would I be correct to assume that if we don't normalize here, we wouldn't hit other optimizations/coalescings down the road?

So, the question I'm pondering is if it'd be better to do a single normalization pass after `CollapseShifts` instead.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -668,14 +713,103 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+      C: Corecursive.Aux[FreeMap, CoEnv[Hole, MapFunc, ?]]): FreeMap =
+      R.cata[FreeMap](fm) {
+        case CoEnv(\/-(MFD(MapFuncsDerived.Typecheck(result, _)))) => result
+        case CoEnv(\/-(MFC(MapFuncsCore.Guard(_, _, result, _)))) => result
+        case otherwise => C.embed(otherwise)
+      }
+
+    def compatibleShifts(l: ShiftGraph, r: ShiftGraph): Boolean = {
+      (l, r) match {
+        // FIXME: compare the struct symbols, not the FreeMaps
```
alissapajer: Another `FIXME`. Is this one hard to implement now for some reason?

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
wemrysi: @djspiewak Aye, we do exactly this when creating the join conditions between shifts that we failed to coalesce.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
jsantos17: This is very similar to the other `FIXME`. It should be fixed with [ch1555](https://app.clubhouse.io/data/story/1555) too.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
alissapajer: In the abstract I like this provenance approach very much and think it's definitely worth exploring!

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
alissapajer: Cool. Actually, could you add that story number to both these `FIXME`s? That way we can keep track of them!

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -409,11 +418,47 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
         }
       }
 
+      def fixCompatible(repair: FreeMapA[ShiftTarget], idStatus: IdStatus, resultIdStatus: IdStatus): FreeMapA[ShiftTarget] =
+        (idStatus, resultIdStatus) match {
+          case (IdOnly, IncludeId) => repair flatMap {
+            case ShiftTarget.LeftTarget => scala.sys.error("ShiftTarget.LeftTarget in CollapseShifts")
+            case ShiftTarget.RightTarget => func.ProjectIndexI(RightTarget, 0)
+            case access @ ShiftTarget.AccessLeftTarget(_) => Free.pure(access)
```
alissapajer: @wemrysi here is an example of where `LeftTarget` and `AccessLeftTarget(Value)` are being treated differently.

@jsantos17 what is the difference here in these two cases? Why is one an error?

djspiewak: One of them cannot exist prior to `ExpandShifts`. This is kind of an artifact of how that code all came together. Definitely something we want to clean up.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -409,11 +418,47 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+          }
+
+          case (ExcludeId, IncludeId) => repair flatMap {
+            case ShiftTarget.LeftTarget => scala.sys.error("ShiftTarget.LeftTarget in CollapseShifts")
+            case ShiftTarget.RightTarget => func.ProjectIndexI(RightTarget, 1)
+            case access @ ShiftTarget.AccessLeftTarget(_) => Free.pure(access)
+          }
+          case _ => repair
```
alissapajer: Why don't we need to fix the `ShiftTarget` in any other cases. Intuitively if the pair were `(IncludeId, ExcludeId)` for example, would we not have to "fix" anything because we already have the id from `IncludeId`?

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
djspiewak: The nice thing about normalizing mid-way (and generally the only reason we do it) is to make the unit tests cleaner.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
djspiewak: Fun story: when I originally started writing `MinimizeAutoJoins`, I was *trying* to use provenance to determine compatibility, but it wasn't a rich enough representation. I'm pretty sure that it now is.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
jsantos17: It's my understanding that there should never be a `LeftTarget` at this point in any `ShiftGraph`. Those are introduced only in the results of a coalesce.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
alissapajer: That's true that it does make the unit tests for `CollapseShifts` cleaner. But this kind of normalization is not the kind of thing that we're going to explicitly test in `CollapseShifts` and so I think the cost of cleaner tests is that a missed normalization case might slip through. Whereas if we had a separate single-pass normalizer, we could unit test that separately.

djspiewak: There is a separate normalization pass already though. It happens in qscript. So I'm not sure we need another one in QSU.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
alissapajer: :fish: :fire: :fish_cake:

alissapajer: Since I'm sure that wasn't clear, I was celebrating our provenance wins by cooking tasty fish.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
jsantos17: I normalized here since it was incredibly difficult to read and debug most `FreeMap`s without it.

alissapajer: Hm, aside from debugging do we need to normalize here? This feels kind of like a whack-a-mole approach, and the reason I'm wary of that kind of approach is because it would be easy for us to start relying on the normalized `FreeMap`s without realizing it. 

@djspiewak if we need to normalize here (for correctness or optimization), then I was arguing that we should do it in a separate pass so it's more explicit. If we don't need to normalize here, then maybe we shouldn't?

alissapajer: :+1: I think that's totally fair, and I could honestly go either way with this one. :smile:

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -87,10 +89,10 @@ object MinimizeAutoJoinsSpec extends Qspec with TreeMatchers with QSUTTypes[Fix]
...
         case Map(Read(_), fm) =>
           // must_=== doesn't work
-          fm.linearize must beTreeEqual(
-            func.Add(
-              func.ProjectKeyS(func.Hole, "foo"),
-              func.ProjectKeyS(func.Hole, "bar")))
+          fm must beTreeEqual(
+            recFunc.Add(
+              recFunc.ProjectKeyS(recFunc.Hole, "foo"),
+              recFunc.ProjectKeyS(recFunc.Hole, "bar")))
```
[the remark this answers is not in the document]
djspiewak: It's actually kind of difficult to test the MAJ phases individually, so we conventionally unit test them within `MinimizeAutoJoinsSpecs`.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
jsantos17: We append `IdStatus`es to create the left side of the tuple. The only case where the `IdStatus` could be `ExcludeId` or `IncludeId` is when two compatible shifts have both `ExcludeId` or both `IncludeId`. Since the resulting `LeftShift` will have the same `IdStatus` the repairs are still valid.

This case happens only when we coalesce something like `a[*] + a[*].b[*]`

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -668,14 +713,103 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+      }
+
+    def compatibleShifts(l: ShiftGraph, r: ShiftGraph): Boolean = {
+      (l, r) match {
+        // FIXME: compare the struct symbols, not the FreeMaps. ch1555
+        case (-\/(QSU.LeftShift(srcL, structL0, _, _, _, rotL)), -\/(QSU.LeftShift(srcR, structR0, _, _, _, rotR))) => {
+          val structL = elideGuards(structL0.linearize)
+          val structR = elideGuards(structR0.linearize)
+
+          srcL.root === srcR.root && structL === structR && rotL === rotR
```
alissapajer: Hmm, so when evaluating a `LeftShift`, mimir doesn't actually use the `ShiftType` since the object/array distinction is carried along in the `FreeMap`s as guards (I think?). Meaning, I wonder if we actually need to be distinguishing if we're shifting an array or object? At what stage do we encode the object/array distinction?

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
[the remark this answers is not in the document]
jsantos17: @alissapajer re integration tests, I don't think this PR changes the results of any query. I did some quick benchmarks and it definitely improves performance.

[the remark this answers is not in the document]
alissapajer: Awesome!

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
alissapajer: Also, I haven't figured out how this is working given that we're eliding guards above...

jsantos17: I'm not very familiar with mimir, but we definitely need to compare rotations here since we disregard `Guard`s when comparing. The `Guard`s are not removed from the generated QScript, they're only removed to compare them.

djspiewak: @alissapajer I haven't actually checked on it, but we should actually use the `ShiftType` in interpreting `LeftShift` in mimir.

alissapajer: Just ran a query on this branch and on master, and in both places the guard is pulled out into a `Map`, so we're good :+1:

alissapajer: @djspiewak I have checked and we are not using the `ShiftType` in `QScriptCorePlanner`.

djspiewak: Sigh… Ok that's definitely a bug that we should fix. @jsantos17 This is probably why things exploded when you tried to omit the `Guard`s entirely.

alissapajer: Oh yes, that would make much sense!

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
alissapajer: Hm, if the the original `LeftShift` was `IdOnly`, then how could we hit this case?

alissapajer: Oh wait, I was confusing myself. All good.

qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -668,14 +713,103 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+        }
+      }
+
+      // Avoid unnecessary reordering if there are no compatible shifts
+      // to put next to each other.
+      if (cs1.all(_._2 === 1))
+        cs
+      else
+        cs1.sortBy {
+          case (_, freqCount) => -freqCount
```
alissapajer: If there are multiple groups of shifts, each with the same `freqCount`, would this not group them all together, meaning they might not end up grouped correctly?

jsantos17: Yes, that'd generate duplicates, but even with perfect sorting this won't coalesce all compatible shifts. We have 

I posted in Discord and it got lost there:

```
The current implementation is not ideal and misses a lot of cases. As an example of something that the current pairwise approach can't handle nicely: a{:_}, a{_:}, b{:_}, b{_:}. MAJ will go left to right on each shift. It correctly notices the first pair and plans it with a single shift. However, coalescing the first pair with the first b left-shift yields a MultiLeftShift with two structs in it. Coalescing the last b left shift with the previous result yields a MultiLeftShift three structs in it. 

So reordering the candidates doesn't help much.  The correct way to handle this would be to coalesce the a and b pairs separately and then coalesce those two into a single MultiLeftShift with two structs. I haven't thought much about this but seems promising. 
```

I created a ticket about this [here](https://app.clubhouse.io/data/story/1473/improve-coalescing-of-shifts-at-the-same-locus). Anyway, I pushed a commit that also considers the `LeftShift` source when reordering shifts just in case.

jsantos17:
@alissapajer can we merge this?

djspiewak:
@jsantos17 Sorry, board meeting. I'm on it! Taking a final look and merging shortly.

PR #3765, opened 2018-08-21
Persist slices in bulk in NIHDB (bypassing JValue)

djspiewak:
This gives us, among other things, type preservation in preparation results. It should also be orders of magnitude faster.

This relies on the new API, `NIHDB.insertSegmentsVerified`. This API is, in isolation, safe, but you cannot mix it with `NIHDB.insert` or `insertVerified`. If you mix the two, you'll end up losing uncooked data in the append logs. If we need mixing, we can add it down the line, it's just more complicated than what I wanted to deal with right now.

[ch1331]

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
+++ b/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
@@ -60,10 +60,9 @@ final class MimirPTableStore[F[_]: Monad: LiftIO] private (
   def write(key: StoreKey, table: PTable): Stream[F, Unit] = {
     val ios = Stream.bracket(cake.createDB(keyToFile(key)).map(_.toOption))({
       case Some((_, _, db)) =>
-        val can = table.canonicalize(Config.maxSliceRows)
-        val com = can.compact(cake.trans.TransSpec1.Id)
+        val can = table.compact(cake.trans.TransSpec1.Id).canonicalize(Config.maxSliceRows)
 
-        fromStreamT(com.slices).zipWithIndex evalMap {
+        fromStreamT(can.slices).zipWithIndex evalMap {
           case (slice, offset) =>
             val segments = SegmentsWrapper.sliceToSegments(offset.toLong, slice)
             IO.fromFutureShift(
--- a/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
+++ b/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
@@ -24,6 +24,7 @@ import quasar.contrib.pathy.{ADir, AFile}
 import quasar.contrib.scalaz.MonadError_
 import quasar.mimir.MimirCake.Cake
 import quasar.yggdrasil.ExactSize
+import quasar.yggdrasil.nihdb.SegmentsWrapper
 import quasar.yggdrasil.vfs.ResourceError
 import quasar.niflheim.NIHDB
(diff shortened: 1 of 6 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/nihdb/SegmentsWrapper.scala
```diff
@@ -99,3 +99,26 @@ case class SegmentsWrapper(segments: Seq[Segment], projectionId: Int, blockId: L
...
+
+    slice.columns.toList map {
+      case (ColumnRef(path, CBoolean), col: ArrayBoolColumn) =>
+        BooleanSegment(id, path, col.definedAt, col.values, length)
+
+      case (ColumnRef(path, tpe: CNullType), col: BitsetColumn) =>
+        NullSegment(id, path, tpe, col.definedAt, length)
+
+      // a =:= e
+      case (ColumnRef(path, tpe: CValueType[e]), col: AccessibleArrayColumn[a]) =>
```
alissapajer: If/when `CArrayType` comes online, I'm pretty sure this will break.

djspiewak: It will.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
```diff
@@ -59,17 +60,20 @@ final class MimirPTableStore[F[_]: Monad: LiftIO] private (
   def write(key: StoreKey, table: PTable): Stream[F, Unit] = {
     val ios = Stream.bracket(cake.createDB(keyToFile(key)).map(_.toOption))({
       case Some((_, _, db)) =>
-        fromStreamT(table.slices).zipWithIndex evalMap {
+        val can = table.compact(cake.trans.TransSpec1.Id).canonicalize(Config.maxSliceRows)
```
alissapajer: We still don't limit the number of columns in a `Slice` in `canonicalize`. Meaning, this would explode us if we throw some wide enough data at it.

To note, `canonicalize` does have a call site already in `cross`. Though to my knowledge the product isn't generating queries that would hit `cross` yet.

yggdrasil/src/test/scala/quasar/yggdrasil/nihdb/SegmentsWrapperSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+              IO.fromFutureShift(
+                IO(nihdb.insertSegmentsVerified(index.toLong, slice.size, segments)))
+          }
+
+          proj <- NIHDBProjection.wrap(nihdb)
+          stream <- proj.getBlockStream(None).toStream
+        } yield {
+          val slices2 = stream.toList.map(_.deref("value"))
+
+          slices.flatMap(_.toRValues) mustEqual slices2.flatMap(_.toRValues)
```
alissapajer: :+1:

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
djspiewak: > Meaning, this would explode us if we throw some wide enough data at it.

Yep. It will also explode in NIHDB if we throw wide enough data at it, too. We need to make sure that the frontend guarantees we're not putting excessively wide data into prepared tables. Additionally, we will need to make sure that this is the case for joins and unions as well. (heads up, @beckyconning)

PR #3766, opened 2018-08-21
Cache holes in branching nodes

djspiewak:
Verified pretty well by the integration tests (though the cartesian case is apparently never hit with large datasets; perhaps for the best).

[ch158]

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
+++ b/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
@@ -25,9 +25,10 @@ import quasar.contrib.iota.copkTraverse
 import quasar.mimir.MimirCake._
 import quasar.qscript._
 
+import quasar.yggdrasil.MonadFinalizers
 import quasar.yggdrasil.TableModule.SortAscending
 
-import cats.effect.LiftIO
+import cats.effect.{IO, LiftIO}
 
 import matryoshka.{Hole => _, _}
 import matryoshka.implicits._
@@ -40,7 +41,7 @@ import scalaz._, Scalaz._
 
 final class EquiJoinPlanner[
     T[_[_]]: BirecursiveT: EqualT: ShowT,
-    F[_]: LiftIO: Monad] extends Logging {
+    F[_]: LiftIO: Monad: MonadFinalizers[?[_], IO]] extends Logging {
 
   def mapFuncPlanner[G[_]: Monad] = MapFuncPlanner[T, G, MapFunc[T, ?]]
 
@@ -63,11 +64,17 @@ final class EquiJoinPlanner[
       val (lkeys, rkeys) = keys.unfzip
 
       for {
+        cachedD <- LiftIO[F].liftIO(src.P.cacheTable(src.table))
+        cachedTable = cachedD.unsafeValue
+        _ <- MonadFinalizers[F, IO].tell(cachedD.dispose :: Nil)
+
+        cachedSrc = src.map(_ => cachedTable): MimirRepr
+
         leftRepr <- lbranch.cataM[F, MimirRepr](
-          interpretM[F, QScriptTotal[T, ?], Hole, MimirRepr](κ(src.point[F]), planQST))
+          interpretM[F, QScriptTotal[T, ?], Hole, MimirRepr](κ(cachedSrc.point[F]), planQST))
 
         rightRepr <- rbranch.cataM[F, MimirRepr](
-          interpretM[F, QScriptTotal[T, ?], Hole, MimirRepr](κ(src.point[F]), planQST))
+          interpretM[F, QScriptTotal[T, ?], Hole, MimirRepr](κ(cachedSrc.point[F]), planQST))
 
         lmerged = src.unsafeMerge(leftRepr)
         ltable = lmerged.table
--- a/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
+++ b/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
@@ -105,7 +105,7 @@ final class EquiJoinPlanner[
 
             for {
               // cache the right side since we're going to restart it
-              cachedRightD <- LiftIO[F].liftIO(rmerged.P.cacheTable(rtable, cartesian))
+              cachedRightD <- LiftIO[F].liftIO(rmerged.P.cacheTable(rtable, false))
               cachedRight = cachedRightD.unsafeValue
               _ <- MonadFinalizers[F, IO].tell(cachedRightD.dispose :: Nil)
             } yield (ltable.cross(cachedRight)(transMiddle), src.unsafeMerge(leftRepr).lastSort)
--- a/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
+++ b/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
@@ -103,7 +103,12 @@ final class EquiJoinPlanner[
           if (cartesian) {
             log.trace("EQUIJOIN: full-cross detected!")
 
-            (ltable.cross(rtable)(transMiddle), src.unsafeMerge(leftRepr).lastSort).point[F]
+            for {
+              // cache the right side since we're going to restart it
+              cachedRightD <- LiftIO[F].liftIO(rmerged.P.cacheTable(rtable, cartesian))
+              cachedRight = cachedRightD.unsafeValue
+              _ <- MonadFinalizers[F, IO].tell(cachedRightD.dispose :: Nil)
+            } yield (ltable.cross(cachedRight)(transMiddle), src.unsafeMerge(leftRepr).lastSort)
           } else {
             log.trace("EQUIJOIN: not a full-cross; sorting and cogrouping")
 
--- a/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
+++ b/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
@@ -62,9 +62,10 @@ final class EquiJoinPlanner[
       }
 
       val (lkeys, rkeys) = keys.unfzip
+      val cartesian = keys.isEmpty
 
       for {
-        cachedD <- LiftIO[F].liftIO(src.P.cacheTable(src.table))
+        cachedD <- LiftIO[F].liftIO(src.P.cacheTable(src.table, cartesian))
         cachedTable = cachedD.unsafeValue
         _ <- MonadFinalizers[F, IO].tell(cachedD.dispose :: Nil)
 
@@ -99,7 +100,7 @@ final class EquiJoinPlanner[
 
         // identify full-cross and avoid cogroup
         resultAndSort <- {
-          if (keys.isEmpty) {
+          if (cartesian) {
             log.trace("EQUIJOIN: full-cross detected!")
 
             (ltable.cross(rtable)(transMiddle), src.unsafeMerge(leftRepr).lastSort).point[F]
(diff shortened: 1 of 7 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/table/VFSColumnarTableModule.scala
```diff
@@ -227,6 +229,115 @@ trait VFSColumnarTableModule extends BlockStoreColumnarTableModule with Logging
...
 
+  /**
+   * Produces a Table which will, when run the first time, write itself out to
+   * NIHDB storage in a temporary directory, which in turn the subsequent reads
+   * will consume. The very first running of the table must be completed entirely
+   * before subsequent reads may proceed. Once the first read is completed, all
+   * subsequent reads may happen in parallel. The disposal action produced will
+   * shut down the temporary database and remove the directory.
+   *
+   * TODO configurable early force
```
alissapajer: Configurable at what level?

djspiewak: Crap I forgot to remove that TODO. It *is* configurable in the parameter on the function it is documenting. :-)

alissapajer: :+1: could you push a change removing it?

mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
```diff
@@ -92,7 +100,7 @@ final class EquiJoinPlanner[
 
         // identify full-cross and avoid cogroup
         resultAndSort <- {
-          if (keys.isEmpty) {
+          if (cartesian) {
```
alissapajer: :+1:

mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
```diff
@@ -61,13 +62,20 @@ final class EquiJoinPlanner[
       }
 
       val (lkeys, rkeys) = keys.unfzip
+      val cartesian = keys.isEmpty
 
       for {
+        cachedD <- LiftIO[F].liftIO(src.P.cacheTable(src.table, cartesian))
```
alissapajer: Hm, if it's not a cartesian, why don't we have to force the caching?

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
```diff
@@ -334,12 +334,19 @@ final class QScriptCorePlanner[
...
     case qscript.Union(src, lBranch, rBranch) =>
       for {
-       leftRepr <- interpretBranch(planQST)(src, lBranch)
-       rightRepr <- interpretBranch(planQST)(src, rBranch)
+        cachedD <- LiftIO[F].liftIO(src.P.cacheTable(src.table, false))
+        cachedTable = cachedD.unsafeValue
+        _ <- MonadFinalizers[F, IO].tell(cachedD.dispose :: Nil)
 
-       rightCoerced = leftRepr.unsafeMerge(rightRepr)
+        cachedSrc = src.map(_ => cachedTable): MimirRepr
```
alissapajer: This line is so silly.

mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
djspiewak: Oh you know what? You're right. `cogroup` has the same problem, not because of the restart (which is what I thought), but because of the simultaneous paging of both sides. I'll update.

mimir/src/main/scala/quasar/mimir/QScriptCorePlanner.scala
djspiewak: Yep. :-(

mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
alissapajer: Other thought: In the case of a `cross`, should we cache the table that is being restarted, as well as `src`? Otherwise we still have to run (in pseudo code) `rbranch.cataM(interpretM(cachedSrc, planQST))` once per restart, which is derived from a `FreeQS` and could potentially be expensive.

djspiewak: The problem is the answer to that question (should we cache the fully-evaluated freeqs?) is entirely dependent on *how* expensive the calculation is. If it's just a trivial derivation from the hole, then we definitely don't want to spend that storage space. On the other hand, if it's really expensive (e.g. imagine if the whole is `Unreferenced` and it's a cross between two separate datasets!), then we actually do need to cache it.

I think that you're right: for the `cross` case, we should cache the right side in addition to the hole, if only because the "two separate datasets" case seems more likely than the self-cartesian case.

yggdrasil/src/main/scala/quasar/yggdrasil/table/VFSColumnarTableModule.scala
```diff
@@ -227,6 +229,118 @@ trait VFSColumnarTableModule extends BlockStoreColumnarTableModule with Logging
...
+   * before subsequent reads may proceed. Once the first read is completed, all
+   * subsequent reads may happen in parallel. The disposal action produced will
+   * shut down the temporary database and remove the directory.
+   *
+   * If `earlyForce` is `true`, then the effect of running the *entire* input
+   * table stream and writing it out to disk will be sequenced into the outer
+   * `IO`. This flag is important in the event that the output `Table` will be
+   * sequenced multiple times *interleaved*, which is of course a deadlock.
+   */
+  def cacheTable(table: Table, earlyForce: Boolean): IO[Disposable[IO, Table]] = {
```
alissapajer: Ideas for unit testing:

We can put an effect in the initial table creation, like incrementing an `AtomicInteger`. Then we cache the table and then perform operations to run the table multiple times (maybe we can just `force` it). We check 1) that the resulting table is as expected and 2) that the `AtomicInteger` was only incremented once.

This wouldn't be testing the `earlyForce` logic in this function, but it would be testing that we have actually cached the table. Perhaps to test the `earlyForce` logic we can do something similar by running the outer effect and verifying that the `AtomicLong` incremented once, before we continue on to run the table.

djspiewak: Great idea!

yggdrasil/src/main/scala/quasar/yggdrasil/table/VFSColumnarTableModule.scala
```diff
@@ -227,6 +229,118 @@ trait VFSColumnarTableModule extends BlockStoreColumnarTableModule with Logging
...
+          st.uncons map {
+            case Some((head, tail)) =>
+              Some(((head, index), (tail, index + 1)))
+
+            case None => None
+          }
+      }
+    }
+
+    val target = table.compact(trans.TransSpec1.Id).canonicalize(Config.maxSliceRows)
```
alissapajer: Fine either way, but we could make this a function `Table#compactAndCanonicalize`.

djspiewak: We could. They're split mostly because they are in fact useful independently, for cases like `Take`/`Drop` (and it's slightly more efficient to have the split control sometimes).

alissapajer: I was thinking more that we keep them both exposed independently, but also add a `compactAndCanonicalize` for the situation when we need them both.

mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
djspiewak: There isn't an easy to way to check directly if the `src` is `Unreferenced` since we're using Matryoshka. We could add a flag to `MimirRepr` or use a different fold, basically. I'm not sure it gains us anything though. If it's a constant long table, then it'll fall under the size limit, which means we'll use `force` rather than paging to NIHDB. `force` is basically instant on a pure table, so it shouldn't matter really at all. That's why I didn't bother trying to detect the case.

alissapajer: I was thinking that by detecting this case, we could determine when we absolutely must branch cache the fully-evaluated left and right side `FreeQS`.

djspiewak: Well, we don't ever need to cache the fully-evaluated left side since it's never restarted. The right side though… Looking at the hole doesn't really tell us anything about right side caching. Even if the hole is unreferenced, it doesn't guarantee that the right side requires caching. Maybe it's just a constant literal that is somehow obfuscated enough to defeat autojoin minimization! Similarly, if the hole is *not* unreferenced, the right side could still be some very complex computation, possibly even a join involving other reads, and we don't want to recompute that when we restart.

Really, we would need to look at the freeqs to be totally sure.

yggdrasil/src/main/scala/quasar/yggdrasil/table/VFSColumnarTableModule.scala
djspiewak: I'd be ok with that. This case has come up a couple times.

yggdrasil/src/main/scala/quasar/yggdrasil/table/VFSColumnarTableModule.scala
```diff
@@ -227,6 +229,118 @@ trait VFSColumnarTableModule extends BlockStoreColumnarTableModule with Logging
...
+                  IO.fromFutureShift(
+                    IO(db.insertSegmentsVerified(offset.toLong, slice.size, segments))).as(slice)
+              }
+
+              persisting ++ StreamT.wrapEffect(cacheTarget.complete((db, dir)).as(StreamT.empty[IO, Slice]))
+            }
+          } else {
+            for {
+              pair <- cacheTarget.get    // wait for the other thread to finish the persistence
+              (db, _) = pair
```
wemrysi: :tada:

djspiewak: Lol. The plugin isn’t applied to these projects.

yggdrasil/src/test/scala/quasar/yggdrasil/nihdb/NIHDBAkkaSetup.scala
```diff
@@ -0,0 +1,54 @@
...
+import quasar.niflheim.{Chef, V1CookedBlockFormat, V1SegmentFormat, VersionedCookedBlockFormat, VersionedSegmentFormat}
+
+import akka.actor.{ActorRef, ActorSystem, Props}
+
+import scala.concurrent.duration._
+
+import java.util.concurrent.{ScheduledThreadPoolExecutor, ThreadFactory}
+import java.util.concurrent.atomic.AtomicInteger
+
+trait NIHDBAkkaSetup {
```
alissapajer: :+1:

yggdrasil/src/test/scala/quasar/yggdrasil/table/VFSColumnarTableModuleSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+
+          _ <- cached.slices.foreach(_ => IO.pure(()))
+
+          i <- cget
+          _ <- IO(i mustEqual 1)
+
+          _ <- cached.slices.foreach(_ => IO.pure(()))
+
+          i <- cget
+          _ <- IO(i mustEqual 1)
```
alissapajer: :+1:

PR #3768, opened 2018-08-23
sanitize datasource config

nicflores:
Sanitize config of secret key.

### DIFF
--- a/api/src/test/scala/quasar/api/datasource/DatasourcesSpec.scala
+++ b/api/src/test/scala/quasar/api/datasource/DatasourcesSpec.scala
@@ -261,6 +261,9 @@ abstract class DatasourcesSpec[
   def refB: F[DatasourceRef[C]] =
     randomName map (DatasourceRef(supportedType, _, validConfigs._2))
 
+  def sanitizedRefB: F[DatasourceRef[C]] =
+    randomName map (DatasourceRef(supportedType, _, validConfigs._2))
+
   def createRef(r: DatasourceRef[C]): F[I] =
     datasources.addDatasource(r) >>= expectSuccess
 
--- a/impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
+++ b/impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
@@ -26,6 +26,7 @@ import java.nio.file.Paths
 import scala.concurrent.ExecutionContext.Implicits.global
 
 import argonaut.Json
+import argonaut.Argonaut.jString
 import cats.effect.{ConcurrentEffect, Timer}
 import fs2.Stream
 import scalaz.{\/, EitherT}
@@ -35,6 +36,10 @@ import shims._
 object LocalDatasourceModule extends LightweightDatasourceModule {
 
   val kind: DatasourceType = LocalType
+
+  def sanitizeConfig(config: Json): Json = {
+    config.hcursor.downField("credentials").downField("secretKey").set(jString("***********")).undo.getOrElse(config)
+  }
 
   def lightweightDatasource[F[_]: ConcurrentEffect: MonadResourceErr: Timer](config: Json)
       : F[InitializationError[Json] \/ Disposable[F, Datasource[F, Stream[F, ?], ResourcePath, Stream[F, Data]]]] = {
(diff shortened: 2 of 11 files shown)


nicflores:
@wemrysi, let me know you don't have time look at this let me know and I can pass this on to @alissapajer. I need some feedback on `MockDatasourceControl.scala`, not sure how deal with the unspecified `C`, so I just returned the same `config`.

impl/src/main/scala/quasar/impl/datasources/DatasourceManagement.scala
```diff
@@ -73,6 +74,11 @@ final class DatasourceManagement[
 
   // DatasourceControl
 
+  def sanitizeRef(ref: DatasourceRef[Json]): DatasourceRef[Json] = {
```
wemrysi: Here, you'll want to lookup the type of the ref in `modules` to obtain it's `DatasourceModule` and then use it to sanitize the config.

If the module doesn't exist, then we could just return `JNull` or `JEmptyObject` for the config, if we want to be conservative.

impl/src/main/scala/quasar/impl/datasources/DefaultDatasources.scala
```diff
@@ -55,7 +55,9 @@ final class DefaultDatasources[
     })
 
   def datasourceRef(datasourceId: I): F[ExistentialError[I] \/ DatasourceRef[C]] =
-    lookupRef[ExistentialError[I]](datasourceId)
+    EitherT(lookupRef[ExistentialError[I]](datasourceId))
+    .flatMap(x => EitherT.rightT(x.pure[F]))
```
wemrysi: The `flatMap` here is unnecessary.

impl/src/test/scala/quasar/impl/datasources/MockDatasourceControl.scala
```diff
@@ -39,6 +39,8 @@ final class MockDatasourceControl[F[_]: Monad, G[_]: PlusEmpty, I: Order, C] pri
     implicit initd: MonadInit[F, I], sdown: MonadShutdown[F, I])
     extends DatasourceControl[F, G, I, C, MockSchemaConfig.type] {
 
+  def sanitizeRef(ref: DatasourceRef[C]): DatasourceRef[C] = ref
```
wemrysi: You could instead accept the sanitize function, `C => C` as an argument  to `MockDatasourceControl`, then it would allow us to test that `DefaultDatasources` properly sanitizes the refs it returns.

impl/src/test/scala/quasar/impl/datasources/DatasourceManagementSpec.scala
```diff
@@ -95,6 +96,10 @@ final class DatasourceManagementSpec extends quasar.Qspec with ConditionMatchers
   val lightMod = new LightweightDatasourceModule {
     val kind = LightT
 
+    def sanitizeConfig(config: Json): Json = {
+      config.hcursor.downField("credentials").downField("secretKey").set(jString("***********")).undo.getOrElse(config)
```
wemrysi: This is fine, but you could also use a much more simple definition to make testing a bit easier (i.e. just always return the string `"sanitized"` or something.

We're not testing what it _means_ to sanitize a config, because that is connector-specific, just that we give the connector a chance to do so.

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
```diff
@@ -36,6 +37,10 @@ object LocalDatasourceModule extends LightweightDatasourceModule {
 
   val kind: DatasourceType = LocalType
 
+  def sanitizeConfig(config: Json): Json = {
```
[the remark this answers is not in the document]
nicflores: good point! I should have realized that.

api/src/test/scala/quasar/api/datasource/DatasourcesSpec.scala
```diff
@@ -261,6 +261,9 @@ abstract class DatasourcesSpec[
   def refB: F[DatasourceRef[C]] =
     randomName map (DatasourceRef(supportedType, _, validConfigs._2))
 
+  def sanitizedRefB: F[DatasourceRef[C]] =
```
wemrysi: Still needed?

wemrysi:
@nicflores I changed the version to `breaking` since it affects binary compatibility with connectors.

PR #3770, opened 2018-08-27
extract Slice.replaceColumnImpl

rintcius:
Extract `Slice.replaceColumnImpl` so that we also replace the column impl where `Slice` is extended (ie in `SegmentsWrapper` and `DerefSlice`).

Also a couple more places to replace `new Slice` with `Slice.apply`


djspiewak:
Both `DerefSlice` and `SegmentsWrapper` are just collections of functions used to create `size` and `columns`. They aren't actually doing anything polymorphic. Which is to say, they could both be pretty trivially rewritten in terms of `Slice.apply`. It seems to me that such an implementation would be better than exposing `replaceColumnsImpl` as an external function, no?

rintcius:
Ah, right, I wasn't really looking at those classes in that mode. Nicer not to expose `replaceColumnsImpl` indeed, and also to simply create a `Slice` via `Slice.apply` instead of extending `Slice`.

PR #3775, opened 2018-08-29
Remove the typechecking phase

alissapajer:
[ch1589]

Commits are self-contained. All commits compile but not all commits test.

`Sql2QueryRegressionSpec` clocks in locally at 3 minutes 43 seconds.

This brings us very close to two more exciting things:
1) We can delete `Data.NA`. I believe that the only remaining place it's used is in `StdLibSpec`. We can add a couple new testing functions so that we can express in the tests that we want a value to be undefined.
2) We can delete `Simplifier`. There are a couple `Simplifier`s that are doing real things, like the one for `SetLib.In`, but we can move that functionality to `ReadLP`, for example.


mimir/src/main/scala/quasar/mimir/StringLib.scala
```diff
@@ -227,8 +230,61 @@ trait StringLibModule extends ColumnarTableLibModule {
...
+
+        case c   :: t =>
+          if ("\\^$.|?*+()[{".contains(c))
+            '\\' +: c +: escape(t)
+          else c +: escape(t)
+
+        case Nil => Nil
+      }
+
+    @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
```
djspiewak: WartRemover isn't actually enabled for mimir, so this is technically redundant.

PR #3779, opened 2018-08-31
QData[RValue]

alissapajer:
Depends on #3777 and #3780.

### DIFF
--- a/blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
+++ b/blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
@@ -161,11 +161,11 @@ object QDataRValue extends QData[RValue] {
   def getObjectValueAt(ac: ObjectCursor): RValue = ac.values(ac.index.toInt)._2
   def stepObject(ac: ObjectCursor): ObjectCursor = ac.copy(index = ac.index + Natural.one)
 
-  type NascentObject = List[(String, RValue)]
+  type NascentObject = Map[String, RValue]
 
-  def prepObject: NascentObject = List[(String, RValue)]()
-  def pushObject(key: String, a: RValue, na: NascentObject): NascentObject = (key, a) +: na // prepend
-  def makeObject(na: NascentObject): RValue = if (na.isEmpty) CEmptyObject else RObject(na.reverse.toMap)
+  def prepObject: NascentObject = Map[String, RValue]()
+  def pushObject(key: String, a: RValue, na: NascentObject): NascentObject = na + ((key, a))
+  def makeObject(na: NascentObject): RValue = if (na.isEmpty) CEmptyObject else RObject(na)
 
   def getMetaValue(a: RValue): RValue = error(s"Unable to represent metadata in `RValue`.")
   def getMetaMeta(a: RValue): RValue = error(s"Unable to represent metadata in `RValue`.")
--- a/blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
+++ b/blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
@@ -21,7 +21,7 @@ import slamdata.Predef._
 import qdata._
 import qdata.time.{DateTimeInterval, OffsetDate}
 
-import spire.math.{Natural, Real}
+import spire.math.Real
 
 import java.math.MathContext
 import java.time.{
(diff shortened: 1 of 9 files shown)


blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
```diff
@@ -0,0 +1,173 @@
...
+  def getArrayAt(ac: ArrayCursor): RValue = ac.values(ac.index.toInt)
+  def stepArray(ac: ArrayCursor): ArrayCursor = ac.copy(index = ac.index + Natural.one)
+
+  type NascentArray = List[RValue]
+
+  def prepArray: NascentArray = List[RValue]()
+  def pushArray(a: RValue, na: NascentArray): NascentArray = a +: na // prepend
+  def makeArray(na: NascentArray): RValue = if (na.isEmpty) CEmptyArray else RArray(na.reverse)
+
+  final case class ObjectCursor(index: Natural, values: List[(String, RValue)])
```
wemrysi: Same argument here as with `ArrayCursor`.

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
```diff
@@ -0,0 +1,173 @@
...
+    case _ => error(s"Expected `RArray` or `CEmptyArray`. Received $a")
+  }
+  def hasNextArray(ac: ArrayCursor): Boolean = ac.values.length > ac.index.toInt
+  def getArrayAt(ac: ArrayCursor): RValue = ac.values(ac.index.toInt)
+  def stepArray(ac: ArrayCursor): ArrayCursor = ac.copy(index = ac.index + Natural.one)
+
+  type NascentArray = List[RValue]
+
+  def prepArray: NascentArray = List[RValue]()
+  def pushArray(a: RValue, na: NascentArray): NascentArray = a +: na // prepend
```
wemrysi: Minor, `::` is more idiomatic when prepending to a list.

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
```diff
@@ -0,0 +1,173 @@
...
+    case RObject(obj) => ObjectCursor(Natural.zero, obj.toList)
+    case CEmptyObject => ObjectCursor(Natural.zero, List[(String, RValue)]())
+    case _ => error(s"Expected `Data.Obj`. Received $a")
+  }
+  def hasNextObject(ac: ObjectCursor): Boolean = ac.values.length > ac.index.toInt
+  def getObjectKeyAt(ac: ObjectCursor): String = ac.values(ac.index.toInt)._1
+  def getObjectValueAt(ac: ObjectCursor): RValue = ac.values(ac.index.toInt)._2
+  def stepObject(ac: ObjectCursor): ObjectCursor = ac.copy(index = ac.index + Natural.one)
+
+  type NascentObject = List[(String, RValue)]
```
wemrysi: I think we could just use a `Map[String, RValue]` here and avoid the list traversals altogether. We lose all insert order when we convert to `Map` at the end, so `List` doesn't appear to gain us anything.

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
```diff
@@ -0,0 +1,173 @@
...
+  }
+  def makeOffsetTime(l: OffsetTime): RValue = COffsetTime(l)
+
+  def getInterval(a: RValue): DateTimeInterval = a match {
+    case CInterval(value) => value
+    case _ => error(s"Expected `CInterval`. Received $a")
+  }
+  def makeInterval(l: DateTimeInterval): RValue = CInterval(l)
+
+  final case class ArrayCursor(index: Natural, values: List[RValue])
```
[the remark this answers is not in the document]
wemrysi: If the contract is that `stepArray` should only be called when `hasNextArray` returns `true` then we could actually use `_.tail` to implement `stepArray`.

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
alissapajer: That's a good point. There was a reason I used `List` but I can't remember what it was, so I'll make the `Map` change and see what falls out :)

alissapajer:
@wemrysi thanks for the suggestions! The `QData[RValue]` implementation largely mirrored the `QData[Data]` implementation, so whatever changes I make, I'll make them over in `QData[Data]` as well, where appropriate.

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
alissapajer: I figured out why I had `List` here. In `Data.Obj` we store the values in a `ListMap[String, Data]` and so order must be preserved in that case. But in `RValue` we use `Map[String, RValue]` so we can indeed use `Map` here!

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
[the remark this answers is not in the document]
alissapajer: I wonder if we have enough or long enough arrays/objects in our integration tests for this fix to make them even faster!

blueeyes/src/main/scala/quasar/precog/common/QDataRValue.scala
wemrysi: Gotcha. We could use `ListMap` in the `Data` instance directly as it preserves insertion order.

alissapajer:

wemrysi:
@alissapajer Would you mind adding an implicit `QData` instance to the companion objects of `Data` and `RValue` so it may be found automatically by implicit search?

alissapajer:
@wemrysi implicits added!

PR #3780, opened 2018-09-01
Specs related upgrades

alissapajer:


frontend/src/main/scala/quasar/frontend/logicalplan/Optimizer.scala
```diff
@@ -331,25 +331,25 @@ final class Optimizer[T: Equal]
   val optimize: T => T =
     NonEmptyList[T => T](
       // Eliminate extraneous constants, etc.:
-      simplify,
+      simplify(_),
```
wemrysi: Any idea why these changes were necessary?

alissapajer: Hm, when I made the commit I assumed it was due to shapeless since that's the only non-test dependency that was updated. But now that I think about it more, I'm unable to fully reason my way through that line of thought since shapeless isn't being directly used here...

Also, I just confirmed that this diff is indeed needed, else we get the error `missing argument list for method simplify in class Optimizer`.

wemrysi: Huh, interesting.

PR #3781, opened 2018-09-01
Double trouble toil and bubble

alissapajer:
Depends on #3777 because it requires 56a09ff.

So, um, we were not creating `CDouble` values in mimir, ever; all non-whole numbers were represented as `CNum`. This is because the `Data => RValue` function was bugged. I discovered this locally on my `Facade` branch (which bypasses `Data` entirely :tada:) when the `trunc` integration test failed.


frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -2118,6 +2128,15 @@ abstract class StdLibSpec extends Qspec {
         "any Dec" >> prop { (x: BigDecimal) =>
```
alissapajer: Note to self that we need to make an issue to look into what `BigDecimal` generator we're using because whatever one it is, it's not generating very useful values because it didn't catch this bug.

rintcius: Related and in addition to that: I was thinking that whenever we find a bug like this, we should at least add one value explicitly to the generator that reproduces the bug. In this way we'll gradually increase quality of our generators and reduce chances of regressions.

PR #3783, opened 2018-09-04
Delete EncodeJson[Type]

alissapajer:


alissapajer:
@wemrysi any reason (maybe SSTs) to keep this around?

wemrysi:
@alissapajer I don't think so, SSTs don't use them and I'm pretty sure the only reason this existed was to render types in structured errors.

PR #3782, opened 2018-09-04
add string to number conversion function

rintcius:
Depends on #3781 

[ch1619]

PR #3784, opened 2018-09-06
add string to number conversion function

rintcius:
cleaned up version of #3782 

[ch1619]


yggdrasil/src/test/scala/quasar/yggdrasil/TransformSpec.scala
```diff
@@ -2242,6 +2242,107 @@ trait TransformSpec extends TableModuleTestSupport with SpecificationLike with S
...
+      1
+    ]""")
+    val sample = SampleData(data.toStream)
+    val table = fromSample(sample)
+
+    val results = toJson(table.transform {
+      ToNumber(Leaf(Source))
+    })
+
+    results.getJValues must_== Stream(JNum(1), JNum(42), JNum(0), JNum(Long.MaxValue), JNum(Long.MinValue))
```
alissapajer: Could we instead use the `JNumLong` constructor itself here instead of `JNum`? Then we wouldn't need the class check two lines down. And the analogous for the other tests in this suite.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -321,32 +321,34 @@ abstract class StdLibSpec extends Qspec {
...
 
-        // TODO: how to express "should execute and may produce any result"
+      "Number" >> {
+        "any BigInt in the domain" >> prop { (x: BigInt) =>
+          unary(Number(_).embed, Data.Str(x.toString), Data.Int(x))
+        }
+
+        "any BigDecimal in the domain" >> prop { (x: BigDecimal) =>
+          unary(Number(_).embed, Data.Str(x.toString), Data.Dec(x))
+        }
```
alissapajer: Can you add a couple more tests generating `Double` and `Long`? The reason being, the `BigInt` and `BigDecimal` generators are not generating (at least not frequently) typical and corner-case `Double` and `Long` values, which are values that are more likely to be in user data. In most user data, we probably won't even have any arbitrary precision values.

As an aside, once we get rid of `Data` we won't have this problem since `RValue` has numeric types that aren't arbitrary precision.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -490,6 +490,45 @@ abstract class Slice { source =>
     Slice (size, columns)
   }
 
+  def toNumber = {
```
alissapajer: Can you include the return type for this?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -490,6 +490,45 @@ abstract class Slice { source =>
...
+
+        RangeUtil.loopDefined(0 to size, c){ i =>
+          val s = c(i)
+          try {
+            val l = java.lang.Long.parseLong(s)
+            lc.update(i, l)
+          } catch {
+            case _: NumberFormatException =>
+              try {
+                val n = BigDecimal(s)
```
alissapajer: Rather than creating the overhead of a `BigDecimal`, we should use `java.lang.Double.parseDouble` to see if we have a `Double`. And then only when we don't have a `Double`, create a `BigDecimal`.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -490,6 +490,45 @@ abstract class Slice { source =>
...
+    val columns = source.columns.get(ColumnRef(CPath.Identity, CString)) match {
+      case Some(c: StrColumn) =>
+        val lc = ArrayLongColumn.empty(size)
+        val dc = ArrayDoubleColumn.empty(size)
+        val nc = ArrayNumColumn.empty(size)
+
+        RangeUtil.loopDefined(0 to size, c){ i =>
+          val s = c(i)
+          try {
+            val l = java.lang.Long.parseLong(s)
```
alissapajer: For `Long` parsing, using [jawn's `parseLong`](https://github.com/non/jawn/blob/3a196ac886cd3053352075cad0fae6a5f26b50e9/util/src/main/scala/jawn/util/package.scala#L21) is "significantly faster" than `java.lang.Long.parseLong` (as stated [here](https://github.com/non/jawn/blob/3a196ac886cd3053352075cad0fae6a5f26b50e9/util/src/main/scala/jawn/util/package.scala#L71)). So we can use that one.

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
```diff
@@ -210,7 +197,7 @@ final class Sql2QueryRegressionSpec extends Qspec {
     // TODO{fs2}: Chunkiness
     val actNormal =
       act.mapChunks(
-        _.map(normalizeJson <<< deleteFields <<< (_.asJson))
```
alissapajer: :+1:

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -490,6 +490,45 @@ abstract class Slice { source =>
     Slice (size, columns)
   }
 
+  def toNumber = {
+    val size = source.size
+    val columns = source.columns.get(ColumnRef(CPath.Identity, CString)) match {
+      case Some(c: StrColumn) =>
+        val lc = ArrayLongColumn.empty(size)
+        val dc = ArrayDoubleColumn.empty(size)
+        val nc = ArrayNumColumn.empty(size)
```
alissapajer: Observing that we're allocating potentially up to three columns more than required, since it's possible that none of the string values will parse into numbers. In the context of this solution, the alternative I can think of would be setting `var lc = _` and then the first time we need to create a `Long` value setting `lc = ArrayLongColumn.empty(size)`. Then we'd have to do `null` checks each time we need to `update` a column value. 

I'm not immediately sure if the gains we'd get from that are worth it. Could you add a `TODO` comment here about how we're allocating extra columns (in some situations) so we can come back to this if needed?

yggdrasil/src/test/scala/quasar/yggdrasil/TransformSpec.scala
rintcius: That won't help because equality is defined like this:

```
  override def equals(other: Any) = other match {
    case JNumLong(n)   => value == n
    case JNumDouble(n) => value == n
    case JNumBigDec(n) => value == n
    case JNumStr(s)    => value.toString == s
    case _             => false
  }
```

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
rintcius: Ah right, there's also https://github.com/bourgesl/jnumbers and was thinking about adding that as a dependency, but thought it'd be better to do that (or something else) separately (I wasn't that much focusing on getting best possible performance in this PR). 

Are you ok addressing this separately, as it looks like yggdrasil is not having jawn as a dep yet? Then we can choose the best lib for these kind of things.

alissapajer: Sure, can you add a `TODO` here?

yggdrasil/src/test/scala/quasar/yggdrasil/TransformSpec.scala
alissapajer: Gotcha, ok.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
rintcius: Right, `Double.parseDouble` does not work. I should have commented this in the first place. Will comment it now.

alissapajer:
@rintcius I'm on mobile and I can't find the approval button, but I approve this. Can you (or anyone else) merge this? Thanks :)

PR #3785, opened 2018-09-06
fix round for doubles

rintcius:
fix round for doubles

... removed round from MathLib because it's broken and we can use rint instead
(both Math.rint and round from sql2 use HALF_EVEN)

[ch1777]


rintcius:
@alissapajer I also want to change the IT `derivedFunctionRound` and all other `derivedFunction` ITs to be based on `numbers.data` instead of `divide.data` but I'll do that as a separate PR once `numbers.data` is available via #3784

PR #3787, opened 2018-09-06
Refactor regression test runner to use portable file paths

wemrysi:
[ch1467]


wemrysi:
@alissapajer That's correct, it wasn't working previously because we were using output from the `posixCodec` in `scala-pathy` to build `java.nio.file.Path` instances, which didn't work on windows as it uses a different path separator.

Instead of doing something specifically for windows, we take the platform agnostic route and build up `java.nio.file.Path` instances programmatically, allowing the JVM to dictate how paths are represented.

alissapajer:
@wemrysi awesome that's a great approach!

PR #3789, opened 2018-09-06
OffsetDateTime toEpochSeconds returns seconds, not milliseconds

jsantos17:
We already record durations for [ch1749], but they're expressed in seconds, not milliseconds.


alissapajer:
@jsantos17 did you intend to close ch1749 with this PR? Including it in the PR description linked it.

alissapajer:
@jsantos17 needs version label. I don't think this affects anything upstream, so `revision`?

Also, once you add the version label, you can merge it.

alissapajer:
Also, I unlinked the ch story and moved it back to `Develop` since this isn't adding the timing info to emails.

jsantos17:
@alissapajer mixed up the PRs. I should've added the story number in the dependent PR instead of here. :)

PR #3791, opened 2018-09-07
fix to_timestamp for integer values which are not a Long

rintcius:
[ch1779]

### DIFF
--- a/frontend/src/test/scala/quasar/std/StdLibSpec.scala
+++ b/frontend/src/test/scala/quasar/std/StdLibSpec.scala
@@ -1517,6 +1517,47 @@ abstract class StdLibSpec extends Qspec {
             Data.Int(BigInt(l)),
             Data.OffsetDateTime(JOffsetDateTime.ofInstant(Instant.ofEpochMilli(l), ZoneOffset.UTC)))
         }
+
+        "arbitrary double" >> prop { (d: Double) =>
+          val n = BigDecimal(d)
+          val data =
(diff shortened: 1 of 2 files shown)


rintcius:
I'll do an IT for this together with the ITs that will be  based on `numbers.data`

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -1517,6 +1517,47 @@ abstract class StdLibSpec extends Qspec {
...
+          val n = BigDecimal(d)
+          val data =
+            // testing with the generated double if it is not a Long or 
+            // if it is an exact double
+            if ((d != d.toLong) || (n.isExactDouble)) Data.Dec(n)
+            // .. but if it is a Long but not an exact double then we test with
+            // d.toLong instead so that we can still test the outcome with exact precision
+            // An example where this happens is: d = -3.2714155255361766E17
+            // BigDecimal(d).toLong = -327141552553617660
+            // d.toLong = -327141552553617664
```
alissapajer: Where did the unit's digit `4` materialize from?

rintcius: I think `-327141552553617664` is the actual value of the generated double and this is approximated as `-3.2714155255361766E17`. I see this approximation would be happening for values `x` where `Math.abs(x)` is big enough, but it surprised me a bit that this is happening for values is in the `Long` range.

PR #3797, opened 2018-09-11
SST performance improvements

wemrysi:
Some low-hanging fruit to improve SST generation performance

1. Make preservation of string structure during compression optional, defaulting to not preserving structure.
2. Optimize computing an SST from a `Chunk` of data.
3. Leverage the fact that SST merging forms a commutative semigroup to divide and conquer, computing multiple chunks in parallel.

### DIFF
--- a/sst/src/main/scala/quasar/sst/compression.scala
+++ b/sst/src/main/scala/quasar/sst/compression.scala
@@ -70,6 +70,7 @@ object compression {
     * also appears in the union.
     */
   def coalescePrimary[J: Order, A: Order: Field: ConvertableTo](
+      stringPreserveStructure: Boolean)(
       implicit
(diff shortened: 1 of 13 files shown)


sst/src/main/scala/quasar/sst/compression.scala
```diff
@@ -70,6 +70,7 @@ object compression {
     * also appears in the union.
     */
   def coalescePrimary[J: Order, A: Order: Field: ConvertableTo](
+      stringPreserveStructure: Boolean)(
```
djspiewak: Minor note: you named this parameter `preserveStructure` below.

wemrysi: I did, but the subject of that function is strings so I didn't think the prefix was necessary.

PR #3799, opened 2018-09-12
Remove `$obj` from precise json.

alissapajer:
We no longer support keys that coincidentally share a name with a precise key.


alissapajer:
Made this breaking, but nothing is using precise json so nothing should break.

PR #3800, opened 2018-09-13
A Datasource has a QueryEvaluator rather than is one

alissapajer:
and other adventures.

[ch1575]

### DIFF
--- a/common/src/main/scala/quasar/common/data/PreciseKeys.scala
+++ b/common/src/main/scala/quasar/common/data/PreciseKeys.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.common.data
+
+object PreciseKeys {
+  val LocalDateTimeKey = "$localdatetime"
+  val LocalDateKey = "$localdate"
+  val LocalTimeKey = "$localtime"
+  val OffsetDateTimeKey = "$offsetdatetime"
+  val OffsetDateKey = "$offsetdate"
+  val OffsetTimeKey = "$offsettime"
+  val IntervalKey = "$interval"
+}
--- a/impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
+++ b/impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
@@ -36,10 +36,10 @@ object QDataPreciseFacade {
     "org.wartremover.warts.Null",
     "org.wartremover.warts.ToString",
     "org.wartremover.warts.Var"))
-  implicit def qdataPrecise[J](implicit qd: QDataEncode[J]): Facade[J] = {
-    import QDataFacade.{qdata => facade}
-
+  implicit def qdataPrecise[J](implicit qd: QDataEncode[J]): Facade[J] =
     new Facade[J] {
+      val facade: Facade[J] = QDataFacade.qdata[J]
+
       def jnull(): J = facade.jnull()
 
       def jfalse(): J = facade.jfalse()
@@ -124,5 +124,4 @@ object QDataPreciseFacade {
           def isObj: Boolean = true
         }
     }
-  }
 }
--- a/impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
+++ b/impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
@@ -18,6 +18,7 @@ package quasar.impl.parse
 
 import slamdata.Predef._
 
+import quasar.common.data.PreciseKeys
 import quasar.time.DateTimeUtil._
 
 import jawn.{Facade, FContext}
@@ -54,14 +55,7 @@ object QDataPreciseFacade {
 
       def objectContext(): FContext[J] =
         new FContext[J] {
-          val LocalDateTimeKey = "$localdatetime"
-          val LocalDateKey = "$localdate"
-          val LocalTimeKey = "$localtime"
-          val OffsetDateTimeKey = "$offsetdatetime"
-          val OffsetDateKey = "$offsetdate"
-          val OffsetTimeKey = "$offsettime"
-          val IntervalKey = "$interval"
-          val ObjKey = "$obj"
+          import PreciseKeys._
 
           var result: qd.NascentObject \/ J = qd.prepObject.left[J]
           var key: String = null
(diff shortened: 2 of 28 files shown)


impl/src/test/scala/quasar/impl/parse/QDataFacadeSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+import slamdata.Predef._
+
+import quasar.Qspec
+import quasar.common.data.Data
+
+import jawn.{Facade, SupportParser}
+import qdata.QDataEncode
+import scalaz.syntax.std.option._
+
+// TODO specs for QDataFacade and QDataPreciseFacade
```
alissapajer: This TODO should be done before merge.

Instead of using `Data` here, I'd like to use `RValue`. I have most of the work done moving `RValue` to `quasar.common.data`. I'll PR that separately and then can merge it in here.

alissapajer:
This is the PR that delivers another speed boost to the integration tests. Locally they run in around 2min 47sec.

impl/src/test/scala/quasar/impl/parse/QDataFacadeSpec.scala
alissapajer: @wemrysi  so I'd to have scalacheck tests for this. I've thought of a couple ways we could do that. 

1) Generate `Arbitrary[RValue]`, turn that into a mimir `Table`, call `renderJson`, then parse through the facade and verify it's the original `RValue`.
2) Generate `Arbitrary[Foo]` where is `Foo` is some json-like representation that we can render into json, render that into json, parse and verify. I could use `jawn.ast.JValue` as it has a `render` function; we'd just need to write a `QDataEncode[jawn.ast.JValue]`. But, precise json is not encoded in `jawn.ast.JValue` and so we couldn't use that to test the precise facade.

I'm not super keen on going through `Table` only because that makes it a little less unit-testy, but we also need to be able to test the precise facade. And I don't think we should write a renderer.

impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
```diff
@@ -0,0 +1,134 @@
...
+      def objectContext(): FContext[J] =
+        new FContext[J] {
+          val LocalDateTimeKey = "$localdatetime"
+          val LocalDateKey = "$localdate"
+          val LocalTimeKey = "$localtime"
+          val OffsetDateTimeKey = "$offsetdatetime"
+          val OffsetDateKey = "$offsetdate"
+          val OffsetTimeKey = "$offsettime"
+          val IntervalKey = "$interval"
+          val ObjKey = "$obj"
```
alissapajer: `ObjKey` can be deleted. And we have these constants accessible somewhere else, I'll push a change.

common/src/main/scala/quasar/common/data/PreciseKeys.scala
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.common.data
+
+object PreciseKeys {
```
wemrysi: :+1:

impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
```diff
@@ -0,0 +1,128 @@
...
+object QDataPreciseFacade {
+
+  @SuppressWarnings(Array(
+    "org.wartremover.warts.AsInstanceOf",
+    "org.wartremover.warts.Equals",
+    "org.wartremover.warts.Null",
+    "org.wartremover.warts.ToString",
+    "org.wartremover.warts.Var"))
+  implicit def qdataPrecise[J](implicit qd: QDataEncode[J]): Facade[J] = {
+    import QDataFacade.{qdata => facade}
```
wemrysi: I'm not sure how much of a performance benefit it will have, if any, but what do you think about instantiating the `QDataFacade.facade` once as a `val` in the precise instance?

alissapajer:
@beckyconning this change removed two intermediate data type conversions when we stream data from a connector into mimir. Thus all queries are improved by this because it's the data transfer step itself that was made faster. With this change, we stream the connector-provided bytes into `RValue` (a json representation), and then transform the stream of `RValue` into mimir's `Table`. For context, our ultimate vision (the evaluating parser) will not have an intermediate in memory representation (so we won't transform to `RValue`), and we'll parse directly into `Table`, which should in theory be even faster.

The S3 connector must change (minorly) when it upgrades to this quasar version, and it will benefit from the change in the same way the local connector does.

Previously the integration tests took about 3min 39sec. So they are about 24% faster.

impl/src/test/scala/quasar/impl/parse/QDataFacadeSpec.scala
wemrysi: @alissapajer Yeah, going through `Table` would also introduce a dependency on `mimir`, which we'd like to avoid.

I'm not sure how to test the precise stuff without a renderer. What about repurposing the `DataCodec` into a `QDataCodec` that works for any `A: QDataEncode: QDataDecode`?

it/src/main/resources/tests/temporal/simple_date_filter.test
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "simple filter using date literals",
+    "name": "foo simple filter using date literals",
```
wemrysi: Remnants of debugging?

run/src/main/scala/quasar/run/ResourceRouter.scala
```diff
@@ -56,8 +57,10 @@ object ResourceRouter {
       case DatasourceResourcePrefix /: UuidString(id) /: qaPath =>
         datasources.map(_.lookup(id) map { d =>
           val qa = d.unsafeValue.fold[QueryAssociate[T, IO]](
-            lw => QueryAssociate.lightweight(f => resultsToIO(lw.evaluate(f))),
-            hw => QueryAssociate.heavyweight(q => resultsToIO(hw.evaluate(q))))
+            lw => QueryAssociate.lightweight(f =>
+              resultsToIO(lw.evaluator[RValue].evaluate(f))),
```
wemrysi: :tada:

impl/src/main/scala/quasar/impl/parse/QDataPreciseFacade.scala
alissapajer: I don't see any downsides to that.

impl/src/test/scala/quasar/impl/parse/QDataFacadeSpec.scala
alissapajer: @wemrysi your idea about codecs led me to this:
```
scala> DataCodec.render(Data.Obj(ListMap(("foo", Data.Int(1)), ("bar", Data.Arr(List(Data.LocalTime(LocalTime.parse("12:05")), Data.True, Data.Null))))))
res14: slamdata.Predef.Option[slamdata.Predef.String] = Some({ "foo": 1, "bar": [ { "$localtime": "12:05:00.000000000" }, true, null ] })
```
So I can actually just generate `Arbitrary[Data]` and render it!

alissapajer: Also, I think a good intermediate goal with `Data` is to move its definition to a test directory, so it's not a huge deal introducing another dependency on it for tests.

alissapajer:
@wemrysi ready for another look!

PR #3801, opened 2018-09-13
add signal to track whether write in MimirPTableStore has completed

rintcius:
[ch1813]
[ch1873]

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
+++ b/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
@@ -33,6 +33,7 @@ import cats.arrow.FunctionK
 import cats.effect.{IO, LiftIO}
 
 import fs2.Stream
+import fs2.async.mutable.Signal
 
 import pathy.Path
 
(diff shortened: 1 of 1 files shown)


rintcius:
@djspiewak I still want to do the clean up part that we discussed, but can't finish that before my days off. I tested this PR in repl and it works fine without the clean up too.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
```diff
@@ -58,26 +59,36 @@ final class MimirPTableStore[F[_]: Monad: LiftIO] private (
...
+                IO(db.insertSegmentsVerified(offset.toLong, slice.size, segments)))
+          }) ++ Stream.eval_(completed.set(true))
+
+        case None =>
+          Stream.empty
+      }, {
+        case Some((blob, version, db)) =>
+          completed.get.flatMap { c =>
+            if (c) cake.commitDB(blob, version, db)
+            else IO.pure(())
```
djspiewak: Should be a `TODO` or something here noting that we need to delete the orphaned version. Additionally, we really need the effect to be *at least* `IO.fromFuture(IO(db.close))`, since we don't want to leak resources even if we're leaking storage space for the time being.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
```diff
@@ -58,26 +59,36 @@ final class MimirPTableStore[F[_]: Monad: LiftIO] private (
...
-        Stream.empty
-    }, {
-      case Some((blob, version, db)) =>
-        cake.commitDB(blob, version, db)
-
-      case None =>
-        IO.pure(())
-    })
+
+    def doWrite(completed: Signal[IO, Boolean]) =
```
djspiewak: Any particular reason to make this a separate inner function? It's only used once.

rintcius: I tend to create functions like this to make the structure more clear. In this case also because otherwise the contents of that function would have been the rhs in a step in a `for`-comprehension.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
rintcius: Added a TODO referring to the follow-up ticket that I previously created.

Ah yeah, I missed closing the db. `IO.fromFuture(IO(db.close))` doesn't seem to be meant to be called from here as it requires an implicit `ActorSystem`. Let me know if the alternative that I used isn't correct though.

rintcius:
[ch1873]

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
djspiewak: Not going to work, unfortunately. In the best case it closes nothing. In the worst case it closes the wrong thing! It's getting the HEAD version of a path that we haven't committed yet, so it's actually *guaranteed* to close something other than the right thing, if it closes anything at all.

I would solve the implicit `ActorSystem` problem by changing the signature of `NIHDB.close` and pushing the `implicit ActorSystem` parameter into the constructor of `NIHDBImpl`.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
djspiewak: Ah I see. It's a pretty subjective issue but I actually prefer the exact opposite. `for`-comprehensions aren't really structural, per se, and understanding what is going on requires understanding every individual right hand side. Breaking things out into one-use functions like this often ends up just obfuscating the work being performed. It's the moral equivalent of doing something like this:

```scala
val foo = 42
val bar = {
 something
 complicated
 here
}

// rewriting as:

def doBar = {
 something
 complicated
 here
}

val foo = 42
val bar = doBar
```

There isn't really any gain, and anyone trying to understand `foo` and `bar` in context is going to have to jump to `doBar` and back again.

Anyway, just a subjective opinion.

PR #3802, opened 2018-09-13
Coalesce mappable regions before ApplyProvenance 

wemrysi:
Inlines nullary regions and coalesces unary ones prior to computing provenance to avoid having to create existentials due to "dynamic" project/injects.

Also appears to speed up compilation in general due to shrinking the size of the graph earlier on in the process.

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/CoalesceUnaryMappable.scala
+++ b/qsu/src/main/scala/quasar/qsu/CoalesceUnaryMappable.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.qsu
+
+import quasar.fp._
+import quasar.qscript.RecFreeS._
+
+import scalaz.syntax.equal._
+
+/** Coalesces adjacent mappable regions of a single root. */
+final class CoalesceUnaryMappable[T[_[_]]] private () extends QSUTTypes[T] {
+  def apply(graph: QSUGraph): QSUGraph =
+    graph rewrite {
+      case g @ MappableRegion.MaximalUnary(src, fm) if g.root =/= src.root =>
+        g.overwriteAtRoot(QScriptUniform.Map(src.root, fm.asRec))
+    }
+}
+
+object CoalesceUnaryMappable {
+  def apply[T[_[_]]](graph: QSUGraph[T]): QSUGraph[T] =
+    (new CoalesceUnaryMappable[T])(graph)
+}
--- a/qsu/src/main/scala/quasar/qsu/InlineNullary.scala
+++ b/qsu/src/main/scala/quasar/qsu/InlineNullary.scala
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


wemrysi:
[ch1466]
[ch1816]

qsu/src/main/scala/quasar/qsu/CoalesceUnaryMappable.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import quasar.fp._
+import quasar.qscript.RecFreeS._
+
+import scalaz.syntax.equal._
+
+/** Coalesces adjacent mappable regions of a single root. */
+final class CoalesceUnaryMappable[T[_[_]]] private () extends QSUTTypes[T] {
+  def apply(graph: QSUGraph): QSUGraph =
+    graph rewrite {
```
djspiewak: This is probably going to be faster if you use `corewrite` rather than `rewrite`, since then we avoid re-traversing graph nodes and re-building `FreeMap`s.

qsu/src/main/scala/quasar/qsu/CoalesceUnaryMappable.scala
```diff
@@ -0,0 +1,36 @@
...
+      case g @ MappableRegion.MaximalUnary(src, fm) if g.root =/= src.root =>
+        g.overwriteAtRoot(QScriptUniform.Map(src.root, fm.asRec))
```
djspiewak: Note to self: we should add the following.

```scala
def modifyAtRoot(f: Symbol => QSU[Symbol]): QSUGraph
def modifyAtRootM[F[_]: Functor](f: Symbol => F[QSU[Symbol]]): F[QSUGraph]
```

It just occurred to me that this pattern (in various forms) is *insanely* common.

qsu/src/main/scala/quasar/qsu/InlineNullary.scala
```diff
@@ -0,0 +1,153 @@
...
+  * eliminating the AutoJoin.
+  */
+final class InlineNullary[T[_[_]]] private () extends QSUTTypes[T] {
+  import QSUGraph.Extractors._
+  import RecFreeS._
+
+  private val namePrefix: String = "iln"
+
+  def apply[F[_]: Monad: NameGenerator: RevIdxM](graph: QSUGraph): F[QSUGraph] =
+    graph rewriteM[F] {
```
djspiewak: TIL you can have explicit type parameters with space syntax. No idea how I didn't realize that was possible before. Excellent!

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -19,7 +19,6 @@ package quasar.yggdrasil.table
 import quasar.blueeyes._, json._
 import quasar.common.data.PreciseKeys
 import quasar.contrib.std.errorImpossible
-import quasar.frontend.data.DataCodec
```
djspiewak: 🎉

djspiewak:
IMO this merits `version: feature`. It's a pretty major improvement, and obviously it enables some important user-facing functionality which is *practically speaking* entirely masked right now.

qsu/src/main/scala/quasar/qsu/CoalesceUnaryMappable.scala
wemrysi: I agree and did so initially, which led to the 2hr regression tests =( I'll have another look and see if I can't find the cause.

qsu/src/main/scala/quasar/qsu/InlineNullary.scala
wemrysi: Same issue as with the other one.

qsu/src/main/scala/quasar/qsu/CoalesceUnaryMappable.scala
wemrysi: Ah, you know what, I think the issue is with the use of `MappableRegion.maximal` we're using it to try and find the maximal regions, but instead of finding a maximal unary or nullary region it finds the largest region it can and then checks if it is unary or nullary, which is much less efficient.

djspiewak: That terrifies me because I have no idea why that would be the case.

Don't worry about tracking it down now. We have bigger fish to fry. But… uh… we should probably try to understand why this happened at some point.

wemrysi: I think I have a quick fix, I'll report back shortly.

PR #3803, opened 2018-09-13
add signal to track whether write in MimirPTableStore has completed (with closing fix)

djspiewak:
Assigning this to @rintcius but I'm going to merge it myself since he's out. This is a straightforward correction of the remaining blocker issue in #3801


rintcius:
@djspiewak thanks!

PR #3805, opened 2018-09-14
More robust test of schema halting after time limit.

wemrysi:


wemrysi:
The previous test could fail due to a race condition.

djspiewak:
I did wonder about that test. :-) We might want to consider increasing `evalDelay` given the sorts of timings we normally see out of travis, but let's see how this goes first.

PR #3808, opened 2018-09-14
Upgrade qdata-version to 2.1.0

alissapajer:


djspiewak:
This is a feature change, rather than revision, since *in qdata* it was a feature change.

PR #3809, opened 2018-09-14
Avoid simplifying constant comparisons in outer joins

jsantos17:
This fixes the old `SimplifyJoin` rewrite. The new integration test will catch this bug when we finally move this rewrite to QSU.

### DIFF
--- a/qscript/src/main/scala/quasar/qscript/rewrites/SimplifyJoin.scala
+++ b/qscript/src/main/scala/quasar/qscript/rewrites/SimplifyJoin.scala
@@ -17,6 +17,7 @@
 package quasar.qscript.rewrites
 
 import quasar.fp._
+import quasar.RenderTreeT
 import quasar.contrib.iota._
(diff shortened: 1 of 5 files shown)


qscript/src/main/scala/quasar/qscript/rewrites/SimplifyJoin.scala
```diff
@@ -80,12 +81,12 @@ object SimplifyJoin {
           // TODO: This can potentially rewrite conditions to try to get left
           //       and right references on distinct sides.
           def alignCondition(l: JoinFunc[T], r: JoinFunc[T]): Option[EquiJoinKey[T]] =
-            if (l.element(LeftSide) && r.element(RightSide) &&
```
jsantos17: Reworked these conditions to simplify them.

jsantos17:
[ch1815]

alissapajer:
Needs version label.

PR #3810, opened 2018-09-17
Implement CONTAINS_KEY

djspiewak:
Heads up, @garyb. Publishing as snapshot version `68.2.0-31c90da` which is eviction compatible with 68.2.0. You should be able to test this within the launcher (by appending version directive `quasar:68.2.0-31c90da`).

@alissapajer I would mark this as ☠️ but github is being weird. Please review and merge asap.

[ch2071]

### DIFF
--- a/frontend/src/test/scala/quasar/std/StdLibSpec.scala
+++ b/frontend/src/test/scala/quasar/std/StdLibSpec.scala
@@ -1531,7 +1531,7 @@ abstract class StdLibSpec extends Qspec {
         "arbitrary double" >> prop { (d: Double) =>
           val n = BigDecimal(d)
           val data =
-            // testing with the generated double if it is not a Long or 
+            // testing with the generated double if it is not a Long or
             // if it is an exact double
             if ((d != d.toLong) || (n.isExactDouble)) Data.Dec(n)
             // .. but if it is a Long but not an exact double then we test with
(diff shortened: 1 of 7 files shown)


frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -3153,6 +3153,40 @@ abstract class StdLibSpec extends Qspec {
...
+        }
+
+        """CONTAINS_KEY("derp", "b")""" >> {
+          binary(
+            ContainsKey(_, _).embed,
+            Data.Str("derp"),
+            Data.Str("b"),
+            Data.Bool(false))
+        }
+      }
```
garyb: I assume the result is undefined if either argument is undefined too? You'll know better than me if that needs testing, just wanted to double check as it's not explicitly covered here.

djspiewak: Oh that does need a test. More importantly, I think this *specific* test case is actually wrong and it should be undefined. I'll fix.

alissapajer:
@djspiewak compilation failed.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
garyb: Ah yeah, it should be undefined from what you said in the issue :smile:

alissapajer: @djspiewak can you push a change with the arguments undefined as well, as @garyb suggested?

djspiewak: One thing at a time. :-) I pushed a fix for the test so it's not exactly in spec.

PR #3812, opened 2018-09-17
Disable publishing documentation artifact.

alissapajer:


alissapajer:
This doesn't disable `sbt doc`. It does disable publishing the documentation.

alissapajer:
I tested it out, and we can add
```
sources in (Compile, doc) := List()
```
to our `buildSettings` (as well as individually to each precog project) if we want to disable `sbt doc` for all projects entirely.

djspiewak:
Anyone who was relying on the existence of the scaladoc artifact in a breaking fashion deserves to be broken. :-D

PR #3813, opened 2018-09-17
Split array_length from length (which is for strings)

djspiewak:
Heads up @garyb 

Had to do this as a `TransSpec` because it fiddles with arrays (which is structural) and it was enormously inefficient to do it without `BitSet` manipulation. As it is, the implementation could probably be improved somewhat, but it's not bad. You'll notice that it is optimized for homogeneous array lengths by reversing the column list (during the sort).

I'll add more tests in mimir in a separate PR. For now, the stdlibspecs should be enough to make forward progress.

[ch2072]


alissapajer:
[ch212]

djspiewak:
@alissapajer Do you think this should be **breaking**? Technically I'm changing SQL², but then again, it wouldn't have worked before anyway.

qscript/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -92,7 +93,6 @@ object MapFunc {
       : scala.PartialFunction[BinaryFunc, (A, A) => MF[A]] = {
     // NB: ArrayLength takes 2 params because of SQL, but we really don’t care
     //     about the second. And it shouldn’t even have two in LP.
```
alissapajer: Delete this comment.

qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -917,6 +921,9 @@ object MapFuncsCore {
   @Lenses final case class JoinSideName[T[_[_]], A](name: Symbol) extends Nullary[T, A]
 
   // array
+  @Lenses final case class ArrayLength[T[_[_]], A](a1: A) extends Unary[T, A]
+
+  // string
   @Lenses final case class Length[T[_[_]], A](a1: A) extends Unary[T, A]
```
alissapajer: We could rename this to `StringLength` for explicitness.

djspiewak: Probably should. Let's ticket that.

qscript/src/main/scala/quasar/qscript/MapFunc.scala
alissapajer: Does it still have two params in LP?

djspiewak: Nope.

qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
alissapajer: ch2076

it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
```diff
@@ -140,13 +140,13 @@ class MimirStdLibSpec extends StdLibSpec with PrecogCake {
...
+        lazy val primary = ((result must haveSize(0)) or
           (result.head must beCloseTo(expected)))
 
-        lazy val fallback = ((result must haveSize(0))) and
-          (expected mustEqual Data.NA)
+        lazy val fallback = (expected mustNotEqual Data.NA) or
+          ((result must haveSize(0)))
 
-        (primary or fallback).toResult
+        (primary and fallback).toResult
```
alissapajer: What changed to require this change in logic? If a `result` has size `0` then the test will pass even if the test expectation was not `Data.NA`.

alissapajer: And the only time a `0` size result is correct is when the test expectation is `Data.NA`.

djspiewak: I probably should have left a comment explaining. This was previously encoded as (a ∧ b) ∨ (c ∧ d), where a and c are mutually exclusive. This resulted in some pretty terrible error messages, stemming from the fact that it's simply logically incorrect. I changed it to (¬a ∨ b) ∧ (¬c ∨ d), which is to say, (a ⇒ b) ∧ (c ⇒ d).

alissapajer: What I'm observing is that if `result must haveSize(0)` returns `true` then the entire spec will pass, regardless of the value of `expected`. And that cannot be correct. Previously if `result must haveSize(0)` returned `true`, it would also have to be the case that `expected mustEqual Data.NA` returned `true` in  order for the spec to pass.

djspiewak: Curse you giant resolve button that looks like reply!

I think you’re right. This tests the converse, rather than the actual assertion.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
```diff
@@ -188,6 +188,34 @@ abstract class StdLibSpec extends Qspec {
...
+      "ArrayLength" >> {
+        "empty" >> {
+          unary(ArrayLength(_).embed, Data.Arr(Nil), Data.Int(0))
+        }
+
+        "singleton" >> {
+          unary(ArrayLength(_).embed, Data.Arr(List(Data.Int(42))), Data.Int(1))
+        }
+
+        "three things" >> {
```
alissapajer: The nature of the unit tests can only test a single row at a time, meaning we cannot test against data that has multiple scalar types at the same array index. Or against data that has arrays of different lengths per row.

Can you include an integration test against some heterogeneous array data? There is special logic in the implementation to make sure we include all the columns at a particular array index, so it would be good to have a test for this. The data could be something like:
```
{"row": 1, "value": [1, "foo", true, null] }
{"row": 2, "value": [1, "bar", "baz"] }
{"row": 3, "value": [{"key": 42, "value": false}, 2, 3, 4, 5] }
{"row": 4, "value": [8, [5], "ack"] }
{"row": 5, "value": [] }
{"row": 6, "value": 19 }
{"row": 7, "value": {} }
{"row": 8, "value": { "name": "yes", "date": 12 } }
{"row": 9 }
```
and the query:
```
select row, array_length(value) from `heterogeneousArray.data`
```

alissapajer: Without this (or without unit tests for `Slice.arrayLength`, which would be another way to go) most of the logic there isn't being tested.

alissapajer:
@djspiewak even if we're changing broken sql2, we're still changing sql2. So I'd say we should make it breaking.

djspiewak:
@alissapajer But is it a breaking change if no one could have had a working query previously? The way I'm thinking about it is this: either there was a non-working query using `array_length` which now works, or there was a query which didn't have anything to do with `array_length` at all. In either case, it's exactly the same as if I *added* the function afresh, which is **feature** not **breaking**.

frontend/src/test/scala/quasar/std/StdLibSpec.scala
djspiewak: I intend to add both.

alissapajer:
@djspiewak if someone found some hacky way to use `array_length` that was taking advantage of its brokenness, then this version would be breaking their query. If we say that we never supported `array_length` previously, then we can go with feature.

djspiewak:
@alissapajer That's kind of what I'm thinking. The only sorts of queries which would have worked would have been something like `select array_length("foo", 1) from ...` (also allowing non-constant strings there), and this would have only worked between when you removed typechecking and now.

it/src/main/resources/tests/arrayLengthHeterogeneous.test
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "array_length over a heterogeneous dataset",
+    "data": "heterogeneous.data",
+    "query": "select array_length(b) from `heterogeneous.data`",
+    "predicate": "exactly",
+    "expected": [2, 0, 1, 6, 7]
```
alissapajer: I'd really prefer this to be 
```
{ "a": 2, "length": 2 }
{ "a": 4, "length": 0 }
{ "a": 6, "length": 1 }
{ "a": 10, "length": 6 }
{ "a": 11, "length": 7 }
```
since it makes it more clear where the results are coming from (and verifies they are coming from the rows we expect them to).

it/src/test/scala/quasar/physical/mimir/MimirStdLibSpec.scala
```diff
@@ -140,13 +140,13 @@ class MimirStdLibSpec extends StdLibSpec with PrecogCake {
...
-        lazy val primary = ((result must haveSize(1)) and
+        lazy val primary = ((result must haveSize(0)) or
           (result.head must beCloseTo(expected)))
 
-        lazy val fallback = ((result must haveSize(0))) and
+        lazy val fallback = ((result must haveSize(1))) or
           (expected mustEqual Data.NA)
 
-        (primary or fallback).toResult
+        (primary and fallback).toResult
```
alissapajer: This looks correct :+1:

it/src/main/resources/tests/arrayLengthHeterogeneous.test
```diff
@@ -0,0 +1,21 @@
...
+      {"a":4,"r":0},
+      {"a":5},
+      {"a":6,"r":1},
+      {"a":7},
+      {"a":8},
+      {"a":9},
+      {"a":10,"r":6},
+      {"a":11,"r":7},
+      {"a":12},
+      {"a":13}
```
alissapajer: outer object concat semantics ftw

djspiewak: Yeah. I considered several other approaches here, but I decided this made the most sense from a test construction standpoint to ensure all the cases were hit.

alissapajer: :+1: this is consistent with what @rintcius did in other tests as well.

PR #3830, opened 2018-09-24
add join ITs

rintcius:
putting PR up to discuss the results..  

looking at the ITs added in this PR, it looks like the semantics of the different joins is as follows:

```
INNER JOIN         left key = right key    
LEFT OUTER JOIN    left key = right key OR left key is defined
RIGHT OUTER JOIN   left key = right key OR                        right key is defined
FULL OUTER JOIN    left key = right key OR left key is defined OR right key is defined
```

This looks actually reasonable to me, but I guess we should instead also be including undefined keys? I.e.:
```
INNER JOIN         left key = right key    
LEFT OUTER JOIN    left key = right key OR left key is un/defined
RIGHT OUTER JOIN   left key = right key OR                           right key is un/defined
FULL OUTER JOIN    left key = right key OR left key is un/defined OR right key is un/defined
```
which means that for `LEFT OUTER JOIN` the left part is always included, for `RIGHT OUTER JOIN` the right part is always included and for `FULL OUTER JOIN` both the left and right part is always included.

relates to ch875


rintcius:
@alissapajer @djspiewak can you confirm that I have to change the 3 outer join ITs added here (by including undefined key values to the result set too)?

rintcius:
@djspiewak ok thanks. Question for `FULL OUTER JOIN`. I guess the left and right `undefined`s need to be summed and not multiplied, right?
I.e the expected result for the test query is:
```
        {"a":1,"b":1,"nr1":1,"nr2":4},
        {"a":2,"nr1":2},
        {"a":2,"nr1":7},
        {"a":3,"nr1":3},
        {"b":4,"nr2":5},
        {"b":4,"nr2":7},
        {"a":6,"b":6,"nr1":8,"nr2":8},
        // 7 rows due to undefineds
        // 3 left undefineds
        {"nr1":4},
        {"nr1":5},
        {"nr1":6},
        // 4 right undefineds
        {"nr2":1},
        {"nr2":2},
        {"nr2":3},
        {"nr2":6}
```
and not:
```
        {"a":1,"b":1,"nr1":1,"nr2":4},
        {"a":2,"nr1":2},
        {"a":2,"nr1":7},
        {"a":3,"nr1":3},
        {"b":4,"nr2":5},
        {"b":4,"nr2":7},
        {"a":6,"b":6,"nr1":8,"nr2":8},
        // 12 rows due to undefineds (3 left x 4 right)
        {"nr1":4,"nr2":1},
        {"nr1":4,"nr2":2},
        {"nr1":4,"nr2":3},
        {"nr1":4,"nr2":6},
        {"nr1":5,"nr2":1},
        {"nr1":5,"nr2":2},
        {"nr1":5,"nr2":3},
        {"nr1":5,"nr2":6},
        {"nr1":6,"nr2":1},
        {"nr1":6,"nr2":2},
        {"nr1":6,"nr2":3},
        {"nr1":6,"nr2":6}
```

djspiewak:
Definitely summed. My guess is the reason they are being multiplied is I believe Mimir considers `undefined = undefined` to be `true`, which is a bit bizarre and also inconsistent with what the rest of quasar thinks it is.

rintcius:
@djspiewak yes, I guessed the same. Can we change `undefined = undefined` to be `false` easily at 1 place?

djspiewak:
@rintcius I don't think so. Or rather, we can change the meaning of the `=` operator itself in one-ish place (https://github.com/djspiewak/quasar/blob/master/yggdrasil/src/main/scala/quasar/yggdrasil/table/SliceTransform.scala#L66-L196), but I don't think that would change the meaning of `cogroup`.

PR #3832, opened 2018-09-24
Delete a bunch more unused qscript rewrites and analyses

alissapajer:


alissapajer:
@wemrysi found more, pushed another commit

PR #3834, opened 2018-09-26
Add integration tests for top-level shifting

alissapajer:


alissapajer:
These are for shift parsing. Getting these in against master so we have evidence that shift parsing isn't changing the results.

PR #3836, opened 2018-09-26
include rows when outer join key is undefined

rintcius:
Now includes rows when outer join key is undefined. This changes the results of left, right and full outer joins if there are undefined join keys.

Full outer joins were originally (that is, with the fix for left and right joins) multiplying rows with undefined keys, but are in ae96456 changed to be summing them instead.

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
+++ b/mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
@@ -156,7 +156,15 @@ final class EquiJoinPlanner[
               }
               newSortOrder = rephrase2(transMiddle, transLKey, transRKey)
               newSortState = newSortOrder.map(order => SortState(None, order :: Nil))
-              joinedTable = lsorted.cogroup(transLKey, transRKey, rsorted)(transLeft, transRight, transMiddle)
+              lcompactSorted = tpe match {
+                case JoinType.LeftOuter | JoinType.FullOuter => lsorted
+                case JoinType.Inner | JoinType.RightOuter => lsorted.compact(transLKey)
+              }
+              rcompactSorted = tpe match {
+                case JoinType.RightOuter | JoinType.FullOuter => rsorted
+                case JoinType.Inner | JoinType.LeftOuter => rsorted.compact(transRKey)
+              }
+              joinedTable = lcompactSorted.cogroup(transLKey, transRKey, rcompactSorted)(transLeft, transRight, transMiddle)
             } yield (joinedTable, newSortState)
           }
         }
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
@@ -622,9 +622,13 @@ trait ColumnarTableModule
       * Cogroups this table with another table, using equality on the specified
       * transformation on rows of the table.
       */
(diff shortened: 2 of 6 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -787,9 +791,19 @@ trait ColumnarTableModule
                   // We're currently in a cartesian.
                   if (lpos < lhead.size && rpos < rhead.size) {
                     comparator.compare(lpos, rpos) match {
-                      case LT if rightStartSliceId == rSliceId =>
```
rintcius: I opted to change the behavior locally only. I guess we could also change the `comparator` instead but then the change would be less local. Thought the latter option would be more appropriate once we start `undefined != undefined`

yggdrasil/src/test/scala/quasar/yggdrasil/CogroupSpec.scala
```diff
@@ -406,14 +406,21 @@ trait CogroupSpec extends TableModuleTestSupport with SpecificationLike with Sca
...
+      parseUnsafe("""{ "val" : 1 }"""),
       parseUnsafe("""{ "val" : 3 }"""),
+      parseUnsafe("""{ "id" : "foo", "val" : 2 }"""),
       parseUnsafe("""{ "id" : "foo", "val" : 4 }"""))))
 
     val expected = Stream(
+      parseUnsafe("""{ "val": 0 }"""),
+      parseUnsafe("""{ "val": 2 }"""),
+      parseUnsafe("""{ "val" : 1 }"""),
+      parseUnsafe("""{ "val" : 3 }"""),
```
rintcius: I don't understand why this doesn't return this instead:
```
      parseUnsafe("""{ "left": 0 }"""),
      parseUnsafe("""{ "left": 2 }"""),
      parseUnsafe("""{ "right" : 1 }"""),
      parseUnsafe("""{ "right" : 3 }"""),
```

... but the IT returns the expected results, so I guess I'm missing something in the impl of `cogroup`

rintcius:
[ch1965]

rintcius:
verified [ch875] : result looks ok now

yggdrasil/src/test/scala/quasar/yggdrasil/CogroupSpec.scala
djspiewak: Note the `Leaf(Source)` transspecs passed to `cogroup` below. So these results are correct.

mimir/src/main/scala/quasar/mimir/EquiJoinPlanner.scala
```diff
@@ -156,7 +156,15 @@ final class EquiJoinPlanner[
               }
               newSortOrder = rephrase2(transMiddle, transLKey, transRKey)
               newSortState = newSortOrder.map(order => SortState(None, order :: Nil))
-              joinedTable = lsorted.cogroup(transLKey, transRKey, rsorted)(transLeft, transRight, transMiddle)
+              lcompactSorted = tpe match {
+                case JoinType.LeftOuter | JoinType.FullOuter => lsorted
+                case JoinType.Inner | JoinType.RightOuter => lsorted.compact(transLKey)
+              }
```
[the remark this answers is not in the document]
rintcius: I think they are covered by the ITs (inner & left/right/full outer self joins) that I added before this PR: https://github.com/slamdata/quasar/pull/3830/files
Let me know if you mean something else.

PR #3840, opened 2018-09-27
Ability to preserve array indices in SSTs

wemrysi:
Updates the array type representation to allow for both known indices of a specific type along with an optional type representing the LUB of the rest of the array, similar to how maps are represented.

Adds a new option to SST schema configuration to preserve a specified number of indices when compressing long arrays.


alissapajer:
@wemrysi is there a ch story for this?

wemrysi:
@alissapajer Yes, ch671, but this doesn't complete it. The corresponding sdbe PR that exposes the option in the HTTP api will do so.

PR #3848, opened 2018-10-04
Make join test pending instead of skipped

alissapajer:
This tests returns results in a few seconds. The results are a very large number of rows each with `{"b":"CHICO"}`, like so:
```
{"b":"CHICO"}
{"b":"CHICO"}
{"b":"CHICO"}
...
```


alissapajer:
@djspiewak @beckyconning bringing this to your attention as joins are coming soon to the product and I don't know why this is failing. Created ch2290.

PR #3851, opened 2018-10-04
I ask nothing more than their Slice on a Plate!

djspiewak:
First cut at `SlicePlate`. There probably should be a ticket for this. I forgot to make one.

Anyway, it's missing tests for hitting column limits and things, but the code which implements that is very simple. I'll come back and write tests which specifically exercise those cases. For now, this is enough to get us started.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -132,6 +132,8 @@ object Dependencies {
   )
 
   def yggdrasil = Seq(
+    "com.slamdata" %% "tectonic" % tectonicVersion,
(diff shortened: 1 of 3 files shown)


djspiewak:
[ch1072]

alissapajer:
Totally fine merging PRs lacking tests to keep us moving. But I wouldn't feel comfortable going into production until the tests are added :)

djspiewak:
@alissapajer It has tests. What is *not* tested is the column/row limit breaking. Everything else is tested and solid though.

PR #3854, opened 2018-10-05
Move `CPath` and rid it of its scary implicits

alissapajer:
`CPath` previously contained numerous `implicit` constructors which would work together (or not) in horrifyingly surprising ways. This PR removes them, as well as removes other unused machinery.

`CPath` also now lives in `quasar.common` since it is going to become our default path representation.


alissapajer:
Forgot about `SlicePlate`. Going to rebase this to get all the commits compiling.

djspiewak:
One caveat… this is technically **breaking** since it affects the published binary API. I think it's very unlikely that anything downstream relies on `CPath`, but perhaps it should at least be **feature** since it adds a new API (if we ignore the fact that it deletes an old one).

alissapajer:
@djspiewak since it's technically breaking, I think we should version as breaking. I don't know of anything that relies on `CPath`, but I think it's best practice to have the version reflect the binary reality.

PR #3856, opened 2018-10-05
Adds ParseInstruction algebra

alissapajer:
Depends on #3854 

[ch2309]


frontend/src/main/scala/quasar/ParseInstruction.scala
```diff
@@ -0,0 +1,65 @@
...
+   *   - `IdOnly` returns the value unwrapped.
+   *   - `ExcludeId` returns the value unwrapped.
+   */
+  final case class Pivot(path: CPath, idStatus: IdStatus, structure: CompositeType, retain: Boolean)
+      extends ParseInstruction
+
+  /* The instructions that must be interpreted sequentially while parsing. These
+   * instructions are not optional and will result in incorrect query results if ignored.
+   */
+  final case class ParseInstructions(instructions: List[Set[ParseInstruction]])
```
alissapajer: I'm not fully convinced yet that a `List[Set[...]]` is what we want here. It's very easy for a `Set[ParseInstruction]` to produce different results, depending on the order the instructions are interpreted. For example, if you have a `Wrap` and then a `Mask` masking out anything that is not an object at the wrapped path, you can get different results depending on the order.

On the other hand, we need a way to represent instructions that must be executed in parallel, for example, if multiple `Mask`s are to be applied, if we apply them sequentially, the first will mask out the values that the second `Mask` instructs us to keep.

We could simply say that there is no guarantee in the order that the inner `Set` will be interpreted, and so it's the constructors job to ensure that they commute. But I feel like there could be some property here we can encode. I have to think about it more.

frontend/src/main/scala/quasar/ParseInstruction.scala
```diff
@@ -0,0 +1,65 @@
...
+  
+  /* Wraps the provided `path` into an object with key `name`, thus adding
+   * another layer of structure. All other paths are retained.
+   */
+  final case class Wrap(path: CPath, name: String) extends ParseInstruction
+  
+  /* Removes all values that are not both at the path `path` and of the type `tpe`.
+   * An empty set indicates that all values should be retained.
+   */
+  final case class Mask(path: CPath, tpe: Set[PrimaryType]) extends ParseInstruction
```
djspiewak: Why isn't this `Set[CType]`?

frontend/src/main/scala/quasar/ParseInstruction.scala
```diff
@@ -0,0 +1,65 @@
...
+   *   1) In the case of a successful pivot, if surrounding structure should be retained.
+   *   2) In the case of an unsuccessful pivot (`path` does not reference a value of
+   *      the provided `structure`), if the row should be returned.
+   *
+   * `idStatus` determines how the values is returned:
+   *   - `IncludeId` wraps the key/value pair in a two element array.
+   *   - `IdOnly` returns the value unwrapped.
+   *   - `ExcludeId` returns the value unwrapped.
+   */
+  final case class Pivot(path: CPath, idStatus: IdStatus, structure: CompositeType, retain: Boolean)
```
djspiewak: TIL `CompositeType`…

mimir/src/test/scala/quasar/mimir/TransSpecable.scala
```diff
@@ -59,42 +60,11 @@ trait TransSpecableModule
       def Cond(node: dag.Cond)(pred: T, left: T, right: T): T
     }
 
-    def isTransSpecable(to: DepGraph, from: DepGraph): Boolean =
```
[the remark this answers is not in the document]
alissapajer: Yeah!

frontend/src/main/scala/quasar/ParseInstruction.scala
alissapajer: I agree that having a semantic of "true parallel" is what we would want here. The "compilation into plates would fail" component is interesting. This would be a real logical layer, probably existing in quasar, that should actually prevent us from constructing these inner `Set`s in the first place. 

I'm tempted to approach this more constructively though. There are a lot of combinations of instructions that do not compose, and so most `Set`s (when arbitrarily constructed) would be rejected as invalid. So, I'd like to come up with a list of the instruction tuplets that we'd actually intend to interpret in parallel. Then we can decide if it makes sense to instead have a more restricted type than `Set[ParseInstruction]`.

alissapajer: We definitely need `Set[Mask]`. I'm not convinced yet that an other operations actually compose, still thinking...

djspiewak: I'm okay with using `Set[ParseInstruction]` for now in the interest of forward progress. I agree I'd rather have uncompilable things be unrepresentable if possible.

alissapajer: Without considering the specific instances of the operations (what fields they are pivoting, etc):
- `Wrap` and `Mask` do not compose.
- `Wrap` does not compose with itself.
- `Pivot` composes with nothing.

That  leaves us only `Set[Mask]`.

@djspiewak from a quasar perspective, it would be easiest (least work) to just allow `Set[Mask]`. But, the goal is to cater to `tectonic`. So, are there any concrete examples of tuples of instructions that compose and that we'd benefit from interpreting in parallel (or need to interpret in parallel) in `tectonic`?

alissapajer: @djspiewak just saw your comment. What do you think of allowing `Set[Mask]` only? I don't actually have a concrete example of what else we'd include in the inner set.

frontend/src/main/scala/quasar/ParseInstruction.scala
alissapajer: Ah, so this is in part an artifact from before `Pivot` had a `retain` field. My original idea was that `Pivot` would always implicitly have `retain = true` and then pre-pivot we'd mask out all values not needed. So, if we knew we were object-pivoting at path `foo.bar.baz` _and_ no other values were needed from that path, then we'd mask out all non-objects at that path.

If we sql2 type-checked dynamic object/array projections, then we could take advantage of including `CompositeType` here, but we're a long way from doing that.

I feel like there _might_ be another reason we still need this, but I don't have a concrete example. So, in order to keep everything as limited as possible, I'll change this to `SimpleType`. If we need to expand that type to include composites, we always can.

Which brings me to the differences between `SimpleType` and `CType`:
- `CType` includes `CEmptyObject`, `CEmptyArray`, and the temporal types. We don't need the temporal types since we don't parse into them in `tectonic`. Do we care about the empty composite types?
- `SimpleType` includes only two numeric types (`Int` and `Dec`) while `CType` includes three (`CDouble`, `CNum`, `CLong`). What's the parsing story for numeric types in `tectonic`?

As a side note, I'd have to move `CValue.scala` out of `blueeyes` if we want to use it here, but that's not a big deal.

frontend/src/main/scala/quasar/ParseInstruction.scala
alissapajer: :+1: to the overlapping ADT, already had that in mind.

If you pivot something that has the same structure as the internal structure, that doesn't compose.

alissapajer:

frontend/src/main/scala/quasar/ParseInstruction.scala
```diff
@@ -0,0 +1,71 @@
...
+   *   2) In the case of an unsuccessful pivot (`path` does not reference a value of
+   *      the provided `structure`), if the row should be returned.
+   */
+  final case class Pivot(path: CPath, idStatus: IdStatus, structure: CompositeParseType)
+      extends ParseInstruction
+
+  /* The instructions that must be interpreted sequentially while parsing. These
+   * instructions are not optional and will result in incorrect query results if ignored.
+   */
+  final case class ParseInstructions(instructions: List[ParseInstruction])
```
djspiewak: Is there value in newtyping this? Also it's a bit awkward: `ParseInstruction.ParseInstructions`.

alissapajer: Ah yeah I had a comment about the newtype in my originally gist, where the comment was "should we newtype this?" :) The `ParseInstruction`s are designed so that an ordered sequence of them can be passed to an interpreter, so I was thinking that the newtype makes this clearer. 

But as I think about it now, I'm not convinced that the newtype actually belongs in `object ParseInstruction`. One way to answer this would be to decide if the `List` is relevant to what it means to talk about a `ParseInstruction` or not.

I'll delete it from here for now. Then when we actually generate these instructions, we can decide then if we'd benefit from a newtype or not. Passing along a `List` might turn out to be just fine.

PR #3857, opened 2018-10-08
Restore string-specific measure to SSTs

wemrysi:
Restores the `Str` case to `TypeStat` in order to collect min/max values in addition to lengths for strings.

[ch2352]

### DIFF
--- a/datagen/src/main/scala/quasar/datagen/dist.scala
+++ b/datagen/src/main/scala/quasar/datagen/dist.scala
@@ -34,6 +34,7 @@ import monocle.syntax.fields.{_2, _3}
 import monocle.std.option.{some => someP}
 import scalaz.{-\/, \/-, ==>>, Bifunctor, Equal, IList, INil, NonEmptyList, Order, Tags}
 import scalaz.Scalaz._
+import scalaz.syntax.tag._
 import spire.algebra.{AdditiveMonoid, Field, IsReal, NRoot}
 import spire.math.ConvertableFrom
 import spire.random.{Dist, Gaussian}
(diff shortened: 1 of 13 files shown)


alissapajer:
Reflecting on how meaningful min/max values for string are. If the string values are all numeric (or all temporal), say, then both the min and max will be numeric (or temporal), which would confirm that _all_ values sampled are numeric (or temporal).

datagen/src/main/scala/quasar/datagen/dist.scala
```diff
@@ -262,27 +263,41 @@ object dist {
...
       case TypeStat.Bool(ts, fs) =>
         val total = ts + fs
-        Dist.weightedMix(
+        some(Dist.weightedMix(
           ((ts / total).toDouble, Dist.constant(true)),
-          ((fs / total).toDouble, Dist.constant(false))
-        ) map (EJson.bool(_))
+          ((fs / total).toDouble, Dist.constant(false))).map(EJson.bool(_)))
 
       case TypeStat.Char(ss, cn, cx) =>
```
alissapajer: Why does it make sense to have both `Char` and `Str` type stats? In the case of a string value, which `TypeStat` do we generate?

datagen/src/main/scala/quasar/datagen/dist.scala
```diff
@@ -262,27 +263,41 @@ object dist {
...
 
       case TypeStat.Dec(ss, mn, mx) =>
-        gaussian(ss)
+        some(gaussian(ss)
           .cata((Dist.gaussian[A] _).tupled, Dist.constant(ss.mean))
-          .map((EJson.dec[J](_)) <<< clamp(mn, mx) <<< ((_: A).toBigDecimal))
+          .map((EJson.dec[J](_)) <<< clamp(mn, mx) <<< ((_: A).toBigDecimal)))
+
+      case TypeStat.Coll(_, _, _) => none
+      case TypeStat.Count(_) => none
```
alissapajer: These two types aren't new; do we never generate them?

datagen/src/main/scala/quasar/datagen/dist.scala
wemrysi: Mostly so we can represent structural strings as `Array[Char]` and accumulate statistics for individual characters.

datagen/src/main/scala/quasar/datagen/dist.scala
wemrysi: Not from this function. We weren't previously either, but I reformulated the function so that it'd benefit from exhaustivity checking as it required modification when I added the new case but the compiler didn't complain due to the use of `collect` in the previous impl.

PR #3862, opened 2018-10-09
Tectonic benchmarks

djspiewak:
This also adds a `Table.parseJson` function which takes a misleading boolean which pretends to handle precise… but doesn't. Support incoming soon! The preliminary results here are really exciting though. Tectonic is between 10% and 90% faster than jawn + qdata + rvalue, and there's plenty of optimization yet to do.

Brings in #3861


yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -364,6 +364,46 @@ trait ColumnarTableModule
...
+        implicit ec: ExecutionContext)
+        : M[Table] = {
+
+      import fs2.{Chunk, Stream}
+      import tectonic.AsyncParser
+
+      val parser = AsyncParser(new SlicePlate, AsyncParser.ValueStream)
+
+      val absorbed: Stream[IO, Chunk[Slice]] =
+        bytes.chunks.evalScan(Chunk.empty[Slice]) { (_, bc) =>
```
rintcius: looks redundant to call `evalScan` without using the accumulator (ie the `_` in `(_, bc)`). 
Would calling a simpler method like `evalMap` work here?

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/TableConstructionBenchmark.scala
```diff
@@ -0,0 +1,205 @@
...
+        RObject(List.tabulate(keysPerObject)(i =>
+          ("k" + ((s * keysPerObject + i) % keys)) -> scalar
+        ).toMap)))
+
+  def createAndConsumeTable(data: fs2.Stream[IO, Byte], bh: Blackhole): IO[Unit] = {
+    def parseWithJawn: fs2.Stream[IO, RValue] = {
+      implicit val facade = qdata.json.QDataFacade.qdata[RValue]
+      val parser = jawn.AsyncParser[RValue](jawn.AsyncParser.ValueStream)
+
+      val absorbtion =
```
rintcius: `absorption`

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/TableConstructionBenchmark.scala
```diff
@@ -0,0 +1,205 @@
...
+        data.chunks.scan(fs2.Chunk.empty[RValue]) { (_, chunk) =>
```
rintcius: similar here: can we use `data.chunks.map`?

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
djspiewak: Uh, yes, `evalMap` would definitely work. I'm not sure why I fixated on scan.

yggdrasilPerf/src/main/scala/quasar/yggdrasil/perf/TableConstructionBenchmark.scala
djspiewak: Yep

PR #3863, opened 2018-10-10
Moved persist out of the PreparationsManager

djspiewak:
`flatMap` for the win. This sequences the effect of *planning* the query with the effect of *running* the query. I should write tests for this… Probably don't merge yet.

[ch2223]

### DIFF
--- a/impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
+++ b/impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
@@ -66,60 +66,61 @@ class PreparationsManager[F[_]: Effect, I, Q, R] private (
       back <- if (check) {
         Condition.abnormal(InProgressError(tableId)).point[F]
       } else {
-        for {
-          result <- evaluator.evaluate(query)
-          persist <- runToStore(tableId, result)
+        val persist = for {
+          result <- Stream.eval(evaluator.evaluate(query))
+          persistStream <- Stream.eval(runToStore(tableId, result))
+          results <- persistStream
+        } yield results
 
-          configured = Stream.eval(F.delay(OffsetDateTime.now())) flatMap { start =>
-            val preparation = Preparation(cancel, start)
-            val halted = persist.interruptWhen(s)
+        val configured = Stream.eval(F.delay(OffsetDateTime.now())) flatMap { start =>
+          val preparation = Preparation(cancel, start)
+          val halted = persist.interruptWhen(s)
 
-            val handled = halted handleErrorWith { t =>
-              val eff = for {
-                end <- F.delay(OffsetDateTime.now())
-                _ <- s.set(true)    // prevent the onComplete handler
+          val handled = halted handleErrorWith { t =>
+            val eff = for {
+              end <- F.delay(OffsetDateTime.now())
+              _ <- s.set(true)    // prevent the onComplete handler
 
-                _ <- notificationsQ.enqueue1(
-                  Some(
-                    PreparationEvent.PreparationErrored(
-                      tableId,
-                      start,
-                      (end.toEpochSecond - start.toEpochSecond).seconds,
-                      t)))
-              } yield ()
+              _ <- notificationsQ.enqueue1(
+                Some(
+                  PreparationEvent.PreparationErrored(
+                    tableId,
+                    start,
+                    (end.toEpochSecond - start.toEpochSecond).seconds,
+                    t)))
+            } yield ()
 
-              Stream.eval_(eff)
-            } onComplete {
-              val eff = for {
-                canceled <- s.get
+            Stream.eval_(eff)
+          } onComplete {
+            val eff = for {
+              canceled <- s.get
 
-                _ <- if (canceled) {
-                  ().point[F]
-                } else {
-                  for {
-                    end <- F.delay(OffsetDateTime.now())
+              _ <- if (canceled) {
+                ().point[F]
+              } else {
+                for {
+                  end <- F.delay(OffsetDateTime.now())
 
-                    _ <- notificationsQ.enqueue1(
-                      Some(
-                        PreparationEvent.PreparationSucceeded(
-                          tableId,
-                          start,
-                          (end.toEpochSecond - start.toEpochSecond).seconds)))
-                  } yield ()
-                }
-              } yield ()
+                  _ <- notificationsQ.enqueue1(
+                    Some(
+                      PreparationEvent.PreparationSucceeded(
+                        tableId,
+                        start,
+                        (end.toEpochSecond - start.toEpochSecond).seconds)))
+                } yield ()
+              }
+            } yield ()
 
-              Stream.eval_(eff)
-            }
-
-            Stream.bracket(
-              F.delay(status.replace(tableId, -\/(cancel), \/-(preparation))))(
-              flag => if (flag) handled else Stream.empty,
-              _ => F.delay(status.remove(tableId, \/-(preparation))).void)
+            Stream.eval_(eff)
           }
 
-          _ <- background(configured.drain)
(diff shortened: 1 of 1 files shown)


alissapajer:
@djspiewak interesting, the `cancel preparation when pending` test failed in appveyor but not on travis. Is that test ib a nondeterministic state?

djspiewak:
Okay, unit tests are now in place verifying this works in a vacuum. I'd like to try an end-to-end test to ensure that it resolves the originating bug, but that doesn't have to hold up the merge. This has been published as **74.0.0-d5c16aa**; @jsantos17, when you get a chance, can you check to see if the original issue is still reproducible with that snapshot?

impl/src/test/scala/quasar/impl/table/PreparationsManagerSpec.scala
```diff
@@ -245,33 +244,33 @@ object PreparationsManagerSpec extends Specification {
...
+        // enqueue the first table, which will never finish preparing
+        _ <- Stream.eval(manager.prepareTable(Id1, ()))
 
-        status1 <- Stream.eval(manager.cancelPreparation(Id))
+        // enqueue the second, which we will test
+        _ <- Stream.eval(manager.prepareTable(Id2, ()))
+        initStatus <- Stream.eval(manager.preparationStatus(Id2))
+        _ <- Stream.eval(IO(initStatus mustEqual Status.Pending))
+
+        status1 <- Stream.eval(manager.cancelPreparation(Id2))
```
alissapajer: This shows we can cancel the second preparation, even though the first preparation hasn't completed, which shows that we are indeed responding immediately from the preparation endpoint. :+1:

impl/src/test/scala/quasar/impl/table/PreparationsManagerSpec.scala
```diff
@@ -348,6 +347,19 @@ object PreparationsManagerSpec extends Specification {
...
     }
+
+    "immediately return when evaluation takes forever" in {
+      val results = for {
+        manager <- PreparationsManager[IO, Unit, Unit, Unit](
+          _ => IO.never)(
+          (_, _) => IO.pure(Stream.empty))
+
+        _ <- Stream.eval(manager.prepareTable((), ()))
+        _ <- Stream.eval(manager.cancelAll)
```
alissapajer: Could we additionally verify here that we have successfully cancelled the preparation?

jsantos17:
@djspiewak I have tested this and aside from OOME also present in master, I discovered the `/table/<uuid>/preparation` endpoint no longer checks for errors in the preparation's query. Preparations with errors in the query are accepted but the preparation never starts. The preparation status isn't updated to reflect the failure either.

Will open a CH ticket to report the OOME in master; but in short, joins with sources larger than 2 megabytes makes quasar crash. Even when the result set is ~50 megabytes.

djspiewak:
@jsantos17 @alissapajer Something is wrong elsewhere then. I just added a unit test which verifies that errors in evaluation are correctly reported by `PreparationsManager`.

jsantos17:
@djspiewak just tested again with a clean database and your latest changes up to `0b4bb4e` and I'm still seeing the same problem. Here are the steps I'm using to reproduce this bug against `slamdata-backend`:

---

Create datasource

```
curl -X POST \
  http://localhost:8080/datasource \
  -H 'Content-Type: application/vnd.slamdata.datasource; version="1"' -vvv \
  -d '{
	"name": "lsource",
	"type": {
	 "name": "local",
	 "version": 1
	},
	"config": {
		"rootDir": "/Users/juanpablo/tmp/lsource",
		"readChunkSizeBytes": "1000000"
	}
}'
```

Responds with:

```
< HTTP/1.1 201 Created
< Location: /datasource/74dc3012-51f3-4ca1-8426-c91ac6ce8f71
< Date: Thu, 11 Oct 2018 22:18:41 GMT
```

Create table:

```
curl -X POST \
  'http://localhost:8080/table' \
  -H 'Content-Type: application/vnd.slamdata.table; version="1"' -vvv \
  -d '{
  "name": "with syntax error",
  "query": "elect t1.a, t2.b, t1.nr as nr1, t2.nr as nr2 FROM `/datasource/74dc3012-51f3-4ca1-8426-c91ac6ce8f71/selfjoin.data` as t1 INNER JOIN `/datasource/74dc3012-51f3-4ca1-8426-c91ac6ce8f71/selfjoin-small.data` as t2 ON t1.a = t2.b",
  "description": {}
}'
```


Responds with:

```
< HTTP/1.1 201 Created
< Location: /table/ea6e3f65-cdd3-42b4-9b6b-7dc11e2bfd03
< Date: Thu, 11 Oct 2018 22:21:12 GMT
< Content-Length: 0
```

Request preparation:

```
curl -X PUT \
  http://localhost:8080/table/ea6e3f65-cdd3-42b4-9b6b-7dc11e2bfd03/preparation \
  -H 'Content-Type: application/vnd.slamdata.preparation; version=1' -vvv \
  -d '{
  "notificationRecipient": "juan@slamdata.com",
  "frontendRoot": "http://localhost:8080/slamdata/#",
  "backendRoot": "http://localhost:8080/"
}
```

Responds with:
(should fail here)

```
< HTTP/1.1 202 Accepted
< Date: Thu, 11 Oct 2018 22:24:06 GMT
< Content-Length: 0
```

Check preparation status:

```
curl localhost:8080/table/ea6e3f65-cdd3-42b4-9b6b-7dc11e2bfd03
```

Responds with:

```
< HTTP/1.1 200 OK
< Content-Type: application/vnd.slamdata.table; version="1"
< Date: Thu, 11 Oct 2018 22:26:57 GMT
< Content-Length: 344
{
  "name": "with syntax error",
  "status": {
    "activity": "not-preparing",
    "preparation": "unprepared"
  },
  "description": {},
  "query": "elect t1.a, t2.b, t1.nr as nr1, t2.nr as nr2 FROM `/datasource/74dc3012-51f3-4ca1-8426-c91ac6ce8f71/selfjoin.data` as t1 INNER JOIN `/datasource/74dc3012-51f3-4ca1-8426-c91ac6ce8f71/selfjoin-small.data` as t2 ON t1.a = t2.b"
}
```

`PUT`ing `localhost:8080/table/ea6e3f65-cdd3-42b4-9b6b-7dc11e2bfd03/preparation` always responds with `200` and the table status does not change.

djspiewak:
@jsantos17 Do you get an email telling you the preparation failed?

jsantos17:
@djspiewak I didn’t configure emails for this test. I can try that too.

djspiewak:
I think you'll probably get an email. Effectively what this does is shift errors in compilation and planning from the initial `POST` to the async response, which is currently just emails.

jsantos17:
@djspiewak just tried that and the server died with a [template-related](https://gist.github.com/jsantos17/920afa529f98eb97ba125a79ade746d7) exception. Tried again in `master` and it worked fine.

>Effectively what this does is shift errors in compilation and planning from the initial POST to the async response, which is currently just emails.

We still need to show the correct status in `/table/<uuid>`, right? I understand the UI polls that endpoint to show the status of preparations.

djspiewak:
@jsantos17 Ticketed:

- https://app.clubhouse.io/data/story/2494/preparation-errors-cause-email-template-explosion
- https://app.clubhouse.io/data/story/2495/defaulttables-needs-to-maintain-error-state-on-async-preparation-failure

I think we should land this PR anyway, despite the fact that there are these other issues. These issues are technically pre-existing.

The real question is: does this PR fix the *freezing* on the POST on large joins? Not the OOM, but just the hang.

jsantos17:
>The real question is: does this PR fix the freezing on the POST on large joins? Not the OOM, but just the hang.

@djspiewak yes, it won't hang anymore. Let's land this and address the two tickets separately.

PR #3864, opened 2018-10-10
Delete RenderQScriptDSL

alissapajer:
I was going to fix (or delete) `RenderQScriptDSLSpec` but then figured out that `RenderQScriptDSL` is no longer in use (and hasn't been since we deleted mongo and the old file system impl).


alissapajer:
And fwiw, `RenderQScriptDSLSpec` has been silently not running for at least a few days, if not longer.

PR #3865, opened 2018-10-10
Optimize provenance computation for Map

wemrysi:
[ch1712]

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
+++ b/qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
@@ -50,8 +50,9 @@ import matryoshka._
 import matryoshka.data.free._
 import matryoshka.implicits._
 import monocle.macros.Lenses
+import monocle.Traversal
 import pathy.Path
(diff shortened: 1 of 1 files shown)


wemrysi:
There aren't any new tests as this is an optimization and should return the same results.

However, it did reveal an issue with the non-optimized version, I've created a separate story to track that work so as not to hold up this fix.

qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
```diff
@@ -281,7 +284,14 @@ final class ApplyProvenance[T[_[_]]: BirecursiveT: EqualT: ShowT] private () ext
 
       case Map(src, fm) =>
         compute1[F](g, src) { sdims =>
-          computeFuncDims(fm.linearize)(κ(sdims)) getOrElse sdims
+          val sigil = dims.prov.fresh()
+          val fdims = computeFuncDims(fm.linearize)(κ(Dimensions.origin(sigil)))
```
djspiewak: You literally just encoded skolemization. Nice!

PR #3869, opened 2018-10-13
Eliminate phantom empty columns in result sets

djspiewak:
The bug was being caused by a column which was part of the intermediate evaluation state, but subsequently got entirely "nulled out" and every value set to `undefined`.  This isn't really a problem for a JSON output, but CSV is a bit pickier. The solution is two-fold. First, we `materialize` every slice upon completion of evaluation. This is probably a good idea anyway. Or at least, I'm asserting without benchmarks that it is. Second, I modified `materialized` to remove any entirely empty columns, since we can do so efficiently at that point.

[ch2383]

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
@@ -977,125 +977,188 @@ abstract class Slice { source =>
     */
   def materialized: Slice = {
     val size = source.size
-    val columns = source.columns lazyMapValues {
-      case col: BoolColumn =>
+    val columns = source.columns flatMap {
+      case (ref, col: BoolColumn) =>
         val defined = col.definedAt(0, source.size)
-        val values = BitSetUtil.filteredRange(0, source.size) { row =>
-          defined(row) && col(row)
+        if (defined.nonEmpty) {
+          val values = BitSetUtil.filteredRange(0, source.size) { row =>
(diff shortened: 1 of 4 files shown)


djspiewak:
This is now published as **76.0.4-da24db7**

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -977,125 +977,188 @@ abstract class Slice { source =>
...
-        Loop.range(0, source.size) { row =>
-          if (defined(row)) values(row) = col(row)
+        if (defined.nonEmpty) {
+          val values  = new Array[OffsetDateTime](source.size)
+          Loop.range(0, source.size) { row =>
+            if (defined(row)) values(row) = col(row)
+          }
+          Some(ref -> ArrayOffsetDateTimeColumn(defined, values))
+        } else {
+          None
```
alissapajer: so much boilerplate...

alissapajer: no way around it without boxing, just noting it...

djspiewak: Yeah, I created most of the above by abusing the heck out of multi-cursor edits.

alissapajer: :smile:

PR #3872, opened 2018-10-15
Rename shifting things into interpreter things

alissapajer:


alissapajer:
Most of this is all changed again on my `RValue` interpreter branch, but it will be easier to get this commit into master now than carry it around locally.

PR #3873, opened 2018-10-15
Tectonic support and tests for every ParseInstruction

djspiewak:
As it says on the tin. Also includes some minor refactoring and renaming.

[ch2585]
[ch2487]


alissapajer:
@djspiewak after merging this I think to ask: why was this `breaking` and not `feature`?

djspiewak:
@alissapajer Because I changed the signatures of `ParseInstruction`. Specifically, `Ids` and `Mask` (nee `Masks`).

PR #3874, opened 2018-10-16
upgrade to fs2 1.0, cats-effect 1.0 and jawn(fs2) 0.13

rintcius:
[ch2446]


yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
```diff
@@ -67,4 +77,26 @@ package object vfs {
       def flatMap[A, B](fa: Free[S, A])(f: A => Free[S, B]): Free[S, B] =
         fa.flatMap(f)
     }
+
+  implicit def contextShiftForS[S[a] <: ACopK[a]](
```
rintcius: not really sure about this implementation

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
```diff
@@ -161,15 +161,6 @@ object FreeVFSSpecs extends Specification with DisjunctionMatchers {
           case CPR(io) => io
         }.replicateM(4)
 
-        _ <- H.pattern[FreeVFS.VFSVersion] {
```
rintcius: I'm puzzled why this needs to be removed (haven't been able to find the change in prod code that explains this)

djspiewak:
[ch2502]

alissapajer:
Needs merge from master

jsantos17:
@djspiewak I started work on updating `quasar-s3` today. I expect to open a PR with those changes tomorrow. Maybe we can wait until then to merge this?

djspiewak:
We should also update sdbe. And get its PR up.

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
```diff
@@ -215,17 +216,17 @@ final class Sql2QueryRegressionSpec extends Qspec {
   /** Returns all the `RegressionTest`s found in the given directory, keyed by
     * file path.
     */
-  def regressionTests[F[_]: Effect: Timer](testDir: JPath)
+  def regressionTests[F[_]: ContextShift: Effect: Timer](testDir: JPath, ec: ExecutionContext)
```
wemrysi: Label as blocking pool.

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
```diff
@@ -109,11 +112,11 @@ final class Sql2QueryRegressionSpec extends Qspec {
...
   val buildSuite =
     for {
-      tdef <- Promise.empty[IO, (Quasar[IO], UUID)]
-      sdown <- Promise.empty[IO, Unit]
+      tdef <- Deferred[IO, (Quasar[IO], UUID)]
+      sdown <- Deferred[IO, Unit]
 
       testsDir <- IO(TestsRoot(Paths.get("")).toAbsolutePath)
-      tests <- regressionTests[IO](testsDir)
+      tests <- regressionTests[IO](testsDir, global)
```
wemrysi: We should create a pool for the blocking reads and not use the `global` one for them.

yggdrasil/src/test/scala/quasar/yggdrasil/vfs/FreeVFSSpecs.scala
wemrysi: If I'm remembering correctly, these tests are fairly brittle as they are essentially asserting about sequences of monadic binds, so if an implementation somewhere changed how many binds it used, it could affect this I think.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/vfs.scala
```diff
@@ -614,22 +606,20 @@ object SerialVFS {
    * TODO: Return a `Stream[F, SerialVFS[F]]` to allow for better resource
    *       handling once we're rid of `BackendModule`.
    */
-  def apply[F[_]: ConcurrentEffect](root: File, pool: ExecutionContext)
+  def apply[F[_]: Concurrent](root: File, pool: ExecutionContext)(implicit cs: ContextShift[POSIXWithIO])
```
wemrysi: Rename `pool` to indicate blocking.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
```diff
@@ -19,25 +19,23 @@ package quasar.yggdrasil.vfs
...
 import java.io.{File, IOException}
 import java.nio.file.{Files, StandardCopyOption}
 import java.util.UUID
+import scala.concurrent.ExecutionContext
 
 object RealPOSIX {
   import POSIXOp._
 
-  def apply[F[_]](root: File)(implicit F: Sync[F]): F[POSIXOp ~> F] = {
+  def apply[F[_]](root: File, ec: ExecutionContext)(implicit F: Sync[F], cs: ContextShift[POSIXWithIO]): F[POSIXOp ~> F] = {
```
wemrysi: Rename to indicate blocking.

run/src/main/scala/quasar/run/Quasar.scala
```diff
@@ -82,15 +81,15 @@ object Quasar {
...
+  def apply[F[_]: ConcurrentEffect: ContextShift: MonadQuasarErr: PhaseResultTell: Timer](
       precog: Precog,
       extConfig: ExternalConfig,
       sstEvalConfig: SstEvalConfig)(
       implicit ec: ExecutionContext)
       : Stream[F, Quasar[F]] = {
 
     for {
-      extMods <- ExternalDatasources[F](extConfig)
+      extMods <- ExternalDatasources[F](extConfig, ec)
```
wemrysi: This shouldn't use the compute pool, we'll either need to accept a blocking pool or create one ourselves.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
```diff
@@ -67,4 +77,26 @@ package object vfs {
...
+    implicit
+    ec: ExecutionContext,
+    I: IO :<<: S)
+      : ContextShift[Free[S, ?]] =
+    new ContextShift[Free[S, ?]] {
+      def shift: Free[S, Unit] = Free.liftF(I(IO.contextShift(ec).shift))
+
+      def evalOn[A](ctx: ExecutionContext)(fa: Free[S, A]): Free[S, A] =
```
wemrysi: Here, I think we want to `fa.mapSuspension` and change any `IO` suspensions to `evalOn(ctx)` using the `ContextShift[IO]`.

rintcius: Hmm, wouldn't that result in delegating too little to be run on `ctx` (i.e.. all non-`IO`s are simply eval-ed as-is)?

BTW Maybe I'm missing something, but I would also have thought that the interface would be something like def `evalOn[A](ec: ExecutionContext)(fa: => F[A]): F[A]` instead

rintcius:
released another snapshot `77.0.5-77dc597`

cc @jsantos17

wemrysi:
@rintcius A separate type for the blocking ones is a great idea. In the past, I've used a `scalaz.Tag` for this purpose, but a value class would also be fine.

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
rintcius: ok thanks for the explanation @wemrysi - I have adapted the implementation

yggdrasil/src/main/scala/quasar/yggdrasil/vfs/package.scala
```diff
@@ -67,4 +76,18 @@ package object vfs {
       def flatMap[A, B](fa: Free[S, A])(f: A => Free[S, B]): Free[S, B] =
         fa.flatMap(f)
     }
+
+  implicit def contextShiftForS[S[a] <: ACopK[a]](
+    implicit
+    ec: ExecutionContext,
```
wemrysi: Sorry for not noticing this before, but I think it'd be better to require an implicit `ContextSwitch[IO]` than to construct it from the implicit `ExecutionContext`.

rintcius: oh, yeah agreed

djspiewak:
I never thought about tagging the blocking pools. Outstanding idea.

rintcius:
Note that GH is not in a healthy state at the moment and that travis is not triggered. Number of commits  in this PR should be 13. If it shows less then try refreshing..

Anyway, published new snapshot 78.1.1-f21a31d  

cc @jsantos17

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
```diff
@@ -22,22 +22,28 @@ import quasar.api.resource.ResourcePath
...
+import cats.effect._
 import fs2.Stream
-import scalaz.{\/, EitherT}
+import scalaz.{EitherT, \/}
 import scalaz.syntax.applicative._
 import shims._
 
 object LocalDatasourceModule extends LightweightDatasourceModule {
 
+  // FIXME this is side effecting
```
rintcius: not ideal but I don't think it's that harmful either (I guess there's only negative impact if this class gets reloaded). 

I tried get this working with an `F.delay(...)` and proper dispose of the `ExecutorService` but gave up because it took too much time.

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
```diff
@@ -22,22 +22,28 @@ import quasar.api.resource.ResourcePath
...
+  val blockingPool: ExecutionContext = ExecutionContext.fromExecutor(Executors.newCachedThreadPool)
+
   val kind: DatasourceType = LocalType
 
   def sanitizeConfig(config: Json): Json = config
 
-  def lightweightDatasource[F[_]: ConcurrentEffect: MonadResourceErr: Timer](config: Json)
+  def lightweightDatasource[F[_]: ConcurrentEffect: ContextShift: MonadResourceErr: Timer](
```
rintcius: tried to get rid of this `ContextShift` but couldn't because I am not able to create a `val cs: ContextShift[F] = ...`

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
wemrysi: How about making it a `private lazy val`? At least then it won't be created until it is used.

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
wemrysi: We actually need the `ContextShift` context bound as it is so computations can be shifted _back_ onto the "default" context for `F`, which will likely be the main compute pool. So this is fine :+1:

wemrysi: Oh, I just noticed the implicit EC, I see what you mean now. Yeah, it's a bit redundant, but ideally we'll get rid of the implicit EC eventually.

rintcius:
@wemrysi PR has been updated

PR #3880, opened 2018-10-16
Add some more Mask tests

alissapajer:


frontend/src/test/scala/quasar/ParseInstructionSpec.scala
```diff
@@ -481,6 +481,66 @@ object ParseInstructionSpec {
...
+        val input = ldjson("""
+          { "a": { "b": 42, "c": true } }
+          """)
+
+        val expected = ldjson("""
+          { "a": { "c": true } }
+          """)
+
+        input must maskInto(".a.c" -> Set(Boolean), ".a.c" -> Set(Number))(expected)
+      }.pendingUntilFixed
```
alissapajer: @djspiewak this one isn't working! Compare to the one below.

alissapajer: Made ch2766

alissapajer: Reminder to self: I'd like to add some triple disjunction tests as well.

alissapajer: Also maybe I misused `maskInto` and should have had a single `Set(Boolean, Number)`? And it's not expected to work like this?

alissapajer: Ok, yeah it's clear that this shouldn't work because we have a `Map[CPath, ...]`.

alissapajer: I'll fix the test.

alissapajer: This is covered by `"retain two scalar types at .a.b"`, so deleting.

frontend/src/test/scala/quasar/ParseInstructionSpec.scala
```diff
@@ -481,6 +481,42 @@ object ParseInstructionSpec {
...
+      "compose disjunctively across paths where one side is false (left) (vector)" in {
+        val input = ldjson("""
+          { "a": { "b": 42, "c": true } }
+          """)
+
+        val expected = ldjson("""
+          { "a": { "c": true } }
+          """)
+
+        input must maskInto(".a" -> Set(Array), ".a.c" -> Set(Boolean))(expected)
```
alissapajer: Oh and it's silly to have both these left and right versions :) fixing...

alissapajer: The methodical portion of my brain was on overdrive and now the logical portion is kicking in... :)

PR #3881, opened 2018-10-16
Removed flaky assertion (there's no way to do it reliably)

djspiewak:
[ch2754]


alissapajer:
Hm, yeah, I don't see a way to test this reliably either. Merging.

PR #3885, opened 2018-10-17
Plate interpretation in Table.parseJson

djspiewak:
Had to move the plates out of mimir and into yggdrasil in order to keep `parseJson`'s implementation confined to that package. Ties the knot on everything mimir-related though.

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
@@ -17,6 +17,7 @@
 package quasar.yggdrasil
 package table
 
+import quasar.ParseInstruction
 import quasar.blueeyes._
 import quasar.common.{CPath, CPathArray, CPathField, CPathIndex, CPathMeta, CPathNode}
 import quasar.contrib.fs2.convert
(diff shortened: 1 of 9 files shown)


alissapajer:
@djspiewak it failed to compile.

djspiewak:
Details…

yggdrasil/src/main/scala/quasar/yggdrasil/table/ColumnarTableModule.scala
```diff
@@ -370,14 +371,30 @@ trait ColumnarTableModule
...
         precise: Boolean = false)(
         implicit ec: ExecutionContext)
         : M[Table] = {
 
       import fs2.{Chunk, Stream}
+      import tectonic.Plate
       import tectonic.json.Parser
 
-      val parser = Parser(new SlicePlate(precise), Parser.ValueStream)
+      val plate = instructions.foldRight(new SlicePlate(precise): Plate[List[Slice]]) {
```
alissapajer: @djspiewak realized while writing my impl (which is basically identical to this :) ) that you'll have to reverse the `List[Instruction]` if you want to `foldRight`. Or we could generate the instructions in the reverse order. Either way.

djspiewak: Actually I'm pretty sure `foldRight` without `reverse` is what I want. A chain of delegating plates is basically the same as a linked list. The first plate which applies to the input stream will be the one which corresponds to the first element of the list.

alissapajer: We're currently generating instructions in this order:
```
List(Mask, Wrap, Pivot)
```
Does this implementation evaluate them in the correct order (`Mask` first, followed by `Wrap`, then `Pivot`)?

djspiewak: It does. :-)

alissapajer: :+1: then we be good :)

PR #3888, opened 2018-10-19
Allow edits of prepared and preparing tables

alissapajer:


alissapajer:
[ch2211]

What's needed on quasar for ch2795.

api/src/test/scala/quasar/api/table/TablesSpec.scala
```diff
@@ -197,6 +197,44 @@ abstract class TablesSpec[F[_]: Monad: Sync, I: Equal: Show, Q: Equal: Show, D,
...
+      } yield {
+        tableResult must beLike {
+          case \/-(t) => t must_= table1
+        }
+
+        prepareResult must beNormal
+      }
+    }
+
+    "successfully replace a prepared table" >>* {
```
alissapajer: We don't have a way to talk about an actively preparing table in `TablesSpec` here, so we only test that we can replace the table when it's `Prepared` but not `Preparing`. 

It looks like I'll be able to add a test in SDBE to check that table replacement works when the status is `Preparing`.

alissapajer: Also I verified this test fails in master :)

alissapajer:
@djspiewak hold off on merge on this this, there might be reason to make it `breaking`, not 100% sure yet.

alissapajer:
Ok we're good to go with `feature`. The exception I was getting in the SDBE test was because the test assumed the response had a body, which it didn't have anymore.

api/src/test/scala/quasar/api/table/TablesSpec.scala
alissapajer: SDBE actually had tests for both cases already! Both tests failed, fixed 'em up.

PR #3890, opened 2018-10-20
Live access

alissapajer:
The quasar side of ch2760

### DIFF
--- a/api/src/test/scala/quasar/api/table/MockTables.scala
+++ b/api/src/test/scala/quasar/api/table/MockTables.scala
@@ -39,8 +39,8 @@ final class MockTables[F[_]: Monad: MockTables.TablesMockState]
   val store = MonadState_[F, IMap[UUID, MockTable]]
 
   def allTables: Stream[F, (UUID, TableRef[String], PreparationStatus)] =
-    Stream.force(store.get.map { s =>
-      Stream.emits(s.toList.map {
+    Stream.force(store.get map { s =>
+      Stream.emits(s.toList map {
         case (uuid, MockTable(table, status)) => (uuid, table, status)
       }).covary[F]
     })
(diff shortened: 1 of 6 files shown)


api/src/test/scala/quasar/api/table/MockTables.scala
```diff
@@ -103,10 +103,11 @@ final class MockTables[F[_]: Monad: MockTables.TablesMockState]
   def cancelPreparation(tableId: UUID): F[Condition[PreparationNotInProgress[UUID]]] =
     store.get.flatMap(stateMap => stateMap.lookup(tableId) match {
       case Some(MockTable(table, PreparationStatus(prep, OngoingStatus.Preparing))) =>
-        store.put {
+        val inserted = store put {
           stateMap.insert(tableId,
             MockTable(table, PreparationStatus(prep, OngoingStatus.NotPreparing)))
-        }.as(Condition.normal())
+        }
+        inserted.as(Condition.normal())
```
djspiewak: 👍

api/src/test/scala/quasar/api/table/MockTables.scala
```diff
@@ -116,31 +117,48 @@ final class MockTables[F[_]: Monad: MockTables.TablesMockState]
...
+    val cancelled = store.get flatMap { stateMap =>
+      val updatedMap = stateMap map {
         case MockTable(ref, PreparationStatus(prep, _)) =>
           MockTable(ref, PreparationStatus(prep, OngoingStatus.NotPreparing))
       }
       store.put(updatedMap)
-    } >> ().point[F]
+    }
+
+    cancelled >> ().point[F]
```
djspiewak: ⛳️ 

```scala
canceled.void
```

impl/src/main/scala/quasar/impl/table/DefaultTables.scala
```diff
@@ -114,10 +116,15 @@ final class DefaultTables[F[_]: Effect, I: Equal, Q, D, S](
   def preparationEvents: Stream[F, PreparationEvent[I]] =
     manager.notifications
 
+  def liveData(tableId: I): F[ExistenceError[I] \/ D] =
+    table(tableId) flatMap {
+      _.traverse(ref => evaluator.evaluate(ref.query))
+    }
```
djspiewak: This is very correct. It's worth noting that if the query is non-linear (i.e. involves branching nodes that are not `Union`), then this effect sequencing will involve evaluating part of the query (to eagerly cache the hole) and thus will not be "live". @beckyconning is aware of this, and it's valid for us to be doing it without checking, I just wanted to underscore the limitation.

PR #3891, opened 2018-10-22
Fix commitdb persistence bug

nicflores:
Fix persistence of indexed store on server restart.

[ch2407]


mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
```diff
@@ -78,8 +78,10 @@ final class MimirPTableStore[F[_]: Monad: LiftIO] private (
       }, {
         case Some((blob, version, db)) =>
           completed.get.flatMap { c =>
-            if (c)
+            if (c) {
               cake.commitDB(blob, version, db)
+              ().pure[IO]
```
djspiewak: This doesn't do what you think it's doing! :-D

I believe you mean: `cake.commitDB(blob, version, db).void`. As it stands, you're throwing away the `commitDB` effect without evaluating it at all. Also the formatting is wrong and you shouldn't have these curly braces here.

alissapajer: Why did this line need to be changed at all?

alissapajer:
We should have tests for the relevant changes. We have `VFSColumnarTableModuleSpec` and `MimirPTableStoreSpec`.

mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala
nicflores: @alissapajer, the change is due to `commitDB` now returning `IO[Boolean]`, but you bring up a good point. @djspiewak maybe I should make `commitDB` return what what was returning before, `IO[Unit]`?

alissapajer:
What kind of integration test did you have in mind?

nicflores:
@alissapajer, something along the lines of a test like this: 
a. start the server
b. make some changes to a store (like add some tokens or rename a table renaming
c. restart the server and make sure the changes persisted

nicflores:
I'm putting this test for to get some feedback. The original fix of `commitDB` was to make sure the `HEAD` pointer was being advanced and that upon restart of the server we picked up the most current state of the previous indexed store. In the test above I execute two `commitDB`s, manually clear the hot cache (the ConcurrentHashMap), and finally assert that the two commits succeeded and that the head pointer is correct. One thing that would be useful is to compare two `nihbd` instances, is there some way of doing this? If so I could use `openDB(path)` and get the `nihdb` object out of its return value and compare it do `db2`.

alissapajer:
Interesting, travis passed, but appveyor failed with what looks legit:
```
[error]   x write and then read back results
[error]    'List()' doesn't have size 1 but size 0 (MimirPTableStoreSpec.scala:84)
[error]   x write repeatedly and then read back last results
[error]    'List()' doesn't have size 1 but size 0 (MimirPTableStoreSpec.scala:109)

[error] Failed: Total 347, Failed 2, Errors 0, Passed 345, Pending 1
[error] Failed tests:
[error] 	quasar.mimir.storage.MimirPTableStoreSpec
```

yggdrasil/src/test/scala/quasar/yggdrasil/table/VFSColumnarTableModuleSpec.scala
```diff
@@ -17,26 +17,80 @@
...
+
+    "check db commits persistence head after clearing hotcache" in {
+
+
+      val path: AFile =  rootDir </> file("foo")
+
+      val test = for {
+        created <- createDB(path) map {
+          case \/-(t) => t
+          case -\/(_) => true must_== false
```
alissapajer: If you want this to fail the test you can use `ko`.

alissapajer: Another option is [this](https://github.com/slamdata/quasar/blob/master/api/src/test/scala/quasar/api/table/TablesSpec.scala#L276). I'm not sure how specs2 results are handled in-line like this; it might result in weird failure messages.

djspiewak: The errors should be fine. I do think that a better approach is something like `isSuccess`, with an added `must be_\/-` assertion.

yggdrasil/src/test/scala/quasar/yggdrasil/table/VFSColumnarTableModuleSpec.scala
```diff
@@ -17,26 +17,80 @@
...
     with nihdb.NIHDBAkkaSetup {
 
   implicit val ExecutionContext = scala.concurrent.ExecutionContext.global
 
+
   "table tempfile caching" should {
+
+    "check db commits persistence head after clearing hotcache" in {
+
+
```
djspiewak: Newlines!

yggdrasil/src/test/scala/quasar/yggdrasil/table/VFSColumnarTableModuleSpec.scala
```diff
@@ -17,26 +17,80 @@
...
+        commit1 must_=== true
+        commit2 must_=== true
+        actualHead must_=== version2
+      }
+
+      test.unsafeRunSync
+    }
+
+
+
```
djspiewak: MOAR NEWLINES

yggdrasil/src/test/scala/quasar/yggdrasil/table/VFSColumnarTableModuleSpec.scala
```diff
@@ -114,7 +168,8 @@ object VFSColumnarTableModuleSpec
     }
   }
 
-  def vfs = ???
+  val base = Files.createTempDirectory("VFSColumnarTableModuleSpec").toFile
+  def vfs = SerialVFS[IO](base, global).unsafeRunSync.unsafeValue
```
djspiewak:
```scala
val (vfs, vfsCleanup) = {
  val d = SerialVFS[IO](base, blockingEC).unsafeRunSync()
  (d.unsafeValue, d.dispose)
}
```

And then at the very *end* of the file, you should have something like:

```scala
step(vfsCleanup.unsafeRunSync())
```

PR #3895, opened 2018-10-23
Implement InterpretedRead rewrites through static array derefs

alissapajer:
[ch2861]


alissapajer:
This also hits `tests/range.test`, which is cool:
```sql
select (topArr[1].botArr[*] .. 9) from `nested.data`
```

PR #3898, opened 2018-10-23
CPath spec

alissapajer:
[ch2354]

Now that we're using `CPath` in some new places, figured we should have some unit tests for it.


common/src/main/scala/quasar/common/CPath.scala
```diff
@@ -72,14 +61,10 @@ sealed trait CPath { self =>
...
   def head: Option[CPathNode] = nodes.headOption
-
   def tail: CPath = CPath(nodes.drop(1): _*)
 
-  def path: String = nodes.mkString("")
-
-  def length: Int = nodes.length
-
-  override def toString = if (nodes.isEmpty) "." else path
+  // TODO we'd like to remove this but a few tests in niflheim depend on it
```
djspiewak: Why do you want to remove it?

common/src/test/scala/quasar/common/CPathSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+package quasar.common
+
+import quasar.Qspec
+
+import scalaz.std.option._
+import scalaz.syntax.std.option._
+
+object CPathSpec extends Qspec {
+
+  "CPath parser" >> {
```
djspiewak: `should`

common/src/test/scala/quasar/common/CPathSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+
+    "parse empty path" >> {
```
djspiewak: `in`

common/src/test/scala/quasar/common/CPathSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+      CPath.parse(".") must_=(
```
djspiewak: `mustEqual` will be much more concise (doesn't require the parens, for example).

common/src/test/scala/quasar/common/CPathSpec.scala
alissapajer: `in` has to be at the innermost level, surrounding a matcher, for it to compile. So like
```
"a" should {
  "b" >> {
    "c" in {
      ko
    }
  }
}

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 2 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: I have been unable to reproduce differences in behavior between `>>` and `should/in` when using `ex` (or in any other situations). Happy to change if we can come up with an example that exhibits different behavior.

File: common/src/main/scala/quasar/common/CPath.scala (hunk 1 above)
Comment (alissapajer) [reply common/src/main/scala/quasar/common/CPath.scala]: Because I'd prefer to use the `Show` instances.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 3 above)
Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: It is at the innermost level here though.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 2 above)
Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: I'll keep an eye out for it. I've reproduced the differences as recently as a few months ago, and it was definitely with `ex`, but I can't remember what test it was on. Maybe `ColumnarTableModuleSpec`? Either way.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 4 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Where is `mustEqual` defined?

Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Specs2. It's literally the default baseline matcher that you want most of the time.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 2 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Also I've only ever seen `ex` working at the innermost nesting level of nesting. If we can make it match on the outer layers, that would be great. But using `should/in` does not make that happen locally.

Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Sometimes `ex` doesn't work even at the innermost level, which was part of what I was experiencing with `>>` a few months ago.

File: common/src/main/scala/quasar/common/CPath.scala (hunk 1 above)
Comment (djspiewak) [reply common/src/main/scala/quasar/common/CPath.scala]: That's fair, but relying exclusively on `Show` results in extremely poor integration with things that are hard-coded to `toString`, such as assertion failures in tests, exceptions, and such. I think for anything that's going to hit `toString`-reliant code like that, we should just have both `Show` and `toString` (similar to how data structures in scalaz and cats generally have both `Equal` and `def equals` implementations).

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 3 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Oh yeah, sorry, at lot of the examples are more than two deep so I was confused about which example your comment was on :)

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 4 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: `must_=` matches using `scalaz.Equal`, which is why I chose it.

Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Ahhh. I didn't realize that. Does it require the parentheses?

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 2 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: It's working at the innermost level, at least in `CPathSpec`. I prefer `>>` at all levels because it makes examples easier to write and to copy-paste around while I'm writing the tests. But if `ex` still doesn't always work with `>>` then I can change it.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 3 above)
Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Yeah definitely on any example which is more than two deep, `>>` is the right way to go on the inner levels. I still prefer `should` at the outer level because it hints at a more uniform convention for naming the labels, but that's a more aesthetic thing.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 4 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: [Here's](https://github.com/slamdata/quasar/blob/master/foundation/src/test/scala/quasar/QuasarSpecification.scala#L57) the def for `must_=`.

Tried it out and parens are not required! I shall remove them :)

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 2 above)
Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: I do like that it's easier to copy/paste things around but that's not usually a problem I run into more than once or twice. I guess this is a more subjective aesthetic thing if `ex` is working on `CPathSpec`. Definitely not going to make a huge fuss over it if it lands as-is.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 4 above)
Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: I wish that definition had been `must_===`, or `mustEquiv` instead. The `_=` suffix has special semantics in Scala.

Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: `must_===` is already taken, not sure where it's imported from though or what equality it uses...

Comment (djspiewak) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: Specs2 defines it. I believe it uses intrinsic equality. If `Qspec` redefined it though, it should take precedence over Specs2's definition.

File: common/src/test/scala/quasar/common/CPathSpec.scala (hunk 3 above)
Comment (alissapajer) [reply common/src/test/scala/quasar/common/CPathSpec.scala]: This is the way I'm using `ex`, btw
```
sbt:root> common/testOnly *CPath* -- -ex prefix is identity --
```

common/src/main/scala/quasar/common/CPath.scala
alissapajer: Fwiw, using `toString` in any context is inconsistent with all components of quasar that didn't get inherited from precog.

The tests using `CPath` would have to change to match using `must_=`, which requires an `scalaz.Equal` and a `scalaz.Show`. I just verified that test errors from a `must_=` are printed using the `Show` instance.

For debugging, you can do `foo.shows`, so it's not much extra overhead imo.

Using `Show` gives you explicit control over how things are printed, which is why I prefer it. Keeping `toString` here isn't horrible though, so I'd also be okay removing the `TODO`.

djspiewak: Calling shows isn’t the overhead. Threading Show everywhere just so we can print *is*, as the qscript project shows quite well. Also you can’t always rely on Show when nested inside another structure which uses intrinsic toString (such as Map or List).

Ultimately, yes, Show is a better solution, but it needs to complement toString on implementations, not replace it. Just like equals. Call sites should always prefer Show if they can, but we can’t control everyone’s call site.

PR #3899, opened 2018-10-23
Run Sql2QueryRegressionSpec against optimized and unoptimized qscript

alissapajer:
[ch2918]

### DIFF
--- a/it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
+++ b/it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
@@ -58,8 +58,10 @@ import scalaz._, Scalaz._
 // import shims._ causes compilation to not terminate in any reasonable amount of time.
 import shims.{monadErrorToScalaz, monadToScalaz}
 
-final class Sql2QueryRegressionSpec extends Qspec {
+abstract class Sql2QueryRegressionSpec extends Qspec {
   import Sql2QueryRegressionSpec._
+
+  val optimizeQScript: Boolean
 
   implicit val ignorePhaseResults: MonadTell_[IO, PhaseResults] =
     MonadTell_.ignore[IO, PhaseResults]
@@ -76,7 +78,7 @@ final class Sql2QueryRegressionSpec extends Qspec {
   /** A name to identify the suite in test output. */
   val suiteName: String = "SQL^2 Regression Queries"
 
-  def Q(testsDir: JPath) = for {
+  def RunQuasar(testsDir: JPath): Stream[IO, (Quasar[IO], UUID)] = for {
     tmpPath <-
       Stream.bracket(IO(Files.createTempDirectory("quasar-test-")))(
         Stream.emit(_),
@@ -86,7 +88,7 @@ final class Sql2QueryRegressionSpec extends Qspec {
 
     evalCfg = SstEvalConfig(1L, 1L, 1L)
 
-    q <- Quasar[IO](precog, ExternalConfig.Empty, SstEvalConfig.single)
+    q <- Quasar[IO](precog, optimizeQScript, ExternalConfig.Empty, SstEvalConfig.single)
 
     localCfg =
       ("rootDir" := testsDir.toString) ->:
@@ -107,7 +109,7 @@ final class Sql2QueryRegressionSpec extends Qspec {
 
   ////
 
-  val buildSuite =
+  val buildSuite: IO[Fragment] =
     for {
       tdef <- Promise.empty[IO, (Quasar[IO], UUID)]
       sdown <- Promise.empty[IO, Unit]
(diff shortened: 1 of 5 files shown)


mimir/src/main/scala/quasar/mimir/evaluate/MimirQScriptEvaluator.scala
```diff
@@ -49,7 +49,7 @@ import shims._
 final class MimirQScriptEvaluator[
     T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT,
     F[_]: LiftIO: Monad: MonadPlannerErr: MonadFinalizers[?[_], IO]: PhaseResultTell] private (
-    cake: Cake)(
+    cake: Cake, optimizeQScript: Boolean)(
```
alissapajer: I aimed to thread this in at the highest level possible. If we like this approach, I'd want to newtype this with a two-type algebra.

repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -61,7 +61,7 @@ object Main extends StreamApp[PhaseResultCatsT[IO, ?]] {
       (dataPath, pluginPath) <- paths[F]
       precog <- Precog.stream(dataPath.toFile).translate(λ[FunctionK[IO, F]](_.to[F]))
       evalCfg = SstEvalConfig(1000L, 2L, 250L)
-      q <- Quasar[F](precog, ExternalConfig.PluginDirectory(pluginPath), evalCfg)
+      q <- Quasar[F](precog, true, ExternalConfig.PluginDirectory(pluginPath), evalCfg)
```
alissapajer: It'd be cool if we could control this via a repl command, but I couldn't see how to make that happen given that it's an argument required to create a `Quasar`.

alissapajer:
Hold off on merge, this isn't actually working...

alissapajer:
omg
```
scala> val x: Boolean = null.asInstanceOf[Boolean]
x: Boolean = false
```

djspiewak:
@alissapajer There's actually a relatively complicated reason for this. You can do something similar with all the primitives:

```scala
val x: Int = null.asInstanceOf[Int]    // => 0
```

The reason is that `null` is technically not a thing that can be assigned to a primitive. Not in Scala's type system, and not in the JVM's. Additionally, `asInstanceOf[Int]` isn't something that can be converted into bytecode. So scalac has to figure out what to do here.

The answer is that it takes the stack slot corresponding to the input value and assumes it is a boxed integer that you're trying to convert, since an unboxed integer would be typed as `Int` (and thus couldn't be `null`) and it could just drop the cast statically. Unboxing code *always* starts with a `null` check which inserts the JVM-default value if the `null` check fails (really, what else would it do?). The default value for `Int` is `0`, and the default value for `Boolean` is `false`.

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
```diff
@@ -58,9 +58,11 @@ import scalaz._, Scalaz._
 // import shims._ causes compilation to not terminate in any reasonable amount of time.
 import shims.{monadErrorToScalaz, monadToScalaz}
 
-final class Sql2QueryRegressionSpec extends Qspec {
+abstract class Sql2QueryRegressionSpec extends Qspec {
   import Sql2QueryRegressionSpec._
 
+  lazy val optimizeQScript: Boolean = null.asInstanceOf[Boolean]
```
alissapajer: Using `IO` (or anything delaying) here would be cleaner, but we have an `buildSuite.unsafeRunSync()` at the top level in the test, so I'll have to see if we can make that work.

alissapajer: No, there's no reason not to use `false`. If we stick with `Boolean` I'll change it to `false`. But I was thinking of not using `Boolean` and using a newtype, in which case I'd let this equal `null` (wouldn't need the `asInstanceof`). And I was saying the newtype would be better because if it used the default value (`null`) we'd get a `NullPointerException` rather than have the test run with the unintended value.

mimir/src/main/scala/quasar/mimir/evaluate/MimirQScriptEvaluator.scala
wemrysi: Regarding naming, I'd prefer this property (and the others like it) be more specific to enabling/disabling the use of parse instructions/parser pushdown, etc. (whatever the proper wording is) rather than optimization as the latter is a bit misleading.

I think we can achieve being able to modify this at runtime with a few modifications

1. Change the signatures of `optimize` and `toTotal` to return their functions in `F[_]`.
2. Create a config type for the `MimirQScriptEvaluator` that contains both the existing `Associates` function and a new property controlling the generation of parse instructions.
3. Replace `AssociatesT` with a `Kleisli` based on this new type.
4. Expose an effectful getter/setter on `MimirQueryFederation` to modify the property.
5. For now, also expose similar getter/setter on `Quasar` (this will go away eventually when we extract `mimir`)

This will allow us to dynamically set the property in the repl as well as run multiple instances of the test suite for both values in the same spec, if desired. The argument to `Quasar` would then serve as the initial value.

alissapajer: Using `Associates` is a good idea! Going to work that through and re-open in a new PR.

PR #3900, opened 2018-10-24
Fixed issue with emiting FinishRow on a singleton vector pivot

djspiewak:
@wemrysi found this one.

### DIFF
--- a/frontend/src/test/scala/quasar/ParseInstructionSpec.scala
+++ b/frontend/src/test/scala/quasar/ParseInstructionSpec.scala
@@ -1020,6 +1020,18 @@ object ParseInstructionSpec {
           input must pivotInto(".a[0]", IdStatus.IncludeId, ParseType.Object)(expected)
         }
       }
+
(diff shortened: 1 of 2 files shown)


frontend/src/test/scala/quasar/ParseInstructionSpec.scala
```diff
@@ -1020,6 +1020,18 @@ object ParseInstructionSpec {
           input must pivotInto(".a[0]", IdStatus.IncludeId, ParseType.Object)(expected)
         }
       }
+
+      "shift an object under .shifted with IncludeId and only one field" in {
```
alissapajer: We have tests quite similar to this already. Was it the "one field of the pivot target" component that makes this unique?

djspiewak: In `SlicePlate`, single-element pivots are completely different from zero or two-or-more element pivots due to the need to renest.

PR #3903, opened 2018-10-24
Implement runtime en/disabling of pushdown evaluation

alissapajer:
And enable pushdown as a repl option
And run the regression tests against en/disabled pushdown

[ch2918]


alissapajer:
@wemrysi I think this is similar to what you had in mind, but if you think anything could still be better let me know!

alissapajer:
@djspiewak @wemrysi this adds runtime toggling of pushdown (whether we rewrite to `InterpretedRead` or not) in the REPL:
```
set pushdown = true | false
```
I tried it out and it's pretty neat. It probably will become the quickest way to verify pushdown results against non-pushdown results.

alissapajer:
Oh and also the REPL feature provides us a first-layer of manual performance checking, comparing pushdown and non-pushdown results. And when we have the tectonic LWC we'll be able to compare against that too!

wemrysi:
@alissapajer Looks great, thanks!

alissapajer:
Having it runtime-changeable was definitely worth it! :)

alissapajer:
Getting this back up to date.

alissapajer:
@wemrysi now up to date with master. Also, I pushed one additional commit renaming `Optimize` to `RewritePushdown`. I'd like to keep the `InterpretedRead` rewrite in one place so that toggling it on/off with the new machinery will continue to make sense. If we have other rewrites we'll put them in a different rewrite step.

connector/src/main/scala/quasar/connector/QScriptEvaluator.scala
```diff
@@ -56,11 +56,11 @@ abstract class QScriptEvaluator[
...
   def execute(repr: Repr): F[R]
 
-  /** Returns the executable representation of the given optimized QScript. */
+  /** Returns the executable representation of the provided QScript. */
   def plan(cp: T[QSM]): F[Repr]
 
-  /** Rewrites the qscript to prepare for optimal evaluation. */
-  def optimize: QSMRewrite[T[QSM]] => QSM[T[QSM]]
+  /** Rewrites the qscript to use InterpretedRead whenever possible. */
+  def rewritePushdown: F[QSMRewrite[T[QSM]] => QSM[T[QSM]]]
```
wemrysi: I think we can keep this as `optimize` as it is implementation specific as to what optimizations are supported, but we can change that later.

alissapajer: Yeah, I was thinking that there might be other optimizations that we want to implement (say to do with how we know mimir evaluates) that are not specifically rewriting to `InterpretedRead`. And so we might not want to put include those in this function because then we lose the ability to just turn off the `InterpretedRead` rewrites. But maybe we'll only want to turn them all off together.

wemrysi: Disabling pushdown is a mimir-specific feature, so it could choose to skip that bit in the implementation of `optimize`.

alissapajer: Right I see your point about it being mimir-specific. Since we only have one `QScriptEvaluator` currently I was thinking really only in terms of mimir.

PR #3907, opened 2018-10-26
compact/canonicalize in execute (after evaluation)

alissapajer:
[ch2936]


alissapajer:
@djspiewak hm so the only place I've found where we're materializing before storing or returning results is in [`SegmentsWrapper.sliceToSegments`](https://github.com/slamdata/quasar/blob/master/yggdrasil/src/main/scala/quasar/yggdrasil/nihdb/SegmentsWrapper.scala#L110), which we call (for one) in [`MimirPTableStore#write`](https://github.com/slamdata/quasar/blob/master/mimir/src/main/scala/quasar/mimir/storage/storage/MimirPTableStore.scala#L82). We need to call `materialized` in `SegmentsWrapper.sliceToSegments` though because an `ArraySegment` contains an array of its actual values.

djspiewak:
Maybe a better approach then would be to make `materialized` check to see if columns are *already* materialized and leave them as-is.

Also I vaguely remember inserting a call to `materialized` somewhere in slamdata-backend.

alissapajer:
@djspiewak I checked SDBE and there are no calls to `materialized`.

djspiewak:
Hurray for faulty memories! Good to know.

PR #3911, opened 2018-10-26
Distinct type for execution contexts used for blocking operations

wemrysi:


yggdrasil/src/main/scala/quasar/yggdrasil/vfs/RealPOSIX.scala
```diff
@@ -48,7 +54,9 @@ object RealPOSIX {
           val ptarget = canonicalize(target).toPath()
 
           val stream =
-            readAll[POSIXWithIO](ptarget, blockingPool, 4096).chunks.map(c => ByteVector(c.toArray))
+            readAll[POSIXWithIO](ptarget, blockingPool.unwrap, 4096)
+              .chunks
+              .map(c => ByteVector(c.toArray))
```
djspiewak: It occurs to me that the performance of this bit is horrible. We should fix it someday.

PR #3914, opened 2018-10-27
Use already-materialized column when materializing

alissapajer:
[ch2957]


yggdrasil/src/main/scala/quasar/yggdrasil/table/Column.scala
```diff
@@ -43,8 +43,28 @@ sealed trait Column {
...
+   * Sets values within [from, to) to the definedness of this column.
+   */
   def definedAt(from: Int, to: Int): BitSet =
-    BitSetUtil.filteredRange(from, to)(isDefinedAt)
+    this match {
+      case col: BitsetColumn =>
+        val bs = col.definedAt.copy()
+        val max = bs.length << 6
+
+        if (from == 0 && to > max) {
```
djspiewak: I think you want `>=` here.

yggdrasil/src/main/scala/quasar/yggdrasil/table/Column.scala
```diff
@@ -604,6 +624,7 @@ trait ArrayColumn[@specialized(Boolean, Long, Double) A] extends BitsetColumn wi
   def update(row: Int, value: A): this.type
   def clear(row: Int): Unit
   def resize(size: Int): ArrayColumn[A]
+  def size: Int
```
djspiewak: I think this is deceptive, since you're defining it to be `defined.size` in some places, which is not necessarily even the size of the array. Under certain circumstances, you could have something like:

```scala
val col: ArrayLongColumn = ...
col.values(col.size - 1)        // => IndexOutOfBoundsException
```

And under other circumstances:

```scala
val col: ArrayLongColumn = ...
col.values.length <= col.size        // => false
```

So basically there really isn't any guaranteed correlation between the value of `size` here and the array. What `size` really represents is something a bit closer to an upper-bound (not even a definite value) on the number of defined rows. At the very least, I would suggest it shouldn't be called `size`, but I would be happier still if it just wasn't exposed. Why do we need this?

yggdrasil/src/main/scala/quasar/yggdrasil/table/Slice.scala
```diff
@@ -977,11 +977,23 @@ abstract class Slice { source =>
     */
   def materialized: Slice = {
     val size = source.size
-    val columns = source.columns flatMap {
+
+    val columns: Map[ColumnRef, Column] = source.columns flatMap {
+      case (ref, col: ArrayColumn[_]) =>
+        val defined = col.definedAt(0, size)
+        if (defined.nonEmpty) {
+          if (col.size == size)
```
djspiewak: If this is the only use of `col.size`, then I think we can more effectively resolve this by adding a check to `resize` to see if the column is guaranteed to already be the designated size. This also accounts for the fact that different columns (such as `ArrayLongColumn` and `MutableEmptyArrayColumn`) have different notions of "size-ness"

yggdrasil/src/main/scala/quasar/yggdrasil/table/Column.scala
alissapajer: Say `to == 13` and `max == 13`. This means that we want to clear the value at index `13`, and that the _length_ (in bits) of the bitset is `13`. This means that the last indexed value of the bitset is at index `12`, so it would be impossible to clear index `13`, so we don't call `bs.clear`. 

So I agree, will change.

alissapajer:
@djspiewak pushed changes so that in `resize`, we check the length of the array of values before doing the resize operation, for all `ArrayColumn`.

I didn't do the analogous for `BitSet`-backed columns because I haven't thought of a good way to do that. Open to suggestions.

PR #3915, opened 2018-10-29
Added tests for tectonic slice breaking and fixed off-by-one issue

djspiewak:
You'll notice that the unit test actually shows that we go one column *over* the limit. This is because we can't know that we've exceeded the column limit… until we actually exceed it. Fixing this in the current implementation would require undoing data that has already been put into columns and then re-doing that work against a new slice, which sounds hard, slow, and probably not worth it for right now. Ticketing for future improvement.

[ch2955]
[ch2350]

### DIFF
--- a/yggdrasil/src/main/scala/quasar/yggdrasil/table/SlicePlate.scala
+++ b/yggdrasil/src/main/scala/quasar/yggdrasil/table/SlicePlate.scala
@@ -259,7 +259,6 @@ private[table] final class SlicePlate(
       val slice = Slice(size, columns.to[λ[α => Map[ColumnRef, Column]]])
       completed += slice
 
-      nextThreshold = defaultMinRows
       size = 0
       columns.clear()
     }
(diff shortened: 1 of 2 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/table/SlicePlate.scala
```diff
@@ -31,7 +31,11 @@ private[table] abstract class EmptyFinishRowPlate[A] extends Plate[A] {
   def finishRow() = ()
 }
 
-private[table] final class SlicePlate(precise: Boolean)
+private[table] final class SlicePlate(
+    precise: Boolean,
+    defaultMinRows: Int = Config.defaultMinRows,
```
wemrysi: :+1:

djspiewak: I think this is actually a really nice idiom (grabbing global config as a parameter default, but keeping it in a constructor parameter).

PR #3917, opened 2018-10-30
Generalize ParseInstruction.Pivot

alissapajer:
```json
{ "a": 42, "b": true, "c": [1, 2, 3], "d": ["hi", "there"] }
```
with `b` and `d` retained and `c` pivoted will first be `Mask`ed, then be `Wrap`ed as so:
```json
{ "locus1": { "retain": { "b": true } }, "locus2": [1, 2, 3], "locus3": { "retain": { "d": ["hi", "there"] } } }
```
and then `Pivot`ed as so:
```scala
Pivot(Map((.locus1, ...), (.locus2, ...), (.locus3, ...)))
```

### DIFF
--- a/yggdrasil/src/test/scala/quasar/yggdrasil/table/TectonicParseInstructionSpec.scala
+++ b/yggdrasil/src/test/scala/quasar/yggdrasil/table/TectonicParseInstructionSpec.scala
@@ -36,7 +36,13 @@ object TectonicParseInstructionSpec extends ParseInstructionSpec {
     evalPlate(stream)(new IdsPlate(_))
 
   def evalPivot(pivot: Pivot, stream: JsonStream): JsonStream =
-    evalPlate(stream)(new PivotPlate(pivot, _))
+    if (pivot.pivots.size == 1)
+      pivot.pivots.head match {
+        case (path, (idStatus, structure)) =>
+          evalPlate(stream)(new PivotPlate(path, idStatus, structure, _))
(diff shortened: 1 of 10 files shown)


yggdrasil/src/test/scala/quasar/yggdrasil/table/TectonicParseInstructionSpec.scala
```diff
@@ -36,7 +36,13 @@ object TectonicParseInstructionSpec extends ParseInstructionSpec {
     evalPlate(stream)(new IdsPlate(_))
 
   def evalPivot(pivot: Pivot, stream: JsonStream): JsonStream =
```
alissapajer: Alternately, we could change the signature of `evalPivot` to take one of each of `CPath`, `IdStatus`, `CompositeParseType` and then have a separate spec for the "multipivot" tests.

alissapajer: This would allow implementors to extend `ParseInstructionSpec.SinglePivotSpec` but not `ParseInstructionSpec.MultiPivotSpec` if they've only implemented one or the other.

djspiewak: I rather like this idea. Especially since we already think we're going to split the plates.

PR #3918, opened 2018-11-01
No more Normalizable

alissapajer:


qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -379,6 +369,10 @@ object MapFuncCore {
       rewrite[T, A],
       ExtractFiltering[T, A]))
 
+  def freeMF[T[_[_]]: BirecursiveT: EqualT, A: Equal: Show](fm: Free[MapFunc[T, ?], A])
```
alissapajer: This was the only useful thing in all of `Normalizable`.

qscript/src/main/scala/quasar/qscript/rewrites/Rewrite.scala
```diff
@@ -233,7 +235,8 @@ class Rewrite[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] extends TTypes[
     ))(prism(ftf))
   }
 
-  private def normalizeWithBijection[F[a] <: ACopK[a]: Functor: Normalizable, G[_]: Functor, A](
+  // FOOBAR 5
```
alissapajer: These are my tracers through the rewrites, without which I'd be lost. It's hard to cherry pick them across branches as when I change things there are conflicts. So if it's okay with folks, I'd like to check them in to master. Once my current phase of this work is done, they'll be easy to grep for and delete.

alissapajer:
@jsantos17 if you're unable to get to this today, could you pass it off to @djspiewak? I want to keep these PRs moving forward so they don't build up as I delete more.

djspiewak:
@jsantos17 Beat you to it! :-D

PR #3919, opened 2018-11-01
Delete some more qscript rewrites

alissapajer:
Depends on #3918


alissapajer:
Aiming to keep both the PRs and the commits as incremental as possible.

alissapajer:
Ok, I verified that we are still generating `InterpretedRead` in this branch, as we should be.

jsantos17:
While checking this I realized we're no longer applying `FreeMap` normalizations to `LeftShift`'s repair. Consider this query generated by the latest frontend:

```
SELECT c.c1 AS A, c.c2 AS B, c.c3 AS C, c.c4 AS D FROM
   (SELECT p5.g7 AS c1, p5.g8 AS c2, p5.g9 AS c3, p5.g10 AS c4 FROM (
            SELECT (
               SELECT * FROM (SELECT ty.previous_addresses[_]{_} FROM `patients.data` AS ty) AS v 
               WHERE type_of(v) = "string") AS g10,
                  (SELECT * FROM (SELECT ty.previous_addresses[_].city FROM `patients.data` AS ty) AS v WHERE type_of(v) = "string") AS g7,
                  (SELECT * FROM (SELECT ty.previous_addresses[_].state FROM `patients.data` AS ty) AS v WHERE type_of(v) = "string") AS g8,
                  (SELECT * FROM (SELECT ty.last_visit FROM `patients.data` AS ty) AS v WHERE type_of(v) = "string") AS g9 FROM `patients.data` AS r6) AS p5) AS c
```

The QScript after this PR is a lot larger than the one from master. [3000 lines QScript here](https://gist.github.com/jsantos17/2a409b788fda0d1c2632c1c71bbbd13e).

PR #3923, opened 2018-11-02
Change evalPivot to evalSinglePivot in ParseInstructionSpec

alissapajer:


djspiewak:
*Technically* this is a breaking change, but only in the testkit, so I think revision is fine.

PR #3926, opened 2018-11-02
Add support for creating Disposable from a cats.effect.Resource

wemrysi:


wemrysi:
We already have something similar defined in `slamdata/quasar-s3` and I also need it for the new URL datasource, so it seemed to make sense to add it to `Disposable`.

nicflores:
Just read the docs on `Resource`, do we need this because we anticipate converting `Resource` into something else?

wemrysi:
@nicflores Yes, as I mentioned above, we're already converting it to `Disposable` in `quasar-s3` and I need to do it as well in the unreleased URL datasource.

PR #3928, opened 2018-11-05
Singleton ds should list resource as a resource

wemrysi:
Fixes an issue where listing the parent path of the singleton resource would return the resource name typed as a prefix instead of a resource.

### DIFF
--- a/connector/src/test/scala/quasar/connector/datasource/SingletonDatasourceSpec.scala
+++ b/connector/src/test/scala/quasar/connector/datasource/SingletonDatasourceSpec.scala
@@ -16,7 +16,7 @@
 
 package quasar.connector.datasource
 
-import slamdata.Predef.{Left, List, String}
+import slamdata.Predef.{Left, List, Some, String}
 
 import quasar.api.datasource.DatasourceType
 import quasar.api.resource._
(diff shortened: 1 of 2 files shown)


connector/src/test/scala/quasar/connector/datasource/SingletonDatasourceSpec.scala
```diff
@@ -40,6 +42,11 @@ object SingletonDatasourceSpec extends DatasourceSpec[IO, List] {
   val nonExistentPath = path / ResourceName("baz")
   def gatherMultiple[A](xs: List[A]) = IO.pure(xs)
 
+  "listing the parent of the resource should return the resource" >>* {
```
alissapajer: Should this test be in `DatasourceSpec` instead?

wemrysi: There's no requirement for a datasource to be nonempty, in general.

wemrysi: Though, I guess we do require it be nonempty in `DatasourceSpec`. The "prefixed child status agrees with pathIsResource" example _would_ catch this if we had it traverse the entire tree, but I'm not sure if that is a good idea as it could be a performance issue if a test datasource had lots of resources.

PR #3929, opened 2018-11-06
Add QScript regression test

alissapajer:
[ch3117]

### DIFF
--- a/it/src/test/scala/quasar/regression/QScriptRegressionSpec.scala
+++ b/it/src/test/scala/quasar/regression/QScriptRegressionSpec.scala
@@ -1,0 +1,190 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.regression
+
+import slamdata.Predef.{Map => SMap, _}
+import quasar._
+import quasar.api.QueryEvaluator
+import quasar.common.{PhaseResults, PhaseResultTell}
+import quasar.contrib.scalaz.{MonadError_, MonadTell_}
+import quasar.qscript._
+import quasar.run.{MonadQuasarErr, QuasarError, Sql2QueryEvaluator, SqlQuery}
+import quasar.sql.Query
+
+import cats.effect.IO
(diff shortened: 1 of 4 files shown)


it/src/test/scala/quasar/regression/RegressionQueryEvaulator.scala
```diff
@@ -0,0 +1,36 @@
...
+import quasar.common.PhaseResultTell
+import quasar.qscript._
+
+import matryoshka._
+import scalaz.Monad
+
+final class RegressionQueryEvaluator[
+    T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT,
+    F[_]: Monad: MonadPlannerErr: PhaseResultTell]
+    extends QueryEvaluator[F, T[QScriptEducated[T, ?]], QScriptCount] {
```
alissapajer: @wemrysi this turned out to be a pretty nifty use of `QueryEvaluator`!

alissapajer:
@jsantos17 @wemrysi we can use this to track how many `LeftShift` there are in queries to make sure our `LeftShift` qsu coalescence doesn't regress. We could also add a counter for `EquiJoin` (or any other nodes) if we want to ensure that all autojoins have been elided, for example.

it/src/test/scala/quasar/regression/QScriptRegressionSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+
+      // ch1473
+      // FIXME this should plan as two LeftShift before optimization, not three
+      val q4 = "select first[_].second{_}.third as value, first[_].second{_:} as key from mydata"
+      q4 in {
+        val result = count(q4)
+
+        result must countShiftedReadAs(0)
+        result must countInterpretedReadAs(1)
+        result must countLeftShiftAs(2)
```
djspiewak: Maybe it would be better to do `1` here then and `pendingUntilFixed` on the whole thing?

it/src/test/scala/quasar/regression/QScriptRegressionSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+      }
+    }
+  }
+
+  def countInterpretedReadAs(expected: Int): Matcher[QScriptCount] =
```
djspiewak: 👍

it/src/test/scala/quasar/regression/QScriptRegressionSpec.scala
alissapajer: One reason for keeping it as is, is so we don't further regress. We've done real work in qsu getting this down to three `LeftShift`, and we wouldn't want to inadvertently make it worse.

Also, it depends on if we think of "too many `LeftShift`s" as a bug or as an opportunity to improve performance. It's not a _correctness_ problem, as the query produces correct results. And fwiw, as we implement more `InterpretedRead` rewrites, some of the other tests in this suite will change their expectations as well.

alissapajer:
@djspiewak can we merge this?

alissapajer:
Since this has been approved, I'm going to merge it. If there is further discussion to be had about the test expectations, we can continue that conversation and push changes if necessary.

PR #3930, opened 2018-11-06
Move RValue to quasar.common.data

alissapajer:
[ch3166]

### DIFF
--- a/blueeyes/src/main/scala/quasar/blueeyes/json/JValue.scala
+++ b/blueeyes/src/main/scala/quasar/blueeyes/json/JValue.scala
@@ -17,8 +17,7 @@
 package quasar.blueeyes.json
 
 import quasar.blueeyes._
-import quasar.common.data.Data
-import quasar.precog.common._
+import quasar.common.data._
 
 import scalaz._, Scalaz._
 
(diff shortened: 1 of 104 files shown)


blueeyes/src/main/scala/quasar/blueeyes/json/JValue.scala
```diff
@@ -107,6 +108,161 @@ object JValue {
...
+    case CLocalDate(v) => JString(v.toString)
+    case CLocalTime(v) => JString(v.toString)
+    case CInterval(v) => JString(v.toString)
+
+    case CArray(_, _) => scala.sys.error("CArray not supported")
+
+    case CUndefined => JUndefined
+  }
+
+  def fromRValueRaw(rvalue: RValue): JValue = rvalue match {
```
alissapajer: There is some duplication among these functions, and it's not clear that we actually need both the "raw" and "not raw" versions. For now, I recreated them here exactly as they are in master; we can look into consolidating them later.

common/src/test/scala/quasar/common/data/RCValueGenerators.scala
```diff
@@ -16,15 +16,17 @@
...
-import quasar.precog.common._
+import slamdata.Predef._
+import quasar.common.data._
 import quasar.pkg.tests._, Gen._
+
 import qdata.time.TimeGenerators._
 
+import scala.sys
+
 trait RCValueGenerators {
```
alissapajer: Discovered that we have a couple `RValue`/`CValue` generators. I kept them both for now; we can consolidate them later.

wemrysi:
@alissapajer thank you for renaming `CValue.scala` to `RValue.scala`!

PR #3933, opened 2018-11-07
Construct Quasar with a List[DatasourceModule]

alissapajer:


run/src/main/scala/quasar/run/Quasar.scala
```diff
@@ -126,7 +116,7 @@ object Quasar {
 
       mr <- Stream.bracket(
         DatasourceManagement[Fix, F, UUID, Double](
-          modules,
+          IMap.fromList(datasourceModules.map(ds => (ds.kind, ds))),
```
alissapajer: Just realized that the `IMap` in the `Quasar.apply` constructor was preserving the property that we can only have one datasource per `DatasourceType`.

wemrysi: Yeah, it'd _seem_ that if you had two `DatasourceModule`s with the same type that they'd be the same module, so in that case it wouldn't really matter if there were more than one, we'll only use one of them.

If one were to reuse a type for multiple modules, then all but one of them would disappear, but the contract for `DatasourceType` is that they're unique per module, so I think that's unavoidable.

Maybe we log a warning when we see a type multiple times? Or we could error and abort altogether.

wemrysi: Alternatively, we could stick with the `IMap` and just validate that the keys match the type of the value and error or maybe evict and warn if not?

alissapajer:

it/src/test/scala/quasar/regression/Sql2QueryRegressionSpec.scala
```diff
@@ -94,7 +94,10 @@ abstract class Sql2QueryRegressionSpec extends Qspec {
 
     evalCfg = SstEvalConfig(1L, 1L, 1L)
 
-    q <- Quasar[IO](precog, ExternalConfig.Empty, SstEvalConfig.single, blockingPool)
+    q <- Quasar[IO](
+      precog,
+      List(DatasourceModule.Lightweight(LocalDatasourceModule)),
```
wemrysi: :tada:

alissapajer: Yeah I'm happy with how this turned out!

PR #3934, opened 2018-11-08
Add LocalParsedDatasource

alissapajer:
[ch2769]

Depends on #3933


alissapajer:
@wemrysi if it's :+1: then I'll do your suggestion separately and not hold up merge for it. The companion object is more in line with the pattern we've been following elsewhere so I'm all for it.

PR #3935, opened 2018-11-08
Add apply constructor for EvaluableLocalDatasource

alissapajer:


impl/src/main/scala/quasar/impl/datasource/local/EvaluableLocalDatasource.scala
```diff
@@ -95,3 +95,12 @@ final class EvaluableLocalDatasource[F[_]: ContextShift: Timer](
   private def toResourceName(jp: JPath): ResourceName =
     ResourceName(jp.getFileName.toString)
 }
+
+object EvaluableLocalDatasource {
```
alissapajer: This does make the construction sites much nicer!

PR #3945, opened 2018-11-14
Delete ExpandDirs and DiscoverPath

alissapajer:
[ch3321]


alissapajer:
I'll be removing `ShiftedRead` and replacing it with `Read` (starting when we compile from `LogicalPlan`) in a separate PR.

djspiewak:
Definitely breaking. I think that if anyone is actually *relying* on this stuff that maybe they deserve to be broken, but it's probably better to have our semver reflect reality.

One thing I'd like to do at some point is reduce the footprint of our public API.

alissapajer:
Travis build failed due to the ch2153 flake. Restarted it.

PR #3946, opened 2018-11-14
Added autobump for azure

djspiewak:


nicflores:
@djspiewak, I think there's a merge conflict.

djspiewak:
@nicflores There is not. I think you have a slightly old version of `sdmerge`

PR #3952, opened 2018-11-16
Add woodstox as a dep as a quick and dirty way to get azure working

rintcius:
[ch3385]


djspiewak:
Does this actually solve the problem? In other words, is something literally walking the classpath entries?

rintcius:
@djspiewak In the repl it does work with this. 
Not sure what you mean with "literally walking the classpath entries". The idea behind this change is to set the properties like that xmlParserFactory correct so that the Netty which has its own EventExecutorGroup runs its threads with these properties.

Did you see my comment in ch3385? With that it may make more sense.

PR #3958, opened 2018-11-19
Delete DeadEnd

alissapajer:
Also not sure the last time it was in use.

Depends on #3957

PR #3959, opened 2018-11-19
Delete reducers/buckets qscript rewrites

alissapajer:
Depends on #3958 

Deletes the qscript rewrites that removed duplicate reducers and buckets, and unused reducers.

These cases shouldn't arise in the first place. If we find they are arising (in the future if/when the product supports bucketed things) we can address it separately then, probably at the source in qsu.


alissapajer:
@jsantos17 oops, I pushed a few more commits to this branch that were intended for another PR.

I don't think the constant folding in `MapFuncCore` is required; the logic I was thinking is important is in `rewrite` [here](https://github.com/slamdata/quasar/blob/master/qscript/src/main/scala/quasar/qscript/MapFuncCore.scala#L444), for example. If there is actually reason to keep around the constant folding, I'll probably just close this PR and reopen with just the earlier commits.

alissapajer:
An integration test failed, so I'm going to just close this and reopen it rebased once I figure out what's going on.

PR #3960, opened 2018-11-20
Remove constant folding from MapFuncCore normalization

alissapajer:


alissapajer:
@jsantos17 I don't think the constant folding in `MapFuncCore` is required; the logic I was thinking is important is in `rewrite` [here](https://github.com/slamdata/quasar/blob/master/qscript/src/main/scala/quasar/qscript/MapFuncCore.scala#L444), for example. Is this required for any rewrites for any reason that you know of?

alissapajer:
Travis failed due to ch3513 (`DatasourceManagementSpec` flake).

jsantos17:
@alissapajer looks good, but there are a few conflicts to resolve.

PR #3966, opened 2018-11-20
Single field select

jsantos17:
TODOs preventing merge of this PR:

- [ ] Testing with the frontend once `slamdata-backend` catches up on bump PRs.

Will assign someone after `TODO`s are done.

Fun fact: the launcher does work with `publishLocal` for `quasar` and `slamdata-backend`. It doesn't work for plugins for some reason tho.

Bug was in the `SQL -> LogicalPlan` phase.


core/src/main/scala/quasar/compile/package.scala
```diff
@@ -103,7 +103,7 @@ package object compile {
   ): Option[String] = {
     val loop: T => (Option[String] \/ (Option[String] \/ T)) =
       _.project match {
-        case Ident(name) if !relationName.element(name)  => some(name).left
+        case Ident(name)                                 => some(name).left
```
jsantos17: Not sure what was the purpose of this check. But it's not necessary and the test suite is green without it.

jsantos17:
[ch63]

djspiewak:
Does this still allow things like the following to work?

```sql
select (select a from foo) + (select b from foo)
```

The frontend actually uses this a fair bit.

jsantos17:
@djspiewak unfortunately no. Something like this works:

```
select (select a from foo).a + (select b from foo).b
```

I haven't done much testing with the frontend, but so far every query generated by the frontend seems to work in this branch.

djspiewak:
@jsantos17 I'm honestly not sure we want to do this then. At least, not quite yet. We do need to address cases like this as part of the eventual SQL compatibility effort, but there are several other similar issues which are similar but not addressed here (such as field reordering). SELECT clauses are just weird and need to be treated more carefully than they are right now in the parser.

jsantos17:
@djspiewak do you know what part of the frontend requires this? I'd like to assess the extent of the breakage induced by this PR.

djspiewak:
@jsantos17 I know I exploited this feature with @garyb when we came up with the encoding for type filtering column selection, so it should be effectively *everything* in the UI. If that isn't broken, then perhaps the impact is less than I'm expecting.

I want to think about this more from the standpoint of what we want to do in general for ANSI SQL compatibility of SELECT clauses.

jsantos17:
@djspiewak I tested the type filtering queries and they're working just fine. I'll see if I can add some tests to make sure they keep working.

garyb:
The exact form we generate for this is a little different:

``` sql
SELECT
  (SELECT * FROM (SELECT ty{_}.dateTime FROM `/data.json` AS ty) AS v WHERE type_of(v) = "number") AS g4 
FROM 
  `/data.json` AS r3
```

But I think even if this does/did break the current codegen in some way we could work around it by just doing the re-projection, if it's as simple that.

jsantos17:
With regards to this query

```
SELECT
  (SELECT * FROM (SELECT ty{_}.dateTime FROM `/data.json` AS ty) AS v WHERE type_of(v) = "number") AS g4 
FROM 
  `/data.json` AS r3
```

The QScript in this branch is identical to the QScript in master. It might be this specific case only. We definitely need more extensive testing with the frontend to see if any behavior changes. Also, happy to implement any changes necessary in the frontend.

jsantos17:
After much investigation I have discovered the frontend is working fine because of a bug in this implementation.

`select baz.foo from bar as baz` produces an unwrapped result, while `select foo from bar` produces a wrapped result. The frontend always uses the former I think.

jsantos17:
I have fixed the bug and it indeed revealed breakage in frontend-related queries. Given the frontend relies on the breaking functionality I'd prefer to run the frontend's end-to-end tests before merging this.

djspiewak:
After a ton of thought and planning, I think I'm okay merging this (once the frontend tests are run; @jsantos17 can you release a snapshot for this PR so that they can?). We're going to need to make a lot more changes here in the future, but for right now, I think this is better than it was.

jsantos17:
I have published this PR as `91.0.6-45814a5`. 

@garyb is it yet possible to run the frontend's end-to-end tests? Happy to do it myself if shown how.

jsantos17:
@garyb that'd be super helpful. I just tried to publish a `slamdata-backend` depending on this snapshot, but the tests still run against the last non-snapshot versions of quasar.

rintcius:
@jsantos17 Did you update `slamdata-backend-version`? It should take the version from there.

garyb:
I'm taking a look at making the arguments overrideable now too.

jsantos17:
@rintcius yes. Published as `27.0.4-bb1d137` if you want to give it a go.

jsantos17:
I'm going to close this. I'd rather fix code generation in the frontend after introducing native support for type-filtering.

PR #3969, opened 2018-11-21
Gracefully handle a missing plugins dir

wemrysi:


impl/src/main/scala/quasar/impl/external/ExternalDatasources.scala
```diff
@@ -63,9 +63,18 @@ object ExternalDatasources extends Logging {
 
     val moduleStream: Stream[F, DatasourceModule] = config match {
       case PluginDirectory(directory) =>
-        convert.fromJavaStream(F.delay(Files.list(directory)))
-          .filter(_.getFileName.toString.endsWith(PluginExtSuffix))
-          .flatMap(loadPlugin[F](_, blockingPool))
+        Stream.eval(F.delay((Files.exists(directory), Files.isDirectory(directory)))) flatMap {
```
djspiewak: You know, I had never thought of tupling to merge two side-effecting calls in this way. That's clever.

wemrysi:
@alissapajer thanks!

PR #3970, opened 2018-11-21
Use QScriptEducated in place of duplicate QScriptRead

alissapajer:


alissapajer:
If someone were using the old alias by name this would be breaking... so `breaking`?

PR #3971, opened 2018-11-21
Delete unnecessary components of coalesceQC

alissapajer:
Depends on #3968 

What's remaining in `coalesceQC` are the rewrites that I'm expecting we'll need, along with the machinery to recurse into `FreeQS`. 

The plan forward (not in this PR) is threefold:
1) We can rewrite what is current `Coalesce` to act directly on `QScriptEducated`. This will remove the many lines of boilerplate machinery and make it easier to work within.
2) `Normalizable` is currently _also_ recursing into `FreeQS`, applying the entire `Rewrite` as well as `MapFunc` normalization (as well as possibly other rabbit hole atrocities). We need to centralize this recursion, and split out the `MapFunc` normalizations from the `Coalesce` normalizations. I think we'll be able to apply the `MapFunc` normalizations as a separate pass after we've coalesced nodes. Imagine two separate `PhaseResult` for these two steps!
3) We are calling `Normalizable` directly twice (for no reason) and calling both `Normalizable` and `Coalesce` who knows how many times via `repeatedly`. There _may_ (but may not) still be a use case for `repeatedly`, but we will stop calling it arbitrarily.

### DIFF
--- a/qscript/src/main/scala/quasar/qscript/rewrites/Coalesce.scala
+++ b/qscript/src/main/scala/quasar/qscript/rewrites/Coalesce.scala
@@ -26,7 +26,6 @@ import quasar.fp.ski._
 import quasar.qscript._
 import quasar.qscript.RecFreeS._
 import quasar.qscript.MapFuncCore._
-import quasar.qscript.MapFuncsCore._
 
 import matryoshka.{Hole => _, _}
 import matryoshka.data._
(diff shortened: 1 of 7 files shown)


alissapajer:
[ch3119]

Going to close ch3119 with this PR, and open new more specific stories for what's remaining.

alissapajer:
@jsantos17 landed #3968

PR #3976, opened 2018-11-28
Add the 'Project' ParseInstruction

wemrysi:
[ch3262]

### DIFF
--- a/impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala
+++ b/impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-package quasar.mimir.evaluate
+package quasar.impl.evaluate
 
-import slamdata.Predef.List
+import slamdata.Predef._
 
 import quasar.{CompositeParseType, IdStatus, ParseInstruction, ParseType}
 import quasar.IdStatus.{ExcludeId, IdOnly, IncludeId}
@@ -24,17 +24,22 @@ import quasar.ParseInstruction.{Ids, Mask, Pivot, Wrap}
 import quasar.common.{CPath, CPathField, CPathIndex, CPathNode}
 import quasar.common.data._
 
+import scala.collection.Iterator
+
+import scalaz.std.anyVal._
+import scalaz.syntax.equal._
 import scalaz.syntax.std.stream._
 
 object RValueParseInstructionInterpreter {
 
+  @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
   def interpret(instructions: List[ParseInstruction], rvalue: RValue): List[RValue] =
     instructions.foldLeft(List[RValue](rvalue)) {
       case (prev, instr @ Mask(_)) =>
         prev.flatMap(interpretMask(instr, _).toList)
 
       case (prev, instr @ Pivot(pivots)) =>
-        if (pivots.size == 1)
+        if (pivots.size === 1)
           pivots.head match {
             case (path, (idStatus, structure)) =>
               prev.flatMap(interpretSinglePivot(path, idStatus, structure, _))
@@ -75,6 +80,7 @@ object RValueParseInstructionInterpreter {
     //
     // build up:
     // only retain if type checks
+    @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
     def inner(masks: Map[List[CPathNode], Set[ParseType]], rvalue: RValue): Option[RValue] = {
 
       val prefixes: Set[CPathNode] = masks.keySet collect {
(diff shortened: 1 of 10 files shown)


yggdrasil/src/main/scala/quasar/yggdrasil/table/ProjectPlate.scala
```diff
@@ -0,0 +1,194 @@
...
+        Signal.SkipRow
+    }
+  }
+
+  override def unnest(): Signal = {
+    super.unnest()
+
+    val signal = if (projectState > FOCUSED) {
+      Signal.SkipRow
+    } else if (focusDepth == 0) {
```
djspiewak: I think this can be replaced by `if (cursor eq Nil)` (where `Nil` is cached in a `val`, of course), which allows you to remove the `focusDepth` stuff.

impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala
```diff
@@ -75,6 +86,7 @@ object RValueParseInstructionInterpreter {
     //
     // build up:
     // only retain if type checks
+    @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
```
djspiewak: Are any of these new recursion suppressions used? I couldn't find any added recursive call sites.

wemrysi: Aye, https://github.com/slamdata/quasar/blob/8a0c7f60a1b04d7c84cc89853642bf64ba87d87b/impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala#L109, for example.

They don't show up in the PR diff though.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ProjectPlate.scala
```diff
@@ -0,0 +1,194 @@
...
+    extends Plate[A]
+    with ContinuingNestPlate[A]
+    with CPathPlate[A] {
+
+  private val SEARCHING: Int = 0
+  private val FOCUSED: Int = 1
+  private val FINISHED_SEARCHING: Int = 2
+  private val FINISHED_FOCUSED: Int = 3
+
+  private var focus: List[CPathNode] = path.nodes
```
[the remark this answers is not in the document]
wemrysi: Since we're projecting, we never have to remember where we came from, just where we're going, so we only care about the head of the focus. Once we match it, we note that we did and toss it.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ProjectPlate.scala
wemrysi: `focusDepth` is there to detect the case where we've partially matched the focus and end up backtracking, meaning we'll never match it and can safely `SkipRow`.

i.e. if our focus is `.a.b.c` and we see `nestMap(a), nestMap(b), tru(), unnest()`, then we know we're never going to see `.c` so we can abort immediately.

wemrysi:
@djspiewak Hold off on merging this, I just realized I think there is a bug in the plate impl where it will project any matching _suffix_ of the plate cursor instead of requiring the focus to match at the root, I'll ping you once I've fixed or confirmed absence.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ProjectPlate.scala
wemrysi: We reset `focusDepth` to `0` every time we match a bit of the focus, so it also tells us when we're done projecting. Maybe it is ill-named?

wemrysi:
@djspiewak Ok, problem solved.

yggdrasil/src/main/scala/quasar/yggdrasil/table/ProjectPlate.scala
```diff
@@ -0,0 +1,206 @@
...
+
+      case _ =>
+        Signal.SkipRow
+    }
+  }
+
+  override def nestArr(): Signal = {
+    super.nestArr()
+
+    projectState match {
```
djspiewak: 🏎 `(projectState: @switch) match` instead, and then apply `@inline` and `final` to all of the states above.

wemrysi: I tried the `@switch` and compiler said it couldn't do it, probably b/c I didn't have the `@inline` `final`, will try it out.

wemrysi: Hm, still no luck with the `@inline` and `final`. I even tried capturing the current value of `projectState` in a local `val`.

djspiewak: Try `private[this]`

wemrysi: No joy there either, I tried making all private `val` and `vars` `private[this]`.

PR #3977, opened 2018-11-29
Allow plugin files to be specified in repl

rintcius:
With this you can start with plugins without any extracting into `~/.config/quasar/plugin`.
Just assemble the plugins with `datasourceAssemblePlugin`.

Example `repl/run` invocation from `sbt`:

`repl/run -P path-to/quasar-datasource-s3/datasource/target/scala-2.12/plugin-build/s3.plugin -P path-to/quasar-datasource-azure/datasource/target/scala-2.12/plugin-build/azure.plugin`

### DIFF
--- a/impl/src/main/scala/quasar/impl/external/ExternalConfig.scala
+++ b/impl/src/main/scala/quasar/impl/external/ExternalConfig.scala
@@ -21,9 +21,8 @@ import slamdata.Predef._
 import java.nio.file.{Files, Path}
 
 import cats.effect.Sync
-import cats.syntax.applicativeError._
 import scalaz.std.list._
-import scalaz.syntax.all._
+import scalaz.syntax.traverse._
 import shims._
 
 sealed trait ExternalConfig extends Product with Serializable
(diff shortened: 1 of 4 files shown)


rintcius:
[ch 3655]

impl/src/main/scala/quasar/impl/external/ExternalConfig.scala
```diff
@@ -46,6 +47,23 @@ object ExternalConfig {
     entriesF.map(ExternalConfig.ExplodedDirs(_))
   }
 
+  def sanitize[F[_]: Sync](cfg: ExternalConfig): F[ExternalConfig] = {
```
rintcius: The logic in `sanitize` is derived from SDBE, but I'm happy to change it, e.g. `raiseError`  instead of ignoring in l.58

rintcius:
[ch3655]

impl/src/main/scala/quasar/impl/external/ExternalConfig.scala
```diff
@@ -46,6 +47,23 @@ object ExternalConfig {
...
+    cfg match {
+      case ExplodedDirs(_) => cfg.pure[F]
+      case PluginDirectory(dir) => cfg.pure[F]
+      case PluginFiles(files) =>
+        for {
+          fileErrs <- files.traverse(path => Sync[F].delay(path).attempt)
```
wemrysi: Unless I'm missing something, this doesn't appear to do anything other than lift each `Path` into `F[_]`.

repl/src/main/scala/quasar/repl/Main.scala
```diff
@@ -86,11 +92,30 @@ object Main extends IOApp {
...
+          files <- opts.pluginFiles
+          qs <- quasarStream[IOT](files, blockingPool)
+                  .evalMap(repl[IOT])
+                  .compile
+                  .last
+                  .run.map(_._2.getOrElse(ExitCode.Success))
+        } yield qs
+      case Failure(f) =>
+        val (s, exitCode) = renderFailure(f, "repl")
+        IO.delay(println(s)) >> ExitCode(exitCode.toInt).pure[IO]
```
wemrysi: :golf: `.as(ExitCode(exitCode.toInt))`

repl/src/main/scala/quasar/repl/Paths.scala
```diff
@@ -38,6 +38,9 @@ object Paths {
       : F[JPath] =
     F.delay(JFiles.createTempFile(prefix, suffix))
 
+  def fromString[F[_]](s: String)(implicit F: Sync[F]): F[JPath] =
+    F.delay(Try(getPath(NonEmptyList(s)))).flatMap(F.fromTry)
```
wemrysi: Is the `Try` necessary? Seems like `F.delay(getPath(...))` is equivalent.

PR #3978, opened 2018-11-29
Move freeMF to MapFuncCore

alissapajer:
Frees qsu from a dependency on `Normalizable`.

### DIFF
--- a/qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
+++ b/qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
@@ -266,7 +266,7 @@ object MapFuncCore {
       : CoMapFuncR[T, A] => CoMapFuncR[T, A] =
     repeatedly(rewrite[T, A])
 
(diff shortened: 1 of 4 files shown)


qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -266,8 +266,10 @@ object MapFuncCore {
       : CoMapFuncR[T, A] => CoMapFuncR[T, A] =
     repeatedly(rewrite[T, A])
 
-  // TODO: This could be split up as it is in LP, with each function containing
-  //       its own normalization.
+  def freeMF[T[_[_]]: BirecursiveT: EqualT, A: Equal: Show](fm: Free[MapFunc[T, ?], A])
```
wemrysi: How about naming this `normalized`?

wemrysi: We could also home this under `MapFunc`, but maybe we can do that in a separate PR since `normalize` and `rewrite` are already here?

PR #3979, opened 2018-11-30
Introducing individual phases for individual qscript rewrites

alissapajer:
[ch3621]

Deletes `Normalizable`, `Coalesce`, `Rewrite`, and `Unirewrite`.

Applies post-graduation transformations as individual phases in `QScriptEvaluator` (with phase results that print in the REPL) in this order:
- `ThetaToEquiJoin`
- `NormalizeQScript`
- `NormalizeQScriptFreeMap`
- `RewritePushdown` (mimir-specific optimization)

This change gives us control over how `QScriptEducated` is transformed into `QScriptNormalized` so that we will be able to generate consistent and optimal `ParseInstruction`s.

The only remaining use of `repeatedly` is in `FreeMap` normalization; we can decide separately if that's still needed.


alissapajer:
Unused imports, hang tight...

djspiewak:
Epic AF

wemrysi:
@alissapajer Did you happen to try using `quasar.qscript.Trans` when redefining these rewrites? It has generalized branch application built in. Even if that doesn't work out, there is still `quasar.qscript.Branches` for traversing all branches.

I've been thinking that a similar class for `FreeMap`s would also be useful, though if we don't have any other use for it than normalizing freemaps, then might not be worth.

alissapajer:
@wemrysi :+1: to using (or modifying, if necessary) the existing `Trans` and `Branches` so that we can use them here and in other places we need to do branch traversals (like in `RewritePushdown`).

jsantos17:
Just started reviewing this. Will test a few frontend queries.

djspiewak:
@alissapajer Future-architecture question: does it make sense to have multiple pushdown rewrites?

alissapajer:
@jsantos17 I just confirmed that the query you found has the same problem in master. I think we're not matching on the pattern where one side of the `ConcatMaps` is a `MakeMap` and the other side isn't. Could you open a new story for this one?

alissapajer:
@djspiewak maybe. My understanding of @wemrysi 's current work is that it's making the pushdown rewrites more individually composable, so I'm not sure if multiple pushdown rewrites (from the point of view of `QScriptEvaluator`) will be useful. Seems like the kind of thing we'll want to let actual implementations drive.

wemrysi:
@jsantos17 @alissapajer that expression is a bit more involved to simplify as we have to statically know all of the structure of `LeftSide` in order to do so safely. It may also contain an entry for `0` that should take precedence due to right-biased `ConcatMaps`.

alissapajer:
@wemrysi oh yeah, very good point.

jsantos17:
@wemrysi right. Makes a lot of sense.

PR #3984, opened 2018-12-05
Repl cleanup and things

djspiewak:
A bunch of changes cleaning up `repl`, trying to make things better, trying to find the memory leak and essentially just failing.

### DIFF
--- a/repl/src/main/scala/quasar/repl/Repl.scala
+++ b/repl/src/main/scala/quasar/repl/Repl.scala
@@ -20,10 +20,10 @@ package repl
 import slamdata.Predef._
 import quasar.build.BuildInfo
 import quasar.common.{PhaseResultListen, PhaseResultTell}
+import quasar.concurrent.BlockingContext
 import quasar.run.{MonadQuasarErr, Quasar}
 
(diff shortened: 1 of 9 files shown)


repl/src/main/scala/quasar/repl/Repl.scala
```diff
@@ -37,42 +37,48 @@ import scalaz._, Scalaz._
...
 
   val F = ConcurrentEffect[F]
 
-  private val read: F[Command] = F.delay(Command.parse(reader.readLine(prompt)))
+  private val read: F[Command] =
+    cs.evalOn(Tag.unwrap(blockingPool))(
+      F.delay(reader.readLine(prompt))).map(Command.parse)
 
-  private def eval(cmd: Command): F[Evaluator.Result[Stream[F, String]]] =
-    F.start(evaluator(cmd)) >>= (_.join)
```
rintcius: IIRC I added this to fix a concurrency issue, but apparently this is not needed anymore.

I think you can reduce the `ConcurrentEffect` constraint now, I guess to `Sync` and `ContextShift`?

PR #3985, opened 2018-12-06
Alternative maths

djspiewak:
The problem here is that we were computing the number of splits correctly (which should have really been more correctly named "number of output slices" or something like that), but the size of those splits was incorrect (we took the ceiling rather than the floor, for… no reason). This in turn meant that our rolling index offset drift off, and we could get negative slices out the other end.

[ch3773]


jsantos17:
Looks like there are some actual `it` failures here.

PR #3986, opened 2018-12-07
Generalize pushdown for all parse instructions

wemrysi:


wemrysi:
[ch3245]
[ch3242]
[ch2862]
[ch3102]

alissapajer:
Found a bug locally where `select a{_} from foo` isn't getting rewritten to `InterpretedRead`. We should test this query on this branch (and add it to `QScriptRegressionSpec`) to confirm it's fixed here.

wemrysi:
@alissapajer confirmed and tests added.

alissapajer:
Travis failed with ch2153; restarted.

wemrysi:
@alissapajer more tests added, ready for another look

PR #3987, opened 2018-12-07
Eliminate Quasar's dependency on mimir

alissapajer:
[ch3725]

### DIFF
--- a/mimir/src/main/scala/quasar/mimir/evaluate/QuasarImpl.scala
+++ b/mimir/src/main/scala/quasar/mimir/evaluate/QuasarImpl.scala
@@ -49,8 +49,9 @@ import matryoshka.data.Fix
 import pathy.Path._
 import shims._
 
-final class QuasarImpl[F[_]: ConcurrentEffect: Monad: MonadQuasarErr: PhaseResultTell: MonadError_[?[_], ResourceError]](
-    precog: Precog)(
+final class QuasarImpl[F[_]: ConcurrentEffect: Monad: MonadQuasarErr: PhaseResultTell: MonadError_[?[_], ResourceError]] private (
+    precog: Precog,
+    pushdown: PushdownControl[F])(
     implicit
     cs: ContextShift[IO],
     ec: ExecutionContext) {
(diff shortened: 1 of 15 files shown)


alissapajer:
Snapshot released as `91.2.5-c14d648`.

mimir/src/main/scala/quasar/mimir/evaluate/QuasarImpl.scala
```diff
@@ -0,0 +1,148 @@
...
+        "UUID",
+        StoreKey.stringIso composePrism stringUuidP),
+      "TableRef",
+      rValueTableRefP(rValueSqlQueryP))
+
+  val pTableStore = MimirPTableStore[F](precog, PreparationLocation)
+
+  val pushdown = Ref.of[F, Pushdown](Pushdown.EnablePushdown).map(p => new PushdownControl(p))
+
+  val federation = MimirQueryFederation[Fix, F](precog, pushdown)
```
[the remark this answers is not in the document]
alissapajer: Changed pushed :+1:

mimir/src/main/scala/quasar/mimir/evaluate/QuasarImpl.scala
```diff
@@ -76,7 +77,7 @@ final class QuasarImpl[F[_]: ConcurrentEffect: Monad: MonadQuasarErr: PhaseResul
 
   val pTableStore = MimirPTableStore[F](precog, PreparationLocation)
 
-  val pushdown = Ref.of[F, Pushdown](Pushdown.EnablePushdown).map(p => new PushdownControl(p))
+  val pushdown = new PushdownControl(Ref.unsafe[F, Pushdown](Pushdown.EnablePushdown))
```
wemrysi: If we're going to use `unsafe` here, then we should wrap the `new QuasarImpl` in an `F.delay` over in `QuasarImpl.apply`. Alternatively the `QuasarImpl` class could accept a `PushdownControl` argument and we construct the `Ref` in `QuasarImpl.apply`.

djspiewak: Is there a way we can just not use `Ref.unsafe`? It's meant for some rather weird situations that are better handled with the 2-effect constructors.

alissapajer: Yeah, we can.

PR #3989, opened 2018-12-11
Lightweight datasource works with InterpretedRead[ResourcePath]

cryogenian:
I'm going to add a couple tests too


connector/src/main/scala/quasar/connector/datasource/SingletonDatasource.scala
```diff
@@ -41,13 +42,15 @@ final class SingletonDatasource[F[_]: MonadResourceErr, G[_]: Applicative, R] pr
 
   val kind: DatasourceType = tpe
 
-  def evaluate(path: ResourcePath): F[R] =
+  def evaluate(iRead: InterpretedRead[ResourcePath]): F[R] = {
+    val path = iRead.path
     if (path === location)
       resource
-    else if (location.relativeTo(path).isDefined)
+    else if (location.relativeTo(iRead.path).isDefined)
```
wemrysi: Mind using either `val path` or `iRead.path` instead of both, for consistency?

impl/src/main/scala/quasar/impl/datasource/local/LocalInterpretedReadDatasource.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import qdata.{QData, QDataDecode, QDataEncode}
+import qdata.json.QDataFacade
+
+import scala.collection.mutable.ArrayBuffer
+
+import scalaz.syntax.tag._
+import shims._
+
+object LocalInterpretedReadDatasource  {
```
wemrysi: I don't understand why this is needed, what do you have in mind?

connector/src/main/scala/quasar/connector/datasource/SingletonDatasource.scala
cryogenian: :+1:

impl/src/main/scala/quasar/impl/datasource/local/LocalInterpretedReadDatasource.scala
cryogenian: Mostly for testing. TBH I didn't totally got the idea of having both `LocalDatasource` and `LocalParsedDatasource` in *main*, they do the same thing, right? Happy to move this to *test* if you wish 🙂

wemrysi: The tests this is used in don't really test anything that isn't introduced in this class, though. So if it didn't exist there wouldn't be anything to test, right? Makes me thing we don't really need it.

As for `LocalParsedDatasource`, I agree it isn't needed in `main` could probably be moved to `test` at some point.

cryogenian: You're totally right! The tests checks only this class and this class isn't used anywhere, I'll remove both 🤦‍♂️

cryogenian:
@wemrysi 

+ `iRead.path` -> `path`
+ Removed `LocalInterpretedReadDatasource` and tests for it

wemrysi:
@cryogenian build failed due to unused imports

PR #3996, opened 2018-12-12
Build cleanup

alissapajer:


alissapajer:
We publish test artifacts so technically this is `breaking`. I want to make it `revision` though.

alissapajer:
Oh interesting. Build failed because we're running
```
      "testOnly -- failtrace" \
      "exclusive:testOnly -- failtrace"
```
in Travis. Do we still care about the `exclusive` variant?

djspiewak:
Ehhhh. I guess I'm okay with that. SDBE's tests might break, but it'll probably be okay.

alissapajer:
I mean, `breaking` is technically correct, so I won't object :)

PR #4000, opened 2018-12-13
Extract mimir

alissapajer:
[ch1052]


alissapajer:
@djspiewak let's merge this after we've seen mimir publish successfully.

PR #4005, opened 2018-12-14
Rename artifacts to remove -internal suffix

alissapajer:


build.sbt
```diff
@@ -187,15 +187,15 @@ lazy val qsu = project
...
+  .settings(name := "quasar-connector")
   .dependsOn(
     foundation % "test->test",
     qscript)
   .settings(commonSettings)
   .enablePlugins(AutomateHeaderPlugin)
 
 lazy val core = project
-  .settings(name := "quasar-core-internal")
+  .settings(name := "quasar-core")
```
alissapajer: I kept this as `quasar-core` (instead of just `quasar`) because the `core` project is in no way the heart of quasar.

djspiewak: Also, I tend to use the following convention:

- If `core` contains *all* of the other projects and thus represents the mega dependency, then just name it after the parent project (e.g. `quasar`)
- If `core` is independent of the other modules and thus represents a split dependency, name it "parent-core" (e.g. `quasar-core`)
- If you aesthetically prefer one thing or another, do that and ignore all the rules, including this one

PR #4008, opened 2018-12-14
Removed quasar-repl direct autobump

djspiewak:


djspiewak:
Nope, we actually need this.

PR #4018, opened 2018-12-17
Rely on publishAndTag branch checking

alissapajer:


djspiewak:
Technically, this removes some branch checking. `publishAndTag` isn't run at all right now on PR branches. It will be run after your change. `after_success` runs on all builds, not just master builds. *However*, because the `after_success` block does not include `set -e`, the `publishAndTag` script will fail… but won't kill the build. I think. We need to make sure that automerging continues to work after this change.

PR #4019, opened 2018-12-17
Don't automatically merge without confirming it's just a revision bump

djspiewak:
Lots of dependencies…

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,13 +50,16 @@ script:
 after_success:
   - |
     if [ $TRAVIS_PULL_REQUEST != "false" ] && [[ $TRAVIS_PULL_REQUEST_BRANCH =~ 'version-bump' ]]; then
-      clone_dir=$(mktemp -d "/tmp/slamdata-bump.XXXXXXXX")
-      git clone --depth 1 https://$GITHUB_TOKEN@github.com/slamdata/devtools.git $clone_dir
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -50,14 +50,17 @@ script:
 after_success:
   - |
     if [ $TRAVIS_PULL_REQUEST != "false" ] && [[ $TRAVIS_PULL_REQUEST_BRANCH =~ 'version-bump' ]]; then
-      clone_dir=$(mktemp -d "/tmp/slamdata-bump.XXXXXXXX")
-      git clone --depth 1 https://$GITHUB_TOKEN@github.com/slamdata/devtools.git $clone_dir
-
-      for t in "$(scripts/listLabels | grep 'version: revision')"; do
-        $clone_dir/bin/sdmerge $TRAVIS_REPO_SLUG $TRAVIS_PULL_REQUEST
-      done
+      if [[ "$(git diff HEAD^ | grep -e '^[-+][0-9]' | sed 's/^[+\-]//' | sed 's/\.[0-9]*$//' | uniq | wc -l)" -eq 2 ]]; then
```
alissapajer: This should be in a script in `sbt-slamdata` rather than copy-pasted across multiple repos.

djspiewak: It should and it will. Doing that takes a bit more restructuring though, because the sbt-slamdata scripts only work with publication right now, and also require decryption. So we need some new functionality in the plugin to enable what you're talking about.

PR #4029, opened 2018-12-19
Generalize Datasources impl to be more composable

wemrysi:


wemrysi:
This is preliminary work for ch3814, we needed a way to affect datasources during the initialization process. The end result is that `DatasourceManagement` was broken apart into more independent, composable pieces and error reporting was able to be converted into middleware.

It all ended up being a bit more far-reaching than anticipated, so giving it a separate PR from the ch3814 work.

nicflores:
@wemrysi, looks good. Approving and merging shortly.

PR #4031, opened 2019-01-03
Use Typecheck instead of Cond when possible

jsantos17:
This does not close ch 366 since this doesn't introduce filtering `MapFunc`s. It simply uses `Typecheck` instead of `Cond` in `FilterToCond` when possible. To showcase what this does, consider the following query generated by the frontend:

```sql
SELECT c.c1 AS a, c.c2 AS b, c.c3 AS c, c.c4 AS d FROM ( 
  SELECT p5.g7 AS c1, p5.g8  AS c2, p5.g9 AS c3, p5.g10 AS c4 
  FROM (SELECT
       (SELECT * FROM (SELECT ty.previous_addresses[_]{_} FROM `patients.data` AS ty) AS v WHERE type_of(v) = "string") AS g10, 
       (SELECT * FROM (SELECT ty.previous_addresses[_].city FROM   `patients.data` AS ty) AS v WHERE  type_of(v) = "string") AS g7, 
       (SELECT * FROM (SELECT ty.previous_addresses[_].state FROM   `patients.data` AS ty) AS v WHERE  type_of(v) = "string") AS g8, 
       (SELECT * FROM (SELECT ty.last_visit FROM   `patients.data` AS ty) AS v WHERE  type_of(v) = "string") AS g9 FROM  `patients.data` AS r6) AS p5) AS c
```

[QScript after this patch](https://gist.github.com/jsantos17/69fe7659edea36445a05b6fa3f5630f6).
[QScript before this patch](https://gist.github.com/jsantos17/3b941f5c3001beed4f8ce6be22d2847d)

Work on the filtering `MapFunc`s (and related frontend changes) are still on the pipeline.

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/minimizers/FilterToCond.scala
+++ b/qsu/src/main/scala/quasar/qsu/minimizers/FilterToCond.scala
@@ -17,21 +17,24 @@
 package quasar.qsu
 package minimizers
 
-import slamdata.Predef._
+import slamdata.Predef.{Map => SMap, _}
 
-import quasar.RenderTreeT
+import quasar.{RenderTreeT, Type}
 import quasar.common.effect.NameGenerator
 import quasar.contrib.iota._
 import quasar.contrib.matryoshka._
+import quasar.ejson.{CommonEJson, Str}
 import quasar.ejson.implicits._
 import quasar.fp._
-import quasar.qscript.{construction, HoleF, MonadPlannerErr, RecFreeS}, RecFreeS._
+import quasar.qscript.{construction, HoleF, MFC, MonadPlannerErr, RecFreeS}, RecFreeS._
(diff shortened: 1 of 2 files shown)


djspiewak:
Oh slick. What are the limitations of this? Is this going to hit every practical frontend query? We can get 366 out of de facto skull status if that's the case.

jsantos17:
Yes, all type filters generated by the frontend will hit this. We might have some problems with [this part](https://gist.github.com/jsantos17/69fe7659edea36445a05b6fa3f5630f6#file-after-typecheck-patch-txt-L84). I remember @wemrysi explained to me simplifying this is not valid in general. 

I'm guessing we'll also need to write the code to convert this QScript to `ParseInstruction`s?

beckyconning:
[ch4467]

PR #4032, opened 2019-01-04
Prefix Resources

wemrysi:
Introduces the ability to query a prefix path (a path that doesn't refer to a leaf resource) resulting in the concatenation of all leaf resources having the prefix.

[ch3814]

### DIFF
--- a/impl/src/main/scala/quasar/impl/datasource/package.scala
+++ b/impl/src/main/scala/quasar/impl/datasource/package.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2014–2018 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.impl
+
+import quasar.api.resource.ResourcePath
+
+import scala.util.Either
+
+import fs2.Stream
+
+package object datasource {
+  type AggregateResult[F[_], R] = Either[R, Stream[F, (ResourcePath, R)]]
+}
--- a/impl/src/main/scala/quasar/impl/datasource/package.scala
+++ b/impl/src/main/scala/quasar/impl/datasource/package.scala
@@ -23,5 +23,6 @@ import scala.util.Either
 import fs2.Stream
 
 package object datasource {
-  type AggregateResult[F[_], R] = Either[R, Stream[F, (ResourcePath, R)]]
+  type AggregateResult[F[_], A] = Stream[F, (ResourcePath, A)]
+  type CompositeResult[F[_], A] = Either[A, AggregateResult[F, A]]
 }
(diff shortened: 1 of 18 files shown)


alissapajer:
@wemrysi any SDBE or frontend changes that will be required to get this into production?

wemrysi:
@alissapajer there are mimir changes needed (ch4298) and possibly some frontend changes, depending on whether they've implemented support for prefix resources in discovery, but the api hasn't changed (we've always been able to report their existence as a datasource could support them directly).

impl/src/main/scala/quasar/impl/datasource/package.scala
```diff
@@ -0,0 +1,27 @@
...
+package quasar.impl
+
+import quasar.api.resource.ResourcePath
+
+import scala.util.Either
+
+import fs2.Stream
+
+package object datasource {
+  type AggregateResult[F[_], R] = Either[R, Stream[F, (ResourcePath, R)]]
```
alissapajer: When is the lhs side of the either used?

impl/src/test/scala/quasar/impl/datasource/local/LocalDatasourceSpec.scala
```diff
@@ -51,12 +51,6 @@ abstract class LocalDatasourceSpec
 
   def gatherMultiple[A](g: Stream[IO, A]) = g.compile.toList
 
-  "listing a file path returns none" >>* {
```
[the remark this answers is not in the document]
wemrysi: The updated `DatasourceSpec` has one, it was underspecified before.

impl/src/main/scala/quasar/impl/datasource/package.scala
alissapajer: I think that would be more clear. :+1: 

And if the `Either` is used in a few places, you could make a second alias for it named something like `ResultOrAggregateResult`, if that were to help with clarity.

alissapajer: Oh also, any reason not to use `scalaz.\/` instead of `scala.util.Either`?

alissapajer: (or the cats version)

alissapajer:
The ch story specifies that we are returning results in the form
```
{ "sourceName": String, "value": <row from source resource> }
```
If someone selected a prefix path `"a"` which has children `"a0"` and `"a1"` then would `"a0"` or `"a1"` be the `"sourceName"` (depending on the origin of the result row)? What if a row has both children as its origin (like in join)? 

Where/how do we keep track of the prefix path names so that they can be encoded in mimir when we return results?

Also thinking forward, we'll want to be able to hand these names to tectonic in situations when we can bypass mimir entirely.

impl/src/main/scala/quasar/impl/datasource/package.scala
djspiewak: Cats doesn't have an `Either` replacement since `scala.util.Either`, as of Scala 2.12, actually forms a default right-biased monad. So there isn't really a reason to use `scalaz.\/` anymore.

alissapajer:
@wemrysi cool, once the `AggregateResult` type is changed, this is good to go! I'll approve now in case you want to merge this and make that change separately; either way is fine with me.

wemrysi:
@alissapajer Thanks! I'll go ahead and make the change in this PR and merge once it builds.

djspiewak:
@alissapajer @wemrysi Hold the merge until after #4033. I just want to make sure I can build a release based on quasar v100.0.x, and I'd ideally like shift coalescence to be in that release so we can test the effect on some of @beckyconning's horrible tables.

alissapajer:
@djspiewak was your PR approval signaling that this can be merge now?

djspiewak:
@alissapajer Yep

PR #4033, opened 2019-01-04
Shift coalescence associativity.

djspiewak:
Normally I would assign this to @jsantos17, but he's out and I'd like to get it in as soon as possible.

What this change does is build a recursive associativity rose trie grouping shifts at each level by parentage, and then subsequently by the linearized, de-`Guard`ed `FreeMap` which makes up the `struct`. We should probably adjust `elideGuards` to take `Typecheck` into account at some point, but I think it's probably still fine.

Once shift chains have been grouped together, the depth of the grouping check is incremented so we move *down* the chain (which is to say *up* the tree). We then perform the same operation at the next level. The concept behind this is you want to make sure that shifts are associated together at the top level if possible, but then if the next level separates them, you need to split them apart. Take `a[*].b[*], a[*].c[*], d[*], a[*].c[*][*]` as an example. You need to group the `a`s at the top-most level, and then group the `c`s at the second level, but then even the `c`s split apart eventually (since one goes to greater depth). This recursion is what was missing from the previous reordering solution.

[ch1473]

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
+++ b/qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
@@ -786,7 +786,7 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
         val parentGroups = wrapped groupBy {
           case (ConsecutiveBounded(_, shifts), _) =>
             // TODO this is pretty darn inefficient
-            shifts.index(depth - 1).map(_.bimap(_.source.root, _.source.root))
+            shifts.index(shifts.length - depth).map(_.bimap(_.source.root, _.source.root))
 
           case _ => None
         }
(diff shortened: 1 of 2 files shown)


qsu/src/main/scala/quasar/qsu/minimizers/CollapseShifts.scala
```diff
@@ -831,8 +759,72 @@ final class CollapseShifts[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT] pr
...
+      } else {
+        val parentGroups = wrapped groupBy {
+          case (ConsecutiveBounded(_, shifts), _) =>
+            // TODO this is pretty darn inefficient
+            shifts.index(shifts.length - depth).map(_.bimap(_.source.root, _.source.root))
+
+          case _ => None
+        }
+
+        // this branch is to keep the tests happy, really
```
wemrysi: :cry:

PR #4041, opened 2019-01-10
Add type filtering MapFuncs

jsantos17:
[ch366]

I have verified frontend queries can be expressed much more succinctly through this functions. The only exception is filters, which hit 2136 when expressed uses these functions instead of `FilterToCond`.

For example, we can rewrite:

```sql
SELECT c.c1 AS A,
       c.c2 AS B,
       c.c3 AS C
FROM
  (SELECT p5.g7 AS c1,
          p5.g8 AS c2,
          p5.g9 AS c3
   FROM
     (SELECT
        (SELECT *
         FROM
           (SELECT ty.previous_addresses[_].city
            FROM `patients.data` AS ty) AS v
         WHERE type_of(v) = "string") AS g7,

        (SELECT *
         FROM
           (SELECT ty.previous_addresses[_].state
            FROM `patients.data` AS ty) AS v
         WHERE type_of(v) = "string") AS g8,

        (SELECT *
         FROM
           (SELECT ty.last_visit
            FROM `patients.data` AS ty) AS v
         WHERE type_of(v) = "string") AS g9
      FROM `patients.data` AS r6) AS p5) AS c
```

into

```sql
SELECT
   as_string(previous_addresses[_].city) AS A,
   as_string(previous_addresses[_].state) AS B,
   as_string(last_visit) AS C FROM `patients.data`
```

These produce the same QScript now.

This PR introduces five functions instead of one since all of our [function definition machinery](https://github.com/slamdata/quasar/blob/master/frontend/src/main/scala/quasar/functions.scala) assumes we want all function parameters to be recursive positions. This is of course undesirable in the case of a type-filtering function. i.e we don't want to allow type-filtering using computed types. The other options is modifying the parser to introduce syntax that enforces one of the parameters is a type.

I'd appreciate opinions on the names of the functions. I'm not sure about `DimensionalEffect`, but `Sifting` seemed correct.


jsantos17:
`feature` instead of `breaking` since the queries generated by the frontend before this change are still valid.

jsantos17:
@wemrysi 

I think I'm going with `ensure_<type>`.  

With regard to `DimensionalEffect`, is it `Mapping` as in "mapping to undefined" in case it's the incorrect type?

wemrysi:
@jsantos17 It is `Mapping` in the functor sense, it can modify values but not alter the dimensionality of the dataset.

qscript/src/main/scala/quasar/qscript/MapFunc.scala
```diff
@@ -84,6 +84,11 @@ object MapFunc {
     case string.Number => a => MFC(C.Number(a))
     case string.Null => a => MFC(C.Null(a))
     case string.ToString => a => MFC(C.ToString(a))
+    case structural.AsNumber => a => MFD(D.Typecheck(a, Type.Numeric))
+    case structural.AsString => a => MFD(D.Typecheck(a, Type.Str))
+    case structural.AsBoolean => a => MFD(D.Typecheck(a, Type.Bool))
+    case structural.AsDate => a => MFD(D.Typecheck(a, Type.DateTime))
```
rintcius: How about keeping this one-to-one with the types? I.e.

```
case structural.EnsureOffsetDateTime => a => MFD(D.Typecheck(a, Type. OffsetDateTime))
case structural.EnsureOffsetTime => a => MFD(D.Typecheck(a, Type. OffsetTime))
... same for other date/time types
```
 
Also good to keep the function names consistent, in the form of `ensure_<typeName>` with the type names how they appear in `type_of(_) = "<typeName>"`, see e.g. https://github.com/slamdata/quasar/blob/ad12fbea192708cdc5f9cdb7fe7ec6a4a0a71ea6/frontend/src/test/scala/quasar/std/StdLibSpec.scala#L144-L181

jsantos17: These types were chosen because those are the ones used by the frontend and I didn't want to introduce a `MapFunc` for every type (11 in total). [I took the types from here](https://github.com/slamdata/slamx/blob/master/src/SlamX/Model/SST.purs#L78). 

The `Coproduct(OffsetDateTime, LocalDateTime)` was introduced since the frontend is looking for `DateTime`s in general, rather than `OffsetDateTime`. In the case of `ensure_number`, we need to keep in mind `type_of` yields `number` for both `Type.Dec` and `Type.Int`. I introduced the `Number` union to preserve this behavior. In general, I leaned toward the more liberal interpretation of what a user might mean when type-filtering for `DateTime`s or numbers. 

I didn't want to introduce eleven functions, especially since we would ideally implement this as a single function. We also need to keep in mind we'd need an extra function for each union we need to define (i.e `Number` or `DateTime`)

jsantos17:
I have updated this PR to use  `ensure_` as a prefix, rather than `as_`

qscript/src/main/scala/quasar/qscript/MapFunc.scala
rintcius: I'm fine with limiting the amount of types to what's used in the frontend. My comment was also not about `number`, that's perfectly fine too. 

I still think we should expose `ensure_ offsetdatetime` instead of `ensure_datetime`, containing only `OffsetDatetime`. I think this is what frontend is able to work with, not `DateTime`s in general.

jsantos17: Ohh I misunderstood then. I'll fix those!

rintcius:
@jsantos17 Hmm, that's still not what I meant. I think it's important that the exposed function names are consistent with the names of the (also exposed) type names, i.e. the strings (not the scalars) you referred to here https://github.com/slamdata/slamx/blob/master/src/SlamX/Model/SST.purs#L78-L84 ), so       `ensure_number`, `ensure_string`, `ensure_boolean`, `ensure_offsetdatetime` (note: no underscore between offset and datetime) and `ensure_null`.

Sorry, if this comes across as too picky, normally I don't care too much about naming but with these function names it is different because they are exposed as part of `sql2` and are very hard to change later.

jsantos17:
Ideally this would be a single function  `typecheck(<value>, <type literal>)` or syntax `<value> :: <type literal>`. Note the second parameter would be a type literal and not a string. For the same  reason type annotations in Scala (or any other language) aren't strings. Because of this I do not believe we want `type_of` to set the gold standard for type names. 

>I think it's important that the exposed function names are consistent with the names of the (also exposed) type names

I'm going to prefix all of these with `_sd` as to make it clear these aren't function intended for users and might go away when we add type-filters and type literals to the parser. These aren't meant to be "exposed" in the same way as other `MapFuncs` are. I think of this as a private API that'll go away soon.

But anyway, I'm eager to get this merged ASAP. What do you think of this?

- `_sd_ensure_number`
- `_sd_ensure_string`
- `_sd_ensure_boolean`
- `_sd_ensure_offsetdatetime`
- `_sd_ensure_null`

core/src/main/scala/quasar/compile/Compiler.scala
```diff
@@ -204,75 +204,80 @@ final class Compiler[M[_], T: Equal]
...
+      CIName("make_map")                  -> structural.MakeMap,
+      CIName("make_array")                -> structural.MakeArray,
+      CIName("map_concat")                -> structural.MapConcat,
+      CIName("array_concat")              -> structural.ArrayConcat,
+      CIName("delete_key")                -> structural.DeleteKey,
+      CIName("contains_key")              -> structural.ContainsKey,
+      CIName("_sd_ensure_number")         -> structural.EnsureNumber,
+      CIName("_sd_ensure_string")         -> structural.EnsureString,
+      CIName("_sd_ensure_boolean")        -> structural.EnsureBoolean,
+      CIName("_sd_ensure_offsetdatetime") -> structural.EnsureOffsetDateTime,
```
jsantos17: Unfortunately I had to reindent this.

jsantos17:
@rintcius I have pushed a new commit with prefixed function names. Let me know if I got it right this time.

rintcius:
>Because of this I do not believe we want type_of to set the gold standard for type names.

Even in this case I'd prefer having consistent names, but we can discuss that later :)

I know postgres supports `CAST(value AS <type literal>)` and `value :: <type literal>` but I'm curious how/if type literals would fit in with standard sql. I think I wouldn't mind a dynamic `typecheck(<value>, <value>)` where the 2nd arg is as dynamic as the first arg, but that may also be something to discuss later.

PR #4044, opened 2019-01-14
Fix handling of DimEdit in ApplyProvenance to account for branching

alissapajer:
[ch2068]
[ch3691]
[ch3999]
[ch3553]
[ch672]

When rewriting a qsu graph with `rewriteM`, we begin at the `Read` (the parent) and then recurse into its children. If a node has multiple immediate children and one of those children is a `DimEdit`, in `ApplyProvenance` we rename the parent to have the name of the `DimEdit`. The bug was that we'd then delete the parent's dimensions since the parent had become orphaned (by its old name). This was a problem because when we'd recurse into the _other_ branch, we'd lookup the dimensions of the parent and find they were not there. This produced the `"Dependent dimensions not found"` error.

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
+++ b/qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
@@ -82,7 +82,6 @@ final class ApplyProvenance[T[_[_]]: BirecursiveT: EqualT: ShowT] private () ext
       case g @ Extractors.DimEdit(src, _) =>
         for {
           _ <- computeDims[X](g)
-          _ <- QAuthS[X].modify(_.supplant(src.root, g.root))
         } yield g.overwriteAtRoot(src.unfold map (_.root))
 
       case g @ Extractors.Transpose(src, retain, rot) =>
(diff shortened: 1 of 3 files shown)


alissapajer:
https://github.com/slamdata/mimir/pull/76 is open with the integration test for ch2068.

I did not verify any of the stories tagged as duplicates of ch2068 as they were complex. QA will verify those and if one of them is not fixed we'll fix it separately.

alissapajer:
@wemrysi I'm going to additionally add a test to `ApplyProvenanceSpec` for the graph
```
    Read
      |
  Transpose
    /     \
DimEdit    Map
  |         |
QSFilter    | 
  \        /
  AutoJoin2
      |
    DimEdit
```
which is the graph I used to track down the bug. The originating query (a minimization of ch2068) is
```sql
SELECT
  (SELECT t4{_} AS DateTime FROM `real-giraffe.json` as t4),
  (SELECT * FROM (SELECT t2{_} FROM `real-giraffe.json` AS t2) AS ot2 WHERE ot2 = "S") AS S
FROM
  `real-giraffe.json`
```

PR #4047, opened 2019-01-16
Add test cases for multi line searching to StdLibSpec

rintcius:
Part of ch895


djspiewak:
I think this should be **breaking**, no?

rintcius:
Hmm, that's true... interesting

PR #4051, opened 2019-01-22
Single field object select

jsantos17:
[ch63]

I have reviewed code generation for the frontend and from what I've seen, it should be unaffected by these changes. The frontend generates names (aliases) for all projections, so single field projections should be already wrapped.

Out of an abundance of caution, I'd rather wait after the Feb 1 release before merging this.


jsantos17:
I'd like to close [ch3682] too, since it was made irrelevant by #4041

alissapajer:
@jsantos17 :+1: on waiting to merge this until after release.

I'm working on some things that are needed for the release, so I might end up delaying review of this for a couple days, if that's ok.

jsantos17:
@alissapajer sounds good. I'll keep the bitrot in check meanwhile.

alissapajer:
@jsantos17 could you merge in master when you have a chance? There are no conflicts, but I want to get a green build against current master.

jsantos17:
@alissapajer Will do!

alissapajer:
@jsantos17 what column names will be generated for the queries
```sql
select foo, foo{*} from bar
select foo{*}, foo[*] from bar
```
Currently for the query
```sql
select foo, foo, foo from bar
```
we generate names `foo`, foo0`, and `foo1. Could we add a couple unit tests showing the names generate in the above cases?

jsantos17:
@alissapajer we do have those tests, but none including shifts/flattens. [I added one here](https://github.com/slamdata/quasar/pull/4051/files#diff-e950b675c95b2d679f94ed86901677a6R1764).

PR #4059, opened 2019-01-23
Fixed collapsing compatible shifts below the top tier

djspiewak:
Also note the extra change to the `QScriptRegressionSpec`: even before @wemrysi's rewrite lands, there are several queries which we can now push into tectonic which we previously could not.

[ch4703]


run/src/test/scala/quasar/QScriptRegressionSpec.scala
```diff
@@ -214,7 +214,17 @@ object QScriptRegressionSpec extends Qspec {
 
         result must countReadAs(0)
         result must countInterpretedReadAs(1)
-        result must countLeftShiftAs(2)
+        result must countLeftShiftAs(0)
```
wemrysi: :tada:

PR #4062, opened 2019-01-25
Change and extend ResourceError

rintcius:
Part of ch3775


rintcius:
For review but not yet for merge

PR #4063, opened 2019-01-25
Prereqs for cartesian rewrite (take 2)

wemrysi:
The remainder after quotienting out the cartesian rewrite.

### DIFF
--- a/frontend/src/test/scala/quasar/ParseInstructionSpec.scala
+++ b/frontend/src/test/scala/quasar/ParseInstructionSpec.scala
@@ -1262,6 +1262,46 @@ object ParseInstructionSpec {
 
         input must pivotInto(".shifted", IdStatus.IncludeId, ParseType.Object)(expected)
       }
+
+      "preserve empty arrays as values of an array pivot" in {
+        val input = ldjson("""
+          [ 1, "two", [] ]
+          [ [] ]
+          [ [], 3, "four" ]
+          """)
+
+        val expected = ldjson("""
+          1
+          "two"
+          []
+          []
+          []
+          3
+          "four"
+        """)
+
+        input must pivotInto(".", IdStatus.ExcludeId, ParseType.Array)(expected)
+      }
+
+      "preserve empty objects as values of an object pivot" in {
+        val input = ldjson("""
+          { "1": 1, "2": "two", "3": {} }
+          { "4": {} }
+          { "5": {}, "6": 3, "7": "four" }
+          """)
+
+        val expected = ldjson("""
+          1
+          "two"
+          {}
+          {}
+          {}
+          3
+          "four"
+        """)
+
+        input must pivotInto(".", IdStatus.ExcludeId, ParseType.Object)(expected)
+      }
     }
 
     def evalPivot(pivot: Pivot, stream: JsonStream): JsonStream
--- a/qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
+++ b/qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
@@ -266,9 +266,17 @@ object MapFuncCore {
       : CoMapFuncR[T, A] => CoMapFuncR[T, A] =
     repeatedly(rewrite[T, A])
 
-  def normalized[T[_[_]]: BirecursiveT: EqualT, A: Equal: Show](fm: Free[MapFunc[T, ?], A])
-      : Free[MapFunc[T, ?], A] =
-    fm.transCata[Free[MapFunc[T, ?], A]](normalize[T, A])
+  def normalized[T[_[_]]: BirecursiveT: EqualT, A: Equal: Show](
+      fm: FreeMapA[T, A])
+      : FreeMapA[T, A] = {
+
+    def norm(fa: FreeMapA[T, A]): Option[FreeMapA[T, A]] = {
+      val next = fa.transCata[Free[MapFunc[T, ?], A]](normalize[T, A])
+      (next =/= fa) option next
+    }
+
+    repeatedly(norm)(fm)
+  }
 
   private def rewrite[T[_[_]]: BirecursiveT: EqualT, A: Equal]:
       CoMapFuncR[T, A] => Option[CoMapFuncR[T, A]] =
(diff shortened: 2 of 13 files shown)


wemrysi:
Snapshot published as `106.0.2-0beacdd`.

qscript/src/test/scala/quasar/qscript/MapFuncCoreSpec.scala
```diff
@@ -284,6 +284,48 @@ final class MapFuncCoreSpec extends Qspec with TTypes[Fix] with TreeMatchers {
       normalize(expr) must beTreeEqual(expect)
     }
 
+    "project key from consequent when projecting from a Cond with an undefined alternate" >> {
```
alissapajer: What motivated the `Cond`-related changes?

run/src/test/scala/quasar/RegressionQScriptEvaluator.scala
```diff
@@ -59,8 +59,8 @@ final class RegressionQScriptEvaluator[
 
   def QSMFunctor: Functor[QSM] = Functor[QSM]
 
-  def optimize: F[QScriptNormalized[T, T[QSM]] => QSM[T[QSM]]] =
-    RewritePushdown[T, QSM, QScriptNormalized[T, ?], ResourcePath].point[F]
+  def optimize(norm: T[QScriptNormalized[T, ?]]): F[T[QSM]] =
+    norm.transCata[T[QSM]](FocusedPushdown[T, QSM, QScriptNormalized[T, ?], ResourcePath]).point[F]
```
alissapajer: `QScriptRegressionSpec` isn't testing against the cartesian rewrite. If/when we move the other pushdown rewrites to thoth, we can move this there.

qscript/src/test/scala/quasar/qscript/MapFuncCoreSpec.scala
wemrysi: We were producing `QScript` like
```
      │  │  ╰─ ProjectKey
      │  │     ├─ ProjectKey
      │  │     │  ├─ ProjectKey
      │  │     │  │  ├─ ProjectKey
      │  │     │  │  │  ├─ Cond
      │  │     │  │  │  │  ├─ Or
      │  │     │  │  │  │  │  ├─ Eq
      │  │     │  │  │  │  │  │  ├─ ProjectKey
      │  │     │  │  │  │  │  │  │  ├─ LeftSide
      │  │     │  │  │  │  │  │  │  ╰─ Constant(Str(cartouche0))
      │  │     │  │  │  │  │  │  ╰─ ProjectIndex
      │  │     │  │  │  │  │  │     ├─ RightSide
      │  │     │  │  │  │  │  │     ╰─ Constant(Int(0))
      │  │     │  │  │  │  │  ╰─ IfUndefined
      │  │     │  │  │  │  │     ├─ ProjectIndex
      │  │     │  │  │  │  │     │  ├─ RightSide
      │  │     │  │  │  │  │     │  ╰─ Constant(Int(0))
      │  │     │  │  │  │  │     ╰─ Constant(Bool(true))
      │  │     │  │  │  │  ├─ ConcatMaps
      │  │     │  │  │  │  │  ├─ LeftSide
      │  │     │  │  │  │  │  ╰─ MakeMap
      │  │     │  │  │  │  │     ├─ Constant(Str(1))
      │  │     │  │  │  │  │     ╰─ ProjectIndex
      │  │     │  │  │  │  │        ├─ RightSide
      │  │     │  │  │  │  │        ╰─ Constant(Int(1))
      │  │     │  │  │  │  ╰─ Undefined
      │  │     │  │  │  ╰─ Constant(Str(original))
      │  │     │  │  ╰─ Constant(Str(left))
      │  │     │  ╰─ Constant(Str(left))
      │  │     ╰─ Constant(Str(0))
```
which, aside from being suboptimal, made it hard to determine whe the final series of projections from `LeftSide`/`RightSide` were.

frontend/src/test/scala/quasar/ParseInstructionSpec.scala
```diff
@@ -1262,6 +1262,46 @@ object ParseInstructionSpec {
 
         input must pivotInto(".shifted", IdStatus.IncludeId, ParseType.Object)(expected)
       }
+
+      "preserve empty arrays as values of an array pivot" in {
```
alissapajer: :+1:

qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -266,9 +266,17 @@ object MapFuncCore {
...
-  def normalized[T[_[_]]: BirecursiveT: EqualT, A: Equal: Show](fm: Free[MapFunc[T, ?], A])
-      : Free[MapFunc[T, ?], A] =
-    fm.transCata[Free[MapFunc[T, ?], A]](normalize[T, A])
+  def normalized[T[_[_]]: BirecursiveT: EqualT, A: Equal: Show](
+      fm: FreeMapA[T, A])
+      : FreeMapA[T, A] = {
+
+    def norm(fa: FreeMapA[T, A]): Option[FreeMapA[T, A]] = {
+      val next = fa.transCata[Free[MapFunc[T, ?], A]](normalize[T, A])
+      (next =/= fa) option next
```
alissapajer: We've had discussions about equality against `FreeMap` in the past... anyone recall what the caveats were?

djspiewak: It depends on how the `FreeMap` has been constructed. We were never able to fully narrow it down, but it appears to be something where the structure is, in one case, constructed directly (e.g. via `embed`), while in the alternative case derived via `flatMap`-based transformations. My guess is that the embedded coyoneda messes up something about how `CoEnv` works, but that's hard to verify.

wemrysi: It's a bit sketchy, we could change to using a boolean state value that gets set to `true` if the rewrite ever changes any layer and use that to decide whether to repeat or not.

alissapajer: So, that adds an additional contract that a single layer changing is enough to change the output. Meaning that if you change one layer and then change another you must guarantee that the output cannot stay the same. I haven't completely convinced myself yet that this would be okay.

Also, there is a `repeatedly` above on line 267; do we still need that one?

wemrysi:
Published as `106.0.2-c1332c1`

PR #4066, opened 2019-01-30
Take the entire provenance chain into account

djspiewak:


djspiewak:
[ch4890]

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -1951,6 +1951,47 @@ object MinimizeAutoJoinsSpec
 
       runOn(qgraph) must haveShiftCount(2)
     }
+
+    // a[*][*], b[*][*]
```
jsantos17: I find it distressing we didn't have any tests to catch this.

djspiewak: Me too! I'm trying not to think about it…

PR #4068, opened 2019-01-31
parallel RValue Cartesian interpreter

alissapajer:
[ch4888]

### DIFF
--- a/impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala
+++ b/impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala
@@ -18,7 +18,7 @@ package quasar.impl.evaluate
 
 import slamdata.Predef._
 
-import cats.effect.Sync
+import cats.effect.Concurrent
 import cats.effect.concurrent.Ref
 
 import quasar.{ParseInstruction, ParseType}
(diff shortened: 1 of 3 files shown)


djspiewak:
Actually **breaking** since it removes an old API.

alissapajer:
@djspiewak btw as you may have noticed, I changed all the names to be the same as the names you used 1) because they are nice names and 2) to remove the overhead of translating between the two places :)

djspiewak:
@alissapajer I noticed and I definitely appreciate it. :-)

impl/src/main/scala/quasar/impl/evaluate/RValueParseInstructionInterpreter.scala
```diff
@@ -179,37 +187,82 @@ object RValueParseInstructionInterpreter {
...
+              math.ceil(headRvs.length.toDouble / numberOfJobs.toDouble).toInt
+
+            val headChunks: List[Array[RValue]] =
+              headRvs.grouped(chunkSize).toList
+
+            val chunks: List[List[(String, Array[RValue])]] = 
+              headChunks.map(chunk => ((headName, chunk)) :: tail)
+
+            chunks.traverse(chunk =>
+              Concurrent[F].start(cross(chunk))).flatMap(_.traverseM(_.join))
```
djspiewak: Minor note, but this could have been `cross(chunk).start` rather than `Concurrent[F].start(cross(chunk))`

PR #4069, opened 2019-02-01
Remove join conditions

jsantos17:
[ch4772]
[ch5044]

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
+++ b/qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
@@ -145,20 +145,6 @@ final class ExpandShifts[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes
       tempShift <- QSUGraph.withName[T, G](NamePrefix)(tempShiftPat)
       commonShift = tempShift.overwriteAtRoot(tempShiftPat.copy(source = commonRoot))
 
-      joinRepair = sameFocus map {
-        case (sym, _, _) =>
-          func.Cond(
-            func.Or(
-              func.Eq(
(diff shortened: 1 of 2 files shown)


qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
```diff
@@ -145,20 +145,6 @@ final class ExpandShifts[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes
       tempShift <- QSUGraph.withName[T, G](NamePrefix)(tempShiftPat)
       commonShift = tempShift.overwriteAtRoot(tempShiftPat.copy(source = commonRoot))
 
-      joinRepair = sameFocus map {
```
djspiewak: I think we can also remove `sameFocus`.

jsantos17: I think we use `sameFocus` for `newDims`. `newDims` is in turned used for `joinDims`, [which is used for side effects](https://github.com/slamdata/quasar/pull/4069/files#diff-1fd1211909c6e5f87e15b209080143caR160). Are the side effects no longer relevant?

djspiewak: Ah that's an interesting question. In theory, removing the join condition without changing that line is going to result in the wrong provenance, I think.

jsantos17:
Build failure is a flaky test. Restarting.

qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
wemrysi: We have to compute the provenance, but we no longer need `sameFocus` to do so. So we can get rid of it and just use the `None` branch on L154 to compute provenance.

PR #4072, opened 2019-02-02
Incremenal RValue parse instruction interpreter

wemrysi:
[ch4995]


wemrysi:
5b310bb is the semantic changes, 7bb4d52 is some code reorg so I recommend only viewing the diff for the first commit.

alissapajer:
@wemrysi have you run the mimir `RValue`-backed integration tests against this change? If not, I think we can still merge and see what happens :) If something failed we'd have to iterate back here in either case.

wemrysi:
@alissapajer am doing so now! I'll push the fix for `chunksPerJob` if everything looks good.

wemrysi:
@alissapajer local run of the `it` tests were green :+1:

PR #4075, opened 2019-02-04
Remove join-condition related side effects

jsantos17:
Not sure if we still need the `MonadCI` constraint since we no longer call `get` on it anywhere. Shall I remove that too?

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
+++ b/qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
@@ -47,9 +47,7 @@ final class ExpandShifts[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes
   private val json = Fixed[T[EJson]]
   private type P = prov.P
   private type QAuthS[F[_]] = MonadState_[F, QAuth]
-  private type MonadCI[F[_]] = MonadState_[F, CompatInfo]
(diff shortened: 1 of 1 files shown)


qsu/src/main/scala/quasar/qsu/ExpandShifts.scala
```diff
@@ -136,36 +133,23 @@ final class ExpandShifts[T[_[_]]: BirecursiveT: EqualT: ShowT] extends QSUTTypes
...
-      compatInfo <- MonadState_[G, CompatInfo].get
-
-      sameFocus = compatInfo findLeft {
-        case (_, rot, qds) => (rotation ≟ rot) && (structDim ≟ qds)
-      }
-
       tempShift <- QSUGraph.withName[T, G](NamePrefix)(tempShiftPat)
       commonShift = tempShift.overwriteAtRoot(tempShiftPat.copy(source = commonRoot))
 
       srcDims <- dimsOf(src.root)
```
wemrysi: This doesn't appear to be referenced anywhere, can be removed.

jsantos17:
@djspiewak why the 🛑label?

djspiewak:
@jsantos17 I'm like, vetoing merge on anything that isn't strictly related to the release tomorrow morning, just to stamp down on all the variables. I don't think there's any problems with this whatsoever, but even the tiniest of risks isn't worth it right now.

PR #4077, opened 2019-02-04
Use the correct parent for determining struct provenance

djspiewak:
It's kind of distressing that we didn't identify this bug before.

[ch5156]
[ch5152]


qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -2202,8 +2203,8 @@ object MinimizeAutoJoinsSpec
           Map(
             LeftShift(
               LeftShift(
-                MultiLeftShift(
-                  MultiLeftShift(
+                LeftShift(
```
wemrysi: :tada:

PR #4079, opened 2019-02-06
Simplify parse instructions

wemrysi:
* Add an `IdStatus` property to `InterpretedRead`
* Remove `ParseInstruction.Ids`
* Remove the `CPath` property from `Pivot`.

[ch4489]
[ch5100]


cryogenian:
Maybe `QueryResult` should have id status too.

wemrysi:
@cryogenian Good call on `Wrap` I'll omit the path from there also. And yes, `QueryResult` should have an `IdStatus` somewhere.

I actually think I'm going to close this PR to make a few more changes that will unify things between `QueryResult` and `InterpretedRead` a bit as well as take this opportunity in breaking things to rename `ParseInstruction`.

PR #4081, opened 2019-02-08
Replace ParseInstructions with ScalarStages

wemrysi:
Renames `ParseInstruction` to `ScalarStage` and introduces `ScalarStages` to describe a sequence row-level transformations on a dataset.

Also simplifies various scalar stages, removing superfluous properties where appropriate.

[ch5100]
[ch4489]


wemrysi:
Published as `114.0.0-709b622`.

PR #4087, opened 2019-02-13
Delete incorrect Eq MapFuncCore normalization case

alissapajer:


qscript/src/main/scala/quasar/qscript/MapFuncCore.scala
```diff
@@ -297,11 +297,6 @@ object MapFuncCore {
 
       case Or(BoolLit(true), _) => some(BoolLit[T, A](true).project)
 
-      case Eq(v1, v2) if v1 ≟ v2 => some(BoolLit[T, A](true).project)
-
-      case Eq(ExtractFunc(Constant(v1)), ExtractFunc(Constant(v2))) =>
-        some(BoolLit[T, A](v1 ≟ v2).project)
```
djspiewak: This one is actually sound, but we may as well let it fall through to mimir. I don't think there are any advantages to eliding it.

alissapajer: Agreed, I meant to make a comment on this one. :+1: But that was my reasoning as well.

alissapajer: We could also remove the other constant value checks from this function, like boolean computations and whatnot.

PR #4089, opened 2019-02-14
Remove focused scalar stage pushdown rewrites

wemrysi:


wemrysi:
This is the quasar side of ch5349.

wemrysi:
Published as `117.0.1-57fd889`

run/src/test/scala/quasar/QScriptRegressionSpec.scala
```diff
@@ -16,44 +16,19 @@
...
-    Sql2QueryEvaluator[Fix, IO, QScriptCount](
-      new RegressionQueryEvaluator[Fix, IO])
-
-  def query(q: String): SqlQuery =
-    SqlQuery(Query(q), Variables(SMap()), Path.rootDir)
+import shims._
 
-  def count(sql2: String): QScriptCount =
-    sql2Evaluator.evaluate(query(sql2)).unsafeRunSync()
+object QScriptRegressionSpec extends CountingRegressionSpec {
```
alissapajer: I like the solution to have this spec in addition to `ScalarPushdownRegressionSpec` over in thoth :+1:

wemrysi: Yeah, at first I moved them all to thoth, but realized we want these here to catch `QSU` problems, I figure they can evolve independently, though we might have examples on the same query in both for more complex/important scenarios.

PR #4090, opened 2019-02-16
Default provenance for MapFunc should be conjunction of inputs

wemrysi:


alissapajer:
@wemrysi test failure in `MinimizeAutoJoinsSpec`

wemrysi:
@alissapajer On it.

qsu/src/test/scala/quasar/qsu/ApplyProvenanceSpec.scala
```diff
@@ -143,10 +143,14 @@ object ApplyProvenanceSpec extends Qspec with QSUTTypes[Fix] {
       tree must haveDimensions(SMap(
         'n0 -> Dimensions.origin(
           P.thenn(
-            P.value(IdAccess.identity('n2)),
+            P.thenn(
+              P.prjValue(J.str("bar")),
+              P.value(IdAccess.identity('n2))),
```
alissapajer: @wemrysi is this (or does this PR affect) the query that we had discussed previously that might have been missing project provenance? You mentioned it in chat at the end of last week, but I can't find reference to which query it actually was.

wemrysi: @alissapajer Yep, this is it. Previously, we'd punt on provenance for `Add` so the `.bar` was missing, now it is present.

wemrysi:
@alissapajer ready for another look

qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
```diff
@@ -278,15 +275,38 @@ final class ApplyProvenance[T[_[_]]: BirecursiveT: EqualT: ShowT] private () ext
...
+
+          val fheads = computeFuncDims(fm.linearize)(κ(Dimensions(heads)))
+
+          def tailNotFound(i: Int): F[NonEmptyList[dims.P]] =
+            MonadPlannerErr[F].raiseError(
+              PlannerError.InternalError.fromMsg(
+                s"Tail with index '$i' not found in '${tails.shows}'"))
+
+          fheads.fold(sdims.point[F])(Dimensions.join.modifyF[F] {
+            case NonEmptyList(h, ICons(dims.prov.prjPath(ejs.int(i)), INil())) =>
```
alissapajer: Why store the index `i` as a component of provenance when you're going to match on it here like this. Or is it used somewhere else too?

wemrysi: I needed a way to hand the synthesized dimensions to `computeFuncDims` and then be able to figure out which result came from which input in order to substitute the tails back in. Since I know nothing can `prjPath` in a `MapFunc`, this seemed like a decent way to do so.

PR #4093, opened 2019-02-21
Provenance fixes for autojoin

wemrysi:
* Flattens sequences prior to computing autojoin conditions so that flatten and shift are treated identically
* Introduces static identities to be able to represent autojoining an inflation with a projection in the join conditions


wemrysi:
Closing this as I'll open a new PR for the new provenance representation.

PR #4095, opened 2019-02-27
`ScalarStagesSpec` pending-configurable per spec and per implementation

alissapajer:
[ch5745]

### DIFF
--- a/frontend/src/test/scala/quasar/ScalarStageSpec.scala
+++ b/frontend/src/test/scala/quasar/ScalarStageSpec.scala
@@ -37,6 +37,11 @@ abstract class ScalarStageSpec
     with ScalarStageSpec.PivotSpec
     with ScalarStageSpec.CartesianSpec
 
+/*
+ * Test names must begin with the prefix specified in their
+ * `override def is` implementation followed by `-#*` in
+ * order for `pendingFragments` to come into effect.
+ */
 object ScalarStageSpec {
 
   def parseNumber(prefix: String, name: String): Option[Int] = {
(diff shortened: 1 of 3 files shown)


alissapajer:
This change will break upstream tests. We've made test-breaking changes **breaking** in the past. I'm open to arguments otherwise though.

frontend/src/test/scala/quasar/ScalarStageSpec.scala
```diff
@@ -1120,8 +1170,41 @@ object ScalarStageSpec {
 
         input must cartesianInto(targets)(expected)
       }
+
+      "cart-7 nested pivoting doesn't produce unnecessary empty fields" in {
```
alissapajer: @cryogenian I went ahead and added your example here.

frontend/src/test/scala/quasar/ScalarStageSpec.scala
```diff
@@ -21,23 +21,51 @@ import slamdata.Predef._
...
     with ScalarStageSpec.CartesianSpec
 
+/*
+ * Test names must begin with the prefix specified in their
+ * `override def is` implementation followed by `-#*` in
+ * order for `pendingFragments` to come into effect.
+ */
 object ScalarStageSpec {
 
+  def parseNumber(prefix: String, name: String): Option[Int] = {
```
alissapajer: At first I didn't have the `prefix` component, but because we mix in all the individual `traits` in `ScalarStageSpec`, it was somehow mutably combining the `override def is` methods, meaning that if you wrote `val idsPendingExamples: Set[Int] = Set(3)`, it would pending example `3` in _all_ the mixed in specs, not just in `IdsSpec`.

alissapajer: Also, I opted for fixing the name prefixes here rather than letting the implementations of the spec parse the test names themselves, because I wanted us to be able to change the test descriptions at will without having upstream tests be affected. And this gives us a nice way to refer to the tests, like "hey did you see mask-52?" "yeah that one's wild!"

PR #4096, opened 2019-02-28
scalar stages test

alissapajer:


frontend/src/test/scala/quasar/ScalarStageSpec.scala
```diff
@@ -1204,6 +1206,35 @@ object ScalarStageSpec {
 
         input must cartesianInto(targets)(expected)
       }
+
+      // a[_] as a, b[_] as b
+      "cart-8 pivoting retains row alignment through undefineds" in {
```
alissapajer: @djspiewak this was an example that you thought might fail in tectonic. It's currently passing in tectonic. Are there any variations you'd like added that you still think might fail?

alissapajer:
Adds an object version (`cart-8`) of the query for ch5371 (the example @cryogenian contrived (`cart-7`)  is actually the exact same example, just for arrays! :) )

PR #4100, opened 2019-03-06
Prepare for aggregated resource querying

rintcius:
For now aggregated resource querying does not return `source` yet. The child query results are simply concatenated. The code still has all the infrastructure though to be able to add the source in a subsequent PR.

Local datasource now excludes hidden files as resources. Otherwise `.DS_Store` prevents local datasource from working with aggregated resource querying.

[ch4298]


impl/src/main/scala/quasar/impl/datasources/CompositeResourceSchema.scala
```diff
@@ -77,16 +76,8 @@ final class CompositeResourceSchema[
...
-      val srcSst = sstQDataEncode.makeString(posixCodec.printPath(rp.toPath))
-
-      val assocs = List(SourceKey -> srcSst, ValueKey -> valSst)
-
-      sstQDataEncode.makeObject(
-        assocs.foldLeft(sstQDataEncode.prepObject) {
-          case (o, (k, v)) => sstQDataEncode.pushObject(k, v, o)
-        })
-    }
+    def attributeSource(rp: ResourcePath)(valSst: S): S =
```
wemrysi: I think it would be better to leave this as it was and just create a new resource schema impl that doesn't add any source/value structure.

impl/src/main/scala/quasar/impl/datasources/middleware/ChildAggregatingMiddleware.scala
```diff
@@ -28,23 +34,46 @@ import cats.syntax.functor._
 
 import fs2.Stream
 
+import matryoshka.data.Fix
+
 import shims._
 
 object ChildAggregatingMiddleware {
   def apply[T[_[_]], F[_]: Monad: MonadResourceErr, I, R](
+      sourceKey: String,
```
wemrysi: Similarly here, I'd just revert all these changes and leave the impl as it was.

djspiewak:
Shouldn't this be **breaking** since it changes the API of `ChildAggregatingDataSource`?

rintcius:
@djspiewak I was considering that to be internal to quasar

rintcius:
Closing this - see https://github.com/slamdata/quasar/pull/4102 for the new one

PR #4101, opened 2019-03-07
more scalar stages tests

alissapajer:
Adds new tests, as well as all three `IdStatus` cases for all the `Pivot`-only tests.

### DIFF
--- a/frontend/src/test/scala/quasar/ScalarStageSpec.scala
+++ b/frontend/src/test/scala/quasar/ScalarStageSpec.scala
@@ -1650,6 +1650,131 @@ object ScalarStageSpec {
 
 	input must interpretInto(stages)(expected)
       }
+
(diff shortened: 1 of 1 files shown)


alissapajer:
Published as `120.0.0-297ffb0`

alissapajer:
@wemrysi might be better to look at the commits individually, as the last two both rename/move a few tests as well as add new `IdStatus` cases to them, while the other commits strictly add new tests.

alissapajer:
Published as `120.0.0-a59904a`

alissapajer:
The additionally pivot tests exposed an `IdOnly` bug in mongo that we didn't know about and an `IncludeId` bug in tectonic (the one I was actually trying to reproduce). So they are pulling their weight already :)

PR #4105, opened 2019-03-11
Minimization2

alissapajer:
Depends on #4104

### DIFF
--- a/frontend/src/test/scala/quasar/ScalarStageSpec.scala
+++ b/frontend/src/test/scala/quasar/ScalarStageSpec.scala
@@ -876,6 +876,54 @@ object ScalarStageSpec {
 
         input must maskInto(".v" -> ColumnType.Top, ".w" -> Set(ColumnType.Object))(expected)
       }
+
+      "mask all values at Top to themselves" >> {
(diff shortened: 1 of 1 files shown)


alissapajer:
@wemrysi I'm going to keep opening these PRs depending on the previous one, as long as they are all versioned at **revision**. When you review, you could review them each individually, or review the most recent one and just merge that, letting all the others follow. Either way :)

frontend/src/test/scala/quasar/ScalarStageSpec.scala
```diff
@@ -875,6 +876,54 @@ object ScalarStageSpec {
...
+        "mask-26 at identity path" in {
+          val input = ldjson("""
+            [[1], {"z":2}, [], {}, null, 42, 42.2, true]
+            {"a":[1], "b":{"z":2}, "c":[], "d":{}, "e":null, "f":42, "g":42.2, "h":true}
+            []
+            {}
+            null
+            42
+            42.2
+            true
```
alissapajer: oops forgot string, I'll push a commit adding a string value

PR #4110, opened 2019-03-15
Add IdOnly and IncludeId tests for nested pivoting in cartouches

alissapajer:


alissapajer:
Published as `120.1.5-4ffca81`

PR #4116, opened 2019-03-27
Add MapBasedDatasource

rintcius:
Add MapBasedDatasource and remove SingletonDatasource and PureDatasource

Not for merge yet, just review. Also needs PRs from quasar-datasource-url and slamx

part of ch6217


rintcius:
@wemrysi I have added support for `prefix-resource`

PR #4117, opened 2019-04-02
Default to OnUndefined.Emit in LeftShift

alissapajer:
[ch6418]

### DIFF
--- a/qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
+++ b/qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
@@ -83,7 +83,13 @@ final class ApplyProvenance[T[_[_]]: BirecursiveT: EqualT: ShowT] private () ext
 
       case g @ Extractors.Transpose(src, retain, rot) =>
         computeDims[X](g) as g.overwriteAtRoot {
-          LeftShift(src.root, recFunc.Hole, retain.fold[IdStatus](IdOnly, ExcludeId), OnUndefined.Omit, RightTarget[T], rot)
+          LeftShift(
+            src.root,
+            recFunc.Hole,
+            retain.fold[IdStatus](IdOnly, ExcludeId),
+            OnUndefined.Emit,
+            RightTarget[T],
+            rot)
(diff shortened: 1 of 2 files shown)


alissapajer:
Surprisingly, there were no unit tests that care one way or another! So I added one :)

alissapajer:
release as `122.0.1-2c54678`

alissapajer:
:stop_sign: is up because a few mimir integration tests are failing. The no-pushdown variant of the integration tests passed, so I believe that the qscript pre-`InterpretedRead` rewrite is correct. But there are test failures in the pushdown variants, meaning that `OnUndefined.Omit` was masking underlying bugs either in the `InterpretedRead` rewrite itself, or in the `ScalarStages` evaluation, or both.

alissapajer:
There 5 are failing in both tectonic and rvalue pushdown in the same way. Test expectations need to be verified.
```
[error]   x flatten a flattened field [./doubleFlatten.test] (64 ms)
[error]   x triple flatten with mixed content [./tripleFlatten.test] (48 ms)
[error]   x shift map keys - distinct wildcard [./shiftMapKeysDistinctWildcard.test] (4 seconds, 195 ms)
[error]   x flatten map key - distinct wildcard [./flattenMapKeyWildcard.test] (3 seconds, 30 ms)
[error]   x double flatten with intervening field [./doubleFlattenWithIntervening.test] (699 ms)
```

ch6421 is tracking `doubleFlattenWithIntervening.test` (bug in thoth). Haven't looked at the others yet.

djspiewak:
@alissapajer It's actually theoretically possible to preserve undefined rows coming out of *tectonic*. Probably not qdata though due to the fact that the facade doesn't expose undefinedness (nor should it). This is a pretty thorny issue, unfortunately, since any outer operation applied to the results of an `InterpretedRead` (or the derived results from some inner operation applied to same!) is going to hit this issue, and MRA is very explicit about what the right answer is.

alissapajer:
Closing since it opened a bigger can of worms than we're going to address right now.

PR #4119, opened 2019-04-10
Support writeback in datasources

jsantos17:
This PR does not break API. I have commented the breakage so we can merge this without affecting other datasources. Once we add the REST interface for destination, we can uncomment the breakage and start working to support write-back.

[ch6446]


connector/src/main/scala/quasar/connector/LightweightDatasourceModule.scala
```diff
@@ -31,6 +33,7 @@ import scalaz.\/
 
 trait LightweightDatasourceModule {
   type DS[F[_]] = Datasource[F, Stream[F, ?], InterpretedRead[ResourcePath], QueryResult[F]]
+  type Dest[F[_]] = Destination[F, Stream[F, ?], Stream[F, Byte]]
```
jsantos17: Should we replace `Stream[F, Byte]` with `QueryResult[F]`? My thought process here is that Quasar should abstract conversion of `QueryResult[F]` to `Stream[F, Byte]`, rather than asking connectors to support all possible `QueryResult[F]`s.

jsantos17:
Should I update `HeavyweightDatasourceModule` to match the API changes in `LightweightDatasourceModule`? I don't know if we still have plans to use `HeavyweightDatasourceModule`.

connector/src/main/scala/quasar/connector/Capability.scala
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package quasar.connector
+
+sealed abstract class Capability
+case object ReadOnly extends Capability
+case object WriteOnly extends Capability
+case object ReadWrite extends Capability
```
djspiewak: Can we tuck these into the companion object?

connector/src/main/scala/quasar/connector/Destination.scala
```diff
@@ -0,0 +1,32 @@
...
+
+/** @tparam F effects
+  * @tparam G multiple results
+  * @tparam I destination input
+  */
+trait Destination[F[_], G[_], I] {
+
+  /** Returns a function within F that can consume input of type I, producing
+    * side effects within G
+    */
```
djspiewak: Not… sure this comment is all that useful. :-) It's literally just the type. I would either remove it or change it to a justification for the type, rather than a natural-language description of literally the type.

connector/src/main/scala/quasar/connector/LightweightDatasourceModule.scala
djspiewak: I agree with your thought process. `Stream[F, Byte]` seems right to me.

connector/src/main/scala/quasar/connector/LightweightDatasourceModule.scala
```diff
@@ -40,4 +43,10 @@ trait LightweightDatasourceModule {
       config: Json)(
       implicit ec: ExecutionContext)
       : F[InitializationError[Json] \/ Disposable[F, DS[F]]]
+
+  // val capability: Capability
+
+  // def destination[F[_]: Effect: ContextShift: MonadResourceErr](
+  //     config: Json)
+  //     : F[InitializationError[Json] \/ Disposable[F, Dest[F]]]
```
djspiewak: One thing we can do to make this less breaking, if we want, is to provide default implementations. It's still a trait, but it has no internal function references, so we could version it as **feature** rather than **breaking**.

connector/src/main/scala/quasar/connector/Destination.scala
```diff
@@ -0,0 +1,32 @@
...
+package quasar.connector
+
+import slamdata.Predef.Unit
+import quasar.api.resource._
+
+/** @tparam F effects
+  * @tparam G multiple results
+  * @tparam I destination input
+  */
+trait Destination[F[_], G[_], I] {
```
djspiewak: Does it make sense to abstract over `G` here? Maybe a better question: will `G` ever be something that *isn't* a stream?

connector/src/main/scala/quasar/connector/Capability.scala
jsantos17: 😳

connector/src/main/scala/quasar/connector/LightweightDatasourceModule.scala
jsantos17: That seems like a good idea! I'll push a commit with a default implementation returning `Left`.

connector/src/main/scala/quasar/connector/Destination.scala
```diff
@@ -0,0 +1,28 @@
...
+  def writeToPath(path: ResourcePath): F[I => G[Unit]]
```
wemrysi: We probably need at least an "unsupported path" error case here.

connector/src/main/scala/quasar/connector/LightweightDatasourceModule.scala
wemrysi: Is there an assumption here about the content of that bytestream? i.e. is it always going to be CSV?

If so, we might want something like `final case class CsvStream[F[_]](columns: List[TableColumn], data: Stream[F, Byte])`.

If not, then I'm not convinced `Stream[F, Byte]` is enough as we'd want to support some sort of negotiation. S3 can store any old bytestream but RedShift can't (nor can MongoDB, if we ever wanted to support it).

Maybe not for now, but we might also want to have the option of preserving type information if a destination supports it (i.e. temporal values).

connector/src/main/scala/quasar/connector/LightweightDatasourceModule.scala
wemrysi: How about we omit `capability` and move the `destination` function into a `DestinationModule`?

Connector implementations then communicate capabilities statically. Their module can implement one of the datasource modules, the destination module or a combination of the two.

We'd just need to update the module loading implementation to check if a loaded class is an instance of the `DestinationModule` in addition to the datasource modules.

This also has the benefit of keeping the concepts cleanly separated in the API. That a connector implements both is coincidental and we don't have any of this "it's a datasource module but not really because it only writes data" tomfoolery.

It also wouldn't break any APIs and be completely opt-in from a connector perspective.

djspiewak: @jsantos17 What happens if a connector is capable of writing, but doesn't have permissions to do so? Is that captured by the `F[I => G[O]]` signature?

jsantos17: `F` in `LightweightDatasourceModule` includes a `MonadResourceErr` constraint in the `destination` method. I'm planning (and currently working) on adding destination-related errors to `ResourceError`.

@wemrysi I like the `DestinationModule` idea. Will implement instead of default implementations in the `LightweightDatasourceModule` since it also fixes the problem with `HeavyweightDatasourceModule`.

connector/src/main/scala/quasar/connector/Destination.scala
jsantos17: I have bound this `F` in `LightweightDatasourceModule` with `MonadResourceErr` constraint. Shouldn't that be enough? As long as we add appropriate errors to `ResourceError`.

djspiewak:
On a more general note, it's worth remembering that this abstraction doesn't work for things like RedShift, BigQuery, or most of our other current destinations. The reason for this is such destinations do not behave as filesystems and do not allow us to simply push data. RedShift, for example, is pretty much hard-coded to a workflow in which RedShift will SSH to *us* (always on port 22), and then execute a command that we give it (usually something like `curl http://localhost...`) to proactively retrieve the data. It has no mechanism for doing the inverse.

So this is good and very important, for sure, but it alone is not sufficient to implement the uber button. Just wanted to clarify that since @wemrysi brought up RedShift as an example.

jsantos17:
@wemrysi 

>We probably need at least an "unsupported path" error case here.

I'm not sure I follow. An error case like an appropriate `ResourceError` for this case?

Otherwise this should be ready for another look.

wemrysi:
@jsantos17 @djspiewak some of the comment's got me thinking about how we might make the API more flexible. Here's a sketch of some types that might serve this purpose, what do you think?

https://gist.github.com/wemrysi/7da42eba7f30967c4a3832e1bf43a701

jsantos17:
>@jsantos17 @djspiewak some of the comment's got me thinking about how we might make the API more flexible. Here's a sketch of some types that might serve this purpose, what do you think?
>
>https://gist.github.com/wemrysi/7da42eba7f30967c4a3832e1bf43a701

@wemrysi with regard to `ResultType` we have something somewhat similar in this PR already as [ResultSet](https://github.com/slamdata/quasar/blob/626e24f9bd1a9322f4202c4d0668fa4317349479/connector/src/main/scala/quasar/connector/ResultSet.scala#L27). In case a connector does not support a given output, the connector should return `Left` or raise a exception through `MonadResourceErr`. To avoid using exceptions as flow control, what if we included a list of supported formats in the `DestinationModule` trait? Something like:

```scala
trait DestinationModule {
  type Dest[F[_]] = Destination[F, Stream[F, ?], ResultSet[F]]

  def supported: Set[ResultType]

  def destination[F[_]: Effect: ContextShift: MonadResourceErr](
      config: Json)
      : F[InitializationError[Json] \/ Disposable[F, Dest[F]]]
}
```
for some suitable `ResultType`

This way we can check a destination's supported format before we attempt to sink a stream into it. I'm not sure what is the advantage of `List[ResultSink[F]]`?

wemrysi:
@jsantos17 
> This way we can check a destination's supported format before we attempt to sink a stream into it.

My goal was to enable precisely this, but without the partiality. If we want to be able to do this prior to initializing any destinations, something like the following might work:

https://gist.github.com/wemrysi/94e83ca2da10f49e762a6bb70d6ecb56

jsantos17:
Oh I see how this encodes more information statically. However, I'm not sure if it's worth complicating the API this much for this purpose.

We could check supported types prior to destination initialization by including a set of `ResultType`s in `DestinationModule`. It's not static but seems simpler.

WDYT @djspiewak, @alissapajer, @rintcius?

djspiewak:
@jsantos17 TBH I rather like @wemrysi's proposal. It's equivalent to including a list of types in the destination module, but with the added information necessary to statically check the expressions derived from this information. Without this, you'll need some form of partiality *somewhere* when using `DestinationModule`, which doesn't really seem necessary.

impl/src/main/scala/quasar/impl/datasource/local/LocalDestination.scala
```diff
@@ -0,0 +1,70 @@
...
+      root: JPath,
+      blockingContext: BlockingContext): LocalDestination[F] =
+    new LocalDestination[F](root, blockingContext)
+}
+
+final class LocalCsvSink[F[_]: Effect: ContextShift] private (
+  root: JPath,
+  blockingContext: BlockingContext) extends ResultSink[F] {
+
+  val resultType: ResultType.Aux[F, ResultType.Csv[F]#T] = ResultType.Csv()
```
jsantos17: I wish I could write this as

```scala
val resultType: ResultType[F] = ResultType.Csv[F]
```

The compiler cannot unify `(List[TableColumn], Stream[F, Byte])` with [`resultType.T` from `ResultSink`](https://github.com/slamdata/quasar/pull/4119/files#diff-1069b433c5bd20aae18d4077bbfde1acR28).

Any ideas?

djspiewak: You could probably write it as `val resultType = ResultType.Csv[F]`. The ascription on the left side of your desired declaration is less specific than the type of the right hand side.

jsantos17:
@wemrysi this should be ready for another look.

impl/src/main/scala/quasar/impl/datasource/local/LocalDatasourceModule.scala
```diff
@@ -17,48 +17,47 @@
...
 
 import argonaut.Json
 import cats.effect._
-import scalaz.{EitherT, \/}
+import scalaz.\/
 import scalaz.syntax.applicative._
 import shims._
 
-object LocalDatasourceModule extends LightweightDatasourceModule {
+object LocalDatasourceModule extends LightweightDatasourceModule with DestinationModule {
```
wemrysi: :golf: You could extract all the destination module things into a `trait LocalDestinationModule extends DestinationModule` and then extend `LocalDestinationModule` at both use sites.

connector/src/main/scala/quasar/connector/DestinationModule.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import quasar.api.datasource.DatasourceError.InitializationError
+import quasar.api.datasource.DestinationType
+
+import argonaut.Json
+import cats.effect.{Effect, ContextShift, Resource}
+import scalaz.\/
+
+trait DestinationModule {
+  def destinationKind: DestinationType
```
wemrysi: Might as well name this `destinationType`.

connector/src/main/scala/quasar/connector/ResultSink.scala
```diff
@@ -0,0 +1,29 @@
...
+import slamdata.Predef.Unit
+
+import quasar.api.resource.ResourcePath
+
+import fs2.Stream
+
+trait ResultSink[F[_]] {
+  val resultType: ResultType[F]
+
+  def apply(dst: ResourcePath, result: resultType.T): F[Stream[F, Unit]]
```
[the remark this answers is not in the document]
jsantos17: I was mostly following `fs2`'s types that define a sink as:

```
type Sink[F[_], I] = Stream[F, I] => Stream[F, Unit] 
```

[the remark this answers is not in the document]
djspiewak: I think this may have been in response to some review I left in a previous version, where the inner effect was parametric, but was *always* instantiated as `Stream[F, ?]` and would practically always have to be instantiated as stream due to other constraints.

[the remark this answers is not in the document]
jsantos17: Yeah, the previous version of this PR had the `Stream[F, ?]` parameterized as `G` to achieve the same types `fs2` uses. This version eliminated that parametricity.

connector/src/main/scala/quasar/connector/DestinationModule.scala
jsantos17: I named it `destinationKind` rather than `destinationType` for consistency with `LightweightDatasourceModule`'s `kind` member.

wemrysi: We only used `kind` because `type` is reserved. I'd suggest we use `destinationType` and later refactor the datasource modules to use `datasourceType`.

connector/src/main/scala/quasar/connector/ResultSink.scala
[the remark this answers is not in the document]
wemrysi: @jsantos17 Hm, but not all inputs will be streams, could we just return `F[Unit]` instead?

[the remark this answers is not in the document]
jsantos17: @wemrysi sure. I can simply compile my `Stream[F, Unit]` into `F[Unit]` in the local implementation.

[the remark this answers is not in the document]
jsantos17: Can we represent all possible sinks with `A => F[Unit]`?

jsantos17:
This should be ready for another look.

impl/src/main/scala/quasar/impl/datasource/local/LocalDestinationModule.scala
```diff
@@ -0,0 +1,45 @@
...
+import quasar.connector.{Destination, DestinationModule, MonadResourceErr}
+
+import argonaut.Json
+import cats.effect.{ContextShift, ConcurrentEffect, Resource, Timer}
+import scalaz.\/
+import scalaz.syntax.applicative._
+import shims._
+
+trait LocalDestinationModule extends DestinationModule {
+  private lazy val blockingPool: BlockingContext =
```
wemrysi: I think we want to make this an abstract `def` so that the concrete modules can provide it. As-is, I'm pretty sure we're creating two blocking pools per concrete module, one for the destination and one for the datasource.

jsantos17: Makes sense, I assumed the caching would prevent this.

wemrysi: @jsantos17 I think it is more due to `private` members shadowing each other. As a pool is defined here and also in the `object`s.

jsantos17: Done!

PR #4121, opened 2019-04-19
Add AggregateResource and PhysicalDatasource

rintcius:
[ch6070]


impl/src/main/scala/quasar/impl/datasource/package.scala
```diff
@@ -16,13 +16,29 @@
...
 import fs2.Stream
+import scalaz.Applicative
 
 package object datasource {
   type AggregateResult[F[_], A] = Stream[F, (ResourcePath, A)]
   type CompositeResult[F[_], A] = Either[A, AggregateResult[F, A]]
+
+  type MonadCreateErr[F[_]] = MonadError_[F, CreateError[Json]]
+
+  def toPhysicalDatasource[F[_]: Applicative: MonadCreateErr, G[_], Q, R](ds: Datasource[F, G, Q, R])
```
rintcius: Couldn't avoid using this method, sadly.

Currently this simply returns a `FailedDatasource` in case it's not an actual `PhysicalDatasource` because we never end up in the `other` branch at this moment. When it becomes necessary we can implement this by filtering `ResourcePathType.Physical` for example.

rintcius:
Ready for review, but needs FE changes before it can be merged

impl/src/main/scala/quasar/impl/datasource/ChildAggregatingDatasource.scala
```diff
@@ -73,25 +78,49 @@ final class ChildAggregatingDatasource[F[_]: Monad: MonadResourceErr, Q, R] priv
...
+
+  // `p / *` exists iff underlying `p` is prefix/prefixresource
+  private def aggPathExists(path: ResourcePath): F[Boolean] =
+    aggPath(path)
+      .map(p => underlying.prefixedChildPaths(p).flatMap(ofPrefix))
+      .getOrElse(false.pure[F])
+
+  // checks whether the provided stream is that of a prefix/prefixresource
+  private def ofPrefix[A](os: Option[Stream[F, A]]): F[Boolean] =
+    os.map(s => s.pull.peek1.flatMap {
```
wemrysi: :golf: `lastOrError` shouldn't be necessary here, `os.traverse(<pull>.stream.compile.last).map(_.join getOrElse false)` is one example that avoids it.

rintcius:
@wemrysi thanks! I didn't think of the `Aux` pattern. I had some difficulties making all the types line up correctly with this, but it worked out nicely in the end.

Ready for another look.

connector/src/main/scala/quasar/connector/Datasource.scala
```diff
@@ -29,6 +29,8 @@ import monocle.{Lens, PLens}
   */
 trait Datasource[F[_], G[_], Q, R] extends QueryEvaluator[F, Q, R] {
 
+  type PathType <: ResourcePathType
```
wemrysi: So it doesn't look like we're really gaining any benefits from having this be a type member, we might as well just make it a type parameter since it looks like every reference to a `Datasource` type involves `Aux`.

connector/src/main/scala/quasar/connector/Datasource.scala
```diff
@@ -41,17 +43,25 @@ trait Datasource[F[_], G[_], Q, R] extends QueryEvaluator[F, Q, R] {
...
 
 object Datasource {
-  def evaluator[F[_], G[_], Q, R]: Lens[Datasource[F, G, Q, R], QueryEvaluator[F, Q, R]] =
-    pevaluator[F, G, Q, R, Q, R]
+  type Aux[F[_], G[_], Q, R, P <: ResourcePathType] = Datasource[F, G, Q, R] { type PathType = P }
+
+  @SuppressWarnings(Array("org.wartremover.warts.AsInstanceOf"))
+  def widenPathType[F[_], G[_], Q, R, PI <: ResourcePathType, PO >: PI <: ResourcePathType](
+      ds: Aux[F, G, Q, R, PI]): Aux[F, G, Q, R, PO] =
+    ds.asInstanceOf[Aux[F, G, Q, R, PO]]
```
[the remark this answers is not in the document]
rintcius: I tried doing that, but that requires a variance annotation I think? Couldn't get that compiling.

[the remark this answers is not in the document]
wemrysi: Ah, yeah you'll need a type lambda like `λ[+a => Aux[F, G, Q, R, a]]`.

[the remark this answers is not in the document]
rintcius: Hmm, I don't see how that changes anything (trying on `Datasource`, now that `Datasource.Aux` is removed). Isn't the problem that the variance annotation still needs to be propagated?

In the end I'm getting:
```
covariant type G occurs in invariant position in type (prefixPath: quasar.api.resource.ResourcePath)F[slamdata.Predef.Option[G[(quasar.api.resource.ResourceName, P)]]] of method prefixedChildPaths
[error]   def prefixedChildPaths(prefixPath: ResourcePath)
[error]       ^
```

This is with defining trait as
```
trait Datasource[+F[_], +G[_], Q, R, +P <: ResourcePathType] extends QueryEvaluator[F, Q, R] {
```

BTW 
```
  // compiles ok
  def something1(x: ResourcePath): P = scala.Predef.??? 

  // covariant type P occurs in invariant position in type (x: quasar.api.resource.ResourcePath)slamdata.Predef.Option[P] of method something2
  def something2(x: ResourcePath): Option[P] = scala.Predef.???

  // covariant type P occurs in invariant position in type (x: quasar.api.resource.ResourcePath)G[P] of method something3
  def something3(x: ResourcePath): G[P] = scala.Predef.???

```

connector/src/main/scala/quasar/connector/Datasource.scala
rintcius: Hmm yes indeed, that works out better I think. It's a bit ironic that that was the first proposed solution, but that's the price of learning I guess ;)

wemrysi: Yeah, that's on me. I thought we'd be able to allow the type member to dictate more result types (and thus avoid more of the tparam propagation), a la `func(ds: Datasource): ds.PathType` but many of the usages don't actually involve a datasource value as we have arguments like `(X, Datasource) => Y` or `F[Datasource]` so we don't have a stable value from which to access the type member.

rintcius: Not on you, I also thought it would work out better with the type member

rintcius:
@wemrysi Ready for another look. 

FYI I'll wait with updating the dependent PRs until this one is approved

PR #4122, opened 2019-04-23
Feature/mapdb indexed store

cryogenian:
Moved from https://github.com/slamdata/mimir/pull/252


cryogenian:
Released as 123.1.0-96ade58

impl/src/main/scala/quasar/impl/storage/MapDBIndexedStore.scala
```diff
@@ -0,0 +1,92 @@
...
+object MapDBIndexedStore {
+  val TREEMAP_NAME: String = "main"
+
+  val ref: Ref[IO, Map[String, DB]] = Ref.unsafe(Map.empty)
+
+  def apply[F[_]: Sync: LiftIO](path: Path): F[IndexedStore[F, String, String]] =
+    makeOrGetDB[F](path) flatMap { db => Sync[F].delay(new MapDBIndexedStore[F](db)) }
+
+  def mkDb[F[_]: Sync](path: Path): F[DB] =
+    Sync[F].delay(DBMaker.fileDB(path.toFile).checksumHeaderBypass.transactionEnable.closeOnJvmShutdown.fileLockDisable.make)
```
alissapajer: We also want `fileMmapEnableIfSupported` if the OS in not Windows.

alissapajer: For [reference](https://jankotek.gitbooks.io/mapdb/content/performance/):

> Mmap files are highly dependent on the operating system. For example, on Windows you cannot delete a mmap file while it is locked by JVM. If Windows JVM dies without closing the mmap file, you have to restart Windows to release the file lock.

> There is also bug in JVM. Mmaped file handles are not released until DirectByteBuffer is GCed. That means that mmap file remains open even after db.close() is called. On Windows it prevents file to be reopened or deleted.

alissapajer: I found two more recommend options that go along with mmapped files:
- `fileMmapPreclearDisable()` (improves performance)
- `cleanerHackEnable()` (unmaps file when it's closed to prevent a race condition when mapped file is read after it's closed, which causes the JVM to crash).

cryogenian:
@alissapajer 

+ Mmaping enabled
+ Replaced `Ref` with `ConcurrentHashMap`

impl/src/main/scala/quasar/impl/storage/MapDBIndexedStore.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  private val isWindows: Boolean =
+    java.lang.System.getProperty("os.name").toLowerCase.indexOf("windows") > -1
+
+  def apply[F[_]: Sync](path: Path): F[IndexedStore[F, String, String]] =
+    makeOrGetDB[F](path) flatMap { db => Sync[F].delay(new MapDBIndexedStore[F](db)) }
+
+  private def mkDb(path: Path): DB = {
+    val rawMaker = DBMaker.fileDB(path.toFile).checksumHeaderBypass.transactionEnable.closeOnJvmShutdown.fileLockDisable
+    val mmapedMaker = if (isWindows) rawMaker.fileMmapEnableIfSupported.fileMmapPreclearDisable.cleanerHackEnable else rawMaker
```
alissapajer: We want the inverse here right? `if (!isWindows)`

cryogenian: :D Right

cryogenian:
123.1.0-6cfc682

impl/src/main/scala/quasar/impl/storage/MapDBIndexedStore.scala
```diff
@@ -0,0 +1,93 @@
...
+    !Option(bmap.remove(k)).isEmpty
+  }}
+
+  private def commit: F[Unit] = Sync[F].delay(db.commit)
+
+}
+object MapDBIndexedStore {
+  val TREEMAP_NAME: String = "main"
+
+  private val cache: ConcurrentHashMap[String, DB] = new ConcurrentHashMap()
```
wemrysi: What problem is this cache solving?

cryogenian: It disallows using more than one instance of `DB` for one particular file. It's like locking but only for one quasar process. Also if (for whatever reason) such `DB` is needed in more than one place it won't raise an error or corrupt underlying file.

cryogenian:
`123.1.0-bfbcfcb`

/cc @wemrysi @alissapajer

impl/src/main/scala/quasar/impl/storage/ConcurrentMapIndexedStore.scala
```diff
@@ -0,0 +1,67 @@
...
+    extends IndexedStore[F, K, V] {
+
+  private val F = Sync[F]
+
+  private def evalOnPool[A](fa: F[A]): F[A] =
+    ContextShift[F].evalOn[A](blockingPool.unwrap)(fa)
+
+  def entries: Stream[F, (K, V)] = for {
+    iterator <- Stream.eval(evalOnPool(F.delay(mp.entrySet.iterator.asScala)))
+    entry <- Stream.fromIterator[F, java.util.Map.Entry[K, V]](iterator)
```
wemrysi: Since we don't know anything about the iterator returned here, I think we'll want to execute all of the the effects in returned `Stream` on the blocking context.

cryogenian:
+ `evalStreamOnPool` added

PR #4124, opened 2019-04-29
Don't ask...

djspiewak:
I'm actually not even certain how the JDBM3 stuff in mimir was working *at all* on `CNum` columns without this change. `scalaz.Order` is not `Serializable`, and `CNum` can appear in `ColumnRef`, which in turn appears in `RowFormat`, which in turn appears in `SortingKeyComparator`, which is set as the comparator on a JDBM3 `TreeMap`, which in turn serializes its comparator, key, and value serializer instances.


rintcius:
Just curious, did you encounter an actual serialization error?
 
May be worth to create a ticket for adding unit tests which ensure that everything that needs to be serializable is actually serializable.

PR #4125, opened 2019-05-02
Aggregate recursively

rintcius:
Depends on #4121 - last 2 commits are for this feature

[ch6576]


rintcius:
will merge together with #4121 - so changed label to breaking

PR #4126, opened 2019-05-08
atomix: AsyncDistributedMapStore, event stream construction and consumption

cryogenian:
This is draft, since there is no tests and some warns appear in build, but the general idea is here already 🙂 

Interesting that actually all but one distributed map implementation based on atomic map implementation. So, for distributed map with gossip protocol it should be safe to use `ConcurrentMapIndexedStore`, but for any other alternatives this is necessary. 

`consume` returns `Deferred` to notify when actually whole stream is consumed, might be useful if we have finite streams (e.g. tests). 

~~Also I think that we probably don't need to  have atomix indexed store, but just initialize atomix distributed map with indexed store content and consume distributed map events to synchronize things.~~
Ignore that, of course we need atomix indexed store :) 

@wemrysi Would you mind to take a quick look? Am I going to right direction?


cryogenian:
Actually we need atomix indexed store 🤦‍♂

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.CompletableFuture
+
+final class AsyncDistributedMapStore[F[_]: Async: ContextShift, K, V](
+    mp: AsyncDistributedMap[K, V])
+    extends IndexedStore[F, K, V] {
+
+  import AsyncDistributedMapStore._
+
+  def entries: Stream[F, (K, V)] = for {
+    iterator <- Stream.eval(Sync[F].delay(mp.entrySet.iterator))
```
wemrysi: Will want to `bracket` here instead of `eval` in order to ensure the `Iterator` is closed upon early stream termination.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+  } yield (entry.getKey, entry.getValue)
+
+  def lookup(k: K): F[Option[V]] =
+    toF(mp get k) map (Option(_))
+
+  def insert(k: K, v: V): F[Unit] =
+    toF(mp.put(k, v)) as (())
+
+  def delete(k: K): F[Boolean] =
+    toF(mp.remove(k)) map { (x: V) => !Option(x).isEmpty }
```
wemrysi: :golf: `Option(x).nonEmpty`

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+}
+
+object AsyncDistributedMapStore {
+  trait IndexedStoreEvent[K, +V]
```
wemrysi: Do we have a use-case in mind for indexed store events? Is this something that we want to add to the `IndexedStore` abstraction rather than just this implementation?

Also, this trait should be `sealed` and extend `Product with Serializable` for nicer inference.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+  def eventStream[F[_]: ConcurrentEffect: ContextShift, K, V](mp: AsyncDistributedMap[K, V]): Stream[F, IndexedStoreEvent[K, V]] = {
+    val F = ConcurrentEffect[F]
+    def run(f: F[Unit]): Unit = F.runAsync(f)(_ => IO.unit).unsafeRunSync
+    def listener(cb: IndexedStoreEvent[K, V] => Unit): MapEventListener[K, V] = { event => event.`type` match {
+      case INSERT => cb(Insert(event.key, event.newValue))
+      case UPDATE => cb(Insert(event.key, event.newValue))
+      case REMOVE => cb(Remove(event.key))
+    }}
+    for {
+      q <- Stream.eval(Queue.bounded[F, IndexedStoreEvent[K, V]](1))
```
wemrysi: I didn't see any backpressure support for atomix events, if there indeed isn't any then we'll need a different consumption strategy. Currently, if the consumer is too slow, we'll end up blocking event listener threads.

Depending on the use case, we could configure a maximum outstanding event size and then drop events if the queue is full.

If consumers need to know if they've missed events, we could come up with another strategy.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+    Stream.unfoldEval(iterator)(getNext)
+  }
+
+  def toF[F[_]: Async: ContextShift, A](cf: CompletableFuture[A]): F[A] = {
+    if (cf.isDone) cf.get.pure[F]
+    else {
+      val fa: F[A] = Async[F].async { cb: (Either[Throwable, A] => Unit) =>
+        val _ = cf.whenComplete { (res: A, t: Throwable) => {
+          val eitherResult: Either[Throwable, A] =
+            if (!Option(t).isEmpty) Left(t) else Right(res)
```
wemrysi: :golf: `Option(t).toLeft(res)`

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+          cb(eitherResult)
+        }}
+      }
+      fa productL ContextShift[F].shift
+    }
+  }
+
+  def consume[F[_]: Concurrent, K, V](
```
wemrysi: If something like this is needed, I'd model it as
```scala
def consume[F[_], K, V](into: IndexedStore[F, K, V]): Sink[F, IndexedStoreEvent[K, V]]
```
and leave execution decisions to the caller.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+  def entries: Stream[F, (K, V)] = for {
+    iterator <- Stream.eval(Sync[F].delay(mp.entrySet.iterator))
+    entry <- fromIterator[F, java.util.Map.Entry[K, V]](iterator)
+  } yield (entry.getKey, entry.getValue)
+
+  def lookup(k: K): F[Option[V]] =
+    toF(mp get k) map (Option(_))
+
+  def insert(k: K, v: V): F[Unit] =
+    toF(mp.put(k, v)) as (())
```
djspiewak: I'm not sure this signature is sufficient anymore. What happens if another node adds the same key?

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  def delete(k: K): F[Boolean] =
```
djspiewak: Similarly here on this signature. What happens if another node changes the value in the middle of this delete? We probably wouldn't want to just delete what they're doing.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
cryogenian: Depends on protocol. If we have strong consistency then the last one wins. If it's anti-entropy protocol then eventually the last one wins, but in the middle some nodes might be inconsistent.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
cryogenian: Yes, I thought about something like this 
+ Create atomix map 
+ Create atomix event stream
+ Create atomix indexed store
+ Create mapdb indexed store 
+ Consume atomix event stream by mapdb indexed store.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
cryogenian: All events must be consumed, `addEventListener` can accept `Executor`, will `BlockingContext` there work for this use case?

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
cryogenian: Ok

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
cryogenian: The changes done after `delete` start will propagate eventually, the latest timestamp wins :)

cryogenian:
**TLDR;** since `compute` like function on gossip based map are based on local node values and messaging doesn't preserve old values it seems that there is no point to use `alter` instead of `insert` and `remove`

I might be wrong but there is only one eventually-consistent map provided by atomix. It's gossip based `DistributedMap`. Atomix docs say that `DistributedMap` is eventually consistent, but in fact the only one implementation that isn't wrapper around `AtomicMap` is gossip. Since we're going for eventual consistency I checked [this](https://github.com/atomix/atomix/blob/893aa1d051a2070cae0d4b28ca4afc3c992e92fd/protocols/gossip/src/main/java/io/atomix/protocols/gossip/map/AntiEntropyMapDelegate.java) the underlying _map_ of gossip based distributed map. 

It appears that messages sent|received by the map delegate actually [has no notion of previous values](https://github.com/atomix/atomix/blob/893aa1d051a2070cae0d4b28ca4afc3c992e92fd/protocols/gossip/src/main/java/io/atomix/protocols/gossip/map/UpdateEntry.java) only [timestamps](https://github.com/atomix/atomix/blob/893aa1d051a2070cae0d4b28ca4afc3c992e92fd/protocols/gossip/src/main/java/io/atomix/protocols/gossip/map/MapValue.java#L29). So, if we 
+ Update key `k` on node `a` and set it's value to `b`
+ Update key `k` on node `b` and set it's value to `c`
+ Update key `k` on node `c` and set it's value to `null` (i.e. remove key `k`)

Then eventually all nodes will have the latest value determined by the operation time. If more than one operation have the same timestamp then state is inconsistent. 
Suppose we 
+ Update key `k` on node `b` if it's value `c` to `null` (i.e. `remove(k, c)`)
+ Update key `k` on node `c` using `putIfAbsent`. 

Checking that previous values is absent or equal to `c` or whatever happens on current `items`. So, if it's for example `b` then both `putIfAbsent` and `remove(k, c)` will do nothing with cluster state. This makes me doubt that there is any pros switching from `put+remove` to something like `alter: (K, Option[V] => Option[V]) => F[Option[V]]`

OTOH, if we use consistent protocol then there is no need in `alter` too, since `CompletableFuture` is resolved when consistency is achieved. (It might be a bit more convenient 🙂)

One more example 
+ Let we have cluster with 1000 nodes (to have chance that message from one node to another can take some time)
+ Select node `a` and node `b` and look at key `k`
+ At time 0 both `a::k` and `b::k` has value `v`
+ At time 1 `a` runs `remove(k, v)`
+ At time 2 `b` runs `putIfAbsent(k, w)`
+ At time 3 `b` takes message to set `k` to `null`

Both values are `null` now, but the message can arrive before `putIfAbsent`! What I am trying to say is that eventual-consistency protocol with `compute` like function based on local values looks useless. Just inserting|removing allows to at least know the last value from the cluster is here, eventually will be here.
In comparison if we have `remove(k)` and `put(k, w)` it's eventually `w`, no chance to end up with `null`.

@djspiewak Did I missed anything? Does `_latest modification wins eventually_ seem OK?

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
djspiewak: My point is that this signature doesn't give us the power to detect and respond to that situation. Callers just have to assume that their call made it.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
djspiewak: Oh I'm sure there's a resolution to it, but the resolution is not exposed to callers of the API in any way. There's no way to respond to such a situation, meaning that things will just silently be incorrect.

wemrysi:
@cryogenian I think your illustration makes sense regarding the lack of atomicity when using the eventually-consistent replication.

Maybe we need a subtype of `IndexedStore` that provides atomic operations that we can implement with `AtomicMap`?

It seems like most of our current use-cases would benefit from atomicity for updates/deletes given our desire to avoid sessions.

cryogenian:
@wemrysi Yep, I'm going to use consistent protocols (atomic maps) for indexed store since we need eventually-consistent protocol for append-only store. The `insert` and `update` functions will return `F[Deferred[F, Something]]` where outer `F` is for sending message and `Deferred` is resolved when whole cluster is updated.

wemrysi:
@cryogenian Ok. Have you considered returning `F[Something]` that (semantically) blocks until consistency is achieved? Users could then leverage a `Concurrent[F]` instance to run the process in the background if desired.

cryogenian:
@wemrysi `F[Something]` is here actually :) To be honest, I'm fine with both `F[Deferred[F, A]]` and `F[A]`. 

On the one hand `F[A]` is cool because we don't need to modify any existing code.
On the other hand `F[A]` assumes that computation might be long and blocking, but `Deferred` yelling _I could be long and blocking, handle me!_

cryogenian:
@wemrysi @djspiewak Can you take another look, please? 

+ `IndexedStore`'s `insert` and `delete` returns `F[Deferred[F, Something]]`, this makes `ConcurrentMapIndexedStore` working for sync variants of atomix consistent maps too. 
+ `bracket` instead of `eval` around iterator
+ `Product with Serializable`
+ Removed `consume` and added `consumeEvent` to `IndexedStore` object. 

I don't think that ability to create eventstream is definitely part of indexed store, e.g. here event stream is constructed from `AsyncDistributedMap` not indexed store itself.

cryogenian:
[ch6814] [ch6815]

impl/src/main/scala/quasar/impl/storage/IndexedStore.scala
```diff
@@ -38,15 +38,26 @@ trait IndexedStore[F[_], I, V] {
...
 }
 
 object IndexedStore extends IndexedStoreInstances {
+  sealed trait Event[K, +V] extends Product with Serializable
+  object Event {
+    final case class Insert[K, V](k: K, v: V) extends Event[K, V]
+    final case class Delete[K](k: K) extends Event[K, Nothing]
+  }
+
+  def consumeEvent[F[_]: Functor, K, V](store: IndexedStore[F, K, V], event: Event[K, V]): F[Unit] = event match {
```
wemrysi: As this doesn't await any of the `Deferred` results, race conditions can lead to out-of-order processing.

impl/src/main/scala/quasar/impl/storage/AsyncDistributedMapStore.scala
```diff
@@ -0,0 +1,119 @@
...
+    val F = ConcurrentEffect[F]
+    def run(f: F[Unit]): Unit = F.runAsync(f)(_ => IO.unit).unsafeRunSync
+    def listener(cb: Event[K, V] => Unit): MapEventListener[K, V] = { event => event.`type` match {
+      case INSERT => cb(Insert(event.key, event.newValue))
+      case UPDATE => cb(Insert(event.key, event.newValue))
+      case REMOVE => cb(Delete(event.key))
+    }}
+
+    for {
+      q <- Stream.eval(Queue.circularBuffer[F, Event[K, V]](128))
```
wemrysi: Probably should make the buffer size should an argument to `eventStream` since exceeding it means lost events. We should also document that semantic.

In the future, we might consider a version of this stream that emits an additional event indicating that the consumer was too slow and missed something, allowing them a chance to restart from a known state.

cryogenian:
Closing in favour #4181

PR #4127, opened 2019-05-09
Vectorized provenance

wemrysi:
Introduces a `Provenance` abstraction that aligns closely with the MRA formalism, replacing the previous implementation. Provides a reasonably efficient implementation loosely based on GSS.

[ch5556]
[ch5432]
[ch5629]


impl/src/main/scala/quasar/impl/provenance/Dim.scala
```diff
@@ -0,0 +1,151 @@
...
+    val project: Prism[Dim[S, V, T], (S, T)] =
+      Prism.partial[Dim[S, V, T], (S, T)] {
+        case Project(s, t) => (s, t)
+      } {
+        case (s, t) => Project(s, t)
+      }
+
+    val scalar: Traversal[Dim[S, V, T], (S, T)] =
+      new Traversal[Dim[S, V, T], (S, T)] {
+        import scalaz.syntax.applicative._
```
djspiewak: Random idiom question: do you think it's better to use the scalaz syntax in cases like this? Or the cats syntaxes with `import shims._` to bridge the gap?

impl/src/main/scala/quasar/impl/provenance/Identities.scala
```diff
@@ -0,0 +1,574 @@
...
+    protected val roots: Set[Int],
+    protected val ends: Set[Int],
+    protected val g: Identities.G[A]) {
+
+    import Identities.{G => IG, MergeState, Node, Vert => IVert}
+
+  /** The number of vectors in the set. */
+  def breadth: Int = {
+    @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+    def go(lvl: Set[Int]): Eval[Int] =
```
djspiewak: <3 `Eval`. It's bizarre how much more often it's used than `Need`/`Value`/`Name`.

impl/src/main/scala/quasar/impl/provenance/Identities.scala
```diff
@@ -0,0 +1,574 @@
...
+
+          nextd.map(max(_, acc))
+        }
+      }
+
+    go(roots, 0).value
+  }
+
+  /** A view of these identities as a set of lists of conjoined regions. */
+  def expanded: NonEmptyList[NonEmptyList[NonEmptyList[A]]] = {
```
djspiewak: Maybe it would be better to type-alias `NonEmptyList` with three different aliases? I'm just trying to find a way to make this easier to parse.

impl/src/main/scala/quasar/impl/provenance/Identities.scala
```diff
@@ -0,0 +1,574 @@
...
+            }
+          }
+
+        zip0(nthis, nthat, nout)
+      }
+
+    zip0(roots, that.roots, Monoid[C].empty)
+  }
+
+  /** Returns whether `this` is equal to `that`. */
```
djspiewak: Best scaladoc NA.

impl/src/main/scala/quasar/impl/provenance/Identities.scala
```diff
@@ -0,0 +1,574 @@
...
+  @SuppressWarnings(Array("org.wartremover.warts.Equals"))
+  def === (that: Identities[A])(implicit A: Order[A]): Boolean = {
```
djspiewak: It's interesting that you put a space after your symbolic functions. Not really wrong, just like… interesting.

impl/src/main/scala/quasar/impl/provenance/ProvImpl.scala
```diff
@@ -0,0 +1,209 @@
...
+import cats.syntax.list._
+import cats.syntax.traverse._
+
+import monocle.Traversal
+
+import scalaz.@@
+import scalaz.Tags.{Conjunction, Disjunction}
+import scalaz.syntax.tag._
+
+trait ProvImpl[S, V, T] extends Provenance[S, V, T] {
```
djspiewak: This whole trait is awesome. So clean.

impl/src/test/scala/quasar/impl/provenance/DimGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+    Arbitrary(Gen.frequency(
+      1 -> Gen.const(D.fresh()),
+      5 -> arbitrary[(V, T)].map(D.inflate(_)),
+      5 -> arbitrary[(S, T)].map(D.inject(_)),
+      5 -> arbitrary[(S, T)].map(D.project(_))))
+  }
+
+  implicit def cogenDim[S: Cogen, V: Cogen, T: Cogen]: Cogen[Dim[S, V, T]] =
+    Cogen((k, dim) => dim match {
+      case d @ Dim.Fresh() => Cogen[Int].perturb(k, d.hashCode)
```
djspiewak: I believe you want `System.identityHashCode(d)` here. `d.hashCode` will be a constant, I believe.

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
```diff
@@ -0,0 +1,395 @@
...
+
+import shims._
+
+object IdentitiesSpec extends Qspec
+    with SpecificationLike
+    with IdentitiesGenerator
+    with CatsNonEmptyListGenerator
+    with Discipline {
+
+  implicit val params = Parameters(maxSize = 10)
```
djspiewak: Probably worth adding the parallelism config.

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
```diff
@@ -0,0 +1,395 @@
...
+  "initRegions" >> {
+    "undefined on singleton vector" >> {
+      Identities(42).initRegions must beNone
+    }
+
+    "elides singleton vectors from set" >> prop { (x: Int, y: Int, ys: NonEmptyList[Int]) =>
+      val a = Identities(x)
+      val b = Identities.fromReducible(ys)
+
+      a.merge(b :+ y).initRegions eqv Some(b)
```
djspiewak: `eqv` rather than `must_===`?

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
```diff
@@ -0,0 +1,395 @@
...
+      ids must beEquivalentToDistinct(vs)
+      //ids.storageSize must_=== (2 * ints.length + 2)
+    }}
+
+    "preserves subsumes efficiently" >> prop { (ints: NonEmptyList[Int], x: Int) =>
+      val vs = vecs(ints :+ x, ints)
+      val ids = Identities.collapsed(vs)
+
+      ids must beEquivalentToDistinct(vs)
+      ids.storageSize must_=== (ints.length + 1)
```
djspiewak: This assertion (and all the ones like it) are so awesome.

impl/src/test/scala/quasar/impl/provenance/ProvImplSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.scalacheck.Arbitrary
+
+import org.specs2.scalacheck.Parameters
+
+object ProvImplSpec extends {
+
+  val prov: Provenance.Aux[Char, Int, Boolean, Uop[Identities[Dim[Char, Int, Boolean]]]] =
+    ProvImpl[Char, Int, Boolean]
+
+  val params = Parameters(maxSize = 6)
```
djspiewak: Parallelism config.

Also holy early initializer syntax, batman.

project/Versions.scala
```diff
@@ -3,8 +3,9 @@ package quasar.project
 object Versions {
   val algebraVersion      = "1.0.0"
   val argonautVersion     = "6.2.2"
+  val catsVersion         = "1.5.0"
   val catsEffectVersion   = "1.1.0"
-  val disciplineVersion   = "0.7.2"
+  val disciplineVersion   = "0.9.0"
```
djspiewak: Aren't we already using ScalaCheck 0.14? Wouldn't that imply we need to use a more recent version of Discipline and cats-laws?

qscript/src/main/scala/quasar/qscript/provenance/AutoJoin.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import shims._
+
+/** Description of an MRA autojoin.
+  *
+  * @tparam S the type of scalar identities
+  * @tparam V the type of vector identities
+  *
+  * @param keys the identity comparisons that form the join condition
+  * @param onUndefined how to handle a row when the identities are undefined
```
djspiewak: Why is this a property of the autojoin?

qscript/src/main/scala/quasar/qscript/provenance/Uop.scala
```diff
@@ -0,0 +1,144 @@
...
+  ////
+
+  @SuppressWarnings(Array("org.wartremover.warts.MutableDataStructures"))
+  private def distinctE[F[_]: Foldable, A: Eq](fa: F[A]): List[A] =
+    fa.foldLeft(ListBuffer[A]()) { (as, a) =>
+      if (as.exists(_ === a))
+        as
+      else
+        as :+ a
+    }.toList
```
djspiewak: You can do this in the other direction and avoid the `ListBuffer`, but I'm not sure if that's worth the effort.

qscript/src/test/scala/quasar/qscript/provenance/CatsNonEmptyListGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import cats.data.NonEmptyList
+
+import org.scalacheck.Cogen
+
+// Copied from cats-laws due to the scalacheck version incompatibility between
+// cats-1.5.0 and quasar, using the instances from `cats` results in
+// NoSuchMethodErrors at runtime.
+//
+// FIXME: Remove once cats upgrades to scalacheck-1.14.x
```
djspiewak: Ahhhhhh, I know why I was confused. Cats already has upgraded to 1.14.0, but for 1.x only on the Scala 2.13 cross-publish. Cats 2.0.0-M1 is on 1.14.0 across the board, but it has some other minor compatibility breakage.

qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
```diff
@@ -42,38 +43,49 @@ import quasar.qscript.{
...
   def QAuthS[F[_]](implicit ev: QAuthS[F]): QAuthS[F] = ev
 
   private type V[F[_], A] = F[ValidationNel[Symbol, A]]
   implicit def V[F[_]: Applicative] = Applicative[F].compose[ValidationNel[Symbol, ?]]
 
-  val dims = QProv[T]
-  val ejs = ejson.Fixed[T[EJson]]
-  val func = construction.Func[T]
-  val recFunc = construction.RecFunc[T]
+  private lazy val B = Bucketing(qprov)
```
djspiewak: Perils of cake!

qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
```diff
@@ -191,62 +160,75 @@ final class ApplyProvenance[T[_[_]]: BirecursiveT: EqualT: ShowT] private () ext
...
-          }
-
-          repairDims getOrElse shiftedDims
+              case _ =>
+                func.Undefined
+            })(sdims, applyIdStatus(idStatus, shiftedDims))
         }
 
+      // FIXME: This is incorrect, we need an indexed inflate identity to
+      //        properly compute this.
```
djspiewak: No worries. It's going away.

qsu/src/main/scala/quasar/qsu/LPtoQS.scala
```diff
@@ -87,7 +97,15 @@ final class LPtoQS[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT]
 }
 
 object LPtoQS {
-  def apply[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT]: LPtoQS[T] = new LPtoQS[T]
+  def apply[T[_[_]]: BirecursiveT: EqualT: ShowT: RenderTreeT](
+      qp: QProv[T])(
```
djspiewak: Is there a reason not to just… pick a `QProv` here? Do we gain anything from pushing this particular decision further to the leaves?

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
```diff
@@ -54,6 +54,8 @@ import scalaz.syntax.either._
 import scalaz.syntax.std.boolean._
 import scalaz.syntax.tag._
 
+import shims.{eqToScalaz, orderToCats, orderToScalaz, showToCats, showToScalaz}
```
djspiewak: Why the specific imports?

qsu/src/test/scala/quasar/qsu/ReifyAutoJoinsSpec.scala
```diff
@@ -19,18 +19,23 @@ package quasar.qsu
...
 import matryoshka.data.Fix
 import matryoshka.data.freeEqual
 import matryoshka.delayEqual
+
 import pathy.Path
 import pathy.Path.Sandboxed
-import scalaz.{EitherT, INil, Need, StateT}
+import scalaz.{EitherT, Need, StateT}
+
+import shims.{eqToScalaz, orderToCats, orderToScalaz}
```
djspiewak: Here as well?

djspiewak:
Also, shouldn't this be **breaking** since the QSU API is changing?

build.sbt
```diff
@@ -239,11 +239,18 @@ lazy val qscript = project
     api,
     frontend % BothScopes)
   .settings(commonSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "org.typelevel" %% "cats-laws"        % catsVersion       % Test,
+      "org.typelevel" %% "cats-kernel-laws" % catsVersion       % Test,
+      "org.typelevel" %% "discipline"       % disciplineVersion % Test))
```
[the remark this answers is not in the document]
wemrysi: :fist_oncoming:

build.sbt
```diff
@@ -283,17 +290,21 @@ lazy val impl = project
...
+      "com.slamdata"   %% "fs2-gzip"         % fs2GzipVersion.value,
+      "com.slamdata"   %% "qdata-tectonic"   % qdataVersion.value,
+      "com.slamdata"   %% "tectonic-fs2"     % tectonicVersion.value,
+      "org.http4s"     %% "jawn-fs2"         % jawnfs2Version,
+      "org.slf4s"      %% "slf4s-api"        % slf4sVersion,
+      "org.typelevel"  %% "cats-laws"        % catsVersion       % Test,
+      "org.typelevel"  %% "cats-kernel-laws" % catsVersion       % Test,
+      "org.typelevel"  %% "discipline"       % disciplineVersion % Test,
+      "org.typelevel"  %% "jawn-argonaut"    % jawnVersion,
+      "org.typelevel"  %% "jawn-util"        % jawnVersion,
```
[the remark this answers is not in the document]
wemrysi: Lol, the best.

impl/src/main/scala/quasar/impl/provenance/Dim.scala
```diff
@@ -0,0 +1,151 @@
...
+
+import scalaz.Applicative
+
+/** A single dimension of an identity vector.
+  *
+  * @tparam S type of scalar identities
+  * @tparam V type of vector identities
+  * @tparam T sorts of identities
+  */
+sealed abstract class Dim[S, V, T]
```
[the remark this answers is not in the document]
wemrysi: Still surprised, or do you think this and the rest of the provenance impl should live elsewhere?

impl/src/main/scala/quasar/impl/provenance/Dim.scala
wemrysi: Sort of depends, if `shims` is already imported, then I'd just use the cats ones, but, depending on the file, `shims` adds a non-trivial amount of compile time so introducing it just for something like this might be too expensive.

Here, using the `scalaz` import seemed reasonable due to the singular, private use-site.

impl/src/main/scala/quasar/impl/provenance/Identities.scala
wemrysi: Yeah, will do :+1:

impl/src/main/scala/quasar/impl/provenance/Identities.scala
wemrysi: I'm a bit of a completionist.

impl/src/main/scala/quasar/impl/provenance/Identities.scala
wemrysi: I find it makes them easier to parse, ymmv.

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
wemrysi: `must_===` doesn't use the `Eq` instance.

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
wemrysi: The ones that are commented out were for the initial `Identities#merge` implementation that attempted to merge after diverging. I left them in case we decide we want to give it another go at some point.

impl/src/test/scala/quasar/impl/provenance/ProvImplSpec.scala
wemrysi: Yeah =/ I needed it to get the various implicit instance types to unify.

qscript/src/main/scala/quasar/qscript/provenance/AutoJoin.scala
wemrysi: We need some notion of the absence of compatibility. When the autojoin is completely empty, then we're fine and just cross. When unions are involved, some components of the union may need to cross whereas others will align and produce join keys.

In `A ⋈ (A | B)`, if the autojoin between the `A`s produce join keys, then they will only be defined when comparing two `A` rows. Without any additional information, this would cause the case when `A` and `B` rows are compared to produce nothing instead of cross.

We have to compute the autojoin in order to detect this scenario, so it seemed reasonable to include it as a property. Knowing whether it is necessary allows us to be potentially more efficient during execution (`EquiJoin` vs `Filter . Cross`).

qscript/src/main/scala/quasar/qscript/provenance/Uop.scala
wemrysi: Yeah, maybe I was thinking a bit too specialized as `foldRight` usually involves a `reverse` in the linear collection implementations. Not sure which is more expensive.

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
djspiewak: Wait… it doesn't? Doesn't it use `scalaz.Equal`?

qscript/src/main/scala/quasar/qscript/provenance/AutoJoin.scala
djspiewak: Ahhhhhhhh. Got it.

qsu/src/main/scala/quasar/qsu/ApplyProvenance.scala
wemrysi: Indeed :disappointed: 

I hadn't intended to make these `trait`s, but ran afoul of scala/bug#5712.

qscript/src/main/scala/quasar/qscript/provenance/Uop.scala
djspiewak: I'm fine with either. Not worth changing if you aren't already in that code.

impl/src/main/scala/quasar/impl/provenance/Dim.scala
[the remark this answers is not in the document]
djspiewak: I think it really comes down to whether or not abstracting `qprov` in **qsu** makes sense. I tend to think that MRA stuff should live in **qsu** (or **qscript**). I think I asked about it elsewhere, but if we're not getting mileage out of the abstraction in qsu, we should probably just concretize it. Obviously it imposes a real syntactic and boilerplate cost on all the QSU phases.

qsu/src/main/scala/quasar/qsu/LPtoQS.scala
wemrysi: The current organization means there isn't a type to pick at this point (since `qsu` doesn't depend on `impl`). If we reorganize things then picking it here would be fine.

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
wemrysi: Nope.

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
wemrysi: With specific imports
```
sbt:root> qsu/test:compile
[info] Compiling 1 Scala source to /home/emrys/work/slamdata/quasar/qsu/target/scala-2.12/test-classes ...
[info] Done compiling.
[success] Total time: 20 s, completed May 10, 2019 10:22:43 AM
```
with wildcard
```
sbt:root> qsu/test:compile
[info] Compiling 1 Scala source to /home/emrys/work/slamdata/quasar/qsu/target/scala-2.12/test-classes ...
[info] Done compiling.
[success] Total time: 284 s, completed May 10, 2019 10:27:36 AM
```
rip.

qsu/src/test/scala/quasar/qsu/ReifyAutoJoinsSpec.scala
wemrysi: Same as above (though not as extreme), compiles in 3 seconds vs 20 seconds.

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
djspiewak: Daaaaaaaamn

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
djspiewak: Wow. I wonder if there's a specs2-cats project somewhere?

impl/src/test/scala/quasar/impl/provenance/IdentitiesSpec.scala
djspiewak: Yeah I noticed that.

impl/src/main/scala/quasar/impl/provenance/Dim.scala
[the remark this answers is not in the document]
wemrysi: I think the abstraction is beneficial from a reasoning/specification standpoint, forcing us to think about what is definitional, making sure behavior/relationships are specified in terms of the core algebra, etc.

The boilerplate cost in this PR is a bit more than I anticipated (I'd envisioned an extra ctor argument in affected phases), though I'd argue most of it isn't due to using an abstraction but to a current limitation of scala. I don't think it introduces much syntactic cost over what was there previously.

As for having the implementation live in qscript/qsu, I think that's fine. I'll move it and also reduce some of the boilerplate in the QSU phases.

alissapajer:
👍 I'll approve and/or comment later today.

qsu/src/test/scala/quasar/qsu/MinimizeAutoJoinsSpec.scala
alissapajer: 284, wow

wemrysi:
@djspiewak Sounds good, I actually have a few more of your comments to address anyhow.

djspiewak:
I'll merge once the build is green!

PR #4128, opened 2019-05-09
Introduce destination-related abstractions

jsantos17:
The main part of this PR are the two interfaces we're going to use to manage destinations:

1. `Destinations`
2. `DestinationManager`

I have implementation of these two traits in the pipeline but I'd like to start reviewing this sooner rather than later. 

You may notice these two traits share many similarities with `Tables`, `Datasources`, `DatasourceManager` and `PreparationsManager`, but I'm not convinced they're similar enough to justify abstraction to eliminate the duplication. Any thoughts are very welcome.

Tagged as `revision` since the changes are purely additive.

[ch6522]


jsantos17:
I'm closing this and opening this same PR again, but including default implementations.

PR #4175, opened 2019-05-17
Use cats.effect.Resource for lifecycle management

wemrysi:
Consistently employ `cats.effect.Resource` for lifecycle management, replacing `quasar.Disposable` and some usages of `fs2.Stream`.


wemrysi:
Released as `126.0.1-58411c3`.

foundation/src/main/scala/quasar/contrib/cats/monadError.scala
```diff
@@ -0,0 +1,43 @@
...
+package quasar.contrib.cats
+
+import quasar.contrib.scalaz.MonadError_
+
+import scala.util.Either
+
+import cats.{Monad, MonadError}
+
+object monadError {
+  implicit def catsMonadErrorMonadError_[F[_], E](implicit F: MonadError[F, E]): MonadError_[F, E] =
```
djspiewak: Shouldn't we put this on the `MonadError_` companion?

impl/src/main/scala/quasar/impl/datasource/ByNeedDatasource.scala
```diff
@@ -32,7 +31,7 @@ import ByNeedDatasource.NeedState
 
 final class ByNeedDatasource[F[_], G[_], Q, R, P <: ResourcePathType] private (
     datasourceType: DatasourceType,
-    mvar: MVar[F, NeedState[F, Disposable[F, Datasource[F, G, Q, R, P]]]])(
+    mvar: MVar[F, NeedState[F, (Datasource[F, G, Q, R, P], F[Unit])]])(
```
djspiewak: To play the devil's advocate… In cases like this, you're just replacing `Disposable` with `Tuple2`. Is that really an improvement? Would it be better to keep the nominal type for cases like this?

foundation/src/main/scala/quasar/contrib/cats/monadError.scala
```diff
@@ -0,0 +1,43 @@
...
+    new MonadError_[F, E] {
+      def raiseError[A](e: E): F[A] = F.raiseError(e)
+      def handleError[A](fa: F[A])(f: E => F[A]): F[A] = F.handleErrorWith(fa)(f)
+    }
+
+  // Not implicit as can easily introduce ambiguity
+  def monadError_CatsMonadError[F[_], E](
```
djspiewak: Randomly, and this is probably a matter for another PR, should we consider using cats-mtl instead?

impl/src/main/scala/quasar/impl/datasources/DefaultDatasourceManager.scala
```diff
@@ -209,27 +214,37 @@ object DefaultDatasourceManager {
...
-          .map(_.map(ManagedDatasource.heavyweight(_)))
+        ByNeedDatasource(ref.kind, mkhw).map(ManagedDatasource.heavyweight(_))
     }
 
+  private def handleInitErrors[F[_]: MonadCreateErr: MonadError[?[_], Throwable], A](
+      kind: DatasourceType,
+      res: => Resource[F, Either[InitializationError[Json], A]])
+      : Resource[F, A] = {
+
+    import quasar.contrib.cats.monadError.monadError_CatsMonadError
```
djspiewak: This… import…

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
```diff
@@ -173,21 +173,22 @@ object PreparationsManager {
...
 
-      back <- emit.concurrently(q.dequeue.parJoin(maxConcurrency)) onComplete {
-        Stream.eval_(notificationsQ.enqueue1(None))
+      pm <- emit.concurrently(q.dequeue.parJoin(maxConcurrency)) onFinalize {
+        notificationsQ.enqueue1(None)
       }
-    } yield back
+    } yield pm
+
+    pmStream.compile.resource.lastOrError
```
djspiewak: Doing this rather than the current `Stream` will result in worse fairness properties and less accurate type signatures for downstream consumers. `Resource` is definitely more appropriate than `Stream` when working with a singleton stream that previously only existed for the purposes of wrapping a bind-scoped `bracket`, but this case is much richer than that. `Stream` is a more appropriate type here.

Another way of conceptualizing this function: it was returning an infinite stream, where all but one of the steps in the stream are non-emitting effects.

impl/src/test/scala/quasar/impl/table/DefaultTablesSpec.scala
```diff
@@ -68,12 +68,12 @@ final class DefaultTablesSpec extends TablesSpec[IO, UUID, String, String, Strin
       def evaluate(query: String): IO[String] = IO(query)
     }
 
-  val manager: List[PreparationsManager[IO, UUID, String, String]] =
+  val manager: PreparationsManager[IO, UUID, String, String] =
     PreparationsManager[IO, UUID, String, String](evaluator)(runToStore)
-      .compile.toList.unsafeRunSync
+      .allocated.map(_._1).unsafeRunSync
```
djspiewak: Wow… The original (deleted) line here was really really broken. I never caught that before. It's probably a coincidence that this test is working at all.

The new line is broken, too: the internal managing job stream will be shut down immediately upon completion. The correct thing to do here is a lot more complicated: we need to keep the shutdown function for the stream and run it in a `step(...)` at the end of the spec.

run/src/main/scala/quasar/run/Quasar.scala
```diff
@@ -70,19 +70,22 @@ object Quasar extends Logging {
...
-      preparationsManager: QueryEvaluator[F, SqlQuery, R] => Stream[F, PreparationsManager[F, UUID, SqlQuery, R]],
+      preparationsManager: QueryEvaluator[F, SqlQuery, R] => Resource[F, PreparationsManager[F, UUID, SqlQuery, R]],
       lookupTableData: UUID => F[Option[R]],
       lookupTableSchema: UUID => F[Option[S]])(
       datasourceModules: List[DatasourceModule],
       sstEvalConfig: SstEvalConfig)(
       implicit
       ec: ExecutionContext)
-      : Stream[F, Quasar[F, R, S]] = {
+      : Resource[F, Quasar[F, R, S]] = {
```
djspiewak: Another area where the new types are lying (and the runtime structures deny the downstream consumer the ability to manage certain things, such as fairness). `Quasar` runs things in the background, and so the more accurate type here *is* `Stream`.

run/src/main/scala/quasar/run/Quasar.scala
```diff
@@ -91,17 +94,17 @@ object Quasar extends Logging {
...
       dsManager <-
-        Stream.resource(DefaultDatasourceManager.Builder[UUID, Fix, F]
+        DefaultDatasourceManager.Builder[UUID, Fix, F]
           .withMiddleware(AggregatingMiddleware(_, _))
           .withMiddleware(ConditionReportingMiddleware(onCondition)(_, _))
-          .build(moduleMap, configured))
+          .build(moduleMap, configured)
 
-      freshUUID = ConcurrentEffect[F].delay(UUID.randomUUID)
+      freshUUID = Sync[F].delay(UUID.randomUUID)
```
djspiewak: Random style question: is it better to summon the weakest typeclass which defines the operation (in this case, `Sync`), or the specific typeclass that you know you happen to have (in this case, `ConcurrentEffect`)?

impl/src/main/scala/quasar/impl/datasource/ByNeedDatasource.scala
wemrysi: Yes, but these cases should only appear internally and it is unlikely we'll use any of the cleanup threading that `Disposable` included. So I guess if we wanted to have a `Disposable` that is equivalent to a nominal `Tuple2`, then sure, but it didn't seem worth it.

foundation/src/main/scala/quasar/contrib/cats/monadError.scala
wemrysi: Probably, less maintenance of our own thing would be nice.

foundation/src/main/scala/quasar/contrib/cats/monadError.scala
wemrysi: We probably can, I'll give it a go. Not sure if it will make it easy to introduce ambiguities with `shims`.

impl/src/main/scala/quasar/impl/datasources/DefaultDatasourceManager.scala
wemrysi: Yeah, it the unfortunate situation where we only have evidence of a `MonadError_` and we need to derive a `cats.MonadError` from it, ugh.

foundation/src/main/scala/quasar/contrib/cats/monadError.scala
djspiewak: Hmm, it shouldn't cause any ambiguities with shims since `MonadError_` is completely separate from the hierarchy. Is there anything that materializes a scalaz instance using a `MonadError_` instance?

run/src/main/scala/quasar/run/Quasar.scala
wemrysi: I prefer the least powerful abstraction, both for the principle and that it can sometimes help one notice that typeclass on the signature is unnecessarily powerful.

djspiewak: I like that, too. I stopped doing it because it was making me add imports, and it also leads to weird things like:

```scala
Sync[F].delay(...) flatMap { _ =>
  Async[F].async(...) { _ => 
    Concurrent[F].start(...)
  }
}
```

Which seems like a smell. Or does it? I still don't know.

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
wemrysi: Would you mind elaborating on the fairness argument? I don't understand the disadvantage.

As for the type signature, I actually think `Resource` is more appropriate than `Stream` here, precisely because it _wasn't_ an infinite stream. If one were to run the previous stream, it would start the job, emit the manager and then immediately stop the job (as you mention below as a problem with the test).

So the running job exists in the resource lifetime of the emitted manager, which I think `Resource` better describes (this is the [scenario](https://github.com/functional-streams-for-scala/fs2/blob/v1.0.4/core/shared/src/main/scala/fs2/Stream.scala#L4056) the `resource` method was introduced for, it would seem).

run/src/main/scala/quasar/run/Quasar.scala
wemrysi: Yeah, I don't know either, that is a bit annoying, to be sure.

foundation/src/main/scala/quasar/contrib/cats/monadError.scala
wemrysi: I was thinking of the case where we derive a `MonadError_` from a `scalaz.MonadError` and `shims` is in scope providing a conversion between `scalaz.MonadError` and `cats.MonadError`. Even if this is possible, implicit prioritization should make it so that one of them would win.

impl/src/test/scala/quasar/impl/table/DefaultTablesSpec.scala
wemrysi: I agree I don't know how the old one ever worked, the new one should work fine, but leak resources (so we should still do what you suggest).

Since we use `allocated`, the job stream is started when run, but we never stop it.

connector/src/main/scala/quasar/connector/DestinationModule.scala
```diff
@@ -19,14 +19,16 @@ package quasar.connector
...
-import scalaz.\/
 
 trait DestinationModule {
   def destinationType: DestinationType
 
   def destination[F[_]: ConcurrentEffect: ContextShift: MonadResourceErr: Timer](
-    config: Json)
-    : F[InitializationError[Json] \/ Resource[F, Destination[F]]]
+      config: Json)
+      : Resource[F, Either[InitializationError[Json], Destination[F]]]
```
jsantos17: Oh right. `Resource[F, A]` is equivalent to `F[Disposable[F, A]]` rather than to `Disposable[F, A]`. Didn't catch that before

jsantos17:
@wemrysi shall we hold this PR until we sort out Mongo? cc @cryogenian

cryogenian:
@jsantos17 The problem has been found :)

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
djspiewak: > Would you mind elaborating on the fairness argument? I don't understand the disadvantage.

It may be irrelevant now that fs2 compiles down to fibers and `Bracket`. I'm not entirely sure. My knowledge on this front is more conceptual than concrete.

In previous versions, compiling a `Stream` to its target effect type would result in a loss of granularity in the scheduler. More concretely, it's the difference between `s1.merge(s2).compile.lastOrError` and `s1.compile.lastOrError.race(s2.compile.lastOrError)`. I actually don't know the extent to which *current* fs2 takes advantage of this, but at least previously, the difference was quite severe. The compilation of the former can take advantage of direct knowledge of `Stream`'s algebra, gaining more granular control over the effects and their scheduling. This in turn allows higher level properties (most notably, fairness and batching) to be optimized, whereas something like `race` is operating at too high of a level, since everything from the original `Stream` is entirely "encoded away."

Again, I'm not sure if that's a concern with *current* fs2, but it most definitely was in prior versions, and back in scalaz-stream.

> As for the type signature, I actually think Resource is more appropriate than Stream here, precisely because it wasn't an infinite stream. If one were to run the previous stream, it would start the job, emit the manager and then immediately stop the job (as you mention below as a problem with the test).

It *is* an infinite stream though. I think of `Stream[F, A]` as something like `InfiniteList[∃ e . F[e] \/ A]`. This stream had only one `A`, but it had an infinite number of effects. You're certainly correct that if someone took the stream and ran it individually using `drain`, `lastOrError`, or anything similar, they would experience serious issues. But isn't the fact that they're trying to take a stream and run it with abbreviation of its results (rather than more appropriately merging it with some other stream) the real problem here?

> So the running job exists in the resource lifetime of the emitted manager, which I think Resource better describes (this is the scenario the resource method was introduced for, it would seem).

I do agree that if the stream is to be run (rather than composed into a broader network of streams), `Resource` is definitely the right choice. And certainly, the scaladoc on that function is quite compelling regarding design intent. Given the documentation on that function, I'm certainly mollified enough to remove my conceptual roadblock, but I will note that the call site immediately takes this `Resource` and re-wraps it in `Stream`, implying that the stream didn't need to be run *in the first place* and could have simply been left as a stream.

Note that the issues with premature termination as you raise them above *also* happen in the re-wrapping case:

```scala
val r = s1.concurrently(s2).compile.resource.lastOrError
val s = Stream.resource(r)
val result = s.compile.lastOrError      // terminates s2 early
```

So there isn't really a free lunch here.

connector/src/main/scala/quasar/connector/DestinationModule.scala
djspiewak: Yeah. `Resource` is sort of like a weird monad transformer.

impl/src/main/scala/quasar/impl/table/PreparationsManager.scala
wemrysi: > In previous versions, compiling a Stream to its target effect type would result in a loss of granularity in the scheduler. More concretely, it's the difference between `s1.merge(s2).compile.lastOrError` and `s1.compile.lastOrError.race(s2.compile.lastOrError)`

Yeah, this very well may still be the case, I'm not familiar enough with the impl to say, however all the cases I've elected to turn a singleton stream into a `Resource` involve `concurrently`, which compiles the "concurrent" stream into the target effect and `Concurrent#start`s it, so it's effects aren't being interleaved with the "main" stream.

> But isn't the fact that they're trying to take a stream and run it with abbreviation of its results (rather than more appropriately merging it with some other stream) the real problem here?

Given the type emitted by the original stream, yes I'd agree that expecting things to work after running it is problematic. However there are plenty of instances where running a stream for its results is perfectly valid. That is basically my point about using `Resource` for things like this, it is more precise because there aren't any instances, semantically (coincidentally, sure), where calling `Resource#use` and still expecting the managed value to be in a usable state afterwards is a valid expectation.

PR #4176, opened 2019-05-21
Change Uop repr to a SortedSet

wemrysi:
We were representing `Uop` with a `List` to try and avoid needing an `Order` instance for `Identities`, however it turned out to be straightforward to implement the latter, so we can now use a `SortedSet` for `Uop` which results in better performance and fixing the issue with `Cogen`.

[ch7021]

### DIFF
--- a/foundation/src/main/scala/quasar/contrib/cats/boolean.scala
+++ b/foundation/src/main/scala/quasar/contrib/cats/boolean.scala
@@ -14,27 +14,22 @@
  * limitations under the License.
  */
 
-package quasar.qsu.mra
+package quasar.contrib.cats
 
-import slamdata.Predef.Set
+import slamdata.Predef.Boolean
 
-import quasar.pkg.tests._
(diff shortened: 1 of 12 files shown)


wemrysi:
@alissapajer The current performance should be quite a bit improved. The semantics of `Uop` have always been set-like, when it was backed by a list we used an `O(n^2)` means of maintaining distinct values when merging (so that we only required `Eq[A]`). Now that we're requiring `Order[A]`, we can use `SortedSet` which gives us `O(n log n)` merging.

alissapajer:
@wemrysi ah! I didn't realize that we were enforcing set semantics previously with list. That's great!

alissapajer:
@wemrysi already merged it, but I meant to ask why the version was **breaking**. Are the changes not entirely contained within `qsu`?

wemrysi:
@alissapajer They are, but I thought `breaking` as the affected APIs are still public.

PR #4181, opened 2019-05-24
Atomix/consistent

cryogenian:
[ch6814] [ch6815]

@wemrysi I agree that we can omit `Deferred` and additional handling is a bit inconvenient, so, back to `F`, 

Also removed streaming and added this task https://app.clubhouse.io/data/story/7102/implement-atomix-distributed-map-backed-with-mapdb-map.

I'm not exactly sure if that's the (I wasn't sure about stream as well) but it seems that integrating mapdb into `AsyncAtomicMap` is better idea than syncing data on every start.


wemrysi:
@cryogenian Would `feature` be more appropriate than `revision` for this PR?

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import fs2.Stream
+
+import io.atomix.core.map.AsyncAtomicMap
+import io.atomix.core.iterator.AsyncIterator
+import io.atomix.utils.time.Versioned
+
+import java.util.concurrent.CompletableFuture
+
+final class AsyncAtomicIndexedStore[F[_]: Async: ContextShift, K, V](
```
dcsobral: I don't recall seeing chained colons before. Maybe I just don't _recall_ it, but, still... What does this expand as?

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def delete(k: K): F[Boolean] =
+    toF(mp.remove(k)) map { (x: Versioned[V]) => Option(x.value).nonEmpty }
+}
+
+object AsyncAtomicIndexedStore {
+  def apply[F[_]: Async: ContextShift, K, V](mp: AsyncAtomicMap[K, V]): IndexedStore[F, K, V] =
+    new AsyncAtomicIndexedStore(mp)
+
+  def fromIterator[F[_]: ContextShift: Async, A](iterator: AsyncIterator[A]): Stream[F, A] = {
```
dcsobral: Is `ContextShift: Async` here as opposed to `Async: ContextShift` used elsewhere on purpose, or does it makes no difference?

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
```diff
@@ -0,0 +1,77 @@
...
+    } yield step
+    Stream.unfoldEval(iterator)(getNext)
+  }
+
+  def toF[F[_]: Async: ContextShift, A](cf: CompletableFuture[A]): F[A] = {
+    if (cf.isDone)
+      cf.get.pure[F]
+    else
+      Async[F].async { cb: (Either[Throwable, A] => Unit) =>
+        val _ = cf.whenComplete { (res: A, t: Throwable) => cb(Option(t).toLeft(res)) }
```
dcsobral: I seem to recall something about `val _ = ...` behaving differently in Dotty.

dcsobral: `Option(throwable).toLeft(result)` -- that's not how I'd write it. Is there a reason why `Option(result).toRight(throwable)` would not work, or is it just personal preference?

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
cryogenian: `Async` and `ContextShift` implicits added to constructor. I.e.
```scala
final class Foo[A: U: W](a: B) === final class Foo[A](a: B)(implict whatever: U[A], whateverElse: W[A])
```

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
cryogenian: That's the same thing :)

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
cryogenian: In general `result` might be `null` e.g. in case of `CompletableFuture[Void]`, OTOH `throwable` is always `null` when computation has no errors.

impl/src/main/scala/quasar/impl/storage/AsyncAtomicIndexedStore.scala
dcsobral: Damn, I *did* know that. :)

PR #4186, opened 2019-05-28
Ensure LocalDatasource paths are confined to config dir

wemrysi:
[ch6644]


wemrysi:
@jsantos17 do you mean the symlinks are listed and, if they point to a directory, may be further descended into?

If so, I don't really see a problem with that, I don't see the `rootDir` as a physical path but a logical one. It is up to the user to add/remove things like that from the configured root directory.

If we want to add a feature to allow control over symlink following, we can. I think a better practice for users would be to create a directory that only contains that which should be available to access, possibly using symlinks to populate it, rather than relying on not following symlinks to prevent access to things.

djspiewak:
> do you mean the symlinks are listed and, if they point to a directory, may be further descended into?
>
> If so, I don't really see a problem with that, I don't see the rootDir as a physical path but a logical one. It is up to the user to add/remove things like that from the configured root directory.

I agree with this. It's up to the user to make sure the jail directory is safe.

PR #4187, opened 2019-05-29
Initialize destinations for destination REST API

jsantos17:
The most relevant changes are in `quasar.run.Quasar`. There are other changes that are necessary for subsequent PRs with the actual API implementation.

Part of [ch7014]

Changes are `breaking` because of changes in `quasar.run.Quasar.apply`.


jsantos17:
Released as `128.1.0-107524a`

jsantos17:
🛑up since I'll need to make changes in this branch for dependent PR reviews.

impl/src/main/scala/quasar/impl/destinations/DefaultDestinations.scala
```diff
@@ -34,7 +34,7 @@ import scalaz.syntax.monad._
 import scalaz.{\/, Equal, IMap, ISet, OptionT, Order}
 import shims._
 
-class DefaultDestinations[F[_]: Sync, I: Equal: Order, C] private (
+class DefaultDestinations[I: Equal: Order, C, F[_]: Sync] private (
```
cryogenian: Why do you prefer this?

api/src/main/scala/quasar/api/datasource/DatasourceType.scala
```diff
@@ -16,23 +16,34 @@
...
 import scalaz.syntax.show._
+import shapeless.{Witness => W}
 
 @Lenses
 final case class DatasourceType(name: DatasourceType.Name, version: Positive)
 
 object DatasourceType extends DatasourceTypeInstances {
-  type Name = datasource.Name
-  type NameP = datasource.NameP
+  type NameP = MatchesRegex[W.`"[a-zA-Z0-9-]+"`.T]
```
cryogenian: Why should this be defined here? Just wondering :)

impl/src/main/scala/quasar/impl/destinations/DefaultDestinations.scala
jsantos17: I wanted to keep it consistent with `DefaultDestinationManager`. `F[_]` at the end won.

api/src/main/scala/quasar/api/datasource/DatasourceType.scala
jsantos17: No good reason. For some reason they aren't found when put in `quasar.api`'s package object :/

jsantos17:
Closing in favor of #4202

PR #4201, opened 2019-06-03
Binary RValue

wemrysi:
Preliminary work for ch7173.

### DIFF
--- a/common/src/main/scala/quasar/common/data/QDataRValue.scala
+++ b/common/src/main/scala/quasar/common/data/QDataRValue.scala
@@ -18,6 +18,11 @@ package quasar.common.data
 
 import slamdata.Predef._
 
+import cats.instances.string._
+import cats.syntax.eq._
+
+import monocle.Optional
+
 import qdata._
 import qdata.time.{DateTimeInterval, OffsetDate}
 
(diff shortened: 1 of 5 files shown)


common/src/main/scala/quasar/common/data/RValue.scala
```diff
@@ -378,8 +386,8 @@ sealed abstract class RValueInstances {
   implicit val show: Show[RValue] =
     Show.showFromToString
 
-  implicit val qdataEncode: QDataEncode[RValue] = QDataRValue
-  implicit val qdataDecode: QDataDecode[RValue] = QDataRValue
+  implicit def qdataEncode: QDataEncode[RValue] = QDataRValue
+  implicit def qdataDecode: QDataDecode[RValue] = QDataRValue
```
rintcius: Curious, why this change from `val` to `def`?

wemrysi: The impl of `QDataRValue` references some `vals` defined in the `RValue` companion leading to NPE due to initialization order.

common/src/main/scala/quasar/common/data/QDataRValue.scala
```diff
@@ -173,16 +194,29 @@ object QDataRValue extends QDataEncode[RValue] with QDataDecode[RValue] {
 
   def getMetaValue(a: RValue): RValue = a match {
     case RMeta(value, _) => value
+    case CBinary(bytes) => RValue.rString(CBinary.toBase64String(bytes))
```
alissapajer: Why the decision to turn the bytes into a string rather than keeping them as a byte array?

PR #4219, opened 2019-06-18
Add scalacheck generators for CPathNode and CPath

wemrysi:


wemrysi:
@alissapajer Ah, I missed that one

PR #4223, opened 2019-06-21
Move MapDB storage implementation into quasar

jsantos17:
[ch7966]

### DIFF
--- a/run/src/main/scala/quasar/run/store/Store.scala
+++ b/run/src/main/scala/quasar/run/store/Store.scala
@@ -42,7 +42,7 @@ object Store {
     path: Path,
     codec: CodecJson[A],
     pool: BlockingContext)
-      : Resource[F, IndexedStore[F, UUID, A]] =
+    : Resource[F, IndexedStore[F, UUID, A]] =
     for {
       db <- mapDB[F](path)
       table <- Resource.liftF(uuidTable[F](db))
--- a/run/src/main/scala/quasar/run/store/Store.scala
+++ b/run/src/main/scala/quasar/run/store/Store.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package quasar.run
+package quasar.run.store
 
 import slamdata.Predef._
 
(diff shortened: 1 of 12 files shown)


jsantos17:
Published as `130.0.15-dcc6d15`

run/src/test/scala/quasar/run/store/StoreCodecSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  val jsonGen: Gen[Json] =
+    Gen.oneOf(
+      List(
+        Json.obj("foo" -> Json.jString("bar")),
+        Json.jString("quux"),
+        Json.jNumber(42L),
+        Json.array(Json.jString("one"), Json.jString("two"))))
+
+  def modifyGen: Gen[DestinationRef[Json] => DestinationRef[Json]] =
```
cryogenian: I can't find if it's used anywhere.

run/src/main/scala/quasar/run/store/Store.scala
```diff
@@ -0,0 +1,103 @@
...
+import shims._
+
+object Store {
+  val MapDBTableName = "default"
+
+  def codecStore[A: Show, F[_]: Sync: ContextShift: MonadError_[?[_], StoreError]](
+    path: Path,
+    codec: CodecJson[A],
+    pool: BlockingContext)
+      : Resource[F, IndexedStore[F, UUID, A]] =
```
cryogenian: Shouldn't this be 4 spaces, not 6?

run/src/test/scala/quasar/run/store/StoreCodecSpec.scala
jsantos17: Oh right. This was used when I tested this is a prism. Will delete.

run/src/main/scala/quasar/run/store/Store.scala
jsantos17: Right. `emacs` likes six for some reason.

jsantos17:
@cryogenian this should be ready for another look.

PR #4224, opened 2019-06-26
Use quasar.api.table.TableColumn instead of custom type in destination API

jsantos17:
[ch7279]


alissapajer:
@jsantos17 a couple compile errors

jsantos17:
@alissapajer this should be ready for another look.

alissapajer:
@jsantos17 we're changing the API that quasar's exposing, so technically this should be **breaking**.

jsantos17:
@alissapajer makes sense! Even if no-one is using this API just yet.

PR #4227, opened 2019-07-02
Improve performance of Provenance#projectStatic

wemrysi:


alissapajer:
Is this to address that query that took 25 minutes to compile? Do we have a ticket for that?

wemrysi:
@alissapajer yes and yes, but unfortunately this doesn't fix the problem, even though it is still an improvement. 7927 is one of the issues and has a link to the other.

alissapajer:
The build hung. I don't think we've been experiencing hung quasar builds in master? So it could be related to this PR.

wemrysi:
@alissapajer Looks like travis was having some issues, going to restart.

wemrysi:
@alissapajer I've managed to reproduce the hang locally, investigating.

wemrysi:
I haven't had time to track down the issue, closing for now.

PR #4230, opened 2019-07-03
add new schema method to Datasources

alissapajer:
[ch8054]

### DIFF
--- a/api/src/main/scala/quasar/api/datasource/Datasources.scala
+++ b/api/src/main/scala/quasar/api/datasource/Datasources.scala
@@ -29,8 +29,10 @@ import scalaz.{\/, ISet}
   * @tparam G multple results
   * @tparam I identity
   * @tparam C configuration
+  * @tparam OldS old schema configuration
+  * @tparam S schema configuration
   */
-trait Datasources[F[_], G[_], I, C, S <: SchemaConfig] {
+trait Datasources[F[_], G[_], I, C, OldS <: SchemaConfig, S <: SchemaConfig] {
   import DatasourceError._
 
   type PathType <: ResourcePathType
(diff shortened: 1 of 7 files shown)


alissapajer:
🛑 is up because we should not merge this until the associated SDBE PR is also ready to merge.

alissapajer:
Published as `131.0.2-eb3582e`

api/src/main/scala/quasar/api/datasource/Datasources.scala
```diff
@@ -83,13 +85,19 @@ trait Datasources[F[_], G[_], I, C, S <: SchemaConfig] {
     *
     * Returns `None` if the resource exists but a schema is not available.
     */
-  def resourceSchema(
+  def oldResourceSchema(
```
wemrysi: What do you think about leaving `resourceSchema` unchanged and adding `newResourceSchema` or `resourceSchema2` instead? This would then be an additive change and reduce downstream incompatibilities.

alissapajer: My logic for this was that this PR is a breaking change either way due to the changes to `Quasar`, and this renaming will make it easier to delete the old method when we eventually do so (although that deletion PR will still be a breaking change since we're removing something). Additionally the only dependency that will be affected by this rename is SDBE, and I also already have to update that for this PR.

But I don't feel really strongly either way, so if it's more standard to be additive, I'm fine with that too.

wemrysi: That's fine, I also don't have a strong preference in this case given it is all temporary. I think the repl will also be affected by this, but it is minor in scope.

PR #4246, opened 2019-07-23
Update/specs2 core 4.6.0

rintcius:


rintcius:
Just checking how the tests go with the merge from master (in particular the updated scala-parser-combinators)

PR #4261, opened 2019-07-24
removed AsyncAtomicIndexedStore, added anti-entropy store and timesta…

cryogenian:
…mped store, made emptyStore a resource in indexed store spec, added indexed store hooks

The same as #4234, but with one commit only 

[ch8215]


build.sbt
```diff
@@ -280,6 +280,7 @@ lazy val core = project
 /** Implementations of the Quasar API. */
 lazy val impl = project
   .settings(name := "quasar-impl")
+  .settings(parallelExecution in Test := false)
```
wemrysi: Would this be necessary if we used distinct ports for the nodes in each of the suites that use `atomix`?

cryogenian: This must be _partially hardcoded_, i.e. `AtomixClusterSpec` uses `6000 + x` and `AntiEntropyStoreSpec` uses `7000 + x`

cryogenian:
@wemrysi Made ports in `AtomixSpec` and `AntiEntropyStore` being per-module unique for every node, removed `parallelExecution in Test := false`

cryogenian:
@wemrysi 

+ `random` takes two element set instead of full `peers` (it might be one element set, but this isn't a big deal I think) 
+ `use` instead of `allocated` in tests
+ Changed log message
+ Changed enqueue function a bit

cryogenian:
@wemrysi 
+ Extracted config into distinct module
+ Added `NodeAddress` and `NodeInfo.address: NodeAddress`, since we don't need ids for seeds

impl/src/main/scala/quasar/impl/cluster/Atomix.scala
```diff
@@ -92,8 +96,12 @@ object Atomix extends Logging {
     def peers: F[Set[MemberId]] =
       F.delay(service.getMembers.asScala.to[Set].map(_.id))
 
-    def random: F[Set[MemberId]] =
-      peers
+    def random: F[Set[MemberId]] = for {
+      ps <- peers
```
wemrysi: ⛳️ I think it'd be a bit more natural to convert `ps` to a `Vector` and then project indices out of it below.

impl/src/main/scala/quasar/impl/cluster/Atomix.scala
```diff
@@ -140,24 +148,32 @@ object Atomix extends Logging {
...
-            case None => F.delay {
-              log.warn(s"incorrect message received\neventType ::: ${eventType}\nsource ::: ${addr}\nbits ::: ${a.toHex}")
-            }
-            case Some(_) if size >= maxItems => F.delay {
-              log.warn(s"there are too much events in subscription queue\neventType ::: ${eventType}\nsize ::: ${size}")
-            }
-            case Some(id) => q.enqueue1((id, d.value))
+      def decodeBits(a: BitVector, cont: P => F[Unit]): F[Unit] = Codec[P].decode(a) match {
+        case Attempt.Failure(_) =>
+          F.delay(log.warn(s"malformed payload received\neventType ::: ${eventType}\nbits ::: ${a.toHex}"))
```
wemrysi: We don't need the newlines here (and elsewhere). Will also make parsing the logs a bit easier without them.

PR #4275, opened 2019-07-29
Clustering/remove max events

cryogenian:
Oops :|


djspiewak:
BTW you should probably PR from your fork, rather than from upstream.

PR #4276, opened 2019-07-30
added waitABit after every broadcast

cryogenian:


cryogenian:
3 runs are OK /cc @djspiewak @alissapajer

PR #4297, opened 2019-08-08
Remove ALL the Cords

jsantos17:
I tried to retain usage of `++` where possible. Switched to `+` where needed.


jsantos17:
[ch8542]

jsantos17:
I see. It should be easy enough to revert the last four commits. I'll close this and open another PR with a cleaner history tho.

PR #4302, opened 2019-08-14
Result push trigger

jsantos17:
[ch8142]

This is the internal implementation of result push. What's missing from this PR and will be implemented under separate tickets:

1. Limiting the number of rows
2. The actual HTTP API (in `slamdata-backend`)
3. A function/module to convert results to CSV. This PR depends on a passed function to accomplish this. I believe results are of type `Stream[F, MimirRepr]`
4. Some sort of integration tests. I have added unit tests but there are so many moving pieces I'd like to add something to `SDBESpec`.


connector/src/main/scala/quasar/connector/ResultSink.scala
```diff
@@ -16,12 +16,31 @@
...
+object ResultSink {
+  type Aux[F[_], RT0 <: ResultType[F]] = ResultSink[F] {
+    type RT = RT0
+  }
+
+  object Csv {
+    // TODO: make @unchecked unnecessary
+    def unapply[F[_]](sink: ResultSink[F]): Option[ResultSink.Aux[F, ResultType.Csv[F]]] =
+      (sink, sink.resultType) match {
+        case (rs: ResultSink.Aux[F, ResultType.Csv[F]] @unchecked, ResultType.Csv()) =>
```
jsantos17: Is there a way to get rid of this `@unchecked`? The cast should be safe since it's pattern matched in parallel with a value that ensures the type  *really* is `ResultSink.Aux[F, ResultType.Csv[F]]`

jsantos17:
Labeled as `breaking` since this changes the destination API slightly. Nobody is using that API yet, but still.

impl/src/main/scala/quasar/impl/push/DefaultResultPush.scala
```diff
@@ -0,0 +1,164 @@
...
+object DefaultResultPush {
+  def apply[F[_]: Concurrent: Timer, T, D, Q, R](
+    lookupTable: T => F[Option[TableRef[Q]]],
+    evaluator: QueryEvaluator[F, Q, R],
+    lookupDestination: D => F[Option[Destination[F]]],
+    jobManager: JobManager[F, T, Nothing],
+    convertToCsv: R => Stream[F, Byte]
+  ): F[DefaultResultPush[F, T, D, Q, R]] = {
+    for {
+      pushStatus <- Ref.of[F, Map[T, Status]](Map.empty[T, Status])
```
jsantos17: Should this be a `ConcurrentHashMap`? `Ref`s are inside an `AtomicReference` anyway.

impl/src/main/scala/quasar/impl/table/DefaultTables.scala
```diff
@@ -189,15 +187,13 @@ object DefaultTables {
   def apply[F[_]: Effect, I: Equal, Q, D, S](
       freshId: F[I],
       tableStore: IndexedStore[F, I, TableRef[Q]],
-      evaluator: QueryEvaluator[F, Q, D],
```
jsantos17: Turns out we were not using this parameter at all.

connector/src/main/scala/quasar/connector/ResultSink.scala
djspiewak: If you remove the ascription, it would work, but then you lose the typecast to the more specific parameter type. The fact that this cast is unchecked is an *accurate* warning by the compiler, so the only way around it would be to remove the cast if you can avoid it. In this case, I think it's unavoidable unless you add a liskov to the `ResultType`.

jsantos17: Will look into Liskov!

impl/src/main/scala/quasar/impl/push/DefaultResultPush.scala
wemrysi: I think `ConcurrentHashMap` will perform better under high load spread out over multiple keys whereas `Ref` will have to retry concurrent updates regardless of whether they're for the same or different keys.

That said, we'll probably need to implement against `IndexedStore` eventually in order to support clustering.

api/src/main/scala/quasar/api/push/ResultPush.scala
```diff
@@ -0,0 +1,44 @@
...
+trait ResultPush[F[_], TableId, DestinationId] {
+  import ResultPushError._
+
+  def start(tableId: TableId, destinationId: DestinationId, path: ResourcePath, format: ResultType[F], limit: Option[Long])
+      : F[Condition[ResultPushError[TableId, DestinationId]]]
+
+  def cancel(tableId: TableId)
+      : F[Condition[ExistentialError[TableId, DestinationId]]]
+
+  def status(tableId: TableId)
```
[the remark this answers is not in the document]
jsantos17: Good observation. Will bring it up with business.

jsantos17:
@wemrysi this should be ready for another look. The last problem with this PR is the use of `@unchecked` to find a `ResultSink.Aux[F, ResultType.Csv[F]]` in `NonEmptyList[ResultSink[F]]`. 

I'm convinced the cast is safe but I don't know how to convince the Scala compiler of the same.

wemrysi:
@jsantos17 Looks good, if you want to remove the `@unchecked`, one way is to add a value of type `cats.evidence.Is` to the `Csv` `ResultType` case, like so

```scala
  final case class Csv[F[_]]() extends ResultType[F] {
    type T = (List[TableColumn], Stream[F, Byte])
    val isT: Is[T, (List[TableColumn], Stream[F, Byte])] = Is.refl
  }
 ```

then when you pattern match on `c @ Csv()` you can use `c.isT` to apply the type refinement to other instances of `resultType.T`.

**Comment by jsantos17** [comment]
Unfortunately it doesn't work even with `isT`.  I'm doing this:

```scala
  object Csv {
    def unapply[F[_]](sink: ResultSink[F]): Option[ResultSink.Aux[F, ResultType.Csv[F]]] =
      sink.resultType match {
        case (c @ ResultType.Csv()) => Some(new ResultSink[F] {
          type RT = ResultType.Csv[F]
          val resultType = ResultType.Csv[F]()

          def apply(dst: ResourcePath, result: resultType.T): F[Unit] =
            (c.isT.substitute[Function2[ResourcePath, ?, F[Unit]]](sink.apply))(dst, result)
        })
        case _ => None
      }
  }
```



i.e for the Scala compiler `sink` and `sink.resultType` are not related in any way. The error is:

```
found   : (quasar.api.resource.ResourcePath, sink.resultType.T) => F[Unit]
required: (quasar.api.resource.ResourcePath, c.T) => F[Unit]
```

It's almost the same type, but not quite

wemrysi:
@jsantos17 Drat.

Do we use `ResultType` for anything else? If not, what if we just get rid of it and push the types into `ResultSink`

```scala
sealed trait ResultSink[F[_]]

object ResultSink {
  final case class Csv[F[_]](run: (ResourcePath, List[ColumnType], Stream[F, Byte]) => F[Unit]) extends ResultSink[F]
}
```

jsantos17:
We do use it elsewhere, but I can fix that. I'll push another commit eliminating the existentials. Hopefully it's not too hard.

jsantos17:
@wemrysi I have removed the type members and the build is green so this should be ready for another look.

impl/src/main/scala/quasar/impl/datasource/local/LocalDestination.scala
```diff
@@ -39,40 +37,27 @@ final class LocalDestination[F[_]: Effect: ContextShift: MonadResourceErr] priva
...
 
-object LocalDestination {
-  def apply[F[_]: Effect: ContextShift: MonadResourceErr](
-      root: JPath,
-      blockingContext: BlockingContext)
-      : LocalDestination[F] =
-    new LocalDestination[F](root, blockingContext)
-}
+  private def csvSink(root: JPath, blockingContext: BlockingContext): ResultSink[F] =
+    ResultSink.Csv[F] {
```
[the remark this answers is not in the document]
jsantos17: Interesting. I distinctly remember the Scala compiler disallowing this and suggesting a `{ case ... }` instead. Good to know! Will merge when green.

